PR #4394, opened 2023-10-02
[Project] Update in the project api of the function usages

davesh0812:
Few update in project api: 
1. `create_function` - new api, create a function object without setting it to the project.
2. `create_model_monitoring_function` - new api, create a monitoring function object without setting it to the project.
3. `set_model_monitoring_function` - new api,  update or add a monitoring function to the project (without deploy).
4. `remove_function` - will remove now also from the db.
5. `remove_model_monitoring_function`- new api, remove a monitoring function from a project and from the db.
6. `list_model_monitoring_functions` - new api, retrieve a list of all the model monitoring functions.

Also add a method  that indicates if a runtime is a model monitoring function - `is_model_monitoring_function`
 

[ML-4706](https://jira.iguazeng.com/browse/ML-4706)


assaf758 (changes requested):

mlrun/projects/project.py
```diff
@@ -1797,28 +1797,126 @@ def set_model_monitoring_application(
...
-        update or add a monitoring application to the project & deploy it.
+        update or add a monitoring function to the project.
+
         examples::
-            project.set_model_monitoring_application(application_class_name="MyApp",
-                                                 image="davesh0812/mlrun-api:1.5.0", name="myApp")
-        :param func:                    function object or spec/code url, None refers to current Notebook
-        :param name:                    name of the function (under the project), can be specified with a tag to support
+            project.set_model_monitoring_function(application_class_name="MyApp",
+                                                 image="mlrun/mlrun-api:1.5.0", name="myApp")
```
alonmr, line 1804: why are you using mlrun-api? this image should not be used by users. if you can't use mlrun/mlrun or some other user facing image, we should create a dedicated image.
also is it available in mlrun 1.5.0? because I see you set the tag to 1.5.0 but this will be released in 1.6.0

mlrun/projects/project.py
```diff
@@ -1797,28 +1797,126 @@ def set_model_monitoring_application(
...
+        :param requirements_file:       Path to a python requirements file
         :param application_class:       Name or an Instance of a class that implementing the monitoring application.
         :param application_kwargs:      Additional keyword arguments to be passed to the
                                         monitoring application's constructor.
         """
 
+        function_object: RemoteRuntime = None
+        (
+            resolved_function_name,
+            function_object,
```
alonmr, line 1827:
```suggestion
        (
            resolved_function_name,
            function_object: RemoteRuntime,
```

mlrun/projects/project.py
```diff
@@ -1797,28 +1797,126 @@ def set_model_monitoring_application(
...
+
+        return function_object
+
+    def create_model_monitoring_function(
+        self,
+        func: str = None,
+        application_class: typing.Union[str, ModelMonitoringApplication] = None,
+        name: str = None,
+        image: str = None,
+        handler=None,
```
alonmr, line 1862: type hint?

mlrun/projects/project.py
```diff
@@ -1797,28 +1797,126 @@ def set_model_monitoring_application(
         **application_kwargs,
     ) -> mlrun.runtimes.BaseRuntime:
         """
-        update or add a monitoring application to the project & deploy it.
+        update or add a monitoring function to the project.
```
alonmr, line 1800: Since you have taken initiative to use capital letters
```suggestion
        Update or add a monitoring function to the project.
```

mlrun/projects/project.py
```diff
@@ -1797,28 +1797,126 @@ def set_model_monitoring_application(
...
+        image: str = None,
+        handler=None,
+        with_repo: bool = None,
+        tag: str = None,
+        requirements: typing.Union[str, typing.List[str]] = None,
+        requirements_file: str = "",
+        **application_kwargs,
+    ) -> mlrun.runtimes.BaseRuntime:
+        """
+        create a monitoring function object without setting it to the project
```
alonmr, line 1870: Why is this needed? what's the use case? Add it to the description?
```suggestion
        Create a monitoring function object without setting it to the project
```

mlrun/projects/project.py
```diff
@@ -1797,28 +1797,126 @@ def set_model_monitoring_application(
...
+
+        examples::
+            project.create_model_monitoring_function(application_class_name="MyApp",
+                                                 image="mlrun/mlrun-api:1.5.0", name="myApp")
```
alonmr, line 1874: same here with the image its internal

mlrun/projects/project.py
```diff
@@ -1797,28 +1797,126 @@ def set_model_monitoring_application(
...
+            handler,
+            with_repo,
+            tag,
+            requirements,
+            requirements_file,
+            **application_kwargs,
+        )
+        return function_object
+
+    def _create_model_monitoring_function_helper(
```
alonmr, line 1907: or give it some other indicative name
```suggestion
    def _instantiate_model_monitoring_function(
```

mlrun/projects/project.py
```diff
@@ -1797,28 +1797,126 @@ def set_model_monitoring_application(
...
+        self,
+        func: typing.Union[str, mlrun.runtimes.BaseRuntime] = None,
+        application_class: typing.Union[str, ModelMonitoringApplication] = None,
+        name: str = None,
+        image: str = None,
+        handler=None,
```
alonmr, line 1913: type hint

mlrun/projects/project.py
```diff
@@ -1947,10 +2048,80 @@ def set_function(
...
+        self._set_function_helper(resolved_function_name, tag, function_object, func)
+        return function_object
+
+    def create_function(
+        self,
+        func: str = None,
+        name: str = "",
+        kind: str = "",
+        image: str = None,
+        handler=None,
```
alonmr, line 2060: type

mlrun/projects/project.py
```diff
@@ -1947,10 +2048,80 @@ def set_function(
...
+        with_repo: bool = None,
+        tag: str = None,
+        requirements: typing.Union[str, typing.List[str]] = None,
+        requirements_file: str = "",
+    ):
+        """
+        create a function object without setting it to the project
```
alonmr, line 2067:
```suggestion
        Create a function object without setting it to the project
```

mlrun/projects/project.py
```diff
@@ -1947,10 +2048,80 @@ def set_function(
...
+            proj.create_function('./src/mycode.py', 'ingest',
+                              image='myrepo/ing:latest', with_repo=True)
+            proj.create_function('http://.../mynb.ipynb', 'train')
+            proj.create_function('./func.yaml')
+            proj.create_function('hub://get_toy_data', 'getdata')
+
+            # set function requirements
+
+            # by providing a list of packages
+            proj.set_function('my.py', requirements=["requests", "pandas"])
```
alonmr, line 2087:
```suggestion
            proj.create_function('my.py', requirements=["requests", "pandas"])
```

mlrun/projects/project.py
```diff
@@ -1947,10 +2048,80 @@ def set_function(
...
+            handler,
+            with_repo,
+            tag,
+            requirements,
+            requirements_file,
+        )
         return function_object
 
-    def _resolved_function(
+    def _create_function_helper(
```
alonmr, line 2124:
```suggestion
    def _instantiate_function(
```

mlrun/projects/project.py
```diff
@@ -2048,7 +2219,7 @@ def _resolved_function(
             func,
         )
 
-    def _set_function(
+    def _set_function_helper(
```
alonmr, line 2222: I think the previous name is fine
```suggestion
    def _set_function(
```

mlrun/projects/project.py
```diff
@@ -2064,28 +2235,28 @@ def _set_function(
         self.spec.set_function(name, function_object, func)
 
     def remove_function(self, name):
-        """remove a function from a project
+        """remove a function from a project and from the db.
```
alonmr, line 2238: this is not backwards compatible.
You can add a flag but default must be false.

alonmr (changes requested):
Some minor stuff

mlrun/projects/project.py
davesh0812: @alonmr ,Are you sure it is a valid syntax?

mlrun/projects/project.py
davesh0812: OK but the problem is that if I removing function and when run `list_functions` I will still get the removed function because we returning the list from the db.

@alonmr

PR #4396, opened 2023-10-02
[Spark] Fix `toPandas` to match pyspark 3.2.3

gtopper:
Fixes
```
ImportError: cannot import name 'DayTimeIntervalType' from 'pyspark.sql.types' (/spark/python/lib/pyspark.zip/pyspark/sql/types.py)
```
in `TestFeatureStoreSparkEngine.test_ingest_with_steps_mapvalues`.

### DIFF
--- a/mlrun/data_types/to_pandas.py
+++ b/mlrun/data_types/to_pandas.py
@@ -12,15 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import warnings
 from collections import Counter
-from typing import List, Optional, Type
-from warnings import catch_warnings, simplefilter, warn
 
 from pyspark.sql.types import (
     BooleanType,
     ByteType,
-    DataType,
-    DayTimeIntervalType,
     DoubleType,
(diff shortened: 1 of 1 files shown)


mlrun/data_types/to_pandas.py
```diff
@@ -145,18 +137,7 @@ def toPandas(spark_df):
...
-                        corrected_panda_types[tmp_column_names[index]] = (
-                            np.object0 if pandas_type is None else pandas_type
-                        )
-
-                    pdf = pd.DataFrame(columns=tmp_column_names).astype(
-                        dtype=corrected_panda_types
-                    )
-                    pdf.columns = spark_df.columns
-                    return pdf
+                    return pd.DataFrame.from_records([], columns=spark_df.columns)
```
tomerm-iguazio, line 140: Regardless of whether the DataFrame is empty or not, ensure that the column types are preserved.

tomerm-iguazio (changes requested):

PR #4401, opened 2023-10-02
[Workflows] Minor tweaks

liranbg:
added reference to remote and remote:local in docstring,  fixed warning not raised for remote:local and when no engine but schedule (which defaults to remote) is selected.

https://jira.iguazeng.com/browse/ML-4721
https://jira.iguazeng.com/browse/ML-4722

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -1265,7 +1265,7 @@ class MlrunProject(ModelObj):
         :param name:          name of the workflow
         :param workflow_path: url/path for the workflow file
         :param embed:         add the workflow code into the project.yaml
-        :param engine:        workflow processing engine ("kfp" or "local")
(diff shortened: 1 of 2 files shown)


mlrun/projects/project.py
```diff
@@ -1265,7 +1265,7 @@ def set_workflow(
         :param name:          name of the workflow
         :param workflow_path: url/path for the workflow file
         :param embed:         add the workflow code into the project.yaml
-        :param engine:        workflow processing engine ("kfp" or "local")
+        :param engine:        workflow processing engine ("kfp", "local", "remote" or "remote:local")
```
alonmr, line 1268: can add that remote is essentially `remote:kfp`

alonmr (approved):

PR #4412, opened 2023-10-04
[Model Monitoring] Create new run db instance for each new model monitoring feature set [1.5.x]

davesh0812:
[ML-4544](https://jira.iguazeng.com/browse/ML-4544)


alonmr (commented):
We need some more fixes wrt feature set run db.
Please see https://github.com/mlrun/mlrun/pull/3972/files#diff-d5d4266b1f536331152555622d1c79211619488f5630d18a9125437a9939e33a
I mistakenly removed it because the comment above was confusing and incorrect.
Please return it but notice that you can't import api in mlrun client code so you need to get the run db instance as you did here in the ingestion job as well.
Hopefully this is the only place this is missing.

PR #4418, opened 2023-10-04
[DB] Fix submit job not using specific db session

liranbg:
Fix submit function from hub due to reused session that have race condition between others using it (singleton-issue). injecting db session given by fastapi instead solves it

https://jira.iguazeng.com/browse/ML-4733

### DIFF
--- a/mlrun/api/api/utils.py
+++ b/mlrun/api/api/utils.py
@@ -46,8 +46,14 @@ from mlrun.api.utils.singletons.logs_dir import get_logs_dir
 from mlrun.api.utils.singletons.scheduler import get_scheduler
 from mlrun.common.helpers import parse_versioned_object_uri
 from mlrun.config import config
+from mlrun.db import RunDBInterface
 from mlrun.errors import err_to_str
-from mlrun.run import import_function, new_function
+from mlrun.run import (
+    compile_imported_function,
+    extend_hub_uri_if_needed,
+    import_function_to_dict,
+    new_function,
+)
 from mlrun.runtimes.utils import enrich_function_from_dict
 from mlrun.utils import get_in, logger
 
@@ -135,7 +141,7 @@ def get_secrets(
 
 def get_run_db_instance(
     db_session: Session,
-):
+) -> RunDBInterface:
     # TODO: getting the run db should be done seamlessly by the run db factory and not require this logic to
     #  inject the session
     db = get_db()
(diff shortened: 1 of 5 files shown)


mlrun/run.py
```diff
@@ -521,9 +528,8 @@ def import_function_to_dict(url, secrets=None):
                 raise ValueError("exec path (spec.command) must be relative")
             url = url[: url.rfind("/") + 1] + code_file
             code = get_object(url, secrets)
-            dir = path.dirname(code_file)
-            if dir:
-                makedirs(dir, exist_ok=True)
+            if path.dirname(code_file):
+                makedirs(path.dirname(code_file), exist_ok=True)
```
alonmr, line 532:
```suggestion
            if code_dir := path.dirname(code_file):
                makedirs(code_dir, exist_ok=True)
```

mlrun/api/api/utils.py
```diff
@@ -168,8 +174,23 @@ def _generate_function_and_task_from_submit_run_body(db_session: Session, data):
         function = new_function(runtime=function_dict)
     else:
         if "://" in function_url:
-            function = import_function(
-                url=function_url, project=task.get("metadata", {}).get("project")
+            project_name = task.get("metadata", {}).get("project")
+            hub_function = function_url.startswith(mlrun.utils.hub_prefix)
```
alonmr, line 178: Can be false as it will be checked in the else clause
```suggestion
            hub_function = False
```

mlrun/api/api/utils.py
```diff
@@ -168,8 +174,23 @@ def _generate_function_and_task_from_submit_run_body(db_session: Session, data):
...
+            db = mlrun.api.api.utils.get_run_db_instance(db_session)
+            if function_url.startswith(mlrun.utils.db_prefix):
+                url = function_url.removeprefix(mlrun.utils.db_prefix)
+                _project_name, name, tag, hash_key = parse_versioned_object_uri(url)
+                runtime = db.get_function(name, _project_name, tag, hash_key)
+                if not runtime:
+                    raise mlrun.errors.MLRunNotFoundError(
+                        f"function {name}:{tag} not found in the DB",
```
alonmr, line 186: I think we should make it a convention to use capital letters in logs and exceptions
```suggestion
                        f"Function {name}:{tag} not found in the DB",
```

alonmr (approved):
Some minor comments

PR #4434, opened 2023-10-05
[Builder] Clone GitHub repository to a directory other than `/tmp`

gtopper:
[ML-4741](https://jira.iguazeng.com/browse/ML-4741)


gtopper:
> Can you explain the original issue? I don't think this will work in open source or with security context other than iguazio user..

@alonmr, please check out the RCA in the ticket and let me know if you have a suggestion. I'll try to rewrite it on Sunday to avoid assuming that the user is iguazio.

PR #4436, opened 2023-10-05
[Model Monitoring] Pass start time + current_stats to the applications and the writer & Change partitioning order.

davesh0812:


jond01:
Superseded by #4624

PR #4444, opened 2023-10-09
[Datastore] Fix reading parquet directory when using datastore profiles

alxtkr77:
https://jira.iguazeng.com/browse/ML-4751
When a datastore profile directory (as opposed to a single file) is utilized to read a pandas dataframe, the appropriate filesystem is already determined and set up. However, it remains unused. This is because, while pandas attempts to create a new instance of the filesystem, mlrun doesn't forward the filesystem to pandas. As a result, pandas encounters an error due to its unfamiliarity with the 'ds://' schema. This PR rectifies this by providing the pre-established filesystem to the pandas parquet reader and removing the 'ds://prfofile' prefix from the directory URL

### DIFF
--- a/mlrun/datastore/base.py
+++ b/mlrun/datastore/base.py
@@ -285,7 +285,14 @@ class DataStore:
         if file_system:
             if self.supports_isdir() and file_system.isdir(file_url) or df_module == dd:
                 storage_options = self.get_storage_options()
-                if storage_options:
+                if url.startswith("ds://"):
+                    parsed_url = urllib.parse.urlparse(url)
+                    url = parsed_url.path[1:]
+                    # Pass the underlying file system
+                    kwargs["filesystem"] = file_system
+                elif storage_options:
+                    # If it's a datastore profile,
+                    # the storage_options are already incorporated into the filesystem object.
                     kwargs["storage_options"] = storage_options
                 df = reader(url, **kwargs)
             else:
--- a/mlrun/datastore/base.py
+++ b/mlrun/datastore/base.py
@@ -291,8 +291,6 @@ class DataStore:
                     # Pass the underlying file system
                     kwargs["filesystem"] = file_system
                 elif storage_options:
-                    # If it's a datastore profile,
-                    # the storage_options are already incorporated into the filesystem object.
                     kwargs["storage_options"] = storage_options
                 df = reader(url, **kwargs)
             else:
--- a/tests/integration/aws_s3/test_aws_s3.py
+++ b/tests/integration/aws_s3/test_aws_s3.py
@@ -205,6 +205,8 @@ class TestAwsS3:
 
     def test_directory(self, use_datastore_profile):
         param = self.s3["ds"] if use_datastore_profile else self.s3["s3"]
+        for p in credential_params:
+            os.environ[p] = config["env"][p]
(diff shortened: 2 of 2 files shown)


tests/integration/aws_s3/test_aws_s3.py
```diff
@@ -198,3 +202,35 @@ def test_using_profile(self, use_datastore_profile):
...
+        data1 = {"Column1": [1, 2, 3], "Column2": ["A", "B", "C"]}
+
+        # Define data for the second DataFrame
+        data2 = {"Column1": [4, 5, 6], "Column2": ["X", "Y", "Z"]}
+
+        # Create the DataFrames
+        df1 = pd.DataFrame(data1)
+        df2 = pd.DataFrame(data2)
+        temp_file1 = tempfile.mktemp(suffix=".parquet")
+        temp_file2 = tempfile.mktemp(suffix=".parquet")
```
tomerm-iguazio, line 221: Please use with statement, in order to delete the file after the test.

tomerm-iguazio: with tempfile.NamedTemporaryFile(suffix='.parquet', delete=False) as temp_file1, \
                tempfile.NamedTemporaryFile(suffix='.parquet', delete=False) as temp_file2:

tomerm-iguazio: with tempfile.NamedTemporaryFile(suffix='.parquet', delete=False) as temp_file1, \
                tempfile.NamedTemporaryFile(suffix='.parquet', delete=False) as temp_file2:

alxtkr77: Why to use  tempfile.NamedTemporaryFile(...) ? I need only name, not opening the file

tests/integration/aws_s3/test_aws_s3.py
```diff
@@ -198,3 +202,35 @@ def test_using_profile(self, use_datastore_profile):
         # cleanup
         for param in params:
             os.environ.pop(param)
+
+    def test_directory(self, use_datastore_profile):
+        param = self.s3["ds"] if use_datastore_profile else self.s3["s3"]
```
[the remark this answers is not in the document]
alxtkr77: Fixed

mlrun/datastore/base.py
```diff
@@ -285,7 +285,14 @@ def reader(*args, **kwargs):
         if file_system:
             if self.supports_isdir() and file_system.isdir(file_url) or df_module == dd:
                 storage_options = self.get_storage_options()
-                if storage_options:
+                if url.startswith("ds://"):
```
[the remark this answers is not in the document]
assaf758: Good observation Tomer, however not that the data-store profile mechanism is intended to be used as part of mlrun abstractions (source, target, data-item etc) - and not to be used as regular new "file-system" with direct pandas API.
The later would require another approach entirely as data-store profiles assume access to mlrun secrets etc.

tests/integration/aws_s3/test_aws_s3.py
tomerm-iguazio: It is better to delete the file after the test.

tomerm-iguazio (approved):
looking good!

PR #4466, opened 2023-10-18
[Pipelines] Raise get pipeline timeout and add logs

alonmr:
Summary formatting is not allowed with list_pipelines so no need to worry about spamming logs


theSaarco (approved):
Looks good. My concern is around the time it takes to `list_runs()` so the logs should cover that area.

PR #4474, opened 2023-10-19
[Log collector] Update last log time every 4KB read

alonmr:
https://jira.iguazeng.com/browse/ML-4791


theSaarco (approved):
Looks good. Maybe to avoid doubts on negative numbers you can have the bytes counter start at 0, and add the number of bytes read, and perform the update if it's >= `common.BytesReadPerLogTimeUpdate`. It's also using the constant just once, rather than twice.
Not mandatory, just a suggestion.

assaf758 (approved):

PR #4488, opened 2023-10-24
[Tests] Minor TODOs and fixes

alonmr:


guy1992l (changes requested):
A side-note: Logging a metric means to log results in an iterative way with respect to the time of logging. It is a valuable feature that is sad we don't have.

PR #4489, opened 2023-10-25
[Debug] Introducing remote debugging available

liranbg:
Currently supports pycharm & remote workflows but that should not be limited to that only.
Explanation is provided on `CONTRIBUTING.md` file


CONTRIBUTING.md
```diff
@@ -263,3 +263,86 @@ pyenv virtualenv mlrun
...
+project = mlrun.load_project(
+    name="load-project-0",
+    context="./load-project-0",
+    url="git://github.com/someone/somewhere.git",
+    clone=True,
+)
+project.set_workflow(
+      "kfpmain",
+      "workflow.py",
+      schedule="*/5 * * * *",
```
alonmr, line 338: is shcedule related?

CONTRIBUTING.md
```diff
@@ -263,3 +263,86 @@ pyenv virtualenv mlrun
...
+      engine="kfp",
+      image="mlrun/mlrun:unstable-dev",
+      remote_debugging={
+          "mode": "pycharm",
+          "port": 17714,
+          "host": "7.tcp.eu.ngrok.io",
```
alonmr, line 344:
```suggestion
          "port": <ngrok-port>
          "host": <ngrok-host>,
```

CONTRIBUTING.md
```diff
@@ -263,3 +263,86 @@ pyenv virtualenv mlrun
...
+      },
+)
+project.run("kfpmain", watch=False, schedule=True)
```
alonmr, line 347: unless its needed of course
```suggestion
project.run("kfpmain", watch=False)
```

server/api/api/endpoints/workflows.py
```diff
@@ -193,6 +193,7 @@ async def submit_workflow(
                 workflow_request=updated_request,
                 db_session=db_session,
                 auth_info=auth_info,
+                remote_debugging=workflow_request.spec.remote_debugging,
```
alonmr, line 196: I believe you can get it from the updated_request instead of adding another param

alonmr (commented):
Awesome stuff!
I don't think you need to pass the param in most of these methods since you pass the workflow spec as well and its already there

CONTRIBUTING.md
liranbg: makes it a remote scheduled (thus using remote)

theSaarco: You mean remote execution of the workflow? You can do the same with the `engine` parameter, setting it to `remote` (or the always confusing `remote:local` value).

PR #4490, opened 2023-10-25
[Model Monitoring] Fix writer pipelines v3io access key

Eyal-Danieli:
Replace user V3IO access key with Pipelines access key when generating V3IO clients through the writer entity. User V3IO access key has no permission to the writer container and therefore it can't write records to kv/tsdb targets.

### DIFF
--- a/mlrun/model_monitoring/writer.py
+++ b/mlrun/model_monitoring/writer.py
@@ -117,14 +117,10 @@ class ModelMonitoringWriter(StepToDict):
         self._create_tsdb_table()
 
     @staticmethod
-    def get_v3io_container(project_name: str) -> str:
-        return f"users/pipelines/{project_name}/monitoring-apps"
-
-    @staticmethod
     def _get_v3io_client() -> V3IOClient:
         return mlrun.utils.v3io_clients.get_v3io_client(
             endpoint=mlrun.mlconf.v3io_api,
-            access_key=os.getenv(ProjectSecretKeys.ACCESS_KEY),
+            access_key=os.getenv(ProjectSecretKeys.PIPELINES_ACCESS_KEY),
         )
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


mlrun/model_monitoring/writer.py
```diff
@@ -124,15 +124,15 @@ def get_v3io_container(project_name: str) -> str:
...
+            access_key=os.getenv(ProjectSecretKeys.PIPELINES_ACCESS_KEY),
         )
 
     @staticmethod
     def _get_v3io_frames_client(v3io_container: str) -> V3IOFramesClient:
         return mlrun.utils.v3io_clients.get_frames_client(
             address=mlrun.mlconf.v3io_framesd,
             container=v3io_container,
-            token=os.getenv(ProjectSecretKeys.ACCESS_KEY, ""),
+            token=os.getenv(ProjectSecretKeys.PIPELINES_ACCESS_KEY, ""),
```
jond01, line 135: From an active deployment:
```sh
k -n default-tenant exec -it nuclio-test-monitoring-app-flow29-model-monitoring-writer-wrknc -- \
printenv | grep ACCESS
```
results in
```txt
MODEL_MONITORING_ACCESS_KEY=***
V3IO_ACCESS_KEY=***
```
`MODEL_MONITORING_PIPELINES_ACCESS_KEY` is not even set in this writer pod. Is it set in your environment?
What error did you face?

jond01: Updating here following our conversation today:
1. `ProjectSecretKeys.PIPELINES_ACCESS_KEY` (= `"MODEL_MONITORING_PIPELINES_ACCESS_KEY"`) indeed doesn't exist in the writer pod
2. The suggestion returns `None`
3. `"V3IO_ACCESS_KEY"` is used as a fallback
4. The saving process works (at least for you)

I need to re-check which one of the two access keys is the right one to use here.

jond01:
It was addressed in #4495 (but still doesn't work for me with the change).

PR #4491, opened 2023-10-25
[ImageBuilder] Fix config with requirements file

alonmr:
https://jira.iguazeng.com/browse/ML-4844


laurybueno (approved):
thanks for the great work, man!

PR #4497, opened 2023-10-29
[MLRun] Extend monitoring capabilities phase 1

alonmr:
## Extending monitoring capabilities with state thresholds
The thresholds represent time limits for the state of pods. If these time limits are exceeded, the pod will be terminated, and the run state will transition to an error/failed state. For example, if a threshold of 1 hour is set to `pending_scheduled` state, and the pod is pending and scheduled after 1 hour, it will be terminated. 

The state thresholds can be set with the runtimes SDK `function.set_state_thresholds({...})` or in a parameter for the `run` method for a single run `function.run(state_thresholds={..})`.
The threshold should be a time string that conforms to timelength package https://github.com/Etorix0005/timelength. For example, to set a threshold of 10 hours, use the string `"10h"`. If not threshold is set, the default configured values will be used. To disable the state threshold, set it to `"-1"`.

The supported states are:
1. pending_scheduled 📆 
2. pending_not_scheduled 🚫 
3. running 🏃 
4. image_pull_backoff 🌐 

In phase 1 there are no actions taken against the runs, just resolving the states

Jira - [ML-3379](https://jira.iguazeng.com/browse/ML-3379)


tests/system/runtimes/test_mpijob.py
```diff
@@ -36,7 +36,7 @@ def test_mpijob_run(self):
             handler="handler",
             project=self.project_name,
             filename=code_path,
-            image="mlrun/mlrun",
+            image="mlrun/blabla",
```
liranbg, line 39: ?

liranbg (changes requested):
minor comments below
overall looking very well!

server/api/runtime_handlers/base.py
```diff
@@ -213,6 +214,14 @@ def monitor_runs(self, db: DBInterface, db_session: Session):
                     exc=err_to_str(exc),
                     traceback=traceback.format_exc(),
                 )
+
+        self._terminate_resource_less_runs(
+            db, db_session, project_run_uid_map, run_runtime_resources_map
+        )
+
+    def _terminate_resource_less_runs(
```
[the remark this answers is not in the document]
alonmr: pycharm complains that `resourceless` is not a word but I guess I'll use it anyway

mlrun/utils/regex.py
```diff
@@ -88,3 +88,7 @@
 # must be alphanumeric or _
 # max 256 length
 v3io_stream_consumer_group = [r"^(?!_)[a-zA-Z0-9_]{1,256}$"]
+
+# A time regex which is used to parse a time string and get the time and the unit
+# e.g. 10s -> 10, s
+time_pattern = r"^(\d+)([smhdw])$"
```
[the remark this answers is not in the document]
alonmr: done

mlrun/runtimes/constants.py
```diff
@@ -53,6 +56,56 @@ def pod_phase_to_run_state(pod_phase):
         }[pod_phase]
 
 
+class ThresholdStates:
+    pending_scheduled = "pending_scheduled"
+    pending_not_scheduled = "pending_not_scheduled"
+    running = "running"
+    image_pull_backoff = "image_pull_backoff"
```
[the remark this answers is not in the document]
alonmr: added in comment

server/api/runtime_handlers/base.py
```diff
@@ -1217,29 +1231,20 @@ def _resolve_resource_state_and_apply_threshold(
...
                     start_time=str(start_time),
                     namespace=namespace,
                 )
-                try:
-                    server.api.utils.singletons.k8s.get_k8s_helper().delete_pod(
-                        runtime_resource_name, namespace, self.pod_grace_period_seconds
-                    )
-                except Exception as exc:
-                    logger.warning(
-                        "Failed to kill pod",
```
liranbg: why not leaving it here until next pr?

tests/api/runtime_handlers/test_kubejob.py
```diff
@@ -578,7 +578,7 @@ async def test_state_thresholds_defaults(self, db: Session, client: TestClient):
         pending_scheduled_pod.status.conditions = [
             k8s_client.V1PodCondition(type="PodScheduled", status="True")
         ]
-        pending_scheduled_pod.status.start_time = datetime.datetime.utcnow() - timedelta(
+        pending_scheduled_pod.status.start_time = datetime.datetime.now() - timedelta(
```
liranbg, line 581: why changing?

liranbg, line 581: why changing?

liranbg (commented):
minor question

tests/api/runtime_handlers/test_kubejob.py
alonmr: had some issues with utcnow. changed to `datetime.now(timezone.utc)` fixed it.

PR #4498, opened 2023-10-30
[Model Monitoring] Add an evidently app to the system test

jond01:
Add a new application with [evidently](https://github.com/evidentlyai/evidently) logic (and package) in addition to the dummy application.

It passes locally.

This PR continues the work on [ML-4762](https://jira.iguazeng.com/browse/ML-4762),  https://github.com/mlrun/mlrun/pull/4478.


assaf758 (approved):
Very nice! :-)
Not blocking for now but we should consider if we can improve the testing somewhat:
- asserting  the kv/tsdb content values
- some checks on the evidently artifacts

PR #4501, opened 2023-10-30
[Workflows] Fail if name handler or path were not given

yanburman:
instead of running arbitrary workflow, force user to provide which workflow to run
https://jira.iguazeng.com/browse/ML-3169


liranbg:
@yanburman  mind the pr title convention ^

PR #4503, opened 2023-10-30
[Project] Restore init function in spec for BC [1.5.x]

alonmr:
We encountered an issue with 1.5.0 where users would sync functions of a project that had a function in an old format of:
~~~
{"name":...
"spec":{
"metadata":{...},
"spec":{...},
}}
~~~
This format can no longer be generated but we still need to support this as we did not migrate it.
This was removed in https://github.com/mlrun/mlrun/pull/4271/files#diff-c972cfb48585c0ee85184b83808a9bf592a0ddd60a07cba789023649354f6cb8L3148

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -3334,7 +3334,11 @@ def _init_function_from_dict(
     url, in_context = project.get_item_absolute_path(url)
 
     if "spec" in f:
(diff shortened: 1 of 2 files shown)


gtopper (approved):
LGTM ✅

PR #4504, opened 2023-10-30
[Project] Deprecate project API `build_image.skip_deployed` parameter [1.5.x]

theSaarco:
The `skip_deployed` parameter was useless anyway since a new function was always created, therefore a build was always triggered (as there was no image in the function status). Since the API is called `build_image`, it makes sense for the user to expect a build will happen once triggered.

See [ML-4865](https://jira.iguazeng.com/browse/ML-4865)


alonmr (approved):

liranbg:
@theSaarco 
cherry-pick to development

PR #4521, opened 2023-11-02
[Datastore] Refactor store profile resolution

alxtkr77:
This pull request updates the mechanism for resolving datastore profiles within datastores. At present, each store individually managed datastore profile resolution, which complicated the process of integrating a new store to support datastore profiles. With this pull request, the responsibility for datastore profile resolution is elevated to a higher level, allowing credentials to be transmitted to the stores via 'secrets'.


mlrun/datastore/redis.py
```diff
@@ -34,27 +35,25 @@ def __init__(self, parent, schema, name, endpoint="", secrets: dict = None):
...
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "Provide Redis username and password only via secrets"
+            )
+        user = self._get_secret_or_env("REDIS_USER", "")
+        password = self._get_secret_or_env("REDIS_PASSWORD", "")
+        host = parsed_endpoint.hostname
+        port = parsed_endpoint.port if parsed_endpoint.port else "6379"
+        schema = parsed_endpoint.scheme
+        if user or password:
+            endpoint = f"{schema}://{user}:{password}@{host}:{port}"
```
theSaarco, line 53: I'm not sure I follow this solution - so you're encoding the secrets in the URL after all? Is this due to the path being passed to storey and other places? Let's discuss this, to make sure I understand this.

mlrun/datastore/redis.py
```diff
@@ -34,27 +35,25 @@ def __init__(self, parent, schema, name, endpoint="", secrets: dict = None):
...
+        self.secure = parsed_endpoint.scheme == "rediss"
+
+        if parsed_endpoint.username or parsed_endpoint.password:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "Provide Redis username and password only via secrets"
+            )
+        user = self._get_secret_or_env("REDIS_USER", "")
+        password = self._get_secret_or_env("REDIS_PASSWORD", "")
+        host = parsed_endpoint.hostname
+        port = parsed_endpoint.port if parsed_endpoint.port else "6379"
```
theSaarco, line 50: Let's move `6379` to be a config parameter - we want to avoid such numeric constants in the code.

mlrun/datastore/base.py
```diff
@@ -699,3 +700,26 @@ def _validate_https_token(self):
...
+    def __init__(self, cls, *args, **kwargs):
+        if not issubclass(cls, fsspec.AbstractFileSystem):
+            raise ValueError("Class must be a subclass of fsspec.AbstractFileSystem")
+
+        class DsSanitizerTemp(cls):
+            @classmethod
+            def _strip_protocol(cls, url):
+                if url.startswith("ds://"):
+                    parsed_url = urlparse(url)
+                    url = parsed_url.path[1:]
```
theSaarco, line 715: Are we expecting a situation where we have something like `ds://` here at all? IIUC, this is created by the stores themselves, which should already get the url without a `ds://` prefix. Maybe I'm missing something.
Also, why calling `super()._strip_protocol()` as well? Are paths like `ds://s3://...` possible?

theSaarco (changes requested):
We're definitely on the right track here - I'm not sure about the logic behind the `redis` changes (I'm not sure how this fits what we've discussed, I'm probably missing it), let's discuss this.

mlrun/datastore/datastore.py
```diff
@@ -199,7 +201,13 @@ def get_or_create_store(self, url, secrets: dict = None) -> (DataStore, str):
                     secrets=self._secrets
                 ).get_datastore_profile(profile_name, project_name)
 
-            datastore_type = datastore.type
+            if secrets and datastore.secrets():
+                secrets = merge(secrets, datastore.secrets())
```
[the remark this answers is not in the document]
alxtkr77: Hmm, in case of Kafka it can be nested dictionary. Let's talk about it

mlrun/datastore/base.py
```diff
@@ -699,3 +700,26 @@ def _validate_https_token(self):
                 f"A AUTH TOKEN should not be provided while using {self._schema} "
                 f"schema as it is not secure and is not recommended."
             )
+
+
+class DsSanitizer:
```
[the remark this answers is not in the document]
alxtkr77: I mean 'ds://path/to' -> 'path/to'. Literally 'ds' prefix removal . Any suggestions for the name?

mlrun/datastore/base.py
alxtkr77: Yes: 
ingest()->featureset.purge_targets()-> driver.purge()->self._get_store().rm(self.get_target_path(), recursive=True)

theSaarco (approved):
Look good, per our discussion. Would still like to move the `6379` to be a defined constant (it really hurts my eyes as-is).
Also, please add a description to the PR prior to merging.

PR #4531, opened 2023-11-05
[LogCollector] Fix get_log size calculation

liranbg:
seems like the size and offset calculation were a bit off wrt calculating the chunk size to read
add some UT to cover its usages

https://jira.iguazeng.com/browse/ML-4747

### DIFF
--- a/server/log-collector/pkg/services/logcollector/server.go
+++ b/server/log-collector/pkg/services/logcollector/server.go
@@ -354,15 +354,10 @@ func (s *Server) GetLogs(request *protologcollector.GetLogsRequest, responseStre
 		return errors.Wrapf(err, "Failed to get log file size for run id %s", request.RunUID)
 	}
 
-	// if size < 0 - we read only the logs we have for this moment in time starting from offset, so GetLogs will be finite.
-	// otherwise, we read only the request size from the offset
-	endSize := currentLogFileSize - request.Offset
-	if request.Size > 0 && endSize > request.Size {
-		endSize = request.Size
-	}
-
 	// if the offset is bigger than the current log file size, return empty response
-	if endSize <= 0 {
+	// this happens when client is ready for next iteration while server did not collect the logs just yet
+	// we send empty response to ensure client will not get stuck and keep retrying
+	if currentLogFileSize-request.Offset <= 0 {
 		if err := responseStream.Send(&protologcollector.GetLogsResponse{
 			Success: true,
 			Logs:    []byte{},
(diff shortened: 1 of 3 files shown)


server/log-collector/pkg/services/logcollector/server.go
```diff
@@ -1055,22 +1066,29 @@ func (s *Server) isLogCollectionRunning(ctx context.Context, runUID, project str
...
-	// if the remaining size is smaller than the buffer size, use the remaining size
-	if endSize-currentOffset < chunkSize {
-		chunkSize = endSize - currentOffset
+	// if we didn't read anything yet, assume read the initial offset
+	if totalLogsSize == 0 {
+		totalLogsSize += initialOffset
 	}
 
-	return chunkSize
+	// if the size we need to read is bigger that buffer, use the buffer size
```
TomerShor, line 1085:
```suggestion
	// if the size we need to read is bigger than the buffer, use the buffer size
```

server/log-collector/pkg/services/logcollector/server.go
```diff
@@ -372,14 +367,27 @@ func (s *Server) GetLogs(request *protologcollector.GetLogsRequest, responseStre
 		return nil
 	}
 
+	var endSize = currentLogFileSize
+
+	// if request size is positive, read only the requested size, add the offset to ensure we read
+	// all requested logs
+	if request.Size > 0 {
+		endSize = request.Size + request.Offset
+	}
```
TomerShor, line 376: not sure I understand why you add the offset here to the size - we use the offset in `readAt` later on, so the size of the read logs doesn't include the offset bytes...

server/log-collector/pkg/services/logcollector/server.go
```diff
@@ -1055,22 +1066,29 @@ func (s *Server) isLogCollectionRunning(ctx context.Context, runUID, project str
...
+	if totalLogsSize+initialOffset >= endSize {
+		return 0
 	}
 
-	// if the remaining size is smaller than the buffer size, use the remaining size
-	if endSize-currentOffset < chunkSize {
-		chunkSize = endSize - currentOffset
+	// if we didn't read anything yet, assume read the initial offset
+	if totalLogsSize == 0 {
+		totalLogsSize += initialOffset
```
TomerShor, line 1082: same here, relates to the comment above

TomerShor (commented):
Grammar nazi and a question

server/log-collector/pkg/services/logcollector/server.go
liranbg: u start to read at offset, and you read until size+offset
if you dont add offset, then you start to read at offset and read until size
so basically, for the latter, if you want to read 100 bytes with offset 10, you essentially read 90
see that getChunkSize uses the endsize to calculate the leftToRead in a new way

server/log-collector/pkg/services/logcollector/server.go
liranbg: resolves it as the answer relates to https://github.com/mlrun/mlrun/pull/4531#discussion_r1383374543

quaark (approved):
Good catch!
The only thing I don't like personally is the name of the var `endSize`. Would much rather it be `endIndex`. But I won't die on that hill :)

server/log-collector/pkg/services/logcollector/server.go
```diff
@@ -1055,22 +1066,29 @@ func (s *Server) isLogCollectionRunning(ctx context.Context, runUID, project str
...
 	}
 
-	return chunkSize
+	// if the size we need to read is bigger than the buffer, use the buffer size
+	leftToRead := endSize - totalLogsSize
```
quaark, line 1086: Actually. Shouldn't this be: ?
```
leftToRead := endSize - totalLogsSize - initialOffset

quaark (changes requested):

quaark (approved):

PR #4534, opened 2023-11-06
[Schedules] Fix `get_schedule` when the last-run was deleted

theSaarco:
When listing schedules, the UI asks for `/projects/<project>/schedules?include_last_run=yes` which tries to enrich the schedule record with the details of the last run executed. 
If deleting the last-run of a given schedule, the last-run uid is still stored in the schedule itself, and it doesn't get removed when the job is deleted from the DB. Therefore, when `_enrich_schedule_with_last_run` tries to retrieve the details of that run (based on its uid) from the DB, the call fails and this fails the `list-schedules` REST call with 404. This situation persists until another run is triggered by the schedule, and the last_run field is updated to point at the new run executed.

The fix is to gracefully handle the situation where the last run is not found in the DB, and also clear the `last_run_uri` from the returned schedule.

Fixes [ML-4902](https://jira.iguazeng.com/browse/ML-4902)

### DIFF
--- a/server/api/utils/scheduler.py
+++ b/server/api/utils/scheduler.py
@@ -906,8 +906,21 @@ class Scheduler:
             run_project, run_uid, iteration, _ = RunObject.parse_uri(
                 schedule_output.last_run_uri
             )
-            run_data = get_db().read_run(db_session, run_uid, run_project, iteration)
-            schedule_output.last_run = run_data
+            try:
+                run_data = get_db().read_run(
+                    db_session, run_uid, run_project, iteration
+                )
+                schedule_output.last_run = run_data
+            except mlrun.errors.MLRunNotFoundError:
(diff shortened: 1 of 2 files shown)


server/api/utils/scheduler.py
```diff
@@ -906,8 +906,21 @@ def _enrich_schedule_with_last_run(
...
+            except mlrun.errors.MLRunNotFoundError:
+                # Possibly the last-run was already deleted (ML-4902). Continue, and clear the last_run_uri in
+                # the response.
+                logger.debug(
+                    "Failed to find the last run for schedule. Continuing",
+                    project=schedule_output.project,
+                    schedule_name=schedule_output.name,
+                    run_uid=run_uid,
+                )
+                schedule_output.last_run_uri = None
```
liranbg, line 923: Id go with 
```
schedule_output.last_run = self._get_last_run()
```
and add a function that will do the try/except
leaving the enrich more concise

PR #4538, opened 2023-11-07
[Runtimes] Log databricks artifact

tomerm-iguazio:
Log Databricks artifacts and tests.

This process is achieved by using the `mlrun_log_artifact` function in the code that runs in Databricks.
Alternatively, if using a handler, return a dictionary from this handler.

The dictionary should have a key for the name of the artifact and a corresponding value for the DBFS path.
The DBFS path format in the code itself should start with /dbfs.

+ added test

[ML-4623](https://jira.iguazeng.com/browse/ML-4623)


mlrun/runtimes/databricks_job/databricks_runtime.py
```diff
@@ -137,3 +160,25 @@ def run(
...
+            existing_data.update(new_data)
+            json_file.seek(0)
+            json.dump(existing_data, json_file)
+    else:
+        parent_dir = os.path.dirname(mlrun_artifacts_path)
+        if parent_dir != '/dbfs':
+            os.makedirs(parent_dir, exist_ok=True)
+        with open(mlrun_artifacts_path, 'w') as json_file:
+            json.dump(new_data, json_file)
+        first_mlrun_artifact = False
```
theSaarco, line 182: What is this used for? Looks like a leftover.

mlrun/runtimes/databricks_job/databricks_runtime.py
```diff
@@ -137,3 +160,25 @@ def run(
 handler_arguments = json.loads(handler_arguments)
 
 """
+
+artifacts_code_template = """\n
+def mlrun_log_artifact(name, path):
```
theSaarco, line 165: This entire code is not really protective in any way - you don't verify that the `path` or the `name` is a string, and you don't really check whether any of the os calls (reading, creating, writing files) actually succeed or not. The end result here may be that user's code would explode due to a problem in our code (which the user may or may not understand). 
In general, this entire code should be treated as a best-effort and it should at least be wrapped with a `try/except` to catch exceptions thrown in between. Should also consider more specific checks about things you can think of.

mlrun/runtimes/databricks_job/databricks_wrapper.py
```diff
@@ -80,21 +102,28 @@ def run_mlrun_databricks_job(
...
         "databricks_run_name", f"mlrun_task_{mlrun_databricks_job_id}"
     )
-
+    current_time = datetime.datetime.utcnow()
+    run_time = current_time.strftime("%H_%M_%S_%f")
+    databricks_run_name = (
+        databricks_run_name
+        if str(mlrun_databricks_job_id) in databricks_run_name
+        else f"{databricks_run_name}_{run_time}"
+    )
```
theSaarco, line 115: Why not always use the `mlrun_databricks_job_id`, like you do if the user didn't provide a run name? Why in this case use the time?

mlrun/runtimes/databricks_job/databricks_wrapper.py
```diff
@@ -44,6 +44,28 @@ def get_task(databricks_run: Run) -> RunTask:
...
+    workspace: WorkspaceClient,
+    artifact_json_path: str,
+    databricks_run_name: str,
+):
+    if not workspace.dbfs.exists(artifact_json_path):
+        return
+    context.logger.info(f"Artifacts found. Run name: {databricks_run_name}")
+    with workspace.dbfs.open(artifact_json_path, read=True) as artifact_file:
+        artifact_json = json.load(artifact_file)
+    for artifact_name, artifact_path in artifact_json.items():
```
theSaarco, line 58: This is where you may fail because you didn't verify the contents of the json are strings - this may cause the runtime code to explode if the user provided other things such as dicts or whatever.

theSaarco (changes requested):
Cool. I think the code needs to be more protective and allow for more flexibility in return values that the user is expected to provide - see my comments.

mlrun/runtimes/databricks_job/databricks_runtime.py
```diff
@@ -41,21 +53,32 @@ def get_internal_parameters(self, runobj: RunObject):
         if not encoded_code:
             return "", original_handler
```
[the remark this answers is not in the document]
tomerm-iguazio: ok tnx!

mlrun/runtimes/databricks_job/databricks_wrapper.py
tomerm-iguazio: add a type decorator in mlrun_log_artifact. tnx!

assaf758 (approved):

PR #4547, opened 2023-11-08
[Pipelines] Clean up Kubeflow specific data from Function definitions after workflow execution

laurybueno:
[ML-4161](https://jira.iguazeng.com/browse/ML-4161)

When a Kubeflow workflow uses pipeline parameters to change resource requests/limits on a function, KFP-specific strings are left on the `_function_objects` dictionary. It's important to stress that this contamination only affects the client's local state and is not stored on the MLRun database since the API has safeguards in place against invalid strings. 

Example:
```
resources': {'requests': {'memory': '{{pipelineparam:op=;name=val}}'},
   'limits': {'memory': '{{pipelineparam:op=;name=val}}'}}
```

There are at least three reasonable strategies for dealing with this problem:
1. Add a filter to `mlrun.db.httpdb.HTTPRunDB.store_function` to avoid sending Kubeflow-specific values when saving the function after a workflow has been run. 
      a. Pros: the solution is straightforward
      b. Cons: if the `resources` field had any values before the workflow code was run, these would be overwritten on the API
2. Prepare a somewhat simple field rollback mechanism for function specs so that after a workflow is run, fields marked as pending rollback can be reverted
      a. Pros: guarantees that original values won't be lost; makes the fix self-contained to the workflow run process, so its consequences don't leak to other tasks on the client
      b. Cons: assumes that reverting field values might be something frequent enough to warrant a structure for it
3. Leverage a hook-style strategy so that `KubeResourceSpec` can proactively queue the field cleanup to be done at the end of a workflow run. In this scenario, `MlrunProject` would have a list of `Callable` to run right after a workflow, and `KubeResourceSpec` could send a function to it to perform the cleanup if appropriate
      a. Pros: ensures that `KubeResourceSpec` is the protagonist of its cleanup strategy
      b. Cons: the simplest way to implement this (that I could think of) would trigger a circular dependency by importing pipeline_context to `mlrun/runtimes/pod.py`. A more complex and scalable strategy might require a new dependency like [pluggy](https://github.com/pytest-dev/pluggy).

On this PR, I tentatively implemented the strategy 2.

### DIFF
--- a/mlrun/projects/pipelines.py
+++ b/mlrun/projects/pipelines.py
@@ -597,7 +597,9 @@ class _KFPRunner(_PipelineRunner):
             try:
                 func.spec.rollback_fields()
             except AttributeError:
-                logger.debug(f"Function of type {type(func)} doesn't require a spec field rollback")
+                logger.debug(
+                    f"Function of type {type(func)} doesn't require a spec field rollback"
+                )
             except Exception as exc:
                 logger.warning(
                     f"Failed to rollback spec fields for function {func.metadata.name}",
--- a/mlrun/projects/pipelines.py
+++ b/mlrun/projects/pipelines.py
@@ -595,15 +595,16 @@ class _KFPRunner(_PipelineRunner):
         # The user provided workflow code might have made changes to function specs that require cleanup
         for func in project.spec._function_objects.values():
             try:
-                func.spec.rollback_fields()
+                func.spec.discard_changes()
             except AttributeError:
                 logger.debug(
-                    f"Function of type {type(func)} doesn't require a spec field rollback"
+                    "Function does not require a field rollback", func_type=type(func)
                 )
             except Exception as exc:
                 logger.warning(
-                    f"Failed to rollback spec fields for function {func.metadata.name}",
+                    "Failed to rollback spec fields for function",
                     project=project,
+                    func_name=func.metadata.name,
                     exc_info=err_to_str(exc),
                 )
 
--- a/mlrun/projects/pipelines.py
+++ b/mlrun/projects/pipelines.py
@@ -591,6 +591,20 @@ class _KFPRunner(_PipelineRunner):
             artifact_path=artifact_path,
             cleanup_ttl=workflow_spec.cleanup_ttl,
         )
+
+        # The user provided workflow code might have made changes to function specs that require cleanup
+        for func in project.spec._function_objects.values():
+            try:
+                func.spec.rollback_fields()
+            except AttributeError:
+                logger.debug(f"Function of type {type(func)} doesn't require a spec field rollback")
+            except Exception as exc:
+                logger.warning(
+                    f"Failed to rollback spec fields for function {func.metadata.name}",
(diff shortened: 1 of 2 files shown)


mlrun/projects/pipelines.py
```diff
@@ -591,6 +591,22 @@ def run(
...
         )
+
+        # The user provided workflow code might have made changes to function specs that require cleanup
+        for func in project.spec._function_objects.values():
+            try:
+                func.spec.rollback_fields()
+            except AttributeError:
+                logger.debug(
+                    f"Function of type {type(func)} doesn't require a spec field rollback"
+                )
```
liranbg, line 602: our logger knows how to handle kwargs so once we change human readable to json, it structure the key-val in an organized way
```suggestion
                logger.debug(
                    "Function does not require field rollback", func_type=type(func),
                )
```

this apply for the rest of the added log lines

mlrun/runtimes/pod.py
```diff
@@ -330,6 +333,10 @@ def _verify_and_set_limits(
                         resource_value,
                     )
 
+    def rollback_fields(self):
+        for k, v in self.__fields_pending_rollback.items():
+            setattr(self, k, v)
```
liranbg, line 338: maybe discard_changes?
we need to doc that function as it is public. perhaps explain the intention and why it is needed (I guess can be taken from PR)

mlrun/runtimes/pod.py
```diff
@@ -338,6 +345,15 @@ def _verify_and_set_requests(
...
         resources = verify_requests(resources_field_name, mem=mem, cpu=cpu)
+        for pattern in mlrun.utils.regex.pipeline_param:
+            if re.match(pattern, str(cpu)) or re.match(pattern, str(mem)):
+                self.__fields_pending_rollback.update(
+                    {
+                        resources_field_name: copy.deepcopy(
+                            getattr(self, resources_field_name)
+                        ),
+                    }
+                )
```
liranbg, line 356: id move to `self._add_field_to_pending_discard(resources_field_name, getattr(self, resources_field_name))` 
that way you can
1. have a bit more concise function
2. if fields_pending_rollback structure is changed, no need to add in many places (assuming more function will reuse that logic

wdyt?

liranbg (commented):
While (1) sounds a bit more intuitive, its con approves why you went with (2). Few suggestions below and great job!

mlrun/projects/pipelines.py
laurybueno: Good point. I improved it for us.

mlrun/runtimes/pod.py
laurybueno: Indeed. I added added a docstring for us. Please take a look to see if it's appropriate.

mlrun/runtimes/pod.py
laurybueno: Good thinking. I made this refactor now.

liranbg (approved):

PR #4550, opened 2023-11-09
[Project] Update in the project api of the function usage, including mm-apps framework

assaf758:
Completing @davesh0812 [PR](https://github.com/mlrun/mlrun/pull/4394).
 
Few update in project api:

1. create_model_monitoring_function - new api, create a monitoring function object without setting it to the project.
2. set_model_monitoring_function - new api (replace set_model_monitoring_function), update or add a monitoring function to the project (without doing deploy_function).
3. remove_model_monitoring_function - new api, remove a monitoring function from the project
4. list_model_monitoring_functions - new api, retrieve a list of all the model monitoring functions

Also add a method that indicates if a runtime is a model monitoring function - is_model_monitoring_function

[ML-4706](https://jira.iguazeng.com/browse/ML-4706)


mlrun/projects/project.py
```diff
@@ -2017,28 +2188,28 @@ def _set_function(
         self.spec.set_function(name, function_object, func)
 
     def remove_function(self, name):
-        """remove a function from a project
+        """remove the specified function from the project
 
         :param name:    name of the function (under the project)
         """
         self.spec.remove_function(name)
+        mlrun.db.get_run_db().delete_function(name=name.lower())
```
alonmr, line 2196: > remove_XXX - APIs that will only remove from project object (the opposite of set)
from https://jira.iguazeng.com/browse/ML-4893

mlrun/projects/project.py
```diff
@@ -1903,18 +2004,88 @@ def set_function(
         self._set_function(resolved_function_name, tag, function_object, func)
         return function_object
 
-    def _resolved_function(
+    def create_function(
```
alonmr, line 2007: not sure why create function is needed? I think its out of context here. Why would a user want to create a function in a project and not set it?
Also, in the examples docstring you allow getting an object and a string but the type hints suggest you only accepts strings.

alonmr (commented):
Minor comments

mlrun/projects/project.py
assaf758: removed, thx!

mlrun/projects/project.py
assaf758: Agree it somewhat out of context so removing.
Btw there is a use-case for not setting into the project - when you don't want to export it

alonmr (approved):

mlrun/projects/project.py
```diff
@@ -3113,16 +3213,23 @@ def list_functions(self, name=None, tag=None, labels=None):
             # convert dict to function objects
             return [mlrun.new_function(runtime=func) for func in functions]
 
-    def list_model_monitoring_applications(self):
-        """Retrieve a list of alll the model monitoring application.
+    def list_model_monitoring_functions(self, name=None, tag=None, labels=None):
```
jond01, line 3216: Type hints?

jond01 (commented):
I had some small comments and questions

mlrun/model_monitoring/tracking_policy.py
```diff
@@ -35,7 +35,7 @@ def __init__(
         ] = mlrun.common.schemas.schedule.ScheduleCronTrigger(minute="0", hour="*/1"),
         default_batch_image: str = "mlrun/mlrun",
         stream_image: str = "mlrun/mlrun",
-        application_batch: bool = False,
+        application_batch: bool = True,
```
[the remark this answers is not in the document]
assaf758: unrelated change.

mlrun/projects/project.py
```diff
@@ -1733,7 +1733,7 @@ def setup(self, save: bool = True) -> "MlrunProject":
         )
         return _run_project_setup(self, setup_file_path, save)
 
-    def set_model_monitoring_application(
+    def set_model_monitoring_function(
```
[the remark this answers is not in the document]
assaf758: well, it's model-monitoring application implemented by mlrun application.

mlrun/projects/project.py
```diff
@@ -1815,30 +1918,23 @@ def set_model_monitoring_application(
...
+        return resolved_function_name, function_object, func
 
     def set_function(
         self,
         func: typing.Union[str, mlrun.runtimes.BaseRuntime] = None,
         name: str = "",
         kind: str = "",
         image: str = None,
-        handler=None,
+        handler: str = None,
```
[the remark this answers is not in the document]
assaf758: ok, will fix the type hints in separate PR

mlrun/projects/project.py
assaf758: same

PR #4552, opened 2023-11-09
[Nuclio] Allow not creating an http trigger for Nuclio functions by default

rokatyy:
Jira - https://jira.iguazeng.com/browse/NUC-58 , https://jira.iguazeng.com/browse/ML-4032

Recently, we implemented a new functionality in Nuclio that allows deploying functions without a default HTTP trigger. This pull request expands this capability in MLRun, enabling users to disable or enable the default HTTP trigger through MLRun.

### DIFF
--- a/mlrun/runtimes/function.py
+++ b/mlrun/runtimes/function.py
@@ -163,6 +163,7 @@ class NuclioSpec(KubeResourceSpec):
         add_templated_ingress_host_mode=None,
         clone_target_dir=None,
         state_thresholds=None,
+        disable_default_http_trigger=None,
     ):
 
         super().__init__(
@@ -209,6 +210,8 @@ class NuclioSpec(KubeResourceSpec):
 
         self.min_replicas = min_replicas or 1
         self.max_replicas = max_replicas or 4
+
+        self.disable_default_http_trigger = disable_default_http_trigger
 
         # When True it will set Nuclio spec.noBaseImagesPull to False (negative logic)
         # indicate that the base image should be pulled from the container registry (not cached)
@@ -682,6 +685,22 @@ class RemoteRuntime(KubeResource):
         raise NotImplementedError(
             "State thresholds do not apply for nuclio as it has its own function pods healthiness monitoring"
         )
+
+    def disable_default_http_trigger(
+        self,
+    ):
+        """
+        Disables nuclio's default http trigger creation
+        """
+        self.spec.disable_default_http_trigger = True
+
+    def enable_default_http_trigger(
+        self,
+    ):
+        """
+        Enables nuclio's default http trigger creation
+        """
+        self.spec.disable_default_http_trigger = False
 
     def _get_state(
         self,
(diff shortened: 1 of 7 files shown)


liranbg (changes requested):
I dont think this is enough. look at 
`deploy_nuclio_function` function and see its flow. basically when you create a nuclio function it is automatically enriched with http ingress and an ingress

then look at "RemoteRuntime" class and see its "with_http" functionality.
Think that user might add http trigger and set disable to true, it doesnt make sense of course and nuclio might return an error explaining that but we want to be ahead of that and to *not* enrich each function with *http trigger* automatically if that option is set *on mlrun side*.

also need take care of "invoke" when function has no http trigger at all. need to make sure it is explained to client that invoking cannot happen because function has no invocation url


Lastly, I think it worth adding that functionality to "cheat-sheet.md" under "#### Nuclio"

rokatyy:
@liranbg agreed that invoking logic is needed and will add docs as well. 

Regarding having custom http trigger when default http trigger creation is disabled - there won't be any error from nuclio side, because technically custom and default http triggers are not the same thing. Nuclio checks disableHTTPtrigger only in case when there are no any custom http trigger. Moreover, crucial to remember about the case when disableHTTPtrigger  isn't set in function config, which means that it is taken from platfrom config, so we cannot guarantee from mlrun side that function will be envokable even if user didn't set disableHTTPtrigger to true on it's side. But I will add error raising in case if user set `disableHTTPtrigger=true`

mlrun/runtimes/function.py
```diff
@@ -878,6 +880,15 @@ def invoke(
...
         """
+
+        # here we check that if default http trigger is disabled, function contains a custom http trigger
+        # Otherwise, the function is not invokable, so we raise an error
+        if "spec.triggers.http" not in self.spec.config and self.spec.disable_default_http_trigger:
+            raise mlrun.errors.MLRunPreconditionFailedError(
+                "Default http trigger creation is disabled and there is no any other custom http trigger, "
+                "so function can not be invoked via http. Either enable default http trigger creation or create"
+                "custom http trigger")
+
```
liranbg: I think this can be changed to inspect the function status and see if it does not have either external/internal invocation url - then not invokable. that as long as path does not include "://" of course.

server/api/crud/runtimes/nuclio/helpers.py
```diff
@@ -92,6 +92,11 @@ def enrich_function_with_ingress(config, mode, service_type):
     # we would enrich it with an ingress
     http_trigger = resolve_function_http_trigger(config["spec"])
     if not http_trigger:
+
+        # if there are no http triggers and default http trigger creation disabled,
+        # function is not invokable
+        if config["spec"].get("disableDefaultHTTPTrigger"):
```
liranbg, line 98: `if config["spec"].get("disableDefaultHTTPTrigger", False):`
just to explcit the "type" of value you'd expect (not string or so)

liranbg (commented):
minor

mlrun/runtimes/function.py
rokatyy: @liranbg this logic already exists in this function below, so should I delete this part then? As an option I can also add this check after 
`path = self._resolve_invocation_url(path, force_external_address)` in case of `not path`. wdyt?

liranbg: sounds good. just make sure you cover it with a test

PR #4555, opened 2023-11-09
[LogCollector] Fix big stop logs

liranbg:
When list of run uids is too big, it exceeds grpc message size and fails with

```
> 2023-11-09 18:35:56,028 [warning] Failed stopping logs for runs. Ignoring: {'exc': '<AioRpcError of RPC that terminated with:\n\tstatus = StatusCode.RESOURCE_EXHAUSTED\n\tdetails = "grpc: received message larger than max (12470570 vs. 4194304)"\n\tdebug_error_string = "UNKNOWN:Error received from peer ipv4:127.0.0.1:8282 {created_time:"2023-11-09T18:35:56.025126602+00:00", grpc_status:8, grpc_message:"grpc: received message larger than max (12470570 vs. 4194304)"}"\n>', 'project': 'load-project-0', 'run_uids':  .....
```

This PR chunk the list of run uids into a smaller portion ensuring the message size is relatively small

https://jira.iguazeng.com/browse/ML-4989

### DIFF
--- a/server/api/main.py
+++ b/server/api/main.py
@@ -747,25 +747,34 @@ async def _stop_logs():
         await fastapi.concurrency.run_in_threadpool(close_session, db_session)
 
 
-async def _stop_logs_for_runs(runs: list):
-    project_to_run_uids = {}
+async def _stop_logs_for_runs(runs: list, chunk_size: int = 10):
+    project_to_run_uids = collections.defaultdict(list)
     for run in runs:
         project_name = run.get("metadata", {}).get("project", None)
         run_uid = run.get("metadata", {}).get("uid", None)
-        project_to_run_uids.setdefault(project_name, []).append(run_uid)
+        project_to_run_uids[project_name].append(run_uid)
 
     for project_name, run_uids in project_to_run_uids.items():
-        try:
-            await server.api.utils.clients.log_collector.LogCollectorClient().stop_logs(
-                project_name, run_uids
-            )
-        except Exception as exc:
-            logger.warning(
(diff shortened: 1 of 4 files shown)


server/api/main.py
```diff
@@ -747,25 +747,34 @@ async def _stop_logs():
...
-            )
-        except Exception as exc:
-            logger.warning(
-                "Failed stopping logs for runs. Ignoring",
-                exc=err_to_str(exc),
-                project=project_name,
-                run_uids=run_uids,
-            )
+        if not run_uids:
+            logger.debug("No runs to stop logs for", project=project_name)
```
TomerShor, line 759: Is this even possible?
You add a project to the dict only if there's a run for that project.

server/api/main.py
```diff
@@ -747,25 +747,34 @@ async def _stop_logs():
         await fastapi.concurrency.run_in_threadpool(close_session, db_session)
 
 
-async def _stop_logs_for_runs(runs: list):
-    project_to_run_uids = {}
+async def _stop_logs_for_runs(runs: list, chunk_size: int = 10):
```
alonmr, line 750: this method is not being run in threadpool. I believe it blocks the main thread like we talked.

server/api/main.py
liranbg: just defensive

server/api/main.py
liranbg: there is nothing to run here in a thread. the code within is fully asyncio compatible (awaitable)

PR #4558, opened 2023-11-12
[Docs] Add warning banner to old versions that this is not the latest version 

jillnogold:


docs/conf.py
```diff
@@ -88,6 +89,11 @@ def current_version():
     ".md": "myst-nb",
 }
 
+# versionwarning configuration
+versionwarning_default_message = (
+    "This is not the latest documentation. See {latest} instead."
```
alonmr, line 94: you need the placeholder to match `versionwarning_message_placeholder` . since you didn't set it, the default is newest. The bellow should work.
```suggestion
    "This is not the latest documentation. See {newest} instead."
```

alonmr (approved):

PR #4560, opened 2023-11-12
[Server] Avoid setting workdir to target dir

gtopper:
[ML-4992](https://jira.iguazeng.com/browse/ML-4992)


theSaarco (approved):
The fix (specifically not using `WORKDIR`) seems ok, though it does have some potential BC issue with user code, especially if the user had some commands running in the dockerfile - if those commands assumed a given workdir (for example using relative paths) they would stop working after this change. 
I think this change therefore calls for some release note at least documenting this behavior change.

PR #4566, opened 2023-11-12
[Builder] Fix resolving secret name for using mlrun default registry

liranbg:
When using the default registry explicitly, the secret name is not automatically populated as opposed to how it works when using "." as image prefix (which turns "." -> "default registry")

This PR simply checks whether the target image uses mlrun default registry and if no secret name is provided (as an override) it simply populates the secret name from mlrun default configuration

https://jira.iguazeng.com/browse/ML-4953

### DIFF
--- a/mlrun/api/utils/builder.py
+++ b/mlrun/api/utils/builder.py
@@ -28,6 +28,7 @@ import mlrun.api.utils.singletons.k8s
 import mlrun.common.constants
 import mlrun.common.schemas
 import mlrun.errors
+import mlrun.model
 import mlrun.runtimes.utils
 import mlrun.utils
 from mlrun.config import config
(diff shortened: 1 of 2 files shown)


theSaarco (approved):
Looks good.
As this requires fast fixing, please consider the following points for subsequent PRs:
* Add a test covering the build API and ensuring the values are filled in properly (don't know if we already have any such tests). I assume you've verified this fixes the issue we see?
* Also, I guess a similar logic is performed someplace in the client-side sdk code, since this issue didn't happen when building from sdk. Should remove the client-side logic and just keep it in the server.

liranbg:
>Add a test covering the build API and ensuring the values are filled in properly (don't know if we already have any such tests). I assume you've verified this fixes the issue we see?

indeed. verified api, ui

liranbg:
> Also, I guess a similar logic is performed someplace in the client-side sdk code, since this issue didn't happen when building from sdk. Should remove the client-side logic and just keep it in the server.

will be done on pr to development. this is a quick fix for 1.5.1

PR #4575, opened 2023-11-14
[Builder] Fix resolving secret name for using mlrun default registry

liranbg:
porting #4566 to development with added tests and re-organization for code that no longer needs to reside within mlrun but move to where it is being used @ server/api.
In addition, removed some unused code

https://jira.iguazeng.com/browse/ML-4953


theSaarco (approved):
LGTM.

PR #4579, opened 2023-11-14
[Datastore] Support Dbfs datastore profiles

tomerm-iguazio:
1) Add support for datastore Profiles.
2) Add tests.

Pay attention!
Support for datastore profiles and dask dataframe (dd) will be added in a future PR.

[ML-4828](https://jira.iguazeng.com/browse/ML-4828)

### DIFF
--- a/mlrun/datastore/base.py
+++ b/mlrun/datastore/base.py
@@ -49,6 +49,8 @@ class FileStats:
 
 
 class DataStore:
+    using_bucket = False
+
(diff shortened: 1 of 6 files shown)


theSaarco (changes requested):
Looks good - I have some concerns about how you handle the `using_bucket` parameter, I think it needs to be much more transparent.

theSaarco (approved):
ok

PR #4587, opened 2023-11-16
[Secrets] Fix v3io access-key that remained in cleartext when creating function through UI

theSaarco:
When saving functions, we replace cleartext v3io access-keys with auth-secrets. To do that we rely on username derived from either the mount path (to conform with how `mlrun.mount_v3io` works) or the env variable `V3IO_USERNAME` which is set in the function spec - the MLRun sdk/client code set this env variable.
When a function is created through the UI with user feeding manual v3io mount, those details do not exist, thus the BE code cannot mask the secret due to lack of username. The result is that the function is stored with the access-key in cleartext.

This fixes this scenario by passing the username from the REST API auth info, and using it if nothing else works.

Fixes [ML-5068](https://jira.iguazeng.com/browse/ML-5068)

### DIFF
--- a/server/api/api/utils.py
+++ b/server/api/api/utils.py
@@ -483,10 +483,12 @@ def ensure_function_auth_and_sensitive_data_is_masked(
 def mask_function_sensitive_data(function, auth_info: mlrun.common.schemas.AuthInfo):
     if not mlrun.runtimes.RuntimeKinds.is_local_runtime(function.kind):
         _mask_v3io_access_key_env_var(function, auth_info)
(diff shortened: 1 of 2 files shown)


alonmr (approved):
Looks good but in general I wonder why not always use the auth info 1st?
I understand its a big change but eventually isn't it more accurate?

theSaarco:
@alonmr - the issue is that a user may create a function that accesses storage using some other user's creds. So, if the user chose to specify a different user name (either through `mount_v3io` or through setting the env variable) we need to respect that **before** taking into account the creds from the REST API.

PR #4589, opened 2023-11-16
[Runtimes] Add set categories method to runtimes

Yacouby:


mlrun/runtimes/base.py
```diff
@@ -196,6 +196,9 @@ def set_label(self, key, value):
         self.metadata.labels[key] = str(value)
         return self
 
+    def set_categories(self, categories: List[str]):
+        self.metadata.categories = categories
```
liranbg, line 200: `mlrun.utils.helpers.as_list` to ensure the input is indeed a list

liranbg (commented):
small

mlrun/runtimes/base.py
Yacouby: Done

PR #4598, opened 2023-11-19
Correct the check for the existence of the HTTP trigger

rokatyy:
[No ticket]

Correct the check for the existence of the HTTP trigger so it work correctly if name of trigger is not equal to `http`

### DIFF
--- a/mlrun/runtimes/function.py
+++ b/mlrun/runtimes/function.py
@@ -906,7 +906,7 @@ class RemoteRuntime(KubeResource):
                 # here we check that if default http trigger is disabled, function contains a custom http trigger
                 # Otherwise, the function is not invokable, so we raise an error
                 if (
-                    "spec.triggers.http" not in self.spec.config
+                    not self._trigger_of_kind_exists(kind="http")
                     and self.spec.disable_default_http_trigger
                 ):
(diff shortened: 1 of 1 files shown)


mlrun/runtimes/function.py
```diff
@@ -952,6 +952,17 @@ def invoke(
...
         return data
 
+    def _trigger_of_kind_exists(self, kind: str) -> bool:
+        if not self.spec.config:
+            return False
+
+        for name, spec in self.spec.config.items():
+            if isinstance(spec, dict):
+                if spec.get("kind") == kind:
+                    return True
```
TomerShor, line 962: What if the `spec` is a dict but does not have a `kind`?
e.g for something other than triggers

we don't want to fail in this case, then:
```
if spec.get("kind", None) == kind:
```

TomerShor (changes requested):

mlrun/runtimes/function.py
rokatyy: @TomerShor None is default value for [`get`](https://www.tutorialspoint.com/python/dictionary_get.htm) function, so we don't need to set it explicitly

TomerShor (approved):

liranbg (approved):

PR #4599, opened 2023-11-19
[Runtimes] Fixed missing `force_build` in `deploy()` for spark and serving runtimes

theSaarco:
Calling `project.build_function()` on a Spark runtime would explode with:
```
TypeError: deploy() got an unexpected keyword argument 'force_build'
```
Reason is that the `force_build` param was missing in the `Spark3Runtime.deploy()` method - it was overlooked when this functionality was implemented.
Same is done for `ServingRuntime` (though it's usually not directly invoked through project functions).
Added the missing parameter.


liranbg (changes requested):
runtimes/server.py also misses this argument

PR #4603, opened 2023-11-19
[Artifacts] Convert legacy artifacts when loading in dict_to_artifact

TomerShor:
In some cases such as `load_model`, the Legacy Artifacts class was still used, causing the v2 artifacts API calls to fail, due the the Artifacts schema enforcing the new artifact format (which includes `spec`, `metadata` and `status`). 

Resolves https://jira.iguazeng.com/browse/ML-4808

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -791,13 +791,6 @@ class HTTPRunDB(RunDBInterface):
             params["tree"] = tree
 
         body = _as_json(artifact)
-
-        logger.info(
-            "TOMER - Sending store artifact request",
-            endpoint_path=endpoint_path,
-            body=body,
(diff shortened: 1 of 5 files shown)


mlrun/db/httpdb.py
```diff
@@ -791,6 +791,13 @@ def store_artifact(
             params["tree"] = tree
 
         body = _as_json(artifact)
+
+        logger.info(
+            "TOMER - Sending store artifact request",
+            endpoint_path=endpoint_path,
+            body=body,
+        )
```
liranbg, line 799: Hi Tomer :)

TomerShor: Hi oopsy

PR #4608, opened 2023-11-20
[Storey] Fix test for aggregations returning NaN

alxtkr77:
Fix system test following changes in '[Storey] Make aggregations return NaN when there is no data except for count' pull request.
https://github.com/mlrun/mlrun/pull/4442


gtopper (approved):
LGTM

PR #4614, opened 2023-11-21
[Pipelines] Patch unsupported features during user workflow parsing

laurybueno:
[ML-4028](https://jira.iguazeng.com/browse/ML-4028)

To add flexibility to our Kubeflow SDK version requirement, we need a system to inform users when they try to use an unsupported feature available on newer versions. This PR introduces such a system.

It's important to note that our goal here is to be user-friendly rather than intrusive. This is why, if the patching of an unsupported feature fails, this system prioritizes giving back control to the user and allowing them to proceed.

Because of where the context manager introduced on this PR has to be used, it has a **lot of disruptive potential to regular user interactions when creating workflows, so an extensive review is recommended** for this PR. 

# Caveats and alternatives
Patching external functionality is hardly ever pretty work, but this PR doesn't employ the best strategy to avoid overengineering a situational feature.

If we ever consider expanding our unsupported features list, we might want to consider more solid approaches, such as:

## Import hooks
Traditionally, this would be the preferred method to solve our current problem. However, it's unlikely to work consistently since the user might import `kfp` before MLRun's execution starts.

That being said, if we would like to research this path further, this snippet would be a sound starting point:
```python
from importlib.abc import PathEntryFinder
import sys

class UnsupportedDependencies(PathEntryFinder):
    __dependencies = {
        "kfp.dsl.ParallelFor": "ParallelFor is not currently supported by MLRun",
    }

    def find_spec(self, fullname, path, target=None):
        if fullname in self.__dependencies.keys():
            raise NotImplementedError(
                self.__dependencies.get(
                    fullname, "This feature is not currently supported by MLRun"
                )
            )
        return None

sys.meta_path.insert(0, UnsupportedDependencies())
```

## Audithooks
Python's Audit Hooks can also be a solid alternative here since they work independently from when an import has taken place. We might need to use memory addresses or object IDs to test if an unsupported Class of function is being invoked. 

A good starting point could be something like this:
```python
from sys import addaudithook

dependencies = ["ParallelFor"]

def unsupported_dependencies(event, arg):
    if event == "object.__getattr__" and any([d in arg[0] for d in dependencies]):
        raise NotImplementedError(f"{arg[0]} is not currently supported by MLRun")

addaudithook(unsupported_dependencies)
```

## Unittest patches
Patching functionality is second nature to `unittest,` so it'd make sense to try and use it here. And since it's part of the standard Python distribution, it's likely to be available on all environments in which we run MLRun.

However, it has its caveats. For one, it only patches references (and not the actual objects), so late imports can override it. Beyond that, `unittest` needs to consider how the user makes the actual import in their file. E.g.: `from kfp.dsl import ParallelFor` requires a different treatment to `from kfp import dsl; dsl.ParallelFor`. The first calls for `mock.patch("<user_file_name>.ParallelFor", ...)` while the second requires `mock.patch("kfp.dsl.ParallelFor", ...)`

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -46,6 +46,7 @@ from ..utils import (
     logger,
     new_pipe_metadata,
     normalize_name,
+    unsupported,
     version,
 )
 from .base import RunDBError, RunDBInterface
@@ -1461,9 +1462,10 @@ class HTTPRunDB(RunDBInterface):
                 cleanup_ttl=cleanup_ttl,
                 op_transformers=ops,
             )
-            kfp.compiler.Compiler().compile(
-                pipeline, pipe_file, type_check=False, pipeline_conf=conf
-            )
+            with unsupported.disable_unsupported_external_features():
+                kfp.compiler.Compiler().compile(
+                    pipeline, pipe_file, type_check=False, pipeline_conf=conf
+                )
 
         if pipe_file.endswith(".yaml"):
             headers = {"content-type": "application/yaml"}
--- a/mlrun/utils/unsupported.py
+++ b/mlrun/utils/unsupported.py
@@ -34,14 +34,14 @@ def disable_unsupported_external_features():
     """
     try:
         originals = []
-        for d in dependencies:
-            module_name, _, callable_name = d.rpartition(".")
+        for dependency in dependencies:
+            module_name, _, callable_name = dependency.rpartition(".")
             try:
                 curr = getattr(import_module(module_name), callable_name)
             except (ImportError, KeyError) as exc:
                 logger.debug(
                     "Module import for dependency patching failed",
-                    dependency=d,
+                    dependency=dependency,
                     exc_info=err_to_str(exc),
                 )
                 continue
@@ -49,7 +49,7 @@ def disable_unsupported_external_features():
             if not isclass(curr):
                 logger.debug(
                     "MLRun only supports patching of features based on classes",
-                    dependency=d,
+                    dependency=dependency,
                 )
                 continue
 
@@ -60,7 +60,7 @@ def disable_unsupported_external_features():
                 }
             )
             curr.__init__ = mock.MagicMock(
-                side_effect=NotImplementedError(f"MLRun does not support {d}")
+                side_effect=NotImplementedError(f"MLRun does not support {dependency}")
             )
         # Signal the caller that the context is ready
         yield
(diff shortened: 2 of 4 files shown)


mlrun/utils/unsupported.py
```diff
@@ -0,0 +1,85 @@
...
+@contextmanager
+def disable_unsupported_external_features():
+    """
+    Disables functionality external to MLRun that is known to cause problems during workflow parsing
+    This function operates on a best-effort basis and lets errors pass mostly silently
+    Only features based on classes are supported since this patches their __init__
+    """
+    try:
+        originals = []
+        for d in dependencies:
```
liranbg, line 37:
```suggestion
        for dependency in dependencies:
```

mlrun/db/httpdb.py
```diff
@@ -1461,9 +1462,10 @@ def submit_pipeline(
                 cleanup_ttl=cleanup_ttl,
                 op_transformers=ops,
             )
-            kfp.compiler.Compiler().compile(
-                pipeline, pipe_file, type_check=False, pipeline_conf=conf
-            )
+            with unsupported.disable_unsupported_external_features():
```
liranbg, line 1465: checkout `min_nuclio_versions` functionality as well, this is how we restrict certain nuclio versions according to the minimal version being supported
perhaps can unite the logic under same idea

liranbg (changes requested):
With that, we need to remove the restriction from kfp < 1.18.13 and allow it to be latest ~1.8

mlrun/db/httpdb.py
laurybueno: I see two main opportunities here: 

**(1) How to detect unsupported features** 
The Nuclio function seems to get its data from MLConf. I don't think we can do that to detect if `ParallelFor` is going to be run by the user. We could inspect their source code directly though.

**(2) How to wrap code that needs to be checked** 
A decorator-style wrapper would work indeed, but it won't really look like a decorator in this case, because we want to wrap the user-defined function (whose pointer is held by the `pipeline` variable). So, it'll work, but it'll look more like `pipeline = disable_unsuported_features(pipeline)` than a true decorator. If that's preferable, let me know. I went with the context manager approach because the idea of making changes to an environment before **and** after something runs felt very context-like to me.

I'm happy to brainstorm some ideas with you. Wanna do a quick PS session together?

mlrun/utils/unsupported.py
laurybueno: That's fair. I made the changes for us.

laurybueno:
After more research, the MLRun Platform Team realized that patching `ParallelFor` is not a requirement to support the latest release of KFP SDK 1.8. So, we can walk away from this strategy.

This PR is closed in favor of https://github.com/mlrun/mlrun/pull/4619

PR #4615, opened 2023-11-21
[Artifacts] Try getting artifact with tag as producer_id if not found

TomerShor:
Earlier clients sometime tried to get artifacts with the `producer_id` (`tree`) or `uid` in the `tag` query param, which resulted in not found artifacts.
Since the old clients use the old `v1` artifact endpoints, in this PR I added a fallback where if the artifact is not found in `get_artifact`, we try to  get the artifact with the tag used as the tree. 

Resolves https://jira.iguazeng.com/browse/ML-4302

### DIFF
--- a/server/api/api/endpoints/artifacts.py
+++ b/server/api/api/endpoints/artifacts.py
@@ -146,16 +146,17 @@ async def get_artifact(
     auth_info: mlrun.common.schemas.AuthInfo = Depends(deps.authenticate_request),
     db_session: Session = Depends(deps.get_db_session),
 ):
-    data = await run_in_threadpool(
-        server.api.crud.Artifacts().get_artifact,
-        db_session,
-        key,
-        tag=tag,
-        iter=iter,
-        project=project,
-        format_=format_,
-    )
-    if not data:
+    try:
+        data = await run_in_threadpool(
+            server.api.crud.Artifacts().get_artifact,
+            db_session,
+            key,
+            tag=tag,
+            iter=iter,
+            project=project,
+            format_=format_,
+        )
+    except mlrun.errors.MLRunNotFoundError:
         # in earlier versions, producer_id and tag got confused with each other,
         # so we try to get the artifact with the given tag as the producer_id before returning an empty response
         data = await run_in_threadpool(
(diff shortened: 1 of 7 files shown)


alonmr (commented):
A few notes/questions:
1. You should remove the changes you took from development to avoid merge conflicts afterwards.
2. What about the other place in the client you added this? is it still needed?

TomerShor:
@alonmr 
1. ✅ 
2. The fallback in the client was trying to find the artifact with `tag=tree`, but since we would only have older clients with new api - I removed it.

server/api/api/endpoints/artifacts.py
```diff
@@ -146,15 +146,28 @@ async def get_artifact(
...
+            server.api.crud.Artifacts().get_artifact,
+            db_session,
+            key,
+            tag=tag,
+            iter=iter,
+            project=project,
+            format_=format_,
+        )
+    except mlrun.errors.MLRunNotFoundError:
+        # in earlier versions, producer_id and tag got confused with each other,
```
alonmr, line 160: i'm not sure we want to chain the exceptions. We might want to see how the response will look like in case you get 404 for both tries.
I think we want to do something like (see `get_or_create_project` for reasoning):
```suggestion
    not_found = False
    try:
        data = await run_in_threadpool(
            server.api.crud.Artifacts().get_artifact,
            db_session,
            key,
            tag=tag,
            iter=iter,
            project=project,
            format_=format_,
        )
    except mlrun.errors.MLRunNotFoundError:
        not_found = True
     if not_found:
         try:
             ...
          except mlrun.errors.MLRunNotFoundError:
              pass
      if not data:
          raise mlrun.errors.MLRunNotFoundError(some error message)
```

TomerShor: The response will be the same as a single try, e.g:
```
<Response [404 Not Found]>
{'detail': "MLRunNotFoundError('Artifact prj/some-key#0:latest not found')"}
```

I can add a log line that we fallback to `tag=tree`, but I don't see why this will happen:
```
 or else the exceptions raised below will be logged along with the "not found" message
```

alonmr (approved):

PR #4623, opened 2023-11-22
[Builder] Add mlrun requirement when installing PyPs on mlrun base image

alonmr:
Add mlrun to the requirements even though it is already installed because we want pip to include mlrun constraints when installing other packages.
A few rules to note:
1. mlrun version specifier overrides everything
2. Take mlrun version from enriched base image tag if possible


theSaarco (approved):
👍

PR #4495, opened 2023-10-26
[Model Monitoring] Add MM applications dashboard to Grafana

Eyal-Danieli:
In this PR:
- Generate V3IO KV Schema file under `users/pipelines/{project_name}/monitoring-apps/{endpoint_id}`. 
- Add a new Grafana dashboard called `Model Monitoring - Applications`. The new Grafana dashboard includes charts and KPIs that are relevant to a specific monitoring application (under a specific model endpoint). Most of the presented stats are taken from the V3IO TSDB. 

A dashboard example:

image

PR #4624, opened 2023-11-22
[Model Monitoring] Calculate `current_stats` for each application and forward to writer

Eyal-Danieli:
- **Grafana**: Update `model-monitoring-applications` dashboard, including fixing an issue regarding filtering model endpoints in the TSDB charts.
- **Apps Parquet**: Rename `batch_controller_parquet` to `apps_parquet`. In addition, we removed the time partitioning. That means, each app parquet will be stored under the following path: 
`<project-name>/artifacts/model-endpoints/apps_parquet/key=<endpoint-id>/<start_infer_time>/<app_name>.parquet`. Note that **we create a parquet file for each application for each time range**. In a future optimization, we will generate a single file for multiple apps that defined under the same interval. 
- **`start_infer_time` and `end_time_time`**: Add `start_infer_time` and `end_infer_time` to the application record result (aka `ModelMonitoringApplicationResult `). These values represent the start and end time of the application window interval. Right now it's the controller interval. Once implemented, we will use the monitoring policy to generate time range for each application.
- **Writer**: Update the KV schema file
- **Rename batch application to controller**: Rename `batch_application.py` to `controller.py`, rename `batch_application_hanlder.py` to `controller_handler.py`, rename `BatchApplicationProcessor` to `MonitoringApplicationController`, and rename controller function name from `model-monitoring-batch-application` to `model-monitoring-controller`

Related JIRA - https://jira.iguazeng.com/browse/ML-4748


mlrun/model_monitoring/writer.py
```diff
@@ -154,12 +155,41 @@ def _update_kv_db(self, event: _AppResultEvent) -> None:
...
+            {
+                "name": WriterEvent.RESULT_VALUE,
+                "type": "double",
+                "nullable": False,
+            },
+            {
+                "name": WriterEvent.RESULT_STATUS,
+                "type": "double",
+                "nullable": False,
+            },
```
jond01, line 192: Could you please change double -> int in kind and status?
https://github.com/mlrun/mlrun/pull/4495/files#r1375888171

jond01 (commented):
If you can address also:
https://github.com/mlrun/mlrun/pull/4495/files#r1375935350
It will be nice

mlrun/model_monitoring/controller.py
```diff
@@ -17,10 +17,10 @@
 import json
 import os
 import re
-from typing import Callable, Tuple
+import typing
+from typing import Callable, List, Tuple
```
jond01, line 21: Please choose one option

Eyal-Danieli: https://github.com/mlrun/mlrun/pull/4624/commits/553be4f8b7dec387a0df8677aaac74cbd76cf2c7#:~:text=from%20typing%20import%20Any%2C%20Callable%2C%20Dict%2C%20List%2C%20Optional%2C%20Tuple

mlrun/model_monitoring/writer.py
Eyal-Danieli: fixed - https://github.com/mlrun/mlrun/pull/4624/commits/f31e00b3c580fcd6c49f4f0e558fcac938c3e1ca
BTW, according to v3io docs, we can only use one of the following types - `string`, `double`, and `long` (https://github.com/v3io/v3io-py/blob/b869b0eb85d9f81fbe83218bac7265de635a6e2c/v3io/dataplane/kv.py#L359C1-L359C58).
However, I tested the flow with the new `int` type and it went well. I suspect that V3IO doesn't really care about the provided type... anyway let's leave it right now with `int`.

Eyal-Danieli:
> If you can address also: https://github.com/mlrun/mlrun/pull/4495/files#r1375935350 It will be nice

@jond01 - fixed https://github.com/mlrun/mlrun/pull/4624/commits/f31e00b3c580fcd6c49f4f0e558fcac938c3e1ca#:~:text=if%20res.status_code%20!%3D%20HTTPStatus.OK.value%3A

mlrun/datastore/targets.py
```diff
@@ -556,7 +556,7 @@ def write_dataframe(
                             break
                 # Partitioning will be performed on timestamp_key and then on self.partition_cols
                 # (We might want to give the user control on this order as additional functionality)
-                partition_cols += self.partition_cols or []
+                partition_cols = (self.partition_cols or []) + partition_cols
```
assaf758, line 559: why do we need to change this? don't we create the partition dirs "manually"?

mlrun/model_monitoring/controller_handler.py
```diff
@@ -14,19 +14,19 @@
 #
 
 import mlrun
-from mlrun.model_monitoring.batch_application import BatchApplicationProcessor
+from mlrun.model_monitoring.controller import MonitoringApplicationProcessor
```
assaf758, line 17: since we decided on "Controller", let's change processor to controller (ex in MonitoringApplicationProcessor)

mlrun/model_monitoring/application.py
```diff
@@ -35,25 +36,28 @@ class ModelMonitoringApplicationResult:
...
-    :param endpoint_id:          (str) ID of the monitored model endpoint.
-    :param schedule_time:        (pd.Timestamp)Timestamp of the monitoring schedule.
-    :param result_name:          (str) Name of the application result.
-    :param result_value:         (float) Value of the application result.
-    :param result_kind:          (ResultKindApp) Kind of application result.
-    :param result_status:        (ResultStatusApp) Status of the application result.
-    :param result_extra_data:    (dict) Extra data associated with the application result.
+    :param application_name:      (str) Name of the model monitoring application.
+    :param endpoint_id:           (str) ID of the monitored model endpoint.
+    :param start_processing_time: (pd.Timestamp) Start time of the monitoring schedule.
```
assaf758, line 41: so start_processing_time would equal last_period_analyzed per the specific app?
end_processing_time = last_processing_time = last_ready_period + base_period?
need to be consistent if the ranges are  "open" or "close" (ie do they include the limit points themselves)

mlrun/model_monitoring/controller.py
```diff
@@ -350,25 +350,29 @@ def _delete_old_parquet(
...
-                    fs.rm(path=file["name"], recursive=True)
-                    # Delete directory
-                    fs.rmdir(path=file["name"])
+            for endpoint in endpoints:
+                parquet_files = fs.listdir(
+                    path=f"{self.parquet_directory}"
+                    f"/key={endpoint[mlrun.common.schemas.model_monitoring.EventFieldType.UID]}"
+                )
+                for file in parquet_files:
+                    if file["mtime"] < time_to_keep:
```
assaf758: why not iterate on the dirnames to understand the processing time of each?
also, if you want, you can't assert that file times are what you expect (and log an error if they don't) but we should still delete - otherwise we will also fail deleting the dir and we are "leaking" files.

mlrun/datastore/targets.py
Eyal-Danieli: We don't need it (old leftovers) :) 
I'll remove it

mlrun/model_monitoring/controller_handler.py
Eyal-Danieli: https://github.com/mlrun/mlrun/pull/4624/commits/c18f2328dc64ac10063eaf59b6ffeb7968bb7743

mlrun/model_monitoring/controller.py
```diff
@@ -312,107 +308,79 @@ def _parse_batch_dict_str(self):
...
+
+            target = mlrun.datastore.targets.BaseStoreTarget(
+                path=self.parquet_directory
+            )
+            store, _ = target._get_store_and_path()
+            fs = store.get_filesystem()
+
+            # calculate time threshold (keep only files from the last 24 hours)
+            time_to_keep = float(
+                (datetime.datetime.now() - datetime.timedelta(days=days)).strftime("%s")
```
jond01, line 326: Question: do we intend to keep using `datetime.datetime.now()` here?
It is plausible unless the system really crashes.

mlrun/model_monitoring/controller.py
```diff
@@ -436,58 +405,51 @@ def _push_to_applications(
...
-        storage_options,
-    ):
+        feature_set: mlrun.common.schemas.FeatureSet,
+        endpoint_id: str,
+        start_infer_time: datetime.datetime,
+        end_infer_time: datetime.datetime,
+        parquet_directory: str,
+        storage_options: dict,
+        application_name: str,
+    ) -> mlrun.feature_store.OfflineVectorResponse:
```
jond01, line 415: Great improvement here 👍  It's much easier to understand now

mlrun/model_monitoring/controller.py
```diff
@@ -436,58 +405,51 @@ def _push_to_applications(
...
-            start_time=start_time,
-            end_time=end_time,
+            start_time=start_infer_time,
+            end_time=end_infer_time,
             timestamp_for_filtering=mm_constants.EventFieldType.TIMESTAMP,
             target=ParquetTarget(
-                path=parquet_directory,
-                time_partitioning_granularity="minute",
+                path=parquet_directory
+                + f"/key={endpoint_id}/{start_infer_time.strftime('%s')}/{application_name}.parquet",
```
jond01, line 448: Minor: I suggest keeping this string as a class constant and formatting it here

jond01 (approved):

mlrun/model_monitoring/controller.py
```diff
@@ -201,80 +201,76 @@ def model_endpoint_process(
...
+                # Continue if not enough events provided since the deployment of the model endpoint
+                except FileNotFoundError:
                     logger.warn(
-                        "Not enough model events since the beginning of the batch interval",
-                        featureset_name=m_fs.metadata.name,
+                        "Parquet not found, probably due to not enough model events",
                         endpoint=endpoint[mm_constants.EventFieldType.UID],
                         min_rqeuired_events=mlrun.mlconf.model_endpoint_monitoring.parquet_batching_max_events,
-                        start_time=start_time,
-                        end_time=end_time,
```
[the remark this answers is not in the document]
Eyal-Danieli: good catch

mlrun/model_monitoring/controller.py
Eyal-Danieli: I don't think it's an issue (assuming that the controller timezone is not changing after the crash). The controller creates the parquets and manage the removal process based on the same timezone.

assaf758 (approved):

PR #4628, opened 2023-11-23
[Runtimes] Remove requirement to provide `image` in `set_function` when using `pull_at_runtime`

theSaarco:
Prior to this fix, the following code would explode due to `set_function` requiring to provide `image`:
```python
project: MlrunProject = mlrun.get_or_create_project(
    name="<project name>", context="./"
)

project.set_source(
    "<git source>",
    pull_at_runtime=True,
)

# KubeJob
func = project.set_function(
    func="code.py",
    name="my-function",
    handler="handler",
    with_repo=True,
    kind="job",
)

...

ValueError: image must be provided with py code files which do not run on 'local' engine kind
```
However, when working with `pull_at_runtime` there is no need to build an image to provide the code (of course, it's still needed when there are extra requirements etc.) as it will be provided at runtime.
This PR fixes this flow, and also fixes some validations in the general build flow across launchers.

[ML-4994](https://jira.iguazeng.com/browse/ML-4994)


mlrun/launcher/remote.py
```diff
@@ -90,9 +90,11 @@ def launch(
                 runtime.deploy(skip_deployed=True, show_on_failure=True)
 
             else:
-                raise mlrun.errors.MLRunRuntimeError(
-                    "Function image is not built/ready, set auto_build=True or use .deploy() method first"
-                )
+                if not runtime.spec.build.load_source_on_run:
```
liranbg, line 93: while this is a warning implying that we might fail, and suggest how to handle when it would fail - perhaps worth adding why we _dont_ raise an exception here and for which cases it is a valid flow

liranbg (approved):

mlrun/launcher/remote.py
alonmr, line 93: I think that checking `load_source_on_run` is not enough. You need to do something like
~~~
require_build = (
            build.commands
            or build.requirements
            or (build.source and not build.load_source_on_run)
        )
~~~
like in `prepare_image_for_deploy`. Maybe extract it to a method under BaseRuntime

server/api/launcher.py
```diff
@@ -268,6 +278,15 @@ def _validate_runtime(
             )
 
         self._validate_state_thresholds(run.spec.state_thresholds)
+
+        if (
+            mlrun.runtimes.RuntimeKinds.requires_image_name_for_execution(runtime.kind)
+            and not runtime.spec.image
```
alonmr, line 284: why not use `is_deployed`? because of spark?

alonmr (commented):
Minor comments/questions

server/api/launcher.py
theSaarco: Yes, Spark is behaving differently and has a `use_default_image` field in the `Spark3JobSpec` class. That's also why `requires_image_name_for_execution` doesn't just look for local kinds but also for Spark.

mlrun/launcher/remote.py
liranbg: I would suggest not to, leaving the server side to be the opinionated one.
perhaps only worth adding a comment that you "alerting the user and that it is up to the server to really validate it"

theSaarco: So that's exactly what's being done here. I rephrased the message a bit to make it clear execution is about to fail.

mlrun/runtimes/base.py
```diff
@@ -800,6 +800,14 @@ def clean_build_params(self):
...
         return self
 
+    def requires_build(self) -> bool:
+        build = self.spec.build
+        return (
+            build.commands
+            or build.requirements
+            or (build.source and not build.load_source_on_run)
+        )
+
```
liranbg, line 810: 🎖️

theSaarco: Credit goes to @alonmr for the suggestion 👏

alonmr (approved):
🔥

PR #4637, opened 2023-11-27
[Auth] Add missing permission check when scheduling workflows

theSaarco:
The new `submit_workflow` endpoint allows scheduling workflows, but doesn't check for `create` permissions on the `schedule` resource. Add this check if a schedule was provided, like we do in `submit_job`.


yonishelach (approved):
LGTM

PR #4652, opened 2023-11-28
[Feature Store] Fix issues with default `start_time` and `end_time`

gtopper:
Fixes `TestFeatureStore::test_passthrough_feature_set` as well as [ML-5145](https://jira.iguazeng.com/browse/ML-5145).

### DIFF
--- a/mlrun/datastore/utils.py
+++ b/mlrun/datastore/utils.py
@@ -93,10 +93,6 @@ def filter_df_start_end_time(
 ) -> typing.Union[pd.DataFrame, typing.Iterator[pd.DataFrame]]:
     if not time_column:
         return df
-    if not start_time and not end_time:
-        df[time_column] = pd.to_datetime(df[time_column])
-        return df
-
     if isinstance(df, pd.DataFrame):
         return _execute_time_filter(df, time_column, start_time, end_time)
     else:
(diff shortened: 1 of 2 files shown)


mlrun/datastore/utils.py
```diff
@@ -91,8 +91,12 @@ def filter_df_start_end_time(
     start_time: pd.Timestamp = None,
     end_time: pd.Timestamp = None,
 ) -> typing.Union[pd.DataFrame, typing.Iterator[pd.DataFrame]]:
-    if not time_column or (not start_time and not end_time):
+    if not time_column:
         return df
+    if not start_time and not end_time:
+        df[time_column] = pd.to_datetime(df[time_column])
```
assaf758, line 97: can you explain more why you need to do this?
do we need to this always? ie without respect to the type of df[time_column]?

gtopper: This code is executed in `_execute_time_filter`. Actually, it's not necessary here, and not correct in all cases. The right solution is to remove `or (not start_time and not end_time)` from the condition, so `_execute_time_filter` will still run.

Either way, the reason is that we need to cast the time column to a datetime type, as it may be a string.

assaf758 (approved):

PR #4653, opened 2023-11-28
[Requirments] Bump storey to v1.6.7

assaf758:


liranbg:
@assaf758  need to remove `pyarrow>=10.0, <13` from requirements.txt as well

alonmr (commented):
Why remove it completely? Seems like we use it on datastore. If its a dependency of another package can we rely on it? Seems unlikely

PR #4659, opened 2023-11-28
[API] Reduce memory footprint of runs monitoring [1.5.x]

alonmr:
https://jira.iguazeng.com/browse/ML-5148

### DIFF
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -99,7 +99,7 @@ default_config = {
     "runs_monitoring_missing_runtime_resources_debouncing_interval": None,
     "monitoring": {
         "runs": {
-            "list_runs_last_update_time_from": 7,  # days
+            "list_runs_time_period_in_days": 7,  # days
         }
     },
     # the grace period (in seconds) that will be given to runtime resources (after they're in terminal state)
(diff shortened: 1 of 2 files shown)


mlrun/config.py
```diff
@@ -97,6 +97,11 @@
     # runs monitoring debouncing interval in seconds for run with non-terminal state without corresponding k8s resource
     # by default the interval will be - (runs_monitoring_interval * 2 ), if set will override the default
     "runs_monitoring_missing_runtime_resources_debouncing_interval": None,
+    "monitoring": {
+        "runs": {
+            "list_runs_last_update_time_from": 7,  # days
```
theSaarco, line 102:
```suggestion
            "list_runs_time_period_in_days": 7,  # days
```
It's not really time-from, the time from is `now` - this parameter.
Also, I think having a `runs` sub-section inside the `monitoring` section is probably an overkill...

theSaarco (approved):
LGTM, minor comment on the parameter name.

mlrun/config.py
alonmr: 👍 about 1
about 2, might be overkill but I think it leaves room for configuring monitoring of other resources and it makes a more structured and organized code. we also already did it in development so better to align it here with the same config path.

PR #4664, opened 2023-11-29
[FeatureStore] Change APIs to be FeatureSet/FeatureVector class methods

alxtkr77:
https://jira.iguazeng.com/browse/ML-4622
Add APIs to FeatureSet and FeatureVector

class FeatureSet:
  ingest() // new method - shadows mlrun.feature_store.api.ingest 
  preview() // new method - shadows mlrun.feature_store.api.preview
  deploy_ingestion_service() - shadows mlrun.feature_store.api.deploy_ingestion_service_v2

class FeatureVector:
  get_offline_features() // new method - shadows mlrun.feature_store.api.get_offline_features 
  get_online_feature_service() new method - shadowsmlrun.feature_store.api.get_online_feature_service

### DIFF
--- a/mlrun/feature_store/feature_set.py
+++ b/mlrun/feature_store/feature_set.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 import warnings
 from datetime import datetime
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 import pandas as pd
 import pytz
@@ -23,6 +23,7 @@ import mlrun
 import mlrun.common.schemas
 
 from ..config import config as mlconf
+from ..data_types import InferOptions
 from ..datastore import get_store_uri
 from ..datastore.sources import BaseSourceDriver, source_kind_to_driver
 from ..datastore.targets import (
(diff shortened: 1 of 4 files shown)


assaf758 (approved):

mlrun/feature_store/feature_set.py
```diff
@@ -982,6 +984,56 @@ def reload(self, update_spec=True):
...
+        self,
+        source,
+        entity_columns: list = None,
+        namespace=None,
+        options: InferOptions = None,
+        verbose: bool = False,
+        sample_size: int = None,
+    ) -> pd.DataFrame:
+        return mlrun.feature_store.api.preview(
+            source, entity_columns, namespace, options, verbose, sample_size
```
theSaarco, line 1022: Where is `self` (should be 1st parameter)

tests/feature-store/test_featureset.py
```diff
@@ -0,0 +1,130 @@
...
+
+    # Assert that mlrun.feature_store.api.preview was called with the correct parameters
+    mock_preview.assert_called_once_with(
+        test_source,
+        test_entity_columns,
+        test_namespace,
+        test_options,
+        test_verbose,
+        test_sample_size,
+    )
```
theSaarco, line 100: Wrong expected param-list. `fs` should be first param.

theSaarco (commented):
The preview API is missing parameter and won't work.
The test doesn't catch it since it looks for a (wrong) param list and doesn't verify functionality.
I suggest going to functional FS tests, and modifying them to support calling both the global API and the object-bases APIs (I think these would be mostly system tests, but still worth it to catch these errors).

PR #4668, opened 2023-11-29
[Pipelines] Add KFP 2.0 compatibility on SDK and API levels

laurybueno:
[ML-4717](https://jira.iguazeng.com/browse/ML-4717)

**The work here isn't final, but it's shared to gather feedback and flag risks.** 

This codebase can compile a local pipeline with KFP SDK 2.0 (in the new IR YAML format), dispatch it to the MLRun API, which enriches it, and pass it along to the KFP API. 

### Useful references
Information about the Python import system and the meta path: 
https://docs.python.org/3/reference/import.html#the-meta-path

List of KFP 1.8~2.0 compatibility issues that we need to keep in mind
https://www.kubeflow.org/docs/components/pipelines/v2/migration/#containerop-support

[ML-4717]: https://iguazio.atlassian.net/browse/ML-4717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/pipelines/kfp/v1_8/__init__.py
liranbg, line 1: reminder to add license content on top of every newly added file

mlrun/projects/pipelines.py
```diff
@@ -537,6 +469,7 @@ def get_run_status(
         pass
 
 
+# TODO: should we move this entire class to a version specific KFP module? Either that or keep dual-compatibility
 class _KFPRunner(_PipelineRunner):
```
liranbg, line 473: I dont think the entire class needs to move but rather the implementation
this unit should hold only mlrun information (sdk) and the impl should be used from the newly pipelines folder

liranbg (commented):
Add the license to each added file
I think overall, small PRs to refactor moving pipelines logic slowly is the best practice here
rest of TODOs might be achievable using following up PRs

mlrun/pipelines/kfp/v1_8/__init__.py
laurybueno: Thanks!

mlrun/projects/pipelines.py
laurybueno: That's a good point.

It seems to me that the main KFP usage points here are the `new_pipe_metadata` usage (which is already a function on the new pipelines module) and the `Compiler()` call under the `save` method. 

Since `Compiler` is also available under KFP 2.0, this might be a case of code that works on both versions. If, later down the line, we find out that this is not the case, we should be able to isolate it on the pipelines module without much trouble.

liranbg (approved):

mlrun/__init__.py
```diff
@@ -35,7 +35,7 @@
 from .execution import MLClientCtx
 from .model import RunObject, RunTemplate, new_task
 from .package import ArtifactType, DefaultPackager, Packager, handler
-from .platforms import VolumeMount, auto_mount, mount_v3io, v3io_cred
+from .platforms import VolumeMount, auto_mount
```
theSaarco, line 38: This will break backwards-compatibility for user's existing code. Anyone currently doing something like -
```
import mlrun
function.apply(mlrun.mount_v3io())
```
Will have broken code, and will need to adjust to the new import path.
So, unless there's a specific reason why this import cannot be here, these imports are needed here (instead of importing from `.platforms` adjust to the new location).

mlrun/pipelines/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class PipelineEngineModuleFinder(MetaPathFinder):
+    @staticmethod
+    def _resolve_module_path(fullname, path):
+        path_prefix = path[0].replace(
+            "mlrun/pipelines",
+            _pipeline_module_locations.get(PIPELINE_COMPATIBILITY_MODE),
+        )
+        path_suffix = fullname.replace("mlrun.pipelines", "")
```
theSaarco, line 43: From a somewhat surface-level read of the path-finder mechanism, I'm not sure how this will work with imports of relative paths? Specifically if importing `..pipelines.something` - this `replace` will not do anything, and it seems that your next replace (`.`->`/`) will change this to something like `//pipelines/something` which doesn't seem to be what we want? You need to replace `..` to `../`, don't you?

mlrun/pipelines/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+
+import sys
+from importlib.abc import MetaPathFinder
+from importlib.util import spec_from_file_location
+
+import kfp
+
+
+def resolve_pipeline_engine():
+    if kfp.__version__.startswith("2.0"):
```
theSaarco, line 24: Maybe `2.0` is too strict? Just thinking that if at some point there'll be a `2.1` version to kfp, it'll probably still need the `kfp-v2.0` module (hopefully it's BC) but this `if` will not catch it and will return `kfp-v1.8`.
Why not being stricter about this check? If it's 2.0, return v2.0, if it's exactly 1.8 return v1.8, else explode. This way at least the developers will be immediately aware when such changes happen.

mlrun/platforms/other.py
```diff
@@ -23,7 +23,7 @@
 from mlrun.errors import MLRunInvalidArgumentError
 from mlrun.utils.helpers import logger
 
-from .iguazio import mount_v3io
+from ..pipelines.iguazio import mount_v3io
```
theSaarco, line 26: Maybe I'm missing something - this file is full of modifiers that use `container_op` for their purposes and reference `kfp.dsl`. Shouldn't they also be moved to the `pipelines` directory like you did for `mount_v3io` and its friends above? Or did you just miss this?

theSaarco (changes requested):
In general this looks like a good approach that will let us have multiple kfp implementations based on version.
The mechanism itself seems legit, as far as I can tell, though I'm not sure how it handles `..` paths. It may be my lack of understanding of the actual internals there.
Also, a concern I have right now is that this change is not BC in terms of user code - this also triggered many changes in this PR to adjust, but will require users to do it as well. See my comments.

mlrun/__init__.py
laurybueno: I didn't realize these were imported by our users from there. I added them back. Thanks, Saar!

mlrun/pipelines/__init__.py
laurybueno: This works with relative imports. Before calling the finders, Python's import system completes the import string.

However, based on your and Liran's feedback, I've changed this import strategy to one that favors static analyzers such as `lint-imports` and PyCharm's autocomplete. It's not completely ready yet, but you can already take a look and give me your thoughts (I've pushed it to this PR).

mlrun/platforms/other.py
laurybueno: These absolutely will be moved. I didn't get around to doing it yet.

laurybueno:
@liranbg @theSaarco Thanks a lot for your early reviews!

I've pushed a new code split strategy for isolating version-specific KFP code as namespaced Python packages (look at the `pipeline-providers` folder).

I'm still working on the installation method for these, but I'll most likely bundle the files on MLRun's build wheel so that KFP 1.8 code acts as default and 2.* can be installed as an extra requirement (`pip install mlrun[kfp2]`). I'm still implementing it, but your feedback is much appreciated already.

laurybueno:
Closed in favor of https://github.com/mlrun/mlrun/pull/5417

PR #4670, opened 2023-11-29
[CLI] Fix local source file in yaml when building function

yanburman:
ML-2007


alonmr (commented):
It's worth adding a system test of build with CLI from function yaml.
See `tests.system.runtimes.test_kubejob.TestKubejobRuntime.test_run_cli_watch_remote_job` for reference but it will use build instead of run of course (don't think we have one like that yet)

PR #4672, opened 2023-11-29
[Artifacts] Improve object hash calculation

TomerShor:
In this PR:
- Remove `labels` and `db_key` from the artifact dict when calculating object hash - When updating an artifact, we need the object hash to be the same as the existing artifact's hash. Including both in the calculation might case a duplicate record to be created instead of updating.
- Don't add `tag` to the artifact dict when creating a new artifact - This solves an issue when the `latest` tag remained in the `metadata` of every newly created artifact, even if the tag itself pointed to a different artifact.
When getting artifacts, the tag is added on runtime from the actual tag in the DB.

Resolves https://jira.iguazeng.com/browse/ML-5153

### DIFF
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -944,18 +944,21 @@ def fill_object_hash(object_dict, uid_property_name, tag=""):
     return uid
 
 
-def fill_artifact_object_hash(
-    object_dict, uid_property_name, iteration=None, producer_id=None
-):
+def fill_artifact_object_hash(object_dict, iteration=None, producer_id=None):
     # remove status, tag, date and old uid from calculation
     object_dict.setdefault("metadata", {})
     tag = object_dict["metadata"].get("tag", None)
+    labels = object_dict["metadata"].pop("labels", None)
     status = object_dict.setdefault("status", {})
     object_dict["metadata"]["tag"] = ""
-    object_dict["metadata"][uid_property_name] = ""
+    object_dict["metadata"]["uid"] = ""
     object_dict["status"] = None
     object_updated_timestamp = object_dict["metadata"].pop("updated", None)
     object_created_timestamp = object_dict["metadata"].pop("created", None)
+
+    # if the artifact is first created, it will not have a db_key, so we need to pop it from the spec
+    # so further updates of the artifacts will have the same hash
+    db_key = object_dict.get("spec").pop("db_key", None)
 
     # make sure we have a key, producer_id and iteration, as they determine the artifact uniqueness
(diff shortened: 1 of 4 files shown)


mlrun/utils/helpers.py
```diff
@@ -944,19 +944,22 @@ def fill_object_hash(object_dict, uid_property_name, tag=""):
     return uid
 
 
-def fill_artifact_object_hash(
-    object_dict, uid_property_name, iteration=None, producer_id=None
-):
+def fill_artifact_object_hash(object_dict, iteration=None, producer_id=None):
```
alonmr, line 947: you can extract the artifact specific attributes needed (db_key, tree, etc.) and then reuse `fill_object_hash` above.

server/api/db/sqldb/db.py
```diff
@@ -758,13 +757,14 @@ def list_artifact_tags(
         artifacts = self.list_artifacts(session, project=project, category=category)
         results = []
         for artifact in artifacts:
-            results.append(
-                (
-                    project,
-                    artifact["spec"].get("db_key"),
-                    artifact["metadata"].get("tag"),
+            if artifact["metadata"].get("tag"):
```
alonmr, line 760: why ignore artifacts without tags? maybe add a comment to explain as well

alonmr: please add a comment why you add the artifacts only if they have tags.

alonmr (commented):
IIUC now if you create artifact A with tag latest and then create artifact B with the same key and no tag it will take the latest tag from artifact A.
Then artifact A has no tag and it will not show in UI or when you list artifacts.
I would expect artifact B to override artifact A in such case and not leave dangling artifacts behind.

TomerShor:
> IIUC now if you create artifact A with tag latest and then create artifact B with the same key and no tag it will take the latest tag from artifact A. Then artifact A has no tag and it will not show in UI or when you list artifacts. I would expect artifact B to override artifact A in such case and not leave dangling artifacts behind.

@alonmr This is not changing the current behavior - when you create 2 (or more) artifacts with the same key, only the last one created will have the tag "latest" (because it is the latest 😉 ).

Why do you say that an artifact without a tag won't show up in the UI? I can see that the UI is querying artifacts without tag filtering: `GET /artifacts?category=other&format=full&best-iteration=true`

In your example - artifact A will remain in the system, just without the latest tag - as it should.

server/api/db/sqldb/db.py
alonmr: my bad :) thought it was listing artifacts.. can delete

PR #4676, opened 2023-11-30
[FeatureStore] Fixing FeatureSet.preview()

alxtkr77:
Fixing error in passing parameters from FeatureSet.preview() to the api.preview() introduced in https://github.com/mlrun/mlrun/pull/4664
In addition to unit tests some system tests are set to use the new FeatureSet/FeatureVector  functions


theSaarco (approved):
👍

assaf758 (approved):

PR #4685, opened 2023-12-03
[API] Reduce memory footprint of runs monitoring [1.3.x] 

theSaarco:
Back-porting 2 PRs to 1.3.x:
#4659 
#4662 

Back-porting had to be done manually due to the significant changes between 1.5.x and 1.3.x. As part of the difficulty to back-port, didn't include the unit test added in #4662.


liranbg:
putting aside for now

PR #4700, opened 2023-12-05
[Serving] Fix YAML serialization error in serving and ingestion service

gtopper:
[ML-5108](https://jira.iguazeng.com/browse/ML-5108)

### DIFF
--- a/mlrun/datastore/targets.py
+++ b/mlrun/datastore/targets.py
@@ -877,7 +877,7 @@ class ParquetTarget(BaseStoreTarget):
         else:
             storage_options = storage_options or self.storage_options
 
-        graph.add_step(
+        step = graph.add_step(
             name=self.name or "ParquetTarget",
             after=after,
             graph_shape="cylinder",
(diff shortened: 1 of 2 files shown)


mlrun/datastore/targets.py
```diff
@@ -891,6 +891,16 @@ def add_writer_step(
...
+            dict_fields=[
+                "path",
+                "columns",
+                "index_cols",
+                "partition_cols",
+                "time_field",
+                "storage_options",
+                "max_events",
+                "flush_after_seconds",
+            ],  # update_last_written is not serializable, so we must exclude it (ML-5108)
```
assaf758, line 903: maybe we can build this list once and store into the class? would also help in keeping this list in sync with the __init__() param list (as will be closer together)
even nicer if you can come with some python meta trick to create this list automatically)

gtopper: Well, the meta trick is what it does if you don't pass it the list 😅. I'll see what I can do.

What I'd really want to do here is not to serialize the writer step at all (because why should be? It's derived from the class – e.g. `storey.ParuqetTarget` from `ParquetTarget`). But I'm afraid that's a much bigger and riskier change to make.

assaf758 (approved):

PR #4709, opened 2023-12-07
[Model Monitoring] Support running apps on "batch infer" results

jond01:
<TBA>


mlrun/model_monitoring/api.py
```diff
@@ -164,6 +165,17 @@ def record_results(
...
+                                     infrastructure in place (i.e. stream path is empty).
+                                     If True, mark the current monitoring window as completed - meaning that
+                                     the current data will be used in the monitoring, without waiting for the
+                                     next monitoring window to start.
+                                     You may want to set it to False if you have multiple results to record, and this
+                                     result is not the last one in the current batch set. In that case, pass True only
+                                     for the last call to this function.
+                                     When the model endpoint has a model monitoring infrastructure in place, this flag
+                                     will produce a warning and will be ignored. Set `last_in_batch_set` to `None` to
+                                     resolve this warning.
```
assaf758, line 178: I suggest some changes in the flag description.
```suggestion
    :param last_in_batch_set:
                                     This flag can (and should only) be used  when the model endpoint does not have a model                 monitoring set.
                                     If True (the default), mark the current monitoring window (on this monitoring endpoint) as completed - meaning that the data inferred so far is assumed to be the total data for this monitoring window.
                                     You may want to set this flag to False if you want to record multiple results in close time proximity ("batch set"). In this case, set this flag to False on all but the last batch in the set.
```

mlrun/model_monitoring/helpers.py
```diff
@@ -89,3 +96,49 @@ def get_connection_string(secret_provider: typing.Callable = None) -> str:
...
+    seconds_delta: int = 2,
+) -> None:
+    """
+    Update the last request field of the model endpoint to be after the current last request time.
+
+    :param project:         Project name.
+    :param model_endpoint:  Model endpoint object.
+    :param db:              DB interface.
+    :param minutes_delta:   Minutes delta to add to the last request time.
+    :param seconds_delta:   Seconds delta to add to the last request time. This is mainly to ensure that the last
```
assaf758, line 115: If I understand correctly, I think the 2 seconds adds a risk of skipping the next period:
if the batch-infer was at 9:59, we will set last_request to 20:01 causing us to "skip" the [10,20] period.

jond01: I think that changing it to 1 second will be ok from both sides.
BTW, I realized that what we really want to bump is "last_updated", so we should add to this delta the 60 seconds wait time

mlrun/model_monitoring/api.py
jond01: fda8d70

assaf758 (approved):

PR #4710, opened 2023-12-07
[Docs] Add creating and using a custom/private hub

jillnogold:
[ML-3370](https://jira.iguazeng.com/browse/ML-3370)

### DIFF
--- a/docs/runtimes/load-from-hub.md
+++ b/docs/runtimes/load-from-hub.md
@@ -35,7 +35,9 @@ Read [CONTRIBUTING.md](https://github.com/mlrun/functions) to learn how to creat
 Make sure your hub source is accessible via github (private is also possible).
 ```
 
-To create a function hub from scratch, the hub structure must be the same as the [MLRun hub](https://github.com/mlrun/marketplace):
+To create a function hub from scratch, the hub structure must be the same as the [MLRun hub](https://github.com/mlrun/marketplace).
+
+The hierarchy must be:
 
 - functions directory
 	- channels directories
--- a/docs/runtimes/load-from-hub.md
+++ b/docs/runtimes/load-from-hub.md
@@ -61,19 +61,26 @@ The hierarchy must be:
 
 
 ### Add a custom hub to the MLRun database
-To add your hub to your MLRun database, run:
+When you add a hub, specify `order=-1` to add it to the top of the list. 
+The list order is relevant when [loading a function](#load-function-example-omit-in-toc):
+if you don't specify a source name, MLRun starts searching for the function with the last added hub.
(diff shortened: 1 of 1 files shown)


yonishelach (approved):
Looks good!

docs/runtimes/load-from-hub.md
```diff
@@ -4,29 +4,84 @@
...
+private_source = mlrun.common.schemas.IndexedHubSource(
+    order=-1,
+	source=mlrun.common.schemas.HubSource(
+		metadata=mlrun.common.schemas.HubObjectMetadata(
+		name="priv", description="a private source"
+		),
+		spec=mlrun.common.schemas.HubSourceSpec(
+		path="/local/path/to/source", channel="development"
+		),
+	)
```
alonmr, line 78: I don't really follow the example and if I struggle, I believe it needs some clarifications.
1. `order=-1` means that it replaces the default IIUC. It should be specified (@yonishelach please confirm)
2. what is local path to source? I expected this to be a url to a git repo. Does local path means its available only for the current user? why do you need a channel if the path is local? @yonishelach
3. please fix the source name to make it a proper example (is priv short for private?)

yonishelach: I took it from the documentation of `create_hub_source` so maybe it is need to be improved as well or I am missing something too. According the documentation there:
```
The global hub source always exists in the list, and is always the last source
(`order = -1`). It cannot be modified nor can it be moved to another order in the list.
```
So is it impossible to assign order=-1 at all?
@theSaarco can you please help us here?

```suggestion
# Add a private source as the last one (will be #1 in the list) TODO: need to be resolved
private_source = mlrun.common.schemas.IndexedHubSource(
    order=-1,
	source=mlrun.common.schemas.HubSource(
		metadata=mlrun.common.schemas.HubObjectMetadata(
		name="private", description="a private source"
		),
		spec=mlrun.common.schemas.HubSourceSpec(
		path="https://mlrun.github.io/marketplace", channel="development"
		),
	)
```

theSaarco: The problem is that you're quoting just part of the docstring. The `create_hub_source` docstring says:
```
MLRun maintains an ordered list of hub sources (“sources”) Each source has
its details registered and its order within the list. When creating a new source, the special order ``-1``
can be used to mark this source as last in the list. However, once the source is in the MLRun list,
its order will always be ``>0``.

The global hub source always exists in the list, and is always the last source
(``order = -1``). It cannot be modified nor can it be moved to another order in the list.
```

I think this should answer your questions. The `-1` is just for the creation flow. Once the source is added it will get a different order which makes it last in the list.

theSaarco (approved):
ok

PR #4713, opened 2023-12-07
[API] Lock abortion of aborting runs

alonmr:
This block users from aborting runs multiple times while another abort run job is already running and it will also help the UI block the abort button if run state is aborting.

Adding a caching mechanism that holds the info about the running background jobs with ttl.
Sending a request to abort checks if there is already a job in the cache, if yes - check its state, if it is running return it.
Otherwise, create a new job and update the cache.

https://jira.iguazeng.com/browse/ML-5279


liranbg (changes requested):
add a unit test that ensure once run abortion failed, it finishes in failed state
in addition, i dont think that conflict should be raised because then an operation that abort a run (being aborted) will simply fail too

I think the solutino is to return the aborting run so the operation calling on it would assume it triggered it successfully. that can also have a unit test
1. create task
2. mock its abortion to last for 10 seconds
3. abort first time (in a thread, or in a non blocking way)
4. abort second time, wait on abortion to finish
5. ensure both (3) and (4) succeeded and run is now aborted

alonmr:
> i dont think that conflict should be raised because then an operation that abort a run (being aborted) will simply fail too
> 
> I think the solutino is to return the aborting run so the operation calling on it would assume it triggered it successfully. 

This is not really possible as long as the run abortion is wrapped in a background task, I would need to get the background task somehow (I don't have its ID) and then override the outer background task with the inner one 🤯 
Also, I see that there is no way to terminate a background task and even though internally we have a timeout attr, fastapi/starlette doesn't support it so maybe it's not a good idea to block the user from sending multiple abort requests

server/api/utils/cache.py
```diff
@@ -0,0 +1,131 @@
...
+        :return:    A CachedObject instance and a boolean indicating whether the object was created.
+        """
+        created = False
+        async with self.lock:
+            obj = self.cache.get(key)
+            if not obj:
+                created = True
+                obj = self._create(key, ttl, cls_kwargs)
+
+            # lock the run specific lock before releasing the global lock
```
theSaarco, line 77:
```suggestion
            # lock the object specific lock before releasing the global lock
```

server/api/utils/cache.py
```diff
@@ -0,0 +1,131 @@
...
+            await obj.lock.acquire()
```
theSaarco, line 78: This has deadlock potential - 
If someone else locked the object and trying to do something requiring the global lock (insert a new object), this context will be stuck on the object lock, while the other context will be stuck waiting for us to release the global lock.
If you keep this order of locks, then it needs to be at least documented or in some other way prevented that the global lock is acquired when an object lock is held.

server/api/utils/cache.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class Cache:
+    """
+    A cache for objects that are created asynchronously and need to be locked while in use.
+    """
+
+    def __init__(self, cls):
+        self.lock = asyncio.Lock()
+        self.cls = cls
```
theSaarco, line 51: Suggest to verify here that `issubclass(cls, CachedObject)`

server/api/api/endpoints/runs.py
```diff
@@ -417,19 +420,79 @@ async def abort_run(
...
+                    run_uid=uid,
+                    background_task_id=background_task_indicator.background_task_id,
+                )
+                return background_task
+
+            # background task is not running, override it in cache with the same lock acquired,
+            # it will generate a new id for the background task
+            background_task_indicator = await _abort_run_background_tasks_cache.create(
+                uid, cache_ttl, lock=background_task_indicator.lock
+            )
```
theSaarco, line 463: Not sure I really like this flow - what is the situation in which the background task is not running but is still in the cache? Is it that the cache ttl has not expired yet? My main concern is the reuse of the lock in two objects, which is never a good idea...
In this case, why not add an API to the cache that will `replace()` a task, and will trigger the cleanup of the previous one immediately, while creating a new object with new uid and new lock?

server/api/utils/cache.py
```diff
@@ -0,0 +1,131 @@
...
+        expiry_delayed_call: asyncio.Handle = None,
+    ):
+        self._object_id = object_id
+        self.lock = lock or asyncio.Lock()
+        self._expiry_delayed_call: typing.Optional[asyncio.Handle] = expiry_delayed_call
+
+    def matches(self, object_id: str) -> bool:
+        return self._object_id == object_id
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
theSaarco, line 39: To make sure I understand - you're using `__exit__` means that anyone using these cached object is always expected to work in a context manager fashion (i.e. `with...` objects). Right? 
Given that this is an infrastructure that other people may choose to use later, then I suggest that you add some documentation to the class with intended usage and maybe a simple code snip of utilizing it.

theSaarco (changes requested):
Looks very good - I have some comments (though not much...) on the infra and how it's used.

server/api/utils/cache.py
alonmr: nice catch, will rethink it

server/api/api/endpoints/runs.py
alonmr: good point again, the thought was to keep the run blocked for other requests coming in but now I understand the issue that the object could be expired by then unless locking the whole cache. will give the `replace` a try.

server/api/utils/cache.py
alonmr: The idea was that I needed a teardown/cleanup and was inspired by `MLClientCtx` which is not really used as a context manager. I didn't test it yet so not sure if it works. Either way I now realize that I need to ensure it is safe to cancel while someone else might have the lock acquired.

alonmr:
Using cache requires aborting runs only from chief. Either we reroute to chief like other endpoints or we need to think of something else that allows reading the run background task from both worker and chief

PR #4717, opened 2023-12-10
[Artifacts] Fix tag migrations

TomerShor:
2 fixes regarding for creating tags when migrating artifacts to the new v2 table:
1. Commit the artifacts first, before the tags and labels, so the artifact object will get a db `id` which will be set in the tags and labels.
2. Search for existing multiple tags in the DB that are attached to the artifact, instead of taking the tag name from the artifact metadata.

Fixes https://jira.iguazeng.com/browse/ML-5220

### DIFF
--- a/server/api/initial_data.py
+++ b/server/api/initial_data.py
@@ -543,7 +543,7 @@ def _migrate_artifacts_batch(
     batch_size: int,
 ):
     new_artifacts = []
-    artifacts_tags_to_migrate = []
+    old_id_to_artifact = {}
     artifacts_labels_to_migrate = []
     link_artifact_ids = []
 
@@ -637,10 +637,10 @@ def _migrate_artifacts_batch(
 
         last_migrated_artifact_id = artifact.id
 
-        # save the artifact's tags and labels to migrate them later
-        tags = _get_tag_names(db, db_session, artifact, artifact_metadata)
-        for tag in tags:
-            artifacts_tags_to_migrate.append((new_artifact, tag))
+        # keep the old tag to artifact mapping, so we can migrate the tags later
+        old_id_to_artifact[artifact.id] = new_artifact
+
+        # save the artifact's labels to migrate them later
         labels = artifact_metadata.get("labels", {})
         if labels:
             artifacts_labels_to_migrate.append((new_artifact, labels))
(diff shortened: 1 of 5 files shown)


server/api/initial_data.py
```diff
@@ -638,8 +638,8 @@ def _migrate_artifacts_batch(
         last_migrated_artifact_id = artifact.id
 
         # save the artifact's tags and labels to migrate them later
-        tag = artifact_metadata.get("tag", "")
-        if tag:
+        tags = _get_tag_names(db, db_session, artifact, artifact_metadata)
+        for tag in tags:
```
theSaarco, line 642: You're doing a query on the tag table per artifact - that's extremely non-efficient. Since you're doing this part in batches anyway, I suggest you list once all the tags whose `obj_id` is within the current batch and keep that in-memory, then use this data-structure to get tag names.
The data structure is not going to be exceedingly large since it's just for a batch of objects of limited size, and it will save you thousands of DB queries which are expensive.

theSaarco (changes requested):
Should perform the tag query in a batched way.

tests/api/db/test_artifacts.py
```diff
@@ -883,6 +904,78 @@ def test_migrate_artifacts_to_v2(db: DBInterface, db_session: Session):
...
+        # perform the migration
+        server.api.initial_data._migrate_artifacts_table_v2(db, db_session)
+
+    # validate the migration succeeded
+    query_all = db._query(
+        db_session,
+        server.api.db.sqldb.models.ArtifactV2,
+    )
+    new_artifacts = query_all.all()
+    assert len(new_artifacts) == 1
```
theSaarco, line 960:
```suggestion
    new_artifact = query_all.one()
```
`one()` asserts if query results in not exactly a single result. It also returns the object, so you don't need `new_artifacts[0]` later.

theSaarco (approved):
Looks good. A very minor suggestion.

liranbg (approved):
minor

tests/api/db/test_artifacts.py
TomerShor: Nice :)

PR #4719, opened 2023-12-10
[Artifacts] Add the option to calculate the size of artifact

Yacouby:
with the need to use the get_filestat functionality from files API, crud/files.py was created, and now both files & artifacts endpoints can use it.

resolves: 
https://jira.iguazeng.com/browse/ML-66


tests/api/api/test_artifacts.py
```diff
@@ -394,3 +396,33 @@ def test_list_artifact_with_multiple_tags(db: Session, client: TestClient):
                 ignore_order=True,
             )
         ) == {}
+
+
+def test_store_artifact_calculate_size(db: Session, client: TestClient):
```
liranbg, line 401: u can take `tmp_path` fixture instead of creating temp file and deleting it at the end

mlrun/db/base.py
```diff
@@ -91,7 +91,9 @@ def del_runs(self, name="", project="", labels=None, state="", days_ago=0):
         pass
 
     @abstractmethod
-    def store_artifact(self, key, artifact, uid, iter=None, tag="", project=""):
+    def store_artifact(
+        self, key, artifact, uid, iter=None, tag="", project="", calc_size=False
+    ):
```
liranbg, line 96: why do we need to take that argument? this granularity is too wide
calc size should happen at best effort anyway

mlrun/db/httpdb.py
```diff
@@ -763,7 +763,9 @@ def del_runs(self, name=None, project=None, labels=None, state=None, days_ago=0)
         error = "del runs"
         self.api_call("DELETE", "runs", error, params=params)
 
-    def store_artifact(self, key, artifact, uid, iter=None, tag=None, project=""):
+    def store_artifact(
+        self, key, artifact, uid, iter=None, tag=None, project="", calc_size=False
+    ):
```
liranbg, line 768: worth removing calc size

server/api/crud/artifacts.py
```diff
@@ -47,6 +50,23 @@ def store_artifact(
...
+                    try:
+                        file_stat = (
+                            server.api.crud.Files().get_filestat_with_project_secrets(
+                                auth_info, path=path
+                            )
+                        )
+                        data["spec"]["size"] = file_stat["size"]
+                    except mlrun.errors.MLRunNotFoundError:  # if the path was not found the size will be N/A
+                        logger.debug("Path was not found", path=path)
+                        pass
```
liranbg, line 68: I think this should always happen. anyway it is wrapped with try/except so worst case scenario, the artifact size would be N/A.
also, worth putting entire logic under "_resolve_artifact_size()`

server/api/crud/files.py
```diff
@@ -0,0 +1,84 @@
...
+import server.api.api.utils
+from mlrun import store_manager
+from mlrun.errors import err_to_str
+from mlrun.utils import logger
+
+
+class Files(
+    metaclass=mlrun.utils.singleton.Singleton,
+):
+    def get_filestat_with_project_secrets(
```
liranbg, line 28: what's the difference between both `get_filestat_with_project_secrets` and `get_filestat` if they both use project secrets? seems redundant. use `get_filestat` only

server/api/crud/files.py
```diff
@@ -0,0 +1,84 @@
...
+        return _get_filestat(schema, path, user, auth_info, secrets=secrets)
+
+
+def _get_filestat(
+    schema: str,
+    path: str,
+    user: str,
+    auth_info: mlrun.common.schemas.AuthInfo,
+    secrets: dict = None,
+):
```
liranbg, line 55: move under the class itself

liranbg (changes requested):

server/api/crud/artifacts.py
```diff
@@ -165,3 +171,15 @@ def _transform_artifact_struct_to_legacy_format(artifact):
         for key, value in artifact[section].items():
             legacy_artifact[key] = value
     return legacy_artifact
+
+
+def _resolve_artifact_size(data, auth_info):
```
liranbg, line 176: move this function to be part of artifact class and use `self._auto_info`
would be much easier

liranbg (commented):
minor suggestion

server/api/crud/artifacts.py
Yacouby: do you mean use self._resolve_artifact_size ?

liranbg: this function is not part of the class that uses it. so better move it to be part of the class uses it. that way, you will have direct access to the auth_info and you wont need to get it as a parameter

server/api/api/endpoints/artifacts.py
```diff
@@ -74,7 +74,7 @@ async def store_artifact(
         "Storing artifact", project=project, uid=uid, key=key, tag=tag, iter=iter
     )
     await run_in_threadpool(
-        server.api.crud.Artifacts().store_artifact,
+        server.api.crud.Artifacts(auth_info=auth_info).store_artifact,
```
liranbg, line 77: u can't pass auth for constructing the artifacts object when it is a singleton or else you will leak first inbound request authentication details for any consecutive requests.

server/api/crud/artifacts.py
```diff
@@ -57,6 +67,22 @@ def store_artifact(
             project,
         )
 
+    def _resolve_artifact_size(self, data):
+        if "spec" in data and "size" not in data["spec"]:
+            if "target_path" in data["spec"]:
+                path = data["spec"].get("target_path")
+                try:
+                    file_stat = server.api.crud.Files().get_filestat(
+                        self._auth_info, path=path
```
liranbg, line 76: this should never be a property of an object that is a singleton

liranbg (changes requested):
Important below

liranbg (approved):

PR #4727, opened 2023-12-11
[Runtimes] Databricks response part 2

tomerm-iguazio:
1) Add try catch to mlrun_log_artifact function.
2) Override the mlrun_log_artifact function if it exists in the code. This way, the user will be able to use dummy functions in their code without editing it before using mlrun.
3) Remove log_result to logs, and change the test to check logs in their new API.
4) Added unit tests for code editing (as string).
5) Added auto generated artifact's name + test.
6) Blocked returns for now.

[ML-4892](https://jira.iguazeng.com/browse/ML-4892)


tests/runtimes/test_databricks.py
```diff
@@ -11,17 +11,74 @@
...
     ):
         databricks_runtime.run(local=True)
+
+
+@pytest.mark.parametrize("user_code", [USER_CODE, USER_CODE_WITH_DUMMY])
+def test_get_internal_parameters(user_code):
+    databricks_runtime = DatabricksRuntime()
+    runobj = RunObject()
+    runobj.metadata.uid = "1"
+    print(runobj.spec.parameters)
```
assaf758, line 53: please remove

assaf758 (approved):

PR #4728, opened 2023-12-11
[API] Delete run logs when deleting run

TomerShor:
When deleting a run, delete its logs using the log collector first, and if it fails call the legacy deletion in which the api deletes the files by itself.

Also - refactored a bit of the code so deleting logs will be in the `Logs` crud, and so we can share code between runs an projects.

Resolves [ML-4883](https://jira.iguazeng.com/browse/ML-4883)

### DIFF
--- a/server/api/crud/runs.py
+++ b/server/api/crud/runs.py
@@ -15,6 +15,7 @@
 import typing
 
 import sqlalchemy.orm
+from fastapi.concurrency import run_in_threadpool
 
 import mlrun.common.schemas
 import mlrun.config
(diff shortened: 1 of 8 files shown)


server/api/crud/runs.py
```diff
@@ -260,3 +263,14 @@ def abort_run(
...
+    @staticmethod
+    async def _post_delete_run(project, uid):
+        if (
+            mlrun.mlconf.log_collector.mode
+            != mlrun.common.schemas.LogsCollectorMode.legacy
+        ):
+            await server.api.crud.Logs().stop_logs_for_run(project, uid)
+            await server.api.crud.Logs().delete_run_logs(project, uid)
+        else:
+            server.api.crud.Logs().delete_run_logs_legacy(project, uid)
```
liranbg, line 276: here too. will block eventloop

alonmr (approved):
lgtm

PR #4729, opened 2023-12-11
[API] Cascade and batch multiple runs deletion

alonmr:
Delete run now deletes all of the runtime resources (pods, crds, logs...)
That means that deleting multiple runs should take deletion of the resources into consideration which makes it a heavy operation.
This PR adds async batch runs deletion.

https://jira.iguazeng.com/browse/ML-5307

### DIFF
--- a/server/api/crud/runs.py
+++ b/server/api/crud/runs.py
@@ -226,16 +226,12 @@ class Runs(
         if (
             state
             and state
-            not in mlrun.runtimes.constants.RunStates.allowed_for_deletion_states()
+            in mlrun.runtimes.constants.RunStates.not_allowed_for_deletion_states()
         ):
             raise mlrun.errors.MLRunBadRequestError(
                 f"Can not delete runs in {state} state"
             )
-        states = (
-            [state]
-            if state
-            else mlrun.runtimes.constants.RunStates.allowed_for_deletion_states()
-        )
+
         if not runs_list:
             start_time_from = None
             if days_ago:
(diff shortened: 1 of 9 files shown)


server/api/crud/runs.py
```diff
@@ -209,19 +212,76 @@ async def delete_run(
...
+            ]:
+                tasks.append(
+                    server.api.db.session.run_function_with_new_db_session(
+                        self.delete_run,
+                        run.uid,
+                        run.iteration,
+                        run.project,
+                    )
+                )
+            results = await asyncio.gather(*tasks, return_exceptions=True)
```
TomerShor, line 267: Although the tasks run concurrently, the deletions in the DB are still done one by one. 
Perhaps we can still use `db.del_runs` that does the deletion for all runs at once?
We will need to know the UIDs of the runs that succeeded and failed the `delete_run` task.

server/api/crud/runs.py
```diff
@@ -209,19 +212,76 @@ async def delete_run(
...
+                    )
+
+            runs_list = runs_list[
+                mlrun.config.config.crud.runs.batch_delete_runs_chunk_size :
+            ]
+
+        if failed_deletions:
+            raise mlrun.errors.MLRunRuntimeError(
+                f"Failed to delete {failed_deletions} run(s). Error: {mlrun.errors.err_to_str(last_exception)}"
+            ) from last_exception
```
TomerShor, line 284: Showing only the last exception might miss some other errors.

TomerShor (commented):
Looking good overall! some minor comments

server/api/crud/runs.py
alonmr: I didn't want to make the error endless in case many runs failed to delete. Either way the errors are logged above.

TomerShor (approved):
LGTM

theSaarco (approved):
Looks good. Have some minor suggestions, but generally approved.

PR #4730, opened 2023-12-12
[Artifacts] Always add `latest` tag when creating an artifact

TomerShor:
In the `create_artifact` flow we want to add the `latest` tag regardless of whether a tag was specified or not.


alonmr (approved):

theSaarco (approved):
LGTM

PR #4731, opened 2023-12-13
[Logs] Refactor get_log & watch_log functions for better log retrieval

Yacouby:
Resolves:
https://jira.iguazeng.com/browse/ML-4749


liranbg (changes requested):
minor

PR #4732, opened 2023-12-13
[Requirments] fsspec dependencies adlfs/s3fs/gcsfs: allow picking-up patch releases of 2023.9

assaf758:


tomerm-iguazio (approved):
All tests passed:)

PR #4734, opened 2023-12-14
[Datastore] bugfix Azure delete

tomerm-iguazio:
We have removed the usage of the BlobServiceClient in the Azure Datastore.
From now on, we will rely solely on the azure fsspec class (AzureBlobFileSystem for the 'az' protocol) in the Datastore.

1) Fixed delete (rm in azure datastore) function.
2) Added tests for delete and download items.

[ML-5303](https://jira.iguazeng.com/browse/ML-5303)

### DIFF
--- a/mlrun/datastore/azure_blob.py
+++ b/mlrun/datastore/azure_blob.py
@@ -15,7 +15,6 @@
 import time
 from pathlib import Path
 
-from azure.storage.blob import BlobServiceClient
 from fsspec.registry import get_filesystem_class
 
 import mlrun.errors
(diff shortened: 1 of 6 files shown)


mlrun/datastore/azure_blob.py
```diff
@@ -15,7 +15,6 @@
 import time
 from pathlib import Path
 
-from azure.storage.blob import BlobServiceClient
```
assaf758, line 18: so we can remove azure-storage-blob from the dependencies?

assaf758 (approved):

PR #4740, opened 2023-12-14
[Linting] Add Ruff linter - replace isort & flake8

jond01:
The first part of [ML-5300](https://jira.iguazeng.com/browse/ML-5300).

- Move the configuration
- Adapt commands
- Adapt requirements
- Fix new reported errors

I recommend reviewing the changes commit by commit.

https://docs.astral.sh/ruff/linter/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -597,10 +597,9 @@ html-docs-dockerized: build-test ## Build html docs dockerized
 		make html-docs
 
 .PHONY: fmt
-fmt: ## Format the code (using black and isort)
+fmt: ## Format the code using black
 	@echo "Running black fmt..."
 	python -m black .
-	python -m isort .
 
 .PHONY: lint-imports
 lint-imports: ## Validates import dependencies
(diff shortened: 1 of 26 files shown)


Makefile
```diff
@@ -597,29 +597,27 @@ html-docs-dockerized: build-test ## Build html docs dockerized
 		make html-docs
 
 .PHONY: fmt
-fmt: ## Format the code (using black and isort)
+fmt: ## Format the code using black
```
alonmr, line 600: make fmt should also format the imports. maybe need to add ruff here as well

pyproject.toml
```diff
@@ -1,7 +1,20 @@
...
-multi_line_output = 3
-extend_skip_glob = ["*venv*", "server/api/proto"]
+[tool.ruff]
+select = [
+    "F", # pyflakes
+    "W", # pycodestyle
+    "E", # pycodestyle
+    "I", # isort
+]
+extend-exclude = [
```
alonmr, line 8: is `*venv*` not needed?

alonmr (commented):
Nice! minor comments

Makefile
liranbg: 👆

pyproject.toml
jond01: No - this pattern is in `.gitignore`, which is respected by default.
https://docs.astral.sh/ruff/settings/#respect-gitignore
https://github.com/mlrun/mlrun/blob/f4b97867841685ff91fa3e8a162030b93f68dbc6/.gitignore#L29

Makefile
jond01: Ok, I've added it to this target: 299314d.
The direction here is to stay with a single command that lints, formats, and checks - it fixes what it can and reports the overall status 🙂

alonmr (approved):

PR #4742, opened 2023-12-15
[Nuclio] Introduce API Gateways

rokatyy:
This pull request introduces Nuclio API gateways to MLRun. Within this PR:

Client-side:

* New project methods: create_or_update/get/list_api_gateways
* Introduction of the `APIGateway` entity (in `runtimes.nuclio`)

Server-side:

* Extension of the Pydantic model for APIGateway to align with the Nuclio API Gateway CRD
* Addition of new routes for storing and retrieving API gateways

Jira - https://jira.iguazeng.com/browse/ML-5371
Next steps (in following PR) described here: https://jira.iguazeng.com/browse/ML-5372

Note about tests: That's OK that backwards compatibility tests failed because it was decided to get rid of `/nuclio/` subpath in API gateway's endpoints


mlrun/runtimes/api_gateway.py
```diff
@@ -0,0 +1,160 @@
...
+
+            if not percentage_1 and percentage_2:
+                percentage_1 = 100 - percentage_2
+            if not percentage_2 and percentage_1:
+                percentage_2 = 100 - percentage_1
+            if percentage_1 and percentage_2:
+                canary = [percentage_1, percentage_2]
+            return functions, canary
+        else:
+            return None, None
```
TomerShor, line 160: Add a comment / log on why you are returning `None` here

TomerShor (changes requested):
Some first comments.

@rokatyy to answer your questions:
> 1. Where to keep api_gateways file? (it is in mlrun.runtimes now which can be not a best place)

IMO we should create a subdirectory `nuclio` inside `mlrun/runtimes` which will include all of the Nuclio relates objects and functions.
We should make sure that everything we move is backwards compatible and is accessible the same as it was before the move (i.e. not to change imports).

> The ticket indicates the addition of API gateway creation functionality. Should users also have the capability to delete these gateways?

Yes, but can be done in a different PR to keep it simple.

server/api/utils/clients/async_nuclio.py
```diff
@@ -102,3 +143,88 @@ def _handle_error_response(
...
+        password=None,
+        canary=None,
+    ) -> dict:
+        if not functions:
+            raise ValueError("functions should contain at least one object")
+        host = (
+            f"{api_gateway_name}-{project_name}.{self._nuclio_domain[self._nuclio_domain.find('.')+1:]}"
+            if not host
+            else host
+        )
```
theSaarco, line 165:
```suggestion
        host = host or f"{api_gateway_name}-{project_name}.{self._nuclio_domain[self._nuclio_domain.find('.')+1:]}"
```

A little bit shorter 😄

mlrun/runtimes/api_gateway.py
```diff
@@ -0,0 +1,170 @@
...
+            return None
+        return cls(
+            project=project,
+            name=spec.get("name"),
+            host=spec.get("host"),
+            path=spec.get("path", ""),
+            description=spec.get("description", ""),
+            functions=functions,
+            username=authentication.get("username"),
+            password=authentication.get("password"),
```
theSaarco, line 142: Do we expect to get the username/password from the BE response when listing GWs (this is where `from_dict` is used currently)?

mlrun/runtimes/api_gateway.py
```diff
@@ -0,0 +1,170 @@
...
+        if len(upstreams) == 1:
+            return [upstreams[0].get("nucliofunction", {}).get("name")], None
+        elif len(upstreams) == 2:
+            canary = [0, 0]
+            functions = [
+                upstreams[0].get("nucliofunction", {}).get("name"),
+                upstreams[1].get("nucliofunction", {}).get("name"),
+            ]
+            percentage_1 = upstreams[0].get("percentage")
+            percentage_2 = upstreams[0].get("percentage")
```
theSaarco, line 159:
```suggestion
            percentage_2 = upstreams[1].get("percentage")
```

mlrun/projects/project.py
```diff
@@ -3446,6 +3448,85 @@ def remove_custom_packager(self, packager: str):
...
+    def create_api_gateway(
+        self,
+        name: str,
+        host: str = None,
+        path: str = "",
+        description: str = "",
+        functions: Union[
+            list[Union[RemoteRuntime, ServingRuntime]],
+            Union[RemoteRuntime, ServingRuntime],
+        ] = [],
```
theSaarco, line 3460: And, I see no validation on projects. Specifically, a GW belongs to a project. Can it sit in front of functions from different projects? If not (which I believe it shouldn't), then you should also validate that the functions you get are from the same project as the one for which you're creating the GW for.

mlrun/projects/project.py
```diff
@@ -3446,6 +3448,85 @@ def remove_custom_packager(self, packager: str):
         """
         self.spec.remove_custom_packager(packager=packager)
 
+    def create_api_gateway(
```
theSaarco, line 3451: Why not get an `APIGateway` object here instead of all of its components?

server/api/api/endpoints/nuclio.py
```diff
@@ -37,7 +39,7 @@ async def list_api_gateways(
     ).list_api_gateways(project)
```
theSaarco, line 39: Another note on previously existing code - why is the router at `/projects/{project}/nuclio/api-gateways`? I feel the `nuclio` part is a bit redundant and you can work with `/projects/{project}/api-gateways`.

server/api/api/endpoints/nuclio.py
```diff
@@ -47,3 +49,37 @@ async def list_api_gateways(
         ),
         auth_info,
     )
+
+
+@router.post("/projects/{project}/nuclio/api-gateways/{gateway}")
+async def create_api_gateway(
+    project: str,
+    gateway: str,
+    functions: list = Query(alias="functions"),
```
theSaarco, line 58: Assuming we keep these params and not move to an object schema, the alias should be `function` since the REST API URL will look something like:
```
POST /projects/{project}/nuclio/api-gateways/{gateway}?function=<func1>&function=<func2>...
```
And fastapi will construct these query params into a list.
Anyway, I'm not sure why the `functions` have to be a query param at all.

mlrun/projects/project.py
rokatyy: @theSaarco I was thinking that it will be more intuitive for user to do 
`project.create_api_gateway(params..)`
 than
 ```
 gw = APIGateway(params)
 project.create_api_gateway(gw)
 ```

mlrun/projects/project.py
rokatyy: @theSaarco that's a very good point, will add validation!

mlrun/projects/project.py
```diff
@@ -3446,6 +3448,85 @@ def remove_custom_packager(self, packager: str):
...
+            username=username,
+            password=password,
+            canary=canary,
+        )
+        ok = mlrun.db.get_run_db().create_api_gateway(
+            gateway_instance, (username, password)
+        )
+        return gateway_instance if ok else None
+
+    def list_api_gateways(self) -> list[mlrun.runtimes.api_gateway.APIGateway]:
```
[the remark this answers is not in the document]
rokatyy: @theSaarco I'd add this within following PR (which will add deletion functionality), just to stay more granular. wdyt?

mlrun/runtimes/api_gateway.py
rokatyy: @theSaarco yes, username/password is returned within BE response. As an example, test `test_list_api_gateways` contains real response tested on my system.

server/api/api/endpoints/nuclio.py
rokatyy: @theSaarco I assume the initial intention behind the 'nuclio' prefix was to establish a sort of conceptual 'separation' to show that we should request nuclio api directly. 

cc: @liranbg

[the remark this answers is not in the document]
theSaarco: @rokatyy - note that the return value from the function will be serialized to JSON by fastapi anyway. Using schema'd objects helps fastapi perform validations on the objects created. It also helps the Swagger UI provide meaningful information to the users based on the schema.

mlrun/projects/project.py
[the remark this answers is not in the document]
theSaarco: No problem, agree that PRs should be kept reasonable in size.

mlrun/runtimes/api_gateway.py
theSaarco: This looks like a bad practice that goes all the way to nuclio - @liranbg , @TomerShor - are there any plans to remove credentials data from gw structured returned in nuclio APIs?

mlrun/projects/project.py
theSaarco: @rokatyy - just as a note, I'm not 100% sure we should block putting a gw in front of cross-project objects, but this needs to be properly considered before we allow such a thing (and as 1st stage implementation it makes sense to block).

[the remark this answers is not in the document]
theSaarco: Thinking a bit more about this - eventually this looks for deployed functions in nuclio, so it's possible that the name itself is enough, but this needs to be verified with what's possible in nuclio - you should be able to deploy multiple versions of the same function at the same time (or can't you?)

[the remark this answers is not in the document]
rokatyy: @theSaarco I'm not in the context in terms of how it works on the MLRun side, but on the Nuclio side, each function should have its unique name, not even within one project, but within the system.

server/api/api/endpoints/nuclio.py
rokatyy: @theSaarco reworked and passing APIGateway entity as you suggested above :)

mlrun/runtimes/api_gateway.py
rokatyy: @theSaarco @TomerShor I'll get rid of all the credentials stored in the APIGateways object and the user will have to pass authentication information when calling the function.

mlrun/runtimes/api_gateway.py
```diff
@@ -0,0 +1,170 @@
...
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "API Gateway name cannot be empty"
+            )
+
+        if canary:
+            if len(functions) != len(canary):
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    "Lengths of function and canary lists do not match"
+                )
+            if sum(canary) != 100:
```
gilad-shaham, line 89: If we're validating, we should also check each value is between 0 and 100

mlrun/runtimes/api_gateway.py
```diff
@@ -0,0 +1,170 @@
...
+        elif len(upstreams) == 2:
+            canary = [0, 0]
+            functions = [
+                upstreams[0].get("nucliofunction", {}).get("name"),
+                upstreams[1].get("nucliofunction", {}).get("name"),
+            ]
+            percentage_1 = upstreams[0].get("percentage")
+            percentage_2 = upstreams[1].get("percentage")
+
+            if not percentage_1 and percentage_2:
```
gilad-shaham, line 161: Similarly here, validate value is not negative or above 100

mlrun/projects/project.py
```diff
@@ -3446,6 +3448,102 @@ def remove_custom_packager(self, packager: str):
...
+        name: str,
+        functions: Union[
+            list[
+                Union[
+                    RemoteRuntime,
+                    ServingRuntime,
+                ]
+            ],
+            Union[RemoteRuntime, ServingRuntime],
+        ] = [],
```
gilad-shaham, line 3462: Is there a meaning to API gateway without any functions?

mlrun/projects/project.py
```diff
@@ -3446,6 +3448,102 @@ def remove_custom_packager(self, packager: str):
...
+    ) -> Union[mlrun.runtimes.api_gateway.APIGateway, None]:
+        """
+        Creates a Nuclio API Gateway. Nuclio docs here: https://docs.nuclio.io/en/latest/reference/api-gateway/http.html
+
+        :param name (str): API Gateway name.
+        :param functions (Union[list[Union[RemoteRuntime, ServingRuntime]], Union[RemoteRuntime, ServingRuntime]]):
+            List of Nuclio functions or Nuclio function.
+
+        :param host (str): API Gateway host (optional).
+        :param path (str): API Gateway path (optional).
```
gilad-shaham, line 3478: I think these 2 parameters should be better explained, and what values they get by default if not specified.
Also, for host, can the user specify any value?

gilad-shaham (commented):
Looks good, a few comments about the parameters

mlrun/projects/project.py
rokatyy: @gilad-shaham That's a very good question. I've checked in the nuclio UI, and we do not allow specifying it manually. However, technically, it's possible to set it manually when making an API request. I'll leave only the 'path' to be specified (it's possible to specify the 'path' in the UI), so we will be aligned with how it works on the UI side.

mlrun/runtimes/api_gateway.py
```diff
@@ -0,0 +1,170 @@
...
+        canary: Union[dict[str, int], None],
+    ):
+        self.project = project
+        self.name = name
+        self.host = host
+        self.functions = functions
+        self.path = path
+        self.description = description
+        self.canary = canary
+        self._nuclio_dashboard_url = mlrun.mlconf.nuclio_dashboard_url
```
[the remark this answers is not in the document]
rokatyy: @theSaarco Added host resolving logic. Currently it is based on `httpdb.list_api_gateways()` method, but added TODO to change it when `httpdb.get_api_gateway()` is implemented in the following PR

mlrun/projects/project.py
TomerShor: @rokatyy @gilad-shaham  in Nuclio, the hostname that you see in the API gateway UI comes from the following template available in the frontend spec:
```
"defaultHTTPIngressHostTemplate": "{{ .ResourceName }}-{{ .ProjectName }}.{{ .Namespace }}.app.dev65.lab.iguazeng.com"
```

This template is configurable on the Nuclio service level, and Mlrun can tell Nuclio to use this template for ingresses by passing `"@nuclio.fromDefault"` (see the code section [here](https://github.com/nuclio/nuclio/blob/development/pkg/platform/kube/platform.go#L1982-L1983)).
However I am not sure it can be done the same with API gateways. If not - perhaps an improvement is required on Nuclio's side to allow it as well.

rokatyy: @TomerShor as an option, we can allow passing empty host (it's not permitted now) and generate it on nuclio side. So mlrun won't take responsibility on this

TomerShor: @rokatyy yes that's exactly what I meant by: "an improvement is required on Nuclio's side to allow it as well."
Can be left as a TODO for now

server/api/utils/clients/async_nuclio.py
```diff
@@ -102,3 +130,81 @@ def _handle_error_response(
...
+                            "name": api_gateway.function[0],
+                        },
+                        "percentage": 0,
+                    }
+                ],
+                "host": host,
+            },
+            "metadata": {
+                "labels": {
+                    "nuclio.io/project-name": project_name,
```
TomerShor, line 169: `"nuclio.io/project-name"` can be a Constant

tests/projects/test_project.py
```diff
@@ -1464,3 +1463,189 @@ def test_load_project_from_yaml_with_function(context):
...
+def test_list_api_gateways(patched_list_api_gateways, context):
+    patched_list_api_gateways.return_value = [
+        {
+            "metadata": {
+                "name": "test",
+                "namespace": "default-tenant",
+                "labels": {
+                    "iguazio.com/username": "admin",
+                    "nuclio.io/project-name": "default",
+                },
```
TomerShor, line 1595: keep the mocks simple, non iguazio specific.
i.e. remove username label and `default-tenant`

mlrun/db/httpdb.py
```diff
@@ -3267,6 +3268,24 @@ def list_api_gateways(self, project=None):
...
+        Creates an API Gateway.
+        :param api_gateway (mlrun.runtimes.api_gateway.APIGateway): API Gateway entity.
+        :param auth (Tuple[str, str]): Pair of (username, password) if authentication is required.
+
+        @return: True if the API Gateway was created successfully, False otherwise
+        """
+        endpoint_path = f"projects/{project}/nuclio/api-gateways/{name}"
+        error = "create api gateways"
+        response = self.api_call("POST", endpoint_path, error, json=api_gateway.dict())
+        return response.ok if response else False
```
TomerShor, line 3287: when `create_api_gateway` will return the object, return it here as well

TomerShor (changes requested):
Some more comments :)

mlrun/projects/project.py
```diff
@@ -3446,6 +3448,128 @@ def remove_custom_packager(self, packager: str):
...
+        canary: Optional[list[int]] = None,
+    ) -> Union[mlrun.runtimes.api_gateway.APIGateway, None]:
+        """
+        Creates a Nuclio API Gateway. Nuclio docs here: https://docs.nuclio.io/en/latest/reference/api-gateway/http.html
+
+        :param name (str): API Gateway name.
+        :param functions (Union[list[Union[RemoteRuntime, ServingRuntime]], Union[RemoteRuntime, ServingRuntime]]):
+            List of Nuclio functions or Nuclio function.
+        :param host (str): API Gateway host name (optional). If not specified, nuclio will automatically set it
+            to `<api-gateway-name>-<project>.<system-domain>`
```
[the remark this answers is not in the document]
rokatyy: @TomerShor I fully deleted `host` param on MLRun client side (will clean up docs as well), it is being generated on server side now. It was done like this because user doesn't have `nuclio_dashboard` info on client side.

mlrun/projects/project.py
```diff
@@ -3446,6 +3448,128 @@ def remove_custom_packager(self, packager: str):
...
+                        f"Input function {func.name} is not a Nuclio function"
+                    )
+                elif hasattr(func, "metadata"):
+                    raise mlrun.errors.MLRunInvalidArgumentError(
+                        f"Input function {func.metadata.name} is not a Nuclio function"
+                    )
+                else:
+                    raise mlrun.errors.MLRunInvalidArgumentError(
+                        "Input function is not a Nuclio function"
+                    )
```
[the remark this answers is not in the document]
rokatyy: @TomerShor exactly, we will raise error here if `if func.kind not in mlrun.runtimes.RuntimeKinds.nuclio_runtimes()`, so it won't be any of those

server/api/api/endpoints/nuclio.py
```diff
@@ -47,3 +48,17 @@ async def list_api_gateways(
...
+    project: str,
+    gateway: str,
+    api_gateway: mlrun.common.schemas.APIGateway,
+    auth_info: mlrun.common.schemas.AuthInfo = Depends(deps.authenticate_request),
+):
+    await server.api.utils.clients.async_nuclio.Client(auth_info).create_api_gateway(
+        project_name=project,
+        api_gateway_name=gateway,
+        api_gateway=api_gateway,
+    )
```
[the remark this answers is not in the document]
rokatyy: @TomerShor I added it recently, but it was lost during many fixes within this PR, my bad, will return it

mlrun/projects/project.py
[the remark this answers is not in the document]
TomerShor: Oh right 🤦 
The shorter version should still apply though, with an extra check perhaps

mlrun/projects/project.py
```diff
@@ -3446,6 +3448,128 @@ def remove_custom_packager(self, packager: str):
...
+                path=path,
+                description=description,
+                username=username,
+                password=password,
+            ),
+        )
+
+        # if api gateway was created and host wasn't defined by user, we request created api gateway to resolve the host
+        if ok:
+            # resolving host name
```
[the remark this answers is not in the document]
rokatyy: @TomerShor we don't have a host parameter on mlrun client side at all

mlrun/db/httpdb.py
rokatyy: @TomerShor which object?

mlrun/runtimes/api_gateway.py
```diff
@@ -0,0 +1,128 @@
...
+import base64
+from typing import Optional, Union
+
+import requests
+
+import mlrun
+import mlrun.common.schemas
+
+BASIC_AUTH_NUCLIO_API_GATEWAY_AUTH_MODE = "basicAuth"
+NO_AUTH_NUCLIO_API_GATEWAY_AUTH_MODE = "none"
```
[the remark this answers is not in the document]
rokatyy: @TomerShor let's firstly cover those methods in our docs - https://docs.nuclio.io/en/latest/reference/api-gateway/http.html

mlrun/projects/project.py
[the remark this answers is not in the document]
TomerShor: Right, but you are not actually resolving the host in the code below the comment, you just get the api gateway object and return it.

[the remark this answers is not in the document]
rokatyy: @TomerShor yes, that's right .. will get rid of it :)

mlrun/db/httpdb.py
rokatyy: resolved as discussed to implement in the next PR

mlrun/common/schemas/api_gateway.py
```diff
@@ -0,0 +1,26 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+import typing
+
+import pydantic
+
+
+class APIGateway(pydantic.BaseModel):
+    function: typing.List[str]
```
theSaarco, line 21: Being picky here - this it's a list, call it `functions`.

mlrun/db/httpdb.py
```diff
@@ -3267,6 +3268,24 @@ def list_api_gateways(self, project=None):
         resp = self.api_call("GET", endpoint_path, error)
         return resp.json()
 
+    def create_api_gateway(
+        self,
+        project: str,
+        name: str,
+        api_gateway: mlrun.common.schemas.APIGateway,
```
theSaarco, line 3275: Note that now you have 2 schemas/representations for the `APIGateway` class - the one in the `schemas` which you reference here but it usually used in server-side code, and the one in `runtimes/api_gateway.py` which is the client-side code.
Since `httpdb` is client-side code, it's actually better to get the client-side object here. For full coverage you can do a `Union` of both classes and handle their to-dict serialization based on the class type. See `create_project` or `create_feature_set` for examples.

mlrun/projects/project.py
```diff
@@ -3446,6 +3448,123 @@ def remove_custom_packager(self, packager: str):
...
+                    "The sum of canary function percents should be equal to 100"
+                )
+        if username and not password:
+            raise mlrun.errors.MLRunInvalidArgumentError("Password is not specified")
+
+        if password and not username:
+            raise mlrun.errors.MLRunInvalidArgumentError("Username is not specified")
+
+        function_names = [func.uri for func in functions]
+        ok = mlrun.db.get_run_db().create_api_gateway(
```
theSaarco, line 3536: You changed that function to get an `APIGateway` object. Need to change here.

theSaarco, line 3536: If you're changing to create an `APIGateway` here and passing the object to the creation call, I think the validations done above (good coverage, BTW!) should go to the `APIGateway` class - this way, the code here would either get an object which is already validated, or would (if you remain with getting the parts of the GW and constructing it here) create the object here, and would fail if something is wrong. 
It makes more sense since the validation logic belongs to the API GW class, not to the project class.

server/api/api/endpoints/nuclio.py
```diff
@@ -47,3 +48,24 @@ async def list_api_gateways(
         ),
         auth_info,
     )
+
+
+@router.post("/projects/{project}/nuclio/api-gateways/{gateway}")
+async def create_api_gateway(
```
theSaarco, line 54: A general question - does the nuclio create-gw API perform create-only (i.e. fails if object already exists) logic or does it do a create-or-replace?
Reason I'm asking is that in MLRun usually we use `POST` for create-only, and `PUT` for create-or-replace, and also the endpoint function would usually be called `store_XXX` in the create-or-replace case.

mlrun/projects/project.py
theSaarco: @rokatyy - yeah, both ways make sense, and it's certainly ok to use this approach. Why not support both options? Allow passing either an `APIGateway` class or the set of params? 
There are use-cases where you already have an APIGateway object (for example - I got the existing GW from DB, modified it and now want to pass it with the changes).

theSaarco (commented):
Some additional comments, based on the revised code.

server/api/api/endpoints/nuclio.py
TomerShor: @theSaarco @rokatyy in Nuclio `POST /api_gateways` fails if an api gateway with the same name already exists.

mlrun/projects/project.py
rokatyy: @theSaarco currently we don't have a gateway entity in DB, so I think we can keep it as simple as possible and gradually complicate the logic (in following PRs)

rokatyy: But that's definitely a good idea

mlrun/db/httpdb.py
rokatyy: @theSaarco Make sense! I've moved all validation logic under ApiGateway (client) class

server/api/api/endpoints/nuclio.py
rokatyy: I've checked Nuclio's `Update` method and it updates the entity if it already exists. So, replaced `POST` with `PUT`.

mlrun/runtimes/api_gateway.py
```diff
@@ -0,0 +1,257 @@
...
+
+            if not percentage_1 and percentage_2:
+                percentage_1 = 100 - percentage_2
+            if not percentage_2 and percentage_1:
+                percentage_2 = 100 - percentage_1
+            if percentage_1 and percentage_2:
+                canary = [percentage_1, percentage_2]
+            return functions, canary
+        else:
+            # The upstream list length should be either 1 or 2; otherwise, we cannot parse the upstream value correctly
```
TomerShor, line 256:
```suggestion
            # Nuclio only supports 1 or 2 upstream functions
```

server/api/utils/clients/async_nuclio.py
```diff
@@ -43,8 +46,33 @@ async def list_api_gateways(self, project_name=None):
...
+            headers[NUCLIO_PROJECT_NAME_HEADER] = project_name
+
+        body = self._generate_nuclio_api_gateway_body(
+            project_name=project_name,
+            api_gateway_name=api_gateway_name,
+            api_gateway=api_gateway,
+        )
+
+        return await self._send_request_to_api(
+            method="PUT",
```
TomerShor, line 71: Will it create a new API gateway even if it does NOT exist?

server/api/api/endpoints/nuclio.py
TomerShor: @rokatyy what @theSaarco meant here is that in mlrun we use `create_...` [POST method] when we only want to create a new object (meaning - fail on existing), and use `store_...` [PUT method] when we want it to "create-or-update" (meaning, create if doens't exist, otherwise update).

Here you did a mix between the two - POST `create_..` but with create-or-update behavior.

TomerShor (changes requested):
Last few comments

server/api/utils/clients/async_nuclio.py
rokatyy: @TomerShor NO, it won't. Because inside of nuclio we do `get` before the resource update. Maybe we can consider moving Update to CreateOrUpdate logic in Nuclio, but for now will leave everything on `POST`

TomerShor: We can implement the Create or Update logic inside mlrun.
Let's leave this PR with `create` only, so we can move on :)

TomerShor (approved):
LGTM 🎉

mlrun/runtimes/api_gateway.py
```diff
@@ -0,0 +1,274 @@
...
+        ],
+    ):
+        if not isinstance(functions, list):
+            functions = [functions]
+
+        # validating functions
+        if not 1 <= len(functions) <= 2:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"Gateway can be created from one or two functions, "
+                f"but the number of functions is passed {len(functions)}"
```
theSaarco, line 213:
```suggestion
                f"the number of functions passed is {len(functions)}"
```

theSaarco (approved):
Looks good! 
One small change in an error string.

mlrun/db/httpdb.py
```diff
@@ -3305,6 +3306,31 @@ def list_api_gateways(self, project=None):
...
+            or :py:class:`~mlrun.common.schemas.APIGateway`: API Gateway entity.
+        :param project: project name. Mandatory if api_gateway is mlrun.common.schemas.APIGateway.
+        :param name: api gateway name. Mandatory if api_gateway is mlrun.common.schemas.APIGateway.
+        @return: True if the API Gateway was created successfully, False otherwise
+        """
+
+        if isinstance(api_gateway, mlrun.runtimes.api_gateway.APIGateway):
+            api_gateway = api_gateway.to_scheme()
+
+        endpoint_path = f"projects/{project}/nuclio/api-gateways/{name}"
```
liranbg, line 3329: this is not our restful convention - when POSTing (aka creating) you do it on the resource bucket (not on the resource instance
```suggestion
        endpoint_path = f"projects/{project}/nuclio/api-gateways"
```

mlrun/projects/project.py
```diff
@@ -3502,6 +3504,75 @@ def remove_custom_packager(self, packager: str):
...
+                    RemoteRuntime,
+                    ServingRuntime,
+                ]
+            ],
+            Union[RemoteRuntime, ServingRuntime],
+        ],
+        path: Optional[str] = "/",
+        description: Optional[str] = "",
+        username: Optional[str] = None,
+        password: Optional[str] = None,
```
liranbg, line 3522:
```suggestion
        basic_auth: Optional[Tuple[str,str]] = None
```

mlrun/projects/project.py
```diff
@@ -3502,6 +3504,75 @@ def remove_custom_packager(self, packager: str):
...
+
+        ok = mlrun.db.get_run_db().create_api_gateway(api_gateway=api_gateway)
+
+        if ok:
+            # TODO: rework to request specific api gateway by name
+            api_gateways_list = self.list_api_gateways()
+            for gw in api_gateways_list:
+                if gw.project == self.name and gw.name == name:
+                    return gw
+        return None
```
liranbg, line 3562: why doesnt the create api gateway returns the created?

mlrun/runtimes/api_gateway.py
```diff
@@ -0,0 +1,274 @@
...
+
+import requests
+
+import mlrun
+import mlrun.common.schemas
+
+from .function import RemoteRuntime
+from .serving import ServingRuntime
+
+BASIC_AUTH_NUCLIO_API_GATEWAY_AUTH_MODE = "basicAuth"
```
liranbg, line 25:
```suggestion
NUCLIO_API_GATEWAY_AUTHENTICATION_MODE_BASIC_AUTH = "basicAuth"
```
note the convention

mlrun/runtimes/api_gateway.py
```diff
@@ -0,0 +1,274 @@
...
+    @staticmethod
+    def _enrich_authentication_mode(username, password):
+        return (
+            NO_AUTH_NUCLIO_API_GATEWAY_AUTH_MODE
+            if username is not None and password is not None
+            else BASIC_AUTH_NUCLIO_API_GATEWAY_AUTH_MODE
+        )
+
+    @staticmethod
+    def _generate_auth(username: str, password: str):
```
liranbg, line 246:
```suggestion
    def _generate_basic_auth(username: str, password: str):
```

server/api/api/endpoints/nuclio.py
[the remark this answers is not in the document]
liranbg: While I understand the motivation here, the caveat is that we now need to maintain the schema on both places which is a huge headache. I would prefer raw JSON and make the swagger refer to nuclio's documentation for more information.
if nuclio adds a field - we want to allow mlrun seamlessly pass it and not omit/ignore those "unknown" fields.
perhaps when setting the pydantic we can set to accept extra fields

mlrun/runtimes/api_gateway.py
```diff
@@ -0,0 +1,274 @@
...
+        token = base64.b64encode(f"{username}:{password}".encode()).decode()
+        return f"Basic {token}"
+
+    @staticmethod
+    def _process_canary(
```
liranbg, line 251:
```suggestion
    def _resolve_canary(
```

server/api/api/endpoints/nuclio.py
liranbg, line 54:
```suggestion
@router.post("/projects/{project}/nuclio/api-gateways")
async def create_api_gateway(
```

server/api/utils/clients/async_nuclio.py
```diff
@@ -43,8 +46,33 @@ async def list_api_gateways(self, project_name=None):
...
+        project_name: str,
+        api_gateway_name: str,
+        api_gateway: mlrun.common.schemas.APIGateway,
+    ):
+        headers = {}
+
+        if project_name:
+            headers[NUCLIO_PROJECT_NAME_HEADER] = project_name
+
+        body = self._generate_nuclio_api_gateway_body(
```
liranbg, line 64:
```suggestion
        body = self._compile_nuclio_api_gateway_body(
```
or resolve 🤷🏼

liranbg (changes requested):
the apigate way scehma missing the name here, also annotation, labels, etc
while I know we have said that it would be a mistake to copy the nuclio schema here too, perhaps we can retain the structure to be the same so we can let the user extend / query it on his own while having pydantic allowing us to have "unknown" fields? that way we will avoid the scenario where "strict" rpc-comunnication (Grpc, etc) omitted unknown fields.

```suggestion
    metadata: ...
    spec: ...
    status: ...
```

sorry for the hassle, I just think it worth it to reduce the "guessing" game we have for MLes understanding what are the capabilities this feature has

mlrun/projects/project.py
```diff
@@ -3502,6 +3504,75 @@ def remove_custom_packager(self, packager: str):
...
+        )
+
+        ok = mlrun.db.get_run_db().create_api_gateway(api_gateway=api_gateway)
+
+        if ok:
+            # TODO: rework to request specific api gateway by name
+            api_gateways_list = self.list_api_gateways()
+            for gw in api_gateways_list:
+                if gw.project == self.name and gw.name == name:
+                    return gw
```
rokatyy, line 3561: review this

mlrun/runtimes/api_gateway.py
```diff
@@ -0,0 +1,274 @@
...
+        return mlrun.common.schemas.APIGateway(
+            functions=self.functions,
+            path=self.path,
+            description=self.description,
+            username=self.__username,
+            password=self.__password,
+            canary=self.canary,
+        )
+
+    def generate_invoke_url(
```
rokatyy, line 132: get_invoke_url

mlrun/runtimes/api_gateway.py
```diff
@@ -0,0 +1,274 @@
...
+NO_AUTH_NUCLIO_API_GATEWAY_AUTH_MODE = "none"
+PROJECT_NAME_LABEL = "nuclio.io/project-name"
+
+
+class APIGateway:
+    def __init__(
+        self,
+        project,
+        name: str,
+        path: str,
```
rokatyy, line 35: to set default one `/`

mlrun/runtimes/api_gateway.py
```diff
@@ -0,0 +1,274 @@
...
+            canary=self.canary,
+        )
+
+    def generate_invoke_url(
+        self,
+    ):
+        return f"{self.host}{self.path}"
+
+    def set_invoke_url(self, invoke_url: str):
+        self._invoke_url = invoke_url
```
rokatyy, line 138: delete this

server/api/api/endpoints/nuclio.py
```diff
@@ -47,3 +48,24 @@ async def list_api_gateways(
...
+    auth_info: mlrun.common.schemas.AuthInfo = Depends(deps.authenticate_request),
+):
+    await server.api.utils.auth.verifier.AuthVerifier().query_project_resource_permissions(
+        mlrun.common.schemas.AuthorizationResourceTypes.api_gateway,
+        project,
+        gateway,
+        mlrun.common.schemas.AuthorizationAction.create,
+        auth_info,
+    )
+    await server.api.utils.clients.async_nuclio.Client(auth_info).create_api_gateway(
```
rokatyy, line 67: return

server/api/utils/clients/async_nuclio.py
```diff
@@ -43,8 +46,33 @@ async def list_api_gateways(self, project_name=None):
...
+            project_name=project_name,
+            api_gateway_name=api_gateway_name,
+            api_gateway=api_gateway,
+        )
+
+        return await self._send_request_to_api(
+            method="POST",
+            url=self._nuclio_dashboard_url,
```
rokatyy, line 72: remove and use it from send_request..

mlrun/runtimes/api_gateway.py
TomerShor, line 1: @rokatyy note that nuclio-related runtimes were moved to a subfolder `mlrun/runtimes/nuclio`. Move this file there as well.

rokatyy:
It has been decided to make some adjustments to this pull request. I have temporarily moved it to draft mode for the purpose of fixing.

server/api/utils/clients/async_nuclio.py
```diff
@@ -18,35 +18,100 @@
...
+        self._username = auth_info.username
         self._auth = aiohttp.BasicAuth(auth_info.username, auth_info.session)
         self._logger = logger.get_child("nuclio-client")
         self._nuclio_dashboard_url = mlrun.mlconf.nuclio_dashboard_url
+        self._nuclio_domain = urllib.parse.urlparse(self._nuclio_dashboard_url).netloc
 
-    async def list_api_gateways(self, project_name=None):
+    async def list_api_gateways(
+        self, project_name=None
+    ) -> dict[str, mlrun.common.schemas.APIGateway]:
```
TomerShor, line 46: why did you choose a `dict` response? is it for an easier permission query in the endpoint?

server/api/utils/clients/async_nuclio.py
```diff
@@ -18,35 +18,100 @@
...
 import mlrun.common.schemas
+import mlrun.errors
+import mlrun.runtimes.nuclio.api_gateway
 import mlrun.utils
 from mlrun.utils import logger
 
 NUCLIO_API_SESSIONS_ENDPOINT = "/api/sessions/"
-NUCLIO_API_GATEWAYS_ENDPOINT = "/api/api_gateways/"
+NUCLIO_API_GATEWAYS_ENDPOINT = "/api/api_gateways/{api_gateway}"
 API_GATEWAY_NAMESPACE_HEADER = "X-Nuclio-Api-Gateway-Namespace"
```
TomerShor, line 28: Align with the rest
```suggestion
NUCLIO_API_GATEWAY_NAMESPACE_HEADER = "X-Nuclio-Api-Gateway-Namespace"
```

server/api/utils/clients/async_nuclio.py
```diff
@@ -18,35 +18,100 @@
...
 NUCLIO_PROJECT_NAME_HEADER = "X-Nuclio-Project-Name"
+NUCLIO_PROJECT_NAME_LABEL = "nuclio.io/project-name"
+IGUAZIO_USERNAME_LABEL = "iguazio.com/username"
+MLRUN_CREATED_LABEL = "mlrun-created"
```
TomerShor, line 32: These labels are not really nuclio-only, so I think the can be in a more common file (like `mlrun/common/constants.py`, or a new `mlrun/common/labels.py`.

Also, @theSaarco do we want to change `mlrun-created=True` to `created-by=mlrun`?

server/api/utils/clients/async_nuclio.py
```diff
@@ -18,35 +18,100 @@
...
+        )
+
+        return await self._send_request_to_api(
+            method=method,
+            path=path,
             headers=headers,
+            json=body,
         )
 
+    def set_iguazio_labels(self, nuclio_object, project_name):
```
TomerShor, line 110: this can be a private method

server/api/utils/clients/async_nuclio.py
```diff
@@ -102,3 +178,18 @@ def _handle_error_response(
         self._logger.warning("Request to nuclio failed. Reason:", **log_kwargs)
 
         mlrun.errors.raise_for_status(response, error_message)
+
+    def _resolve_nuclio_api_gateway(
```
TomerShor, line 182: Not really resolving anything. perhaps Enriching
```suggestion
    def _enrich_nuclio_api_gateway(
```

server/api/api/endpoints/nuclio.py
```diff
@@ -21,29 +22,122 @@
...
     project: str,
     auth_info: mlrun.common.schemas.AuthInfo = Depends(deps.authenticate_request),
 ):
+    """
     await server.api.utils.auth.verifier.AuthVerifier().query_project_permissions(
         project_name=project,
         action=mlrun.common.schemas.AuthorizationAction.read,
         auth_info=auth_info,
     )
+    """
```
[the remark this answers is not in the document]
rokatyy: @TomerShor I'll delete it, leftovers after testing it

server/api/utils/clients/async_nuclio.py
rokatyy: @TomerShor because the Nuclio api returns the dictionary

mlrun/common/schemas/api_gateway.py
```diff
@@ -0,0 +1,93 @@
...
+    metadata: APIGatewayMetadata
+    spec: APIGatewaySpec
+    status: Optional[APIGatewayStatus]
+
+    class Config:
+        extra = pydantic.Extra.allow
+
+
+class APIGateways(pydantic.BaseModel):
+    api_gateways: typing.Optional[dict[str, APIGateway]] = {}
```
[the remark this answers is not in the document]
rokatyy: @TomerShor I followed the same way it done in projects schema. How you would suggest to do it instead?

mlrun/projects/project.py
```diff
@@ -3546,6 +3547,77 @@ def remove_custom_packager(self, packager: str):
...
+        return [
+            mlrun.runtimes.nuclio.api_gateway.APIGateway.from_scheme(gateway_dict)
+            for gateway_dict in gateways_list.api_gateways.values()
+        ]
+
+    def get_api_gateway(
+        self,
+        name: Optional[str],
+        api_gateway: Optional[mlrun.runtimes.nuclio.api_gateway.APIGateway],
+    ) -> mlrun.runtimes.nuclio.api_gateway.APIGateway:
```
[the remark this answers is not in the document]
rokatyy: @TomerShor it was my idea with a future extension. It might be useful to request actual `Status` of gateway. Also, user might update the api-gateway in one jupyter instance and upload the changes in another instance.

mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -0,0 +1,300 @@
...
+        if (
+            self.authentication_mode
+            == NUCLIO_API_GATEWAY_AUTHENTICATION_MODE_BASIC_AUTH
+            and not auth
+        ):
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "API Gateway invocation requires authentication. Please pass credentials"
+            )
+        if auth:
+            headers["Authorization"] = self._generate_basic_auth(*auth)
```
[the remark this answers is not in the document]
rokatyy: @TomerShor then we can do this in a following PR. I'd not touch it in this PR as this code is already tested. wdyt?

mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -0,0 +1,300 @@
...
+        self,
+        method="POST",
+        headers: dict = {},
+        auth: Optional[tuple[str, str]] = None,
+        **kwargs,
+    ):
+        if not self.invoke_url:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "Invocation url is not set. Set up gateway's `invoke_url` attribute."
+            )
```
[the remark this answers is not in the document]
rokatyy: @TomerShor sounds like idea for the next PR as well

server/api/api/endpoints/nuclio.py
```diff
@@ -21,29 +22,122 @@
...
+    client = server.api.utils.clients.async_nuclio.Client(auth_info)
+    api_gateways = await client.list_api_gateways(project)
+    await client.close_session()
```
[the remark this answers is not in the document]
rokatyy: @TomerShor make sense!

mlrun/common/schemas/api_gateway.py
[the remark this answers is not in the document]
TomerShor: OK I see. I would just go with `APIGatewaysOutput` then, so it is clear that it's for the list endpoint output.

mlrun/runtimes/nuclio/api_gateway.py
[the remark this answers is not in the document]
TomerShor: Add a TODO then

mlrun/runtimes/nuclio/api_gateway.py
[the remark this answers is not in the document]
rokatyy: @TomerShor I'm adding everything to the jira task

mlrun/db/httpdb.py
```diff
@@ -3340,20 +3341,61 @@ def verify_authorization(
             body=dict_to_json(authorization_verification_input.dict()),
         )
 
-    def list_api_gateways(self, project=None):
+    def list_api_gateways(self, project=None) -> mlrun.common.schemas.APIGateways:
```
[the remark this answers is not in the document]
rokatyy: We return here what server part returns us

mlrun/common/schemas/api_gateway.py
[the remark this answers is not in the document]
rokatyy: renamed

mlrun/projects/project.py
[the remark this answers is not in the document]
rokatyy: Okay, then will leave only the name.

server/api/utils/clients/async_nuclio.py
rokatyy: @TomerShor moved it to `constants.py`. 

Do not resolve conversation until @theSaarco response.

theSaarco: @rokatyy - The MLRun ticket [ML-4243](https://jira.iguazeng.com/browse/ML-4243) defines `mlrun/system-generated`, with potentially a `mlrun/system-generated-reason` (which I don't think is needed in this case).

mlrun/common/schemas/api_gateway.py
```diff
@@ -0,0 +1,92 @@
...
+            return cls.none
+        elif authentication_mode == "basicAuth":
+            return cls.basic
+        else:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"Authentication mode `{authentication_mode}` is not supported",
+            )
+
+
+class APIGatewayMetadata(pydantic.BaseModel):
```
liranbg, line 39:
```suggestion
class _APIGatewayBaseModel(pydantic.BaseModel):
    class Config:
        extra = pydantic.Extra.allow

class APIGatewayMetadata(_APIGatewayBaseModel):
```

and all rest of classes can inherit the config extra = allow

mlrun/db/nopdb.py
```diff
@@ -506,6 +506,19 @@ def get_hub_item(
...
+        api_gateway: mlrun.runtimes.nuclio.APIGateway,
+    ) -> mlrun.common.schemas.APIGateway:
+        pass
+
+    def list_api_gateways(self, project=None):
+        pass
+
+    def get_api_gateway(self, name, project=None):
+        pass
+
```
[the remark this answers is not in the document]
liranbg: @rokatyy then implement on db while adding "raise NotImplementedError"
see how `sql.db` implements `create_project_secrets`

mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -0,0 +1,302 @@
...
+        ):
+            api_gateway.spec.authentication = mlrun.common.schemas.APIGatewayBasicAuth(
+                username=self.__username, password=self.__password
+            )
+        return api_gateway
+
+    def get_invoke_url(
+        self,
+    ):
+        return urljoin(self.host, self.path)
```
liranbg, line 166:
```suggestion
   @propery
    def invoke_url(
        self,
    ):
        return urljoin(self.host, self.path)
```

mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -0,0 +1,302 @@
...
+        ],
+        description: str = "",
+        path: str = "/",
+        authentication_mode: Optional[
+            str
+        ] = NUCLIO_API_GATEWAY_AUTHENTICATION_MODE_NONE,
+        host: Optional[str] = None,
+        canary: Optional[list[int]] = None,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
```
liranbg, line 56: instead of passing the auth mode + its options
read:
None -> ()
Basic -> (username, password)
IGZ -> (accessToken)
SSO -> (JWT)
and so on, just accept an argument that abstracts the authentication (an interface) mode. each concrete (none, basic, igz, sso) will implement some functionality that extends a given api gateway structure

server/api/api/endpoints/nuclio.py
```diff
@@ -31,19 +37,103 @@ async def list_api_gateways(
         action=mlrun.common.schemas.AuthorizationAction.read,
         auth_info=auth_info,
     )
+    client = server.api.utils.clients.async_nuclio.Client(auth_info)
+    try:
+        api_gateways = await client.list_api_gateways(project)
+    finally:
+        await client.close_session()
```
liranbg, line 44: why do you need to explicitly close the session? that's some boilerplate here.
u may close the session on `__del__` functionality or, if you want to do it faster than python's GC then provide a `@contextlib.asynccontextmanager` decorator on top of the class that will close the session for you
e.g:

```python
  api_gateways = await client.list_api_gateways(project)
```

while the class implements `__aenter__`
```python
class Client:
  def __init__(auth_info):
    self._auth_info = auth_info
  async def __aenter__():
    await self._session.initialize()
    yield self
  async def __aexit__(*args, **kwargs):
    await self._session.close()
```

(that's just an example of course

server/api/api/endpoints/nuclio.py
```diff
@@ -31,19 +37,103 @@ async def list_api_gateways(
...
+                api_gateway_name=gateway,
+                api_gateway=api_gateway,
+                create=True,
+            )
+        else:
+            await client.store_api_gateway(
+                project_name=project,
+                api_gateway_name=gateway,
+                api_gateway=api_gateway,
+            )
```
liranbg, line 131:
```suggestion
        create = not await client.api_gateway_exists(
            name=gateway,
            project_name=project,
        )
        await client.store_api_gateway(
                project_name=project,
                api_gateway_name=gateway,
                api_gateway=api_gateway,
                create=create,
            )
```

i think shorter and concise

server/api/utils/clients/async_nuclio.py
```diff
@@ -18,35 +18,100 @@
...
+NUCLIO_API_GATEWAY_NAMESPACE_HEADER = "X-Nuclio-Api-Gateway-Namespace"
 NUCLIO_PROJECT_NAME_HEADER = "X-Nuclio-Project-Name"
+NUCLIO_PROJECT_NAME_LABEL = "nuclio.io/project-name"
+IGUAZIO_USERNAME_LABEL = "iguazio.com/username"
 
 
 class Client:
     def __init__(self, auth_info: mlrun.common.schemas.AuthInfo):
         self._session = None
+        self._username = auth_info.username
```
liranbg, line 38: why do you need it explicitly here?

server/api/utils/clients/async_nuclio.py
```diff
@@ -55,13 +120,18 @@ async def _ensure_async_session(self):
                 logger=logger,
             )
 
+    async def close_session(self):
+        if self._session:
+            await self._session.close()
+            self._session = None
```
liranbg, line 126: use some sort of `@contextlib.asynccontextmanager` functionality. better and simplify the boilerplating

liranbg (changes requested):
very well! getting much closer to nail this PR
some recommendation and suggestion to make the code more concise and DRY-er :)

mlrun/common/schemas/api_gateway.py
rokatyy: Nice idea!

mlrun/runtimes/nuclio/api_gateway.py
rokatyy: @liranbg this will be done in the following PR if you don't mind,  just not to add more classes/functionality to this one. It's already 1k changes. I've added link to this comment in jira task - https://jira.iguazeng.com/browse/ML-5372

server/api/utils/clients/async_nuclio.py
rokatyy: @liranbg to set `iguazio.com/username` labels below

server/api/api/endpoints/nuclio.py
rokatyy: @liranbg agree, reworked! Thanks for the detailed example :)

TomerShor (changes requested):

requirements.txt
```diff
@@ -41,3 +41,4 @@ deprecated~=1.2
 jinja2~=3.1, >=3.1.3
 anyio~=3.7
 orjson~=3.9
+python-multipart~=0.0.6
```
liranbg, line 44: Not sure what uses it?

liranbg (commented):
Much better now 👌

requirements.txt
rokatyy: @liranbg It was needed for old version of PR. I've removed it.

server/api/utils/clients/async_nuclio.py
```diff
@@ -18,13 +18,17 @@
...
+import mlrun.errors
+import mlrun.runtimes.nuclio.api_gateway
 import mlrun.utils
+from mlrun.common.constants import MLRUN_CREATED_LABEL
 from mlrun.utils import logger
 
 NUCLIO_API_SESSIONS_ENDPOINT = "/api/sessions/"
-NUCLIO_API_GATEWAYS_ENDPOINT = "/api/api_gateways/"
-API_GATEWAY_NAMESPACE_HEADER = "X-Nuclio-Api-Gateway-Namespace"
+NUCLIO_API_GATEWAYS_ENDPOINT = "/api/api_gateways/{api_gateway}"
```
TomerShor:
```suggestion
NUCLIO_API_GATEWAYS_ENDPOINT_TEMPLATE = "/api/api_gateways/{api_gateway}"
```

TomerShor (approved):
Nicely done 🆗 (with a tiny comment).

Regarding BC tests - I think that's ok because you renamed an endpoint (removed `/nuclio/`) - FYI @liranbg @alonmr

PR #4744, opened 2023-12-17
[Artifacts] Remove `legacy` format

TomerShor:
In the artifact refactor we removed all legacy artifacts, so we do not support this artifact format anymore.
If a request is made with a `format=legacy` query, it will get a 422 (Unprocessable Entity) response from FastAPI.


theSaarco (approved):
👍

PR #4750, opened 2023-12-17
[Artifacts] Get most recent artifacts in project summaries

TomerShor:
Project summaries was using querying artifacts with the `latest` tag, causing the number of artifacts to be no as expected.
Instead, project summaries shows the "most recent" artifacts with distinct keys - meaning it should count only one artifact per key, according to `max(updated)`.

Relates to https://jira.iguazeng.com/browse/ML-5312


liranbg (approved):

theSaarco (approved):
LGTM

PR #4763, opened 2023-12-19
[API] Fix passing project spec params as int object

Yacouby:


server/api/db/sqldb/db.py
```diff
@@ -1963,7 +1963,8 @@ def _normalize_project_parameters(project: mlrun.common.schemas.Project):
         # remove leading & trailing whitespaces from the project parameters keys and values to prevent duplications
         if project.spec.params:
             project.spec.params = {
-                key.strip(): value.strip() for key, value in project.spec.params.items()
+                str(key).strip(): str(value).strip()
```
liranbg, line 1966:
```suggestion
                str(key).strip(): value.strip() if isinstance(value, str) else value
```

PR #4766, opened 2023-12-19
[Runtimes] Databricks output as logs

tomerm-iguazio:
Add databricks output to logs....
Including non-logs info - such as metadata.

[ML-5310](https://jira.iguazeng.com/browse/ML-5310)

### DIFF
--- a/mlrun/runtimes/databricks_job/databricks_wrapper.py
+++ b/mlrun/runtimes/databricks_job/databricks_wrapper.py
@@ -208,4 +208,4 @@ def run_mlrun_databricks_job(
     logger.info(f"logs:\n{run_output.logs}")
     run_output_dict = run_output.as_dict()
     run_output_dict.pop('logs', None)
-    logger.info(f"Run output and metadata:\n{run_output_dict}")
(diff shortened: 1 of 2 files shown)


mlrun/runtimes/databricks_job/databricks_wrapper.py
```diff
@@ -206,3 +206,6 @@ def print_status(run: Run):
 
     logger.info(f"job finished: {run.run_page_url}")
     logger.info(f"logs:\n{run_output.logs}")
+    run_output_dict = run_output.as_dict()
```
assaf758, line 209: in case of exception in the outer try-catch, I think run_output is not defined

tomerm-iguazio: In the outer try-catch there is only try and finally. If there will be an error, it will not catch it, and will raise it (python will also run the finally part in this scenario).

so this code will not run in the case of exception....

tomerm-iguazio: add comment.

assaf758 (approved):

PR #4768, opened 2023-12-19
[MAKE] Make awk expression compatible with older awk versions

yanburman:


server/log-collector/Makefile
```diff
@@ -17,7 +17,7 @@ MLRUN_DOCKER_REPO ?= mlrun
 MLRUN_DOCKER_TAG ?= $(shell echo "$(MLRUN_VERSION)" | sed -E 's/\+/\-/g')
 MLRUN_DOCKER_IMAGE_PREFIX := $(if $(MLRUN_DOCKER_REGISTRY),$(strip $(MLRUN_DOCKER_REGISTRY))$(MLRUN_DOCKER_REPO),$(MLRUN_DOCKER_REPO))
 
-GO_VERSION=$(shell awk '/go ([0-9]\.[0-9]{2})/ {print $$2}' go.mod)
+GO_VERSION=$(shell awk '/^go ([0-9]\.[0-9])/ {print $2}' go.mod)
```
liranbg, line 20: another way is 
`grep -m1 'go ' go.mod | cut -d' ' -f2`
if awk is problematic in that regard

PR #4770, opened 2023-12-19
[Functions] If a function is available in local memory, it can be preserved during sync

laurybueno:
[ML-3680](https://jira.iguazeng.com/browse/ML-3680)

This minor tweak to `sync_functions` allows it to reuse function objects if they are already available in memory.

This is not intended as a performance optimization but as a fix to the following use case:

1. user creates a function
2. saves it
3. tweak its resources
4. saves it again
5. runs workflow that executes said function while using pipeline engine `kfp`

The above scenario spawns a function execution without any information on its resources field. This occurs because `project.run` triggers a [new project sync](https://github.com/mlrun/mlrun/blob/0d4ef3a1d21e520c1c6f95baf0e4b4d0f9514b0d/mlrun/projects/project.py#L2589), ultimately leading to the [recreation of the function object](https://github.com/mlrun/mlrun/blob/0d4ef3a1d21e520c1c6f95baf0e4b4d0f9514b0d/mlrun/projects/project.py#L2390) using the barebones dictionary available on `project.spec._function_definitions` as a reference.

### Note to maintainers
This PR should only be merged once development on 1.7.0 starts.


tests/projects/test_project.py
```diff
@@ -86,6 +86,22 @@ def test_sync_functions_with_names_different_than_default(rundb_mock):
...
 
+def test_sync_functions_preserves_existing(rundb_mock):
+    project = mlrun.new_project("project-name", save=False)
+    project.set_function("hub://describe", "describe")
+    project.set_function("hub://auto-trainer", "auto-trainer")
+
+    old_trainer = project.spec._function_objects.pop("auto-trainer")
+    old_describe = project.spec._function_objects["describe"]
+
+    project.sync_functions(always=False)
```
theSaarco, line 97: For completeness, may be interesting to check the situation where you pass `names` of functions, to verify the same logic holds for specific functions as well. Just note that if you want to test the 1st-pass behavior, make sure to reset the project's `_initialized` property before (as it gets set on `sync_functions`).

theSaarco (approved):
@laurybueno - looks good. Had a minor suggestion on tests.
Just to verify - did you go through places in the code where `sync_functions` is called and verified that the (somewhat different) meaning of the `always` parameter and the sync logic make sense in their context?

tests/projects/test_project.py
laurybueno: Good points, Saar!

I improved our test coverage. Please, check if this is what you expected.

laurybueno:
Thanks for your review, @theSaarco!

As far as I can tell, `sync_functions` is only used with the `always` parameter in one place on our codebase: https://github.com/mlrun/mlrun/blob/c67c84a0013daca4a694fa342c83f54362328b33/mlrun/projects/project.py#L2622

I believe our change on this PR is safe in this context, but I would love to hear your assessment as well.

PR #4772, opened 2023-12-20
[API] Add model monitoring dashboard URL to frontend spec 

gtopper:
[ML-4894](https://jira.iguazeng.com/browse/ML-4894)

### DIFF
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -650,7 +650,7 @@ default_config = {
         # used for igz client when emitting events
         "access_key": "",
     },
-    "external_grafana_url": "",
+    "grafana_url": "",
 }
 
 _is_running_as_api = None
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -650,6 +650,7 @@ default_config = {
         # used for igz client when emitting events
         "access_key": "",
     },
+    "external_grafana_url": "",
 }
 
 _is_running_as_api = None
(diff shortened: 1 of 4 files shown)


quaark (approved):

tests/api/api/test_frontend_spec.py
```diff
@@ -158,7 +158,7 @@ def test_get_frontend_spec_jobs_dashboard_url_resolution(
         == f"{grafana_url}/d/mlrun-jobs-monitoring/mlrun-jobs-monitoring?orgId=1"
         f"&var-groupBy={{filter_name}}&var-filter={{filter_value}}"
     )
```
assaf758, line 160: you can also assert on frontend_spec.model_monitoring_dashboard_url

tests/api/api/test_frontend_spec.py
```diff
@@ -158,7 +158,7 @@ def test_get_frontend_spec_jobs_dashboard_url_resolution(
...
-    server.api.utils.clients.iguazio.Client().try_get_grafana_service_url.assert_called_once()
+    server.api.utils.clients.iguazio.Client().try_get_grafana_service_url.assert_called()
```
assaf758, line 161: you can assert it's called twice, no?

mlrun/config.py
```diff
@@ -650,6 +650,7 @@
         # used for igz client when emitting events
         "access_key": "",
     },
+    "external_grafana_url": "",
```
assaf758, line 653: i'm not sure "external" is very informative - in both cases (iguazio / ce) grafana is some pod in the k8s.
why not just `grafana_url` (that happens to be configured in ce case, but possibly in other cases)

assaf758 (approved):

PR #4774, opened 2023-12-20
[Pipelines] Fix Non-deterministic Remote Pipeline Notifications

quaark:
The Remote Pipeline Runner wasn't really waiting for the pipeline, causing the notifications to be sent too early and therefore not always contain all the run's data.


mlrun/projects/pipelines.py
```diff
@@ -1081,6 +1081,7 @@ def load_and_run(
         raise RuntimeError(f"Workflow {workflow_log_message} failed") from run.exc
 
     if wait_for_completion:
+        run.wait_for_completion()
```
liranbg, line 1084: worth wrapping in try/except to ensure that following lines would be executed in case of an error

PR #4781, opened 2023-12-20
[Pipelines] Default workflow status to unknown if status is missing

liranbg:
https://jira.iguazeng.com/browse/ML-5195


alonmr (approved):
Generally looks good but I suggest moving `format_run` to kfpops.py because its not clear that it is meant for pipeline runs

PR #4784, opened 2023-12-21
[Project] Delete Project Background Task

quaark:
## New Project Delete Behaviour
https://jira.iguazeng.com/browse/IG-22194

### When MLRun is leader (Open Source):
1. Create a background task for deleting the project
2. The background task will delete the project resources and then project itself

### When MLRun is a follower:
Due to the nature of the project deletion flow, we need to wrap the task as:
```
MLRunDeletionWrapperTask(LeaderDeletionJob(MLRunDeletionTask))
```
The flow is as follows:
1. Create MLRunDeletionWrapperTask
2. MLRunDeletionWrapperTask will send a request to the projects leader to delete the project
3. MLRunDeletionWrapperTask will wait for the project to be deleted using LeaderDeletionJob job id
    **During (In leader)**:
    1. Create LeaderDeletionJob
    2. LeaderDeletionJob will send a second delete project request to the follower
    3. LeaderDeletionJob will wait for the project to be deleted using the MLRunDeletionTask task id
        **During (Back here in follower)**:
        1. Create MLRunDeletionTask
        2. MLRunDeletionTask will delete the project resources and then project itself.
        3. Finish MLRunDeletionTask
    4. Finish LeaderDeletionJob

## Changes
- Added v2 DELETE /projects endpoint which behaves in the way stated above (for `check` strategy, no BG task is spawned, the check is performed and returned immediately)
- HTTPDB calls new endpoint and waits for BG task.
- Kept v1 DELETE /projects endpoint for backwards compatibility
- Added v2 /healthz endpoint (useful for the project leader to see whether v2 api exists in this mlrun server or not)
- Internal BG Tasks can now have a given name and timeout
- reroute request to chief can now have a configurable api version (set to v2 in our v2 DELETE /projects but left as default for the v1 DELETE /projects endpoint)


server/api/api/endpoints/healthz_v2.py
TomerShor, line 1: Is this actually needed?
Can't we reuse the v1 `healthz` as the v2_router's `healthz` as well?

server/api/api/endpoints/projects_v2.py
```diff
@@ -0,0 +1,193 @@
...
+            auth_info=auth_info,
+        )
+
+
+def _is_request_from_leader(
+    projects_role: typing.Optional[mlrun.common.schemas.ProjectsRole],
+) -> bool:
+    if projects_role and projects_role.value == mlrun.mlconf.httpdb.projects.leader:
+        return True
+    return False
```
TomerShor, line 175: duplicate function from `projects.py`, can be moved to a helper function.

TomerShor (changes requested):
Nice one mate,
Only a couple of minor suggestions.

Also - can it would be nice to see a new test that validates the background tasks or a system test of sorts.

TomerShor (approved):
LGTM 👍

alonmr (commented):
Looks great! Minor comments

quaark:
> Nice one mate, Only a couple of minor suggestions.
> 
> Also - can it would be nice to see a new test that validates the background tasks or a system test of sorts.

It's a bit of an issue/already covered. The MLRun leader flow is covered in regular integration tests that already test deleting the project. For the full flow only a system test will do. However, all the system tests are already creating and deleting projects. So I wonder if it's worth making them longer for something they already do. The test itself will literally be:
```
def test_delete_project():
    pass
```

alonmr (approved):
Looks good! Couple of minor suggestions

PR #4786, opened 2023-12-21
[Tests] Change test target path

tomerm-iguazio:
Change large df ingest test target path.

Following https://github.com/mlrun/mlrun/pull/4777


tests/system/feature_store/test_feature_store.py
```diff
@@ -785,13 +785,13 @@ def test_ingest_large_parquet(self):
             "Column1": range(0, num_rows),
             "Column2": np.random.choice(["A", "B", "C"], size=num_rows),
         }
-
+        path = f"{str(self.results_path / _generate_random_name())}.parquet"
```
gtopper, line 788:
```suggestion
        path = f"{self.results_path / _generate_random_name()}.parquet"
```

gtopper (approved):

PR #4790, opened 2023-12-21
[Makefile] Only run fmt and lint on tracked files

gtopper:


liranbg:
@gtopper - why? perhaps i would like to enjoy the feature for non-tracked files as well (e.g. file to be added to code base)

Makefile
```diff
@@ -66,6 +66,8 @@ MLRUN_PYTHON_VERSION_SUFFIX := $(if $(INCLUDE_PYTHON_VERSION_SUFFIX),$(MLRUN_ANA
 MLRUN_OLD_VERSION_ESCAPED = $(shell echo "$(MLRUN_OLD_VERSION)" | sed 's/\./\\\./g')
 MLRUN_BC_TESTS_OPENAPI_OUTPUT_PATH ?= $(shell pwd)
 
+CHECKED_IN_PYTHON_FILES := $(shell git ls-files | grep '\.py$$')
```
jond01, line 69: This doesn't catch notebooks

jond01 (commented):
What scenario does it solve? Untracked files are not checked or formatted by Ruff anyway.
Also, we plan to add pre-commit which handles that more accurately.

gtopper:
@liranbg if it's to be added to the code base, then it's a tracked file.

liranbg:
> @liranbg if it's to be added to the code base, then it's a tracked file.

well, on regular basis I add the file only after I have all code in place, `make fmt` wont format until I add to git 
overall I think this is error prone (as @jond01 suggested above, doesnt catch notebooks and etc)

Can you elaborate what trigger you opening this pr?

gtopper:
@liranbg
> on regular basis I add the file only after I have all code in place, make fmt wont format until I add to git

It doesn't seem like having to git add files that you're going to git add anyway is much of a downside. It's good practice actually.

> I think this is error prone (as @jond01 suggested above, doesnt catch notebooks and etc)

Actually, the formatter and linter are currently _not_ picking up notebooks. If you try to explicitly run the linter on the notebooks, you get a lot of failures.

> Can you elaborate what trigger you opening this pr?

I keep one-offs in there, and them getting picked up by the formatter and linter is naturally not great.

jond01:
> > I think this is error prone (as @jond01 suggested above, doesnt catch notebooks and etc)
> 
> 
> 
> Actually, the formatter and linter are currently _not_ picking up notebooks. If you try to explicitly run the linter on the notebooks, you get a lot of failures.

The formatter does run on the notebooks by configuration.

> > Can you elaborate what trigger you opening this pr?
> 
> 
> 
> I keep one-offs in there, and them getting picked up by the formatter and linter is naturally not great.

Can you be more explicit? Your situation isn't clear.

Makefile
gtopper: The current code doesn't run on the notebooks either.

gtopper:
@liranbg, @jond01, let's set up a call next week to discuss?

gtopper:
Superseded by #4951.

PR #4794, opened 2023-12-21
[Model Monitoring] Upgrade evidently and specify the supported version

jond01:
Resolves [ML-5374](https://jira.iguazeng.com/browse/ML-5374), [ML-5345](https://jira.iguazeng.com/browse/ML-5345).

- Since evidently broke the interface between patches, raise a warning when the current version is not "compatible".
- Upgrade to version [0.4.11](https://github.com/evidentlyai/evidently/releases/tag/v0.4.11).

### DIFF
--- a/mlrun/model_monitoring/evidently_application.py
+++ b/mlrun/model_monitoring/evidently_application.py
@@ -19,20 +19,37 @@ from typing import Union
 import pandas as pd
 import semver
 
+from mlrun.errors import MLRunIncompatibleVersionError
 from mlrun.model_monitoring.application import ModelMonitoringApplication
 
 SUPPORTED_EVIDENTLY_VERSION = semver.Version.parse("0.4.11")
+
+
+def _check_evidently_version(*, cur: semver.Version, ref: semver.Version) -> None:
+    if ref.is_compatible(cur) or (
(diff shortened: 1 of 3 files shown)


assaf758 (commented):
with evidently publishing a new release every ~two weeks (usually less) I think it's too strict to warn when 0.4.12 is not used (as it means this warning will show almost always).
If know about lower limit - by all means, let's specify it, and even raise exception. For upper limit I suggest warning when the 'minor' is changed

jond01:
> with evidently publishing a new release every ~two weeks (usually less) I think it's too strict to warn when 0.4.12 is not used (as it means this warning will show almost always). If know about lower limit - by all means, let's specify it, and even raise exception. For upper limit I suggest warning when the 'minor' is changed

I agree that this warning is a bit annoying when everything works, but 0.y.z versions are generally not compatible with each other:
- https://semver.org/#spec-item-4
- https://python-semver.readthedocs.io/en/stable/usage/check-compatible-semver-version.html

I changed the logic as you suggested, just note that an incompatibility issue may reoccur without warning (0.y.z -> 0.y.<z+1>).

assaf758 (approved):

PR #4795, opened 2023-12-21
[Workflows] When the user tries to run a non-existing workflow, raise a clear error message

laurybueno:
[ML-5346](https://jira.iguazeng.com/browse/ML-5346)

Raise an error with a clear message when the user mistakenly tries to run a workflow not defined within the current project.


liranbg (approved):

mlrun/projects/project.py
```diff
@@ -2594,6 +2594,8 @@ def run(
 
         if not name and not workflow_path and not workflow_handler:
             raise ValueError("Workflow name, path, or handler must be specified")
+        if name not in self.spec._workflows.keys():
```
alonmr, line 2597: should be after line 2600 since it is valid to specify a workflow name that doesn't exist along side a workflow path or callable handler.
~~~
if workflow_path or (workflow_handler and callable(workflow_handler)):
~~~

laurybueno: Good thinking, Alon! I changed the position of my check. 

When you can, please see if it makes more sense now.

PR #4796, opened 2023-12-21
[Linting] Add pyupgrade

jond01:
Upgrade MLRun codebase to Python 3.9.

- Add the rule to Ruff's config
- Configure the Python target version to 3.9
- Apply the auto-fixes and manual fixes

https://github.com/asottile/pyupgrade
https://docs.astral.sh/ruff/rules/#pyupgrade-up

I reviewed all of the changes (321 / 321 files viewed).


liranbg:
@jond01 while I approve this PR (thought it has conflicts and etc) - I think it would be safer to merge once 1.7 development started to avoid any side effects we are yet to understand

liranbg (approved):

mlrun/datastore/datastore_profile.py
```diff
@@ -30,7 +30,7 @@
 class DatastoreProfile(pydantic.BaseModel):
     type: str
     name: str
-    _private_attributes: typing.List = ()
+    _private_attributes: list = ()
```
jond01, line 33: Unrelated to this change but it's a wrong type hint (empty tuple assignment), pydantic handles it in an inconsistent way

mlrun/model.py
```diff
@@ -1284,7 +1284,7 @@ def ui_url(self) -> str:
         """UI URL (for relevant runtimes)"""
         self.refresh()
         if not self._status.ui_url:
-            print("UI currently not available (status={})".format(self._status.state))
+            print(f"UI currently not available (status={self._status.state})")
```
jond01, line 1287: A print in the code... should be a log.
Can be caught by the `T201` rule:
https://docs.astral.sh/ruff/rules/#flake8-print-t20

assaf758 (approved):

quaark:
> I'm not a fan of "star-gating" arguments, but please note - the pyupgrade _package_ is not added here. We use Ruff linter and opt in to pyupgrade _rules_, implemented by Ruff: https://github.com/astral-sh/ruff/tree/main/crates/ruff_linter/src/rules/pyupgrade Ruff has some more stars, but please not get into that. I know great projects with much less stars.

I've been looking at the original pyupgrade fork here - https://github.com/asottile/pyupgrade
While I agree that we shouldn't star gate projects. I still hesitate with inserting still relatively new products that change code flow.

> The last commit is from less than a month ago: [asottile/pyupgrade@85c6837](https://github.com/asottile/pyupgrade/commit/85c6837431a2eebdc9809f80b7adc529525ea73c) image
> 
> (If you meant to Ruff's pyupgrade - it's from just 4 days ago: [astral-sh/ruff@e50603c](https://github.com/astral-sh/ruff/commit/e50603caf62dd9c45c7eb95038ed80103cc3ab95).)

You're correct I got confused with the last release. Seems they are still working on it yeah.

> Well, I didn't see there any bugs, mainly feature requests. And that's a common case for stable projects, in this case only 14 open and 400 closed... Do you have a specific issue in mind? And I think it means this project isn't really new 😉
> All of pyupgrade rules are quite simple - nearly just formatting:
> 
> * https://docs.astral.sh/ruff/rules/#pyupgrade-up
> * https://github.com/asottile/pyupgrade?tab=readme-ov-file#implemented-features
> 
> I guess you mean to the changes in this commit: [b8bf281](https://github.com/mlrun/mlrun/commit/b8bf28191c64559252f03e6ffd50581fda541346). They are not done by ruff/pyupgrade, but by myself as a part of this PR.

If that was done by you then I am a bit more ok with the changes.
This issue really rubbed me the wrong way - https://github.com/asottile/pyupgrade/issues/918, This is a mishandling of changing `if` logic, it's not simple formatting.

> This is just a tool, and it scanned and fixed here issues in ~30% of the Python files in MLRun codebase without issues (at least to my review). I think it's enough. If we find future issues with it, we can always ignore specific rules or disable it. If you are ok with the changes introduced in this PR, I don't see a reason to worry too much about potential issues.

What about future changes when it comes to upgrade to 3.10 and 3.11? Is pyupgrade mature enough for that? I'm not sure. But I see we can (and have) in the config limited it to conform to py39 anyway.

You have mostly convinced me and I am going to approve the PR on my end. But I feel this is something we should keep in mind and remember to not trust this blindly in our next python version upgrade.

quaark (approved):

PR #4798, opened 2023-12-22
[BackwardsCompatibility] Fix upgrading to mlrun 1.6.x from versions < 1.4

liranbg:
Then might happen because older mlrun versions might stored a pickle where its classes where belong to mlrun.api and not mlrun.common.schemas
this PR intend to bring back the infamous yet needed mlrun.api.schemas so that older clients < 1.4 mlrun would be able to seamlessly upgrade to mlrun 1.6

[ML-5368](https://jira.iguazeng.com/browse/ML-5368)


theSaarco (approved):
Sadly, I approve. 
Is this change enough to support future upgrades (once we remove these extra files in 1.7)? Or do we also need to do data-migration reading all projects and writing them back?

alonmr (approved):

liranbg:
> Sadly, I approve. Is this change enough to support future upgrades (once we remove these extra files in 1.7)? Or do we also need to do data-migration reading all projects and writing them back?

we will need it until all projects are migrated to be JSON and not BLOBS

PR #4801, opened 2023-12-24
[Project] Add `tree` argument to `get_artifact`

TomerShor:
Add an optional argument to get an artifact with a specific `tree` (`producer_id`).

Additionally, improve the naming and docstring of `mlrun.utils.parse_artifact_uri` for better usability.

Related to https://jira.iguazeng.com/browse/ML-4974

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -3148,7 +3148,7 @@ class MlrunProject(ModelObj):
             mock=mock,
         )
 
-    def get_artifact(self, key, tag=None, iter=None):
+    def get_artifact(self, key, tag=None, iter=None, tree=None):
         """Return an artifact object
 
         :param key: artifact key
@@ -3157,7 +3157,9 @@ class MlrunProject(ModelObj):
         :return: Artifact object
         """
         db = mlrun.db.get_run_db(secrets=self._secrets)
-        artifact = db.read_artifact(key, tag, iter=iter, project=self.metadata.name)
+        artifact = db.read_artifact(
+            key, tag, iter=iter, project=self.metadata.name, tree=tree
+        )
         return dict_to_artifact(artifact)
 
     def list_artifacts(
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -3154,6 +3154,7 @@ class MlrunProject(ModelObj):
         :param key: artifact key
         :param tag: version tag
         :param iter: iteration number (for hyper-param tasks)
+        :param tree: the producer id (tree)
         :return: Artifact object
         """
         db = mlrun.db.get_run_db(secrets=self._secrets)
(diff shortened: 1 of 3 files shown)


mlrun/projects/project.py
```diff
@@ -3148,7 +3148,7 @@ def deploy_function(
             mock=mock,
         )
 
-    def get_artifact(self, key, tag=None, iter=None):
+    def get_artifact(self, key, tag=None, iter=None, tree=None):
```
alonmr, line 3151: Add the tree to the docstring

mlrun/utils/helpers.py
```diff
@@ -660,11 +660,18 @@ def dict_to_json(struct):
...
 
 def parse_artifact_uri(uri, default_project=""):
-    uri_pattern = r"^((?P<project>.*)/)?(?P<key>.*?)(\#(?P<iteration>.*?))?(:(?P<tag>.*?))?(@(?P<uid>.*))?$"
+    """parse artifact uri into project, key, tag, iter, tree
+    uri format: [<project>/]<key>[#<iter>][:<tag>][@<tree>]
+
+    :param uri:            uri to parse
+    :param default_project: default project name if not in uri
+    :returns: project, key, iter, tag, tree
+    """
```
alonmr, line 669:
```suggestion
    """
    Parse artifact URI into project, key, tag, iter, tree
    URI format: [<project>/]<key>[#<iter>][:<tag>][@<tree>]

    :param uri:            uri to parse
    :param default_project: default project name if not in URI
    :returns: A tuple of:
              [0] = ...
              [1] = ...
    """
```

mlrun/utils/helpers.py
```diff
@@ -660,11 +660,18 @@ def dict_to_json(struct):
...
+    uri_pattern = r"^((?P<project>.*)/)?(?P<key>.*?)(\#(?P<iteration>.*?))?(:(?P<tag>.*?))?(@(?P<tree>.*))?$"
```
alonmr, line 670: Not sure I follow why uid is no longer part of the artifact uri

alonmr (commented):
A few comments. Also, I don't see `tree` in the get artifact endpoint params. Looks like it will be ignored.

mlrun/utils/helpers.py
TomerShor: The `tree` was called `uid` before the refactor, and this is what actually was in the URI.
Now, UID is auto-generated in the BE, and this value was never a part of the URI.

TomerShor:
> Also, I don't see `tree` in the get artifact endpoint params. Looks like it will be ignored.

@alonmr see https://github.com/mlrun/mlrun/blob/development/server/api/api/endpoints/artifacts_v2.py#L193
Since `project.py` is a client side function, it will only use the v2 endpoints.

PR #4810, opened 2023-12-24
[API] Remove deprecated artifacts endpoints

alonmr:
Self explanatory 
https://jira.iguazeng.com/browse/ML-4899

### DIFF
--- a/server/api/api/endpoints/artifacts.py
+++ b/server/api/api/endpoints/artifacts.py
@@ -242,24 +242,6 @@ async def list_artifacts(
     }
 
 
-async def delete_artifacts_legacy(
-    project: str = mlrun.mlconf.default_project,
-    name: str = "",
-    tag: str = "",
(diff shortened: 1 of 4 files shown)


server/api/api/endpoints/artifacts.py
```diff
@@ -270,13 +242,6 @@ async def list_artifacts(
...
 
 
-# TODO: remove /artifacts in 1.6.0
-@router.delete(
-    "/artifacts",
-    deprecated=True,
-    description="/artifacts is deprecated in 1.4.0 and will be removed in 1.6.0, "
-    "use /projects/{project}/artifacts instead",
-)
 async def delete_artifacts_legacy(
```
TomerShor, line 245: can also be removed if no endpoint leads to it anymore

TomerShor (changes requested):

PR #4816, opened 2023-12-25
[Doc] Add `KubeResource` to docs and fix some malformed doctrings

theSaarco:
The `KubeResource` class has documentation for several functions which are not documented elsewhere (for example `with_security_context`). Added this class to the documentation so that it's visible.
Also fixed some small docstring issues (mostly fixing `:return:` to `:returns:`).

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -499,6 +499,7 @@ class HTTPRunDB(RunDBInterface):
             - state - The state of the runtime object which generates this log, if it exists. In case no known state
               exists, this will be ``unknown``.
             - content - The actual log content.
+
             * in case size = -1, return the state and the final offset
         """
         if size is None:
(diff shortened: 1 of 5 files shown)


mlrun/db/httpdb.py
```diff
@@ -2995,14 +2996,15 @@ def create_model_monitoring_controller(
...
+
         :param project:                  Project name.
         :param default_controller_image: The default image of the model monitoring controller job. Note that the writer
                                          function, which is a real time nuclio functino, will be deployed with the same
                                          image. By default, the image is mlrun/mlrun.
         :param base_period:              Minutes to determine the frequency in which the model monitoring controller job
                                          is running. By default, the base period is 5 minutes.
-        :return: model monitoring controller job as a dictionary. You can easily convert the resulted function into a
-                 runtime object by calling ~mlrun.new_function.
+        :returns: model monitoring controller job as a dictionary. You can easily convert the resulted function into a
```
jillnogold, line 3006: :returns: model monitoring controller job as a dictionary. You can easily convert the returned function into a

mlrun/runtimes/pod.py
```diff
@@ -901,6 +901,10 @@ def get_modifier(self):
 
 
 class KubeResource(BaseRuntime):
+    """
+    A parent class for runtimes which generate k8s resources when executing.
```
jillnogold, line 905: A parent class for runtimes that generate k8s resources when executing.

PR #4820, opened 2023-12-26
[KubeResource] Rename state threshold `running` to `executing`

alonmr:
https://jira.iguazeng.com/browse/ML-5373


liranbg (approved):
makes much more sense now!

PR #4822, opened 2023-12-26
[API] Force delete pods after grace period passed

alonmr:
This was raised by databricks runtime as it uses a grace period of 60 seconds to lay off some time for the cancel task execution.
When testing state thresholds it was apparent that for states like pending scheduled and image pull backoff, the pod would hang on terminating forever, for these states, the cancel task is not needed therefore we can force deletion.

https://jira.iguazeng.com/browse/ML-5271

### DIFF
--- a/server/api/runtime_handlers/base.py
+++ b/server/api/runtime_handlers/base.py
@@ -804,6 +804,8 @@ class BaseRuntimeHandler(ABC):
         namespace: str,
         deleted_pods: List[Dict],
         label_selector: str = None,
+        timeout: int = 180,
+        force: bool = False,
     ):
         deleted_pod_names = [pod_dict["metadata"]["name"] for pod_dict in deleted_pods]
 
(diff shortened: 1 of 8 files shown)


server/api/runtime_handlers/base.py
```diff
@@ -968,8 +983,18 @@ def _delete_pod_resources(
...
+        timeout = 180
+        force = False
+        # Deleting pods in specific states with non 0 grace period can cause the pods to be stuck in terminating state,
+        # so we're forcing deletion after the grace period passed in this case.
+        if self.pod_grace_period_seconds > 0:
+            timeout = self.pod_grace_period_seconds + 10
+            force = True
+        self._wait_for_pods_deletion(
+            namespace, deleted_pods, label_selector, timeout, force
+        )
```
liranbg, line 997: as discussed offline, I think at this point you can force regardless if `self.pod_grace_period_seconds` is set or not.

liranbg: and when you delete the pod few lines above, you can skip using the explicit grace period

liranbg (changes requested):

liranbg (approved):

PR #4824, opened 2023-12-26
[Model Monitoring] Always add ±inf bins to histograms

jond01:
Fixes [ML-5387](https://jira.iguazeng.com/browse/ML-5387).

In "batch infer" model endpoints, these bins might not be present, resulting in an inconsistent behavior between shared and unshared features (e.g., "target").
It is beneficial and required to add these bins, as some "live" data may fall under them.

### DIFF
--- a/mlrun/common/model_monitoring/helpers.py
+++ b/mlrun/common/model_monitoring/helpers.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-#
 
 import sys
 import typing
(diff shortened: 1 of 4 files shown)


mlrun/common/model_monitoring/helpers.py
```diff
@@ -106,10 +105,21 @@ def _get_edges(hist: Histogram) -> BinEdges:
...
 def pad_hist(hist: Histogram) -> None:
-    """Add [-inf, x_0] and [x_n, inf] bins to the histogram inplace"""
+    """
+    Add [-inf, x_0] and [x_n, inf] bins to the histogram inplace unless present
+    """
     counts = _get_counts(hist)
     edges = _get_edges(hist)
 
+    def is_padded() -> bool:
+        factor = 1000
```
assaf758, line 115: why do we need the factor, padded histogram should have value +- _MAX_FLOAT at the edges..

assaf758 (approved):
nice!
added some minor comment/question however approved the PR

mlrun/common/model_monitoring/helpers.py
jond01: Right, I was worried about compatibility issues. I checked now and the value we use here:
https://github.com/mlrun/mlrun/blob/b322c4b8dbf1a8ec75bcc390407af5cf429ea90f/mlrun/common/model_monitoring/helpers.py#L30
Is not platform dependent:
https://docs.python.org/3/library/sys.html#sys.float_info

So this can be removed. I'll open a fix to remove this factor.

assaf758: Thanks!

jond01: https://github.com/mlrun/mlrun/pull/4827

PR #4825, opened 2023-12-26
[Artifacts] Fix timezone in created timestamp

TomerShor:
Artifacts created timestamps are saved in the DB as strings without timezone, and a milliseconds.
When first creating an artifact, the created timestamp includes timezone, but when updating it may not include it.
This can cause errors when updating artifacts and trying to force a format with timezone offset using `datetime.strptime`.

The fix is to always make sure the timestamp exists, whether creating a new artifact or updating it.

Fixes https://jira.iguazeng.com/browse/ML-5389


quaark (approved):
spoke offline about minor change. Otherwise LGTM! :)

PR #4826, opened 2023-12-26
[Spark] Make spark work with azure datastore profile

alxtkr77:
https://jira.iguazeng.com/browse/ML-5258

### DIFF
--- a/mlrun/datastore/azure_blob.py
+++ b/mlrun/datastore/azure_blob.py
@@ -151,19 +151,19 @@ class AzureBlobStore(DataStore):
                     if st[key] and st[key] != parsed_credential.get(key):
                         if key == "account_name":
                             raise mlrun.errors.MLRunInvalidArgumentError(
-                                f"'{key}' from the storage_options does not match corresponding connection string - \
-                                    {st[key]} != {parsed_credential.get(key)}"
+                                f"Storage option for '{key}' is '{st[key]}',"
+                                "which does not match corresponding connection string '{parsed_credential.get(key)}'"
                             )
                         else:
                             raise mlrun.errors.MLRunInvalidArgumentError(
-                                f"'{key}' from the storage_options does not match corresponding connection string"
+                                f"'{key}' from storage options does not match corresponding connection string"
                             )
                     st[key] = parsed_credential.get(key)
 
         account_name = st.get("account_name")
         if not account_name:
             raise mlrun.errors.MLRunInvalidArgumentError(
-                "The 'account_name' is absent both in 'storage_settings' and 'connection_string'"
+                "The 'account_name' is absent both in storage settings and connection string"
             )
         if primary_url:
             if primary_url.startswith("http://"):
--- a/mlrun/datastore/azure_blob.py
+++ b/mlrun/datastore/azure_blob.py
@@ -16,6 +16,7 @@ import time
 from pathlib import Path
 from urllib.parse import urlparse
 
+from azure.storage.blob._shared.base_client import parse_connection_str
 from fsspec.registry import get_filesystem_class
 
 import mlrun.errors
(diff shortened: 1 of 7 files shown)


mlrun/datastore/utils.py
```diff
@@ -15,21 +15,42 @@
...
-from urllib.parse import parse_qs, urlparse
+from urllib.parse import parse_qs, urlparse, urlunparse
 
 import pandas as pd
 
 import mlrun.datastore
 
 
-def store_path_to_spark(path):
+def store_path_to_spark(path, spark_options={}):
```
gtopper, line 25: Better avoid default of dict, even though it's safe here, but it raises warnings in IDE and looks suspicious to anyone reading the code.
```suggestion
def store_path_to_spark(path, spark_options=None):
```
Screenshot 2023-12-27 at 20 05 11

tests/system/feature_store/test_spark_azure.py
```diff
@@ -0,0 +1,145 @@
...
+    ParquetTarget,
+)
+from tests.system.base import TestMLRunSystem
+from tests.system.feature_store.expected_stats import expected_stats
+
+
+@TestMLRunSystem.skip_test_if_env_not_configured
+# Marked as enterprise because of v3io mount and remote spark
+@pytest.mark.enterprise
+class TestFeatureStoreAzureSparkEngine(TestMLRunSystem):
```
gtopper, line 37: Seems to me like duplicating the original test suite, with all of its utilities such as `run_local`, `spark_image_deployed`, `test_branch`, etc, is not a good idea. Better make a common base class. This common base would of course undo the duplication added in #4787 for S3.

gtopper (changes requested):

tests/system/feature_store/test_spark_azure.py
alxtkr77: On this matter, I hold a different viewpoint. When I attempted to augment `test_spark_engine.py` with S3 integration, the outcome was overly complex and challenging to comprehend or maintain. Subsequently, it dawned on me that our primary focus is testing the setup, such as the need for additional installations like `azure-storage-blob` and varying datastore profiles, rather than the test itself. The apparent duplication in the test is essentially just the necessary boilerplate.

gtopper: Well, I definitely wouldn't augment the original class, but rather extract a base to deduplicate about 100 lines of code and documentation that are duplicated per suite (potentially including the test itself, which is almost identical). Otherwise, it's x3 duplication, with the potential to grow (e.g. for gcs).

And even failing that, I can't see why we would want to duplicate the docstring verbatim.

mlrun/datastore/azure_blob.py
```diff
@@ -135,3 +136,69 @@ def listdir(self, key):
...
+                        else:
+                            raise mlrun.errors.MLRunInvalidArgumentError(
+                                f"'{key}' from the storage_options does not match corresponding connection string"
+                            )
+                    st[key] = parsed_credential.get(key)
+
+        account_name = st.get("account_name")
+        if not account_name:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "The 'account_name' is absent both in 'storage_settings' and 'connection_string'"
```
gtopper, line 166: I'm pretty sure there is no such thing as "storage_settings".

mlrun/datastore/azure_blob.py
```diff
@@ -135,3 +136,69 @@ def listdir(self, key):
...
+        if st.get("connection_string"):
+            primary_url, _, parsed_credential = parse_connection_str(
+                st.get("connection_string"), credential=None, service=service
+            )
+            for key in ["account_name", "account_key"]:
+                if parsed_credential.get(key):
+                    if st[key] and st[key] != parsed_credential.get(key):
+                        if key == "account_name":
+                            raise mlrun.errors.MLRunInvalidArgumentError(
+                                f"'{key}' from the storage_options does not match corresponding connection string - \
```
gtopper, line 154:
```suggestion
                                f"'{key}' from storage options does not match corresponding connection string - \
```
I don't think `storage_options` as a variable or function is user-facing.

mlrun/datastore/azure_blob.py
```diff
@@ -135,3 +136,69 @@ def listdir(self, key):
...
+                                    {st[key]} != {parsed_credential.get(key)}"
```
gtopper, line 155:
```suggestion
                                f"Storage option for '{key}' is '{st[key]}', which does not match corresponding connection string '{parsed_credential.get(key)}'"
```
Maybe this is a bit clearer.

mlrun/datastore/azure_blob.py
```diff
@@ -135,3 +136,69 @@ def listdir(self, key):
     def rm(self, path, recursive=False, maxdepth=None):
         path = self._convert_key_to_remote_path(key=path)
         super().rm(path=path, recursive=recursive, maxdepth=maxdepth)
+
+    def get_spark_options(self):
+        res = {}
+        st = self.get_storage_options()
+        service = "blob"
```
gtopper, line 143: I think it's fine to inline. Not like we're going to add more services.

alxtkr77: you never know . I would leave it

tests/system/feature_store/test_spark_azure.py
alxtkr77: Will try to do something about

tests/system/feature_store/spark_hadoop_test_base.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+
+    @classmethod
+    def configure_namespace(cls, test_suite_name="hadoop"):
+        cls.project_name = "fs-system-spark-engine-" + test_suite_name
+        cls.ds_profile_name = "ds_profile_spark_" + test_suite_name
+        cls.remote_function_name = (
+            "remote-spark-default-image-deploy-temp-" + test_suite_name
+        )
+        cls.spark_service_name = "dummy-spark-" + test_suite_name
```
gtopper, line 77: I don't think the service name should change between suites... The suite doesn't create the service – it depends on the system to already have it running. We don't want to have to create a service per suite.

Actually, this should be the same as in `TestFeatureStoreSparkEngine`, getting it from `env.yml`:
```python
    @classmethod
    def _init_env_from_file(cls):
        env = cls._get_env_from_file()
        if cls.run_local:
            cls.spark_service = None
        else:
            cls.spark_service = env["MLRUN_SYSTEM_TESTS_DEFAULT_SPARK_SERVICE"]
```

tests/system/feature_store/spark_hadoop_test_base.py
```diff
@@ -0,0 +1,154 @@
...
+        cls.pq_source = "testdata.parquet"
+        cls.pq_target = "testdata_target"
+
+    @classmethod
+    def configure_image_deployment(
+        cls,
+        deployment_type=Deployment.Remote,
+        additional_pip_pacakges=None,
```
gtopper, line 85: Typo
```suggestion
        additional_pip_packages=None,
```

tests/system/feature_store/spark_hadoop_test_base.py
```diff
@@ -0,0 +1,154 @@
...
+        if not cls.spark_image_deployed:
+            from mlrun import get_run_db
+            from mlrun.run import new_function
+
+            sj = new_function(kind="remote-spark", name=cls.remote_function_name)
+            if additional_pip_pacakges:
+                sj.spec.build.commands = [f"pip install {additional_pip_pacakges}"]
+            sj.spec.build.image = "." + cls.remote_function_name
+            sj.with_spark_service(spark_service=cls.spark_service_name)
+            sj.spec.image_pull_policy = "Always"
```
gtopper, line 103: Does this actually have any effect in this context? I think it probably doesn't...

gtopper (changes requested):
Nice! Looks much better. Still found a couple of things to fix and improve.

tests/system/feature_store/spark_hadoop_test_base.py
alxtkr77: Somehow without this setting the test did not fetch the image

mlrun/datastore/utils.py
```diff
@@ -15,21 +15,43 @@
...
     elif path.startswith("v3io:///"):
         path = "v3io:" + path[len("v3io:/") :]
+    elif path.startswith("az://"):
+        account_key = None
+        path = "wasbs:" + path[len("az:") :]
+        prefix = "spark.hadoop.fs.azure.account.key."
+        if spark_options:
+            for key in spark_options:
+                if key.startswith(prefix):
+                    account_key = key[len(prefix) :]
```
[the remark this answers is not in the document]
alxtkr77: There is only one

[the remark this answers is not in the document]
gtopper: So best add a break I think.

tests/system/feature_store/spark_hadoop_test_base.py
gtopper: It doesn't really make sense though... This function just deploys the image. It's not used by the test.

gtopper (approved):

PR #4829, opened 2023-12-27
[Logs] Validate positive offset when getting logs

TomerShor:
When getting logs, the offset is an offset inside the log file.
Passing a negative offset can cause several issues when calculating the chunk sizes and result with large logs returning in the responses, and errors such as:
```
MLRunRuntimeError: ("Connection broken: InvalidChunkLength(got length b'', 0 bytes read)", InvalidChunkLength(got length b'', 0 bytes read))
```

Resolves https://jira.iguazeng.com/browse/ML-5391

### DIFF
--- a/server/api/api/endpoints/logs.py
+++ b/server/api/api/endpoints/logs.py
@@ -79,6 +79,10 @@ async def get_log(
         server.api.api.deps.get_db_session
     ),
 ):
+    if offset < 0:
+        raise mlrun.errors.MLRunInvalidArgumentError(
+            "Offset cannot be negative",
+        )
     await server.api.utils.auth.verifier.AuthVerifier().query_project_resource_permissions(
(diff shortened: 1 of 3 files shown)


server/api/api/endpoints/logs.py
```diff
@@ -79,6 +79,10 @@ async def get_log(
         server.api.api.deps.get_db_session
     ),
 ):
+    if offset < 0:
```
theSaarco, line 82: Consider using:
```
offset: int = Query(0, ge=0),
```
Which will enforce `>=0` already in FastAPI, and should also display the limitation in the Swagger page. Only thing is to validate the error message is clear in that case.

theSaarco (approved):
Looks good. Have a suggestion for different implementation of the check in the BE.

PR #4830, opened 2023-12-27
[Model Monitoring] Fix delete old parquet function

davesh0812:
Parquet files are stored on the user side, while the controller operates on the server side. To delete these Parquet files through the controller, it's necessary to establish a file system with the appropriate v3io access key for the relevant permissions.

[ML-4982](https://jira.iguazeng.com/browse/ML-4982)

### DIFF
--- a/mlrun/datastore/v3io.py
+++ b/mlrun/datastore/v3io.py
@@ -86,7 +86,9 @@ class V3ioStore(DataStore):
                 ) from exc
             return None
         if v3io_access_key:
-            self._filesystem = fsspec.filesystem("v3io", v3io_access_key=v3io_access_key, v3io_api=mlrun.mlconf.v3io_api)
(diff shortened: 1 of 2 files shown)


jond01 (approved):
Nice 👍

mlrun/datastore/v3io.py
```diff
@@ -85,7 +85,12 @@ def get_filesystem(self, silent=True):
                     "v3iofs or storey not installed, run pip install storey"
                 ) from exc
             return None
-        self._filesystem = fsspec.filesystem("v3io", **self.get_storage_options())
```
alxtkr77, line 88: I would advise against altering the store.get_filesystem() API since it's an integral part of the DataStore base class. Instead, I would add a new user_store within the controller, incorporating the user's v3io_api_key for specific functionality.

alxtkr77 (approved):

PR #4832, opened 2023-12-27
[Model Monitoring] Save only the new fields that generated by the user to ModelMonitoringApplicationResult 

davesh0812:
[ML-5414](https://jira.iguazeng.com/browse/ML-5414)


assaf758 (approved):

mlrun/model_monitoring/application.py
```diff
@@ -284,15 +260,33 @@ def __init__(
...
+            ],
+            mm_constant.WriterEvent.END_INFER_TIME: origin_event[
+                mm_constant.ApplicationEvent.END_INFER_TIME
+            ],
+            mm_constant.WriterEvent.CURRENT_STATS: json.dumps(
+                origin_event[mm_constant.ApplicationEvent.CURRENT_STATS]
+            ),
+        }
+        for result in results:
+            data = result.to_dict().update(metadata)
```
jond01, line 289: This statement sets `data` to `None` - you misuse the `update` method.
https://docs.python.org/3/library/stdtypes.html#dict.update

mlrun/model_monitoring/application.py
```diff
@@ -284,15 +260,33 @@ def __init__(
...
         """
         Push application results to the monitoring writer stream.
 
-        :param event: Monitoring result(s) to push.
+        :param event: Monitoring result(s) to push and the original event from the controller.
         """
         self._lazy_init()
-        for result in event:
-            data = result.to_dict()
+        results, origin_event = event
```
jond01, line 270: `application_results, application_event` is more readable IMO

jond01 (changes requested):
Please see the first comment and run the system test

mlrun/model_monitoring/application.py
davesh0812: ops thanks:)

tests/system/feature_store/test_feature_store.py
```diff
@@ -4548,7 +4548,7 @@ def test_merge_different_number_of_entities(self):
         vec = fstore.FeatureVector(
             "vector_partyaccount", features, join_graph=join_graph
         )
-        df = fstore.get_offline_features(vec).to_dataframe()
+        df = vec.get_offline_features().to_dataframe()
```
jond01, line 4551: How are these changes in this file related to the PR?

mlrun/model_monitoring/application.py
```diff
@@ -284,17 +260,35 @@ def __init__(
...
+                mm_constant.ApplicationEvent.END_INFER_TIME
+            ],
+            mm_constant.WriterEvent.CURRENT_STATS: json.dumps(
+                application_event[mm_constant.ApplicationEvent.CURRENT_STATS]
+            ),
+        }
+        for result in application_results:
+            result.to_dict().update(metadata)
+            logger.info(f"Pushing data = {result} \n to stream = {self.stream_uri}")
+            self.output_stream.push([result])
```
jond01, line 291: I still don't see how it would work. You update a dictionary object returned by `result.to_dict()`, but pass `result` which is not the same object.

I kindly ask you to run the system test and request a review only after it passes.

jond01 (changes requested):

assaf758 (commented):
@davesh0812, does the test-app system-test pass with this last revision of the PR?

PR #4835, opened 2023-12-27
[Project] Enable adding git to a project after project creation

Yacouby:
Creating a project using mlrun.get_or_create_project() without specifying init_git=True and then using create_remote() was causing an error even if git is already initialized in the current context.

resolves :
https://jira.iguazeng.com/browse/ML-5078

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -2292,8 +2292,7 @@ class MlrunProject(ModelObj):
         :param name:   name for the remote (default is 'origin')
         :param branch: Git branch to use as source
         """
-        if not self.spec.repo and not self.is_git_initialized():
-            raise ValueError("git repo is not valid/initialized")
+        self._ensure_git_repo()
         self.spec.repo.create_remote(name, url=url)
         url = url.replace("https://", "git://")
         if not branch:
@@ -2306,19 +2305,25 @@ class MlrunProject(ModelObj):
         self.spec._source = self.spec.source or url
         self.spec.origin_url = self.spec.origin_url or url
 
-    def is_git_initialized(self):
+    def _ensure_git_repo(self):
+        if self.spec.repo:
+            return
         context = self.context
         git_dir_path = os.path.join(context, ".git")
 
-        if os.path.exists(git_dir_path) and os.path.isdir(git_dir_path):
+        if not os.path.exists(git_dir_path):
+            logger.warning("Git repository not initialized. initializing now")
+            self.spec.repo = git.Repo.init(context)
+        else:
+            # git already initialized
             try:
                 self.spec.repo = git.Repo(context)
-                return True
             except git.InvalidGitRepositoryError:
                 logger.warning("Invalid Git repository")
             except Exception as e:
-                logger.error(f"Error initializing Git repository: {e}")
-        return False
+                logger.error(
+                    f"Error initializing Git repository: {mlrun.errors.err_to_str(e)}"
+                )
 
     def push(
         self,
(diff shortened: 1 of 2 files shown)


mlrun/projects/project.py
```diff
@@ -2291,8 +2292,8 @@ def create_remote(self, url, name="origin", branch=None):
         :param name:   name for the remote (default is 'origin')
         :param branch: Git branch to use as source
         """
-        if not self.spec.repo:
-            raise ValueError("git repo is not set/defined")
+        if not self.spec.repo and not self.is_git_initialized():
+            raise ValueError("git repo is not valid/initialized")
```
alonmr, line 2296:
```suggestion
        if not self.spec.repo:
           self._ensure_git_repo()
```

mlrun/projects/project.py
```diff
@@ -2305,6 +2306,20 @@ def create_remote(self, url, name="origin", branch=None):
         self.spec._source = self.spec.source or url
         self.spec.origin_url = self.spec.origin_url or url
 
+    def is_git_initialized(self):
```
alonmr, line 2309:
```suggestion
    def _ensure_git_repo(self):
```

mlrun/projects/project.py
```diff
@@ -2305,6 +2306,20 @@ def create_remote(self, url, name="origin", branch=None):
...
+        context = self.context
+        git_dir_path = os.path.join(context, ".git")
+
+        if os.path.exists(git_dir_path) and os.path.isdir(git_dir_path):
```
alonmr, line 2313: check if path does not exist, then log a warning that git was not initialized and we are about to do it. use
```
repo = git.Repo.init(context)
```
otherwise, create the repo from the context like you did below

alonmr (changes requested):
I think we can make it easier on the user and just create the git init if it doesn't exist instead of dragging him around with error messages saying do that that and that.

mlrun/projects/project.py
```diff
@@ -2292,7 +2293,7 @@ def create_remote(self, url, name="origin", branch=None):
         :param branch: Git branch to use as source
         """
         if not self.spec.repo:
-            raise ValueError("git repo is not set/defined")
+            self._ensure_git_repo()
```
alonmr, line 2296: Can make it a bit cleaner if you would move the if above to inside the function.
So that inside the function you will have:
~~~
if self.spec.repo:
    return
~~~

mlrun/projects/project.py
```diff
@@ -2305,6 +2306,22 @@ def create_remote(self, url, name="origin", branch=None):
...
+            logger.warning("Git repository not initialized. initializing now")
+            self.spec.repo = git.Repo.init(context)
+        else:
+            # git already initialized
+            try:
+                self.spec.repo = git.Repo(context)
+            except git.InvalidGitRepositoryError:
+                logger.warning("Invalid Git repository")
+            except Exception as e:
+                logger.error(f"Error initializing Git repository: {e}")
```
alonmr, line 2323:
```suggestion
                logger.error(f"Error initializing Git repository: {mlrun.errors.err_to_str(e)}")
```

tests/projects/test_project.py
```diff
@@ -1477,3 +1477,30 @@ def test_load_project_from_yaml_with_function(context):
             )
             == {}
         )
+
+
+def test_project_create_remote():
+    # test calling create_remote without git_init=True on project creation
+
+    # Set the current working directory to a temporary directory
+    temp_dir = tempfile.mkdtemp()
```
alonmr, line 1486: can do it with context manager for easier cleanup:
~~~
with tempfile.TemporaryDirectory() as tmp_dir:
~~~

alonmr (commented):
Looks great! some minor comments for tidiness

PR #4840, opened 2023-12-28
[DB] Fix reusing db sessions

liranbg:
https://jira.iguazeng.com/browse/ML-5424

ensured db session is created for when listing function hubs to avoid race condition accessing to db session

### DIFF
--- a/mlrun/run.py
+++ b/mlrun/run.py
@@ -367,7 +367,7 @@ def import_function(url="", secrets=None, db="", project=None, new_name=None):
         if not runtime:
             raise KeyError(f"function {name}:{tag} not found in the DB")
     else:
-        url, is_hub_uri = extend_hub_uri_if_needed(url, db)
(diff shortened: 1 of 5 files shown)


mlrun/run.py
```diff
@@ -367,7 +367,7 @@ def import_function(url="", secrets=None, db="", project=None, new_name=None):
         if not runtime:
             raise KeyError(f"function {name}:{tag} not found in the DB")
     else:
-        url, is_hub_uri = extend_hub_uri_if_needed(url)
+        url, is_hub_uri = extend_hub_uri_if_needed(url, db)
```
alonmr, line 370: seems like db is expected to be a string
```
:param db: optional, mlrun api/db path
```

liranbg: few lines up it changes it to rundb db is a string

alonmr: you are in the else clause

liranbg: oh, right right

PR #4842, opened 2023-12-28
[Project] Set default build base image

alonmr:
https://jira.iguazeng.com/browse/ML-4784

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -3035,7 +3035,6 @@ class MlrunProject(ModelObj):
         overwrite_build_params: bool = False,
         requirements_file: str = None,
         extra_args: str = None,
-        force_build: bool = False,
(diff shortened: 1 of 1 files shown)


mlrun/projects/project.py
```diff
@@ -3035,7 +3035,6 @@ def build_image(
         overwrite_build_params: bool = False,
         requirements_file: str = None,
         extra_args: str = None,
-        force_build: bool = False,
```
liranbg, line 3038: Better leave for bc but not using it

theSaarco: +1 to that - keep it, mark deprecated, and warn that it's ignored and always `True`.

alonmr: it was introduced in 1.6.0 so it's safe to delete ;)

PR #4843, opened 2023-12-28
[K8s] Map k8s api exceptions to mlrun exceptions

alonmr:
When there is no error handling, we can use a decorator to catch the ApiException and match it with an MLRun exception.
When there is some error handling, do it inside the function.
This will help giving users more indicative error instead of some 500 Internal error with no error message.
Also, added a retry for store secret in case of race condition where secret was created just before we were about to create it.

https://jira.iguazeng.com/browse/ML-5338


server/api/utils/singletons/k8s.py
```diff
@@ -46,6 +46,21 @@ def get_k8s_helper(namespace=None, silent=True, log=False) -> "K8sHelper":
...
+    """
+    A decorator for calls to k8s api when no error handling is needed.
+    Raises the matching mlrun exception to the status code.
+    """
+
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except ApiException as exc:
+            mlrun.errors.raise_for_status_code(exc.status, message=exc.reason)
```
liranbg, line 59: you lose traceback this one, try making it maybe
```suggestion
            raise mlrun.errors.raise_for_status_code(exc.status, message=exc.reason) from exc
```

and raise_for_status_code will simply return the exception (not raising it)

server/api/utils/singletons/k8s.py
```diff
@@ -454,7 +504,8 @@ def delete_secrets(
                 return None
             else:
                 logger.error(
-                    f"failed to retrieve k8s secret: {mlrun.errors.err_to_str(exc)}"
+                    "Failed to retrieve k8s secret",
+                    exc=mlrun.errors.err_to_str(exc),
                 )
```
liranbg, line 509: add `secret_name` as log param

server/api/utils/singletons/k8s.py
```diff
@@ -388,16 +421,32 @@ def store_secrets(
...
-            self.v1api.create_namespaced_secret(namespace, k8s_secret)
-            return mlrun.common.schemas.SecretEventActions.created
+            try:
+                self.v1api.create_namespaced_secret(namespace, k8s_secret)
+                return mlrun.common.schemas.SecretEventActions.created
+            except ApiException as exc:
+                if exc.status == 409 and retry_on_conflict:
+                    logger.warning(
+                        "Secret was created while we tried to create it, retrying...",
+                        exc=mlrun.errors.err_to_str(exc),
```
liranbg, line 439: add secret name as log param

server/api/utils/singletons/k8s.py
```diff
@@ -388,16 +421,32 @@ def store_secrets(
             # If secret doesn't exist, we'll simply create it
             if exc.status != 404:
                 logger.error(
-                    f"failed to retrieve k8s secret: {mlrun.errors.err_to_str(exc)}"
+                    "Failed to retrieve k8s secret", exc=mlrun.errors.err_to_str(exc)
```
liranbg, line 424: add secret name as log param

server/api/utils/singletons/k8s.py
```diff
@@ -269,9 +293,12 @@ def get_project_vault_secret_name(
             # It's valid for the service account to not exist. Simply return None
             if exc.status != 404:
                 logger.error(
-                    f"failed to retrieve service accounts: {mlrun.errors.err_to_str(exc)}"
+                    "Failed to retrieve service accounts",
+                    exc=mlrun.errors.err_to_str(exc),
```
liranbg, line 297: add service account name to log param?

server/api/utils/singletons/k8s.py
```diff
@@ -172,8 +191,10 @@ def get_pod(self, name, namespace=None, raise_on_not_found=False):
             return api_response
         except ApiException as exc:
             if exc.status != 404:
-                logger.error(f"failed to get pod: {mlrun.errors.err_to_str(exc)}")
-                raise exc
+                logger.error("Failed to get pod", exc=mlrun.errors.err_to_str(exc))
```
liranbg, line 194: add pod name to log param

liranbg (changes requested):
overall looking good, just add the resource name when logging the failure/warning

liranbg (approved):

PR #4846, opened 2023-12-30
[CI] Updating system test

liranbg:
1. Updated prepare script to leverage new provctl features + tweaks and bug fixes
2. Run system tests against self-hosted runner

since we run tests under github self hosted runner - we use containers for each job steps. that means, system test should/cannot run dockerized because we dont have docker mounted to the job steps container. that means, system test will now run from the container itself that runs the jobs. for simplicity, ive removed the need of schemas to be compiled for system test as those schemas are not needed by the system test container

TODO: following this PR:
1. create system test container based on python 3.9 that will have all requirements pre-installed. this will speed up each test suite bootstrapping
2. use github api to get latest 1 hour old commit sha (instead of cloning)


.github/workflows/system-tests-enterprise.yml
```diff
@@ -194,12 +194,9 @@ jobs:
...
         # Get the tested mlrun version
         echo "unstable_version_prefix=$( \
-          cd /tmp && \
-          git clone --single-branch --branch ${{ steps.current-branch.outputs.name }} https://github.com/mlrun/mlrun.git mlrun-upstream 2> /dev/null && \
-          cd mlrun-upstream && \
-          cat automation/version/unstable_version_prefix && \
-          cd .. && \
-          rm -rf mlrun-upstream)" >> $GITHUB_OUTPUT
+          curl https://raw.githubusercontent.com/mlrun/mlrun/${{ steps.current-branch.outputs.name }}/automation/version/unstable_version_prefix \
+        )" >> $GITHUB_OUTPUT
```
alonmr, line 198:
```suggestion
         >> $GITHUB_OUTPUT
```

.github/workflows/system-tests-enterprise.yml
```diff
@@ -83,39 +85,37 @@ jobs:
...
+      image: python:3.9
     needs: [system-test-cleanup]
     # let's not run this on every fork, change to your fork when developing
     if: github.repository == 'mlrun/mlrun' || github.event_name == 'workflow_dispatch'
 
     steps:
     - uses: actions/checkout@v3
+    - name: Install dependencies
+      run: |
+        apt-get update -qqy && apt-get install -y sshpass jq curl gnupg nodejs
```
alonmr, line 103: I'm guessing nodejs is needed for some specific tests?

alonmr (approved):
Looks good! Minor comment and question

.github/workflows/system-tests-enterprise.yml
liranbg: to run some github actions steps

.github/workflows/system-tests-enterprise.yml
liranbg: see line 196, it opens (" so the closure is at 198

PR #4849, opened 2023-12-31
[Build] Update build command

liranbg:
https://jira.iguazeng.com/browse/ML-5435 - [https://packaging.python.org/en/latest/discussions/setup-py-deprecated/](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/:)

license update - https://jira.iguazeng.com/browse/ML-5436

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -152,11 +152,17 @@ jobs:
           cache: pip
       - name: Install dependencies
         run: |
-          pip install -r automation/requirements.txt
+          # needed for updating the version file prior to publishing
+          python -m pip install -r automation/requirements.txt
+          
+          # TODO: move to a separated requirements file
+          python -m pip install \
+            build~=1.0 \
+            wheel~=0.42 \
+            twine~=4.0
       - name: Build & push to pypi
         if: github.event.inputs.skip_publish_pypi != 'true'
         run: |
-          pip install twine wheel build
           export TWINE_USERNAME=${{ secrets.PYPI_USERNAME }}
           export TWINE_PASSWORD=${{ secrets.PYPI_PASSWORD }}
           MLRUN_VERSION="$INPUT_VERSION" make publish-package
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -156,7 +156,7 @@ jobs:
       - name: Build & push to pypi
         if: github.event.inputs.skip_publish_pypi != 'true'
         run: |
-          pip install twine wheel
+          pip install twine wheel build
           export TWINE_USERNAME=${{ secrets.PYPI_USERNAME }}
           export TWINE_PASSWORD=${{ secrets.PYPI_PASSWORD }}
           MLRUN_VERSION="$INPUT_VERSION" make publish-package
--- a/Makefile
+++ b/Makefile
@@ -416,7 +416,7 @@ build-test-system: compile-schemas update-version-file ## Build system tests doc
 
 .PHONY: package-wheel
 package-wheel: clean update-version-file ## Build python package wheel
-	python setup.py bdist_wheel
+	python -m build --wheel
 
 .PHONY: publish-package
 publish-package: package-wheel ## Publish python package wheel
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -416,7 +416,7 @@ build-test-system: compile-schemas update-version-file ## Build system tests doc
 
 .PHONY: package-wheel
 package-wheel: clean update-version-file ## Build python package wheel
-	python setup.py bdist_wheel
+	python -m build --wheel
```
alonmr, line 419: this requires `build` python package. add it to dev-requiremetns

.github/workflows/release.yaml
```diff
@@ -156,7 +156,7 @@ jobs:
       - name: Build & push to pypi
         if: github.event.inputs.skip_publish_pypi != 'true'
         run: |
-          pip install twine wheel
+          pip install twine wheel build
```
jond01, line 159: 1. Is `wheel` required?
2. I recommend tracking the versions of the required 3rd party packages in
   https://github.com/mlrun/mlrun/blob/97969648941d362cd235b05b844221353c842073/dev-requirements.txt#L1-L2
  and installing compatible versions from this file with the `--constraint` option: https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-c

Makefile
jond01, line 419: It's generally recommended to upload also the source distribution and not only the wheel:
https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#id1

But maybe let's keep it for another time and document in Jira

Makefile
```diff
@@ -428,10 +428,8 @@ test-publish: package-wheel ## Test python package publishing
 
 .PHONY: clean
 clean: ## Clean python package build artifacts
-	rm -rf build
-	rm -rf dist
-	rm -rf mlrun.egg-info
-	find . -name '*.pyc' -exec rm {} \;
+	rm -rf build dist mlrun.egg-info
```
jond01, line 431: Nit: why `-f`?

setup.py
```diff
@@ -55,7 +55,7 @@ def version():
     long_description_content_type="text/markdown",
     author="Yaron Haviv",
     author_email="yaronh@iguazio.com",
-    license="MIT",
+    license="Apache License Version 2.0",
```
jond01, line 58: I think that it should be:
```suggestion
    license="Apache-2.0",
```
https://spdx.org/licenses/Apache-2.0.html

liranbg: id go with full name (`Apache License 2.0`)

Makefile
liranbg: to avoid errors of deleting the directotories

.github/workflows/release.yaml
liranbg: whatever performs this command needs to install `dev-requirements.txt` - updated the job step as well

liranbg: I am taking my words back. installing `dev-requirements.txt` is an overkill to this step as `dev-requirements.txt` includes ml frameworks for development and not for release purposes

Makefile
liranbg: https://github.com/mlrun/mlrun/pull/4849#discussion_r1438879946

.github/workflows/release.yaml
jond01: I meant to:
```sh
pip install twine build -c dev-requirements.txt
```
This installs only the required packages but with the constraints listed in the file

liranbg: and @jond01  `--constraint` wont work because `dev-requirements` has `extras` (see `hypothesis[numpy]`).
I will default to inline it on job step and add a TODO to reunite requiremnts files better

jond01: 👍 
Regarding the `wheel` package, I'm not sure it's required.
And regarding `build` - let's add it to `dev-requirements.txt` anyway to allow smooth local builds

liranbg: I would add that to a separate requirements txt alltogether

PR #4852, opened 2023-12-31
[HTTPDB] Keep watching on created and aborting states

TomerShor:
When "watching" a run, keep watching if the run is in transitional states, which include:
- `pending`
- `running`
- `created`
- `aborting`

Also, use constants instead of raw strings.

Resolves https://jira.iguazeng.com/browse/ML-5430

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -574,7 +574,12 @@ class HTTPRunDB(RunDBInterface):
             else:
                 nil_resp += 1
 
-            if watch and state in ["pending", "running"]:
+            if watch and state in [
+                mlrun.runtimes.constants.RunStates.pending,
+                mlrun.runtimes.constants.RunStates.running,
+                mlrun.runtimes.constants.RunStates.created,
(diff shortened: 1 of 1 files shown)


mlrun/db/httpdb.py
```diff
@@ -574,7 +574,12 @@ def watch_log(self, uid, project="", watch=True, offset=0):
             else:
                 nil_resp += 1
 
-            if watch and state in ["pending", "running"]:
+            if watch and state in [
+                mlrun.runtimes.constants.RunStates.pending,
+                mlrun.runtimes.constants.RunStates.running,
+                mlrun.runtimes.constants.RunStates.created,
```
liranbg, line 580: im not sure created is a legit state to watch on as it might not change overtime but would remain as is
I would omit that

TomerShor: But if the run state is `created` at the time of checking, the watch would stop before it even started running, which means that this won't wait for the run to be completed at all.

PR #4860, opened 2024-01-02
[Model Monitoring] Improve the controller job's logs

jond01:
Resolves [ML-5347](https://jira.iguazeng.com/browse/ML-5347).

### DIFF
--- a/mlrun/model_monitoring/controller.py
+++ b/mlrun/model_monitoring/controller.py
@@ -317,21 +317,25 @@ class MonitoringApplicationController:
 
     def run(self):
         """
-        Main method for run all the relevant monitoring application on each endpoint
+        Main method for run all the relevant monitoring applications on each endpoint
         """
         try:
             endpoints = self.db.list_model_endpoints(uids=self.model_endpoints)
-            application = mlrun.get_or_create_project(
+            monitoring_functions = mlrun.get_or_create_project(
                 self.project
             ).list_model_monitoring_functions()
-            if application:
-                applications_names = list({app.metadata.name for app in application})
+            if monitoring_functions:
+                applications_names = list(
+                    {app.metadata.name for app in monitoring_functions}
+                )
             else:
-                logger.info("There are no monitoring application found in this project")
+                self.context.logger.info(
+                    "No monitoring functions found", project=self.project
+                )
                 applications_names = []
 
         except Exception as e:
-            logger.error("Failed to list endpoints", exc=e)
+            self.context.logger.error("Failed to list endpoints", exc=e)
             return
         if endpoints and applications_names:
             # Initialize a process pool that will be used to run each endpoint applications on a dedicated process
--- a/mlrun/model_monitoring/controller.py
+++ b/mlrun/model_monitoring/controller.py
@@ -488,11 +488,10 @@ class MonitoringApplicationController:
                         model_monitoring_access_key=model_monitoring_access_key,
                         parquet_target_path=parquet_target_path,
                     )
-        except Exception as e:
-            logger.error(
+        except Exception:
+            logger.exception(
                 "Encountered an exception",
                 endpoint_id=endpoint[mm_constants.EventFieldType.UID],
-                exc=e,
             )
 
     def _delete_old_parquet(self, endpoints: list[dict[str, Any]], days: int = 1):
--- a/mlrun/model_monitoring/controller.py
+++ b/mlrun/model_monitoring/controller.py
@@ -219,7 +219,7 @@ class _BatchWindowGenerator:
         cls, first_request: Optional[str], endpoint: str
     ) -> Optional[int]:
         if not first_request:
-            logger.warn(
+            logger.debug(
                 "There is no first request time for this endpoint.",
                 endpoint=endpoint,
                 first_request=first_request,
--- a/mlrun/model_monitoring/controller.py
+++ b/mlrun/model_monitoring/controller.py
@@ -17,7 +17,6 @@ import datetime
 import json
 import os
 import re
-from contextlib import contextmanager
 from typing import Any, Iterator, NamedTuple, Optional, Union, cast
 
 from v3io.dataplane.response import HttpResponseError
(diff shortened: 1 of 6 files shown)


mlrun/model_monitoring/controller.py
```diff
@@ -55,29 +68,32 @@ def __init__(
...
+                    table_path=self._endpoint,
+                    key=self._application,
+                )
         except HttpResponseError as err:
-            logger.warn(
-                "Failed to get the last analyzed time for this endpoint and application, "
+            logger.info(
+                "No last analyzed time for this endpoint and application, "
                 "as this is probably the first time this application is running. "
                 "Using the first request time instead.",
```
assaf758, line 91:
```suggestion
                "Using the first-request time as last-analyzed.",
```

mlrun/model_monitoring/controller.py
```diff
@@ -55,29 +68,32 @@ def __init__(
...
                 endpoint=self._endpoint,
                 application=self._application,
                 first_request=self._first_request,
-                error=err,
             )
+            logger.debug("Error while getting last analyzed time", err=err)
```
assaf758, line 96: why an extra log?

mlrun/model_monitoring/controller.py
```diff
@@ -55,29 +68,32 @@ def __init__(
...
+            with _adapt_logger_level(logger=logger, level=logging.ERROR):
+                data = self._kv_storage.get(
+                    container=self._v3io_container,
+                    table_path=self._endpoint,
+                    key=self._application,
+                )
         except HttpResponseError as err:
-            logger.warn(
-                "Failed to get the last analyzed time for this endpoint and application, "
+            logger.info(
```
assaf758, line 88: I think we should have app invocation counter. this way log an error if it's not the first invocation, and silently ignore this error (since it's expected) on the first invocation.

assaf758 (commented):
Nice work!
I do have some comments.

mlrun/model_monitoring/controller.py
```diff
@@ -426,22 +434,18 @@ def model_endpoint_process(
                         parquet_target_path = offline_response.vector.get_target_path()
 
                         if len(df) == 0:
-                            logger.warn(
-                                "Not enough model events since the beginning of the batch interval",
-                                featureset_name=m_fs.metadata.name,
+                            logger.info(
+                                "Empty window",
```
davesh0812, line 438:
```suggestion
                                "During this time window, the endpoint has not received any data.",
```

mlrun/model_monitoring/controller.py
```diff
@@ -426,22 +434,18 @@ def model_endpoint_process(
...
                                 endpoint=endpoint[mm_constants.EventFieldType.UID],
-                                min_required_events=mlrun.mlconf.model_endpoint_monitoring.parquet_batching_max_events,
                                 start_time=start_infer_time,
                                 end_time=end_infer_time,
                             )
                             continue
 
-                    # Continue if not enough events provided since the deployment of the model endpoint
                     except FileNotFoundError:
```
davesh0812, line 445: As we see the get offline function creates parquet even if it is empty, so I think we need to catch it here (maybe the last catch is enough), and if we catching it we need to change the logger level to warn/error.

davesh0812 (changes requested):

mlrun/model_monitoring/helpers.py
```diff
@@ -55,24 +56,22 @@ def get_stream_path(project: str = None, application_name: str = None):
...
     :return:           Monitoring parquet target path.
     """
-
-    project_obj = mlrun.get_or_create_project(name=project)
-    artifact_path = project_obj.spec.artifact_path
+    artifact_path = project.spec.artifact_path
     # Generate monitoring parquet path value
     parquet_path = mlrun.mlconf.get_model_monitoring_file_target_path(
-        project=project,
+        project=project.name,
```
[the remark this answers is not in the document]
jond01: It's the same

mlrun/model_monitoring/controller.py
```diff
@@ -55,29 +68,32 @@ def __init__(
...
         self._step = timedelta_seconds
 
     def _get_last_analyzed(self) -> Optional[int]:
         try:
-            data = self._kv_storage.get(
-                container=self._v3io_container,
-                table_path=self._endpoint,
-                key=self._application,
-            )
+            with _adapt_logger_level(logger=logger, level=logging.ERROR):
```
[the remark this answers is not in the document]
jond01: b2d599a:
```
[warning] Response error: Request failed with status 404: b'{\n\t"ErrorCode": -2,\n\t"ErrorMessage": "No such file or directory"\n}'
```
This happens before the first window (no last analyzed for the app).

mlrun/model_monitoring/controller.py
jond01: To log the raw error but at a debug level.
This log is usually uninformative.

mlrun/model_monitoring/controller.py
jond01: Created [ML-5496](https://jira.iguazeng.com/browse/ML-5496) following our discussion

mlrun/model_monitoring/controller.py
jond01: eeb9f74

mlrun/model_monitoring/controller.py
jond01: d711f75

mlrun/model_monitoring/controller.py
jond01: 64f1f75

mlrun/model_monitoring/controller.py
```diff
@@ -30,10 +32,21 @@
 from mlrun.datastore.targets import ParquetTarget
 from mlrun.model_monitoring.batch import calculate_inputs_statistics
 from mlrun.model_monitoring.helpers import get_monitoring_parquet_path, get_stream_path
-from mlrun.utils import logger
+from mlrun.utils import Logger, logger
 from mlrun.utils.v3io_clients import get_v3io_client
 
 
+@contextmanager
+def _adapt_logger_level(logger: Logger, level: int) -> Iterator[None]:
```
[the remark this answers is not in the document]
jond01: I removed this utility, using now a new logger instead with `create_logger`.
1729b76

assaf758 (approved):
Looks good!

PR #4864, opened 2024-01-03
[Artifacts] Fix artifact tagging

TomerShor:
Some fixes regarding artifact tagging:
1. When tagging artifacts, remove the same tag for artifacts with the same `key` and `project` but with a different `producer_id` before adding the new tags, so we won't have multiple artifacts from different producers with the same tag.
2. Consider iteration when tagging artifacts - In a hyper param run, all created iterations must have the "latest" tag, and not only the most recent iteration.

Resolves https://jira.iguazeng.com/browse/ML-5432


server/api/db/sqldb/db.py
```diff
@@ -3100,6 +3155,29 @@ def _re_tag_existing_object(
 
         return None
 
+    def _re_tag_existing_artifact(
```
theSaarco, line 3158: Why is this needed? There's no code in this PR that calls it?

server/api/db/sqldb/db.py
```diff
@@ -897,6 +884,74 @@ def delete_tag_from_artifacts(
...
+                    artifact.Tag,
+                    name=tag_name,
+                    project=project,
+                )
+                .join(
+                    ArtifactV2,
+                )
+                .filter(
+                    ArtifactV2.key == artifact.key,
+                    ArtifactV2.project == project,
```
theSaarco, line 910: Being extra-picky - these predicates are redundant (the query on the tag covers those already)

server/api/db/sqldb/db.py
```diff
@@ -897,6 +884,74 @@ def delete_tag_from_artifacts(
...
+                    name=tag_name,
+                    project=project,
+                    obj_name=artifact.key,
+                )
+                .join(
+                    ArtifactV2,
+                )
+                .filter(
+                    ArtifactV2.key == artifact.key,
+                    ArtifactV2.project == project,
```
theSaarco, line 934: redundant

tests/api/db/test_artifacts.py
```diff
@@ -853,6 +854,83 @@ def test_list_artifact_for_tagging_fallback(
...
+
+        # list artifact with "latest" tag - should return all artifacts
+        artifacts = db.list_artifacts(db_session, project=project, tag="latest")
+        assert len(artifacts) == iteration_number
+
+        # mark iteration 3 as the best iteration
+        best_iteration = 3
+        self._mark_best_iteration_artifact(
+            db, db_session, project, artifact_key, artifact_tree, best_iteration
+        )
```
theSaarco, line 886: Instead of writing an extra function, you could have already marked that in the `store_artifact` above (with the `best_iteration` param), or did you want to specifically test the usage of link artifacts for that?

theSaarco (approved):
Looks good. I have some comments, none critical - please take a look.

server/api/db/sqldb/db.py
TomerShor: Leftover from some of my tests. Removing.

tests/api/db/test_artifacts.py
TomerShor: @theSaarco I wanted to both test the link artifact functionality and to make sure the tagging isn't affected by the iterations / best iteration.

theSaarco (approved):
Sababa

PR #4865, opened 2024-01-03
[Background Tasks] Lock Internal Background Tasks by Kind

quaark:
- Split Background task file to module
- Introduce background task kinds
- Lock Background tasks so only one task of a certain kind can be running concurrently
- Delete old background tasks that are finished (keep history of one previous task)
- Added Async Multithreaded test client to background task UT in order to test this concurrency feature.


server/api/utils/background_tasks/internal.py
```diff
@@ -230,3 +172,32 @@ def _generate_background_task(
         return mlrun.common.schemas.BackgroundTask(
             metadata=metadata, spec=spec, status=status
         )
+
+    def _get_active_task_name_by_kind(self, kind: str):
+        return self._background_tasks_kind_locks.get(kind, (None, None))[0]
+
+    def _get_previous_task_name_by_kind(self, kind: str):
```
TomerShor, line 179: When/where is this being used? 
and why are we saving the previous task at all?

TomerShor (commented):
just a question.
Otherwise looks good!

server/api/utils/background_tasks/internal.py
```diff
@@ -20,112 +20,32 @@
...
 class InternalBackgroundTasksHandler(metaclass=mlrun.utils.singleton.Singleton):
     def __init__(self):
         self._internal_background_tasks: typing.Dict[
             str, mlrun.common.schemas.BackgroundTask
         ] = {}
 
+        # contains a lock for each background task kind, with the following format:
+        # {kind: [active_name, previous_name]}
+        self._background_tasks_kind_locks: typing.Dict[
+            str, typing.Tuple[typing.Optional[str], typing.Optional[str]]
```
liranbg, line 41: id suggest to `kind` it as a const of kinds so that it would be easier to search usages of such

server/api/utils/background_tasks/internal.py
```diff
@@ -134,8 +54,15 @@ def create_background_task(
         # sanity
         if name in self._internal_background_tasks:
             raise RuntimeError("Background task name already exists")
-        background_task = self._generate_background_task(name)
+
+        if self._get_active_task_name_by_kind(kind):
+            raise mlrun.errors.MLRunConflictError(
+                f"Background task of kind {kind} already running"
```
liranbg, line 60:
```suggestion
                f"Background task of kind '{kind}' already running"
```

server/api/utils/background_tasks/kinds.py
```diff
@@ -0,0 +1,19 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+class BackgroundTaskKinds:
+    db_migrations = "db.migrations"
+    project_deletion = "project.deletion.{0}"
```
liranbg, line 19: related to comment above, change `kind: str` -> `kind: BackgroundTaskKinds`

server/api/utils/background_tasks/project.py
```diff
@@ -0,0 +1,114 @@
...
+                state=mlrun.common.schemas.BackgroundTaskState.failed,
+                error=err_str,
+            )
+        else:
+            server.api.utils.singletons.db.get_db().store_background_task(
+                db_session,
+                name,
+                project=project,
+                state=mlrun.common.schemas.BackgroundTaskState.succeeded,
+            )
```
liranbg, line 114:
```suggestion
     state=suceeded
     error = None
     try:
       ...
     except Exception as exc:
        error = err_to_str(exc)
        state = failed

     finally:

        server.api.utils.singletons.db.get_db().store_background_task(
          db_session,
          name,
          project=project,
          state=state,
          error=error,
      )
```

liranbg (changes requested):
minor comments / suggestions
in addition to that, have a look at `endpoints/runs.py` for route "/projects/{project}/runs/{uid}/abort" which also uses background task mechanism. perhaps it needs an update as well? (using kind)

quaark:
> projects/{project}/runs/{uid}/abort

Abort run uses ProjectBackgroundTask, which doesn't necessarily run on the chief, it cannot have this same implementation. Only InternalBackgroundTasks e.g. migrations, project deletion (which isn't a project BG task, as it can't be related to something it is deleting itself).

server/api/utils/background_tasks/internal.py
quaark: But then we can't use templates to create custom kind strings. Enforcing the kind being in a subset of constants will limit the functionality of the feature.

server/api/utils/background_tasks/kinds.py
quaark: see my reply to above comment

server/api/utils/background_tasks/internal.py
liranbg: perhaps u can const the template, the idea is to provide a way to find usages of background kinds

liranbg (approved):

server/api/utils/background_tasks/internal.py
quaark: You're right it currently isn't anything using this method. Though once I implement delete project again, I see a use case for a scenario where you delete project X. And after it is deleted, you might want to inspect the job that deleted it. That endpoint will use this method (Once the task is no longer active, it is immediately previous).

PR #4866, opened 2024-01-03
[Model Monitoring] Improve the model monitoring application Python API

jond01:
Resolves [ML-5456](https://jira.iguazeng.com/browse/ML-5456).

* Rename the "Application" base classes to "ApplicationBase"
* Inherit from `ABC`, mark the run method as an `abstractmethod` to block direct instantiation by mistake
* Rename the `run_application` method of the `ApplicationBase` class to `do_tracking`
* Fix some docstrings + a typo

### DIFF
--- a/mlrun/model_monitoring/application.py
+++ b/mlrun/model_monitoring/application.py
@@ -148,8 +148,9 @@ class ModelMonitoringApplication(StepToDict, ABC):
         """
         raise NotImplementedError
 
-    @staticmethod
+    @classmethod
     def _resolve_event(
+        cls,
         event: dict[str, Any],
     ) -> Tuple[
         str,
@@ -183,10 +184,10 @@ class ModelMonitoringApplication(StepToDict, ABC):
         end_time = pd.Timestamp(event[mm_constant.ApplicationEvent.END_INFER_TIME])
         return (
             event[mm_constant.ApplicationEvent.APPLICATION_NAME],
-            ModelMonitoringApplication._dict_to_histogram(
+            cls._dict_to_histogram(
                 json.loads(event[mm_constant.ApplicationEvent.CURRENT_STATS])
             ),
-            ModelMonitoringApplication._dict_to_histogram(
+            cls._dict_to_histogram(
                 json.loads(event[mm_constant.ApplicationEvent.FEATURE_STATS])
             ),
             ParquetTarget(
--- a/mlrun/model_monitoring/application.py
+++ b/mlrun/model_monitoring/application.py
@@ -73,7 +73,7 @@ class ModelMonitoringApplicationBase(StepToDict, ABC):
     example for very simple custom application::
         # mlrun: start-code
         class MyApp(ApplicationBase):
-            def run_application(
+            def do_tracking(
                 self,
                 sample_df_stats: pd.DataFrame,
                 feature_stats: pd.DataFrame,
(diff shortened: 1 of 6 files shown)


mlrun/model_monitoring/application.py
```diff
@@ -145,13 +143,14 @@ def run_application(
         :param endpoint_id:             (str) ID of the monitored model endpoint
         :param output_stream_uri:       (str) URI of the output stream for results
 
-        :returns:                       (ModelMonitoringApplicationResult) or
-                                        (list[ModelMonitoringApplicationResult]) of the application results.
+        :returns:                       (ApplicationResult) or
+                                        (list[ApplicationResult]) of the application results.
```
gtopper, line 147: I wonder if it's really necessary to duplicate the type annotations in this way.

mlrun/model_monitoring/application.py
```diff
@@ -82,43 +83,42 @@ def run_application(
...
+        results = self.run(*resolved_event)
         results = results if isinstance(results, list) else [results]
         return results, event
 
     def _lazy_init(self, app_name: str):
         self.context = self._create_context_for_logging(app_name=app_name)
 
-    def run_application(
+    @abstractmethod
+    def run(
```
davesh0812, line 121: I don't see any benefits to this change only can make a confusion to the `run()` of the nuclio function or the `run()` method of the project class.

Also you can see [here](https://jira.iguazeng.com/browse/ML-5295?focusedCommentId=121022&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-121022) amit's comment, and according to it maybe the right name can be 'analyze' or something like that.

mlrun/model_monitoring/application.py
```diff
@@ -31,7 +32,7 @@
 
 
 @dataclasses.dataclass
-class ModelMonitoringApplicationResult:
+class ApplicationResult:
```
davesh0812, line 35:
```suggestion
class MonitoringApplicationResult:
```

mlrun/model_monitoring/application.py
```diff
@@ -64,15 +65,15 @@ def to_dict(self):
         }
 
 
-class ModelMonitoringApplication(StepToDict):
+class ApplicationBase(StepToDict, ABC):
```
davesh0812, line 68:
```suggestion
class MonitoringApplicationBase(StepToDict, ABC):
```

mlrun/model_monitoring/evidently_application.py
```diff
@@ -64,13 +64,14 @@ def _check_evidently_version(*, cur: semver.Version, ref: semver.Version) -> Non
     from evidently.utils.dashboard import TemplateParams
 
 
-class EvidentlyModelMonitoringApplication(ModelMonitoringApplication):
+class EvidentlyApplicationBase(ApplicationBase):
```
davesh0812, line 67:
```suggestion
class EvidentlyMonitoringApplicationBase(ApplicationBase):
```

davesh0812 (changes requested):
Regarding the modified class names, I believe that the term "Monitoring" should be included in the class names to provide context. Currently, these classes represent a monitoring application and its results, and the inclusion of "Monitoring" in their names would better reflect their purpose. :)

mlrun/model_monitoring/application.py
jond01: 3a6bf78

mlrun/model_monitoring/evidently_application.py
jond01: 7eb34b8

mlrun/model_monitoring/application.py
jond01: a34682e

mlrun/model_monitoring/application.py
jond01: Do you mean the docstring duplication of the type hint in the signature?

assaf758 (approved):

PR #4869, opened 2024-01-04
[Project] Delete Project Background Task Mark II

quaark:
## New Project Delete Behaviour
Redoing https://github.com/mlrun/mlrun/pull/4784, as using background task names introduced a regression where projects could only be deleted once (if a new project with the same name was created, we wouldn't be able to delete it)

https://jira.iguazeng.com/browse/IG-22194

### When MLRun is leader (Open Source):
1. Create a background task for deleting the project
2. The background task will delete the project resources and then project itself

### When MLRun is a follower:
Due to the nature of the project deletion flow, we need to wrap the task as:
```
MLRunDeletionWrapperTask(LeaderDeletionJob(MLRunDeletionTask))
```
The flow is as follows:
1. Create MLRunDeletionWrapperTask
2. MLRunDeletionWrapperTask will send a request to the projects leader to delete the project
3. MLRunDeletionWrapperTask will wait for the project to be deleted using LeaderDeletionJob job id
    **During (In leader)**:
    1. Create LeaderDeletionJob
    2. LeaderDeletionJob will send a second delete project request to the follower
    3. LeaderDeletionJob will wait for the project to be deleted using the MLRunDeletionTask task id
        **During (Back here in follower)**:
        1. Create MLRunDeletionTask
        2. MLRunDeletionTask will delete the project resources and then project itself.
        3. Finish MLRunDeletionTask
    4. Finish LeaderDeletionJob

## Changes
- Added v2 `DELETE /projects/<name>` endpoint which behaves in the way stated above (for `check` strategy, no BG task is spawned, the check is performed and returned immediately)
- Added v2 `GET /projects/<name>/deletion-background-task` if we ever want to inspect the BG task that deleted a project.
- HTTPDB calls new endpoint and waits for BG task.
- Kept v1 `DELETE /projects/<name>` endpoint for backwards compatibility
- Added v2 `GET /healthz` endpoint (useful for the project leader to see whether v2 api exists in this mlrun server or not)
- Internal BG Tasks can now have a timeout
- reroute request to chief can now have a configurable api version (set to v2 in our v2 DELETE /projects but left as default for the v1 DELETE /projects endpoint)


server/api/api/endpoints/projects_v2.py
```diff
@@ -0,0 +1,176 @@
...
+        return await chief_client.delete_project(
+            name=name, request=request, api_version="v2"
+        )
+
+    # check if project exists
+    try:
+        get_project_member().get_project(db_session, name, auth_info.session)
+    except mlrun.errors.MLRunNotFoundError:
+        logger.info("Project not found, nothing to delete", project=name)
+        return fastapi.Response(status_code=http.HTTPStatus.NO_CONTENT.value)
```
alonmr, line 80: I wonder if we can do this before re-routing to chief wdyt?

server/api/api/endpoints/projects_v2.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+
+    background_task_kind = background_task_kind_format.format(project_name)
+    try:
+        return server.api.utils.background_tasks.InternalBackgroundTasksHandler().get_active_background_task_by_kind(
+            background_task_kind,
+            raise_on_not_found=True,
+        )
+    except mlrun.errors.MLRunNotFoundError:
+        return server.api.utils.background_tasks.InternalBackgroundTasksHandler().create_background_task(
```
alonmr, line 149: think its better not to nest under try except because it chains the exception.
see `get_or_create_project` for reference

alonmr (commented):
Looks good! minor comment / suggestions

alonmr (approved):

PR #4873, opened 2024-01-04
[CI] Enable log collector in deployer and OS system tests

alonmr:
Add options in the deployer script to:
1. override the log collector image
2. disable log collector (was always disabled, now enabled by default)
3. disableable components will always be part of the helm command - if they were enabled, the script would not specify it and count on the default value. Now it will be explicit.

Add precondition check for get log size endpoint - supported only with log collector.

This will make the open source system tests run with log collector.

### DIFF
--- a/server/api/crud/logs.py
+++ b/server/api/crud/logs.py
@@ -99,23 +99,34 @@ class Logs(
         logger.debug("Getting log size for run", project=project, run_uid=run_uid)
         if (
             mlrun.mlconf.log_collector.mode
+            == mlrun.common.schemas.LogsCollectorMode.best_effort
+        ):
+            try:
+                return await self._get_log_size_from_log_collector(project, run_uid)
+            except Exception as exc:
+                if mlrun.mlconf.log_collector.verbose:
+                    logger.warning(
+                        "Failed to get logs from logs collector, falling back to legacy method",
+                        exc=exc,
+                    )
+                return self._get_log_size_legacy(project, run_uid)
+
+        elif (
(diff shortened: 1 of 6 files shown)


quaark (approved):

server/api/crud/logs.py
```diff
@@ -97,6 +97,14 @@ async def get_log_size(
...
     ):
         logger.debug("Getting log size for run", project=project, run_uid=run_uid)
+        if (
+            mlrun.mlconf.log_collector.mode
+            == mlrun.common.schemas.LogsCollectorMode.legacy
+        ):
+            raise mlrun.errors.MLRunPreconditionFailedError(
+                "Cannot get log size in legacy log collector mode",
+            )
+
```
liranbg, line 107: it will fail log_collector.py `get_logs` function that calls it for legacy mode

liranbg (approved):

server/api/crud/logs.py
```diff
@@ -97,19 +97,36 @@ async def get_log_size(
...
+            mlrun.mlconf.log_collector.mode
+            == mlrun.common.schemas.LogsCollectorMode.best_effort
+        ):
+            try:
+                return await self._get_log_size_from_log_collector(project, run_uid)
+            except Exception as exc:
+                if mlrun.mlconf.log_collector.verbose:
+                    logger.warning(
+                        "Failed to get logs from logs collector, falling back to legacy method",
+                        exc=exc,
```
liranbg, line 116: Err to str

PR #4878, opened 2024-01-07
[Project] Enable adding git to a project after project creation

Yacouby:
review made in :
https://github.com/mlrun/mlrun/pull/4835

resolves :
https://jira.iguazeng.com/browse/ML-5078

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -17,7 +17,6 @@ import glob
 import http
 import importlib.util as imputil
 import json
-import os
 import pathlib
 import shutil
 import tempfile
@@ -2293,8 +2292,7 @@ class MlrunProject(ModelObj):
         :param name:   name for the remote (default is 'origin')
         :param branch: Git branch to use as source
         """
-        if not self.spec.repo and not self.is_git_initialized():
-            raise ValueError("git repo is not valid/initialized")
+        self._ensure_git_repo()
         self.spec.repo.create_remote(name, url=url)
         url = url.replace("https://", "git://")
         if not branch:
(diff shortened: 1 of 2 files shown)


mlrun/projects/project.py
```diff
@@ -2306,6 +2306,26 @@ def create_remote(self, url, name="origin", branch=None):
...
+
+        if not os.path.exists(git_dir_path):
+            logger.warning("Git repository not initialized. initializing now")
+            self.spec.repo = git.Repo.init(context)
+        else:
+            # git already initialized
+            try:
+                self.spec.repo = git.Repo(context)
+            except git.InvalidGitRepositoryError:
+                logger.warning("Invalid Git repository")
```
alonmr, line 2323: I don't see a reason to catch the exceptions here just to log some warning and continue.
You can create the repo without a try clause

alonmr (changes requested):
Looks great! minor comment

mlrun/projects/project.py
```diff
@@ -2306,6 +2306,19 @@ def create_remote(self, url, name="origin", branch=None):
         self.spec._source = self.spec.source or url
         self.spec.origin_url = self.spec.origin_url or url
 
+    def _ensure_git_repo(self):
+        if self.spec.repo:
+            return
+        context = self.context
+        git_dir_path = os.path.join(context, ".git")
```
alonmr, line 2313: you already have path imported so you don't need to import os
```suggestion
        git_dir_path = path.join(context, ".git")
```

alonmr (approved):

PR #4879, opened 2024-01-07
[Databricks] Bypass Intermediate completed state on abortion

alonmr:
https://jira.iguazeng.com/browse/ML-5348

Raise a dedicated exception when run was canceled to signal that the run was aborted and the state should not be changed.


mlrun/runtimes/databricks_job/databricks_wrapper.py
```diff
@@ -199,14 +205,20 @@ def print_status(run: Run):
...
-        run_output = workspace.jobs.get_run_output(get_task(run).run_id)
+
+        task_run_id = get_task(run).run_id
+        run_output = workspace.jobs.get_run_output(task_run_id)
     finally:
         workspace.dbfs.delete(script_path_on_dbfs)
         workspace.dbfs.delete(artifact_json_path)
 
+    run_result_state = run_output.metadata.state.result_state
+    if run_result_state == RunResultState.CANCELED:
```
tomerm-iguazio, line 216: why this condition is needed?

alonmr: The bug was that the run state moved from aborting to completed and then to aborted when it should have moved from aborting straight to aborted.
we need to raise an exception if the task was aborted so that it won't be marked as completed/error.
I check that the task was aborted by checking the result state - if it is canceled - it was aborted

PR #4881, opened 2024-01-07
[Workflow] Ensure "owner" label exists on runner

TomerShor:
Ensure "owner" label exists on the workflow runner.
Owner is taken from the `V3IO_USERNAME` env var or the logged-in username.

Also, moved the owner label enrichment to a common helper function so it can be reused across the codebase.

Resolves https://jira.iguazeng.com/browse/ML-5412

### DIFF
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -14,6 +14,7 @@
 
 import enum
 import functools
+import getpass
 import hashlib
 import inspect
 import itertools
@@ -534,6 +535,11 @@ def match_labels(labels, conditions):
         else:
             match = match and (condition.strip() in labels)
     return match
+
+
+def enrich_owner_label(labels: dict, owner: str = None):
+    if "owner" not in labels:
+        labels["owner"] = owner or os.environ.get("V3IO_USERNAME") or getpass.getuser()
 
 
 def match_times(time_from, time_to, obj, key):
--- a/mlrun/kfpops.py
+++ b/mlrun/kfpops.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import getpass
 import json
 import os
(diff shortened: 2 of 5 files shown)


mlrun/utils/helpers.py
```diff
@@ -536,6 +537,11 @@ def splitter(verb, text):
     return match
 
 
+def enrich_owner_label(labels: dict, owner: str = None):
```
alonmr, line 540: Let's generalize it. We can use Enum for the RunLabels. The helper will enrich the given label something like:
```
def enrich_labels(labels: dict, labels_to_enrich: list = None):
       run_labels = {RunLabels.owner: ....., .....}
       labels_to_enrich = labels_to_enrich or RunLabels.all()
       for label_to_enrich in labels_to_enrich:
             ....
```
Note owner and v3io_user is duplication we want to remove someday

TomerShor: It's a good idea, but enrichment is usually for default values. 
I didn't find any other default labels other than `owner` and `v3io_user`. Do you know of any others?

Also, if we want to remove `v3io_user`, shouldn't it be written somewhere?

alonmr: > Also, if we want to remove v3io_user, shouldn't it be written somewhere?

There is a TODO somewhere.
I'm not aware of more run labels ATM but it shouldn't be an issue

mlrun/runtimes/utils.py
```diff
@@ -31,6 +33,7 @@
 from ..artifacts import TableArtifact
 from ..config import config
 from ..utils import get_in, helpers, logger, verify_field_regex
+from .constants import RunLabels
```
alonmr, line 36: Preferably don't use relative imports because they can cause circular import errors

alonmr (approved):
🔥

mlrun/runtimes/utils.py
TomerShor: Done. fixed the others as well

liranbg (approved):

PR #4884, opened 2024-01-08
[Artifacts] Fix list artifacts legacy tree as tag

alonmr:
Search for artifacts setting tag as tree (legacy).
Fix type hints and docstrings.
Add tree parameter for project SDK.


server/api/api/endpoints/artifacts.py
```diff
@@ -232,6 +232,25 @@ async def list_artifacts(
         format_=format_,
     )
 
+    # in earlier versions, producer_id and tag got confused with each other,
+    # so we combine the results with the given tag as the producer_id
+    if tag:
```
TomerShor, line 237: I think it is better to fallback than to just merge

```suggestion
    if not artifacts and tag:
        # in earlier versions, producer_id and tag got confused with each other,
        # so we search for results with the given tag as the producer_id
```

TomerShor (changes requested):
Minor

TomerShor (approved):

PR #4887, opened 2024-01-08
[Artifacts] Don't overwrite `metadata.key` when storing artifact

TomerShor:
Aligning `metadata.key` with `db_key` breaks backwards compatibility and should still be allowed, as there are different usages for saving an artifact with the same `metadata.key` with different keys in the db.

Reverting part of #4864 and #4876.


alonmr (approved):

liranbg (changes requested):
can you UT that? so we wont discover that on system test but on CI?

PR #4892, opened 2024-01-09
[Project] Take project source into account when running `build_image`

theSaarco:
The `project.build_image()` function did not take into account the project source. If the user is setting a source to the project (for example `git` source) and set `pull_at_runtime = False`, then it is expected that images used would contain the project source in them. This worked for specific functions when building them, but not in the project level.
The expectation is that running this code:
```
    name="project-name", context="./"
)

project.set_source(
    "git://some/repo",
    pull_at_runtime=False
)

project.build_image(image=".some-project-image")
```
Would build the `.some-project-image` image with the source in it, such that running a function will work without having to specify an image:
```
func = project.set_function(handler="package.function", name="func", kind="job")
func.save()
project.run_function("func", params={...})
```
This would work even without having the source code available on the dev site, and will not require pulling the code at execution time.

This fixes [ML-5481](https://jira.iguazeng.com/browse/ML-5481)

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -3048,9 +3048,12 @@ class MlrunProject(ModelObj):
         overwrite_build_params: bool = False,
         requirements_file: str = None,
         extra_args: str = None,
+        target_dir: str = None,
     ) -> typing.Union[BuildStatus, kfp.dsl.ContainerOp]:
         """Builder docker image for the project, based on the project's build config. Parameters allow to override
         the build config.
+        If the project has a source configured and pull_at_runtime is not configured, this source will be cloned to the
+        image built. The `target_dir` parameter allows specifying the target path where the code will be extracted.
 
         :param image: target image name/path. If not specified the project's existing `default_image` name will be
                         used. If not set, the `mlconf.default_project_image_name` value will be used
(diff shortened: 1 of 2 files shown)


mlrun/projects/project.py
```diff
@@ -3115,6 +3115,13 @@ def build_image(
 
             function = mlrun.new_function("mlrun--project--image--builder", kind="job")
 
+            if self.spec.source and not self.spec.load_source_on_run:
+                function.with_source_archive(
+                    source=self.spec.source,
+                    workdir=self.spec.workdir,
```
alonmr, line 3121: workdir is not part of the build job but it is set during runtime (before starting the run)

tests/projects/test_project.py
```diff
@@ -47,6 +48,37 @@ def assets_path():
     return pathlib.Path(__file__).absolute().parent / "assets"
 
 
+@pytest.fixture
+def remote_builder_mock(monkeypatch):
+    class _MockDb:
```
alonmr, line 53: why setting the class inside the fixture? it can be done like our k8s secrets mock

alonmr (commented):
Nice! Minor comments

mlrun/projects/project.py
alonmr: huh you are right :( that's my fault.. we should move it to the build spec

tests/projects/test_project.py
theSaarco: Fixed.

mlrun/projects/project.py
theSaarco: It's also not part of `project.set_source`, and it does make sense. I think for the project use-case, it's ok to have it as a parameter for `build_image`, which is what I did.

alonmr (approved):

PR #4893, opened 2024-01-09
[Artifacts] Remove updatable fields from UID calc

alonmr:
### Bug
Creating a model artifact, adding evaluation values to the model, storing the artifact again expecting it to update to logged one but instead it creates a new model artifact.

### Fix
Remove updatable spec fields from the UID hash calculation. It is expected that if the model metrics / content was changed but the file path is the same, the artifact will be updated instead of creating a new artifact.

### DIFF
--- a/.run/Patch remote.run.xml
+++ b/.run/Patch remote.run.xml
@@ -1,6 +1,7 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Patch remote" type="PythonConfigurationType" factoryName="Python">
     <module name="mlrun" />
+    <option name="ENV_FILES" value="" />
     <option name="INTERPRETER_OPTIONS" value="" />
     <option name="PARENT_ENVS" value="true" />
     <envs>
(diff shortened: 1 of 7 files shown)


.run/Patch remote.run.xml
TomerShor, line 1: What is this for?

TomerShor (approved):
looks good

.run/Patch remote.run.xml
alonmr: its a run configuration for patching mlrun script. I removed the log collector flag. I think that by default it should not patch the log collector

theSaarco (approved):
👍

tests/api/db/test_artifacts.py
```diff
@@ -1262,6 +1262,35 @@ def test_migrate_artifact_v2_tag(self, db: DBInterface, db_session: Session):
...
+
+        # update spec that should not be ignored
+        model_body["spec"]["model_file"] = "some/path"
+        db.store_artifact(db_session, artifact_key, model_body)
+        artifacts = db.list_artifacts(db_session)
+        assert len(artifacts) == 2
+        for model in artifacts:
+            assert model["metadata"]["key"] == artifact_key
+
+        assert artifacts[1]["spec"]["model_file"] == "some/path"
```
TomerShor, line 1292: is the order guaranteed?

alonmr: don't know, fixed it for u

PR #4898, opened 2024-01-10
[Model Monitoring] Change default last analyzed to be the latest between first request time or last update time

davesh0812:
[ML-5497](https://jira.iguazeng.com/browse/ML-5497)

### DIFF
--- a/mlrun/model_monitoring/controller.py
+++ b/mlrun/model_monitoring/controller.py
@@ -57,9 +57,9 @@ class _BatchWindow:
         self._first_request = first_request
         self._kv_storage = get_v3io_client(endpoint=mlrun.mlconf.v3io_api).kv
         self._v3io_container = self.V3IO_CONTAINER_FORMAT.format(project=project)
-        self._start = self._get_last_analyzed()
         self._stop = last_updated
         self._step = timedelta_seconds
+        self._start = self._get_last_analyzed()
 
     def _get_last_analyzed(self) -> Optional[int]:
         try:
@@ -80,7 +80,13 @@ class _BatchWindow:
             )
 
             return max(
-                self._first_request, self._stop - cast(int, datetime.timedelta(days=1))
+                self._first_request,
+                int(
+                    (
+                        datetime.datetime.fromtimestamp(self._stop)
+                        - datetime.timedelta(days=1)
+                    ).timestamp()
+                ),
             )
 
         last_analyzed = data.output.item[mm_constants.SchedulingKeys.LAST_ANALYZED]
--- a/mlrun/model_monitoring/controller.py
+++ b/mlrun/model_monitoring/controller.py
@@ -69,7 +69,7 @@ class _BatchWindow:
                 key=self._application,
             )
         except HttpResponseError as err:
-            logger.warn(
+            logger.info(
                 "Failed to get the last analyzed time for this endpoint and application, "
                 "as this is probably the first time this application is running. ",
                 "Using the latest between first request time or last update time minus one day instead.",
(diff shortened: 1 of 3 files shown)


mlrun/model_monitoring/controller.py
```diff
@@ -71,14 +71,23 @@ def _get_last_analyzed(self) -> Optional[int]:
...
-            return self._first_request
+
+            return max(
+                self._first_request,
+                int(
+                    (
+                        datetime.datetime.fromtimestamp(self._stop)
+                        - datetime.timedelta(days=1)
+                    ).timestamp()
+                ),
```
jond01, line 89: To avoid timezones mess from `fromtimestamp`:
```suggestion
                self._stop - int(datetime.timedelta(days=1).total_seconds()),
```
https://docs.astral.sh/ruff/rules/call-datetime-fromtimestamp/

mlrun/model_monitoring/controller.py
```diff
@@ -71,14 +71,23 @@ def _get_last_analyzed(self) -> Optional[int]:
         except HttpResponseError as err:
             logger.warn(
                 "Failed to get the last analyzed time for this endpoint and application, "
-                "as this is probably the first time this application is running. "
-                "Using the first request time instead.",
+                "as this is probably the first time this application is running. ",
+                "Using the latest between first request time or last update time instead.",
```
jond01, line 75: Log the time used as "last_analyzed" and where it comes from (stop - 1d / first_request).

mlrun/runtimes/serving.py
```diff
@@ -201,7 +201,7 @@ def __init__(
         self.function_refs = function_refs or []
         self.graph_initializer = graph_initializer
         self.error_stream = error_stream
-        self.track_models = track_models
+        self.track_models = track_models or False
```
jond01, line 204: ?

mlrun/model_monitoring/controller.py
```diff
@@ -71,14 +71,19 @@ def _get_last_analyzed(self) -> Optional[int]:
...
                 application=self._application,
                 first_request=self._first_request,
+                last_update=self._stop,
                 error=err,
             )
-            return self._first_request
+
+            return max(
+                self._first_request,
+                self._stop - int(datetime.timedelta(days=1).total_seconds()),
```
jond01, line 85: I'm ok with it for now, but note that this `days=1` constant comes with a price:
This means that monitoring time windows longer than one day will not work.
For example, consider an app that expects one-week windows. This logic limits the difference between the "start" (last_analyzed - L85) to "stop" (last_updated) to one day at most. The app will never run, as the first one-week window will not be completed.

FYI @assaf758.

jond01 (approved):

mlrun/model_monitoring/controller.py
davesh0812: Good point when we will have a policies, we will need to change it according to it.

assaf758: I think we can overcome this issue by updating the the KV last_analyzed with value we are returning.

mlrun/model_monitoring/controller.py
```diff
@@ -71,14 +71,20 @@ def _get_last_analyzed(self) -> Optional[int]:
         except HttpResponseError as err:
             logger.warn(
                 "Failed to get the last analyzed time for this endpoint and application, "
-                "as this is probably the first time this application is running. "
-                "Using the first request time instead.",
+                "as this is probably the first time this application is running. ",
```
assaf758, line 74: Let's check and exception (or introduce exception types to v3io-py)  and log an error only if it's not 404/not found.
On the other hand, send info-sev log when we init the last_analyzed (see other comment)

mlrun/model_monitoring/controller.py
```diff
@@ -69,16 +69,25 @@ def _get_last_analyzed(self) -> Optional[int]:
...
                 application=self._application,
                 first_request=self._first_request,
+                last_update=self._stop,
                 error=err,
             )
-            return self._first_request
+
+            # TODO : Change the timedelta according to the policy.
+            first_period_in_seconds = max(
+                3600, self._step
```
jond01, line 85: This is 1hr, change to:
```suggestion
                86400, self._step
```
Or even better:
```suggestion
                datetime.timedelta(days=1).total_seconds(), self._step
```

davesh0812: @jond01 you review it to fast.. thanks anyway :)

jond01 (approved):

assaf758 (approved):

PR #4901, opened 2024-01-10
[Datastore] Rename access key to access_key_id in s3.

tomerm-iguazio:
[ML-5504](https://jira.iguazeng.com/browse/ML-5504)


mlrun/datastore/s3.py
```diff
@@ -133,19 +133,19 @@ def get_storage_options(self):
         force_non_anonymous = self._get_secret_or_env("S3_NON_ANONYMOUS")
         profile = self._get_secret_or_env("AWS_PROFILE")
         endpoint_url = self._get_secret_or_env("S3_ENDPOINT_URL")
-        key = self._get_secret_or_env("AWS_ACCESS_KEY_ID")
+        key_id = self._get_secret_or_env("AWS_ACCESS_KEY_ID")
```
assaf758, line 136: so let's call it access_key_id here as well..

assaf758 (approved):

PR #4906, opened 2024-01-10
[Functions] Add more labels to the image builder pods to facilitate resource usage tracking

laurybueno:
[ML-3728](https://jira.iguazeng.com/browse/ML-3728)

Add more MLRun-related labels to pods that build function and project images in order to facilitate monitoring.

The added labels are:
```
mlrun/name
mlrun/function
mlrun/tag
```  

It's worth noting that `mlrun/class` and `mlrun/project` are already present on builder pods.

### DIFF
--- a/server/api/utils/singletons/k8s.py
+++ b/server/api/utils/singletons/k8s.py
@@ -591,7 +591,7 @@ class BasePod:
         project=None,
         default_pod_spec_attributes=None,
         resources=None,
-        labels={},
+        labels=None,
     ):
         self.namespace = namespace
         self.name = ""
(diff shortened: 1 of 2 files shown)


server/api/utils/singletons/k8s.py
```diff
@@ -591,6 +591,7 @@ def __init__(
         project=None,
         default_pod_spec_attributes=None,
         resources=None,
+        labels={},
```
liranbg, line 594: value is mutable, change to `None` and set within function if its value is None

liranbg (changes requested):

server/api/utils/singletons/k8s.py
laurybueno: Good thinking. I've improved it for us. Please see if it makes more sense now.

PR #4917, opened 2024-01-11
[Artifacts] Retain `db_key` when logging artifact

TomerShor:
When logging an artifact with an existing `spec.db_key`, do not overwrite it with the resolved key, but retain it.

This was an issue mainly when loading projects with existing artifacts that had a custom db_key.

Resolves https://jira.iguazeng.com/browse/ML-4608


tests/projects/test_local_pipeline.py
```diff
@@ -71,7 +71,8 @@ def test_import_artifacts(self, rundb_mock):
             assert artifact.metadata.key in expected_body_map
             assert expected_body_map[artifact.metadata.key] == artifact._get_file_body()
 
-            some_artifact = project.get_artifact(artifact.metadata.key)
+            db_key = artifact.db_key or artifact.metadata.key
+            some_artifact = project.get_artifact(db_key)
```
alonmr, line 75: why was this needed? I would expect that
```
project.get_artifact(artifact.metadata.key)
``` 
will still work

tests/system/projects/test_project.py
```diff
@@ -1119,6 +1119,69 @@ def test_project_build_config_export_import(self):
...
+        project = mlrun.get_or_create_project(project_1_name, context=context)
+
+        # create artifact with an explicit db_key
+        artifact_key = "artifact_key"
+        artifact_db_key = "artifact_db_key"
+        project.log_artifact(artifact_key, db_key=artifact_db_key, body="test")
+
+        # validate that the artifact is in the db
+        artifacts = project.list_artifacts(name=artifact_db_key)
+        print("TOMER - artifacts 1", artifacts)
```
alonmr, line 1140: print

tests/system/projects/test_project.py
```diff
@@ -1119,6 +1119,69 @@ def test_project_build_config_export_import(self):
...
+            key=artifact_new_key, artifact=Artifact.from_dict(artifacts[0])
+        )
+        project.save()
+
+        # create a project from the same spec
+        project2 = mlrun.load_project(context=context, name=project_2_name)
+
+        # validate that the artifact was saved with the db_key
+        artifacts = project2.list_artifacts(name=artifact_db_key)
+        print("TOMER - artifacts 2", artifacts)
```
alonmr, line 1156: print

alonmr (commented):
Nice ! minor question and comments

mlrun/artifacts/manager.py
```diff
@@ -200,8 +200,8 @@ def log_artifact(
                 # and receive back all the runs that are associated with his search result.
                 db_key = producer.name + "_" + key
             else:
-                db_key = key
-        item.db_key = db_key if db_key else ""
+                db_key = item.db_key or key
```
liranbg, line 203: mind explaining the flows and why using key vs db_key would make sense in that regard?

tests/projects/test_local_pipeline.py
TomerShor: The way we get/list artifacts with `key` is by filtering the db `key` column value according to the given key.
If an artifact is saved with a custom `db_key`, that will be the value in the `key` db column.
To query an artifact according to the value in `metadata.key` would mean to decode the blob and then filter - which is exactly what we tried to avoid in the refactor.

In this PR, the change is to use the value in `spec.db_key` as the key when you import/register an artifact, if it already exists.
the artifact used in the test has a `spec.db_key = x` value.

mlrun/artifacts/manager.py
TomerShor: The `db_key` variable determines what `key` the artifact is saved with in the DB.

When you log an artifact without specifying a `db_key` explicitly, `item.db_key` will be empty - so we want to take the `key` and use that as the actual key.
If an explicit `db_key` was set, we want to use that.

This change mainly affects imported artifacts that have an explicit `db_key` value already set (either the same as `key` or a different value).
If it is set - we don't want to override it and we want to keep the same behavior.

PR #4918, opened 2024-01-13
[Artifacts] Dataset metadata generation must be mindful of differences among header, preview and schema

laurybueno:
[ML-5395](https://jira.iguazeng.com/browse/ML-5395)

To allow for an `index` column on the dataset Preview tab on the UI, we must preserve it in the dataframe as an explicit column until we parse the table schema.

This PR is a follow-up of https://github.com/mlrun/mlrun/pull/4778.


mlrun/artifacts/dataset.py
```diff
@@ -283,14 +283,16 @@ def update_preview_fields_from_df(
         if artifact.spec.length > preview_rows_length and not ignore_preview_limits:
             preview_df = df.head(preview_rows_length)
 
-        # reset index while dropping existing index
-        # that way it wont create another index if one already there
-        preview_df = preview_df.reset_index(drop=True)
```
liranbg, line 288: Ci fails because there is another instance of such flow (code dup) @ lines 516

liranbg: It actually needs to be implemented as well. the obsoletion is pushed to mlrun 1.7 so that is needed as well.

laurybueno: I have just updated it for us. Please take a look and see if it looks reasonable to you. Thanks!

PR #4919, opened 2024-01-14
[Projects] Fix `DELETE /api/v2/projects` not returning 403 on permission denied

quaark:
https://jira.iguazeng.com/browse/ML-5520
The permissions check failure was happening inside the task. Added the check before the task so we can raise and return 403.


server/api/api/endpoints/projects_v2.py
```diff
@@ -63,6 +63,15 @@ async def delete_project(
         logger.info("Project not found, nothing to delete", project=name)
         return fastapi.Response(status_code=http.HTTPStatus.NO_CONTENT.value)
 
+    # usually the CRUD for delete project will check permissions, however, since we are running the crud in a background
+    # task, we need to check permissions here. skip permission check if the request is from the leader.
+    if not server.api.utils.helpers.is_request_from_leader(auth_info.projects_role):
```
liranbg, line 68: i was thinking about the boilerplate, I think worth shoving the `is_request_from_leader` inside the `query_project_permissions`

quaark: That's not safe in my opinion. `query_project_permissions` isn't only called on projects CRUD, but on other sub resources as well, where this `if` is *not* boilerplated. If it were everywhere, I'd agree, but I wouldn't add this to all permission checks ever.

PR #4923, opened 2024-01-14
[Datastore] Filter out 'none' values when passing storage options for Azure

alxtkr77:
https://jira.iguazeng.com/browse/ML-5518
https://jira.iguazeng.com/browse/ML-5521

Hadoop configurations for Azure Spark were based on the presumption that underlying storage options are present only if they are non-empty. However, this assumption proved to be inaccurate. The proposed Pull Request (PR) addresses this issue by excluding empty storage options, thereby ensuring the proper functioning of Spark.

### DIFF
--- a/mlrun/datastore/azure_blob.py
+++ b/mlrun/datastore/azure_blob.py
@@ -57,7 +57,7 @@ class AzureBlobStore(DataStore):
         return self._filesystem
 
     def get_storage_options(self):
-        return dict(
+        res = dict(
             account_name=self._get_secret_or_env("account_name")
             or self._get_secret_or_env("AZURE_STORAGE_ACCOUNT_NAME"),
             account_key=self._get_secret_or_env("account_key")
@@ -74,6 +74,9 @@ class AzureBlobStore(DataStore):
             or self._get_secret_or_env("AZURE_STORAGE_SAS_TOKEN"),
             credential=self._get_secret_or_env("credential"),
         )
+        # Filter out None values
+        res = {k: v for k, v in res.items() if v}
+        return res
 
     def _convert_key_to_remote_path(self, key):
         key = key.strip("/")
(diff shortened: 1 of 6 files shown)


mlrun/datastore/azure_blob.py
```diff
@@ -74,6 +74,9 @@ def get_storage_options(self):
             or self._get_secret_or_env("AZURE_STORAGE_SAS_TOKEN"),
             credential=self._get_secret_or_env("credential"),
         )
+        # Filter out None values
+        res = {k: v for k, v in res.items() if v}
```
gtopper, line 78: This will also filter out empty strings, empty data structures, `False`, and 0 numericals. It looks like only empty strings apply here. Did you mean to filter these out?

Either way, it seems like either the condition or the comment should be updated.

alxtkr77: Oh, this python...
Anyway - looks like there is a broader issue with the get_storage_options()

mlrun/datastore/base.py
```diff
@@ -107,7 +114,7 @@ def _get_secret_or_env(self, key, default=None, prefix=None):
 
     def get_storage_options(self):
         """get fsspec storage options"""
-        return None
+        return self._sanitize_storage_options(None)
```
gtopper, line 117: I think you can just return `{}` here.

mlrun/datastore/google_cloud_storage.py
```diff
@@ -59,12 +59,12 @@ def get_storage_options(self):
...
                 # If it's not json, handle it as a filename
                 token = credentials
-            return dict(token=token)
+                return self._sanitize_storage_options(dict(token=token))
         else:
             logger.info(
                 "No GCS credentials available - auth will rely on auto-discovery of credentials"
             )
-            return None
+            return self._sanitize_storage_options(None)
```
gtopper, line 67: Here as well, simpler to just return `{}`.

mlrun/datastore/base.py
alxtkr77: I want to enforce the pattern - always return via _sanitized_(...). If tomorrow we decide return None instead of {}  for example

mlrun/datastore/google_cloud_storage.py
alxtkr77: same is here

mlrun/datastore/base.py
gtopper: Doesn't really seem worthwhile, but I guess it's not very important either.

gtopper (approved):
LGTM otherwise

PR #4924, opened 2024-01-14
[Datastore] Refactor internal get_filesystem() method

alxtkr77:
https://jira.iguazeng.com/browse/ML-5510
Refactoring an internal get_filesystem() method

### DIFF
--- a/mlrun/datastore/azure_blob.py
+++ b/mlrun/datastore/azure_blob.py
@@ -20,7 +20,6 @@ from azure.storage.blob._shared.base_client import parse_connection_str
 from fsspec.registry import get_filesystem_class
 
 import mlrun.errors
-from mlrun.errors import err_to_str
 
 from .base import DataStore, FileStats, makeDatastoreSchemaSanitizer
 
@@ -34,16 +33,15 @@ class AzureBlobStore(DataStore):
     def __init__(self, parent, schema, name, endpoint="", secrets: dict = None):
         super().__init__(parent, name, schema, endpoint, secrets=secrets)
 
-    def get_filesystem(self):
+    @property
+    def filesystem(self):
         """return fsspec file system object, if supported"""
         if self._filesystem:
             return self._filesystem
         try:
             import adlfs  # noqa
         except ImportError as exc:
-            raise ImportError(
-                f"Azure adlfs not installed, run pip install adlfs, {err_to_str(exc)}"
-            )
+            raise ImportError("Azure adlfs not installed") from exc
         # in order to support az and wasbs kinds.
         filesystem_class = get_filesystem_class(protocol=self.kind)
         self._filesystem = makeDatastoreSchemaSanitizer(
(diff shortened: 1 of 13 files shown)


mlrun/datastore/v3io.py
```diff
@@ -73,18 +73,16 @@ def url(self):
...
             import v3iofs  # noqa
         except ImportError as exc:
-            if not silent:
-                raise ImportError(
-                    "v3iofs or storey not installed, run pip install storey"
-                ) from exc
-            return None
+            raise ImportError(
+                "v3iofs or storey not installed, run pip install storey"
+            ) from exc
```
gtopper, line 85: `v3iofs` is a base requirement of mlrun. What does this have to do with storey?

mlrun/datastore/s3.py
```diff
@@ -109,18 +109,14 @@ def get_spark_options(self):
...
             return self._filesystem
         try:
             import s3fs  # noqa
         except ImportError as exc:
-            if not silent:
-                raise ImportError(
-                    "AWS s3fs not installed, run pip install s3fs"
-                ) from exc
-            return None
+            raise ImportError("AWS s3fs not installed, run pip install s3fs") from exc
```
gtopper, line 119:
```suggestion
            raise ImportError("AWS s3fs not installed") from exc
```
The user should _not_ install the latest s3fs. They should install `mlrun[complete]` or  `mlrun[s3]` with a version that is compatible with the mlrun service on their env.

mlrun/datastore/azure_blob.py
```diff
@@ -86,12 +83,12 @@ def _convert_key_to_remote_path(self, key):
 
     def upload(self, key, src_path):
         remote_path = self._convert_key_to_remote_path(key)
-        self._filesystem.put_file(src_path, remote_path, overwrite=True)
+        self.get_filesystem().put_file(src_path, remote_path, overwrite=True)
```
gtopper, line 86: I think if we're going to replace it everywhere, it makes sense to make it a `@property`.

mlrun/datastore/azure_blob.py
```diff
@@ -33,20 +33,17 @@ class AzureBlobStore(DataStore):
...
         try:
             import adlfs  # noqa
         except ImportError as exc:
-            if not silent:
-                raise ImportError(
-                    f"Azure adlfs not installed, run pip install adlfs, {err_to_str(exc)}"
-                )
-            return None
+            raise ImportError(
+                f"Azure adlfs not installed, run pip install adlfs, {err_to_str(exc)}"
```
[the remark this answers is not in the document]
gtopper: I realize it wasn't introduced in this PR, though.

mlrun/datastore/s3.py
gtopper: I realize it wasn't introduced in this PR, though.

mlrun/datastore/v3io.py
gtopper: I realize it wasn't introduced in this PR, though.

mlrun/datastore/azure_blob.py
alxtkr77: get_filesystem() is a public method in the DataStore class and is overridden in subclasses as part of the public interface. Would it be appropriate to convert it into a @property instead?

gtopper (approved):
Looks awesome now!

mlrun/datastore/azure_blob.py
gtopper: I don't think it's really meant to be user facing, but if in doubt, consider adding this to the base class:
```python
def get_filesystem(self):
    return self.filesystem
```

assaf758: Let's add it with deprecation warning

mlrun/datastore/base.py
```diff
@@ -91,6 +92,16 @@ def uri_to_kfp(endpoint, subpath):
     def uri_to_ipython(endpoint, subpath):
         return ""
 
+    # TODO: remove in 1.8.0
+    @deprecated(
+        version="1.8.0",
+        reason="'get_filesystem()' will be removed from this file in 1.8.0, use "
```
assaf758, line 98: Please remove "from this file"

PR #4930, opened 2024-01-15
[Core] Fix deprecated numpy type

alxtkr77:
https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

### DIFF
--- a/mlrun/data_types/to_pandas.py
+++ b/mlrun/data_types/to_pandas.py
@@ -178,7 +178,7 @@ def toPandas(spark_df):
         if isinstance(field.dataType, IntegralType) and pandas_col.isnull().any():
             dtype[fieldIdx] = np.float64
         if isinstance(field.dataType, BooleanType) and pandas_col.isnull().any():
-            dtype[fieldIdx] = np.object
+            dtype[fieldIdx] = np.object_
 
     df = pd.DataFrame()
     for index, t in enumerate(dtype):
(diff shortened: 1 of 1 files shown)


mlrun/data_types/to_pandas.py
```diff
@@ -178,7 +178,7 @@ def toPandas(spark_df):
         if isinstance(field.dataType, IntegralType) and pandas_col.isnull().any():
             dtype[fieldIdx] = np.float64
         if isinstance(field.dataType, BooleanType) and pandas_col.isnull().any():
-            dtype[fieldIdx] = np.object
+            dtype[fieldIdx] = np.object_
```
gtopper, line 181: The recommendation is to replace this deprecated type with a plain `object`, not with the internal type.

```suggestion
            dtype[fieldIdx] = object
```

gtopper (changes requested):

mlrun/data_types/to_pandas.py
alxtkr77: Same goes for  np.float64. Lets be consistent

gtopper: Please see https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated:
> For float and complex you can use float64 and complex128 if you wish to be more explicit about the precision.

gtopper: But feel free to change it to `float` if you think that makes more sense.

alxtkr77: np.float64 is in may places - I don't want to change it. So I don't agree with you, but I don't feel it is important enough. Changing it as per your suggestion

gtopper (approved):

PR #4936, opened 2024-01-16
[Tests] Fix `AttributeError` error due to typo

gtopper:
And add a regression test.

[ML-5537](https://jira.iguazeng.com/browse/ML-5537)


assaf758 (commented):
The fix is correct of course, but seeing the test it makes it apperant that limiting the max_worker will work only if the add_nuclio_trigger() is after the function.spec.* setting

gtopper:
Right @assaf758. This method only ever gets called internally, long after the user defined the spec.

assaf758 (approved):

PR #4938, opened 2024-01-16
[Execution] Fix non logging worker lables

alonmr:
The host label is needed to resolve if the worker is the logging worker or not.
Do not set labels from non logging workers
https://jira.iguazeng.com/browse/ML-5536

### DIFF
--- a/mlrun/execution.py
+++ b/mlrun/execution.py
@@ -977,7 +977,7 @@ class MLClientCtx(object):
         """
         # If it's a OpenMPI job, get the global rank and compare to the logging rank (worker) set in MLRun's
         # configuration:
-        if self._labels.get("kind", "job") == "mpijob":
+        if self._labels.get("kind", "job") == "mpijob" and self._labels.get("host"):
             # The host (pod name) of each worker is created by k8s, and by default it uses the rank number as the id in
             # the following template: ...-worker-<rank>
             rank = int(self._labels["host"].rsplit("-", 1)[1])
--- a/mlrun/execution.py
+++ b/mlrun/execution.py
@@ -201,10 +201,7 @@ class MLClientCtx(object):
     @property
     def labels(self):
         """Dictionary with labels (read-only)"""
-        labels = deepcopy(self._labels)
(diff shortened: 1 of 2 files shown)


mlrun/execution.py
```diff
@@ -201,7 +201,10 @@ def out_path(self):
     @property
     def labels(self):
         """Dictionary with labels (read-only)"""
-        return deepcopy(self._labels)
+        labels = deepcopy(self._labels)
+        if not self.is_logging_worker():
+            labels.pop("host", None)
+        return labels
```
liranbg, line 207: please explain in a comment why do you pop such labels
and, in addition, maybe add a declaritve list of labels that we dont want to be used/read from non-logging worker
another idea is to have `labels`/`non_logging_worker_labels`

liranbg (changes requested):

mlrun/execution.py
```diff
@@ -1090,6 +1091,15 @@ def _write_tmpfile(self, json=None):
...
                 fp.close()
 
+    def _cleanse_labels(self):
+        """Remove labels that should not be stored to the corresponding run"""
+        labels = deepcopy(self._labels)
+        # Remove host (pod name) label if it's not a logging worker:
+        if not self.is_logging_worker():
+            labels.pop("host", None)
+
+        return labels
```
liranbg, line 1101: the cleansing shoudl either take an argument of labels to cleanse, or the function name should be changed to `_resolve_cleansed_labels` or you can `_resolve_labels_for_update()` without feeding it a labels argument

the reason im saying that is because when you want to `_x_something` where x is the action, you would expect to feed the function with an argument as oppose to where x is "resolve" or "get" where you want it to return you something / compile something for you

liranbg (changes requested):
minor
also, add a unit test to ensure upon update, no "host" (unless logging worker) in labels

liranbg (changes requested):
perhaps we need to update labels from the logging worker at all and drop that field for other runs?

liranbg (approved):

PR #4939, opened 2024-01-16
[Datastore] Remove run ID from `StreamTarget` path

gtopper:
[ML-5484](https://jira.iguazeng.com/browse/ML-5484)

### DIFF
--- a/tests/datastore/test_targets.py
+++ b/tests/datastore/test_targets.py
@@ -27,7 +27,8 @@ class MockGraph:
         self.kwargs = kwargs
 
 
-def test_stream_target():
+# ML-5484
+def test_stream_target_path_is_without_run_id():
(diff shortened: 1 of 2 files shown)


tests/datastore/test_targets.py
```diff
@@ -0,0 +1,40 @@
...
+    def __init__(self):
+        self.args = None
+        self.kwargs = None
+
+    def add_step(self, *args, **kwargs):
+        self.args = args
+        self.kwargs = kwargs
+
+
+def test_stream_target():
```
assaf758, line 30: Please rename the test (or/and add comment) to explain that what is actually being tested (that existence of run_id attribute does not effect the stream_path).

assaf758 (approved):

PR #4943, opened 2024-01-16
[Run] Save enriched function to db when submitting job schedule with custom spec

TomerShor:
When submitting a scheduled batch run and overriding the function's configuration, the task still keeps a reference for the original function URI, with the default values.
So although the actual run includes the custom spec, this affects editing a schedule (in the UI) by not populating the config fields with the correct custom values.

To overcome this, we create a new versioned function in the DB with the enriched configuration, and set its URI on the scheduled task, same as we do with hub functions.
Now, the task will reference the enriched function instead of the original one.

Fixes https://jira.iguazeng.com/browse/ML-5111


alonmr (approved):
🔥

PR #4945, opened 2024-01-16
[Databricks] Fix abortion state transitions

alonmr:
https://jira.iguazeng.com/browse/ML-5517
Some changes to make sure that databricks runs do not go from aborting to error before moving to aborted

### DIFF
--- a/mlrun/runtimes/constants.py
+++ b/mlrun/runtimes/constants.py
@@ -166,6 +166,17 @@ class RunStates(object):
         ]
 
     @staticmethod
+    def abortion_states():
+        return [
+            RunStates.aborted,
+            RunStates.aborting,
+        ]
+
+    @staticmethod
+    def error_and_abortion_states():
+        return list(set(RunStates.error_states()) | set(RunStates.abortion_states()))
+
+    @staticmethod
     def non_terminal_states():
         return list(set(RunStates.all()) - set(RunStates.terminal_states()))
 
--- a/mlrun/runtimes/constants.py
+++ b/mlrun/runtimes/constants.py
@@ -162,6 +162,7 @@ class RunStates(object):
     def error_states():
         return [
             RunStates.error,
+            RunStates.aborting,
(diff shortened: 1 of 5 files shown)


mlrun/runtimes/constants.py
```diff
@@ -162,6 +162,7 @@ def terminal_states():
     def error_states():
         return [
             RunStates.error,
+            RunStates.aborting,
```
liranbg, line 165: this is not true, aborting is a transitional state, not error state

mlrun/runtimes/local.py
```diff
@@ -487,6 +488,11 @@ def exec_from_params(handler, runobj: RunObject, context: MLClientCtx, cwd=None)
             else:
                 val = handler(**kwargs)
             context.set_state("completed", commit=False)
+        except mlrun.errors.MLRunTaskCancelledError as exc:
+            logger.warning("Run was aborted", err=err_to_str(exc))
+            # Run was aborted, the state run state is updated by the abort job, no need to commit again
+            context.set_state(mlrun.runtimes.constants.RunStates.aborted, commit=False)
+            commit = False
```
liranbg, line 495: looking down, logger returns to old level - dont you want to do it as well? perhaps we need to conform both behavior

liranbg (changes requested):

mlrun/runtimes/local.py
alonmr: the set logger level is duplicated with one at the end of the function. I have merged them with try, finally clause

mlrun/runtimes/constants.py
alonmr: I think that aborting should be considered an error state. Think that if you want to check if a run is in error state you would also want to know if its aborting so its either that or do
```
run.state in (error_states + [aborting])
```
everywhere

liranbg (approved):

PR #4946, opened 2024-01-16
[Model Monitoring] Fix bumping last_request without controller runs

jond01:
Fixes [ML-5542](https://jira.iguazeng.com/browse/ML-5542).
Assuming controller runs in 1.6.0 is problematic, as we still support the legacy model monitoring without the newer application framework enabled.
If no controller runs are found - don't bump the last_request field of the model endpoint.

### DIFF
--- a/mlrun/model_monitoring/helpers.py
+++ b/mlrun/model_monitoring/helpers.py
@@ -36,6 +36,10 @@ class _BatchDict(typing.TypedDict):
     days: int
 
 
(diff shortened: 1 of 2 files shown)


assaf758 (approved):
Looks good.
And indeed we have some holes in 1.6, and as we discussed we probably need to revise our handling of batch-infer for non-serving MEP. But for now I think it's ok.

PR #4948, opened 2024-01-17
[Model Monitoring] Update deprecated `get_offline_features` calls

jond01:
Fixes [ML-5488](https://jira.iguazeng.com/browse/ML-5488).
Update only model monitoring usages to avoid noisy logs.

### DIFF
--- a/mlrun/model_monitoring/api.py
+++ b/mlrun/model_monitoring/api.py
@@ -612,7 +612,7 @@ def read_dataset_as_dataframe(
         if label_columns is None:
             label_columns = dataset.status.label_column
         # Get the features and parse to DataFrame:
-        dataset = mlrun.feature_store.get_offline_features(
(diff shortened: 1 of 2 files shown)


mlrun/model_monitoring/api.py
```diff
@@ -612,7 +612,7 @@ def read_dataset_as_dataframe(
         if label_columns is None:
             label_columns = dataset.status.label_column
         # Get the features and parse to DataFrame:
-        dataset = mlrun.feature_store.get_offline_features(
+        dataset = dataset.get_offline_features(
             dataset.uri, drop_columns=drop_columns
```
gtopper, line 616:
```suggestion
            drop_columns=drop_columns
```

jond01: Missed that. Thanks!

gtopper (approved):
LGTM

PR #4951, opened 2024-01-17
[Linting] Add pre-commit

jond01:
Resolves [ML-5290](https://jira.iguazeng.com/browse/ML-5290).

https://pre-commit.com/

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-#
+
 name: CI
 
 on:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -49,6 +49,8 @@ jobs:
         python -m pip install --upgrade pip~=23.2.1
         pip install pre-commit==3.6.0
     - name: Lint
+      env:
+        PRE_COMMIT_COLOR: always
       run: make lint
 
     # perform copyright check here, to avoid spinning up another runner needlessly
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,11 +36,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        # 3.9 is the current >= 1.3.0 python version
         python-version: ["3.9"]
     steps:
     - uses: actions/checkout@v3
-    - name: Set up python ${{ matrix.python-version }}
+    - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
@@ -48,7 +47,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip~=23.2.1
-        pip install -r dev-requirements.txt
+        pip install pre-commit==3.6.0
     - name: Lint
       run: make lint
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,4 @@
-# Copyright 2023 Iguazio
+# Copyright 2024 Iguazio
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,17 @@
+# Copyright 2023 Iguazio
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 repos:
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -595,25 +595,9 @@ html-docs-dockerized: build-test ## Build html docs dockerized
...
 .PHONY: lint
-lint: fmt-check lint-imports ## Run lint on the code
-
-.PHONY: fmt-check
-fmt-check: ## Check the code (using ruff)
-	@echo "Running ruff checks..."
-	python -m ruff check --exit-non-zero-on-fix
-	python -m ruff format --check
+lint:
+	pre-commit run --all-files
```
gtopper, line 600: Basically what this does is equivalent to lint then fmt. If the format is wrong, a subsequent run will pass because the first run already formatted the code.

That's probably fine, but I thought it useful to point out the change in behaviour.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,30 @@
+# Copyright 2023 Iguazio
```
gtopper, line 1:
```suggestion
# Copyright 2024 Iguazio
```

.github/workflows/ci.yaml
```diff
@@ -36,19 +36,18 @@ jobs:
...
+    - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
         cache: pip
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip~=23.2.1
-        pip install -r dev-requirements.txt
+        pip install pre-commit==3.6.0
```
gtopper, line 50: Shouldn't this match the requirement in `dev-requirements.txt`?
Either `pre-commit~=3.6.0` or `pre-commit~=3.6` would probably be in order.

gtopper: I wonder whether the linter really doesn't need the requirements installed to properly validate (e.g. to validate correct number of arguments is passed to a function imported from a module in required library).

pyproject.toml
```diff
@@ -5,10 +5,6 @@ select = [
     "E", # pycodestyle
     "I", # isort
 ]
-extend-exclude = [
-    "server/api/proto",
-    "playground",
-]
```
gtopper, line 11: Why is this deleted?

.pre-commit-config.yaml
jond01: This is the suggested template:
https://github.com/mlrun/mlrun/blob/ba399e9f487b769324d9f9e98a8a047ea9c39a74/.github/workflows/ci.yaml#L68

And I think that it should be changed for all the files 🤔
I'll try updating it here, but it's a wider issue

.github/workflows/ci.yaml
jond01: > Shouldn't this match the requirement in `dev-requirements.txt`? Either `pre-commit~=3.6.0` or `pre-commit~=3.6` would probably be in order.

I preferred stability in the CI and therefore specified `==`. I can change it to `~=`, but then it's less predictable. Let me know what you think.



> I wonder whether the linter really doesn't need the requirements installed to properly validate (e.g. to validate correct number of arguments is passed to a function imported from a module in required library).

"the linter" - what do you mean by that? pre-commit? Ruff?
The hooks are not local, and pre-commit manages the environments for each hook separately.
Please clarify this point with an explicit example.

pyproject.toml
jond01: It wasn't needed in the first place. It's even more redundant now - as pre-commit runs only on tracked files, so there's no need to ignore untracked folders explicitly.

.pre-commit-config.yaml
jond01: Ok, it worked 🙂
96bb5d0

Makefile
jond01: Right, good point. I'll write it in the message.
One may choose a Ruff integration for their IDE and catch this even earlier. I recommend Ruff's VSCode extension for VSCode users.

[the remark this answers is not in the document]
jond01: > missing target note, 

Added in a4b959a.

> also, will that fail if a fix was applied? we need to in order to fail CI

In that case, pre-commit fails -> the step fails -> the CI linting job fails.

.pre-commit-config.yaml
gtopper: It's the year the file was added. You don't need to update all the existing files.

.github/workflows/ci.yaml
gtopper: > "the linter" - what do you mean by that? pre-commit? Ruff?

Ruff.

> The hooks are not local, and pre-commit manages the environments for each hook separately.

Not sure what this means or how it applies here.

> Please clarify this point with an explicit example.

I did, in parentheses. Doesn't ruff need to have the dependencies installed to know if a function is called with valid arguments?

gtopper: Ok, never mind, it just doesn't take the arguments passed at all...

jond01: > Ok, never mind, it just doesn't take the arguments passed at all...

Right. There are other tools that check that, for example: mypy ([ML-5304](https://jira.iguazeng.com/browse/ML-5304)). Pylint also does some static analysis checks, but that part is not in Ruff AFAIK.

gtopper (approved):
Very nice! ✅

Makefile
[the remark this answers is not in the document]
jond01: > > also, will that fail if a fix was applied? we need to in order to fail CI
> 
> In that case, pre-commit fails -> the step fails -> the CI linting job fails.

See https://github.com/mlrun/mlrun/pull/4951/commits/d0ff837ce77d017c74a96baa56ee2819a74d8df7.

jond01:
We currently do not plan to add pre-commit. It can serve as a future reference though in case pre-commit is added.

PR #4954, opened 2024-01-17
[Projects] Delete Project Background Task Support for Iguazio <= 3.5.4

quaark:
In Iguazio <= 3.5.4, we don't support mlrun's new delete background task. Therefore to assure the timeout fix also for older systems we use a slightly different behavior designed by @alonmr:

In order to improve stability with legacy iguazio versions the following changes will be done in mlrun projects API:
1. Projects v1 - 
	- Gets the request from old clients and zebo < 3.5.5.
	- When it gets the request from zebo it needs to create a background task that deletes the project but not wait for it. 
	- Instead it will immediately respond with a 200ish response even though the deletion is still in progress because the timeout in zebo is short and the client is actually waiting for the wrapper request.
2. Wrapper request (v1 and v2) - 
	- When mlrun initially gets the request, it will forward it to zebo. 
	- Now zebo will respond quickly but the project wasn't deleted yet. 
	- The wrapper request should query the project in the DB and wait until it is deleted. 
	- This needs to be in v1 as well for backwards compatibility.
3. Monitoring -
	- No changes here - just note that if the project was deleted in zebo, but not yet in mlrun, then in the next cycle mlrun will archive the project


server/api/api/endpoints/projects.py
```diff
@@ -196,6 +196,14 @@ async def delete_project(
         chief_client = server.api.utils.clients.chief.Client()
         return await chief_client.delete_project(name=name, request=request)
 
+    if server.api.utils.helpers.is_request_from_leader(auth_info.projects_role):
+        # here in DELETE v1/projects, we can assume that if the request is from the leader, then the leader is
+        # iguazio <= 3.5.4. In this older version, the leader isn't waiting for the background task from v2 to complete.
+        # In order for this request not to time out, we want to force the request to not wait for completion and return
+        # 202 immediately. Later in the project deletion wrapper task, after the iguazio job is completed, we will
+        # wait for the project to be removed from the DB and complete the task.
+        wait_for_completion = False
```
alonmr, line 205: wait_for_completion means that we don't wait for the zebo job.
what we want is to not wait from the deletion in mlrun.
you need to create a background task that deletes the project in mlrun and not wait for it.

server/api/api/endpoints/projects_v2.py
```diff
@@ -137,6 +137,10 @@ def _get_or_create_project_deletion_background_task(
            4. Finish LeaderDeletionJob
         4. Finish MLRunDeletionWrapperTask
     """
+    # The project deletion wrapper should wait for the project deletion to complete. This is a backwards compatibility
+    # feature for when working with iguazio <= 3.5.4 that does not support background tasks and therefore doesn't wait
+    # for the project deletion to complete.
+    wait_for_project_deletion = True
```
alonmr, line 143: wait_for_project_deletion should True when igz <= 3.5.4 and request is not from leader.

server/api/api/endpoints/projects_v2.py
```diff
@@ -179,7 +185,20 @@ async def _delete_project(
     project_name: str,
     deletion_strategy: mlrun.common.schemas.DeletionStrategy,
     auth_info: mlrun.common.schemas.AuthInfo,
+    wait_for_project_deletion: bool,
 ):
+    def _verify_project_is_deleted():
+        try:
+            get_project_member().get_project(
```
alonmr, line 192: this should wait for the mlrun background job that was triggered in the v1 endpoint (not yet implemented)
but since you don't have the task name you can query the project in mlrun DB

alonmr (changes requested):
A few notes and just to make the flow clearer.
1. v2 endpoint triggeres the zebo job and waits until the mlrun project is no longer in the DB.
2. zebo sends request to follower
3. v1 endpoint (requested by zebo) starts a background task and sends accepted to zebo.

that's it, the v2 endpoint is waiting for the project to be deleted in mlrun

server/api/api/utils.py
```diff
@@ -1058,3 +1063,118 @@ def artifact_project_and_resource_name_extractor(artifact):
...
+    project_name: str,
+    deletion_strategy: mlrun.common.schemas.DeletionStrategy,
+    auth_info: mlrun.common.schemas.AuthInfo,
+    wait_for_project_deletion: bool,
+):
+    def _verify_project_is_deleted():
+        background_task_handler = (
+            server.api.utils.background_tasks.InternalBackgroundTasksHandler()
+        )
+        active_task = background_task_handler.get_active_background_task_by_kind(
```
alonmr, line 1151: shouldn't you check if the background task finished?

alonmr (commented):
Looks great! minor comment. Would be nice to add a UT as well

alonmr (approved):

server/api/db/sqldb/db.py
```diff
@@ -2351,6 +2354,7 @@ def _get_project_record(
         name: str = None,
         project_id: int = None,
         raise_on_not_found: bool = True,
+        with_for_update: bool = False,
```
liranbg, line 2357: can be removed

server/api/crud/projects.py
```diff
@@ -206,7 +206,7 @@ def delete_project_resources(
                 )
 
     def get_project(
-        self, session: sqlalchemy.orm.Session, name: str
+        self, session: sqlalchemy.orm.Session, name: str, with_for_update: bool = False
```
liranbg: same, remove `with_for_update`

alonmr (approved):
lgtm

PR #4958, opened 2024-01-17
[V3io datastore] use requests session when doing put in chunks

assaf758:


mlrun/datastore/base.py
```diff
@@ -665,9 +665,10 @@ def http_head(url, headers=None, auth=None):
     return response.headers
 
 
-def http_put(url, data, headers=None, auth=None):
+def http_put(url, data, headers=None, auth=None, session=None):
```
gtopper, line 668: This method should be moved to `V3ioStore` (which is the only class that uses it), and it should always reuse the session to avoid recreating the HTTP connection.

mlrun/datastore/v3io.py
```diff
@@ -144,13 +145,15 @@ def _put(self, key, data, max_chunk_size: int = ONE_GB):
             data = memoryview(data)
         except TypeError:
             pass
+        requests_session = requests.Session()
```
gtopper, line 148: As mentioned above, we should probably reuse the session at the class level, but as the code is right now, the session should be closed, preferably like this:
```suggestion
       with requests.Session() as requests_session:
```

mlrun/datastore/base.py
assaf758: Created https://jira.iguazeng.com/browse/ML-5554

gtopper (approved):

PR #4961, opened 2024-01-18
[Artifacts] Fix tag deletion to target only specified artifacts

Yacouby:
The previous implementation of _delete_artifacts_tags gets a list of artifact objects, but then extract their key and deletes all the  tags that belong to artifact with those keys. 
The code should be much more precise and only the specific artifacts chosen should get their tags removed.

Resolves:
https://jira.iguazeng.com/browse/ML-4678


TomerShor (approved):
LGTM ✌️

alonmr (approved):

PR #4966, opened 2024-01-19
[Workflows] Disregard project name on named runs while filtering

laurybueno:
[ML-5540](https://jira.iguazeng.com/browse/ML-5540)

When listing runs, MLRun's UI removes the project string from their names to avoid visual repetition and allow for a cleaner list for the end user since the project's name prefixes every run in a project. However, this change is only done to runs explicitly named by the user when triggered. The ones that weren't named keep the project's name on the UI listing.

Despite being reasonable, this frontend-only change creates an unexpected behavior when a user searches for runs using the project name as the query. In such case, all existing project runs will be listed despite, apparently, only the non-named runs having the `project_name` string on their name. In other words, the named runs shouldn't appear since they don't show the project name at all to the user.

To address the inconsistency, this PR adds a new parameter to the `/projects/<project_name>/pipelines` API route: `name--contains`. When used, the MLRun API will employ its contents on an in-memory filter for all runs returned by Kubeflow. This new filter will evaluate named runs without considering the project's name and allow non-named runs to be filtered with the project's name. This change should make the filtering process work as expected from the perspective of a UI user.

# An important caveat
The new filtering will present a problem when the user requests a list of runs for all projects simultaneously, in other words, when `/projects/*/pipelines` is called.

In this case, MLRun makes a paginated request to Kubeflow but delegates to the client the responsibility of making further calls for each page. When this happens, MLRun cannot guess the total number of records that match the filter. The number informed by the Kubeflow API might not be valid since it doesn't consider the in-memory filtering that MLRun will do on each page. In this case, I suggest we return `-1` on the `total_size` field to indicate to a client that the total number of records is unknown.

### DIFF
--- a/server/api/api/endpoints/pipelines.py
+++ b/server/api/api/endpoints/pipelines.py
@@ -42,6 +42,7 @@ async def list_pipelines(
     sort_by: str = "",
     page_token: str = "",
     filter_: str = Query("", alias="filter"),
+    name__contains: str = Query("", alias="name--contains"),
     format_: mlrun.common.schemas.PipelinesFormat = Query(
         mlrun.common.schemas.PipelinesFormat.metadata_only, alias="format"
     ),
@@ -76,6 +77,7 @@ async def list_pipelines(
             sort_by,
             page_token,
             filter_,
+            name__contains,
             computed_format,
             page_size,
         )
--- a/server/api/api/endpoints/pipelines.py
+++ b/server/api/api/endpoints/pipelines.py
@@ -42,7 +42,7 @@ async def list_pipelines(
     sort_by: str = "",
     page_token: str = "",
     filter_: str = Query("", alias="filter"),
-    name__contains: str = Query("", alias="name--contains"),
+    name_contains: str = Query("", alias="name-contains"),
     format_: mlrun.common.schemas.PipelinesFormat = Query(
         mlrun.common.schemas.PipelinesFormat.metadata_only, alias="format"
     ),
@@ -77,7 +77,7 @@ async def list_pipelines(
             sort_by,
             page_token,
             filter_,
-            name__contains,
+            name_contains,
             computed_format,
             page_size,
         )
--- a/server/api/crud/pipelines.py
+++ b/server/api/crud/pipelines.py
@@ -46,7 +46,7 @@ class Pipelines(
         sort_by: str = "",
         page_token: str = "",
         filter_: str = "",
-        name__contains: str = "",
+        name_contains: str = "",
         format_: mlrun.common.schemas.PipelinesFormat = mlrun.common.schemas.PipelinesFormat.metadata_only,
         page_size: typing.Optional[int] = None,
     ) -> typing.Tuple[int, typing.Optional[int], typing.List[dict]]:
@@ -82,7 +82,7 @@ class Pipelines(
                 run_project = self.resolve_project_from_pipeline(run_dict)
                 if run_project == project:
                     project_runs.append(run_dict)
-            runs = self._filter_runs_by_name(project_runs, project, name__contains)
+            runs = self._filter_runs_by_name(project_runs, name_contains)
             total_size = len(runs)
             next_page_token = None
         else:
@@ -94,7 +94,7 @@ class Pipelines(
                 filter=filter_,
             )
             runs = [run.to_dict() for run in response.runs or []]
-            runs = self._filter_runs_by_name(runs, project, name__contains)
+            runs = self._filter_runs_by_name(runs, name_contains)
             next_page_token = response.next_page_token
             # In-memory filtering turns Kubeflow's counting inaccurate if there are multiple pages of data
             # so don't pass it to the client in such case
@@ -441,13 +441,11 @@ class Pipelines(
                 return data.get("id", "")
         return ""
 
-    @staticmethod
-    def _filter_runs_by_name(runs: list, project_name: str, target_name: str) -> list:
+    def _filter_runs_by_name(self, runs: list, target_name: str) -> list:
         """Filter runs by their name while ignoring the project string on them
 
         :param runs: list of runs to be filtered
-        :param project_name: string to be disregarded while filtering
-        :param name: string that should be part of a valid run name
+        :param target_name: string that should be part of a valid run name
         :return: filtered list of runs
         """
         if not target_name:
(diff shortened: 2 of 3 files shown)


server/api/api/endpoints/pipelines.py
```diff
@@ -42,6 +42,7 @@ async def list_pipelines(
     sort_by: str = "",
     page_token: str = "",
     filter_: str = Query("", alias="filter"),
+    name__contains: str = Query("", alias="name--contains"),
```
theSaarco, line 45: Why the double dash? Could use `name-contains` for this param name.

server/api/crud/pipelines.py
```diff
@@ -433,3 +440,25 @@ def _get_experiment_id_from_run(run: dict) -> str:
...
                 return data.get("id", "")
         return ""
+
+    @staticmethod
+    def _filter_runs_by_name(runs: list, project_name: str, target_name: str) -> list:
+        """Filter runs by their name while ignoring the project string on them
+
+        :param runs: list of runs to be filtered
+        :param project_name: string to be disregarded while filtering
+        :param name: string that should be part of a valid run name
```
theSaarco, line 450:
```suggestion
        :param target_name: string that should be part of a valid run name
```

server/api/crud/pipelines.py
```diff
@@ -433,3 +440,25 @@ def _get_experiment_id_from_run(run: dict) -> str:
...
+        :return: filtered list of runs
+        """
+        if not target_name:
+            return runs
+
+        def filter_by(run):
+            run_name = run.get("name", "")
+            if project_name != "*":
+                run_name = run_name.replace(project_name + "-", "")
```
theSaarco, line 459: At this point, each run has a project in it - we fill that as part fo the `list_pipelines` flow (in `_format_run`). So, instead of passing a project to this function, you can simply use each run's project name to perform the removal - this will work also for the `*` case.

server/api/crud/pipelines.py
```diff
@@ -433,3 +440,25 @@ def _get_experiment_id_from_run(run: dict) -> str:
             ):
                 return data.get("id", "")
         return ""
+
+    @staticmethod
+    def _filter_runs_by_name(runs: list, project_name: str, target_name: str) -> list:
```
theSaarco, line 445: I think you don't need the `project_name` parameter - see my other comment on this function.

theSaarco (changes requested):
Very nice, and good test coverage. I have a suggestion that will improve the `*` case handling.

alonmr (commented):
Not much to add over Saar's comments. Looks good!

server/api/api/endpoints/pipelines.py
```diff
@@ -76,6 +77,7 @@ async def list_pipelines(
             sort_by,
             page_token,
             filter_,
+            name__contains,
```
liranbg, line 80:
```suggestion
            name_contains,
```

server/api/crud/pipelines.py
```diff
@@ -46,6 +46,7 @@ def list_pipelines(
         sort_by: str = "",
         page_token: str = "",
         filter_: str = "",
+        name__contains: str = "",
```
liranbg, line 49:
```suggestion
        name_contains: str = "",
```

liranbg (commented):
minor, I think pretty obvious given @theSaarco review

server/api/api/endpoints/pipelines.py
laurybueno: Some ORMs like to split variable names from operation names with a clear separator (one that hardly ever would be part of a variable name). But I do see that we don't do this on MLRun, so it'd just be a noisy addition to the codebase. 

I removed it for us.

server/api/crud/pipelines.py
laurybueno: Good one!

server/api/crud/pipelines.py
laurybueno: I didn't realize that. Thanks for pointing it out, Saar. I improved the code accordingly.

server/api/crud/pipelines.py
laurybueno: for sure. I removed it for us.

server/api/api/endpoints/pipelines.py
laurybueno: absolutely. Thanks for the reminder.

server/api/crud/pipelines.py
laurybueno: Took care of it here too. Thanks!

server/api/crud/pipelines.py
```diff
@@ -433,3 +440,25 @@ def _get_experiment_id_from_run(run: dict) -> str:
...
+        :param runs: list of runs to be filtered
+        :param target_name: string that should be part of a valid run name
+        :return: filtered list of runs
+        """
+        if not target_name:
+            return runs
+
+        def filter_by(run):
+            run_name = run.get("name", "")
+            if self.resolve_project_from_pipeline(run) != "*":
```
theSaarco, line 456: I don't think `resolve_project_from_pipeline` returns `*` under any scenario. I believe the case you try to handle here is the situation where the function doesn't return anything?

theSaarco (changes requested):
Great. One change is still needed in the implmentation.

server/api/crud/pipelines.py
laurybueno: Good one, Saar!

This, however, brings us to a gap in the task requirements. Do we want `name-contains` not to consider the project's name even when the API request specifies all projects? If so, my last commit should clear any remaining problems. If not, we might need to pass `project_name` to this function after all.

What do you think would be best?

theSaarco: @laurybueno - indeed, the filtering logic should not be affected by the list-query scope. From user perspective, the workflow name is the part that's after the project name prefix (which is how the user called the workflow to begin with), and this shouldn't change because the query was for all projects. So, each workflow name should always be stripped of its respective project name before doing the `name-contains` filter. The only exception is the case where the workflow name was originally empty, in which case we assume the workflow name is the project name (not performing stripping). 
Looks like your new code fits this description (with a small glitch, see my new comment).

server/api/crud/pipelines.py
```diff
@@ -433,3 +440,23 @@ def _get_experiment_id_from_run(run: dict) -> str:
...
+
+        :param runs: list of runs to be filtered
+        :param target_name: string that should be part of a valid run name
+        :return: filtered list of runs
+        """
+        if not target_name:
+            return runs
+
+        def filter_by(run):
+            run_name = run.get("name", "").replace(
```
theSaarco, line 455: Using `replace` is too permissive - it will remove project name from anywhere in the string, not just the prefix. Should be replaced with something using `startswith` or since we're in py3.9, you can use `str.removeprefix()`.

theSaarco (changes requested):
One more small thing... 🙂

theSaarco (approved):
❤️

liranbg (approved):

PR #4969, opened 2024-01-21
[Iguazio] Cache project deletion and creation jobs to avoid consecutive requests

TomerShor:
An error was seen in Iguazio where multiple delete project requests were sent to the Iguazio leader, and several deletion jobs were triggered although only one actually deleted the project.
This resulted in hanging jobs and request timeouts in the mlrun client invoking the delete requests.

To overcome this, we introduce project deletion / creation job caches.
When MLRun's Iguazio client calls Iguazio's delete project endpoint it will cache the project name to Iguazio job id. 
That way, any subsequent requests will return the same job instead of triggering a new one.
After waiting for the job to complete we clean the cache. 
In addition, when the job is cached there is a scheduled invalidation task to remove the project from the cache if it wasn't removed earlier.

Furthermore, if a new project deletion request arrived, and there is already a cached job for this project, we check if that project is already done and if so - remove it manually from the cache.
This can occur if the project was re-created and deleted again before the cache TTL has reached.

The same solution is implemented on the create project method, for protection there as well.

Resolves https://jira.iguazeng.com/browse/ML-5555

### DIFF
--- a/server/api/utils/clients/iguazio.py
+++ b/server/api/utils/clients/iguazio.py
@@ -19,6 +19,7 @@ import datetime
 import enum
 import http
 import json
+import threading
 import typing
 import urllib.parse
 
@@ -34,6 +35,7 @@ import mlrun.utils.helpers
 import mlrun.utils.singleton
 import server.api.utils.projects.remotes.leader as project_leader
 from mlrun.utils import get_in, logger
+from server.api.utils.periodic import run_function_periodically
 
 
 class JobStates:
@@ -72,6 +74,54 @@ class SessionPlanes:
         ]
 
 
+class JobCache:
+    """
+    Cache for create/delete project jobs.
+    This cache is used to avoid consecutive create/delete jobs for the same project,
+    while the jobs are still in progress.
+    """
+
+    def __init__(self):
+        self._delete_jobs_cache = {}
+        self._create_jobs_cache = {}
+        self._delete_lock = threading.Lock()
+        self._create_lock = threading.Lock()
+
+    def get_delete_jobs_cache(self) -> typing.Dict[str, dict]:
+        return self._delete_jobs_cache
+
+    def get_create_jobs_cache(self) -> typing.Dict[str, dict]:
+        return self._create_jobs_cache
+
+    def get_delete_job_id(self, project: str) -> typing.Optional[dict]:
+        return self._delete_jobs_cache.get(project, {}).get("job_id")
+
+    def get_create_job_id(self, project: str) -> typing.Optional[dict]:
+        return self._create_jobs_cache.get(project, {}).get("job_id")
+
+    def set_delete_job(self, project: str, job_id: dict):
+        with self._delete_lock:
+            self._delete_jobs_cache[project] = {
+                "job_id": job_id,
+                "timestamp": datetime.datetime.now(),
+            }
+
+    def set_create_job(self, project: str, job_id: str):
+        with self._delete_lock:
+            self._delete_jobs_cache[project] = {
+                "job_id": job_id,
+                "timestamp": datetime.datetime.now(),
+            }
+
+    def remove_delete_job(self, project: str):
+        with self._delete_lock:
+            del self._delete_jobs_cache[project]
+
+    def remove_create_job(self, project: str):
+        with self._create_lock:
+            del self._create_jobs_cache[project]
+
+
 class Client(
     project_leader.Member,
     metaclass=mlrun.utils.singleton.AbstractSingleton,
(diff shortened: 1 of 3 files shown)


server/api/utils/clients/iguazio.py
```diff
@@ -72,6 +74,87 @@ def all():
...
+class JobCache:
+    """
+    Cache for create/delete project jobs.
+    This cache is used to avoid consecutive create/delete jobs for the same project,
+    while the jobs are still in progress.
+    """
+
+    def __init__(self, ttl: str):
+        self._delete_jobs = {}
+        self._create_jobs = {}
```
alonmr, line 86: create is not needed?

server/api/utils/clients/iguazio.py
```diff
@@ -72,6 +74,87 @@ def all():
...
+        self._delete_locks = {}
+        self._create_locks = {}
+
+        self._ttl = humanfriendly.parse_timespan(ttl)
+
+    def get_delete_lock(self, project: str):
+        if project not in self._delete_locks:
+            self._delete_locks[project] = threading.Lock()
```
alonmr, line 94: what if you get a context switch between those lines?

server/api/utils/clients/iguazio.py
```diff
@@ -191,50 +278,43 @@ def delete_project(
...
-            mlrun.common.schemas.Project(
-                metadata=mlrun.common.schemas.ProjectMetadata(name=name)
+        with self._job_cache.get_delete_lock(name):
+            # check if project is already being deleted
+            job_id = self._job_cache.get_delete_job_id(name)
+
+            # the existing job might have already been completed, but the cache was not yet invalidated
+            # in that case, we want to create a new deletion job
+            if job_id and self._is_job_done(session, job_id):
+                self._job_cache.remove_delete_job(name)
```
alonmr, line 288: you need to acquire the lock and ensure you are deleting the correct job id

server/api/utils/clients/iguazio.py
```diff
@@ -783,6 +926,25 @@ def _handle_error_response(
...
+        """
+        try:
+            response = self._get_job_from_iguazio(session, job_id)
+            return (
+                response["data"]["attributes"]["state"] in JobStates.terminal_states()
+            )
+        except requests.HTTPError as exc:
+            if exc.response.status_code == http.HTTPStatus.NOT_FOUND.value:
+                # job not found - consider it as done
+                return True
```
alonmr, line 945: I don't think you should decide the 404 is done here. You should raise mlrun not found error and decide what you want to do outside

alonmr (commented):
I didn't review the creation part because i thought we said we would remove it. anyway a few comments for now

server/api/utils/clients/iguazio.py
TomerShor: The lock is already acquired in this block, a few lines above it :
```
with self._job_cache.get_delete_lock(name):
```

server/api/utils/clients/iguazio.py
TomerShor: removed

server/api/utils/clients/iguazio.py
TomerShor: locked 🔒

server/api/utils/clients/iguazio.py
```diff
@@ -484,10 +589,7 @@ def _get_project_from_iguazio(
 
     def _wait_for_job_completion(self, session: str, job_id: str, error_message: str):
         def _verify_job_in_terminal_state():
-            response = self._send_request_to_api(
-                "GET", f"jobs/{job_id}", "Failed getting job from Iguazio", session
-            )
-            response_body = response.json()
+            response_body = self._get_job_from_iguazio(session, job_id)
```
liranbg, line 592:
```suggestion
            job_response_body = self._get_job_from_iguazio(session, job_id)
```

liranbg (changes requested):
overall looking good, some minor suggestion / fixes are required

tests/api/utils/clients/test_iguazio.py
```diff
@@ -810,6 +811,132 @@ async def test_delete_project_without_wait(
...
+    job_id = "some-job-id"
+
+    client._job_cache.__getattribute__(f"set_{cache_kind}_job")(project_name, job_id)
+
+    assert (
+        client._job_cache.__getattribute__(f"get_{cache_kind}_job_id")(project_name)
+        == job_id
+    )
+
+    await asyncio.sleep(3)
```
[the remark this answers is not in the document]
TomerShor: reduced `ttl` to 1 sec in the test

quaark (approved):

PR #4970, opened 2024-01-21
[Artifacts] Overwrite `spec.db_key` when importing artifact with `new_key` arg

TomerShor:
Following https://github.com/mlrun/mlrun/pull/4917, we retain existing `spec.db_key` when logging artifacts.

When importing artifacts using `project.import_artifact`, we can set an explicit `new_key`. This `new_key` was set in the artifact's metadata, but not in the spec db_key property, causing artifacts with an existing `db_key` to be saved in the db with it instead of the new key.

In this PR - if `new_key` is given it will overwrite both `metadata.key` and `spec.db_key`.

Fixes https://jira.iguazeng.com/browse/ML-5544

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -1672,6 +1672,10 @@ class MlrunProject(ModelObj):
             artifact.metadata.updated = None
             artifact.metadata.tag = tag or artifact.metadata.tag
             if new_key:
+                if artifact.spec.db_key:
+                    logger.warning(
+                        f"Overwriting artifact old db_key '{artifact.spec.db_key}' with new key '{new_key}'"
+                    )
                 artifact.spec.db_key = new_key
             return artifact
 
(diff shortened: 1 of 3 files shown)


mlrun/projects/project.py
```diff
@@ -1671,6 +1671,12 @@ def get_artifact(spec):
             artifact.metadata.project = self.metadata.name
             artifact.metadata.updated = None
             artifact.metadata.tag = tag or artifact.metadata.tag
+            if new_key:
+                if artifact.spec.db_key:
+                    logger.warning(
+                        f"Overwriting artifact old db_key '{artifact.spec.db_key}' with new key '{new_key}'"
+                    )
```
liranbg, line 1678: add the "side-effect" of that ("artifact can be fetched via the new_key only")

liranbg (approved):

PR #4976, opened 2024-01-21
[Artifacts] log_artifact: check file existence for local filesystem path

tomerm-iguazio:
[ML-4873](https://jira.iguazeng.com/browse/ML-4873)


mlrun/artifacts/manager.py
```diff
@@ -121,6 +122,26 @@ def __init__(
         self.input_artifacts = {}
         self.artifacts = {}
 
+    @staticmethod
+    def validate_artifact_location(item, path, body):
```
theSaarco, line 126: Would change the name a bit to something like `ensure_artifact_source_file_exists`. Also, please add some comments here explaining the logic and why you do the exceptions that you do (if we have body, then no file is needed, model artifact is a directory etc.).

mlrun/utils/helpers.py
```diff
@@ -1579,3 +1580,9 @@ def to_parquet(df, *args, **kwargs):
 def is_ecr_url(registry: str) -> bool:
     # example URL: <aws_account_id>.dkr.ecr.<region>.amazonaws.com
     return ".ecr." in registry and ".amazonaws.com" in registry
+
+
+def get_local_file_schema() -> List:
+    # The expression `list(string.ascii_lowercase)` generates a list of lowercase alphabets,
+    # which corresponds to drive letters in Windows file paths such as `C:/Windows/path`.
```
theSaarco, line 1587: Nice! Didn't think of it...

tests/artifacts/test_artifacts.py
```diff
@@ -344,6 +344,22 @@ def test_log_artifact_with_target_path_and_upload_options():
                 assert logged_artifact.metadata.hash is None
 
 
+@pytest.mark.parametrize(
+    "local_path, fail", [("s3://path/file.txt", False), ("/not_exists/file.txt", True)]
```
theSaarco, line 348: Since you did advanced validation on local paths, might as well check it here... Add some test cases for things like `file://some/path/file.txt` or `C:/hello/world.bat`.

theSaarco (approved):
Looks good! Have some improvement suggestions, see comments.

mlrun/utils/helpers.py
tomerm-iguazio: tnx!

mlrun/artifacts/manager.py
```diff
@@ -121,6 +122,28 @@ def __init__(
         self.input_artifacts = {}
         self.artifacts = {}
 
+    @staticmethod
+    def ensure_artifact_source_file_exists(item, path, body):
+        # If the body exists, the source path does not have to be exists.
```
assaf758, line 127:
```suggestion
        # If the body exists, the source path does not have to exists.
```

mlrun/artifacts/manager.py
```diff
@@ -121,6 +122,28 @@ def __init__(
...
+        parsed_url = urlparse(path)
+        schema = parsed_url.scheme
+        #  we are not checking remote paths yet.
+        if schema and schema not in get_local_file_schema():
+            return
+        if schema.lower() == "file":
+            path = parsed_url.path
+        if not exists(path):
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"Failed to log an artifact, path does not exists. path:{path}"
```
assaf758, line 144:
```suggestion
                f"Failed to log an artifact, file does not exists at path {path}"
```

PR #4988, opened 2024-01-22
[API] Fix background jobs not playing nice with fastapi middlewares

liranbg:
Observed on some envs that mlrun api getting stuck executing background jobs when more than one endpoints requires it and a background job is running (in the background, duh)

Ive yet to fully dive deep to understand what exactly blocks it but looking at https://github.com/encode/starlette/issues/919 it  seems that we have experienced the same symptoms, thus, I migrated our fastapi-like middlewares to be wsgi-like middlewares. havent managed to reproduce it and seems that this fix nails it.

Reference for the future
1. works when the requests do come and go to the same endpoint using background tasks
2. blocks when requests come to different endpoints, while both require fastapi background task
3. happen transiently and under unknown clear circumstances 

In addition, enhanced the logger middleware to take the request id from x-request-id so that requests getting to mlrun from nginx (or any load balancer that sets such header) would carry on as mlrun's request id. this is useful when requests travel between workers and chief

https://jira.iguazeng.com/browse/ML-5571


server/api/middlewares/ensure_be_version.py
```diff
@@ -0,0 +1,55 @@
...
+        self.app = app
+        self._backend_version = backend_version
+
+    async def __call__(
+        self, scope: "Scope", receive: "ASGIReceiveCallable", send: "ASGISendCallable"
+    ) -> None:
+        """
+        This middleware ensures response header includes backend version
+        """
+        if scope["type"] not in ("http", "websocket"):
```
alonmr, line 43: where do we use websockets?

server/api/middlewares/ui_clear_cache.py
```diff
@@ -0,0 +1,80 @@
...
+            mlrun.common.schemas.constants.HeaderNames.ui_version, ""
+        )
+
+        async def send_wrapper(message: Message) -> None:
+            if message["type"] == "http.response.start":
+                if (
+                    ui_version
+                    and ui_version != config.version
+                    and not self._is_development_version()
+                ):
```
quaark, line 65:
```suggestion
            if (
                message["type"] == "http.response.start"
                and ui_version
                and ui_version != config.version
                and not self._is_development_version()
            ):
```

quaark (changes requested):
Very minor, otherwise looks good!

server/api/middlewares/ensure_be_version.py
liranbg: we dont, that is mainly asgi convention of writing middlewares

quaark (approved):

PR #4994, opened 2024-01-23
[Model Monitoring] Log windows result in the controller job

jond01:
Resolves [ML-5453](https://jira.iguazeng.com/browse/ML-5453).
Since all apps have the same policy, log the results in the format:
```
{"model-endpoint": ["window-i-start", "window-i+1-start"]}
```

When an application is triggered - the UI shows an explicit result in the run's row:
image

The relevant system test passed (`pytest tests/system/model_monitoring/test_app.py::TestMonitoringAppFlow`).


assaf758 (approved):

assaf758 (approved):
Nice debugging addition!

PR #5006, opened 2024-01-25
[DB] Update labels for existing features if needed

TomerShor:
**Bug:**
A feature set has features without any labels.
When adding a feature with the same name and  new labels, the labels are not saved in the database.

**RCA:**
In the DB layer, when we update the feature (in `_update_feature_set_features`) we only look at the `features_to_add`, according to their name.
Since the features already exist on the feature set, we ignore them and don't update their labels.

**Fix:**
Check the existing features labels compared to the "new" feature labels, and update them if needed.

Resolves https://jira.iguazeng.com/browse/ML-5574

### DIFF
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -2936,9 +2936,10 @@ class SQLDB(DBInterface):
                     ),
                     None,
                 )
-                # update it with the new labels in case they were changed
-                labels = feature_dict.get("labels") or {}
-                update_labels(feature, labels)
+                if feature:
+                    # update it with the new labels in case they were changed
+                    labels = feature_dict.get("labels") or {}
+                    update_labels(feature, labels)
 
     @staticmethod
     def _update_feature_set_entities(feature_set: FeatureSet, entity_dicts: List[dict]):
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -2927,7 +2927,7 @@ class SQLDB(DBInterface):
                 update_labels(feature, labels)
                 feature_set.features.append(feature)
             elif feature_name not in features_to_remove:
-                # check if the existing feature's labels were changed, and update them if needed
+                # get the existing feature from the feature set
                 feature = next(
                     (
                         feature
@@ -2936,11 +2936,9 @@ class SQLDB(DBInterface):
                     ),
                     None,
                 )
+                # update it with the new labels in case they were changed
                 labels = feature_dict.get("labels") or {}
-                if not server.api.utils.helpers.are_object_labels_equal_to_labels_dict(
-                    feature.labels, labels
-                ):
-                    update_labels(feature, labels)
+                update_labels(feature, labels)
 
     @staticmethod
     def _update_feature_set_entities(feature_set: FeatureSet, entity_dicts: List[dict]):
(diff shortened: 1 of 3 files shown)


server/api/db/sqldb/db.py
```diff
@@ -2925,6 +2926,21 @@ def _update_feature_set_features(
...
+                        feature
+                        for feature in feature_set.features
+                        if feature.name == feature_name
+                    ),
+                    None,
+                )
+                labels = feature_dict.get("labels") or {}
+                if not server.api.utils.helpers.are_object_labels_equal_to_labels_dict(
+                    feature.labels, labels
+                ):
```
alonmr, line 2942: I don't see why you need this check. Just update the labels

server/api/db/sqldb/db.py
```diff
@@ -2925,6 +2926,21 @@ def _update_feature_set_features(
...
+                feature = next(
+                    (
+                        feature
+                        for feature in feature_set.features
+                        if feature.name == feature_name
+                    ),
+                    None,
+                )
+                labels = feature_dict.get("labels") or {}
+                if not server.api.utils.helpers.are_object_labels_equal_to_labels_dict(
```
quaark, line 2940: Use `deepdiff`, no need to implement it yourself

server/api/db/sqldb/db.py
```diff
@@ -2925,6 +2926,21 @@ def _update_feature_set_features(
...
+            elif feature_name not in features_to_remove:
+                # check if the existing feature's labels were changed, and update them if needed
+                feature = next(
+                    (
+                        feature
+                        for feature in feature_set.features
+                        if feature.name == feature_name
+                    ),
+                    None,
+                )
```
quaark, line 2938: While this is cool, I got confused reading this, and assumed the comment of 'checking if the existing feature's labels were changed' was part of this generator with the filter.
I'd either change the comment to something like:
"If the feature already exists, and its labels were changed, update it with the new labels"
To explicitly express the steps you are taking in the code.

quaark (changes requested):
Looks good! some minor stuff

server/api/db/sqldb/db.py
quaark: Actually yeah I agree with Alon on this one

TomerShor: but why update if there is nothing to update?

quaark: It's the same runtime and memory usage as checking first and then not updating

TomerShor: alright then, I'll remove the entire extra logic :)

server/api/db/sqldb/db.py
TomerShor: removed the check altogether

server/api/db/sqldb/db.py
TomerShor: Updated the comments

alonmr (approved):
Nice LGTM

server/api/db/sqldb/db.py
```diff
@@ -2925,6 +2926,19 @@ def _update_feature_set_features(
                 )
                 update_labels(feature, labels)
                 feature_set.features.append(feature)
+            elif feature_name not in features_to_remove:
+                # get the existing feature from the feature set
+                feature = next(
```
liranbg, line 2931: what happen when `feature` is None?

TomerShor: if the feature is not in `features_to_remove` and not in `features_to_remove`, it is in the current features.
I will add a check just in case.

PR #5010, opened 2024-01-25
[Docs] Setting `min_replicas` and `max_replicas` in `KafkaSource`

jillnogold:
[ML-5545](https://jira.iguazeng.com/browse/ML-5545)

### DIFF
--- a/docs/serving/graph-ha-cfg.md
+++ b/docs/serving/graph-ha-cfg.md
@@ -22,7 +22,7 @@ lost. The mechanism of Window ACK provides a solution for such
 stateful event processing.
 
 ```{admonition} Note
-For statefeul functions, each worker has its own state.
+For statefull functions, each worker has its own state. See, for example, {py:meth}`~storey.transformations.MapWithState`.
 ```
 
 With Window ACK, the consumer group's committed offset is delayed by one window, committing the offset at (processed event num – window). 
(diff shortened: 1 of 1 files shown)


docs/serving/graph-ha-cfg.md
```diff
@@ -32,9 +36,15 @@ graph structure and should be calculated accordingly. The following sections exp
 
 A consumer function is essentially a Nuclio function with a stream trigger. As part of the trigger, you can set a consumer group.  
 
-When the consumer function is part of a graph then the consumer function’s number of replicas is derived from the number of shards and is 
-therefore nonconfigurable. The same applies to the number of workers in each replica, which is set to 1 and is not configurable.  
-
+When the consumer function is part of a graph then the number of replicas per function depends on the  source:
+- StreamSource: The number of replicas is derived from the number of shards and is therefore nonconfigurable. The same applies to the number of workers in each replica, which is set to 1 and is not configurable.
```
gtopper, line 40: I think this is clearer (because it being "the same" is initially ambiguous):
```suggestion
- `StreamSource`: The number of replicas is derived from the number of shards and is therefore nonconfigurable. Furthermore, the number of workers in each replica is set to 1 and is also not configurable.  
```

docs/serving/graph-ha-cfg.md
```diff
@@ -32,9 +36,15 @@ graph structure and should be calculated accordingly. The following sections exp
 
 A consumer function is essentially a Nuclio function with a stream trigger. As part of the trigger, you can set a consumer group.  
 
-When the consumer function is part of a graph then the consumer function’s number of replicas is derived from the number of shards and is 
-therefore nonconfigurable. The same applies to the number of workers in each replica, which is set to 1 and is not configurable.  
-
+When the consumer function is part of a graph then the number of replicas per function depends on the  source:
```
gtopper, line 39: I don't think this is specific to the function being "a part of a graph" (meaning a multi-function graph?). Rather, it's always the case that we applies this logic to control the number of replicas.

docs/serving/graph-ha-cfg.md
```diff
@@ -21,6 +21,10 @@ performs additional processing of events after the function handler returns, the
 lost. The mechanism of Window ACK provides a solution for such 
 stateful event processing.
 
+```{admonition} Note
+For statefeul functions, each worker has its own state.
```
gtopper, line 25: Maybe mention `MapWithState` as an example of this.

docs/serving/graph-ha-cfg.md
jillnogold: updated

docs/serving/graph-ha-cfg.md
jillnogold: updated

docs/serving/graph-ha-cfg.md
jillnogold: added

docs/serving/graph-ha-cfg.md
```diff
@@ -21,6 +21,10 @@ performs additional processing of events after the function handler returns, the
 lost. The mechanism of Window ACK provides a solution for such 
 stateful event processing.
 
+```{admonition} Note
+For statefull functions, each worker has its own state. See, for example, {py:meth}`~storey.transformations.MapWithState`.
```
gtopper, line 25:
```suggestion
For stateful functions, each worker has its own state. See, for example, {py:meth}`~storey.transformations.MapWithState`.
```

PR #5012, opened 2024-01-25
[Monitoring] Fix inconsistent runs overriding

alonmr:
https://jira.iguazeng.com/browse/ML-4991

Prefer patch run instead of store run when possible (run exists) so that only given attributes will be updated instead of overriding the full run struct in DB.


liranbg (approved):
🎸

PR #5013, opened 2024-01-26
[Datastore] Forbid unlisted parameters in datastore profile initialization

alxtkr77:
https://jira.iguazeng.com/browse/ML-5595


assaf758 (changes requested):
instead of "forbid", use pydentic.Extra.forbid

assaf758 (approved):

PR #5014, opened 2024-01-27
[Logger] Use orjson to dump log lines

liranbg:
Making it a bit faster but the main reason for this change is to enable logging of pydantic structure / arbitrary datetime objects when being logged


alonmr (approved):
LGTM

PR #5019, opened 2024-01-28
[DB] Lock db rows when tagging artifacts

TomerShor:
We have seen cases where parallel runs with the same "producer_id" are all trying to log artifacts with the same tag simultaneously.
This sometimes caused multiple "latest" tags pointing to artifacts with the same key, iteration and producer id, due to the tagging race conditions.

To avoid that, we want to lock the row of the specific artifacts that are being tagged.
This is done using SQL's "FOR UPDATE" (see [this](https://www.cockroachlabs.com/blog/select-for-update/) for more info).
The artifacts will be locked for the entire transaction, until the session is committed.

Resolves https://jira.iguazeng.com/browse/ML-5596

### DIFF
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -902,8 +902,8 @@ class SQLDB(DBInterface):
             session,
             ArtifactV2,
             project=project,
-        ).filter(
-            ArtifactV2.key.in_([artifact.key for artifact in artifacts])
+        ).filter(ArtifactV2.key.in_([artifact.key for artifact in artifacts])).order_by(
+            ArtifactV2.key.asc()
         ).populate_existing().with_for_update().all()
 
         objects = []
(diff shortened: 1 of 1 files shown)


alonmr (changes requested):
A few things to note here:
1. You shouldn't lock rows in a for loop. You risk deadlocks (can be avoided by sorting the artifacts list before locking)
2. It is recommended to use `populate_existing` when using `with_for_update` by sqlalchemy.
3. I don't think this fixes the issue because if 2 latest tags were created that means that both requests didn't find the row (because it didn't exist). locking a row that doesn't exist does nothing.

server/api/db/sqldb/db.py
```diff
@@ -896,9 +896,19 @@ def tag_artifacts(
         artifacts,
         project: str,
     ):
+        # to avoid multiple runs trying to tag the same artifacts simultaneously,
```
alonmr, line 899: What happens if some of the artifacts are locked? I assume it will wait until it can lock all rows but does it mean that the other rows are locked until the lock we wait for is released? I would assume that mysql has a good mechanism for it (FCFS?) but we should verify it

liranbg (approved):

PR #5023, opened 2024-01-29
[Datastore] Clarify error messages when using spark to ingest unsupported sources

alxtkr77:
https://jira.iguazeng.com/browse/ML-5579

### DIFF
--- a/mlrun/datastore/sources.py
+++ b/mlrun/datastore/sources.py
@@ -1020,6 +1020,12 @@ class KafkaSource(OnlineSource):
 
         return function
 
+    def to_spark_df(self, session, named_view=False, time_field=None, columns=None):
+        raise NotImplementedError(
+            "Conversion of a source of type 'KafkaSource' "
+            "to a Spark dataframe is not possible, as this operation is not supported by Spark"
+        )
+
 
 class SQLSource(BaseSourceDriver):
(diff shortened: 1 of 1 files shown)


mlrun/datastore/sources.py
```diff
@@ -118,7 +118,10 @@ def to_spark_df(self, session, named_view=False, time_field=None, columns=None):
             if named_view:
                 df.createOrReplaceTempView(self.name)
             return self._filter_spark_df(df, time_field, columns)
-        raise NotImplementedError()
+        raise NotImplementedError(
+            f"Conversion of a source of type '{type(self).__name__}' "
+            "to a Spark dataframe is not possible, as this operation is not supported"
```
assaf758, line 123:
```suggestion
            "to a Spark dataframe is not possible, as this operation is not supported by Spark"
```

gtopper: Pretty sure spark supports reading from SQL, but `SQLSource` doesn't support  spark. So it's probably not right to blame spark, at least not in all cases.

assaf758: Following Gal's comment, I suggest we limit my change to KafkaSource, and leave your comment for the rest of the sources.

assaf758 (approved):

PR #5027, opened 2024-01-30
[Datasets] Support uploading dataset body when it is not a dataframe

TomerShor:
When logging a dataset artifact and uploading to the target path, it was only uploaded if it contained a dataframe.
For other cases, such as where the dataset was an imported parquet file, it was just ignored and wasn't actually uploaded to the data store, causing it to be basically useless.

In this PR, when uploading a dataset we check if the dataframe is set, and if not we try to upload the dataset's body.

Resolves https://jira.iguazeng.com/browse/ML-5613


liranbg (approved):
🚀

PR #5031, opened 2024-01-30
[Artifacts] Warn loading legacy artifacts

alonmr:
Users should be warned that loading legacy artifacts will not be supported in 2 minor versions.
This ensures each time we load one, we tell the users that they need to save the artifact in the new format.

### DIFF
--- a/mlrun/artifacts/base.py
+++ b/mlrun/artifacts/base.py
@@ -16,6 +16,7 @@ import os
 import pathlib
 import tempfile
 import typing
+import warnings
 import zipfile
 
 import yaml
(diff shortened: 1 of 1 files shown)


mlrun/artifacts/base.py
```diff
@@ -1067,6 +1068,17 @@ def convert_legacy_artifact_to_new_format(
...
         )
 
+    artifact_key = legacy_artifact_dict.get("key", "")
+    artifact_tag = legacy_artifact_dict.get("tag", "")
+    if artifact_tag:
+        artifact_key = f"{artifact_key}:{artifact_tag}"
+    # TODO: remove in 1.8.0
+    warnings.warn(
+        f"Converting legacy artifact '{artifact_key}' to new format. This will not be supported in MLRun 1.8.0. "
+        f"Make sure to save the artifact/project in the new format.",
```
TomerShor, line 1078: Are you sure saving the artifact will achieve the desired effect?

alonmr: It depends on what the user is doing. It is possible that they loaded the artifact from artifact yaml (import_artifact) and not from project yaml. They will need to save the artifact or the project depending on what they loaded.

TomerShor (approved):

PR #5039, opened 2024-01-31
[Packager] Add timestamp to archive directory path for uniqueness

TomerShor:
The packager saves file artifacts as zip archives.
When zip artifacts are exported and the imported, the files lose their `.zip` suffix, but still remain extractable zip files.

When the packager then tries to unpack these archives, it assumes the files has a `.zip` suffix, and tries to extract them to a directory of the same name (e.g. extract `my-artifact.zip` to `my-artifact` dir). 
The directory creation fails in case the suffix is missing, because a file of the same name already exists.

The simple solution is to just add a random suffix (`datetime.now()`) to the directory name, as it is just a temporary directory. The directory can then be created and the artifacts are usable in the run.

Fixes https://jira.iguazeng.com/browse/ML-5614

### DIFF
--- a/mlrun/package/utils/_archiver.py
+++ b/mlrun/package/utils/_archiver.py
@@ -12,12 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-import datetime
 import os
 import tarfile
 import zipfile
 from abc import ABC, abstractmethod
 from pathlib import Path
+
+import mlrun.utils
 
 from ._supported_format import SupportedFormat
 
@@ -104,7 +105,7 @@ class _ZipArchiver(_Archiver):
         # Create the directory path, add timestamp to avoid collisions:
         directory_path = output_path / archive_path.stem
         directory_path = directory_path.with_name(
-            f"{directory_path.name}-{datetime.datetime.now()}"
+            f"{directory_path.name}-{mlrun.utils.now_date().isoformat()}"
         )
         os.makedirs(directory_path)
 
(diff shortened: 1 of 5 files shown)


mlrun/package/utils/_archiver.py
```diff
@@ -100,8 +101,11 @@ def extract_archive(cls, archive_path: str, output_path: str) -> str:
         archive_path = Path(archive_path)
         output_path = Path(output_path)
 
-        # Create the directory path:
+        # Create the directory path, add timestamp to avoid collisions:
         directory_path = output_path / archive_path.stem
+        directory_path = directory_path.with_name(
+            f"{directory_path.name}-{datetime.datetime.now()}"
```
liranbg, line 107:
```suggestion
            f"{directory_path.name}-{mlrun.utils.now_date()}"
```

liranbg (approved):

mlrun/package/utils/_archiver.py
```diff
@@ -18,6 +18,8 @@
 from abc import ABC, abstractmethod
 from pathlib import Path
 
+import mlrun.utils
```
guy1992l, line 21: I don't get why you need to import `mlrun.utils` just for `datetime.now()`, the `mlrun.utils.now_date()` is kinda redundant in my opinion.

guy1992l (commented):
Great catch! <3

mlrun/package/utils/_archiver.py
liranbg: well it gives something (timezone aware)

PR #5040, opened 2024-02-01
[Tests] Fix K8s Secrets Mock + Project Deletion UT uses both API versions

quaark:
Explanation for fixing the k8s secrets mock:
There was a weird case where `test_delete_project_with_resources` was using the `unversioned_client` while `k8s_secrets_mock` was itself calling the `client` fixture. When the test previously used `client` they were synched as the same client, but when using `unversioned_client` it initialized two fixtures, each one initialized a new FastAPI app on the same thread. So in the end of the test one app teared down successfully, but when the second app tried to teardown the event loop was already closed. And the teardown fails.


alonmr (approved):
🔥

tests/api/api/test_projects.py
```diff
@@ -193,31 +199,31 @@ def test_delete_project_with_resources(
     )
 
     # deletion strategy - check - should fail because there are resources
-    response = client.delete(
-        f"projects/{project_to_remove}",
+    response = unversioned_client.delete(
+        f"{api_version}/projects/{project_to_remove}",
```
TomerShor, line 203: Move the URL to a variable since it is the same for every request in the test

TomerShor (changes requested):

TomerShor (approved):
Noice 👌

PR #5043, opened 2024-02-01
[Dataset] Don't fail uploading dataset if `df` or `body` are empty

TomerShor:
An exception was mistakenly added in https://github.com/mlrun/mlrun/pull/5027 when there is no dataframe or body to the upload in a dataset artifact.
In this PR we maintain the previous behavior of simply setting the artifact's size and hash to `None` if there is no df to upload.

Resolves https://jira.iguazeng.com/browse/ML-5645


liranbg (changes requested):
Make sure you add ut coverage for this specific flow

liranbg (approved):

PR #5045, opened 2024-02-01
[Datastore] Fix previewing artifacts associated with datastore profile

alxtkr77:
https://jira.iguazeng.com/browse/ML-5638
https://jira.iguazeng.com/browse/ML-5634

### DIFF
--- a/mlrun/datastore/datastore_profile.py
+++ b/mlrun/datastore/datastore_profile.py
@@ -367,7 +367,7 @@ class DatastoreProfile2Json(pydantic.BaseModel):
             )
 
 
-def datastore_profile_read(url):
+def datastore_profile_read(url, project_name=""):
     parsed_url = urlparse(url)
     if parsed_url.scheme.lower() != "ds":
         raise mlrun.errors.MLRunInvalidArgumentError(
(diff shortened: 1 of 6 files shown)


mlrun/datastore/datastore_profile.py
```diff
@@ -367,15 +367,15 @@ def safe_literal_eval(value):
...
+def datastore_profile_read(url, project_name=""):
     parsed_url = urlparse(url)
     if parsed_url.scheme.lower() != "ds":
         raise mlrun.errors.MLRunInvalidArgumentError(
             f"resource URL '{url}' cannot be read as a datastore profile because its scheme is not 'ds'"
         )
 
     profile_name = parsed_url.hostname
-    project_name = parsed_url.username or mlrun.mlconf.default_project
+    project_name = project_name or mlrun.mlconf.default_project
```
gtopper, line 378: Are you sure we don't want to keep `parsed_url.username` as an option?

alxtkr77: I am.  This was supposed to be a way to pass project name , and now we remove it (see https://jira.iguazeng.com/browse/ML-5634)

gtopper (approved):
LGTM ✅

PR #5058, opened 2024-02-04
[Model Monitoring] Fix path filtering according to nuclio changes

davesh0812:
Works also on previous nuclio versions.
According to @TomerShor the changes are only when using `V3ioStream`.
[ML-5659](https://jira.iguazeng.com/browse/ML-5659)

### DIFF
--- a/mlrun/model_monitoring/stream_processing.py
+++ b/mlrun/model_monitoring/stream_processing.py
@@ -180,7 +180,7 @@ class EventStreamProcessor:
 
         apply_event_routing()
 
-        # Step 2 - Filter out events with '-' in path which indicates that the event is supposed to be processed
+        # Step 2 - Filter out events with '-' in the path basename from going forward
         # through the next steps of the stream graph
         def apply_storey_filter_stream_events():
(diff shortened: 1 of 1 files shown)


Eyal-Danieli (approved):
cool cool cool
We just need to make sure that it works in CE (prometheus) as expected. Anyway, it seems correct.

mlrun/model_monitoring/stream_processing.py
```diff
@@ -180,14 +180,14 @@ def apply_event_routing():
 
         apply_event_routing()
 
-        # Step 2 - Filter out events with no '-' in path which indicates that the event is supposed to be processed
+        # Step 2 - Filter out events with '-' in path which indicates that the event is supposed to be processed
         # through the next steps of the stream graph
```
assaf758, line 184:
```suggestion
        # Step 2 - Filter out events with '-' in the path basename from going forward
        # through the next steps of the stream graph
```

assaf758 (changes requested):

assaf758 (approved):

PR #5067, opened 2024-02-05
[Tests] Use a unique name for the monitoring app sys test

jond01:
A workaround for [ML-5588](https://jira.iguazeng.com/browse/ML-5588). Add a datetime-based suffix to the project name for uniqueness.

I checked it locally twice and it passed.


Eyal-Danieli (approved):
Great

PR #5068, opened 2024-02-05
[Model Monitoring] Saving feature_stats as empty dict and not as null in the model_endpoint

davesh0812:
[ML-5673](https://jira.iguazeng.com/browse/ML-5673)


Eyal-Danieli (approved):
nice nice nice

PR #5072, opened 2024-02-06
[Project] Fix DELETE Project not returning 412 when Nuclio Functions Exist

quaark:
https://jira.iguazeng.com/browse/ML-5671

Before the background task refactor, the delete request would pass mlrun validations and then be forwarded to Iguazio, where zebo would validate against nuclio. Now as the request to Iguazio is nested within the background task, the validation against nuclio is also nested in the task.
In order to perform the validation before spawning the BG task, mlrun must perform the deletion check manually against nuclio (regardless of its leader/follower status) as Iguazio only supports `restricted` and `cascading` (not `check`) and we want this fix to work with old Iguazio versions as well.


server/api/db/sqldb/db.py
```diff
@@ -2417,6 +2418,12 @@ def verify_project_has_no_related_resources(self, session: Session, name: str):
             name, feature_vectors, "feature_vectors"
         )
 
+        # verify project can be deleted in nuclio
```
alonmr, line 2421: I think `_verify_project_has_no_external_resources` is a more suiting place for this check

alonmr (approved):

PR #5084, opened 2024-02-07
[CI] Extend PR title checker for side branches

alonmr:
If the PR is not for development, ensure that the title ends with  `[<branch-name>]`


liranbg (changes requested):
I think that, instead of validation that, I would suggest doing a bot that align the pr title when it is a pr against a side release branch (x.y.z) because I would not expect people to put `feature/x-y-z` on pr title

PR #5090, opened 2024-02-08
[Artifacts] Add Index artifact key and query tags with new session

TomerShor:
Following adding the locks when tagging artifacts in https://github.com/mlrun/mlrun/pull/5019, we have seen that the locks was not enough to prevent multiple processes from creating new tags for artifacts with the same key+project+producer_id+iteration, causing multiple tags for artifacts with the same key.

After some serious investigation, we have realized that even with the locking, the different parallel requests are not aware of 
tags that the other requests have just created.

To fix this, we introduced 2 more fixes:
1. Add indexes to the ArtifactV2.key column - We saw that selecting the artifacts with "FOR UPDATE" does indeed lock the records, but also adds some locks on all of the other artifact records in the table. This is due to the locking done with filtering on the artifacts' keys, which are not indexed. The DB iterates over ALL of the artifacts to locate the ones with the provided key. Adding an index allows locking only the selected records without actually going over the entire table.
2. Query existing tags with a new session - Right before creating a new tag, we check if a tag already exists. If so, we update its `obj_id` with the artifact id at hand. Since multiple parallel sessions couldn't find the newly created artifacts, querying them with a new session acts as a "refresh", and allows finding all of the new tags in the DB.

Resolves https://jira.iguazeng.com/browse/ML-5596


server/api/db/sqldb/db.py
```diff
@@ -899,15 +900,30 @@ def tag_artifacts(
         artifacts,
         project: str,
     ):
+        artifacts_keys = [artifact.key for artifact in artifacts]
+        logger.debug(
+            "Locking artifacts in db before tagging artifacts",
+            project=project,
+            tag=tag_name,
+            artifacts_keys=artifacts_keys,
+        )
```
theSaarco, line 909: I know this is debug log, but it still is too verbose for production code...

server/api/db/sqldb/db.py
```diff
@@ -971,6 +993,13 @@ def tag_artifacts(
         # this will also release the locks on the artifacts' rows
         self._commit(session, objects)
 
+        logger.debug(
+            "Locking artifacts in before tagging artifacts",
+            project=project,
+            tag=tag_name,
+            artifacts_keys=artifacts_keys,
+        )
```
theSaarco, line 1001: Same.

theSaarco (approved):
Amazing ❤️ 
I think the debug logs should not remain once we figured out the issue.
Also, is it possible to generate a unit-test (or integration-test) that will simulate this situation? I know it's kinda complex, since we need parallel invocations so it needs a thread-pool or whatever, but it can serve as a great framework for testing similar flows for other DB objects in the future. We are time-constrained now, but a future PR would be great.

server/api/db/sqldb/db.py
TomerShor: @theSaarco Since we are locking the DB I think it will be beneficial for us to leave these logs here.

server/api/db/sqldb/db.py
TomerShor: same as above, but also fixed the log message

PR #5096, opened 2024-02-08
[Project] Wait for project to be deleted in Nuclio when deleting project

TomerShor:
When deleting a project, deleting model endpoint resources deletes a directory that contains files that can be used by Nuclio functions - specifically function with V3IO Stream trigger.
The client that consumes from the V3IO stream, uses a state file to store information about the stream consumption. This state file is stored in the directory mentioned above.

There are cases where MLRrun deletes the directory before the Nuclio function is deleted. When the stream client sees that its state file doesn't exist - it creates it again, essentially recreates the previously deleted directory.
When the function is eventually deleted, it does not delete this state file or its parent directories (by design).

This can cause conflicts later when mlrun tries to create this directory for a **new project of the same name**, but there is already an existing directory with files.

To overcome this, when we delete a project - before deleting model endpoint resources, wait for the project (including its functions) is deleted in Nuclio.

**Note:** Although an async nuclio client already exists, I decided to extend the sync nuclio client `get_project` functionality to minimize the code changes using the async client will require (risky this late in the 1.6.x game).

Resolves https://jira.iguazeng.com/browse/ML-5588

### DIFF
--- a/server/api/crud/projects.py
+++ b/server/api/crud/projects.py
@@ -31,7 +31,7 @@ import server.api.utils.events.events_factory as events_factory
 import server.api.utils.projects.remotes.follower as project_follower
 import server.api.utils.singletons.db
 import server.api.utils.singletons.scheduler
-from mlrun.utils import logger
+from mlrun.utils import logger, retry_until_successful
 from server.api.utils.singletons.k8s import get_k8s_helper
 
 
@@ -160,6 +160,7 @@ class Projects(
         self,
         session: sqlalchemy.orm.Session,
         name: str,
+        auth_info: mlrun.common.schemas.AuthInfo = mlrun.common.schemas.AuthInfo(),
     ):
         # Delete schedules before runtime resources - otherwise they will keep getting created
         server.api.utils.singletons.scheduler.get_scheduler().delete_schedules(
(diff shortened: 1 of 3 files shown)


server/api/crud/projects.py
```diff
@@ -384,3 +388,46 @@ async def _calculate_pipelines_counters(
...
+                nuclio_client.get_project(session, project_name)
+                raise Exception(
+                    f"Project not deleted in nuclio yet. Project: {project_name}"
+                )
+            except mlrun.errors.MLRunNotFoundError:
+                logger.debug(
+                    "Nuclio project deleted",
+                    project_name=project_name,
+                )
+                return True
```
quaark, line 414:
```suggestion
            try:
                nuclio_client.get_project(session, project_name)
            except mlrun.errors.MLRunNotFoundError:
                logger.debug(
                    "Nuclio project deleted",
                    project_name=project_name,
                )
                return True
            else:
                raise Exception(
                    f"Project not deleted in nuclio yet. Project: {project_name}"
                )
```
Better to raise outside of the try

quaark (changes requested):
Looks good!
One minor thing

liranbg (approved):

liranbg (approved):

PR #5104, opened 2024-02-11
[API] Check background task state on project deletion wait [1.6.x]

alonmr:
https://jira.iguazeng.com/browse/ML-5701

Fail fast when the background task failed instead of waiting for 30 mins for nothing.

### DIFF
--- a/server/api/api/utils.py
+++ b/server/api/api/utils.py
@@ -1186,6 +1186,12 @@ async def _delete_project(
 
 
 def verify_project_is_deleted(project_name, auth_info):
+    background_task_name = (
+        server.api.utils.background_tasks.BackgroundTaskKinds.project_deletion.format(
+            project_name
+        )
+    )
+
     def _verify_project_is_deleted():
         try:
             server.api.db.session.run_function_with_new_db_session(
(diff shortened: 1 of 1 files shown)


quaark (approved):
Looks Good!

tests/api/api/test_projects.py
```diff
@@ -970,15 +971,80 @@ def test_delete_project_not_found_in_leader(
...
+    ],
+)
+def test_delete_project_fail_fast(
+    unversioned_client: TestClient,
+    mock_project_follower_iguazio_client,
+    delete_api_version: str,
+) -> None:
+    # Set the igz version for the project leader mock
+    # We only test igz version < 3.5.5 flow because from 3.5.5 iguazio waits for the inner background task to
+    # finish so the wrapper task does not wait for the inner task
```
TomerShor, line 1007: isn't the fail fast what you are testing? so it happens only for 3.5.5, no?

alonmr: We only wait for the project deletion on iguazio < 3.5.5.
From iguazio 3.5.5 zebo waits for the inner bg task so the wrapper task waits on the zebo job and we know that if the zebo job succeeded/failed then the inner bg task finished. Therfore there is nothing to test on igz 3.5.5 because the flow doesn't reach the wait for deletion part.

PR #5106, opened 2024-02-11
[Project] Ensure nuclio function pods are deleted

liranbg:
As written on code, nuclio first delete CRDs, then its controller is triggered to delete the runtime resources. In order to ensure project directory is safely deleted, mlrun needs for nuclio resources to get deleted before it continues

https://jira.iguazeng.com/browse/ML-5588


alonmr (approved):
lgtm

PR #5109, opened 2024-02-11
[Model Monitoring] Improve the error messages in drift table creation

jond01:
Fixes [ML-5265](https://jira.iguazeng.com/browse/ML-5265).

This change covers two cases:
- Extra features in the inputs (current data), as in the original issue.
- Missing keys, e.g. "mean", from existing features in the inputs. This happens when there's a placeholder for a label, but a new label name is used.

### DIFF
--- a/mlrun/model_monitoring/features_drift_table.py
+++ b/mlrun/model_monitoring/features_drift_table.py
@@ -93,7 +93,6 @@ class FeaturesDriftTablePlot:
 
     def produce(
         self,
-        features: list[str],
         sample_set_statistics: dict,
         inputs_statistics: dict,
         metrics: dict[str, Union[dict, float]],
@@ -113,7 +112,7 @@ class FeaturesDriftTablePlot:
         """
         # Plot the drift table:
         figure = self._plot(
-            features=features,
+            features=list(inputs_statistics.keys()),
             sample_set_statistics=sample_set_statistics,
             inputs_statistics=inputs_statistics,
             metrics=metrics,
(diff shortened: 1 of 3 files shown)


mlrun/model_monitoring/features_drift_table.py
```diff
@@ -93,7 +93,6 @@ def __init__(self):
 
     def produce(
         self,
-        features: list[str],
```
Eyal-Danieli, line 96: remove `features` from the docstring as well

mlrun/model_monitoring/features_drift_table.py
```diff
@@ -279,6 +278,7 @@ def _plot_feature_row_table(
         input_statistics: dict,
         metrics: dict,
         row_height: int,
+        feature: str,
```
Eyal-Danieli, line 281: add it to the docstring to keep consistency with the other function params

Eyal-Danieli: actually why do you prefer to pass `feature` here if you already have `feature_name`? perhaps it's better to pass only `feature_name` and change the variable name when parsing it into html format. something like 
`parsed_feature_name = "<br>".join(self._separate_feature_name(feature_name))`

Eyal-Danieli (commented):
nice
few little comments/suggestions

mlrun/model_monitoring/features_drift_table.py
jond01: a79ba82

mlrun/model_monitoring/features_drift_table.py
jond01: Right, great catch! 😅
Done in c6f8e0d.

Eyal-Danieli (approved):

PR #5111, opened 2024-02-12
[Kafka] Default `max_workers` to 1 instead of 4

gtopper:
[ML-5693](https://jira.iguazeng.com/browse/ML-5693)


mlrun/runtimes/nuclio/serving.py
```diff
@@ -496,10 +496,6 @@ def _add_ref_triggers(self):
                     if brokers:
                         brokers = brokers.split(",")
                     topic, brokers = parse_kafka_url(stream.path, brokers)
-                    max_workers_default = 4
```
assaf758, line 499: why not change default to 1?

gtopper: 1 is the default in nuclio-jupyter. We could repeat it here if we want.

assaf758 (approved):

PR #5115, opened 2024-02-12
[Artifacts] Fix import artifact bug

tomerm-iguazio:
1) Add item.target_path condition.
2) Add regression test.

[ML-5702](https://jira.iguazeng.com/browse/ML-5702)


theSaarco (approved):
LGTM.

PR #5116, opened 2024-02-12
[Project] Fix Project Deletion Failure due to Project Sync Race Condition

quaark:
Make project sync skip any projects that are currently being deleted by inspecting the existing background tasks.
Also we realize project sync does not need to run on all workers anymore, so we can change it to a chief only operation and rely on the in-memory state of the internal background tasks.


server/api/utils/projects/follower.py
```diff
@@ -47,6 +48,10 @@ class Member(
 ):
     def initialize(self):
         logger.info("Initializing projects follower")
+        self._chief = (
```
alonmr, line 51:
```suggestion
        self._is_chief = (
```

server/api/utils/projects/follower.py
```diff
@@ -66,32 +71,24 @@ def initialize(self):
...
-                "Initial projects sync failed",
-                exc=err_to_str(exc),
-                traceback=traceback.format_exc(),
-            )
-        self._start_periodic_sync()
+        if self._chief:
+            try:
+                # full_sync=True was a temporary measure to handle the move of mlrun from single instance to
+                # chief-worker model.
+                # TODO: remove full_sync=True in 1.7.0 if no issues arise
```
alonmr, line 78: we do full_sync only on initialize and not on periodic so you will still need it here even of you don't sync in workers

alonmr (commented):
looks good! minor comments

server/api/utils/projects/follower.py
alonmr: discussed offline

liranbg (changes requested):
overall, looking good to me. add some UT to make sure no sync may occur during project deletion.

liranbg (approved):

PR #5132, opened 2024-02-13
[Docs] Improve feature store documentation; sources and targets; partitioning

jillnogold:
[ML-3386](https://jira.iguazeng.com/browse/ML-3386), [ML-5680](https://jira.iguazeng.com/browse/ML-5680), [ML-5602](https://jira.iguazeng.com/browse/ML-5602)


docs/feature-store/sources-targets.md
```diff
@@ -0,0 +1,47 @@
+(sources-targets)=
+# Sources and targets
```
assaf758, line 2: Since sources and targets are so central to the feature-store, I would:
1. remove the Sources and Targets H1 heading, and instead gives sources and targets a H1 heading
2. Move  "Partitioning on Parquet target" into "Parquet target" section, where partitioning will be discussed (maybe we will add more special info on other targets as the time goes)

docs/feature-store/sources-targets.md
```diff
@@ -0,0 +1,47 @@
...
+- [Sources](#sources)
+- [Targets](#targets)
+- [Partitioning on Parquet target](#partitioning-on-parquet-target)
+
+
+
+## Sources
+| Class name                                                                                                        | Description                                                                     | storey | spark | pandas |batch or event|
+| --------------------------------------------------                                                                | ---------------------------------                                               | ---    | ---   | ---    |---|
+| [mlrun.datastore.BigQuerySource](../api/mlrun.datastore.html#mlrun.datastore.BigQuerySource)                      | Reads Google BigQuery query results as input source for a flow.                 | N      | Y     | Y      |batch|
```
assaf758, line 14: since batch-or-event is actually an attribute of the source, I think it's misleading to provide this info alongside columns that convey whether a particular ingestion engine support it.
Instead, I would add it to the Description:
`Reads Google BigQuery query results as input source for a flow ("Batch" source)`
`Sets the kafka source for the flow (Event-based source)`

docs/feature-store/sources-targets.md
```diff
@@ -0,0 +1,47 @@
...
+| [mlrun.datastore.StreamSource](../api/mlrun.datastore.html#mlrun.datastore.StreamSource)                          | Sets the stream source for the flow. If the stream doesn’t exist it creates it. | Y      | N     | N      |event|
+
+## Targets
+| Class name                                                                                                    | Description                                                                                        | storey | spark | pandas |Online or offline|
+| --------------------------------------------------                                                            | -------------------------------------------------------                                            | ---    | ---   | ---    |---|
+| [mlrun.datastore.CSVTarget](https://storey.readthedocs.io/en/latest/api.html#storey.targets.CSVTarget)        | Writes events to a CSV file.                                                                       | Y      | Y     | Y      |offline|
```
assaf758, line 27: Same here, offline/online is an attribute of the target unrelated to whether a particular ingestion engine support it.
Suggest to add to the Description column.

docs/feature-store/sources-targets.md
```diff
@@ -0,0 +1,47 @@
...
+| mlrun.datastore.RedisNoSqlTarget                                                                              | Persists the data in Redis table to its associated storage by key.                                 | Y      | Y     | N      |Online|
+| mlrun.datastore.SqlTarget                                                                                     | Persists the data in SQL table to its associated storage by key.                                   | Y      | N     | Y      |Offline (default)|
+| [mlrun.datastore.ParquetTarget](https://storey.readthedocs.io/en/latest/api.html#storey.targets.ParquetTarget)| The Parquet target storage driver, used to materialize feature set/vector data into parquet files. | Y      | Y     | Y      |Online|
+| [mlrun.datastore.StreamTarget](https://storey.readthedocs.io/en/latest/api.html#storey.targets.StreamTarget)  | Writes all incoming events into a V3IO stream.                                                     | Y      | N     | N      |Offline|
+
+## Partitioning on Parquet target
```
assaf758, line 35: we should mention the 1024 partitions here.

jillnogold: @assaf758 - what are 1024 partitions?

docs/feature-store/sources-targets.md
```diff
@@ -0,0 +1,67 @@
...
+Partitioning, when configured correctly, improves read performance by reducing the amount of data that needs to be 
+processed for any given function, for example, when reading back a limited time range with `get_offline_features()`.
+
+```{note} 
+When partitioning with a pandas engine, the `ParquetTarget(path="my_file.parquet")` writes to a single Parquet file. 
+`ingest()` calls `df.to_parquet()` once, with all the data. A single `df.to_parquet()` call cannot write to more than 1024 partitions.
+```
+storey processes the data row by row (as a streaming engine, it doesn't get all the data up front, so it needs to process row by row). 
+These rows are batched together according to the partitions defined, and they are 
+written to each partition separately. (Therefore, storey does not have the 1024 limitation.)
```
assaf758, line 51: When using the pandas engine for ingestion, pandas incur a maximum limit of 1024 partitions on each ingestion.
If the data being ingested spans over more than 1024 partitions, the ingestion will fail.
Decreasing the number of partitions can be achieved by filtering by time or increasing the `time_partitioning_granularity`

jillnogold: @assaf758 what do you mean by time filtering? Isn't that <time_partitioning_granularity>

assaf758: time-filter is a feature of sources - for example ParquetSource has start_filter/end_filter.
time_partitioning_granularity is the granularity (the time range that each partition holds)

assaf758 (approved):

PR #5145, opened 2024-02-15
[Docs] Remove duplication from generated API docs

gtopper:


assaf758 (approved):

mlrun/datastore/sources.py
```diff
@@ -863,7 +861,7 @@ def __init__(
         **kwargs,
     ):
         """
-        Sets stream source for the flow. If stream doesn't exist it will create it
+        Sets stream source for the flow. If stream doesn't exist it will create it.
```
jillnogold, line 864: Sets the stream source for the flow. If the stream doesn't exist it will create it.

jillnogold (approved):

PR #5170, opened 2024-02-19
[API] Cache Get client-spec

liranbg:
No reason to resolve it every time, enough to simply cache it for 60s per client version matrix.
the "ttl_seconds" is simply a way to invalidate cache based on time

### DIFF
--- a/server/api/api/utils.py
+++ b/server/api/api/utils.py
@@ -14,6 +14,7 @@
 #
 import collections
 import copy
+import functools
 import json
 import re
 import traceback
@@ -62,6 +63,11 @@ from server.api.utils.singletons.scheduler import get_scheduler
 def log_and_raise(status=HTTPStatus.BAD_REQUEST.value, **kw):
     logger.error(str(kw))
     raise HTTPException(status_code=status, detail=kw)
+
+
+@functools.lru_cache
+def expiring_lru_cache(ttl_seconds: int, func: typing.Callable, *args, **kwargs):
+    return func(*args, **kwargs)
 
 
 def log_path(project, uid) -> Path:
--- a/tests/api/api/test_client_spec.py
+++ b/tests/api/api/test_client_spec.py
@@ -26,6 +26,7 @@ import mlrun.common.schemas
 import mlrun.errors
 import mlrun.runtimes
 import mlrun.utils.version
+import server.api.crud.client_spec
 
 
 def test_client_spec(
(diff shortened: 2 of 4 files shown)


server/api/api/utils.py
```diff
@@ -64,6 +65,18 @@ def log_and_raise(status=HTTPStatus.BAD_REQUEST.value, **kw):
     raise HTTPException(status_code=status, detail=kw)
 
 
+@functools.lru_cache
```
theSaarco, line 68: I would add a relatively small `maxsize` here (16 or something like that), otherwise the cache would inflate to 1440 entities (# of minutes in a day), due to using `ttl_seconds` as parameter, each holding a full `ClientSpec` in it.
Also, it's confusing to call the param `ttl_seconds` since it actually represents the minutes, not how many seconds to live. As a matter of fact, each cached value will last a full day given the current implementation.

tests/api/api/test_client_spec.py
```diff
@@ -217,3 +218,20 @@ def test_client_spec_response_based_on_client_version(
...
+        server.api.crud.client_spec.ClientSpec,
+        "get_client_spec",
+        return_value=client_spec,
+    ) as mocked_get_client:
+        response = client.get("client-spec")
+        assert response.status_code == http.HTTPStatus.OK.value
+        cached_response = client.get("client-spec")
+        assert cached_response.status_code == http.HTTPStatus.OK.value
+        assert response.json() == cached_response.json()
+        assert mocked_get_client.call_count == 1
```
theSaarco, line 237: If you're counting calls, it would be better to call the client API several times and still make sure the actual function is only called once.

liranbg: will do

tests/api/api/test_client_spec.py
```diff
@@ -152,6 +156,8 @@ def test_client_spec_response_based_on_client_version(
     assert response_body["kfp_image"] == "mlrun/mlrun:unstable"
     assert response_body["dask_kfp_image"] == "mlrun/ml-base:unstable"
 
+    # clear case for next scenario
```
TomerShor, line 159:
```suggestion
    # clear cash for next scenario
```

TomerShor (approved):
Cool stuff!
Only a minor typo

tests/api/api/test_client_spec.py
theSaarco: Should actually be `clear cache`.

theSaarco (approved):
LGTM

PR #5171, opened 2024-02-20
[Project] Support loading project from files with both `yaml` and `yml` extensions

TomerShor:
If a `project.yml` file exists, there is no reason for `mlrun.load_project` to fail.

As an improvement, support both `yaml` and `yml` extensions for project and function files.


quaark (approved):
🚀

liranbg (approved):

PR #5172, opened 2024-02-20
[K8s] Support email-like usernames

liranbg:
https://jira.iguazeng.com/browse/ML-5721


alonmr (commented):
wrt to labels looks good.
what about user project? i.e. when we concatenate the username to the project name

liranbg:
> wrt to labels looks good. what about user project? i.e. when we concatenate the username to the project name

handled already

PR #5177, opened 2024-02-20
[SQLAlchemy] Recycle & ping connection prior using them

liranbg:
It has been observed that not using mlrun for very long of period forcing mysql to stale the connection, later on discovering it while using connection to perform db operations. this pr ensure sqlalchemy recycle the connection pool in a default of 1 hour + perform a pre-ping prior using the connection to perform db operation

https://jira.iguazeng.com/browse/ML-5772

### DIFF
--- a/mlrun/common/db/sql_session.py
+++ b/mlrun/common/db/sql_session.py
@@ -63,20 +63,11 @@ def _init_engine(dsn=None):
         if max_overflow is None:
             max_overflow = config.httpdb.max_workers
 
-        # tests connections for liveness upon each checkout
-        pool_pre_ping = config.httpdb.db.connections_pool_pre_ping
-        if pool_pre_ping is None:
-            pool_pre_ping = True
-
-        # his setting causes the pool to recycle connections after the given number of seconds has passed
-        pool_recycle = config.httpdb.db.connections_pool_recycle
-        if pool_recycle is None:
-            pool_recycle = 60 * 60
         kwargs = {
             "pool_size": pool_size,
             "max_overflow": max_overflow,
-            "pool_pre_ping": pool_pre_ping,
-            "pool_recycle": pool_recycle,
+            "pool_pre_ping": config.httpdb.db.connections_pool_pre_ping,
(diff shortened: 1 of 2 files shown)


mlrun/common/db/sql_session.py
```diff
@@ -62,9 +62,21 @@ def _init_engine(dsn=None):
         max_overflow = config.httpdb.db.connections_pool_max_overflow
         if max_overflow is None:
             max_overflow = config.httpdb.max_workers
+
+        # tests connections for liveness upon each checkout
+        pool_pre_ping = config.httpdb.db.connections_pool_pre_ping
+        if pool_pre_ping is None:
+            pool_pre_ping = True
+
+        # his setting causes the pool to recycle connections after the given number of seconds has passed
```
theSaarco, line 71:
```suggestion
        # This setting causes the pool to recycle connections after the given number of seconds has passed
```

mlrun/common/db/sql_session.py
```diff
@@ -62,9 +62,21 @@ def _init_engine(dsn=None):
         max_overflow = config.httpdb.db.connections_pool_max_overflow
         if max_overflow is None:
             max_overflow = config.httpdb.max_workers
+
+        # tests connections for liveness upon each checkout
+        pool_pre_ping = config.httpdb.db.connections_pool_pre_ping
+        if pool_pre_ping is None:
+            pool_pre_ping = True
```
theSaarco, line 69: Why not set this value already in the config?

mlrun/common/db/sql_session.py
```diff
@@ -62,9 +62,21 @@ def _init_engine(dsn=None):
...
+
+        # his setting causes the pool to recycle connections after the given number of seconds has passed
+        pool_recycle = config.httpdb.db.connections_pool_recycle
+        if pool_recycle is None:
+            pool_recycle = 60 * 60
```
theSaarco, line 74: Same - why not in the config?

theSaarco (approved):
Looks fine, I would prefer some small changes - see comments.

PR #5181, opened 2024-02-20
[Artifacts] Add validation to oversized inline artifact store

yaelgen:
resolves: [ML-3889](https://jira.iguazeng.com/browse/ML-3889)

### DIFF
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -271,9 +271,8 @@ def validate_artifact_key_name(
     )
 
 
-def validate_inline_artifact_body_size(body: str) -> None:
-    body_length = len(body.encode("utf-8"))
-    if body_length > MYSQL_MEDIUMBLOB_BYTES:
+def validate_inline_artifact_body_size(body: typing.Union[str, bytes]) -> None:
+    if len(body) > MYSQL_MEDIUMBLOB_BYTES:
         raise mlrun.errors.MLRunPreconditionFailedError(
             "The body of the artifact exceeds the maximum allowed size. "
             "Please reduce the size of the artifact body and try again."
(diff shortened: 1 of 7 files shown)


mlrun/utils/helpers.py
```diff
@@ -270,6 +271,15 @@ def validate_artifact_key_name(
     )
 
 
+def validate_inline_artifact_body_size(body: str) -> None:
+    body_length = len(body.encode("utf-8"))
+    if body_length > MYSQL_MEDIUMBLOB_BYTES:
+        raise mlrun.errors.MLRunPreconditionFailedError(
+            "The body of the artifact exceeds the maximum allowed size. "
+            "Please reduce the size of the artifact body and try again."
+        )
```
liranbg, line 280: I would go with BadRequest
and I think we should discourage using inline artifacts anyway and instead of asking to reduce, perhaps encourage the clients to not used embed artifacts.

```suggestion
        raise mlrun.errors.MLRunBadRequestError(
            "The body of the artifact exceeds the maximum allowed size. "
            "Avoid embedding the artifact body. This increases the size of the project yaml file and could affect the project during loading and saving. More information is available at https://docs.mlrun.org/en/latest/projects/automate-project-git-source.html#setting-and-registering-the-project-artifacts"
        )
```

liranbg (changes requested):
very well and neat job!
Id stretch the tests and maybe change the logic a bit because artifact body may not be string but perhaps byte string or any bytes ("\xf0\x9f\x98\x86" / "\xf0") that may  not be utf8 encodable.

TomerShor (approved):
Approved with a question

liranbg (approved):

PR #5183, opened 2024-02-21
[Nuclio] Disable project creation on deploy

alonmr:
If the project doesn't exist than something else went wrong.
When mlrun is not leader, nuclio will go to the leader to create the project but the project already exists there so it will fail.


TomerShor (approved):
Nice touch

PR #5186, opened 2024-02-22
[Datastore] Add HDFS Datastore

gtopper:
[ML-5726](https://jira.iguazeng.com/browse/ML-5726)

### DIFF
--- a/mlrun/datastore/base.py
+++ b/mlrun/datastore/base.py
@@ -147,6 +147,10 @@ class DataStore:
     def url(self):
         return f"{self.kind}://{self.endpoint}"
 
+    @property
+    def spark_url(self):
+        return self.url
+
     def get(self, key, size=None, offset=0):
         pass
 
@@ -320,31 +324,17 @@ class DataStore:
             raise Exception(f"File type unhandled {url}")
 
         if file_system:
-            if (
-                self.supports_isdir()
-                and file_system.isdir(file_url)
-                or self._is_dd(df_module)
-            ):
-                storage_options = self.get_storage_options()
-                if url.startswith("ds://"):
-                    parsed_url = urllib.parse.urlparse(url)
-                    url = parsed_url.path
-                    if self.using_bucket:
-                        url = url[1:]
-                    # Pass the underlying file system
-                    kwargs["filesystem"] = file_system
-                elif storage_options:
-                    kwargs["storage_options"] = storage_options
-                df = reader(url, **kwargs)
-            else:
-                file = url
-                # Workaround for ARROW-12472 affecting pyarrow 3.x and 4.x.
-                if file_system.protocol != "file":
-                    # If not dir, use file_system.open() to avoid regression when pandas < 1.2 and does not
-                    # support the storage_options parameter.
-                    file = file_system.open(url)
-
-                df = reader(file, **kwargs)
+            storage_options = self.get_storage_options()
+            if url.startswith("ds://"):
+                parsed_url = urllib.parse.urlparse(url)
+                url = parsed_url.path
+                if self.using_bucket:
+                    url = url[1:]
+                # Pass the underlying file system
+                kwargs["filesystem"] = file_system
+            elif storage_options:
+                kwargs["storage_options"] = storage_options
+            df = reader(url, **kwargs)
         else:
             temp_file = tempfile.NamedTemporaryFile(delete=False)
             self.download(self._join(subpath), temp_file.name)
--- a/mlrun/datastore/datastore.py
+++ b/mlrun/datastore/datastore.py
@@ -94,6 +94,10 @@ def schema_to_store(schema):
         from .dbfs_store import DBFSStore
 
         return DBFSStore
+    elif schema == "hdfs":
+        from .hdfs import HdfsStore
+
+        return HdfsStore
     else:
         raise ValueError(f"unsupported store scheme ({schema})")
 
@@ -185,7 +189,14 @@ class StoreManager:
         store, subpath = self.get_or_create_store(
             url, secrets=secrets, project_name=project
         )
-        return DataItem(key, store, subpath, url, meta=meta, artifact_url=artifact_url)
+        return DataItem(
+            key,
+            store,
+            subpath,
+            store.url + subpath,
+            meta=meta,
+            artifact_url=artifact_url,
+        )
 
     def get_or_create_store(
         self, url, secrets: dict = None, project_name=""
--- a/mlrun/datastore/datastore_profile.py
+++ b/mlrun/datastore/datastore_profile.py
@@ -295,6 +295,30 @@ class DatastoreProfileAzureBlob(DatastoreProfile):
         return res if res else None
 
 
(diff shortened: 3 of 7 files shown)


mlrun/datastore/base.py
```diff
@@ -147,6 +147,10 @@ def _get_secret(self, key: str, default=None):
     def url(self):
         return f"{self.kind}://{self.endpoint}"
 
+    @property
+    def spark_url(self):
+        return self.url
+
```
alxtkr77, line 153: Looks like redundant to mlrun.datastore.utis.store_path_to_spark(). No need to introduce a new API in my opinion

mlrun/datastore/base.py
```diff
@@ -320,31 +324,17 @@ def reader(*args, **kwargs):
...
-            else:
-                file = url
-                # Workaround for ARROW-12472 affecting pyarrow 3.x and 4.x.
-                if file_system.protocol != "file":
-                    # If not dir, use file_system.open() to avoid regression when pandas < 1.2 and does not
-                    # support the storage_options parameter.
-                    file = file_system.open(url)
-
-                df = reader(file, **kwargs)
+            storage_options = self.get_storage_options()
```
alxtkr77, line 327: no special care for files anymore?

mlrun/datastore/base.py
gtopper: The suggestion being that we add a case to replace `webhdfs://` with `hdfs://` in `store_path_to_spark()`?

mlrun/datastore/base.py
gtopper: Nope, the whole case [was added for an issue with pandas<1.2](https://github.com/mlrun/mlrun/pull/1049/files#diff-e6c49fcaee9a2be90cbb9d67f1efa5c48fa7e9bd008732d4cad59ee6a57d9ccaR213), which we no longer support.

mlrun/datastore/datastore.py
```diff
@@ -185,7 +189,14 @@ def object(
         store, subpath = self.get_or_create_store(
             url, secrets=secrets, project_name=project
         )
-        return DataItem(key, store, subpath, url, meta=meta, artifact_url=artifact_url)
+        return DataItem(
+            key,
+            store,
+            subpath,
+            store.url + subpath,
```
alxtkr77, line 196: Are you sure all the stores has been tested with this change? I would use 'store.url + subpath if url and url.startswith("ds://") else url' - if the purpose is it to handle datastore profiles

mlrun/datastore/sources.py
```diff
@@ -377,7 +377,7 @@ def get_spark_options(self):
         if self.path and self.path.startswith("ds://"):
             store, path = mlrun.store_manager.get_or_create_store(self.path)
             storage_spark_options = store.get_spark_options()
-            path = store.url + path
+            path = store.spark_url + path
```
alxtkr77, line 380: redundant to store_path_to_spark

mlrun/datastore/targets.py
```diff
@@ -930,7 +930,7 @@ def get_spark_options(self, key_column=None, timestamp_key=None, overwrite=True)
                 self.get_target_path()
             )
             storage_spark_options = store.get_spark_options()
-            path = store.url + path
+            path = store.spark_url + path
```
alxtkr77, line 933: Makes no sense. You later pass path to store_path_to_spark(), so this function should build the spark_url basing on the path and the storage_options

tests/system/feature_store/test_spark_engine.py
```diff
@@ -80,7 +81,7 @@ class TestFeatureStoreSparkEngine(TestMLRunSystem):
     pq_source = "testdata.parquet"
     pq_target = "testdata_target"
     csv_source = "testdata.csv"
-    run_local = False
+    run_local = True
```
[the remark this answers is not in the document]
gtopper: Nope.

mlrun/datastore/base.py
gtopper: That can be done, but the issue is the different port.

mlrun/datastore/targets.py
gtopper: Resolving duplicate comment.

mlrun/datastore/targets.py
```diff
@@ -1072,7 +1072,7 @@ def get_spark_options(self, key_column=None, timestamp_key=None, overwrite=True)
...
```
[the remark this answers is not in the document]
gtopper: Resolving duplicate comment.

mlrun/datastore/datastore_profile.py
```diff
@@ -295,6 +295,30 @@ def secrets(self) -> dict:
...
+            res["HDFS_HOST"] = self.host
+        if self.port:
+            res["HDFS_PORT"] = self.port
+        if self.port:
+            res["HDFS_HTTP_PORT"] = self.http_port
+        if self.user:
+            res["HDFS_USER"] = self.user
+        return res or None
+
+    def url(self, subpath):
```
alxtkr77, line 318: You ether provide port and host as part of the url , or as part of the storage parameters in fsspec. I think url should return "webhdfs://' or 'webhdfs:/'

mlrun/datastore/base.py
alxtkr77: This is should be like in S3: You work with url without the hostname/port :
val df = spark.read.text("hdfs:///user/hadoop/kuku.txt")

BUT: in the spark options you add {"spark.hadoop.fs.defaultFS":"hdfs://localhost:9000"}

mlrun/datastore/sources.py
```diff
@@ -209,25 +208,17 @@ def to_step(self, key_field=None, time_field=None, context=None):
...
 
     def get_spark_options(self):
+        store, path = mlrun.store_manager.get_or_create_store(self.path)
+        result = {
+            "path": store.spark_url + path,
+            "format": "csv",
+            "header": "true",
+            "inferSchema": "true",
+        }
         if self.path and self.path.startswith("ds://"):
```
alxtkr77, line 218: I don't think we need this 'ds' special case anymore. We should always return {**result, **storage_spark_options}. The same for all the further cases

mlrun/datastore/redis.py
```diff
@@ -163,3 +163,7 @@ def rm(self, key, recursive=False, maxdepth=None):
                 self.redis.delete(k)
         else:
             self.redis.delete(key)
+
+    @property
+    def spark_url(self):
+        return ""
```
alxtkr77, line 169: Doesn't look right - should be same as url()

gtopper: `store_path_to_spark()` does this:
```python
schemas = ["redis://", "rediss://", "ds://"]
if any(path.startswith(schema) for schema in schemas):
    url = urlparse(path)
    if url.path:
        path = url.path
```
If you simplify that, it's just `""` (since the path is appended to the result by the caller).

alxtkr77 (approved):

PR #5197, opened 2024-02-25
[Artifacts] Persist db key when migrating to artifact v2 table

TomerShor:
We noticed that when migrating the artifacts table to the new v2 table, the new artifacts are saved with the `key` taken from `metadata.key` only.
However, some artifacts have different `db_keys`, meaning the key they are saved with in the DB can be different then the value in `metadata.key`.

Since we want to be backwards compatible, we want to persist the original key in the DB, **with a caveat**:
Artifacts from a hyper param run, with iterations, had a key in the DB that was a concatenation `{iter}-{key}`. We do not persist this behavior, but remove the `iter` part of the key, and save the iteration in a different column.
This should not affect reading/listing artifacts with iterations, as this concatenation was done behind the scenes.

Resolves https://jira.iguazeng.com/browse/ML-5799

### DIFF
--- a/server/api/initial_data.py
+++ b/server/api/initial_data.py
@@ -581,9 +581,8 @@ def _migrate_artifacts_batch(
             or artifact_dict.get("spec", {}).get("db_key")
             or artifact_metadata.get("key", "")
         )
-        key_parts = key.split("-")
-        if len(key_parts) > 1 and iteration and key_parts[0] == str(iteration):
-            key = "-".join(key_parts[1:])
+        if iteration and key.startswith(f"{iteration}-"):
+            key = key[len(f"{iteration}-") :]
         new_artifact.key = key
 
         # best iteration
--- a/server/api/initial_data.py
+++ b/server/api/initial_data.py
@@ -575,7 +575,9 @@ def _migrate_artifacts_batch(
         iteration = artifact_metadata.get("iter", None)
         new_artifact.iteration = int(iteration) if iteration else 0
 
-        # key - the artifact's key, without iteration if it is attached to it
+        # key - retain the db key to ensure BC of reading artifacts by the index key.
+        # if iteration is concatenated to the key, remove it as this was only handled in the backend,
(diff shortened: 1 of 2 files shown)


server/api/initial_data.py
```diff
@@ -571,14 +571,20 @@ def _migrate_artifacts_batch(
...
-        # key - the artifact's key, without iteration if it is attached to it
-        key = artifact_metadata.get("key", "")
-        new_artifact.key = key
-
         # iteration - the artifact's iteration
         iteration = artifact_metadata.get("iter", None)
         new_artifact.iteration = int(iteration) if iteration else 0
 
+        # key - the artifact's key, without iteration if it is attached to it
+        key = (
```
alonmr, line 579: Does it mean that we will never be able to remove the db_key from the spec?
I don't really understand what was the issue. The query is built in the API and users only use `metadata.key` don't they? The backend knows that the key was changed so it should be able to make the adjustment.

server/api/initial_data.py
```diff
@@ -571,14 +571,20 @@ def _migrate_artifacts_batch(
...
+            artifact.key
+            or artifact_dict.get("spec", {}).get("db_key")
+            or artifact_metadata.get("key", "")
```
theSaarco, line 582: Is this really expected to be meaningful? Won't `artifact.key` always have a value, since it's the `key` from the DB table?

TomerShor: I guess not, we will also have `artifact.key`

server/api/initial_data.py
TomerShor: The problem is that the key isn't the Metadata can be different than the db_key.
When reading/listing artifacts, you can only query on the key that is saved in the DB -> `db_key`.

So if users used to query on the DB key, after the migration they wouldn't be able to.

Even in the new API, the BE doesn't auto-resolve the two options, it can only filter on the key column (the rest is kept in the full object).

alonmr (approved):

PR #5198, opened 2024-02-26
[Serving] Allow hundreds of models per serving function

yanburman:
ML-4248


mlrun/serving/server.py
```diff
@@ -312,9 +312,20 @@ def wait_for_completion(self):
...
-    data = os.environ.get("SERVING_SPEC_ENV", "")
-    if not data:
-        raise MLRunInvalidArgumentError("failed to find spec env var")
+    data = None
+
+    try:
+        with open("/model-conf/serving_spec.json") as f:
+            data = f.read()
+    except FileNotFoundError:
+        pass
```
liranbg, line 321: comment / explain the scenario here

server/api/crud/runtimes/nuclio/function.py
```diff
@@ -248,6 +248,24 @@ def _compile_function_config(
     _resolve_and_set_base_image(function, config, client_version, client_python_version)
     function_name = _set_function_name(function, config, project, tag)
 
+    serving_spec = function._get_serving_spec()
+    if serving_spec is not None:
+        confmap_name = "model-conf-" + function_name
```
liranbg, line 253: this is error prone as function name + your prefix might exceed the k8s name limitation of 63 characters. 
I suggest moving this entire "system-generation" resources under k8s_helper functionality so we can later have a single place (class) to see all resources created by mlrun for mlrun purposes
the function (create_or_update) would receive config-map data and name (that we will truncate upon need)
the function would return the generated config map name, enrich it with a label that includes its original name.

liranbg, line 253: this is error prone as function name + your prefix might exceed the k8s name limitation of 63 characters. 
I suggest moving this entire "system-generation" resources under k8s_helper functionality so we can later have a single place (class) to see all resources created by mlrun for mlrun purposes
the function (create_or_update) would receive config-map data and name (that we will truncate upon need)
the function would return the generated config map name, enrich it with a label that includes its original name.

server/api/utils/singletons/k8s.py
```diff
@@ -540,6 +540,49 @@ def delete_secrets(
         self.v1api.delete_namespaced_secret(secret_name, namespace)
         return mlrun.common.schemas.SecretEventActions.deleted
 
+    def create_confmap(self, name: str, data: dict, namespace="", labels=None):
```
liranbg, line 543:
```suggestion
    def create_configmap(self, name: str, data: dict, namespace="", labels=None):
```

server/api/utils/singletons/k8s.py
```diff
@@ -540,6 +540,49 @@ def delete_secrets(
...
+        except ApiException as exc:
+            if exc.status == 409:
+                try:
+                    self.v1api.replace_namespaced_config_map(
+                        name, namespace=namespace, body=body
+                    )
+                except ApiException as exc2:
+                    logger.error(
+                        f"failed to replace k8s ConfigMap {name}: {mlrun.errors.err_to_str(exc2)}"
+                    )
```
liranbg, line 564: use structured log and not templated and use capital letter for first word
```suggestion
                    logger.error(
                        "Failed replacing k8s config map", name=name, exc=mlrun.errors.err_to_str(exc2)"
                    )
```

server/api/utils/singletons/k8s.py
```diff
@@ -540,6 +540,49 @@ def delete_secrets(
...
+
+        body = client.V1ConfigMap(
+            api_version="v1",
+            kind="ConfigMap",
+            metadata=client.V1ObjectMeta(name=name, labels=labels),
+            data=data,
+        )
+
+        try:
+            self.v1api.create_namespaced_config_map(namespace=namespace, body=body)
```
liranbg, line 554: I suggest doing `"get" / "create" | "update"`

server/api/utils/singletons/k8s.py
```diff
@@ -540,6 +540,49 @@ def delete_secrets(
...
+        namespace = self.resolve_namespace(namespace)
+
+        try:
+            self.v1api.delete_namespaced_config_map(
+                name=name,
+                namespace=namespace,
+            )
+        except ApiException as exc:
+            logger.error(
+                f"failed to delete k8s ConfigMap {name}: {mlrun.errors.err_to_str(exc)}"
```
liranbg, line 583: see log comment

server/api/utils/singletons/k8s.py
```diff
@@ -540,6 +540,49 @@ def delete_secrets(
...
+                    )
+                except ApiException as exc2:
+                    logger.error(
+                        f"failed to replace k8s ConfigMap {name}: {mlrun.errors.err_to_str(exc2)}"
+                    )
+                    raise exc2
+
+            else:
+                logger.error(
+                    f"failed to create k8s ConfigMap {name}: {mlrun.errors.err_to_str(exc)}"
```
liranbg, line 569: see log comment

liranbg (changes requested):
Overall idea looks ok, some implementation suggestions and comments

server/api/utils/singletons/k8s.py
```diff
@@ -540,6 +540,64 @@ def delete_secrets(
         self.v1api.delete_namespaced_secret(secret_name, namespace)
         return mlrun.common.schemas.SecretEventActions.deleted
 
+    def create_configmap(
```
alonmr: use `raise_for_status_code` or `mlrun.errors.err_for_status_code` to map the exception to an mlrun exception

alonmr (commented):
Looks really good 👍  minor comment. I assume the config map deletion waits for https://jira.iguazeng.com/browse/ML-3432 . We need to make sure we follow on that.

mlrun/serving/server.py
```diff
@@ -308,14 +309,34 @@ def wait_for_completion(self):
         return self.graph.wait_for_completion()
 
 
+def _get_serving_spec():
```
alonmr: private function below public ones

server/api/crud/runtimes/nuclio/function.py
```diff
@@ -183,6 +184,37 @@ def _compile_function_config(
     # resolve env vars before compiling the nuclio spec, as we need to set them in the spec
     env_dict, external_source_env_dict = _resolve_env_vars(function)
 
+    project = function.metadata.project or "default"
```
alonmr:
```suggestion
    project = function.metadata.project or mlrun.mlconf.default_project
```

alonmr (approved):
LGTM

mlrun/serving/server.py
```diff
@@ -311,11 +312,8 @@ def wait_for_completion(self):
 def v2_serving_init(context, namespace=None):
     """hook for nuclio init_context()"""
 
-    data = os.environ.get("SERVING_SPEC_ENV", "")
-    if not data:
-        raise MLRunInvalidArgumentError("failed to find spec env var")
-    spec = json.loads(data)
     context.logger.info("Initializing server from spec")
+    spec = _get_serving_spec()
     server = GraphServer.from_dict(spec)
```
liranbg: I think it worth shoving the `_get_serving_spec` functionality onto GraphServer/ModelObj itself because
1. the feature provided here might serve other runtime that would have overwhelming amount of envvars
2. we need to remove as much code as possible from here, leaving it concise for the user (as he might see it on the function code itself)

server/api/crud/runtimes/nuclio/function.py
```diff
@@ -183,6 +184,37 @@ def _compile_function_config(
     # resolve env vars before compiling the nuclio spec, as we need to set them in the spec
     env_dict, external_source_env_dict = _resolve_env_vars(function)
 
+    project = function.metadata.project or mlrun.mlconf.default_project
+    tag = function.metadata.tag
+
+    serving_spec_volume = None
+    serving_spec = function._get_serving_spec()
+    if serving_spec is not None:
```
liranbg, line 192: when would it be none?

server/api/crud/runtimes/nuclio/function.py
```diff
@@ -183,6 +184,37 @@ def _compile_function_config(
...
+            confmap_name = k8s_helper.create_configmap(
+                mlrun.common.constants.MLRUN_MODEL_CONF,
+                function_name,
+                {"serving_spec.json": serving_spec},
+                labels={mlrun.common.constants.MLRUN_CREATED_LABEL: "true"},
+            )
+            volume_name = mlrun.common.constants.MLRUN_MODEL_CONF
+            volume_mount = {
+                "name": volume_name,
+                "mountPath": f"/opt/{mlrun.common.constants.MLRUN_MODEL_CONF}",
```
liranbg:
```suggestion
                "mountPath": f"/tmp/mlrun/{mlrun.common.constants.MLRUN_MODEL_CONF}",
```

liranbg: I think it works better when running security context

server/api/crud/runtimes/nuclio/function.py
```diff
@@ -248,6 +278,9 @@ def _compile_function_config(
     _resolve_and_set_base_image(function, config, client_version, client_python_version)
     function_name = _set_function_name(function, config, project, tag)
 
+    if serving_spec_volume is not None:
+        mlrun.utils.update_in(config, "spec.volumes", serving_spec_volume, append=True)
```
liranbg, line 284: since you append, will redeploy of same function recreate the configmap over and over again?

server/api/utils/singletons/k8s.py
```diff
@@ -540,6 +540,65 @@ def delete_secrets(
         self.v1api.delete_namespaced_secret(secret_name, namespace)
         return mlrun.common.schemas.SecretEventActions.deleted
 
+    @raise_for_status_code
+    def create_configmap(
```
liranbg:
```suggestion
    def ensure_configmap(
```
or another name, as what you do here is create/replace and not just "create"

server/api/utils/singletons/k8s.py
```diff
@@ -540,6 +540,65 @@ def delete_secrets(
...
+            self.v1api.delete_namespaced_config_map(
+                name=name,
+                namespace=namespace,
+            )
+        except ApiException as exc:
+            logger.error(
+                "Failed to delete k8s config map",
+                name=name,
+                exc=mlrun.errors.err_to_str(exc),
+            )
```
liranbg, line 627: forgot to raise?

liranbg (changes requested):

server/api/utils/singletons/k8s.py
yanburman: what will user do with that error? error in delete is almost never something you can handle

server/api/crud/runtimes/nuclio/function.py
yanburman: no, since code above it just runs over the volumes without looking at what  was there before

server/api/crud/runtimes/nuclio/function.py
yanburman: in case it is just a nuclio function without serving

server/api/crud/runtimes/nuclio/function.py
TomerShor: This path is also used in `_get_serving_spec`, so better move it to a constant that can be reused

server/api/crud/runtimes/nuclio/function.py
```diff
@@ -183,6 +184,37 @@ def _compile_function_config(
     # resolve env vars before compiling the nuclio spec, as we need to set them in the spec
     env_dict, external_source_env_dict = _resolve_env_vars(function)
 
+    project = function.metadata.project or mlrun.mlconf.default_project
+    tag = function.metadata.tag
+
+    serving_spec_volume = None
+    serving_spec = function._get_serving_spec()
+    if serving_spec is not None:
+        if len(serving_spec) >= mlrun.mlconf.httpdb.nuclio.serving_spec_env_cutoff:
```
TomerShor, line 195: Add a comment explaining why we are doing this

server/api/utils/singletons/k8s.py
```diff
@@ -540,6 +540,65 @@ def delete_secrets(
...
+        body = client.V1ConfigMap(
+            kind="ConfigMap",
+            metadata=client.V1ObjectMeta(name=name, labels=labels),
+            data=data,
+        )
+
+        try:
+            self.v1api.create_namespaced_config_map(namespace=namespace, body=body)
+        except ApiException as exc:
+            if exc.status == 409:
```
TomerShor:
```suggestion
            if exc.status == HTTPStatus.CONFLICT.value:
```

TomerShor (changes requested):
minor comments

mlrun/serving/server.py
```diff
@@ -307,15 +308,33 @@ def wait_for_completion(self):
...
+        except FileNotFoundError:
+            # we will have the serving spec in either mounted config map
+            # or env depending on the size of the spec and configuration
+            pass
+
+        if data is None:
+            data = os.environ.get("SERVING_SPEC_ENV", "")
+            if not data:
+                raise MLRunInvalidArgumentError(
+                    "Failed to find spec env var or config file"
```
TomerShor:
```suggestion
                    "Failed to find serving spec in env var or config file"
```

mlrun/serving/server.py
```diff
@@ -307,15 +308,33 @@ def wait_for_completion(self):
...
+    @staticmethod
+    def get_serving_spec():
+        data = None
+
+        try:
+            with open(mlrun.common.constants.MLRUN_SERVING_SPEC_PATH) as f:
+                data = f.read()
+        except FileNotFoundError:
+            # we will have the serving spec in either mounted config map
+            # or env depending on the size of the spec and configuration
```
TomerShor: That's right, but inside the `except` we already know that it is not a mounted config map

server/api/crud/runtimes/nuclio/function.py
TomerShor: I meant to comment the "why" instead of the "what" - something like:

since environment variables have a limited size, large serving specs are stored in config maps that are mounted to the pod

tests/api/runtimes/test_serving.py
```diff
@@ -166,22 +194,30 @@ def _assert_deploy_spec_has_secrets_config(self, expected_secret_sources):
                 # TODO: Vault: uncomment when vault returns to be relevant
                 # "MLRUN_SECRET_STORES__VAULT__ROLE": f"project:{self.project}",
                 # "MLRUN_SECRET_STORES__VAULT__URL": mlconf.secret_stores.vault.url,
-                # For now, just checking the variable exists, later we check specific contents
-                "SERVING_SPEC_ENV": None,
             }
+            if not use_config_map:
+                expected_env["SERVING_SPEC_ENV"] = None
```
TomerShor, line 200: leave the comment
```suggestion
                # For now, just checking the variable exists, later we check specific contents
                expected_env["SERVING_SPEC_ENV"] = None
```

docs/serving/writing-custom-steps.ipynb
liranbg, line 1: I suggest moving. `get_serving_spec` under mlrun.utils.helpers and use that

mlrun/serving/server.py
```diff
@@ -307,15 +308,34 @@ def wait_for_completion(self):
         """wait for async operation to complete"""
         return self.graph.wait_for_completion()
 
+    @staticmethod
+    def get_serving_spec():
```
liranbg: code reuse with mlrun.utils.helpes

server/api/utils/singletons/k8s.py
```diff
@@ -540,6 +541,68 @@ def delete_secrets(
...
+                logger.error(
+                    "Failed to create k8s config map",
+                    name=name,
+                    exc=mlrun.errors.err_to_str(exc),
+                )
+                raise exc
+        return name
+
+    @raise_for_status_code
+    def delete_configmap(self, name: str, namespace: str = "", raise_on_error=False):
```
liranbg:
```suggestion
    def delete_configmap(self, name: str, namespace: str = "", raise_on_error=True):
```

liranbg (commented):
Last round

server/api/utils/singletons/k8s.py
```diff
@@ -540,6 +541,68 @@ def delete_secrets(
...
+        self,
+        resource: str,
+        name: str,
+        data: dict,
+        namespace: str = "",
+        labels: dict = None,
+    ):
+        namespace = self.resolve_namespace(namespace)
+
+        name = f"{resource}-{name}"
```
liranbg: I suggest handling (enforce) the resource name to adhere k8s name length limit (64)
note that if you trim+rand then you would need to make sure you dont create many duplicates every time you "ensure" a too-long-named resource. maybe add a label of "mlrun-function-name" or so, and get existing configmap by that label value to understand whether you need to get or create

server/api/utils/singletons/k8s.py
```diff
@@ -540,6 +542,96 @@ def delete_secrets(
         self.v1api.delete_namespaced_secret(secret_name, namespace)
         return mlrun.common.schemas.SecretEventActions.deleted
 
+    @staticmethod
+    def _generate_rand_string(length):
+        return "".join(random.choice(string.ascii_letters) for _ in range(length))
```
liranbg: move below public (we first show API functions, then public functions  then private functions)

liranbg (approved):

PR #5199, opened 2024-02-26
[Project] Return not found status code when project not found

moranbental:
The worker return **NO_CONTENT** status code instead of **NOT_FOUND** status code when a project deletion request is sent and the project cannot be found in the DB.
https://jira.iguazeng.com/browse/IG-22546


moranbental:
This update will break the mlrun client, therefore we check the status code in the backend to see whether it is 204 (no content), which indicates that the project does not exist in mlrun.

moranbental:
This update will break the mlrun client, therefore we check the status code in the backend to see whether it is 204 (no content), which indicates that the project does not exist in mlrun.

PR #5204, opened 2024-02-27
[Security] Static analysis issues remediation [1.6.x]

liranbg:
Most issues are pretty basic with no real impact, but getting those fixes in are better for getting the report looks much better.

there should not back any non-backwards compatibility issues here

### DIFF
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -288,7 +288,6 @@ default_config = {
         "state": "online",
         "retry_api_call_on_exception": "enabled",
         "http_connection_timeout_keep_alive": 11,
-
         # http client used by httpdb
         "http": {
             # when True, the client will verify the server's TLS
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -291,7 +291,7 @@ default_config = {
         # http client used by httpdb
         "http": {
             # when True, the client will verify the server's TLS
-            # sset to False for backwards compatibility.
+            # set to False for backwards compatibility.
             "verify": False,
         },
         "db": {
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -288,6 +288,13 @@ default_config = {
         "state": "online",
         "retry_api_call_on_exception": "enabled",
         "http_connection_timeout_keep_alive": 11,
+
+        # http client used by httpdb
+        "http": {
+            # when True, the client will verify the server's TLS
+            # sset to False for backwards compatibility.
+            "verify": False,
+        },
         "db": {
             "commit_retry_timeout": 30,
             "commit_retry_interval": 3,
--- a/mlrun/kfpops.py
+++ b/mlrun/kfpops.py
@@ -91,11 +91,12 @@ def write_kfpmeta(struct):
         elif key in results:
             val = results[key]
         try:
-
             # NOTE: if key has "../x", it would fail on path traversal
             path = os.path.join(KFP_ARTIFACTS_DIR, key)
             if not mlrun.utils.helpers.resolve_safe_path(KFP_ARTIFACTS_DIR, path):
-                logger.warning("Path traversal is not allowed ignoring", path=path, key=key)
+                logger.warning(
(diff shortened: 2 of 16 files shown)


server/api/utils/clients/nuclio.py
```diff
@@ -237,7 +237,13 @@ def _send_request_to_api(
         if auth_info:
             auth = auth_info.to_nuclio_auth_info().to_requests_auth()
 
-        response = self._session.request(method, url, verify=False, auth=auth, **kwargs)
+        response = self._session.request(
+            method,
+            url,
+            verify=mlrun.config.config.httpdb.http.verify,
```
theSaarco, line 243: General question - now all clients (`httpdb`, `nuclio`, `iguazio`) are using the same `http.verify` config parameter. Question is - do you see a situation where we want to use a different mode per client? For example, if I set this parameter to `True`, it assumes that all communication goes through SSL (i.e. through ingress) - is this always the case when calling `nuclio`/`iguazio`? 
Maybe it's better to set a different parameter per client? TBH, I'm not sure what the answer is...

mlrun/kfpops.py
```diff
@@ -41,8 +41,8 @@
 
 # default KFP artifacts and output (ui metadata, metrics etc.)
 # directories to /tmp to allow running with security context
-KFPMETA_DIR = os.environ.get("KFPMETA_OUT_DIR", "/tmp")
-KFP_ARTIFACTS_DIR = os.environ.get("KFP_ARTIFACTS_DIR", "/tmp")
+KFPMETA_DIR = "/tmp"
+KFP_ARTIFACTS_DIR = "/tmp"
```
theSaarco, line 45: I'm ok with removing this, but I think we need at least a RN that notes we no longer allow the use of these env variables, just in case someone was doing it.

theSaarco (approved):
Looks good, but I do have one open question that I'm not sure about.

mlrun/utils/helpers.py
```diff
@@ -1622,3 +1622,11 @@ def get_local_file_schema() -> List:
     # The expression `list(string.ascii_lowercase)` generates a list of lowercase alphabets,
     # which corresponds to drive letters in Windows file paths such as `C:/Windows/path`.
     return ["file"] + list(string.ascii_lowercase)
+
+
+def resolve_safe_path(base, filepath, is_symlink=False):
```
alonmr, line 1627: seems more fitting ,also please add type hints
```suggestion
def is_safe_path(base, filepath, is_symlink=False):
```

server/api/utils/clients/nuclio.py
liranbg: that is not a real case, it is either you have tls termination with valid cert or not

mlrun/config.py
```diff
@@ -288,6 +288,12 @@
         "state": "online",
         "retry_api_call_on_exception": "enabled",
         "http_connection_timeout_keep_alive": 11,
+        # http client used by httpdb
+        "http": {
+            # when True, the client will verify the server's TLS
+            # sset to False for backwards compatibility.
+            "verify": False,
```
alonmr, line 295: agree with Saar we might want to split this to nuclio, httpdb, etc.
I think it should be configured for each client and also have an option to always verify.
maybe something like:
```
# list of clients to verify TLS, options are nuclio, httpdb, iguazio or "*" to verify for all.
"verify": []
```

mlrun/config.py
```diff
@@ -288,6 +288,12 @@
         "state": "online",
         "retry_api_call_on_exception": "enabled",
         "http_connection_timeout_keep_alive": 11,
+        # http client used by httpdb
+        "http": {
+            # when True, the client will verify the server's TLS
+            # sset to False for backwards compatibility.
```
alonmr, line 294:
```suggestion
            # set to False for backwards compatibility.
```

alonmr (commented):
looks good, minor comments

PR #5209, opened 2024-02-28
[Documentation] Add `DatastoreProfileHdfs` to datastore page

gtopper:


docs/store/datastore.md
```diff
@@ -300,8 +301,20 @@ ParquetTarget(path="ds://profile-name/aws_bucket/path/to/parquet.pq")
...
+profile = DatastoreProfileHdfs(name="profile-name")
+ParquetTarget(path="ds://profile-name/path/to/parquet.pq")
+```
 
+`DatastoreProfileHdfs` init parameters:
+- `name` &mdash; Name of the profile
+- `host` &mdash; HDFS namenode host
+- `port` &mdash; HDFS namenode port
+- `http_port` &mdash; WebHDFS port
+- `user` &mdash; User name. Currently, this has no effect when using Spark. Optional. Defaults to the value of the `HADOOP_USER_NAME` environment variable, or else the current user's user name.
```
assaf758, line 317:
```suggestion
- `user` &mdash; User name. Optional - defaults to the value of the `HADOOP_USER_NAME` environment variable, or else the current user's user name. Spark always use the HADOOP_USER_NAME environment variable.
```

gtopper: That's not exactly right though. Spark also falls back on the user name. The only difference is it does all this at the time the spark context is created. Will update accordingly.

gtopper: Pushed new phrasing we decided on together. Please review @jillnogold.

assaf758 (approved):

PR #5210, opened 2024-02-28
[Model Monitoring] Implement TSDB abstraction

Eyal-Danieli:
Add Time Series DB abstraction that will be used for storing the model monitoring real time data. The PR includes:

- A new `tsdb` package under `mlrun/model_monitoring/db/tsdb`. The abstract class called `TSDBConnector` and it includes the main applications that are being used throughout the model monitoring processes.
- An implementation of `V3IOTSDBConnector` that is stored under `mlrun/model_monitoring/db/tsdb/v3io_connector`. Note that this object is based on V3IO `frames` client that handles the communications with V3IO TSDB. 
- The TSDB type is defined under a new configuration value called `tsdb_connector_type`. At the moment, the default value is `v3io-tsdb`.
- Create the TSDB application result table during the writer deployment (Instead of trying to generate it on every writer call).  

Related JIRAs - https://iguazio.atlassian.net/browse/ML-5795, https://iguazio.atlassian.net/browse/ML-5796


mlrun/model_monitoring/stream_processing.py
```diff
@@ -321,77 +330,12 @@ def apply_storey_sample_window():
         # Steps 20-21 - Prometheus branch
         if not mlrun.mlconf.is_ce_mode():
```
davesh0812, line 324: I think you should move this check to `get_tsdb_connector` line 36 :)

mlrun/model_monitoring/db/tsdb/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def to_tsdb_connector(self, project: str, **kwargs) -> TSDBConnector:
+        """
+        Return a TSDBConnector object based on the provided enum value.
+        :param project: The name of the project.
+        :return: `TSDBConnector` object.
+        """
+
+        if self == self.v3io_tsdb:
+            from .v3io.v3io import V3IOTSDBConnector
```
davesh0812: I think its better to change the file name to be v3io.v3io_connector or v3io.connector

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -0,0 +1,105 @@
...
+        different key metric dictionaries. This data is being used by the monitoring dashboards in
+        grafana.
+        There are 3 different key metric dictionaries that are being generated throughout these steps:
+        - base_metrics (average latency and predictions over time)
+        - endpoint_features (Prediction and feature names and values)
+        - custom_metrics (user-defined metrics)
+        """
+        pass
+
+    def write_application_event(self, event: dict):
```
davesh0812:
```suggestion
    def write_application_result(self, event: dict):
```

better to be explicit because metrics will be write to  different tables.

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -0,0 +1,105 @@
...
+        """
+        pass
+
+    def delete_tsdb_resources(self):
+        """
+        Delete all project resources in the TSDB connector, such as model endpoints data and drift results.
+        """
+        pass
+
+    def get_model_endpoint_real_time_metrics(
```
davesh0812, line 75: please move this function to crud as we talked.

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -0,0 +1,105 @@
...
+# limitations under the License.
+#
+
+
+from abc import ABC
+
+import pandas as pd
+
+
+class TSDBConnector(ABC):
```
davesh0812, line 24: Add 2 kinds of error one for raise from the `write` functions and one to raise from the `get_record` function, and maybe consider to add another one to the initialize function.

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -0,0 +1,105 @@
...
+    def __init__(self, project: str):
+        """
+        Initialize a new TSDB connector.
+        :param project: the name of the project.
+        """
+        self.project = project
+
+    def apply_monitoring_stream_steps(self, graph):
```
davesh0812, line 44: Nice idea!!

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -0,0 +1,105 @@
...
+
+from abc import ABC
+
+import pandas as pd
+
+
+class TSDBConnector(ABC):
+    def __init__(self, project: str):
+        """
+        Initialize a new TSDB connector.
```
davesh0812: Please add all the tables names (descriptions) that this class suppose to connect to.

mlrun/model_monitoring/writer.py
```diff
@@ -153,30 +115,10 @@ def _update_kv_db(self, event: _AppResultEvent) -> None:
 
     def _update_tsdb(self, event: _AppResultEvent) -> None:
         event = _AppResultEvent(event.copy())
-        event[WriterEvent.END_INFER_TIME] = datetime.datetime.fromisoformat(
-            event[WriterEvent.END_INFER_TIME]
+        tsdb_connector = mlrun.model_monitoring.get_tsdb_connector(
```
davesh0812: Please move to the `__init__()` method.
Also you can remove this function and to run `self.tsdb_connector.write_application_event(event=event)`
from the `do()` method.

Can be done also for the kv.

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -284,27 +283,23 @@ def delete_model_endpoints_resources(self, endpoints: list[dict[str, typing.Any]
...
-        store, _, _ = mlrun.store_manager.get_or_create_store(tsdb_path)
-        store.rm(tsdb_path, recursive=True)
+        tsdb_connector = mlrun.model_monitoring.get_tsdb_connector(
+            project=self.project,
+            access_key=self.access_key,
+            container=self.container,
+        )
+        tsdb_connector.delete_tsdb_resources()
+
+        if mlrun.mlconf.model_endpoint_monitoring.tsdb_connector_type == "v3io-tsdb":
```
davesh0812: I think it better to move this inside to `delete_tsdb_resources` of the v3io connector

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -284,27 +283,23 @@ def delete_model_endpoints_resources(self, endpoints: list[dict[str, typing.Any]
...
-        # Final cleanup of tsdb path
-        tsdb_path.replace("://u", ":///u")
-        store, _, _ = mlrun.store_manager.get_or_create_store(tsdb_path)
-        store.rm(tsdb_path, recursive=True)
+        tsdb_connector = mlrun.model_monitoring.get_tsdb_connector(
+            project=self.project,
+            access_key=self.access_key,
+            container=self.container,
+        )
+        tsdb_connector.delete_tsdb_resources()
```
davesh0812: Like discussed I think you need to had `endpoints` param to this function.
And also to call this function from `ModelEndpoints.delete_model_endpoints_resources()` and not from the kv/sql store.

mlrun/model_monitoring/db/tsdb/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+    v3io_tsdb = "v3io-tsdb"
+
+    def to_tsdb_connector(self, project: str, **kwargs) -> TSDBConnector:
+        """
+        Return a TSDBConnector object based on the provided enum value.
+        :param project: The name of the project.
+        :return: `TSDBConnector` object.
+        """
+
+        if self == self.v3io_tsdb:
```
davesh0812, line 36: check if not in ce mode :)

server/api/crud/model_monitoring/deployment.py
```diff
@@ -242,6 +242,10 @@ def deploy_model_monitoring_writer_application(
                 writer_data=fn.to_dict(),
                 writer_ready=ready,
             )
+            # Create tsdb table for model monitoring application results
+            self._create_tsdb_application_tables(
+                project=fn.metadata.project, access_key=self.model_monitoring_access_key
```
davesh0812: No need to pass this, can cause a bug for user without admin permissions.
Because we are in the server side the env access key is all ready pipelines's.

server/api/crud/model_monitoring/deployment.py
```diff
@@ -587,6 +591,27 @@ def is_monitoring_stream_has_the_new_stream_trigger(self) -> bool:
...
+        """Each project writer service writes the application results into a single TSDB table and therefore the
+        target table is created during the writer deployment"""
+
+        if (
+            mlrun.mlconf.model_endpoint_monitoring.tsdb_connector_type
+            == mm_constants.TSDBTarget.V3IO_TSDB
+        ):
+            tsdb_configurations = {
+                "access_key": kwargs["access_key"],
+            }
```
davesh0812: Because you don't need special access key this check is redundant, and you can use `get_tsdb_connector()` without extra kwargs.

davesh0812 (commented):
Nice job! some comments, we discussed about most of themץ

mlrun/model_monitoring/db/tsdb/v3io/stream_graph_steps.py
```diff
@@ -0,0 +1,117 @@
...
+
+import mlrun.feature_store.steps
+from mlrun.common.schemas.model_monitoring import (
+    EventFieldType,
+    EventKeyMetrics,
+    EventLiveStats,
+)
+
+
+class ProcessBeforeTSDB(mlrun.feature_store.steps.MapClass):
```
gtopper, line 24: We just duplicate the whole class from `stream_processing.py`?

mlrun/model_monitoring/db/tsdb/v3io/v3io.py
```diff
@@ -0,0 +1,365 @@
...
+import mlrun.utils.v3io_clients
+from mlrun.utils import logger
+
+_TSDB_BE = "tsdb"
+_TSDB_RATE = "1/s"
+
+
+class V3IOTSDBConnector(mlrun.model_monitoring.db.TSDBConnector):
+    """
+    Handles the TSDB operations when the TSDB connector is from type V3IO. To manage these operations we use V3IO Frames
```
gtopper:
```suggestion
    Handles the TSDB operations when the TSDB connector is of type V3IO. To manage these operations we use V3IO Frames
```

mlrun/model_monitoring/db/tsdb/v3io/v3io.py
```diff
@@ -0,0 +1,365 @@
...
+        tsdb_batching_timeout_secs: int = 300,
+    ):
+        """
+        Apply TSDB steps on the provided monitoring graph. Throughout these steps, the graph stores live data of
+        different key metric dictionaries.This data is being used by the monitoring dashboards in
+        grafana. Results can be found under  v3io:///users/pipelines/project-name/model-endpoints/events/.
+        In that case, we generate 3 different key  metric dictionaries:
+        - base_metrics (average latency and predictions over time)
+        - endpoint_features (Prediction and feature names and values)
+        - custom_metrics (user-defined metrics
```
gtopper:
```suggestion
        - custom_metrics (user-defined metrics)
```

mlrun/model_monitoring/db/tsdb/v3io/v3io.py
```diff
@@ -0,0 +1,365 @@
...
+        """
+
+        # Step 12 - Before writing data to TSDB, create dictionary of 2-3 dictionaries that contains
```
gtopper: It doesn't seem like this Step X stuff belong in the connector class, and could easily become wrong if someone adds a step earlier in the flow, elsewhere in the code.

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+from abc import ABC
+
+import pandas as pd
+
+
+class TSDBConnector(ABC):
+    def __init__(self, project: str):
+        """
```
gtopper, line 26: The constructor isn't normally documented, and the relevant information goes into the class docstring instead.

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -0,0 +1,105 @@
...
+        columns: list[str] = None,
+        filter_query: str = "",
+        start: str = "now-1h",
+        end: str = "now",
+    ) -> pd.DataFrame:
+        """
+        Getting records from TSDB data collection.
+        :param table:            Path to the collection to query.
+        :param columns:          Columns to include in the result.
+        :param filter_query:     Filter expression.
```
gtopper: That doesn't really explain what a filter expression/query is though.

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -0,0 +1,105 @@
...
+        self,
+        table: str,
+        columns: list[str] = None,
+        filter_query: str = "",
+        start: str = "now-1h",
+        end: str = "now",
+    ) -> pd.DataFrame:
+        """
+        Getting records from TSDB data collection.
+        :param table:            Path to the collection to query.
```
gtopper: Tables being located at paths is a rather V3IO-specific thing.

mlrun/model_monitoring/db/tsdb/v3io/v3io.py
```diff
@@ -0,0 +1,365 @@
...
+    ) -> pd.DataFrame:
+        """
+         Getting records from V3IO TSDB data collection.
+        :param table:            Table key to query. The table key should exist in the class tables dictionary.
+        :param columns:          Columns to include in the result.
+        :param filter_query:     V3IO filter expression. The expected filter expression includes different conditions,
+                                 divided by ' AND '.
+        :param start:            The start time of the metrics. Can be represented by a string containing an RFC 3339
+                                 time, a Unix timestamp in milliseconds, a relative time (`'now'` or
+                                 `'now-[0-9]+[mhd]'`, where `m` = minutes, `h` = hours, and `'d'` = days), or 0 for the
```
gtopper: `s` for second is supported as well.

mlrun/model_monitoring/stream_processing.py
```diff
@@ -137,6 +137,12 @@ def _initialize_v3io_configurations(
         self.tsdb_batching_max_events = tsdb_batching_max_events
         self.tsdb_batching_timeout_secs = tsdb_batching_timeout_secs
 
+        self.tsdb_configurations = {
+            "access_key": self.v3io_access_key,
```
gtopper: Not specifically regarding the addition here, but this is no way to treat an access key...

gtopper (commented):
Some comments. The main thing that troubles me about this PR though is that the abstraction is added with only one implementation (V3IO), so it remains to be seen if it would even hold for TDEngine.

mlrun/model_monitoring/db/tsdb/v3io/v3io.py
```diff
@@ -0,0 +1,365 @@
...
+        for table in application_tables:
+            logger.info("Creating table in V3IO TSDB", table=table)
+            self._frames_client.create(
+                backend=_TSDB_BE,
+                table=self.tables[table],
+                if_exists=IGNORE,
+                rate=_TSDB_RATE,
+            )
+
+    def apply_monitoring_stream_steps(
```
gtopper: Doesn't this function need to be broken up? It seems to contain some steps that are not related to V3IO.

gtopper: I kind of expect the abstraction to be at the `apply_tsdb_target` level.

mlrun/model_monitoring/db/tsdb/base.py
Eyal-Danieli: the thing is that we might have different tables structures across different TSDB solutions. Anyway, I'll add here a detailed description of the different types of monitoring data.

mlrun/model_monitoring/db/tsdb/base.py
Eyal-Danieli: Since some of the functionality might be relatet to V3IO (such as how we process the df that is returned from get_records()), i kept it in the v3io connector. I did remove the relevant func from the kv_store and moved it into the crud part. 
Anyway, if we will see that this behaviour is repeating in other TSDBs (TDEngine) then I will move all of the logic to the crud part. Let me know if it's not clear

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
Eyal-Danieli: as we agreed, in the meantime we won't pass endpoints to the delete_tsdb_resources. Anyway it delete all of the resources at the moment.

mlrun/model_monitoring/db/tsdb/v3io/stream_graph_steps.py
Eyal-Danieli: no, it's invalid :) I'll remove it from stream_processing.py

mlrun/model_monitoring/db/tsdb/base.py
Eyal-Danieli: yeah you right but the thing is that the filter expression structure depends on the TSDB type (V3IO/TDEngine..). I'll try to document it better.

mlrun/model_monitoring/db/tsdb/base.py
Eyal-Danieli: actually the docstring here is misleading. It's the table name (e.g. `app-results`, `events`..), not a path. I'll fix it.

mlrun/model_monitoring/db/tsdb/v3io/v3io.py
Eyal-Danieli: Got it, I'll update the description and remove the "hard-coded" step numbering.

mlrun/model_monitoring/db/tsdb/v3io/v3io.py
Eyal-Danieli: I'm not sure we are going to repeat these steps exactly in other TSDBs. Therefore, I preferred to keep it on the V3IO TSDB level at the moment.
 If for example we'll see that for TDEngine we need to do the exact logic then I'll move the "none-V3IO" steps into a higher level.

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -0,0 +1,122 @@
...
+        pass
+
+    def delete_tsdb_resources(self):
+        """
+        Delete all project resources in the TSDB connector, such as model endpoints data and drift results.
+        """
+
+        pass
+
+    def get_model_endpoint_real_time_metrics(
```
davesh0812, line 75: Please mention that those metrics are located and gets from the events table (at least for now).

server/api/crud/model_monitoring/deployment.py
```diff
@@ -587,6 +589,23 @@ def is_monitoring_stream_has_the_new_stream_trigger(self) -> bool:
...
 
+    @staticmethod
+    def _create_tsdb_application_tables(project: str):
+        """Each project writer service writes the application results into a single TSDB table and therefore the
+        target table is created during the writer deployment"""
+
+        if (
+            mlrun.mlconf.model_endpoint_monitoring.tsdb_connector_type
+            == mm_constants.TSDBTarget.V3IO_TSDB
+        ):
```
davesh0812: Please remove this check, you already check it inside `get_tsdb_connector()`

server/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -546,18 +538,21 @@ def delete_model_endpoints_resources(project_name: str):
...
             ),
         )
-        endpoints = endpoint_store.list_model_endpoints()
+        endpoint_store.delete_model_endpoints_resources()
 
-        # Delete model endpoints resources from databases using the model endpoint store object
-        endpoint_store.delete_model_endpoints_resources(endpoints)
+        # Delete model monitoring TSDB resources
+        tsdb_connector = mlrun.model_monitoring.get_tsdb_connector(
+            project=project_name, access_key=auth_info.data_session
```
davesh0812: again you don't need here to provide `access key` you are already on the server, and if the user that will run this api call won't have admin permissions this call will fail.

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -0,0 +1,122 @@
...
+
+import mlrun.common.schemas.model_monitoring.constants as mm_constants
+
+
+class TSDBConnector(ABC):
+    def __init__(self, project: str):
+        """
+        Initialize a new TSDB connector. The connector is used to interact with the TSDB and store monitoring data.
+        At the moment we have 3 different types of monitoring data:
+        - real time performance metrics: real time performance metrics that are being calculated by the model
```
davesh0812, line 29:
```suggestion
        - real time performance metrics (named events): real time performance metrics that are being calculated by the model
```

mlrun/common/schemas/model_monitoring/constants.py
```diff
@@ -228,6 +229,12 @@ class MonitoringFunctionNames(MonitoringStrEnum):
     WRITER = "model-monitoring-writer"
 
 
+class V3IOTSDBTables(MonitoringStrEnum):
```
davesh0812: It spouse to be valid for all the tsdb's engines?  
```suggestion
class MonitoringTSDBTables(MonitoringStrEnum):
```

davesh0812 (approved):
LGTM.

Added few more comments :)

mlrun/db/httpdb.py
```diff
@@ -3167,14 +3161,8 @@ def get_model_endpoint(
...
-        :param start:                      The start time of the metrics. Can be represented by a string containing an
-                                           RFC 3339 time, a Unix timestamp in milliseconds, a relative time (`'now'` or
-                                           `'now-[0-9]+[mhd]'`, where `m` = minutes, `h` = hours, and `'d'` = days), or
-                                           0 for the earliest time.
-        :param end:                        The end time of the metrics. Can be represented by a string containing an
-                                           RFC 3339 time, a Unix timestamp in milliseconds, a relative time (`'now'` or
-                                           `'now-[0-9]+[mhd]'`, where `m` = minutes, `h` = hours, and `'d'` = days), or
-                                           0 for the earliest time.
+        :param start:                      The start time of the metrics.
+        :param end:                        The end time of the metrics.
```
gtopper: Not really sure why all the information regarding these parameters was deleted?
(other occurrences as well)

mlrun/model_monitoring/db/tsdb/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        valid_values = list(cls.__members__.keys())
+        raise mlrun.errors.MLRunInvalidArgumentError(
+            f"{value} is not a valid tsdb, please choose a valid value: %{valid_values}."
+        )
+
+
+def get_tsdb_connector(project: str, **kwargs) -> TSDBConnector:
+    """
+    Getting the TSDB connector type based on mlrun.config.model_endpoint_monitoring.tsdb_connector_type.
```
gtopper:
```suggestion
    Get the TSDB connector type based on mlrun.config.model_endpoint_monitoring.tsdb_connector_type.
```

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def write_application_event(
+        self,
+        event: dict,
+        kind: mm_constants.WriterEventKind = mm_constants.WriterEventKind.RESULT,
+    ):
+        """
+        Write a single application or metric to TSDB.
+
+        :raise mlrun.errors.MLRunInvalidArgumentError: If an error occurred while writing the event.
```
gtopper, line 64:
```suggestion
        :raises mlrun.errors.MLRunInvalidArgumentError: If an error occurred while writing the event.
```

gtopper: Are you sure this is the right exception to raise? Sounds more like an `MLRunRuntimeError`.

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -0,0 +1,123 @@
...
+        Getting records from TSDB data collection.
+        :param table:            Table name, e.g. 'metrics', 'app_results'.
+        :param columns:          Columns to include in the result.
+        :param filter_query:     Optional filter expression as a string. The filter structure depends on the TSDB
+                                 connector type.
+        :param start:            The start time of the metrics.
+        :param end:              The end time of the metrics.
+
+        :return: DataFrame with the provided attributes from the data collection.
+        :raise:  MLRunInvalidArgumentError if the provided table wasn't found.
```
gtopper, line 113:
```suggestion
        :raises:  MLRunInvalidArgumentError if the provided table wasn't found.
```

gtopper: This one sounds more like an `MLRunNotFoundError`.

mlrun/model_monitoring/db/tsdb/v3io/stream_graph_steps.py
```diff
@@ -0,0 +1,117 @@
...
+            processed[EventKeyMetrics.CUSTOM_METRICS] = {
+                EventFieldType.RECORD_TYPE: EventKeyMetrics.CUSTOM_METRICS,
+                **event[EventFieldType.METRICS],
+                **base_event,
+            }
+
+        return processed
+
+
+class FilterAndUnpackKeys(mlrun.feature_store.steps.MapClass):
```
gtopper, line 92: Is this meant to be duplicated verbatim from `mlrun/model_monitoring/stream_processing.py`?

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -0,0 +1,407 @@
...
+        apply_tsdb_target(name="tsdb3", after="FilterNotNone")
+
+    def write_application_event(
+        self,
+        event: dict,
+        kind: mm_constants.WriterEventKind = mm_constants.WriterEventKind.RESULT,
+    ):
+        """
+        Write a single result or metric to TSDB.
+        """
```
```suggestion
        """Write a single result or metric to TSDB"""
```

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -0,0 +1,407 @@
...
+                self._frames_client.delete(
+                    backend=mlrun.common.schemas.model_monitoring.TimeSeriesConnector.TSDB,
+                    table=table,
+                )
+            except v3io_frames.errors.DeleteError as e:
+                if "No TSDB schema file found" not in str(e):
+                    logger.warning(
+                        f"Failed to delete TSDB table '{table}'",
+                        err=mlrun.errors.err_to_str(e),
+                    )
```
gtopper: And if it's not, we just suppress the error entirely?

mlrun/db/httpdb.py
Eyal-Danieli: I'll bring it back though my concern was that this information is only relevant to V3IO and this API is going to serve other TSDBs as well.

mlrun/model_monitoring/db/tsdb/v3io/stream_graph_steps.py
Eyal-Danieli: no ;/

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
Eyal-Danieli: agree, it's an old `if` that seems quite redundant here. We need to log the err either way. I'll update it.

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -0,0 +1,405 @@
...
+                                 `'s'` = seconds), or 0 for the earliest time.
+        :param end:              The end time of the metrics. Can be represented by a string containing an RFC 3339
+                                 time, a Unix timestamp in milliseconds, a relative time (`'now'` or
+                                 `'now-[0-9]+[mhd]'`, where `m` = minutes, `h` = hours, `'d'` = days, and
+                                 `'s'` = seconds), or 0 for the earliest time.
+        :return: DataFrame with the provided attributes from the data collection.
+        :raise:  MLRunNotFoundError if the provided table wasn't found.
+        """
+        if table not in self.tables:
+            raise mlrun.errors.MLRunInvalidArgumentError(
```
gtopper: Hm, which error is it? `MLRunNotFoundError` or `MLRunInvalidArgumentError`?

gtopper (approved):

PR #5212, opened 2024-02-29
[Serving] Undo code simplification that breaks with older nuclio

gtopper:
Changing the handler to a regular method that may still return a coroutine object as part of #5203 is correct, but older (<1.12.10) nuclio versions explicitly check whether the handler method itself is a coroutine, rather than whether it returned a coroutine object.

This PR effectively reverts the last commit ([23be466](https://github.com/mlrun/mlrun/pull/5203/commits/23be466b3ff62882be5e724acf8623e8a755bdce)) in #5203, which aimed to simplify the code, not taking older nuclio versions into account.

For reference, the nuclio code in question is [here](https://github.com/nuclio/nuclio/commit/af713bc4e65a6b7324dabf5c354f532cacab9d78#diff-725534bace834296150d3e0c0525fd5ef3ec7b57ddaa193e5cc98ed5d062e2eaL425).


gtopper:
We need to use `SyncEmitSource` when nuclio is older than 1.12.10, or simply require this as the minimum nuclio version for mlrun 1.7.0.

PR #5224, opened 2024-03-03
[Project] Use job kind by default if not specified

Yacouby:
when using project.set_function(), or when running CLI build command with a source function, 
use kind="job" if the user did not specify it.

Resolves:
https://iguazio.atlassian.net/browse/ML-886

### DIFF
--- a/mlrun/__main__.py
+++ b/mlrun/__main__.py
@@ -505,6 +505,8 @@ def build(
         if kfp:
             print("Runtime:")
             pprint(runtime)
+        # use kind = "job" by default if not specified
+        runtime.setdefault("kind", "job")
         func = new_function(runtime=runtime)
 
     elif func_url:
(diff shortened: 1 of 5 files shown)


mlrun/run.py
```diff
@@ -389,6 +389,9 @@ def import_function_to_dict(url, secrets=None):
     code = get_in(runtime, "spec.build.functionSourceCode")
     update_in(runtime, "metadata.build.code_origin", url)
     cmd = code_file = get_in(runtime, "spec.command", "")
+    # use kind = "job" by default if not specified
+    if "kind" not in runtime:
+        runtime["kind"] = "job"
```
TomerShor: for the CLI `build` command, this will only handle the flow where the (optional) `func_url` argument is passed.
What about the case where a dict is passed to the `--runtime` flag, and it doesn't contain a kind?
Add a unittest for this case as well

alonmr: use `runtime.set_default("kind", "job")`

mlrun/__main__.py
```diff
@@ -505,6 +505,9 @@ def build(
         if kfp:
             print("Runtime:")
             pprint(runtime)
+        # use kind = "job" by default if not specified
+        if "kind" not in runtime:
```
alonmr: same here

alonmr (approved):

PR #5234, opened 2024-03-04
[Application] Introducing application runtime phase I

alonmr:
# Application Runtime

MLRun allows user to deploy models and expose them for serving purposes. 

We currently lack a way to enable users deploy an application atop those models. e.g. have a model that differentiate between cat and a dog and expose it while having a UI that allows the user to upload image to invoke the model with a given input. To do so, we introduce a way to deploy web application given user container image, port, and a command to run the http server.

The runtime is based on top of Nuclio and will add the application as a side-car to a Nuclio function while the actual function is a reverse proxy to that application.
The reason we want to use Nuclio for this runtime is to leverage its features such as scaling, triggers, monitoring, API gateway, etc.

## Phase I

This phase is about making it work.
1. Configure the side car application.
2. Add the reverse proxy to the side car.

## Usage example
~~~
# Create an application runtime (with pre built image)
application = project.set_function(name="my-vizro-dashboard", kind="application", image="repo/my-wabapp:latest")

# Set the port the side-car listens on
application.set_internal_app_port(port=8080)

# Deploy
application.deploy()
~~~

Jira - https://iguazio.atlassian.net/browse/ML-4601


mlrun/runtimes/nuclio/function.py
```diff
@@ -958,6 +957,40 @@ def invoke(
...
+    def _with_sidecar(self, name: str, image: str, port: int = None):
+        """
+        Add a sidecar container to the function pod
+        :param name:    Sidecar container name
+        :param image:   Sidecar container image
+        :param port:    Sidecar container port
+        """
+        # TODO: validate image on server side
+        if not name:
+            raise ValueError("Sidecar name must be specified")
```
TomerShor: Why not just enrich with `f"{name}-sidecar"` here?

tests/runtimes/test_application.py
```diff
@@ -0,0 +1,91 @@
...
+
+def test_create_application_runtime():
+    fn: mlrun.runtimes.ApplicationRuntime = mlrun.code_to_function(
+        "application-test", kind="application", image="mlrun/mlrun"
+    )
+    assert fn.kind == mlrun.runtimes.RuntimeKinds.application
+    assert fn.spec.image == "mlrun/mlrun"
+    assert fn.metadata.name == "application-test"
+    assert (
+        "Ly8gQ29weXJpZ2h0IDIwMjMgSWd1YXppbwovLwovLyBMaWN"
```
TomerShor: I assume this is base64 of the go reverse proxy func? perhaps add a comment on this

mlrun/projects/project.py
```diff
@@ -3846,6 +3846,18 @@ def _init_function_from_dict(
                 tag=tag,
             )
 
+    elif image and kind in mlrun.runtimes.RuntimeKinds.nuclio_runtimes():
```
TomerShor, line 3899: I'm missing a validation that `image` exists if the runtime kind is `application`. we can't guess what image to run in the sidecar...

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -0,0 +1,208 @@
...
+        # Save the application image in the status to allow overriding it with the reverse proxy entry point
+        if self.spec.image and (
+            not self.status.application_image
+            or self.spec.image != self.status.container_image
+        ):
+            self.status.application_image = self.spec.image
+            self.spec.image = ""
+
+        if self.status.container_image:
+            self.from_image(self.status.container_image)
```
TomerShor, line 264: What is this for, redployments?

TomerShor (commented):
Looking really good overall!
some comments below

mlrun/runtimes/nuclio/application/application.py
alonmr: yes, turns out this doesn't really work because we need to remove the code from the spec as well. fixing

mlrun/projects/project.py
alonmr: discussed offline

mlrun/run.py
```diff
@@ -718,35 +720,45 @@ def update_common(fn, spec):
...
 
     if kind == RuntimeKinds.databricks and not embed_code:
-        raise ValueError("databricks tasks only support embed_code=True")
+        raise ValueError("Databricks tasks only support embed_code=True")
+
+    if kind == RuntimeKinds.application:
+        if handler:
+            raise ValueError("Handler is not supported for application runtime")
+        filename = str(ApplicationRuntime.reverse_proxy_file_path)
+        handler = "Handler"
```
theSaarco:
```suggestion
        filename, handler = ApplicationRuntime.get_filename_and_handler()
```
It's better to ask the `ApplicationRuntime` class what these are - add a class method to the class to return them. This way any change in these will be local to the runtime class, and will not require searching through the code to find the reference here.

mlrun/run.py
```diff
@@ -718,35 +720,45 @@ def update_common(fn, spec):
...
+        _is_nuclio = _kind.startswith("nuclio")
+        _sub_kind = (
+            _kind[_kind.find(":") + 1 :] if _is_nuclio and ":" in _kind else None
+        )
+        if _kind == RuntimeKinds.serving:
+            _is_nuclio = True
+            _sub_kind = serving_subkind
+        elif _kind == RuntimeKinds.application:
+            _is_nuclio = True
+        return _is_nuclio, _sub_kind
```
theSaarco: Since we're making these changes - isn't it better to move this to the `RuntimeKinds` class? Like `resolve_nuclio_runtime`?

mlrun/runtimes/__init__.py
```diff
@@ -211,6 +216,18 @@ def requires_image_name_for_execution(kind):
         # both spark and remote spark uses different mechanism for assigning images
         return kind not in [RuntimeKinds.spark, RuntimeKinds.remotespark]
 
+    @staticmethod
+    def resolve_nuclio_runtime(kind: str, sub_kind: str):
```
theSaarco, line 220: Maybe add a sanity check to verify that `kind` is a nuclio kind, and explode otherwise.

mlrun/runtimes/nuclio/application/__init__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2023 Iguazio
```
theSaarco:
```suggestion
# Copyright 2024 Iguazio
```

😃

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -0,0 +1,209 @@
...
+        self.application_image = application_image or None
+
+
+class ApplicationRuntime(RemoteRuntime):
+    kind = "application"
+    reverse_proxy_file_path = pathlib.Path(__file__).parent / "reverse_proxy.go"
+
+    @min_nuclio_versions("1.12.7")
+    def __init__(self, spec=None, metadata=None):
+        super().__init__(metadata, spec)
```
theSaarco: It's terrible that the order of parameters is different between this class and its `super()`. How did that happen? As far as I can tell, the parent is `KubeResource` which also does the same reversing (there's `RemoteRuntime` which doesn't implement `__init__`). So, does this even work?

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -0,0 +1,209 @@
...
+        return self._status
+
+    @status.setter
+    def status(self, status):
+        self._status = self._verify_dict(status, "status", ApplicationStatus)
+
+    def set_internal_application_port(self, port: int):
+        port = int(port)
+        if port < 0 or port > 65535:
+            raise ValueError("Port must be in the range 0-65535")
```
theSaarco: A user may be tempted to just update this directly as a property (i.e. `fn.spec.internal_application_port = XXX`) overriding this check. Even more so since there's no getter, so to see the port you need to look at `fn.spec.internal_application_port`.
I would add also a setter on the spec class with the same validation, or just remove it from here and just call the setter.

mlrun/runtimes/nuclio/function.py
```diff
@@ -958,6 +957,38 @@ def invoke(
             data = json.loads(data)
         return data
 
+    def _with_sidecar(self, name: str, image: str, port: int = None):
```
theSaarco: Since this is now a general nuclio functionality, why not make this a public function, allowing users to add sidecars to nuclio functions not just in the application runtime case?

mlrun/runtimes/nuclio/function.py
```diff
@@ -958,6 +957,38 @@ def invoke(
...
+        """
+        Add a sidecar container to the function pod
+        :param name:    Sidecar container name
+        :param image:   Sidecar container image
+        :param port:    Sidecar container port
+        """
+        # TODO: validate image on server side
+        name = name or f"{self.metadata.name}-sidecar"
```
theSaarco, line 979: There's no default value to `name`, so a user will have to make an effort and call `_with_sidecar(name=None,...)` to justify this check. I think you should either make `name` optional, or explode if `not name`.

mlrun/runtimes/nuclio/function.py
```diff
@@ -958,6 +957,38 @@ def invoke(
...
+        sidecar = self._set_sidecar(name)
+        if image:
+            sidecar["image"] = image
+
+        if port:
+            sidecar["ports"] = [
```
theSaarco: It looks like the sidecar function in nuclio supports multiple ports. Why not expose it here, and allow sending a list of ports to this function? Maybe change `port` to be `ports: Optional[Union[int, list[int]]] = None`?

tests/runtimes/test_application.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+    assert fn.kind == mlrun.runtimes.RuntimeKinds.application
+    assert fn.spec.image == "mlrun/mlrun"
+    assert fn.metadata.name == "application-test"
+    # base64 prefix of the reverse proxy code
+    assert (
+        "Ly8gQ29weXJpZ2h0IDIwMjMgSWd1YXppbwovLwovLyBMaWN"
+        in fn.spec.build.functionSourceCode
+    )
+    assert fn.spec.function_handler == "reverse_proxy:Handler"
```
theSaarco: Once you add the file-path and handler as class methods for the app runtime, you can use them to construct this value, rather than use a fixed string.

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -0,0 +1,209 @@
...
+            external_invocation_urls=external_invocation_urls,
+            build_pod=build_pod,
+            container_image=container_image,
+        )
+        self.application_image = application_image or None
+
+
+class ApplicationRuntime(RemoteRuntime):
+    kind = "application"
+    reverse_proxy_file_path = pathlib.Path(__file__).parent / "reverse_proxy.go"
```
theSaarco: Out of curiosity - does nuclio understand that it needs to use Golang runtime based on the suffix of the file? There doesn't seem to be any indication in the code here about it.

theSaarco (changes requested):
Cool stuff. Have some comments and suggestions.

mlrun/runtimes/nuclio/application/__init__.py
alonmr: nitpicker 🐛 😄 
needs to be changed in all files

mlrun/runtimes/nuclio/application/application.py
alonmr: oh wow fixed. copied it from KubeResource and fixed that one as well.

mlrun/runtimes/nuclio/application/application.py
alonmr: nuclio-jupyter enriches the spec.runtime from the file extension

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -0,0 +1,281 @@
...
+            image=self.status.application_image,
+            ports=self.spec.internal_application_port,
+            command=self.spec.command,
+            args=self.spec.args,
+        )
+        self.set_env("SIDECAR_PORT", self.spec.internal_application_port)
+        self.set_env("SIDECAR_HOST", "http://localhost")
+
+    @classmethod
+    def get_filename_and_handler(cls) -> (str, str):
```
TomerShor, line 281: I don't think it should be a public function...
Is is a `classmethod` only for testing purposes? if so, I think it's not really needed to be exposed to users.

mlrun/runtimes/nuclio/application/reverse_proxy.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+func Handler(context *nuclio.Context, event nuclio.Event) (interface{}, error) {
+    reverseProxy := context.UserData.(map[string]interface{})["reverseProxy"].(*httputil.ReverseProxy)
+    sidecarUrl := context.UserData.(map[string]interface{})["server"].(string)
+
+    // populate reverse proxy http request
+    httpRequest, err := http.NewRequest(event.GetMethod(), event.GetPath(), bytes.NewReader(event.GetBody()))
+    if err != nil {
+        return nil, err
```
TomerShor:
```suggestion
    if err != nil {
        context.Logger.ErrorWith("Failed to create a reverse proxy request")
        return nil, err
```

tests/runtimes/test_application.py
```diff
@@ -0,0 +1,197 @@
...
+        "application-test", kind="application", image="mlrun/mlrun"
+    )
+    assert fn.kind == mlrun.runtimes.RuntimeKinds.application
+    assert fn.spec.image == "mlrun/mlrun"
+    assert fn.metadata.name == "application-test"
+    # base64 prefix of the reverse proxy code
+    assert (
+        "Ly8gQ29weXJpZ2h0IDIwMjQgSWd1YXppbwovLwovLyBMaWN"
+        in fn.spec.build.functionSourceCode
+    )
```
TomerShor: Thinking about it again, this is not very resilient to changes in the reverse proxy source code, let alone if you are reusing it in other tests.
Better to automate it (using the `filepath` below)

tests/runtimes/test_application.py
```diff
@@ -0,0 +1,197 @@
...
+
+    filepath, expected_handler = (
+        mlrun.runtimes.ApplicationRuntime.get_filename_and_handler()
+    )
+    expected_filename = pathlib.Path(filepath).name
+    expected_function_handler = f"{expected_filename.split('.')[0]}:{expected_handler}"
+    assert fn.spec.function_handler == expected_function_handler
```
TomerShor: duplicate of the test above it - move to common private function

tests/runtimes/test_application.py
```diff
@@ -0,0 +1,197 @@
...
+    fn.deploy()
+    assert fn.spec.config["spec.sidecars"] == [
+        {
+            "image": image,
+            "name": "application-test-sidecar",
+            "ports": [{"containerPort": 8080, "name": "http", "protocol": "TCP"}],
+        }
+    ]
+    assert fn.status.application_image == image
+    assert not fn.spec.image
```
TomerShor: This is also duplicated multiple times and in multiple tests - create a function for it

mlrun/runtimes/nuclio/application/application.py
alonmr: it's needed for `code_to_function`

theSaarco (approved):
Looks good 👍 .

TomerShor (approved):
Good work mate :)

setup.py
```diff
@@ -59,6 +59,7 @@ def version():
             "server",
         ]
     ),
+    package_data={"": ["*.go"]},
```
liranbg: provide the path and dont take all go files

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -0,0 +1,281 @@
...
+        if self.spec.image and (
+            not self.status.application_image
+            or self.spec.image != self.status.container_image
+        ):
+            self.status.application_image = self.spec.image
+            self.spec.image = ""
+
+        if self.status.container_image:
+            self.from_image(self.status.container_image)
+            self.spec.build.functionSourceCode = ""
```
liranbg, line 267:
```suggestion
            # nuclio implementation detail - when providing the image and emptying out the source code,
            # nuclio skips rebuilding the image and simply takes the prebuilt image
            self.spec.build.functionSourceCode = ""
```

liranbg (commented):
minor

PR #5239, opened 2024-03-05
[Tests] Fix running test in case kube config present

yanburman:


quaark (approved):
Spoke to @yanburman offline about the reasoning for this as this issue happens when mlrun identifies a k8s server instance that we don't want the test to use. 
While I approve of this change to ensure the specific test passes, I'm not to keen on adding fixtures to tests without clear reasons for them, as that has caused issues in the past, especially with `k8s_secrets_mock`. I'm wondering if there's a simple solution maybe we can add this fixture as a default UT fixture in the conftest. @alonmr WDYT?

alonmr:
>  I'm wondering if there's a simple solution maybe we can add this fixture as a default UT fixture in the conftest. @alonmr WDYT?

Agree we should have a way to mock the k8s without adding unrelated fixtures. Note we also have `mock_k8s_helper`.
@yanburman see how we use `MockedK8sHelper`. Will that fix it here or is there a reason to use the `K8sSecretsMock` instead?

PR #5245, opened 2024-03-05
[Projects] Fix second create_remote not overriding current remote

yaelgen:
resolves: [ML-5079](https://jira.iguazeng.com/browse/ML-5079)

[ML-5079]: https://iguazio.atlassian.net/browse/ML-5079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -2352,6 +2352,38 @@ class MlrunProject(ModelObj):
         self.spec._source = self.spec.source or url
         self.spec.origin_url = self.spec.origin_url or url
 
+    def set_remote(self, url, name="origin", branch=None, overwrite=False):
+        """create or update a remote for the project git repository.
+
+        This method allows you to manage remote repositories associated with the project.
+        It checks if a remote with the specified name already exists. If it does, you have
+        the option to overwrite it with a new URL or keep the existing remote.
+        If a remote with the same name does not exist, it will be created.
+
+        :param url: remote git url
+        :param name: name for the remote (default is 'origin')
+        :param branch: Git branch to use as source
+        :param overwrite: if `overwrite` is True, update the existing remote with the given name if it already exists.
+                          If set to False (default),
+                          raise an error when attempting to create a remote with a name that already exists
+        :raises MLRunConflictError: If a remote with the same name already exists and overwrite
+                                     is set to False.
+        """
+        if self._remote_exists(name):
+            if overwrite:
+                self.spec.repo.delete_remote(name)
+            else:
+                raise mlrun.errors.MLRunConflictError(
+                    f"Remote '{name}' already exists in the project, "
+                    f"Each remote in the project must have a unique name."
+                    "Use overwrite=True to override the remote, or choose a different name."
+                )
+        self.create_remote(url=url, name=name, branch=branch)
+
+    def _remote_exists(self, name):
+        """check if a remote with the given name already exists"""
+        return any(remote.name == name for remote in self.spec.repo.remotes)
+
     def _ensure_git_repo(self):
         if self.spec.repo:
             return
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -34,6 +34,7 @@ import kfp
 import nuclio
 import requests
 import yaml
+from deprecated import deprecated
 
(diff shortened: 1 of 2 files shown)


tests/projects/test_project.py
```diff
@@ -1562,6 +1562,65 @@ def test_project_create_remote():
...
+            name=name,
+        )
+        with expected:
+            project.set_remote(
+                url=set_url,
+                name=set_name,
+                overwrite=overwrite,
+            )
+
+            if not name == set_name:
```
TomerShor:
```suggestion
            if name != set_name:
```

TomerShor (commented):
Looking good!

I think there is no real reason to have both `create_remote` and `set_remote`, as they provide the same functionality but with an extra feature in `set`.
We can maybe deprecate `create_remote` (add deprecation warning and remove in 2 versions), and encourage users to use only `set_remote`.

quaark (changes requested):
Spoke with @yaelgen offline, I feel a better solution instead of changing create to set (with override defaulting to False), we should extend the API to include create (never overrides), set (override defaults to True and can be set to False) and remove

mlrun/projects/project.py
```diff
@@ -2333,13 +2333,52 @@ def pull(
...
+        This method allows you to manage remote repositories associated with the project.
+        It checks if a remote with the specified name already exists.
+
+        If a remote with the same name does not exist, it will be created.
+        If a remote with the same name already exists,
+        the behavior depends on the value of the 'overwrite' flag:
+
+        - If 'overwrite' is set to True (default), the existing remote will be overwritten with the new URL.
+        - If 'overwrite' is set to False, a ConflictError will be raised, indicating that the remote already exists
+          and cannot be overwritten.
```
TomerShor: kind of a duplicate of lines 2367-2370, I think we can remove this and keep the information in the params below.

mlrun/projects/project.py
```diff
@@ -2333,13 +2333,52 @@ def pull(
...
+        :raises MLRunConflictError: If a remote with the same name already exists and overwrite
+                                     is set to False.
+        """
+        if self._remote_exists(name):
+            if overwrite:
+                self.spec.repo.delete_remote(name)
+            else:
+                raise mlrun.errors.MLRunConflictError(
+                    f"Remote '{name}' already exists in the project, "
+                    f"Each remote in the project must have a unique name."
```
```suggestion
                    f"each remote in the project must have a unique name."
```

mlrun/projects/project.py
```diff
@@ -2333,13 +2333,52 @@ def pull(
...
+                    "Use 'set_remote' with 'override=True' inorder to update the remote, or choose a different name."
```
TomerShor, line 2375:
```suggestion
                    "Use 'override=True' to update the remote, or choose a different name."
```

mlrun/projects/project.py
```diff
@@ -2352,6 +2391,24 @@ def create_remote(self, url, name="origin", branch=None):
...
+        This method removes the remote repository associated with the specified name from the project's Git repository.
+
+        :param name: Name of the remote to remove.
+        """
+        if self._remote_exists(name):
+            self.spec.repo.delete_remote(name)
+        else:
+            raise mlrun.errors.MLRunNotFoundError(
+                f"Remote '{name}' not found in the project's Git repository."
+            )
```
TomerShor: Why raising an exception and not making it idempotent?
Remove a non-existent remote - it is basically removed successfully

TomerShor (changes requested):

mlrun/projects/project.py
yaelgen: I actually mentioned the `set_remote` on purpose here, since now this error can be thrown by calling `create_remote`, so in case the user isn't familiar with this function.

TomerShor: Alright 👍

TomerShor (approved):
LGTM 👍

quaark (approved):
Great stuff!

liranbg (approved):

PR #5255, opened 2024-03-06
[Docs] Change order of FV topics, add comment about timestamp of last event

jillnogold:


docs/feature-store/feature-vectors.md
```diff
@@ -237,7 +237,7 @@ resp = fs.get_offline_features(
 The online feature vector provides real-time feature vectors to the model using the latest data available.
 
 First create an `Online Feature Service` using {py:meth}`~mlrun.feature_store.get_online_feature_service`. Then feed the `Entity` of the 
-feature vector to the service and receive the latest feature vector.
+feature vector to the service and receive the latest feature vector. (The timestamp of the last event is not returned with `get_online_feature_service` / `svc.get`.)
```
assaf758:
```suggestion
feature vector to `get_online_feature_service` and receive the latest value of the feature vector. Note that the response contains only the features - the timestamp (of the last event that updated the feature sets) is not part of the response. 
```

assaf758 (commented):
See comment inline.
Regarding section names - In the start of featue-vectors.md change: # Creating and using feature vectors
to # Feature vectors

assaf758 (approved):

PR #5264, opened 2024-03-10
[Logger] Fix populating log formatter [1.6.x]

liranbg:
Ensure logger formatter is set. since `config.py` is populated after the logger is created, it is required to set it explicitly as provided within this pr

https://iguazio.atlassian.net/browse/ML-5763


quaark (approved):
🚀

PR #5265, opened 2024-03-10
[Pipelines] Omit `sort_by` and `filter` query params in paginated requests [1.6.x]

TomerShor:
**Bug:**
When listing pipelines, we iteratively call the `list_runs` KFO API, and provide 3 relevant parameters:

- `page_token` - the token used for pagination.
- `sort_by` and `filter` - parameters that tell KFP how to perform the query.

In the KFP API,  `page_token` cannot be provided along with the `sort_by` or `filter` parameters, since they are already contained in the parameters used to generate the paginated query (in the token).
When iterating over pipelines and giving a `sort_by` or `filter` value, the 2nd call will fail since it uses the token returned from the 1st page response.

**Fix:**
Omit the `sort_by` and `filter` params when the `page_token` param is not None.
`filter` is already omitted in one request in such case, this PR aligns the second option and the `sort_by` param. 

Fixes https://iguazio.atlassian.net/browse/ML-3638

### DIFF
--- a/server/api/crud/pipelines.py
+++ b/server/api/crud/pipelines.py
@@ -72,7 +72,7 @@ class Pipelines(
                 response = kfp_client._run_api.list_runs(
                     page_token=page_token,
                     page_size=mlrun.common.schemas.PipelinesPagination.max_page_size,
-                    sort_by=sort_by,
+                    sort_by=sort_by if page_token == "" else "",
                     filter=filter_ if page_token == "" else "",
                 )
                 run_dicts.extend([run.to_dict() for run in response.runs or []])
@@ -90,8 +90,8 @@ class Pipelines(
                 page_token=page_token,
                 page_size=page_size
                 or mlrun.common.schemas.PipelinesPagination.default_page_size,
-                sort_by=sort_by,
-                filter=filter_,
+                sort_by=sort_by if page_token == "" else "",
+                filter=filter_ if page_token == "" else "",
             )
             runs = [run.to_dict() for run in response.runs or []]
             runs = self._filter_runs_by_name(runs, name_contains)
--- a/server/api/crud/pipelines.py
+++ b/server/api/crud/pipelines.py
@@ -66,9 +66,9 @@ class Pipelines(
         if project != "*":
             run_dicts = []
             while page_token is not None:
-                # kfp doesn't allow us to pass both a page_token and the filter. When we have a token from previous
-                # call, we will strip out the filter and use the token to continue (the token contains the details of
-                # the filter that was used to create it)
+                # kfp doesn't allow us to pass both a page_token and the `filter` and `sort_by` params.
+                # When we have a token from previous call, we will strip out the filter and use the token to continue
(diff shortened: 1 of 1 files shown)


server/api/crud/pipelines.py
```diff
@@ -90,8 +90,8 @@ def list_pipelines(
                 page_token=page_token,
                 page_size=page_size
                 or mlrun.common.schemas.PipelinesPagination.default_page_size,
-                sort_by=sort_by,
-                filter=filter_,
+                sort_by=sort_by if page_token == "" else "",
+                filter=filter_ if page_token == "" else "",
```
theSaarco: These are not needed, since in this flow (querying across projects with `*` for project name) the BE does not do any iteration internally, but rather just passes the parameters that the user provided to kfp. In this case, the responsibility for the validity of the params falls on the users and we should not make decisions for them.

I don't think there's anything wrong with doing this, but TBH I'm not sure what my opinion on this is. After all, even putting a warning here won't really matter much since the user will not see it (it will be in the MLRun service logs).

theSaarco (approved):
Looks good, though I have some philosophical musings over the fix. Feel free to observe my thinkings as reflected in the comment.

server/api/crud/pipelines.py
TomerShor: I guess you're right, we can let the call fail because we don't iterate by ourselves, so they user will get an error status code.
But, will the `list_runs` call raise an exception in such case?

server/api/crud/pipelines.py
```diff
@@ -72,7 +72,7 @@ def list_pipelines(
                 response = kfp_client._run_api.list_runs(
                     page_token=page_token,
                     page_size=mlrun.common.schemas.PipelinesPagination.max_page_size,
-                    sort_by=sort_by,
+                    sort_by=sort_by if page_token == "" else "",
```
liranbg, line 75: just a suggestion -
```suggestion
                   # omit the sort_by as page token is provided. kf stores the query parameters when page token is provided
                    sort_by=sort_by if not page_token else "",
```

server/api/crud/pipelines.py
theSaarco: Yes, since kfp will raise an exception (you can try to run it on a setup without your fix and see). I do suspect that the error will not be extremely visible to the user... Trying to do something on my setup, looks like it's returning a 500 with 
```
{
  "detail": "ApiException()"
}
```
That is not very helpful.
Given this, it's probably better to either make this succeed (keep your validation) or return a specific error message that explained what happened.

server/api/crud/pipelines.py
TomerShor: There's already a comment about omitting `filter` above the function call, adding `sort_by` to it as well.

server/api/crud/pipelines.py
TomerShor: In that case, I might go with a try-except and wrap the exception in a more informative error

PR #5269, opened 2024-03-10
[System test] Fix remote runner not saving project [1.6.x]

alonmr:
Remote runner should save the project either when:
1. in load only flow
2. user did not give a source param


liranbg (commented):
missing ut coverage

liranbg (approved):

PR #5280, opened 2024-03-12
[Model Monitoring] Log artifacts in the histogram data drift app

jond01:
Resolves [ML-5775](https://iguazio.atlassian.net/browse/ML-5775).

* Add Plotly table artifact
* Add JSON artifact
* Log the general data drift result

image

API change in `do_tracking`:
`sample_df_stats` and `feature_stats` are now the raw dictionaries, as the `pd.DataFrame` did not include all the information needed from the stats, specifically the x-edges of the bins.

The relevant system tests passed.

[ML-5775]: https://iguazio.atlassian.net/browse/ML-5775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/model_monitoring/api.py
```diff
@@ -704,8 +704,8 @@ def perform_drift_analysis(
         drift_detected_threshold=drift_threshold,
     )
 
-    # Drift table plot
-    html_plot = FeaturesDriftTablePlot().produce(
+    # Drift table artifact
+    plotly_artifact = FeaturesDriftTablePlot().produce(
```
Eyal-Danieli, line 708: note that the `drift_results` here is from type `DriftResultType` that is defined within `batch.py` as 
`DriftResultType = tuple[mlrun.common.schemas.model_monitoring.DriftStatus, float]`

while you changed the corresponded param in `produce` function into a **different** `DriftResultType` that is defined within `features_drift_table.py` as 
`DriftResultType = tuple[
    mlrun.common.schemas.model_monitoring.constants.ResultStatusApp, float
]`

mlrun/model_monitoring/application.py
```diff
@@ -84,16 +84,16 @@ class ModelMonitoringApplicationBase(StepToDict, ABC):
         class MyApp(ApplicationBase):
             def do_tracking(
                 self,
-                sample_df_stats: pd.DataFrame,
-                feature_stats: pd.DataFrame,
+                sample_df_stats: mlrun.common.model_monitoring.helpers.FeatureStats,
+                feature_stats: mlrun.common.model_monitoring.helpers.FeatureStats,
```
Eyal-Danieli, line 88: Make sure that you didn't break any backwards compatibility + how does it affect the user app code? when he defines a custom application should he look up for this pre-defined type FeatureStats?

Eyal-Danieli (commented):
Nice! 
A couple of comments to address :)

mlrun/model_monitoring/application.py
jond01: > Make sure that you didn't break any backwards compatibility

We change the API here, but we plan to address the BC issue in `do_tracking` as a part of a more general API improvement: [ML-5494](https://iguazio.atlassian.net/browse/ML-5494).
This should be at least documented in a release note.

> how does it affect the user app code? when he defines a custom application should he look up for this pre-defined type FeatureStats?

Users who want the previous data frames can use `self.dict_to_histogram`.

mlrun/model_monitoring/applications/histogram_data_drift.py
```diff
@@ -115,86 +119,168 @@ def __init__(self, value_classifier: Optional[ValueClassifier] = None) -> None:
...
         results.append(
             ModelMonitoringApplicationResult(
                 name="general_drift",
                 value=value,
                 kind=self.METRIC_KIND,
-                status=self._value_classifier.value_to_status(value),
+                status=status,
             )
         )
+        self.context.log_results(
```
davesh0812: I am not sure this is a good behavior we want to use in the default app, because in the near future the UI will take the results from the kv and will show them properly (with the status, time and etc.).

mlrun/model_monitoring/application.py
davesh0812: @Eyal-Danieli @jond01 Lets talk about all the BC tomorrow in meeting.

davesh0812 (commented):
Looks good, just few comments :)

mlrun/model_monitoring/api.py
jond01: Good catch.
The coming plan is to remove the `perform_drift_analysis` method altogether as a part of the batch removal, so it's in a temporary phase now.

Edit:
[ML-5752](https://iguazio.atlassian.net/browse/ML-5752) will address this.

mlrun/model_monitoring/applications/histogram_data_drift.py
jond01: > in the near future the UI will take the results from the kv and will show them properly (with the status, time and etc.).

Right, but is there any harm in logging selected results inside the app?
I think that's a nice ability

davesh0812: It is a nice ability!
But I think that if you want to use it you can write different values and not the result itself because we already writing it to the kv and we will show it in more accurate way. 
Maybe you can write the other results you got such that the mean of the 3 metric you calculate.

jond01: I removed this in f8eb3d7

mlrun/model_monitoring/application.py
jond01: We agreed to address the BC issue as a part of [ML-5494](https://iguazio.atlassian.net/browse/ML-5494), which is directly related.

Eyal-Danieli (approved):

PR #5283, opened 2024-03-13
[Project] Retain producers of exported artifacts

TomerShor:
Previously, when artifacts where imported either via `load_project` or `import_artifact`, their `producer` property (+tree) were overridden by the registering project information.
In case artifacts were created by runs, overriding their `producer` causes them to lose some information.

To fix this, we retain the registered artifacts' producers according to the following logic:

- If artifact was originally created by a project, we can re-assign it to the new producer, since they are going to belong to this new project anyway. 
- If the artifact was originally registered by a run (producer type is run), we retain the producer, and check if the artifact already exists in the system:
    - If it does - then there's no need to re-register it.
    - If not - register it with the original producer.

Resolves https://iguazio.atlassian.net/browse/ML-4134


mlrun/projects/project.py
```diff
@@ -1402,6 +1395,12 @@ def register_artifacts(self):
                     artifact.src_path = path.join(
                         self.spec.get_code_path(), artifact.src_path
                     )
+                producer = self._resolve_artifact_producer(artifact, project_tag)
+                if producer.name != self.metadata.name:
```
alonmr: merge the ifs with `and`

mlrun/projects/project.py
```diff
@@ -1498,12 +1497,14 @@ def log_artifact(
...
             artifact_path, self.metadata.name
         )
-        producer = ArtifactProducer(
-            "project",
-            self.metadata.name,
-            self.metadata.name,
-            tag=self._get_hexsha() or str(uuid.uuid4()),
-        )
+        producer = self._resolve_artifact_producer(item)
+        if producer.name != self.metadata.name:
```
alonmr, line 1504: same

mlrun/projects/project.py
```diff
@@ -3761,6 +3767,79 @@ def _validate_file_path(self, file_path: str, param_name: str):
                 f"<project.spec.get_code_path()>/<{param_name}>)."
             )
 
+    def _resolve_artifact_producer(
+        self,
+        artifact: typing.Union[str, Artifact, DatasetArtifact, ModelArtifact],
```
alonmr: not needed
```suggestion
        artifact: typing.Union[str, Artifact],
```

mlrun/projects/project.py
```diff
@@ -3761,6 +3767,79 @@ def _validate_file_path(self, file_path: str, param_name: str):
...
+        project_producer_tag=None,
```
alonmr: type hint

alonmr: it seems that you don't really need this param

mlrun/projects/project.py
```diff
@@ -3761,6 +3767,79 @@ def _validate_file_path(self, file_path: str, param_name: str):
...
+        return ArtifactProducer(
+            "project",
+            self.metadata.name,
+            self.metadata.name,
+            tag=project_producer_tag,
+        )
+
+    def _resolve_existing_artifact(
+        self, item, tag=None
+    ) -> typing.Optional[typing.Union[Artifact, DatasetArtifact, ModelArtifact]]:
```
alonmr: same comments about the type hints

mlrun/projects/project.py
```diff
@@ -3761,6 +3767,79 @@ def _validate_file_path(self, file_path: str, param_name: str):
...
+                    tag=item.tag,
+                    iter=item.iter,
+                    tree=item.tree,
+                )
+            if existing_artifact is not None:
+                return existing_artifact.from_dict(existing_artifact)
+        except mlrun.errors.MLRunNotFoundError as exc:
+            logger.debug(
+                "No existing artifact was found",
+                exc=exc,
```
alonmr: use `err_to_str`

mlrun/projects/project.py
```diff
@@ -1498,12 +1497,14 @@ def log_artifact(
...
-        producer = ArtifactProducer(
-            "project",
-            self.metadata.name,
-            self.metadata.name,
-            tag=self._get_hexsha() or str(uuid.uuid4()),
-        )
+        producer = self._resolve_artifact_producer(item)
+        if producer.name != self.metadata.name:
+            # the artifact producer is retained, so log it only if it doesn't already exist
+            if existing_artifact := self._resolve_existing_artifact(
```
alonmr, line 1506: what if the artifact changed and we want to update it?

tests/projects/test_project.py
```diff
@@ -900,6 +901,50 @@ def test_import_artifact_using_relative_path():
...
+    # we set the producer as dict so the export will work
+    artifact.producer = mlrun.artifacts.ArtifactProducer(
+        kind="run",
+        project=project_1.name,
+        name=run_name,
+        tag=run_tag,
+    ).get_meta()
+
+    # export the artifact
+    artifact_path = f"{str(assets_path())}/my-artifact.yaml"
```
alonmr: use `tests/test_results` instead of the assets path or a tmpdir

alonmr (commented):
A few comments, also missing test for when we change the producer to the new project producer

mlrun/projects/project.py
TomerShor: here it's won't work because I need the `existing_artifact` variable later to return it

mlrun/projects/project.py
TomerShor: I do because of https://github.com/mlrun/mlrun/pull/5283/files#diff-c972cfb48585c0ee85184b83808a9bf592a0ddd60a07cba789023649354f6cb8R1378
I want all artifacts in the list to have the same project tag

mlrun/projects/project.py
TomerShor: Removed it, I don't really care about the exception.

mlrun/projects/project.py
```diff
@@ -3761,6 +3767,79 @@ def _validate_file_path(self, file_path: str, param_name: str):
...
+                )
+
+        # the artifact is not registered, register it with the project as the producer
+        project_producer_tag = (
+            project_producer_tag or self._get_hexsha() or str(uuid.uuid4())
+        )
+        return ArtifactProducer(
+            "project",
+            self.metadata.name,
+            self.metadata.name,
```
```suggestion
            kind="project",
            project=self.metadata.name,
            name=self.metadata.name,
```

It's better to use named params, and especially in this case where it's not obvious from the actual parameters what each is used for.

mlrun/projects/project.py
```diff
@@ -3761,6 +3767,79 @@ def _validate_file_path(self, file_path: str, param_name: str):
...
+                existing_artifact = self.get_artifact(key=item, tag=tag)
+            else:
+                existing_artifact = self.get_artifact(
+                    key=item.key,
+                    tag=item.tag,
+                    iter=item.iter,
+                    tree=item.tree,
+                )
+            if existing_artifact is not None:
+                return existing_artifact.from_dict(existing_artifact)
```
theSaarco, line 3843: Is it even possible to get `None` here?

mlrun/artifacts/base.py
```diff
@@ -88,9 +88,10 @@ class ArtifactSpec(ModelObj):
         "db_key",
         "extra_data",
         "unpackaging_instructions",
+        "producer",
```
theSaarco, line 91: Why was this change needed? This means the `producer` will now be in the result of `base_dict()` for artifacts - is this intended?

mlrun/projects/project.py
```diff
@@ -3761,6 +3767,79 @@ def _validate_file_path(self, file_path: str, param_name: str):
...
+                return ArtifactProducer(
+                    name=producer_dict.get("name", ""),
+                    kind=producer_dict.get("kind", ""),
+                    project=producer_dict.get("project", ""),
+                    tag=producer_dict.get("tag", ""),
+                )
+
+        # the artifact is not registered, register it with the project as the producer
+        project_producer_tag = (
+            project_producer_tag or self._get_hexsha() or str(uuid.uuid4())
```
theSaarco: The `self._get_hexsha() or str(uuid.uuid4())` expression is now returning in at least 3 locations in the project class, looks like a good time to put it in a function/property.

theSaarco (commented):
Looks good. Some minor comments, mostly on style.

mlrun/artifacts/base.py
TomerShor: When setting an artifact on the project it is using the `base_dict()`, so the `producer` property was not exported because it was a part of the `_extra_fields`.
This is a crucial part to maintain the producers when loading a project - to actually have it on the exported spec.

mlrun/projects/project.py
TomerShor: That's a good question actually. 
As per @theSaarco we shouldn't re-register artifacts if we retain their producer, but indeed the spec etc can change.

mlrun/projects/project.py
TomerShor: It is now, see my comment in https://github.com/mlrun/mlrun/pull/5283/files#diff-c972cfb48585c0ee85184b83808a9bf592a0ddd60a07cba789023649354f6cb8R3338-R3339

mlrun/artifacts/base.py
theSaarco: So this means that this fix will not work for any project exported prior to this PR, right?
Also, need to make sure there are no BC issues there (I don't think there are, but needs to be verified).

mlrun/projects/project.py
```diff
@@ -3761,6 +3769,80 @@ def _validate_file_path(self, file_path: str, param_name: str):
...
+                existing_artifact = self.get_artifact(
+                    key=item.key,
+                    tag=item.tag,
+                    iter=item.iter,
+                    tree=item.tree,
+                )
+            if existing_artifact is not None:
+                return existing_artifact.from_dict(existing_artifact)
+        except mlrun.errors.MLRunNotFoundError:
+            logger.debug(
```
alonmr, line 3845: add the artifact key, tag etc.

mlrun/projects/project.py
theSaarco: Good question indeed.
Of course this is only expected if we're re-importing on the same env, since it's a run-id that already exists in the system. Even then, I tend to think it's a strange situation in which the same run-id had generated a different version of the same artifact. It seems more likely that the same artifact was modified after the run was done, in which case it would be probably better to maintain the manual changes done to the artifact.
I guess if the user wants to "reset" the artifact, it will need to be deleted before the import.

alonmr: ok, that makes sense. I suggest adding a log that says the artifact is already registered so we're skipping it.

alonmr (commented):
LGTM. Couple of logging requests

mlrun/artifacts/base.py
TomerShor: It obviously won't fix projects that were exported before this fix, so I'm not sure what BC issues we should check here..

theSaarco: It's not obvious that projects exported before the fix won't be fixed - it's the situation because the export doesn't contain the information needed for the fix.
As for BC, just to make sure that now that artifact export/import contains the producer field in the base export fields, that artifacts that were exported before won't fail importing due to it (again, I'm pretty sure they won't fail, just something to pay attention to).

mlrun/projects/project.py
```diff
@@ -1402,6 +1395,14 @@ def register_artifacts(self):
...
                         self.spec.get_code_path(), artifact.src_path
                     )
+                producer = self._resolve_artifact_producer(artifact, project_tag)
+                if (
+                    producer.name != self.metadata.name
+                    and self._resolve_existing_artifact(
+                        artifact,
+                    )
+                ):
+                    continue
```
liranbg:
```suggestion
                # ensures existing artifacts produced by others are not registered
                producer = self._resolve_artifact_producer(artifact, project_tag)
                if (
                    producer.name != self.metadata.name
                    and self._resolve_existing_artifact(
                        artifact,
                    )
                ):
                    continue
```

liranbg (commented):
really minor

PR #5294, opened 2024-03-17
[CI] Release bundle tutorials add welcome

daniels290813:
https://iguazio.atlassian.net/browse/IG-22441

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -264,7 +264,11 @@ jobs:
       - uses: actions/checkout@v3
       - name: Create tutorials tar
         run: |
-          tar -cvf mlrun-tutorials.tar docs/tutorials
+          mkdir jupyter-welcome
+          wget -c https://github.com/v3io/tutorials/blob/development/welcome.ipynb -o jupyter-welcome/welcome.ipynb
+          wget -c https://github.com/v3io/tutorials/blob/development/README.md -o jupyter-welcome/README.md
+          tar -cvf mlrun-tutorials.tar docs/tutorials jupyter-welcome
+          rm -rf jupyter-welcome
       - name: Add tutorials tar to release
         uses: ncipollo/release-action@v1
         with:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -264,11 +264,10 @@ jobs:
       - uses: actions/checkout@v3
       - name: Create tutorials tar
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -264,7 +264,11 @@ jobs:
       - uses: actions/checkout@v3
       - name: Create tutorials tar
         run: |
-          tar -cvf mlrun-tutorials.tar docs/tutorials
+          mkdir jupyter-welcome
```
alonmr: why create this dir instead of downloading the files into `docs/tutorials`?

.github/workflows/release.yaml
```diff
@@ -264,7 +264,11 @@ jobs:
...
+          wget -c https://github.com/v3io/tutorials/blob/development/welcome.ipynb -o jupyter-welcome/welcome.ipynb
```
alonmr: Pulling from development is dangerous. Someone can easily change the file names or whatever and break release without knowing. Let's create a new branch in `v3io/tutorials` and use that instead. Call it something like `mlrun-release-stable`.

.github/workflows/release.yaml
daniels290813: You are right that I can insert the new files into docs folder (not tutorials since we want README and welcome next to the tutorials folder and not inside).
@alonmr You think putting them in docs is better ?

.github/workflows/release.yaml
daniels290813: Alright no sweat, will wait @aviaIguazio to do that since i don't have permissions

.github/workflows/release.yaml
alonmr: let's use `docs/jupyter-welcome` then. Just because I think that using the root directory is not safe since we don't know what it contains (since we're using github runners).
Also, use `mkdir -p`

.github/workflows/release.yaml
daniels290813: @alonmr Will change branch soon

.github/workflows/release.yaml
daniels290813: README and welcome moved inside docs folder

.github/workflows/release.yaml
daniels290813: Changed branch to mlrun-release-stable

liranbg (approved):

alonmr (approved):

PR #5297, opened 2024-03-17
[Linting] Check copyright notice in Python file in `make lint`

jond01:
Opt-in the preview flake8-copyright CPY001 rule:
* https://docs.astral.sh/ruff/rules/#flake8-copyright-cpy
* https://docs.astral.sh/ruff/settings/#lintflake8-copyright

This allows faster feedback on missing copyright notice at the top of Python source files.

Note:
* Checks only Python files.
* It checks only for a generic header, not the full one.
* This is a preview rule.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -610,10 +610,10 @@ lint-imports: ## Validates import dependencies
 	lint-imports
 
 .PHONY: lint
-lint: fmt-check lint-imports ## Run lint on the code
+lint: lint-check lint-imports ## Run lint on the code
 
-.PHONY: fmt-check
-fmt-check: ## Check the code (using ruff)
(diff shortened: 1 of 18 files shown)


liranbg (commented):
Does that enforce copyright line or entire license block?
Also, does it mean we can remove the CI step using this linter?

jond01:
Please check out the description +

> Does that enforce copyright line or entire license block?

Only the copyright line, including the author:
https://docs.astral.sh/ruff/settings/#lint_flake8-copyright_notice-rgx

> Also, does it mean we can remove the CI step using this linter?

No, as it only checks Python source files. The CI step also checks YAML files, for example.

PR #5304, opened 2024-03-19
[Model Monitoring] Add MySQL implementation for the mm application framework

Eyal-Danieli:
As part of removing the V3IO resources dependency in the model monitoring application flow, in this PR we add support in SQL target for storing the model monitoring application results as well as the monitoring schedules values that are being managed by the monitoring controller.  

This implementation also includes adding an abstract layer for these operations, such as updating the last analyzed value for a certain application. This abstraction can be used in the future for additional databases such as PostgresSQL or Redis. 

In a more detailed overview, the PR includes the following changes:
- A new package called `db` that is stored under `mlrun.model_monitoring.db` and includes the required operations of the model monitoring databases. At the moment this package includes a sub-package called `stores` that handles the SQL/V3IO-KV storage. In the next phases, we will add another sub-package for the TSDB operations. 
-  As for the MySQL implementation, we added two new tables - `app_results` and `monitoring_schedules`, corresponded to the current V3IO-KV schemas. Note that even though these tables have no relationship with each other, they both includes a foreign key - `endpoint_id`.
- As for the CRUD operations, we applied the previous `SQLModelEndpointStore` operations and used them for the current structure. The new updated object layer called `SQLStoreBase`. In general, instead of using `get_model_endpoint_store()` to initialize a new store base object, we will use a new method called `get_store_object()`. The old `get_model_endpoint_store()` is deprecated and will be removed in 1.9.0.   
- Adjustments to support `Kafka Stream` that can be used instead of `V3IO Stream`. 
- Instantiate the `PushToMonitoringWriter` step using MLRun API instead of the application client side. That is important for getting the relevant stream path secret key, especially when using kafka. 

Jira - https://iguazio.atlassian.net/browse/ML-5955


mlrun/model_monitoring/db/stores/__init__.py
```diff
@@ -16,60 +16,61 @@
...
     v3io_nosql = "v3io-nosql"
     SQL = "sql"
 
-    def to_endpoint_store(
+    def to_object_store(
         self,
         project: str,
         access_key: str = None,
-        endpoint_store_connection: str = None,
+        store_connection: str = None,
```
davesh0812: As we discussed I think this the `store_connection` param need to be reach only from the env/secret and no as param.

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -49,7 +51,7 @@ def __init__(
         Initialize SQL store target object.
 
         :param project:               The name of the project.
-        :param sql_connection_string: Valid connection string or a path to SQL database with model endpoints table.
+        :param sql_connection_string: Valid connection string or a path to SQL database with model monitoring tables.
```
davesh0812: Also here `sql_connection_string` param need to be reach only from the env/secret and no as param.

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -62,22 +64,114 @@ def __init__(
             )
         )
 
-        self.table_name = (
+        self._engine = get_engine(dsn=self.sql_connection_string)
+
+    def _init_tables(self):
```
davesh0812, line 64: I saw that in all the action methods you run the relevant _init table method... can we just run this method one time from the class initialisation?

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -15,29 +15,31 @@
...
 import mlrun.model_monitoring.helpers
 from mlrun.common.db.sql_session import create_session, get_engine
 from mlrun.utils import logger
 
-from .model_endpoint_store import ModelEndpointStore
-from .models import get_model_endpoints_table
 
-
-class SQLModelEndpointStore(ModelEndpointStore):
+class SQLStoreBase(mlrun.model_monitoring.db.StoreBase):
```
davesh0812, line 32: Some general thought  - Are we checking somewhere if the user provide the relevant secret? If not we can probably do so in the `enable_model_monitoring` in the next pr..

mlrun/model_monitoring/db/stores/base/store.py
```diff
@@ -143,3 +143,41 @@ def get_endpoint_real_time_metrics(
         """
 
         pass
+
+    @abstractmethod
+    def write_application_result(self, event: dict[str, typing.Any]):
```
davesh0812, line 147: We don't to have also `get_application_result` in the abstraction? I think it can be necessary.. but we can do it in the future.

mlrun/model_monitoring/controller.py
```diff
@@ -65,27 +62,23 @@ def __init__(
...
-            )
-        except HttpResponseError as err:
+        monitoring_schedules = mlrun.model_monitoring.get_store_object(
+            project=self.project
+        )
+        last_analyzed = monitoring_schedules.get_last_analyzed(
+            endpoint_id=self._endpoint,
+            application_name=self._application,
+        )
+        if not last_analyzed:
```
davesh0812: Suggestion - Can you maybe raise  form the store object `MLRunNotFoundError` in case there the table don't contains the desired record.

mlrun/model_monitoring/db/stores/base/store.py
```diff
@@ -143,3 +143,41 @@ def get_endpoint_real_time_metrics(
...
+        Write a new application result event in the target table.
+
+        :param event: An event dictionary that represents the application result, should be corresponded to the
+                      schema defined in the :py:class:`~mlrun.common.schemas.model_monitoring.constants.WriterEvent`
+                      object.
+        """
+        pass
+
+    @abstractmethod
+    def get_last_analyzed(self, endpoint_id: str, application_name: str) -> int:
```
davesh0812, line 158: As discussed : In the sql this function creates new record with null but in the v3io it doesnt, I think that it should do the same.
I suggest that the update method would only update and to write also a `create_last_analyzed_record` method, that we will call from the controller in case the record is not created yet (Also I think it is better to raise an error in case we don't have a record as I wrote in another comment).

davesh0812, line 158: Future suggestion - add get_last_analyzed per endpoint only to improve our performance in the controller, and to access only one time per endpoint.

davesh0812 (commented):
Nice job!
Just a few suggestions, not sure if they needs to be part of this pr maybe for the future.

mlrun/common/schemas/model_monitoring/constants.py
```diff
@@ -177,6 +182,16 @@ class PrometheusMetric:
...
 
 
+class PrometheusEndpoints(StrEnum):
+    MODEL_MONITORING_METRICS = "/model-monitoring-metrics"
+    MONITORING_BATCH_METRICS = "/monitoring-batch-metrics"
+    MONITORING_DRIFT_STATUS = "/monitoring-drift-status"
+
+    @classmethod
+    def list(cls):
+        return list(map(lambda c: c.value, cls))
```
jond01: This logic is duplicated from `WriterEvent` - please create a shared class.

mlrun/db/base.py
```diff
@@ -509,9 +509,7 @@ def create_model_endpoint(
         self,
         project: str,
         endpoint_id: str,
-        model_endpoint: Union[
-            mlrun.model_monitoring.model_endpoint.ModelEndpoint, dict
-        ],
+        model_endpoint: Union[mlrun.model_monitoring.ModelEndpoint, dict],
```
jond01, line 512: Nice!

mlrun/model_monitoring/batch.py
```diff
@@ -32,6 +32,7 @@
 import mlrun.common.schemas.model_monitoring
 import mlrun.data_types.infer
 import mlrun.feature_store as fstore
+import mlrun.model_monitoring.db
```
jond01: Looks unnecessary, but this module is going to be removed anyway

mlrun/projects/project.py
```diff
@@ -40,6 +40,7 @@
 import mlrun.db
 import mlrun.errors
 import mlrun.k8s_utils
+import mlrun.model_monitoring.helpers
```
jond01: Looks unnecessary (?)

tests/api/crud/model_monitoring/test_model_endpoints.py
```diff
@@ -45,13 +49,25 @@ def _patch_external_resources() -> Iterator[None]:
...
+    mock.kv.get = Mock(side_effect=HttpResponseError)
+    with patch(
+        "mlrun.utils.v3io_clients.get_v3io_client",
+        return_value=mock,
+    ):
+        yield
+
+
 @pytest.mark.usefixtures("_patch_external_resources")
+@pytest.mark.usefixtures("mock_kv")
```
jond01: 1. Autouse in this scope is probably unwanted - it affects the whole module.
2. `@pytest.mark.usefixtures("mock_kv")` is redundant if it's an autoused fixture.
3. You may pass multiple fixtures to `@pytest.mark.usefixtures`.
4. You duplicated `tests/model_monitoring/test_helpers.py`.

tests/model_monitoring/test_helpers.py
```diff
@@ -169,7 +169,7 @@ def mock_kv() -> Iterator[None]:
         mock = Mock(spec=["kv"])
         mock.kv.get = Mock(side_effect=HttpResponseError)
         with patch(
-            "mlrun.model_monitoring.controller.get_v3io_client",
+            "mlrun.utils.v3io_clients.get_v3io_client",
             return_value=mock,
         ):
             yield
```
jond01, line 175: See the previous comment

mlrun/model_monitoring/db/stores/sqldb/__init__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2023 Iguazio
```
jond01: Nit: should probably be 2024. The same applies to all new files.

tests/model_monitoring/test_store.py
```diff
@@ -0,0 +1,185 @@
...
+#
+
+import string
+import time
+import typing
+from random import choice, randint
+from typing import Optional
+
+import pytest
+from _pytest.fixtures import FixtureRequest
```
jond01: Nit:
```suggestion
from pytest import FixtureRequest
```

mlrun/model_monitoring/controller.py
```diff
@@ -65,27 +62,23 @@ def __init__(
...
         The start and stop time are in seconds since the epoch.
         """
+        self.project = project
         self._endpoint = endpoint
         self._application = application
         self._first_request = first_request
-        self._kv_storage = get_v3io_client(
-            endpoint=mlrun.mlconf.v3io_api,
-            # Avoid noisy warning logs before the KV table is created
-            logger=create_logger(name="v3io_client", level="error"),
```
jond01, line 74: Why did you remove it from the new module? It will result in noisy logs.
Can the log level be set per request?

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -394,6 +395,103 @@ def get_endpoint_real_time_metrics(
...
+                ),
+                table_path=endpoint_id,
+                key=application_name,
+            )
+            return data.output.item[
+                mlrun.common.schemas.model_monitoring.SchedulingKeys.LAST_ANALYZED
+            ]
+        except v3io.dataplane.response.HttpResponseError as err:
+            logger.debug("Error while getting last analyzed time", err=err)
+            return None
```
jond01: See the previous comment

mlrun/model_monitoring/db/__init__.py
```diff
@@ -12,12 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from sqlalchemy.ext.declarative import declarative_base
+# flake8: noqa  - this is until we take care of the F401 violations with respect to __all__ & sphinx
```
jond01: Is it necessary?

jond01 (commented):
23.5 / 31 files viewed :)

I still need to review the MySQL parts.

mlrun/model_monitoring/db/stores/__init__.py
Eyal-Danieli: As we talked, the user should access the store object using the `get_store_object` that doesn't support this store_connection param. The reason I left it here it's only because we use through the tests to generate a dummy object directly. Anyway, we can remove it and adjust the tests but we can do it in a different PR.

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
Eyal-Danieli: similar answer, but in the bottom line we do enrich the connection string using the project secret

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
Eyal-Danieli: The thing is that sometimes we don't need to initialize all of the tables. For example, when we write an application result there is not reason to initialize the model endpoints table. It's just more efficient.

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
Eyal-Danieli: Although it the recommended approach to provide the connection string as a secret, we also support a pre-defined config (`mlrun.mlconf.model_endpoint_monitoring.endpoint_store_connection`) that can be used for the connection string. Therefore I don't see any reason why should we verify that the user has provided the connection string using the secret approach.

mlrun/model_monitoring/db/stores/base/store.py
Eyal-Danieli: agree + simliar answer for `list_application_result` and `update_application_result`. Anyway, as you suggested, it's a different PR.

mlrun/model_monitoring/db/stores/base/store.py
Eyal-Danieli: cool, in general the new SQL structure already support it (by passing an empty application name) but we need to do a similar implementation for the KV structure.

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
davesh0812: @Eyal-Danieli I meant that we need to check in the `enable_model_monitoring` that the user provide the relevant secret/env that appropriate to the env `mlrun.mlconf.model_endpoint_monitoring.store_type`... I not sure but I think we talked about it after your comment.

mlrun/common/schemas/model_monitoring/constants.py
Eyal-Danieli: added a new shared class - `MonitoringStrEnum`
https://github.com/mlrun/mlrun/pull/5304/files#:~:text=class%20MonitoringStrEnum(,%2C%20cls))

mlrun/projects/project.py
Eyal-Danieli: removed

tests/api/crud/model_monitoring/test_model_endpoints.py
Eyal-Danieli: good catch, it is redundant in this case. I've removed this fixture.

tests/model_monitoring/test_helpers.py
Eyal-Danieli: here it's not redundant... we have to mock this function otherwise the test will fail

tests/model_monitoring/test_store.py
Eyal-Danieli: fixed, and also updated a similar behaviour within test_writer.py

mlrun/model_monitoring/db/stores/sqldb/__init__.py
Eyal-Danieli: https://github.com/mlrun/mlrun/pull/5304/files#:~:text=%40%40%20%2D0%2C0%20%2B1%2C13%20%40%40-,%23%20Copyright%202024%20Iguazio,-%23

mlrun/model_monitoring/db/__init__.py
Eyal-Danieli: removed

mlrun/model_monitoring/controller.py
Eyal-Danieli: what noisy logs? in the app pod? with the `create_logger` func I still see a lot of logs. I tried to run the flow with and without the `create_logger` and didn't see any difference. Can you be more specific?

tests/api/crud/model_monitoring/test_model_endpoints.py
Eyal-Danieli: 1. autouse disabled 
2. autouse disabled :) 
3. fixed
4. here is part of the api tests while in `test_helpers` it's a basic unittest on the client side and also it's defined as a `staticmethod` of an existing object. We can't share this fixture unless we define it in some kind of a sharing conftest.py.

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
Eyal-Danieli: We agreed to add verification as part of adding the creds to the enable_model_monitoring API - https://iguazio.atlassian.net/jira/software/c/projects/ML/issues/ML-5963

mlrun/model_monitoring/controller.py
Eyal-Danieli: https://github.com/mlrun/mlrun/pull/5304/commits/ecae9118c61db11c9f124b552d65402e254a63e9

mlrun/model_monitoring/db/stores/base/store.py
Eyal-Danieli: fixed following our talk - https://github.com/mlrun/mlrun/commit/ecae9118c61db11c9f124b552d65402e254a63e9 
don't think it's required to create a separate API for `create_last_analyzed_record` because anyway we only doing that once through the SQL update process (V3IO KV call put which updates/creates in the same call)

tests/model_monitoring/test_store.py
```diff
@@ -0,0 +1,183 @@
...
+            WriterEvent.START_INFER_TIME: "2023-09-19 14:26:06.501084",
+            WriterEvent.END_INFER_TIME: "2023-09-19 16:26:06.501084",
+            WriterEvent.APPLICATION_NAME: "dummy-app",
+            WriterEvent.RESULT_NAME: "data-drift-0",
+            WriterEvent.RESULT_KIND: 0,
+            WriterEvent.RESULT_VALUE: 0.32,
+            WriterEvent.RESULT_STATUS: request.param,
+            WriterEvent.RESULT_EXTRA_DATA: "",
+        }
+    )
```
jond01: Nit:
You don't use parametrization, so remove this:
```suggestion
@pytest.fixture()
def event(request: pytest.FixtureRequest) -> _AppResultEvent:
    return _AppResultEvent(
        {
            WriterEvent.ENDPOINT_ID: "some-ep-id",
            WriterEvent.START_INFER_TIME: "2023-09-19 14:26:06.501084",
            WriterEvent.END_INFER_TIME: "2023-09-19 16:26:06.501084",
            WriterEvent.APPLICATION_NAME: "dummy-app",
            WriterEvent.RESULT_NAME: "data-drift-0",
            WriterEvent.RESULT_KIND: 0,
            WriterEvent.RESULT_VALUE: 0.32,
            WriterEvent.RESULT_STATUS: 0,
            WriterEvent.RESULT_EXTRA_DATA: "",
        }
    )
```

mlrun/model_monitoring/db/stores/__init__.py
```diff
@@ -16,60 +16,61 @@
...
 import mlrun.common.schemas.secret
 import mlrun.errors
 
-from .model_endpoint_store import ModelEndpointStore
+from .base import StoreBase
 
 
-class ModelEndpointStoreType(enum.Enum):
-    """Enum class to handle the different store type values for saving a model endpoint record."""
+class ObjectStoreType(enum.Enum):
```
jond01: I suggest renaming this class to `ObjectStoreFactory`, as it's more indicative.

mlrun/model_monitoring/controller.py
jond01: Okay, maybe it's resolved in later versions. We had an issue with that in the past.

mlrun/model_monitoring/db/stores/__init__.py
```diff
@@ -16,60 +16,61 @@
...
         :param secret_provider:           An optional secret provider to get the connection string secret.
 
-        :return: `ModelEndpointStore` object.
+        :return: `StoreBase` object.
 
         """
 
-        if self.value == ModelEndpointStoreType.v3io_nosql.value:
-            from .kv_model_endpoint_store import KVModelEndpointStore
+        if self.value == ObjectStoreType.v3io_nosql.value:
```
jond01: Suggestion - why not simply:
```suggestion
        if self == self.v3io_nosql:
```

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -256,6 +349,142 @@ def list_model_endpoints(
...
+        """
+        self._init_monitoring_schedules_table()
+        application_filter_dict = self.filter_endpoint_and_application_name(
+            endpoint_id=endpoint_id, application_name=application_name
+        )
+        monitoring_schedule_record = self._get(
+            table=self.MonitoringSchedulesTable, **application_filter_dict
+        )
+        if not monitoring_schedule_record:
+            raise mlrun.errors.MLRunNotFoundError(
```
jond01, line 434: I suggest using a different error class, specific to this case.
`MLRunNotFoundError` better suits a faulty response from the API server.

tests/api/crud/model_monitoring/test_model_endpoints.py
jond01: Regarding 4 - I think it's better not to duplicate, but I can tolerate this one :)

tests/model_monitoring/test_store.py
```diff
@@ -0,0 +1,183 @@
...
+        application_name=event[WriterEvent.APPLICATION_NAME],
+    )
+
+    assert last_analyzed == epoch_time
+
+    # Clean resources
+    sql_store.delete_last_analyzed(mock_endpoint_1.metadata.uid)
+    _clean_model_endpoint_from_db(
+        store=sql_store, endpoint_id=mock_endpoint_1.metadata.uid
+    )
```
jond01: If it's required to not pollute sequential runs of this test, I think it should be a context manager, possibly in a fixture.

mlrun/model_monitoring/db/stores/sqldb/models/base.py
```diff
@@ -82,3 +94,64 @@ class ModelEndpointsBaseTable(BaseModel):
...
+    )
+
+    result_name = Column(
+        WriterEvent.RESULT_NAME,
+        String(40),
+    )
+
+    start_infer_time = Column(
+        WriterEvent.START_INFER_TIME,
+        TIMESTAMP,
```
jond01: Can you set it to `TIMESTAMP(timezone=True)` for TZ support? (applies to all the relevant cols)
https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.TIMESTAMP

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -394,6 +395,128 @@ def get_endpoint_real_time_metrics(
...
+                "nullable": False,
+            }
+        ]
+        res = self.client.kv.create_schema(
+            container=v3io_monitoring_apps_container,
+            table_path=endpoint_id,
+            key=mlrun.common.schemas.model_monitoring.WriterEvent.APPLICATION_NAME,
+            fields=fields,
+        )
+        if res.status_code != HTTPStatus.OK.value:
```
jond01: I believe that's redundant as `HTTPStatus` is an `IntEnum`:
```suggestion
        if res.status_code != HTTPStatus.OK:
```

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -11,33 +11,34 @@
...
 # limitations under the License.
-#
 
 import json
 import typing
+import uuid
 from datetime import datetime, timezone
 
 import pandas as pd
 import sqlalchemy as db
```
jond01: I suggest removing this alias, as it's somewhat confusing with the new `mlrun.model_monitoring.db` subpackage.
```suggestion
import sqlalchemy
```

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -87,21 +180,18 @@ def write_model_endpoint(self, endpoint: dict[str, typing.Any]):
...
-
-            # Convert the result into a pandas Dataframe and write it into the database
-            endpoint_df = pd.DataFrame([endpoint])
+        # Adjust timestamps fields
+        endpoint[mlrun.common.schemas.model_monitoring.EventFieldType.FIRST_REQUEST] = (
+            datetime.now(timezone.utc)
+        )
+        endpoint[mlrun.common.schemas.model_monitoring.EventFieldType.LAST_REQUEST] = (
+            datetime.now(timezone.utc)
+        )
```
jond01: I suggest using `mlrun.utils.datetime_now()` (equivalent) + 
```suggestion
        # Adjust timestamps fields
        endpoint[mlrun.common.schemas.model_monitoring.EventFieldType.FIRST_REQUEST] = endpoint[mlrun.common.schemas.model_monitoring.EventFieldType.LAST_REQUEST] = (
            mlrun.utils.datetime_now()
        )
```

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -256,6 +349,142 @@ def list_model_endpoints(
...
+            self._update(
+                attributes=event,
+                table=self.ApplicationResultsTable,
+                **application_filter_dict,
+            )
+        else:
+            # Write a new application result
+            event[mlrun.common.schemas.model_monitoring.EventFieldType.UID] = (
+                uuid.uuid4().hex
+            )
```
jond01: What is this UUID used for?

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -256,6 +349,142 @@ def list_model_endpoints(
...
+        )
+        monitoring_schedule_record = self._get(
+            table=self.MonitoringSchedulesTable, **application_filter_dict
+        )
+        if not monitoring_schedule_record:
+            # Add a new record with empty last analyzed value
+            self._write(
+                table=mlrun.common.schemas.model_monitoring.FileTargetKind.MONITORING_SCHEDULES,
+                event={
+                    mlrun.common.schemas.model_monitoring.SchedulingKeys.UID: uuid.uuid4().hex,
```
jond01, line 464: Same Q :)

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -333,6 +562,26 @@ def _validate_labels(
...
 
+    @staticmethod
+    def filter_endpoint_and_application_name(
+        endpoint_id: str, application_name: str
+    ) -> dict[str, str]:
+        """Generate a dictionary filter for endpoint id and application name"""
+        if not endpoint_id and not application_name:
+            raise mlrun.errors.MLRunNotFoundError(
+                "Please provide a valid endpoint_id and/or application_name"
+            )
```
jond01:
```suggestion
        if not endpoint_id and not application_name:
            raise ValueError(
                "Please provide a valid endpoint_id and/or application_name"
            )
```

jond01 (commented):
Great job! I have some more comments.

mlrun/model_monitoring/controller.py
```diff
@@ -301,7 +296,7 @@ def __init__(
             f"Initializing {self.__class__.__name__}", project=project
         )
 
-        self.db = mlrun.model_monitoring.get_model_endpoint_store(project=project)
+        self.db = mlrun.model_monitoring.get_store_object(project=project)
```
davesh0812, line 295: As discussed try to pass this db session to the _BatchWindow & _BatchWindowGenerator instead of using get_store_object in each _BatchWindow run... I hope it wont make issues.

mlrun/model_monitoring/controller.py
```diff
@@ -124,11 +117,13 @@ def _update_last_analyzed(self, last_analyzed: int) -> None:
             application=self._application,
             last_analyzed=last_analyzed,
         )
-        self._kv_storage.put(
-            container=self._v3io_container,
-            table_path=self._endpoint,
-            key=self._application,
-            attributes={mm_constants.SchedulingKeys.LAST_ANALYZED: last_analyzed},
+        monitoring_schedules = mlrun.model_monitoring.get_store_object(
```
davesh0812: related to the prev comment

mlrun/model_monitoring/db/stores/sqldb/models/__init__.py
```diff
@@ -25,3 +29,21 @@ def get_model_endpoints_table(
...
     return SQLiteModelEndpointsTable
+
+
+def get_application_result_table(
+    connection_string: Optional[str] = None,
+) -> Union[type[MySQLApplicationResultTable], type[SQLiteApplicationResultTable]]:
+    """Return ModelEndpointsTable based on the provided connection string"""
+    if connection_string and "mysql:" in connection_string:
+        return MySQLApplicationResultTable
+    return SQLiteApplicationResultTable
```
davesh0812, line 40:
```suggestion
    connection_string = connection_string or mlrun.model_monitoring.helpers.get_connection_string(
                secret_provider=secret_provider
            )
    if connection_string and "mysql:" in connection_string:
        return MySQLApplicationResultTable
    return SQLiteApplicationResultTable
```

mlrun/model_monitoring/db/stores/sqldb/models/__init__.py
```diff
@@ -25,3 +29,21 @@ def get_model_endpoints_table(
...
+
+
+def get_monitoring_schedules_table(
+    connection_string: Optional[str] = None,
+) -> Union[type[MySQLMonitoringSchedulesTable], type[SQLiteMonitoringSchedulesTable]]:
+    """Return ModelEndpointsTable based on the provided connection string"""
+    if connection_string and "mysql:" in connection_string:
+        return MySQLMonitoringSchedulesTable
+    return SQLiteMonitoringSchedulesTable
```
davesh0812: same as prev

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
Eyal-Danieli: I prefer to keep consistency with MLRun in this case. You can have a look at `server.api.db.sqldb.db.py` and see that we are using the same error on empty get calls. Anyway we can discuss and apply a new behaviour in a different PR.

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
Eyal-Danieli: For enrich the new `uid` primary key column in the `app_result` table. We can't use the app name as a primary key because it's not unique. An alternative can be using app_name+endpoint_id as a composite key but I preferred to use an internal UID and the other columns (endpoint_id & app_name) as foreign keys to ensure data integrity.

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
Eyal-Danieli: Same A :)

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
Eyal-Danieli: instead of `ValueError` I will use `mlrun.errors.MLRunBadRequestError` because it's indeed a bad request error...

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
jond01: I see.
In that case, why not use the more standard auto-increment in the table:
https://docs.sqlalchemy.org/en/14/core/metadata.html#sqlalchemy.schema.Column.params.autoincrement

Eyal-Danieli: Just an update - the new `uid` column will be represented by the endpoint_id, application name and result name. With that, we can optimize the update query each time we reach the DB to update an application record.
In other words, the new `uid` is basically `<endpoint_id>_<app_name>_<result_name>`.

mlrun/model_monitoring/controller.py
Eyal-Danieli: Updated. It work partially because we can't pass the same db session into different processes. However, we can use a single DB session for all the required operations within a single process (e.g. get_last_analyzed + update_last_analyzed).

mlrun/model_monitoring/db/stores/sqldb/models/__init__.py
Eyal-Danieli: as we discussed, we removed the connection_string argument from the sql_store + update the get tables functions to be a private functions.

tests/model_monitoring/test_store.py
```diff
@@ -0,0 +1,264 @@
...
+                model=f"model_{randint(0, 100)}:v{randint(0, 100)}",
+                model_class="classifier",
+            ),
+            status=mlrun.common.schemas.ModelEndpointStatus(state=state),
+        )
+
+    @staticmethod
+    @pytest.fixture
+    def event(
+        request: pytest.FixtureRequest,
```
jond01: Minor: this is not used:
```suggestion
```

tests/model_monitoring/test_store.py
```diff
@@ -0,0 +1,264 @@
...
+                WriterEvent.RESULT_VALUE: 0.32,
+                WriterEvent.RESULT_STATUS: 0,
+                WriterEvent.RESULT_EXTRA_DATA: "",
+            }
+        )
+
+    @staticmethod
+    @pytest.fixture
+    def event_v2(
+        request: pytest.FixtureRequest,
```
jond01: Same:
```suggestion
```

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -0,0 +1,662 @@
...
+        :param access_key:       V3IO access key that will be used for generating Frames client object. If not
+                                 provided, the access key will be retrieved from the environment variables.
+
+        :return: A dictionary of metrics in which the key is a metric name and the value is a list of tuples that
+                 includes timestamps and the values.
+        """
+        # # TODO : Implement this method once Perometheus is supported
+        logger.warning(
+            "Real time metrics service using Prometheus will be implemented in 1.4.0"
+        )
```
jond01, line 660: Probably an outdated message 🤔

davesh0812 (approved):

jond01 (approved):
Well done!
Next time - if possible, please try to separate such a big change into smaller chunks. It can help both the you and the reviewers :)

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
Eyal-Danieli: yes anyway we are going to address it in 1.7.0 so I leave it for now

PR #5327, opened 2024-03-26
[Nuclio] API Gateway authentication improvements

rokatyy:
This PR implements part of https://iguazio.atlassian.net/browse/ML-5372: 
* Implements `with_basic_auth`
* Use HTTPBasicAuth instead of custom header generation when invoke gateway
* Pass authentication interface instead of (authmode, authentication)

### DIFF
--- a/mlrun/runtimes/nuclio/api_gateway.py
+++ b/mlrun/runtimes/nuclio/api_gateway.py
@@ -51,6 +51,11 @@ class APIGatewayAuthenticator(typing.Protocol):
         else:
             return NoneAuth()
 
+    def to_scheme(
+        self,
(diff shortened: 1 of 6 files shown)


mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -28,6 +30,56 @@
 PROJECT_NAME_LABEL = "nuclio.io/project-name"
 
 
+class APIGatewayAuthenticator(typing.Protocol):
```
liranbg, line 33: [TIL](https://peps.python.org/pep-0544/)! 🚀

liranbg (approved):
👏🏼

PR #5332, opened 2024-03-27
[Projects] Enrich function node selector from project node selector

yaelgen:
follow up for: [ML-3874](https://iguazio.atlassian.net/browse/ML-3874)
This PR introduces an enrichment of the function node selector with the project's default function node selector, as per discussions on the ticket we've decided to enrich only with the project level, ignoring the common/config level. 

[ML-3874]: https://iguazio.atlassian.net/browse/ML-3874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/runtimes/pod.py
```diff
@@ -1057,6 +1057,12 @@ def is_env_exists(self, name):
                 return True
         return False
 
+    def enrich_runtime_spec(
+        self,
+        project_node_selector: dict[str, str],
+    ):
+        self.spec.node_selector = {**project_node_selector, **self.spec.node_selector}
+
```
quaark: This merge logic is very ambiguous, it might be clear to someone who knows the `**` notation very well and how it decides orders of precedence. But it's worth leaving here a very detailed description on how this merge behaves and what it does.
If this merge exists elsewhere in the same ambiguous fashion, lets create a mlrun helper function `merge_dicts` with the detailed description there

quaark (changes requested):
Looks good code-wise!
The code needs a bit more documentation though

liranbg (approved):
Missing some runtimes such as - Nuclio, Dask, MPIjob, Spark, Databricks job
Can be done in a follow up PR - thus approving this one

PR #5333, opened 2024-03-27
[Pagination] Phase II - Pagination Cache Lifecycle

quaark:
https://iguazio.atlassian.net/browse/ML-5802

- Add pagination cache CRUD, monitor function and cleanup function:
  - monitor: with ttl and max table size from config:
    - Queries the pagination cache records with last accessed time older than now - ttl and deletes them
    - If the remaining amount of records is larger than max table size, delete oldest records to bring us down to that size
  - cleanup: removes all pagination cache records
- pagination cache monitor runs periodically with interval from config
- pagination cache cleanup runs at startup if db isn't from scratch (empty/new)
- UTs

* **TODO**: Add the final parts of the cache lifecycle (requires code from next phases):
  * create new record when receiving paginated requests
  * delete record when paginated request reaches last page


server/api/crud/pagination_cache.py
```diff
@@ -0,0 +1,95 @@
...
+        db.list_paginated_query_cache_record(session, as_query=True).delete()
+        session.commit()
+
+    @staticmethod
+    def monitor_pagination_cache(session: sqlalchemy.orm.Session):
+        """
+        Monitor the pagination cache and remove records that are older than the cache TTL, and if the cache table
+        reached the max size, remove the oldest records.
+        """
+        cache_ttl = mlrun.config.config.httpdb.pagination_cache.ttl
```
liranbg: I would delete cache_ttl +1 to avoid 0 as a second that being used and removed at same time
e.g at 0 seconds and 400ms used would use the pagination while 0 is considered as time expired

server/api/crud/pagination_cache.py
```diff
@@ -0,0 +1,95 @@
...
+    @staticmethod
+    def delete_pagination_cache_record(session: sqlalchemy.orm.Session, key: str):
+        db = server.api.utils.singletons.db.get_db()
+        db.delete_paginated_query_cache_record(session, key)
+
+    @staticmethod
+    def cleanup_pagination_cache(session: sqlalchemy.orm.Session):
+        db = server.api.utils.singletons.db.get_db()
+        db.list_paginated_query_cache_record(session, as_query=True).delete()
+        session.commit()
```
liranbg: why do you need to cmmit here? why not callee commit on the session?

server/api/crud/pagination_cache.py
```diff
@@ -0,0 +1,95 @@
...
+            - datetime.timedelta(seconds=cache_ttl),
+            as_query=True,
+        ).delete()
+
+        all_records_query = db.list_paginated_query_cache_record(session, as_query=True)
+        table_size = all_records_query.count()
+        if table_size > table_max_size:
+            records = all_records_query.limit(table_size - table_max_size)
+            for record in records:
+                db.delete_paginated_query_cache_record(session, record.key)
```
liranbg: that could be very slow, isnt it? why can't expose a function to delete multiple keys at once? 
e.g.: limit by 20, delete 20 queries at once?

liranbg (changes requested):

server/api/crud/pagination_cache.py
quaark: I tried, you get an exception from sql alchemy that you can't call delete on a query that used limit. I have no idea why it won't let us do that :(

server/api/crud/pagination_cache.py
quaark: I don't understand the question. I am calling commit on the session

server/api/crud/pagination_cache.py
quaark: managed to fix it to a single commit though at least

server/api/crud/pagination_cache.py
liranbg: why do you commit here? why not let the callee of `cleanup_pagination_cache` or the session manager commit once done with the session?

liranbg (approved):

PR #5334, opened 2024-03-27
[ModelObj] Filter warnings when using to_dict

alonmr:
https://iguazio.atlassian.net/browse/ML-6034

When part of the spec fields are deprecated we may still be showing them as part of the object dictionary but they are not really used by the user.
The attribute `clone_target_dir` is a reference to `build.source_code_target_dir` and if the latter has a value, the former will warn when using `to_dict`.


liranbg (approved):
❤️

PR #5335, opened 2024-03-27
[Pipelines] Fix get workflow id timeouts and make it configurable per run

alonmr:
https://iguazio.atlassian.net/browse/ML-6026

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -2748,7 +2748,6 @@ class MlrunProject(ModelObj):
         source: str = None,
         cleanup_ttl: int = None,
         notifications: list[mlrun.model.Notification] = None,
-        get_workflow_id_timeout: int = None,
     ) -> _PipelineRunStatus:
         """Run a workflow using kubeflow pipelines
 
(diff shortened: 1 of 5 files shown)


mlrun/projects/project.py
```diff
@@ -2748,41 +2748,39 @@ def run(
...
-                          list of notifications to send for workflow completion
-        :returns: run id
+                          For using the pre-defined workflow's schedule, set `schedule=True`
+        :param timeout:   Timeout in seconds to wait for pipeline completion (watch will be activated)
+        :param source:    Remote source to use instead of the actual `project.spec.source` (used when engine is remote).
+                          For other engines the source is to validate that the code is up-to-date
+        :param cleanup_ttl: Pipeline cleanup ttl in secs (time to wait after workflow completion, at which point the
+                            workflow and all its resources are deleted)
+        :param notifications:           List of notifications to send for workflow completion
+        :param get_workflow_id_timeout: Timeout in seconds to wait for the workflow to start and get its ID.
```
liranbg: why do we need it as a parameter if it can be used by
a. envvar (assumption)
b. override config py directly?

alonmr: can override the config of course, I thought it would be more intuitive as a param

liranbg (approved):

PR #5337, opened 2024-03-28
[Pagination] Phase III - Paginator Logic & Finalize Pagination Cache Lifecycle

quaark:
https://iguazio.atlassian.net/browse/ML-5803

Implement paginator wrapper logic that wraps paginated functions with handling the cached record CRUD.

When requesting to paginate a request:
- If the method isn't supported raise NotImplemented
- If no page size or token is given, return all items - no pagination
- If token was given, check db cache records
  - Record doesn't exist? - 404
  - Record belongs to other user? - 403
  - Update record's last_access time and current page
- Call supported method with page number, page size
  - method raised StopIteration? - Clear cache record from db
- return response with pagination info (token, page number, page size)

### DIFF
--- a/server/api/utils/pagination.py
+++ b/server/api/utils/pagination.py
@@ -40,10 +40,6 @@ class PaginatedMethods:
     def get_method(cls, method_name: str) -> typing.Callable:
         return cls._method_map[method_name]
 
-    @staticmethod
-    def get_name(method: typing.Callable) -> str:
-        return method.__name__
-
 
 class Paginator(metaclass=mlrun.utils.singleton.Singleton):
     def __init__(self):
--- a/server/api/utils/pagination.py
+++ b/server/api/utils/pagination.py
@@ -1,0 +1,148 @@
+# Copyright 2023 Iguazio
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


liranbg (approved):

server/api/utils/pagination.py
```diff
@@ -0,0 +1,148 @@
...
+import server.api.crud
+from mlrun.utils import logger
+
+
+class PaginatedMethods:
+    _methods: list[typing.Callable] = [
+        # TODO: add methods when they implement pagination
+        # server.api.crud.Runs().list_runs,
+    ]
+    _method_map = {method.__name__: method for method in _methods}
```
theSaarco, line 31: If you implemented `get_name`, why not use it here?

tests/api/utils/test_pagination.py
```diff
@@ -0,0 +1,363 @@
...
+    with pytest.raises(mlrun.errors.MLRunAccessDeniedError):
+        paginator.paginate_request(
+            db, paginated_method, auth_info_2, pagination_info["token"]
+        )
+
+    logger.info(
+        "Requesting second page without auth info, should raise AccessDeniedError"
+    )
+    with pytest.raises(mlrun.errors.MLRunAccessDeniedError):
+        paginator.paginate_request(db, paginated_method, None, pagination_info["token"])
```
theSaarco, line 211: What about cleaning up the pagination cache? Aren't you concerned about pagination cache leaks across tests?

theSaarco (commented):
Nice work. Have some small comments/questions.

tests/api/utils/test_pagination.py
quaark: Not really as we delete the db when the test goes down. But can add it to the teardown

server/api/utils/pagination.py
quaark: Sure 👍

quaark: Spoke offline removed get_name

PR #5341, opened 2024-03-28
[Application] Support building the application image

alonmr:
# Building application image with MLRun
Upon deployment of an application MLRun will build the sidecar image if it requires building.
Once the build is complete, we will configure the sidecar and reverse proxy.
The last step is to deploy the function in nuclio.

## Usage example
```
# Specify source to be loaded on build time or run time
application = project.set_function(name="my-vizro-dashboard", image="my-base/image", kind="application", requirements=["vizro"])

# Provide code artifacts 
application.with_source_archive("git://github.com/org/repo#my-branch", pull_at_runtime=False)

# Build the application image via MLRun and deploy the Nuclio function
# Optionally add mlrun
application.deploy(with_mlrun=False)
```

## Implementation details

Build image code moved from `KubeJobRuntime` to `BaseRuntime` for it to be available to nuclio runtimes and application runtime in particular.
New endpoints for nuclio deployment in order to separate nuclio deployment and mlrun build flows as application runtime uses both.

https://iguazio.atlassian.net/browse/ML-5965


mlrun/db/httpdb.py
```diff
@@ -1393,6 +1394,8 @@ def get_builder_status(
         :param last_log_timestamp:  Last timestamp of logs that were already retrieved. Function will return only logs
                                     later than this parameter.
         :param verbose:             Add verbose logs into the output.
+        :param mlrun_build:         Whether the build is an MLRun build. If not provided, will be inferred from the
```
liranbg: so why explicit? I think that the abstraction is good enough. adding another parameter makes the function too complicated.

mlrun/db/httpdb.py
```diff
@@ -1384,6 +1384,7 @@ def get_builder_status(
         logs: bool = True,
         last_log_timestamp: float = 0.0,
         verbose: bool = False,
+        mlrun_build: Optional[bool] = None,
```
liranbg: imo can be removed

mlrun/db/base.py
```diff
@@ -641,6 +641,7 @@ def get_builder_status(
         logs: bool = True,
         last_log_timestamp: float = 0.0,
         verbose: bool = False,
+        mlrun_build: Optional[bool] = None,
```
liranbg: imo can be removed

mlrun/launcher/client.py
```diff
@@ -47,7 +47,9 @@ def prepare_image_for_deploy(runtime: "mlrun.runtimes.BaseRuntime"):
         If build is needed, set the image as the base_image for the build.
         If image is not given set the default one.
         """
-        if runtime.kind in mlrun.runtimes.RuntimeKinds.nuclio_runtimes():
+        if runtime.kind in set(
+            mlrun.runtimes.RuntimeKinds.nuclio_runtimes()
+        ).difference(set(mlrun.runtimes.RuntimeKinds.application)):
```
liranbg: why do you omit application from the kind? worth specifying

mlrun/runtimes/pod.py
```diff
@@ -1312,6 +1313,156 @@ def validate_and_enrich_service_account(
...
+            and not build.extra
+            and with_mlrun
+        ):
+            logger.info(
+                "Running build to add mlrun package, set "
+                "with_mlrun=False to skip if its already in the image"
+            )
+        return with_mlrun
+
+    def _build_image(
```
liranbg, line 1392: why isnt it on mlrun server side?

liranbg (commented):
nicely done!
missing a basic system test / documentation

mlrun/launcher/client.py
theSaarco: Even further on this - can you give this reduced set a meaningful designation? For example "pure nuclio runtimes" or something? If so, I would put this in another accessor in `RuntimeKinds` rather than calculate it here.

server/api/api/endpoints/functions.py
```diff
@@ -718,16 +719,26 @@ def _build_function(
...
 
-        is_nuclio_runtime = fn.kind in RuntimeKinds.nuclio_runtimes()
+        # TODO:  force_build is only used for mlrun build flow, for nuclio runtimes users can specify force_build=True,
+        #  but it is redundant and ignored. Once force_build is removed from RemoteRuntime.deploy() in 1.9.0,
+        #  we can simplify the logic here to:
+        #  is_nuclio_deploy = fn.kind in RuntimeKinds.nuclio_runtimes() and not force_build
+        is_nuclio_deploy = False
+        if fn.kind in set(RuntimeKinds.nuclio_runtimes()).difference(
+            {RuntimeKinds.application}
+        ):
```
theSaarco: Since this is used multiple times, I would certainly move it to another `RuntimeKinds` function.

mlrun/runtimes/nuclio/function.py
```diff
@@ -543,11 +543,23 @@ def deploy(
         :param project:    project name
         :param tag:        function tag
         :param verbose:    set True for verbose logging
-        :param auth_info:  service AuthInfo
+        :param auth_info:  service AuthInfo (deprecated)
```
theSaarco:
```suggestion
        :param auth_info:  service AuthInfo (deprecated and ignored)
```
This should be noted - deprecation may mean that it's being phased out but still used. In this case it's simply ignored.

mlrun/runtimes/nuclio/function.py
```diff
@@ -543,11 +543,23 @@ def deploy(
...
         :param builder_env: env vars dict for source archive config/credentials e.g. builder_env={"GIT_TOKEN": token}
-        :param force_build: set True for force building the image
+        :param force_build: set True for force building the image (deprecated)
```
theSaarco: same.

theSaarco (commented):
Partial review.

mlrun/runtimes/pod.py
alonmr: this is client code, it calls the remote builder with httpdb

mlrun/launcher/client.py
alonmr: 👍

mlrun/db/httpdb.py
alonmr: This is needed to determine which builder status to get, nuclio or mlrun builders. For application runtime we need both and there is no way to tell from the server in which flow we are.

server/api/api/endpoints/functions.py
```diff
@@ -428,7 +429,7 @@ async def build_status(
         )
 
     # nuclio deploy status
-    if fn.get("kind") in RuntimeKinds.nuclio_runtimes():
+    if not mlrun_build and fn.get("kind") in RuntimeKinds.nuclio_runtimes():
```
liranbg: I think we can omit this variable and build the image using a dummy-function to simplify the flow

server/api/api/endpoints/functions.py
```diff
@@ -718,16 +719,24 @@ def _build_function(
...
 
-        is_nuclio_runtime = fn.kind in RuntimeKinds.nuclio_runtimes()
+        # TODO:  force_build is only used for mlrun build flow, for nuclio runtimes users can specify force_build=True,
+        #  but it is redundant and ignored. Once force_build is removed from RemoteRuntime.deploy() in 1.9.0,
+        #  we can simplify the logic here to:
+        #  is_nuclio_deploy = fn.kind in RuntimeKinds.nuclio_runtimes() and not force_build
+        is_nuclio_deploy = False
+        if fn.kind in RuntimeKinds.pure_nuclio_deployed_runtimes():
+            is_nuclio_deploy = True
+        elif fn.kind == RuntimeKinds.application and not force_build:
```
liranbg: I think it worth leaving the force build in because nuclio does support cache invalidation - we just need to make sure we pass it along

server/api/crud/runtimes/nuclio/function.py
```diff
@@ -172,6 +172,22 @@ def get_nuclio_deploy_status(
         return state, address, name, last_log_timestamp, text, function_status
 
 
+def pure_nuclio_deployed_restricted():
```
liranbg, line 175: document the idea behind the reason why it is needed

liranbg (commented):
Food for thoughts

tests/api/runtimes/test_application.py
```diff
@@ -0,0 +1,83 @@
...
+        return "application"
+
+    def _execute_run(self, runtime, **kwargs):
+        # deploy_nuclio_function doesn't accept watch, so we need to remove it
+        kwargs.pop("watch", None)
+        server.api.crud.runtimes.nuclio.function.deploy_nuclio_function(
+            runtime, **kwargs
+        )
+
+    def _generate_runtime(
```
liranbg: move private under

liranbg (approved):

mlrun/db/httpdb.py
```diff
@@ -1380,6 +1366,95 @@ def remote_builder(
...
+                req["builder_env"] = builder_env
+            _path = f"projects/{func.metadata.project}/functions/{func.metadata.name}/deploy"
+            resp = self.api_call("POST", _path, json=req)
+        except OSError as err:
+            logger.error(f"error submitting deploy task: {err_to_str(err)}")
+            raise OSError(f"error: cannot submit deploy, {err_to_str(err)}")
+
+        if not resp.ok:
+            logger.error(f"bad resp!!\n{resp.text}")
+            raise ValueError("bad function run response")
```
theSaarco:
```suggestion
            logger.error(f"deploy nuclio - bad response:\n{resp.text}")
            raise ValueError("bad nuclio deploy response")
```

mlrun/db/httpdb.py
```diff
@@ -1380,6 +1366,95 @@ def remote_builder(
...
+        try:
+            req = {
+                "function": func.to_dict(),
+            }
+            if builder_env:
+                req["builder_env"] = builder_env
+            _path = f"projects/{func.metadata.project}/functions/{func.metadata.name}/deploy"
+            resp = self.api_call("POST", _path, json=req)
+        except OSError as err:
+            logger.error(f"error submitting deploy task: {err_to_str(err)}")
```
theSaarco:
```suggestion
            logger.error(f"error submitting deploy nuclio task: {err_to_str(err)}")
```

server/api/crud/model_monitoring/deployment.py
```diff
@@ -246,6 +247,73 @@ def deploy_model_monitoring_writer_application(
...
+                    else:
+                        stream_args = (
+                            config.model_endpoint_monitoring.serving_stream_args
+                        )
+                        access_key = os.environ.get("V3IO_ACCESS_KEY")
+                        kwargs = {}
+                    if mlrun.mlconf.is_explicit_ack(version=resolve_nuclio_version()):
+                        kwargs["explicit_ack_mode"] = "explicitOnly"
+                        kwargs["worker_allocation_mode"] = "static"
+                    server.api.api.endpoints.functions.create_model_monitoring_stream(
```
davesh0812:
```suggestion
                    server.api.api.endpoints.nuclio.create_model_monitoring_stream(
```
for now just fix this, in the future we will move this function to `MonitoringDeployment` class

davesh0812 (commented):
Just tiny comment from my side :)

server/api/api/endpoints/nuclio.py
```diff
@@ -146,3 +162,438 @@ async def delete_api_gateway(
 
     async with server.api.utils.clients.async_nuclio.Client(auth_info) as client:
         return await client.delete_api_gateway(project_name=project, name=gateway)
+
+
+@router.post("/projects/{project}/functions/{name}/deploy")
```
theSaarco: I think we might want to make it clear that this API is for nuclio only (to not confuse with building/deploying jobs). Why not have `nuclio` in the endpoint?
Maybe `/projects/{project}/nuclio/{name}/deploy`?

mlrun/projects/project.py
```diff
@@ -2180,16 +2180,13 @@ def _instantiate_function(
         if func is None and not _has_module(handler, kind):
             # if function path is not provided and it is not a module (no ".")
             # use the current notebook as default
-            if not is_ipython:
-                raise ValueError(
-                    "Function path or module must be specified (when not running inside a Notebook)"
-                )
-            from IPython import get_ipython
+            if is_ipython:
+                from IPython import get_ipython
```
theSaarco, line 2184: Before this change if this was not `is_ipython` this method would explode and now it won't. Is this intentional?

theSaarco (approved):
Looks very good. I think I'd name the new nuclio endpoints differently, but other than that it looks good.

mlrun/projects/project.py
alonmr: Yes because I wanted to allow creating nuclio functions without code (application runtime is created with an image instead). If no code is given and it's not a nuclio runtime, it will explode in `_init_function_from_dict`

PR #5344, opened 2024-03-28
[Model Monitoring] Deploy the default app on the go

jond01:
Resolves [ML-5865](https://iguazio.atlassian.net/browse/ML-5865).
Move the deployment to the backend in the background.
This enables a smoother and non-blocking experience with the `project.enable_model_monitoring()` call.

The relevant system tests passed.

[ML-5865]: https://iguazio.atlassian.net/browse/ML-5865?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/db/base.py
+++ b/mlrun/db/base.py
@@ -14,7 +14,7 @@
 
 import datetime
 from abc import ABC, abstractmethod
-from typing import Any, Optional, Union
+from typing import Optional, Union
 
 import mlrun.common.schemas
 import mlrun.model_monitoring
--- a/mlrun/db/base.py
+++ b/mlrun/db/base.py
@@ -14,7 +14,7 @@
 
 import datetime
 from abc import ABC, abstractmethod
-from typing import Optional, Union
+from typing import Any, Optional, Union
 
 import mlrun.common.schemas
 import mlrun.model_monitoring
@@ -722,9 +722,11 @@ class RunDBInterface(ABC):
         pass
 
     def enable_model_monitoring(
-        self,
-        project: str,
-        base_period: int = 10,
-        image: str = "mlrun/mlrun",
+        self, project: str, base_period: int = 10, image: str = "mlrun/mlrun"
     ):
         pass
+
+    def deploy_histogram_data_drift_app(
+        self, project: str, image: str = "mlrun/mlrun"
+    ) -> dict[str, Any]:
+        raise NotImplementedError
--- a/mlrun/db/base.py
+++ b/mlrun/db/base.py
@@ -723,10 +723,10 @@ class RunDBInterface(ABC):
 
     def enable_model_monitoring(
         self, project: str, base_period: int = 10, image: str = "mlrun/mlrun"
-    ) -> dict[str, Any]:
+    ) -> None:
         raise NotImplementedError
 
     def deploy_histogram_data_drift_app(
         self, project: str, image: str = "mlrun/mlrun"
-    ) -> dict[str, Any]:
+    ) -> None:
         raise NotImplementedError
--- a/mlrun/db/base.py
+++ b/mlrun/db/base.py
@@ -722,7 +722,11 @@ class RunDBInterface(ABC):
         pass
 
     def enable_model_monitoring(
-        self, project: str, base_period: int = 10, image: str = "mlrun/mlrun"
+        self,
+        project: str,
+        base_period: int = 10,
+        image: str = "mlrun/mlrun",
+        deploy_histogram_data_drift_app: bool = True,
     ) -> None:
         raise NotImplementedError
 
--- a/mlrun/db/base.py
+++ b/mlrun/db/base.py
@@ -723,8 +723,8 @@ class RunDBInterface(ABC):
 
     def enable_model_monitoring(
         self, project: str, base_period: int = 10, image: str = "mlrun/mlrun"
-    ):
-        pass
+    ) -> dict[str, Any]:
+        raise NotImplementedError
 
     def deploy_histogram_data_drift_app(
         self, project: str, image: str = "mlrun/mlrun"
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -1868,16 +1868,6 @@ class MlrunProject(ModelObj):
             requirements_file,
             **application_kwargs,
         )
-        models_names = "all"
-        function_object.set_label(
-            mm_constants.ModelMonitoringAppLabel.KEY,
-            mm_constants.ModelMonitoringAppLabel.VAL,
-        )
-        function_object.set_label("models", models_names)
-
-        if not mlrun.mlconf.is_ce_mode():
-            function_object.apply(mlrun.mount_v3io())
-
         # save to project spec
         self.spec.set_function(resolved_function_name, function_object, func)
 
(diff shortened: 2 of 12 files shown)


server/api/crud/model_monitoring/deployment.py
```diff
@@ -479,6 +487,56 @@ def _check_if_already_deployed(
...
+            logger.info("Ensured the histogram data drift function auth")
+
+        graph = func.set_topology(mlrun.serving.states.StepKinds.flow)
+        first_step = graph.to(class_name="HistogramDataDriftApplication")
+        first_step.to(
+            class_name=mlrun.model_monitoring.application.PushToMonitoringWriter(
+                project=self.project,
+                writer_application_name=mm_constants.MonitoringFunctionNames.WRITER,
+            )
+        ).respond()
```
davesh0812: Please take the same code we using in `project._instantiate_model_monitoring_function()` and copy the relevant part to one function that can be used from both server and mlrun - I think you can put it under `/mlrun/mlrun/model_monitoring/api.py`,  and if you want to take it to next step you can copy 'set_model_monitoring_function()' to this file and when we will be able in the future to add another api of set_mm_function.

mlrun/projects/project.py
```diff
@@ -2041,22 +2041,14 @@ def enable_model_monitoring(
...
-                    pathlib.Path(__file__).parent.parent
-                    / "model_monitoring/applications/histogram_data_drift.py"
-                ),
-                name=mm_constants.MLRUN_HISTOGRAM_DATA_DRIFT_APP_NAME,
-                application_class="HistogramDataDriftApplication",
-                image=image,
-            )
-            fn.deploy()
+            result |= db.deploy_histogram_data_drift_app(project=self.name, image=image)
+        return result
```
davesh0812: This dictionary is very confusing and I don't think we need to return it to the user.
You can maybe print it as a debug but no more

mlrun/projects/project.py
```diff
@@ -2041,22 +2041,14 @@ def enable_model_monitoring(
             )
             image = default_controller_image
         db = mlrun.db.get_run_db(secrets=self._secrets)
-        db.enable_model_monitoring(
+        result = db.enable_model_monitoring(
```
davesh0812: please add option to deploy the histogram_data_drift_app from this api as well, to avoid unnecessary api calls.

mlrun/projects/project.py
```diff
@@ -2041,22 +2041,14 @@ def enable_model_monitoring(
...
-                func=str(
-                    pathlib.Path(__file__).parent.parent
-                    / "model_monitoring/applications/histogram_data_drift.py"
-                ),
-                name=mm_constants.MLRUN_HISTOGRAM_DATA_DRIFT_APP_NAME,
-                application_class="HistogramDataDriftApplication",
-                image=image,
-            )
-            fn.deploy()
+            result |= db.deploy_histogram_data_drift_app(project=self.name, image=image)
```
davesh0812: You don't think we need also sdk for this?

mlrun/projects/project.py
```diff
@@ -2012,7 +2012,7 @@ def enable_model_monitoring(
         base_period: int = 10,
         image: str = "mlrun/mlrun",
         deploy_histogram_data_drift_app: bool = True,
```
davesh0812: I remember we agreed on adding an waiting flag with default set to false.
You can use `_wait_for_function_deployment()`.

mlrun/projects/project.py
```diff
@@ -2081,20 +2073,21 @@ def update_model_monitoring_controller(
...
 
-    def disable_model_monitoring(self):
+    def disable_model_monitoring(
+        self, *, delete_histogram_data_drift_app: bool = True
+    ) -> None:
+        """
+        Disable model monitoring by deleting the underlying functions infrastructure from MLRun database.
+
+        :param delete_histogram_data_drift_app: Whether to delete the histogram data drift app.
+        """
```
davesh0812: In the end of 1.7.0 this function will delete also the nuclio pods of those function, and without this ability this function is more harm than good, therefore I think for now lets mention it in the docs and maybe we can write also that this function is not valid for now.

davesh0812 (changes requested):
Nice!
I had few questions and requests :)

mlrun/projects/project.py
jond01: The API call costs nothing, and I prefer not to add the app to the infrastructure deployment.

For reference, the session is not created for every request:
https://github.com/mlrun/mlrun/blob/65a14413a56c3dbb4ab54df2417c1938142e0e46/mlrun/db/httpdb.py#L250

Edit: done. It's now a single API call.

mlrun/projects/project.py
jond01: It's out of scope for the current PR, but I agree it can be helpful. Let's address it in another PR.

mlrun/projects/project.py
```diff
@@ -1935,48 +1927,40 @@ def create_model_monitoring_function(
...
+        handler: typing.Optional[str] = None,
+        with_repo: typing.Optional[bool] = None,
+        tag: typing.Optional[str] = None,
+        requirements: typing.Union[str, list[str], None] = None,
         requirements_file: str = "",
         **application_kwargs,
     ) -> tuple[str, mlrun.runtimes.BaseRuntime, dict]:
+        import mlrun.model_monitoring.api
+
         function_object: RemoteRuntime = None
```
jond01, line 1945: This import is inside the method to avoid a circular import when it's done on the module level

mlrun/db/base.py
```diff
@@ -726,5 +726,11 @@ def enable_model_monitoring(
         project: str,
         base_period: int = 10,
         image: str = "mlrun/mlrun",
-    ):
-        pass
+        deploy_histogram_data_drift_app: bool = True,
+    ) -> None:
+        raise NotImplementedError
```
davesh0812, line 731: All the other unimplemented methods in this class are passed not raising anything, I think better to stick with the same behavior.

mlrun/db/base.py
```diff
@@ -726,5 +726,11 @@ def enable_model_monitoring(
...
+
+    def deploy_histogram_data_drift_app(
+        self, project: str, image: str = "mlrun/mlrun"
+    ) -> None:
+        raise NotImplementedError
```
davesh0812, line 736: same here

server/api/crud/model_monitoring/deployment.py
```diff
@@ -479,6 +488,37 @@ def _check_if_already_deployed(
                 pass
         logger.info(f"Deploying {function_name} function", project=self.project)
 
+    def deploy_histogram_data_drift_app(self, image: str) -> None:
```
davesh0812, line 491: Return should be the same as the other deploy functions

davesh0812 (approved):
LGTM, 
few comments, subject to @assaf758  decision

assaf758 (approved):
Looks good.
Merging this PR, but @jond01 please address the comments in a follow-up PR.

mlrun/db/base.py
assaf758: agree should be consistent, we need to decide if should raise an expception or to pass.

jond01: Addressed in #5418 as well as the other comments.

PR #5367, opened 2024-04-02
[API] Add alerts handling

yanburman:
Following enhancements are still not part of this PR:
* cool_down_period support
* message formatting based on user supplied pattern
* SDK enhancements and templates
* Model endpoint URI in notification
* Getter for events (stretch goal)
https://iguazio.atlassian.net/issues/ML-6027 as part of a bigger https://iguazio.atlassian.net/issues/ML-5287


server/api/crud/alerts.py
```diff
@@ -0,0 +1,279 @@
...
+            h = int(date_str.split("h")[0].strip())
+        elif date_str.endswith("m"):
+            m = int(date_str.split("m")[0].strip())
+        elif date_str.endswith("s"):
+            s = int(date_str.split("s")[0].strip())
+
+        if s == 0 and m == 0 and h == 0 and d == 0:
+            return None
+
+        return datetime.timedelta(days=d, hours=h, minutes=m, seconds=s)
```
quaark: Use `humanfriendly.parse_timespan` to convert the string to seconds, so something like:
```
datetime.timedelta(seconds=humanfriendly.parse_timespan(date_str))

File: server/api/db/sqldb/db.py [hunk 2]
```diff
@@ -4180,6 +4182,179 @@ def create_data_version(self, session, version):
...
+            if not isinstance(notif["when"], list):
+                notif["when"] = [notif["when"]]
+            return notif
+
+        alert.notifications = [
+            mlrun.common.schemas.notification.Notification(**_adjust_obj(x))
+            for x in self._get_db_notifications(
+                session, AlertConfig, parent_id=alert.id
+            )
+        ]
```
quaark: Please don't use such generic/shortened variable and method names and please add an explanation, why remove the secret params?
```suggestion
        def _enrich_notification(_notification):
            _notification = _notification.to_dict()
            del _notification["secret_params"]

            if not isinstance(_notification["when"], list):
                _notification["when"] = [_notification["when"]]
            return _notification

        alert.notifications = [
            mlrun.common.schemas.notification.Notification(**_enrich_notification(notification))
            for notification in self._get_db_notifications(
                session, AlertConfig, parent_id=alert.id
            )
        ]
```

server/api/db/sqldb/models/models_mysql.py
```diff
@@ -172,6 +172,41 @@ class Notification(Base, mlrun.utils.db.BaseModel):
     return Notification
 
 
+def make_alert_state(table):
+    class AlertState(Base, mlrun.utils.db.BaseModel):
+        __tablename__ = f"{table}_alert_states"
```
quaark: Why is this a generic table? It's creating only one table called `alert_configs_alert_states`. Are there going to be `alert_states` of an object that isn't an `alert_config`?
Feels a bit redundant to me

tests/integration/sdk_api/alerts/test_alerts.py
```diff
@@ -0,0 +1,563 @@
...
+            alert_entity_project,
+            alert_name,
+            alert_entity_kind,
+            alert_entity_project,
+            alert_summary,
+            event_name,
+        )
+        self._post_event(alert_entity_project, event_name, alert_entity_kind)
+
+    def _create_alerts_test(self, project_name, alert1, alert2):
```
quaark, line 179: I feel like a lot of the checks here can be done cleaner with a for loop / pytest.parametrize on `(alert_config, expectation)` (See other places in mlrun that use `expectation`)

quaark (changes requested):
In general this looks really good. There are a few minor things that bothered me and a couple of questions.
The code dup in NotificationPusher is ok for now IMO, We can think of a future refactor PR to try and minimize the dup but it isn't terrible right now as they are different enough in implementation.

mlrun/common/schemas/client_spec.py
```diff
@@ -66,3 +66,4 @@ class ClientSpec(pydantic.BaseModel):
     logs: typing.Optional[dict]
     packagers: typing.Optional[dict]
     external_platform_tracking: typing.Optional[dict]
+    alerts_mode: typing.Optional[str]
```
theSaarco, line 69: I think we should also report this mode in the `frontend_spec` as well, so that the UI can present or hide alert configuration once it's implemented.

mlrun/db/httpdb.py
```diff
@@ -3804,6 +3815,81 @@ def warn_on_s3_and_ecr_permissions_conflict(func):
...
+        Retrieve an alert.
+        :param alert_name: name of the alert to retrieve.
+        :param project: project that the alert belongs to.
+        :return: the alert object.
+        """
+        project = project or config.default_project
+        endpoint_path = f"projects/{project}/alerts/{alert_name}"
+        error_message = f"get alert {project}/alerts/{alert_name}"
+        response = self.api_call("GET", endpoint_path, error_message)
+        return response.json()
```
theSaarco: On a get call, it's standard to return the schema object created from the response JSON.

mlrun/db/httpdb.py
```diff
@@ -3804,6 +3815,81 @@ def warn_on_s3_and_ecr_permissions_conflict(func):
                 default_docker_registry=config.httpdb.builder.docker_registry,
             )
 
+    def generate_event(self, name, event_data, project=""):
```
theSaarco: Does this only support the object as dictionary in `event_data`? In most APIs we enable passing either a dict or the object itself.

mlrun/model_monitoring/writer.py
```diff
@@ -171,6 +172,27 @@ def _update_tsdb(self, event: _AppResultEvent) -> None:
...
+                "id": uid,
+            }
+            event_kind = (
+                alert_constants.EventKind.DRIFT_DETECTED
+                if drift_status == ResultStatusApp.detected
+                else alert_constants.EventKind.DRIFT_SUSPECTED
+            )
+            event_data = mlrun.common.schemas.Event(
+                kind=event_kind, entity=entity, value=drift_value
+            ).dict()
```
theSaarco: If the httpdb method would support the object itself, then there will be no need to call `.dict()` on the object.

mlrun/projects/project.py
```diff
@@ -3773,6 +3774,62 @@ def delete_api_gateway(
...
+        Create/modify an alert.
+        :param alert_data: the data of the alert.
+        :param alert_name: name of the alert.
+        """
+        db = mlrun.db.get_run_db(secrets=self._secrets)
+        if alert_name is None:
+            alert_name = alert_data.name
+        return db.store_alert_config(alert_name, alert_data.dict(), self.metadata.name)
+
+    def get_alert_config(self, alert_name):
```
```suggestion
    def get_alert_config(self, alert_name) -> AlertConfig:
```
This is missing a hint as well as a :return: docstring notation.
Since the project object is client-facing, it's usually "talking" objects, so should return the actual `AlertConfig` (see comment on httpdb for returning the actual object).

mlrun/projects/project.py
```diff
@@ -3773,6 +3774,62 @@ def delete_api_gateway(
...
+    def delete_alert_config(self, alert_data: AlertConfig = None, alert_name=None):
+        """
+        Delete an alert.
+        :param alert_data: the data of the alert.
+        :param alert_name: name of the alert to delete.
+        """
+        if alert_data is None and alert_name is None:
+            raise ValueError(
+                "At least one of alert_data or alert_name must be provided"
+            )
```
theSaarco, line 3841: Should also check what happens if both are set, at least verify they align (or fail, depending on your definition of the API).

mlrun/projects/project.py
```diff
@@ -3773,6 +3774,62 @@ def delete_api_gateway(
...
+    def reset_alert_config(self, alert_data=None, alert_name=None):
+        """
+        Reset an alert.
+        :param alert_data: the data of the alert.
+        :param alert_name: name of the alert to reset.
+        """
+        if alert_data is None and alert_name is None:
+            raise ValueError(
+                "At least one of alert_data or alert_name must be provided"
+            )
```
theSaarco, line 3860: Same - check if they are both set.

mlrun/utils/notifications/notification/slack.py
```diff
@@ -80,22 +84,46 @@ def _generate_slack_data(
...
 
         return data
 
+    def _get_alert_fields(
+        self,
+        alert: mlrun.common.schemas.AlertConfig,
+        event_data: mlrun.common.schemas.Event,
+    ) -> list:
+        line = [
+            self._get_slack_row(f":face_with_spiral_eyes: {alert.name} has occurred"),
```
theSaarco:
```suggestion
            self._get_slack_row(f":face_with_spiral_eyes: {alert.name} alert has occurred"),
```

Is 😵‍💫 the international icon of alerts? I would replace with something more conventional such as `:warning:` or something like this?

server/api/api/endpoints/alerts.py
```diff
@@ -0,0 +1,210 @@
...
+    )
+
+    resp = await run_in_threadpool(
+        server.api.crud.Alerts().get_enriched_alert, db_session, project, name
+    )
+
+    if resp is None:
+        raise mlrun.errors.MLRunNotFoundError(
+            f"Alert {name} for project {project} not found"
+        )
```
theSaarco: Most cases the CRUD object (or actually the DB code) would raise the exception if not found, removing the need to forward the `None` response through the layers all the way to the endpoint just to raise an exception there. 
Is there a specific reason you prefer returning `None`?

server/api/api/endpoints/alerts.py
```diff
@@ -0,0 +1,210 @@
...
+        chief_client = server.api.utils.clients.chief.Client()
+        return await chief_client.reset_alert(
+            project=project, name=name, request=request
+        )
+
+    logger.debug(f"Resetting alert {name}", project=project)
+
+    return await run_in_threadpool(
+        server.api.crud.Alerts().reset_alert, db_session, project, name
+    )
```
theSaarco, line 203: Here, for example, you raise the exception in the CRUD if the object doesn't exist. Should decide on an approach and be consistent (prefer to be consistent with the rest of the endpoints and raise internally).

server/api/crud/alerts.py
```diff
@@ -0,0 +1,269 @@
...
+        return server.api.utils.singletons.db.get_db().get_alert(session, project, name)
+
+    def get_alert_by_id(
+        self,
+        session: sqlalchemy.orm.Session,
+        alert_id: str,
+    ) -> mlrun.common.schemas.AlertConfig:
+        return server.api.utils.singletons.db.get_db().get_alert_by_id(
+            session, alert_id
+        )
```
theSaarco: Is this used anywhere? Looks like this is internal functionality that is needed internally by the DB, but not used externally. Why do you expose it in the CRUD?

server/api/crud/events.py
```diff
@@ -0,0 +1,72 @@
...
+        return bool(event_data.is_valid())
+
+    def add_event(self, project, name, alert_name):
+        self._cache.setdefault((project, name), []).append(alert_name)
+
+    def remove_event(self, project, name):
+        del self._cache[(project, name)]
+
+    def delete_alert_events(self, project):
+        to_delete = [name for proj, name in self._cache if proj == project]
```
theSaarco, line 45:
```suggestion
        to_delete = [name for proj, name in self._cache.items() if proj == project]
```

`in self._cache` would return only the keys and this line would raise an exception. Do you have a unit test covering this API?

server/api/crud/events.py
```diff
@@ -0,0 +1,72 @@
...
+
+        return bool(event_data.is_valid())
+
+    def add_event(self, project, name, alert_name):
+        self._cache.setdefault((project, name), []).append(alert_name)
+
+    def remove_event(self, project, name):
+        del self._cache[(project, name)]
+
+    def delete_alert_events(self, project):
```
theSaarco:
```suggestion
    def delete_project_alert_events(self, project):
```

server/api/crud/events.py
```diff
@@ -0,0 +1,72 @@
...
+        if event_data.entity.project != project:
+            return False
+
+        return bool(event_data.is_valid())
+
+    def add_event(self, project, name, alert_name):
+        self._cache.setdefault((project, name), []).append(alert_name)
+
+    def remove_event(self, project, name):
+        del self._cache[(project, name)]
```
theSaarco: The names of these functions are confusing - `add_event` might sound like the operation you do to raise an event. Maybe call them `add_event_configuration` and `remove_event_configuration` or something.

server/api/crud/events.py
```diff
@@ -0,0 +1,72 @@
...
+import mlrun.utils.singleton
+import server.api.api.utils
+import server.api.utils.singletons.db
+from mlrun.utils import logger
+
+
+class Events(
+    metaclass=mlrun.utils.singleton.Singleton,
+):
+    _cache = {}
```
theSaarco: Add a small comment or a type hint on the structure of this cache (it's a dict from string to list, right?)

server/api/db/sqldb/db.py
```diff
@@ -4180,6 +4183,182 @@ def create_data_version(self, session, version):
         self._upsert(session, [data_version_record])
 
     # ---- Background Tasks ----
```
theSaarco: Move this comment to after your new block (before the actual background tasks).

server/api/db/sqldb/db.py
```diff
@@ -4180,6 +4183,182 @@ def create_data_version(self, session, version):
...
+                session, AlertConfig, name=name, project=project
+            ).one_or_none()
+        finally:
+            pass
+
+    def _get_alert_by_id(self, session, alert_id: int) -> AlertConfig:
+        try:
+            return self._query(session, AlertConfig, id=alert_id).one_or_none()
+        finally:
+            pass
```
theSaarco: Looks like the difference between this and the `get_alert_by_id` is that this returns the DB object, while the other method returns the schema. Some questions:
1. Why not have the other method call this one?
2. Also - why do you need the `try` and `finally` here?
3. And finally, I suggest using a different name to indicate they are returning different objects.

server/api/crud/alerts.py
```diff
@@ -0,0 +1,269 @@
...
+            )
+            notification_object.validate_notification()
+
+        if alert.entity.project != project:
+            raise mlrun.errors.MLRunBadRequestError(
+                f"Invalid alert entity project ({alert.entity.project}) for alert {name} for project {project}"
+            )
+
+    @staticmethod
+    def _string2datetime(date_str):
```
quaark: More accurate name
```suggestion
    def _string_to_timedelta(date_str):
```
And maybe a good idea to move this to a `mlrun.common.utils.helpers`

server/api/db/sqldb/db.py
```diff
@@ -4180,6 +4183,182 @@ def create_data_version(self, session, version):
         self._upsert(session, [data_version_record])
 
     # ---- Background Tasks ----
+
+    def create_alert(
+        self, session, alert: mlrun.common.schemas.AlertConfig
+    ) -> mlrun.common.schemas.AlertConfig:
+        alert_record = self._transform_alert_config_schema_to_record(alert)
+        self._upsert(session, [alert_record])
```
theSaarco: When we do `create_alert` the expectation is that the method would explode if the object already exists (otherwise, use `store_alert`). You don't check it here.

quaark (changes requested):
Just another minor thing

server/api/db/sqldb/db.py
```diff
@@ -4180,6 +4183,182 @@ def create_data_version(self, session, version):
...
+
+        state = AlertState(count=0, parent_id=alert_record.id)
+        self._upsert(session, [state])
+        return self._transform_alert_config_record_to_schema(alert_record)
+
+    def store_alert(
+        self, session, alert: mlrun.common.schemas.AlertConfig
+    ) -> mlrun.common.schemas.AlertConfig:
+        alert_record = self._get_alert_by_id(session, alert.id)
+        alert_record.full_object = alert.dict()
```
theSaarco: `alert_record` may be `None` at this stage, causing this to explode.

server/api/db/sqldb/db.py
```diff
@@ -4180,6 +4183,182 @@ def create_data_version(self, session, version):
...
+
+        self._delete_alert_notifications(session, alert.name, alert, alert.project)
+        self._store_notifications(
+            session, AlertConfig, alert.notifications, alert_record.id, alert.project
+        )
+
+        self._upsert(session, [alert_record, alert_state])
+        return self._transform_alert_config_record_to_schema(
+            self._get_alert_by_id(session, alert_record.id)
+        )
```
```suggestion
        return self.get_alert_by_id(session, alert_record.id)
```

theSaarco (changes requested):
Partial review - submitting since I've accumulated already many comments. I did go over most of the code, except the DB schemas and the tests. Will review them later.

server/api/crud/events.py
yanburman: actually, there is a test for exactly this case. the test is called test_alert_after_project_deletion() in integration tests
the reason this works is that the key is actually (project, name)

mlrun/model_monitoring/writer.py
```diff
@@ -201,4 +223,12 @@ def do(self, event: _RawEvent) -> None:
         self._update_tsdb(event)
         self._update_kv_db(event)
         _Notifier(event=event, notification_pusher=self._custom_notifier).notify()
+
+        if mlrun.mlconf.alerts.mode == mlrun.common.schemas.alert.AlertsModes.enabled:
+            self._generate_event_on_drift(
+                event[WriterEvent.ENDPOINT_ID],
+                event[WriterEvent.RESULT_STATUS],
+                event[WriterEvent.RESULT_VALUE],
+                self.project,
```
davesh0812, line 234: Need to add here also the result name & app_name (WriterEvent.RESULT_NAME, WriterEvent.APP_NAME).
Also it can be very helpful if we can add free text for drift description (can be empty for now)

mlrun/model_monitoring/writer.py
```diff
@@ -201,4 +223,12 @@ def do(self, event: _RawEvent) -> None:
         self._update_tsdb(event)
         self._update_kv_db(event)
         _Notifier(event=event, notification_pusher=self._custom_notifier).notify()
```
davesh0812, line 227: If I understand correctly the `_Notofier` is deprecated and can be deleted no?

mlrun/common/schemas/alert.py
```diff
@@ -0,0 +1,122 @@
...
+
+class EventEntity(pydantic.BaseModel):
+    kind: EventEntityKind
+    project: str
+    id: str
+
+
+class EventKind(StrEnum):
+    DRIFT_DETECTED = "drift_detected"
+    DRIFT_SUSPECTED = "drift_suspected"
```
davesh0812, line 37: I think we need only one kind for mm alert or one extra generic kind for passing a threshold, because the result status is changing to be custom.
What is the differences between this 2 event kind?
Can the user add custom event kind?

mlrun/model_monitoring/writer.py
```diff
@@ -171,6 +172,27 @@ def _update_tsdb(self, event: _AppResultEvent) -> None:
                 event=event,
             )
 
+    @staticmethod
+    def _generate_event_on_drift(uid, drift_status, drift_value, project_name):
```
davesh0812: please add type hints :)

tests/system/alerts/test_alerts.py
```diff
@@ -0,0 +1,250 @@
...
+                    "webhook": "some-webhook",
+                },
+            },
+            {
+                "kind": "webhook",
+                "name": "drift2",
+                "message": "A drift was detected",
+                "severity": "warning",
+                "when": ["now"],
+                "condition": "failed",
```
davesh0812, line 202: I am not sure what does the condition meaning, but it not suppose to be related to drift detection?

tests/system/alerts/test_alerts.py
```diff
@@ -0,0 +1,250 @@
...
+        # in order to trigger the periodic monitor runs function, to detect the failed run and send an event on it
+        time.sleep(35)
+
+        # Validate that the notifications was sent on the failed job
+        expected_notifications = ["notification failure"]
+        self._validate_notifications_on_nuclio(
+            nuclio_function_url, expected_notifications
+        )
+
+    def test_drift_detection_alert(self):
```
davesh0812, line 78: nice :)

mlrun/db/httpdb.py
```diff
@@ -3812,6 +3823,81 @@ def warn_on_s3_and_ecr_permissions_conflict(func):
...
+        """
+        Generate an event.
+        :param name: name of the event.
```
jillnogold: :param name: The name of the event.
        :param event_data: The data of the event.
        :param project: The project that the event belongs to.

mlrun/db/httpdb.py
```diff
@@ -3812,6 +3823,81 @@ def warn_on_s3_and_ecr_permissions_conflict(func):
...
+        project = project or config.default_project
+        endpoint_path = f"projects/{project}/events/{name}"
+        error_message = f"post event {project}/events/{name}"
+        body = _as_json(event_data)
+        self.api_call("POST", endpoint_path, error_message, body=body)
+
+    def store_alert_config(self, alert_name, alert_data, project=""):
+        """
+        Create/modify an alert.
+        :param alert_name: name of the alert.
```
jillnogold: :param alert_name: The name of the alert.
        :param project: The project that the alert belongs to.
        return: The created/modified alert.

mlrun/db/httpdb.py
```diff
@@ -3812,6 +3823,81 @@ def warn_on_s3_and_ecr_permissions_conflict(func):
...
+        endpoint_path = f"projects/{project}/alerts/{alert_name}"
+        error_message = f"put alert {project}/alerts/{alert_name}"
+        body = _as_json(alert_data)
+        response = self.api_call("PUT", endpoint_path, error_message, body=body)
+        return response.json()
+
+    def get_alert_config(self, alert_name, project=""):
+        """
+        Retrieve an alert.
+        :param alert_name: name of the alert to retrieve.
```
jillnogold: :param alert_name: The name of the alert to retrieve.
        :param project: The project that the alert belongs to.
        :return: The alert object.

mlrun/db/httpdb.py
```diff
@@ -3812,6 +3823,81 @@ def warn_on_s3_and_ecr_permissions_conflict(func):
...
+        """
+        project = project or config.default_project
+        endpoint_path = f"projects/{project}/alerts/{alert_name}"
+        error_message = f"get alert {project}/alerts/{alert_name}"
+        response = self.api_call("GET", endpoint_path, error_message)
+        return response.json()
+
+    def list_alerts_configs(self, project=""):
+        """
+        Retrieve list of alerts of a project.
```
jillnogold, line 3944: Retrieve the list of alerts for a project.
        :param project: The project name.
        :return: All of the alert objects of the project.

mlrun/db/httpdb.py
```diff
@@ -3812,6 +3823,81 @@ def warn_on_s3_and_ecr_permissions_conflict(func):
...
+        """
+        project = project or config.default_project
+        endpoint_path = f"projects/{project}/alerts"
+        error_message = f"get alerts {project}/alerts"
+        response = self.api_call("GET", endpoint_path, error_message)
+        return response.json()
+
+    def delete_alert_config(self, alert_name, project=""):
+        """
+        Delete an alert.
```
jillnogold, line 3959: Delete an alert.
        :param alert_name: The name of the alert to delete.
        :param project: The project that the alert belongs to.

mlrun/db/httpdb.py
```diff
@@ -3812,6 +3823,81 @@ def warn_on_s3_and_ecr_permissions_conflict(func):
...
+        """
+        project = project or config.default_project
+        endpoint_path = f"projects/{project}/alerts/{alert_name}"
+        error_message = f"delete alert {project}/alerts/{alert_name}"
+        self.api_call("DELETE", endpoint_path, error_message)
+
+    def reset_alert_config(self, alert_name, project=""):
+        """
+        Reset an alert.
+        :param alert_name: name of the alert to reset.
```
jillnogold: :param alert_name: The name of the alert to reset.
        :param project: The project that the alert belongs to.

mlrun/projects/project.py
```diff
@@ -3797,6 +3798,62 @@ def delete_api_gateway(
 
         mlrun.db.get_run_db().delete_api_gateway(name=name, project=self.name)
 
+    def store_alert_config(self, alert_data: AlertConfig, alert_name=None):
+        """
+        Create/modify an alert.
+        :param alert_data: the data of the alert.
```
jillnogold: :param alert_data: The data of the alert.

mlrun/projects/project.py
```diff
@@ -3797,6 +3798,62 @@ def delete_api_gateway(
...
+        """
+        db = mlrun.db.get_run_db(secrets=self._secrets)
+        if alert_name is None:
+            alert_name = alert_data.name
+        return db.store_alert_config(alert_name, alert_data.dict(), self.metadata.name)
+
+    def get_alert_config(self, alert_name):
+        """
+        Retrieve an alert.
+        :param alert_name: name of the alert to retrieve.
```
jillnogold: :param alert_name: -- The name of the alert to retrieve.

mlrun/projects/project.py
```diff
@@ -3797,6 +3798,62 @@ def delete_api_gateway(
...
+        """
+        Retrieve list of alerts of a project.
+        """
+        db = mlrun.db.get_run_db(secrets=self._secrets)
+        return db.list_alerts_configs(self.metadata.name)
+
+    def delete_alert_config(self, alert_data: AlertConfig = None, alert_name=None):
+        """
+        Delete an alert.
+        :param alert_data: the data of the alert.
```
jillnogold: :param alert_data: The data of the alert.
        :param alert_name: The name of the alert to delete.

mlrun/projects/project.py
```diff
@@ -3797,6 +3798,62 @@ def delete_api_gateway(
...
+            )
+        db = mlrun.db.get_run_db(secrets=self._secrets)
+        if alert_data:
+            alert_name = alert_data.name
+        return db.delete_alert_config(alert_name, self.metadata.name)
+
+    def reset_alert_config(self, alert_data=None, alert_name=None):
+        """
+        Reset an alert.
+        :param alert_data: the data of the alert.
```
jillnogold: :param alert_data: The data of the alert.
        :param alert_name: The name of the alert to reset.

davesh0812 (commented):
Nice! few comments from model monitoring point of view

mlrun/model_monitoring/writer.py
Yacouby: yes, after a conversation with @jond01 , it was agreed that it will be deleted after this PR is merged

davesh0812: No problem, please erase this line from the writer, and about the `_Notifier` class @jond01 will erase it with all the relevant tests in separate pr.

mlrun/common/schemas/alert.py
Yacouby: for now the user can not add a custom event kind, since it is not in the scope of the current requirements..

tests/system/alerts/test_alerts.py
Yacouby: no, it is related to the notifications itself and not the drift detection

mlrun/model_monitoring/writer.py
jond01: > yes, after a conversation with @jond01 , it was agreed that it will be deleted after this PR is merged

@Yacouby - that is not exactly what we talked about IIRC.
Since the alerts will be disabled by default in the coming 1.7, we will remove the custom notifier in the writer only in 1.8.

Yacouby: it has not been decided yet that the alerts will be disabled by default in 1.7,
but either way we will remove the notifier accordingly when it will be enabled.

mlrun/utils/notifications/notification/console.py
```diff
@@ -36,6 +36,7 @@ def push(
         ] = mlrun.common.schemas.NotificationSeverity.INFO,
         runs: typing.Union[mlrun.lists.RunList, list] = None,
         custom_html: str = None,
+        alert: mlrun.common.schemas.AlertConfig = None,
```
liranbg, line 39: I think you're missing here the event data (note that the push function is common signature between couple of implementation

server/api/api/endpoints/alerts.py
```diff
@@ -0,0 +1,203 @@
...
+    if (
+        mlrun.mlconf.httpdb.clusterization.role
+        != mlrun.common.schemas.ClusterizationRole.chief
+    ):
+        chief_client = server.api.utils.clients.chief.Client()
+        return await chief_client.reset_alert(
+            project=project, name=name, request=request
+        )
+
+    logger.debug(f"Resetting alert {name}", project=project)
```
liranbg:
```suggestion
    logger.debug("Resetting alert", project=project, name=name)
```

true for all occurrences

quaark (approved):
Great stuff!
Looks good to me

liranbg (commented):
did not cover everything but raising 2 comments for now

server/api/api/endpoints/alerts.py
```diff
@@ -0,0 +1,203 @@
...
+    )
+    await server.api.utils.auth.verifier.AuthVerifier().query_project_permissions(
+        project,
+        mlrun.common.schemas.AuthorizationAction.read,
+        auth_info,
+    )
+
+    return await run_in_threadpool(
+        server.api.crud.Alerts().list_alerts, db_session, project
+    )
```
theSaarco, line 125: Note that since we (at least on paper) support fine-grained permissions, this list of alerts should be further verified for permissions before returning to user. See examples in other list APIs.
Given that we currently not really support this, it can be done later, in a subsequent PR, but please keep that in mind as future work. We still need to support this.

server/api/crud/alerts.py
```diff
@@ -0,0 +1,260 @@
...
+    def _normalize_events(obj, period):
+        now = datetime.datetime.now(tz=datetime.timezone.utc)
+        events = obj["events"]
+        for event in events:
+            if isinstance(event, str):
+                event_time = datetime.datetime.fromisoformat(event)
+            else:
+                event_time = event
+            if event_time > now + period:
+                events.remove(event)
```
theSaarco: This seems strange to me - aren't the events here the set of events kept in the alert state? How can these events be newer than `now + period`? Or are we expecting `period` to be negative?
Just looking at the code I would expect this to be `now > event_time + period` or something similar.
Please verify this, and if it isn't any of the reasons I mentioned, then please add a comment to explain the logic here since it is confusing.

tests/integration/sdk_api/alerts/test_alerts.py
```diff
@@ -0,0 +1,538 @@
...
+import mlrun
+import mlrun.common.schemas
+import mlrun.common.schemas.alert as alert_constants
+import mlrun.utils
+import tests.integration.sdk_api.base
+from mlrun.utils import logger
+
+
+class TestAlerts(tests.integration.sdk_api.base.TestMLRunIntegration):
+    def test_alert_operations(self):
```
theSaarco, line 28: Why not split this to smaller tests?

theSaarco (approved):
LGTM. 
Have some comments, including one the looks like a code mistake, but that is probably due to me not understanding the code properly.
Note also there is at least one thing that is missing and will need to go into next PR.

mlrun/model_monitoring/writer.py
davesh0812: Ok so if you enabling the alerts in 1.7.0 don't forget to delete it or just send me alert and we will do it :)

mlrun/common/schemas/alert.py
davesh0812: Ok but can you add another kind like I mention?
@theSaarco what do you think?

theSaarco: We have discussed adding something generic like `METRIC_THRESHOLD_PASSED` that can be used for custom metrics to alert on passing a threshold. It can be added here or in subsequent PR, depending on when you need this type of event in the code.

tests/integration/sdk_api/alerts/test_alerts.py
yanburman: integ tests are long to setup/teardown. the reasoning was less runtime

server/api/crud/alerts.py
yanburman: you are right. good catch

mlrun/common/schemas/alert.py
davesh0812: Sounds good, lets discuss it before and agree on the name and the relevant values :)  
But we will need it to 1.7.0

liranbg (approved):

PR #5371, opened 2024-04-03
[Tests] Add dask coverage in datastore tests

tomerm-iguazio:
1) Support dask data frames in all datastores' tests.
2) Use the same files for each integration test class.
3) Improved the uniformity of the tests.

[ML-5905](https://iguazio.atlassian.net/browse/ML-5905)

[ML-5905]: https://iguazio.atlassian.net/browse/ML-5905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/integration/aws_s3/test_aws_s3.py
```diff
@@ -74,59 +70,111 @@ def _make_target_names(
...
             secret_key=self._secret_access_key,
         )
-        register_temporary_client_datastore_profile(profile)
+        register_temporary_client_datastore_profile(self.profile)
+
+    @pytest.fixture(autouse=True)
+    def setup_before_each_test(self, use_datastore_profile):
+        if use_datastore_profile:
+            os.environ["AWS_ACCESS_KEY_ID"] = "wrong_access_key"
+            os.environ["AWS_SECRET_ACCESS_KEY"] = "wrong_token"
```
gtopper, line 107: This deserves a comment.

tests/integration/aws_s3/test_aws_s3.py
```diff
@@ -74,59 +70,111 @@ def _make_target_names(
...
+            os.environ["AWS_SECRET_ACCESS_KEY"] = self._secret_access_key
+
+    def _setup_df_dir(self, use_datastore_profile, file_format, reader):
+        param = self.s3["ds"] if use_datastore_profile else self.s3["s3"]
+        directory = f"/{file_format}s_{uuid.uuid4()}"
+        s3_directory_url = param["bucket_path"] + self.run_dir + directory
+        df1_path = join(self.assets_path, f"test_data.{file_format}")
+        df2_path = join(self.assets_path, f"additional_data.{file_format}")
+
+        #  upload
```
gtopper:
```suggestion
        # upload
```

tests/integration/aws_s3/test_aws_s3.py
```diff
@@ -203,78 +251,107 @@ def test_using_profile(self, use_datastore_profile):
...
+
+    @pytest.mark.parametrize(
+        "file_format, reader, reader_args",
+        [
+            ("parquet", dd.read_parquet, {}),
+            ("csv", dd.read_csv, {}),
+            ("json", dd.read_json, {"orient": "values"}),
+        ],
+    )
+    def test_as_df_dd(
```
gtopper: This test is a duplicate of `test_as_df`, expect for the `as_df` at the end. Why not just keep the original test and add the dask check at the end? It saves us from
* Duplicating a bunch of tests
* Effectively doubling the size of the parameter matrix

tests/integration/azure_blob/test_azure_blob.py
```diff
@@ -76,32 +77,74 @@
...
+    @classmethod
+    def create_fs(cls, storage_options):
+        #  Create filesystem object only once - for teardown_class method.
+        if not cls._azure_fs:
+            azure_fs = AzureBlobFileSystem(storage_options)
+            try:
+                azure_fs.info(cls._bucket_name)  # in order to check connection ...
+                cls._azure_fs = azure_fs
+            except Exception:
+                pass
```
gtopper: How come we just ignore any exceptions?

tests/integration/aws_s3/test_aws_s3.py
```diff
@@ -13,18 +13,20 @@
 # limitations under the License.
 #
 import os
-import random
-import tempfile
 import uuid
+from os.path import abspath, dirname, join
```
gtopper: These tests already use `pathlib`. Is there any justification to also using `os.path`? Also, in such a case it's better to `import os.path` and use the fully qualified name so that it's clear what these functions are.

tests/integration/aws_s3/test_aws_s3.py
```diff
@@ -74,59 +70,111 @@ def _make_target_names(
...
+        )
+
+    @classmethod
+    def teardown_class(cls):
+        test_dir = f"{cls._bucket_name}{cls.test_dir}"
+        if not cls._fs:
+            return
+        if cls._fs.exists(test_dir):
+            cls._fs.delete(test_dir, recursive=True)
+            logger.debug("test directory has been cleaned.")
```
gtopper:
```suggestion
            logger.debug("test directory has been deleted")
```

tests/system/datastore/test_v3io.py
```diff
@@ -185,8 +216,21 @@ def test_v3io_large_object_put(self):
...
-    def test_put_get_and_download(self):
-        data_item, _ = self._get_data_item()
+    @pytest.mark.parametrize("use_secrets_as_parameters", [True, False])
+    def test_put_get_and_download(
+        self, use_datastore_profile, use_secrets_as_parameters
+    ):
+        secrets = {}
+        if use_secrets_as_parameters:
+            os.environ["V3IO_ACCESS_KEY"] = "wrong_token"
+            #  Verify that we are using the profile secret:
```
gtopper:
```suggestion
            # Verify that we are using the profile secret:
```

tests/system/datastore/test_v3io.py
```diff
@@ -51,51 +59,74 @@ def setup_class(cls):
...
+        dataframes_dir = f"/{file_format}_{uuid.uuid4()}"
+        dataframes_url = f"{self.run_dir_url}{dataframes_dir}"
+        df1_path = join(self.assets_path, f"test_data.{file_format}")
+        df2_path = join(self.assets_path, f"additional_data.{file_format}")
 
-        dataitem_url = f"{self.object_dir_url}/df_{uuid.uuid4()}.{file_extension}"
-
-        uploaded_data_item = mlrun.run.get_dataitem(dataitem_url)
-        uploaded_data_item.upload(second_file_path)
+        #  upload
```
gtopper: Really not sure what this double spacing is about... Is this the work of some IDE?
```suggestion
        # upload
```

gtopper (commented):
Please note that some comments may refer to a specific line but are intended for all occurrences of the raised issue in the code.

tests/integration/aws_s3/test_aws_s3.py
tomerm-iguazio: tnx for the comment! done.

tests/integration/azure_blob/test_azure_blob.py
tomerm-iguazio: added warning log, also in test_v3io.py

tests/system/datastore/test_v3io.py
```diff
@@ -248,67 +295,72 @@ def test_rm(self):
...
+            ("parquet", pd.read_parquet, dd.read_parquet, False),
+            ("csv", pd.read_csv, dd.read_csv, True),
         ],
     )
-    def test_check_read_df_dir(
-        self,
-        file_extension: str,
-        reader: callable,
+    def test_as_df_directory(
+        self, use_datastore_profile, file_format, pd_reader, dd_reader, reset_index
```
gtopper: `use_datastore_profile` is not used.

tests/integration/aws_s3/test_aws_s3.py
```diff
@@ -203,78 +228,73 @@ def test_using_profile(self, use_datastore_profile):
...
+
+    @pytest.mark.parametrize(
+        "file_format, pd_reader, dd_reader, reset_index",
+        [
+            ("parquet", pd.read_parquet, dd.read_parquet, False),
+            ("csv", pd.read_csv, dd.read_csv, True),
+        ],
+    )
+    def test_as_df_directory(
+        self, use_datastore_profile, file_format, pd_reader, dd_reader, reset_index
```
gtopper: `use_datastore_profile` is not used.

tests/integration/aws_s3/test_aws_s3.py
```diff
@@ -203,78 +228,73 @@ def test_using_profile(self, use_datastore_profile):
...
+        dataframe_url = f"{self.run_dir_url}/{filename}"
+        local_file_path = os.path.join(self.assets_path, f"test_data.{file_format}")
+
+        source = pd_reader(local_file_path, **reader_args)
+        upload_data_item = mlrun.run.get_dataitem(dataframe_url)
+        upload_data_item.upload(local_file_path)
+        response = upload_data_item.as_df(**reader_args)
+        pd.testing.assert_frame_equal(source, response)
+
+        # dd
```
gtopper:
```suggestion
        # dask
```
(everywhere)

tests/integration/azure_blob/test_azure_blob.py
```diff
@@ -76,32 +78,57 @@
...
             for env_var in env_vars:
                 os.environ.pop(env_var, None)
 
         test_params = AUTH_METHODS_AND_REQUIRED_PARAMS.get(auth_method)
+        if use_datastore_profile:
+            self._bucket_url = f"ds://{self.profile_name}/{self._bucket_name}"
+        else:
+            self._bucket_url = "az://" + self._bucket_name
+        self.run_dir_url = f"{self._bucket_url}/{self.run_dir}"
+        self.object_url = self.run_dir_url + "/" + self.object_file
```
gtopper: It's a bit inconsistent to use f-string on one line and then not use it in exactly the same situation on the next line.

tests/integration/google_cloud_storage/test_google_cloud_storage.py
```diff
@@ -102,46 +106,99 @@ def _setup_profile(self, profile_auth_by):
...
+            ("credentials_file", True),
+            ("serialized_json", False),
+            ("serialized_json", True),
+        ],
+    )
+    def test_perform_google_cloud_storage_tests(
+        self, use_datastore_profile, setup_by, use_secrets
+    ):
+        #  TODO split to smaller tests by datastore conventions.
+        self.setup_mapping[setup_by](self, use_datastore_profile, use_secrets)
```
gtopper: This mechanism doesn't really make sense to me... It's a global setup, that registers a datastore profile and sets environment variables, but these are never deregistered or unset. So the leftovers from a run with one parameter set can affect subsequent runs.

tests/system/datastore/test_v3io.py
```diff
@@ -51,42 +59,70 @@ def setup_class(cls):
         }
         with open(cls.test_file_path) as f:
             cls.test_string = f.read()
-        cls.test_dir_path = "/bigdata/v3io_tests"
-        cls.v3io_test_dir_url = "v3io://" + cls.test_dir_path
+        cls.test_dir = "/bigdata/v3io_tests"
+        cls.test_dir_url = "v3io://" + cls.test_dir
+        cls.run_dir = f"{cls.test_dir}/run_{uuid.uuid4()}"
```
gtopper: Here too, not very consistent with the f-strings. I see that this applies in several places so I won't comment on each and every one to avoid spamming this PR.

tests/integration/aws_s3/test_aws_s3.py
```diff
@@ -60,73 +57,101 @@ def aws_s3_configured(extra_params=None):
...
         )
-        register_temporary_client_datastore_profile(profile)
+        register_temporary_client_datastore_profile(self.profile)
+
+    @pytest.fixture(autouse=True)
+    def setup_before_each_test(self, use_datastore_profile):
+        # We give priority to profiles, then to secrets, and finally to environment variables.
+        # We want to ensure that we test these priorities in the correct order.
+        mlrun.datastore.store_manager.reset_secrets()
+
```
gtopper:
```suggestion
        mlrun.datastore.store_manager.reset_secrets()

        # We give priority to profiles, then to secrets, and finally to environment variables.
        # We want to ensure that we test these priorities in the correct order.
```
or alternatively remove the blank line. Otherwise, it's not obvious what code this comment is referring to.

tests/integration/aws_s3/test_aws_s3.py
```diff
@@ -60,73 +57,101 @@ def aws_s3_configured(extra_params=None):
...
+        data_item.put(self.test_string)
+        df_url = f"{self.run_dir_url}/df_{uuid.uuid4()}.csv"
+        df_data_item = mlrun.run.get_dataitem(df_url, secrets=secrets)
         df_data_item.put(test_df_string)
 
         response = data_item.get()
-        assert response.decode() == test_string, "Result differs from original test"
+        assert (
+            response.decode() == self.test_string
+        ), "Result differs from original test"
```
gtopper:
```suggestion
        )
```
I know it was there before, but this assertion message is more confusing than helpful.

Naturally this refers to all occurrences, not just this line.

tests/integration/azure_blob/test_azure_blob.py
gtopper: No one will read this. Why not fail the test?

tests/integration/aws_s3/test_aws_s3.py
gtopper: Naturally, I was referring to all occurrences of this, not just in this one test file.

tests/integration/test_dbfs_store/test_dbfs_store.py
```diff
@@ -50,65 +52,59 @@ class TestDBFSStore:
...
     with config_file_path.open() as fp:
         config = yaml.safe_load(fp)
-    parquets_dir = "/parquets"
-    csv_dir = "/csv"
-    test_file_path = str(here / "test.txt")
-    json_path = str(here / "test_data.json")
-    parquet_path = str(here / "test_data.parquet")
-    additional_parquet_path = str(here / "additional_data.parquet")
-    csv_path = str(here / "test_data.csv")
-    additional_csv_path = str(here / "additional_data.csv")
```
gtopper, line 60: Moving this stuff to `setup_class` doesn't sit well with PyCharm. Why not leave it where it is?

And of course, I'm referring to all occurrences in this PR.

tests/system/datastore/test_v3io.py
```diff
@@ -248,67 +295,72 @@ def test_rm(self):
...
+        dt1.upload(src_path=df1_path)
+        dt2.upload(src_path=df2_path)
+        dt_dir = mlrun.run.get_dataitem(dataframes_url)
+        df1 = pd_reader(df1_path)
+        df2 = pd_reader(df2_path)
+        expected_df = pd.concat([df1, df2], ignore_index=True)
+        tested_df = dt_dir.as_df(format=file_format)
+        if reset_index:
+            tested_df = tested_df.sort_values("id").reset_index(drop=True)
+        assert_frame_equal(tested_df, expected_df)
```
gtopper:
```suggestion
        pd.testing.assert_frame_equal(tested_df, expected_df)
```
inconsistent usage.

gtopper (changes requested):
Nice. It's a big PR... Please assume all comments refer to all occurrences of the same pattern.

tests/integration/aws_s3/test_aws_s3.py
tomerm-iguazio: I can not find another pathlib import

tests/integration/google_cloud_storage/test_google_cloud_storage.py
tomerm-iguazio: os.environ reset every run in `setup_before_each_test`.
Regarding profiles - because they have the same profile name, it gets overwritten every time it runs (every time we create a new profile), but I take your comment and I added a remove to the profiles by profile_name.

tests/integration/test_dbfs_store/test_dbfs_store.py
tomerm-iguazio: I removed unnecessary parts of the code, leaving only what's essential. I moved the essential part to setup_class in order to serve multiple purposes:

1) It runs once and can be utilized across the entire class.
2) I'm creating a base class that every test class for each datastore will utilize.
Establishing standards for each subclass requires setting up these class attributes.

This is the first step in order to do so.

tests/integration/google_cloud_storage/test_google_cloud_storage.py
```diff
@@ -60,34 +58,39 @@ def google_cloud_storage_configured():
...
 @pytest.mark.parametrize("use_datastore_profile", [False, True])
 class TestGoogleCloudStorage:
-    #  must be static, in order to get access from setup_class or teardown_class.
-    @staticmethod
-    def clean_test_directory(bucket_name, object_dir, gcs_fs):
-        test_dir = f"{bucket_name}/{object_dir}/"
-        if gcs_fs.exists(test_dir):
-            gcs_fs.delete(test_dir, recursive=True)
+    assets_path = os.path.join(os.path.dirname(os.path.dirname(__file__)), "assets")
+    _bucket_name = config["env"].get("bucket_name")
```
gtopper:
```suggestion
    bucket_name = config["env"].get("bucket_name")
```
Unless I'm missing something, there's no reason for this attribute in particular to be private.

tests/integration/google_cloud_storage/test_google_cloud_storage.py
```diff
@@ -60,34 +58,39 @@ def google_cloud_storage_configured():
...
+    test_dir = "test_mlrun_gcs_objects"
+    run_dir = f"{test_dir}/run_{uuid.uuid4()}"
+    profile_name = "gcs_profile"
+    credentials_path = config["env"].get("credentials_json_file")
+    test_file = os.path.join(assets_path, "test.txt")
```
gtopper, line 67: I'm not really sure I understand why in some places it's `os.path.join`, and some places it's just `f"{...}/{...}"`.

tests/integration/google_cloud_storage/test_google_cloud_storage.py
```diff
@@ -101,169 +104,201 @@ def _setup_profile(self, profile_auth_by):
...
+            ("credentials_file", False),
+            ("credentials_file", True),
+            ("serialized_json", False),
+            ("serialized_json", True),
+        ],
+    )
+    def test_perform_google_cloud_storage_tests(
+        self, use_datastore_profile, setup_by, use_secrets
+    ):
+        #  TODO split to smaller tests by datastore conventions.
```
gtopper:
```suggestion
        # TODO: split to smaller tests by datastore conventions
```

tests/system/datastore/test_v3io.py
```diff
@@ -13,79 +13,95 @@
...
     @classmethod
     def setup_class(cls):
         super().setup_class()
-        cls.test_file_path = str(cls.get_assets_path() / "test.txt")
-        test_parquet_path = str(cls.get_assets_path() / "testdata_short.parquet")
-        test_csv_path = str(cls.get_assets_path() / "testdata_short.csv")
-        test_json_path = str(cls.get_assets_path() / "testdata_short.json")
+        cls.token = os.environ.get("V3IO_ACCESS_KEY")
+        assets_path = str(cls.get_assets_path())
+        cls.test_file_path = os.path.join(assets_path, "test.txt")
```
gtopper, line 49: The `cls.*` stuff can move to the class level as well.

tests/system/datastore/test_v3io.py
```diff
@@ -185,8 +201,21 @@ def test_v3io_large_object_put(self):
...
-    def test_put_get_and_download(self):
-        data_item, _ = self._get_data_item()
+    @pytest.mark.parametrize("use_secrets_as_parameters", [True, False])
+    def test_put_get_and_download(
+        self, use_datastore_profile, use_secrets_as_parameters
+    ):
+        secrets = {}
+        if use_secrets_as_parameters:
+            os.environ["V3IO_ACCESS_KEY"] = "wrong_token"
+            # Verify that we are using the profile secret:
```
gtopper: Maybe clarify that we set a wrong token to make sure that the right token is picked from the datastore profile.

tests/system/datastore/test_v3io.py
```diff
@@ -185,8 +201,21 @@ def test_v3io_large_object_put(self):
...
 
-    def test_put_get_and_download(self):
-        data_item, _ = self._get_data_item()
+    @pytest.mark.parametrize("use_secrets_as_parameters", [True, False])
+    def test_put_get_and_download(
+        self, use_datastore_profile, use_secrets_as_parameters
+    ):
+        secrets = {}
+        if use_secrets_as_parameters:
+            os.environ["V3IO_ACCESS_KEY"] = "wrong_token"
```
gtopper, line 214: The thing with setting the environment variable like this is that it's sticky across tests, and can cause situations where which tests are run and in what order can affect whether a test will pass or fail. It's better to save the original and reset the variable in a finally clause.

gtopper (commented):
Looks good to me with some minor comments.

tests/integration/google_cloud_storage/test_google_cloud_storage.py
tomerm-iguazio: os.path.join for local file system join (should work on windows too).

{...}/{...} for remote path join.

tests/system/datastore/test_v3io.py
tomerm-iguazio: I can not, get_assets_path() is not configure in the class level yet here...

tests/system/datastore/test_v3io.py
tomerm-iguazio: I have a fixture that accomplishes this and runs before each test.
It is a better solution than adding a 'finally' statement at the end of each test.

tests/integration/google_cloud_storage/test_google_cloud_storage.py
gtopper: I meant everywhere though, including `TestAlibabaOssDataStore`.

tests/system/datastore/test_v3io.py
gtopper: You can move `cls.token = os.environ.get("V3IO_ACCESS_KEY")`. You're right about the rest of them.

tests/system/datastore/test_v3io.py
gtopper: The test is setting `os.environ["V3IO_ACCESS_KEY"] = "wrong_token"` and never resets it back. If you try to run multiple suites, anything that runs after your suite will have their `V3IO_ACCESS_KEY` corrupted. This can fail a test like `test_masked_access_key`, depending on the order in which the suites happen to run.

tomerm-iguazio: please look at :
```

@pytest.fixture(autouse=True)
    def setup_before_each_test(self, use_datastore_profile):
```

tomerm-iguazio: and anyway, in v3io we inherit from TestMLRunSystem class.
In `setup_method` you overwrite this env variable in `self._setup_env(self._get_env_from_file())`

tests/system/datastore/test_v3io.py
tomerm-iguazio: resolved, tnx!

tomerm-iguazio: reverted, V3IO_ACCESS_KEY is not loaded in the class level

tests/integration/azure_blob/test_azure_blob.py
```diff
@@ -74,34 +75,60 @@
...
+    def teardown_method(self, method):
+        for auth, auth_list in AUTH_METHODS_AND_REQUIRED_PARAMS.items():
+            if auth.startswith("env"):
+                for env_parameter in auth_list:
+                    if config["env"].get(env_parameter, None):
+                        os.environ[env_parameter] = config["env"].get(env_parameter)
+
+    @classmethod
+    def create_fs(cls, storage_options):
+        #  Create filesystem object only once - for teardown_class method.
```
gtopper:
```suggestion
        # Create filesystem object only once
```
Pretty sure we create it for the tests, not the teardown 😛

gtopper (approved):

PR #5374, opened 2024-04-03
[Nuclio] API Gateway system test

rokatyy:
System test which covers:
* simple gateway creation
* gateway with basic auth creation
* canary function

[Passed locally]
https://iguazio.atlassian.net/browse/ML-6023


mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -300,6 +300,8 @@ def invoke_url(
 
         :return: (str) The invoke URL.
         """
+        if not self.host.startswith("http"):
+            self.host = f"http://{self.host}"
```
liranbg: Worth using https unless user provide explicit with http

tests/system/runtimes/test_nuclio.py
```diff
@@ -582,3 +582,76 @@ def test_hyper_run(self):
...
+
+    def _get_basic_gateway(self):
+        return mlrun.runtimes.nuclio.api_gateway.APIGateway(
+            project=self.project_name, functions=self.f1, name=self.gw_name
+        )
+
+    def _cleanup_gateway(self):
+        self.project.delete_api_gateway(self.gw_name)
+
+    def _deploy_function(self, replicas=1, suffix=""):
```
liranbg, line 630: move under public functions

tests/system/runtimes/test_nuclio.py
```diff
@@ -582,3 +582,76 @@ def test_hyper_run(self):
...
+        api_gateway = self.project.store_api_gateway(api_gateway)
+        retry = 0
+        while retry < max_retries:
+            retry += 1
+            result = api_gateway.invoke(auth=auth)
+            if result.status_code == 404:
+                time.sleep(timeout)
+                continue
+            else:
+                return api_gateway
```
liranbg: why wouldnt it be ready once `store` is performed?

liranbg (commented):
Minor suggestions and comments

tests/system/runtimes/test_nuclio.py
rokatyy: @liranbg because from nuclio side dashboard returns 201 after CRD creation, but controller should catch this change and create ingress (usually it takes ~1s)

mlrun/runtimes/nuclio/api_gateway.py
rokatyy: @liranbg there will be no certificate error then, do we want to skip certification check in that case?

liranbg: if user wants to avoid certification issue, he would pass "verify=False"

rokatyy: @liranbg fixed

tests/system/runtimes/test_nuclio.py
liranbg: I think we should make MLRun API aware of it and ensure it returns the created/updated api gateway only once it is usable (without probing it) so I think we can
1. mlrun api creates/updates api gw on nuclio
2. mlrun api waits until ingress record status block is populated (for nuclio < 1.13)
3. mlrun api returns the api gw



2/3 can be done using by either
1. background tasks
2. return the created entity and expose functionality to "ensure_ingress_readiness()" which will use mlrun api to ensure in the background (short polling)

rokatyy: @liranbg reworked

liranbg (approved):
Looking good. comment for followup PR - make sure to encapsulate the "wait_for_ready" to be inside store or update ingress. if it takes more than 30 seconds, add a debug log for the client knowing you are waiting for ingress status to be populated. if it exceeds 90seconds timeout - raise an exception telling the user that ingress status is not populated and ingress may not be usable for invocation.

PR #5375, opened 2024-04-03
[Tensorflow] Fix import

gtopper:
[ML-5916](https://iguazio.atlassian.net/browse/ML-5916)

tensorflow 2.16.1 breaks `tensorflow.keras.callbacks`.

`tensorflow.python.keras.callbacks` works with tensorflow 2.16.1 as well as with previous versions.

[ML-5916]: https://iguazio.atlassian.net/browse/ML-5916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/frameworks/tf_keras/callbacks/logging_callback.py
+++ b/mlrun/frameworks/tf_keras/callbacks/logging_callback.py
@@ -17,7 +17,7 @@ from typing import Callable, Union
 import numpy as np
 import tensorflow as tf
 from tensorflow import Tensor, Variable
-from tensorflow.keras.callbacks import Callback
+from tensorflow.python.keras.callbacks import Callback
 
 import mlrun
 
--- a/mlrun/frameworks/tf_keras/callbacks/logging_callback.py
+++ b/mlrun/frameworks/tf_keras/callbacks/logging_callback.py
@@ -15,9 +15,14 @@
 from typing import Callable, Union
 
 import numpy as np
+import semver
 import tensorflow as tf
 from tensorflow import Tensor, Variable
-from tensorflow.python.keras.callbacks import Callback
+
+if semver.VersionInfo.parse(tf.__version__) < semver.VersionInfo(2, 6):
+    from tensorflow.keras.callbacks import Callback
+else:
+    from keras.callbacks import Callback
 
 import mlrun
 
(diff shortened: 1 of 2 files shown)


guy1992l (changes requested):
Good catch Gal. For backward compatibility please check tf.__version__ and import accordingly

gtopper:
@guy1992l, from my testing, `import tensorflow.python.keras.callbacks` works with older versions as well.

mlrun/frameworks/tf_keras/callbacks/logging_callback.py
```diff
@@ -17,7 +17,7 @@
 import numpy as np
 import tensorflow as tf
 from tensorflow import Tensor, Variable
-from tensorflow.keras.callbacks import Callback
+from tensorflow.python.keras.callbacks import Callback
```
guy1992l:
```suggestion
if tf.__version__.split(".")[1] < 6:
    tensorflow.keras.callbacks import Callback
else:
    from keras.callbacks import Callback
```

mlrun/frameworks/tf_keras/mlrun_interface.py
```diff
@@ -19,15 +19,15 @@
...
-from tensorflow.keras.callbacks import (
+from tensorflow.keras.optimizers import Optimizer
+from tensorflow.python.keras.callbacks import (
     BaseLogger,
     Callback,
     CSVLogger,
     ModelCheckpoint,
     ProgbarLogger,
     TensorBoard,
 )
```
guy1992l:
```suggestion
if tf.__version__.split(".")[1] < 6:
    from tensorflow.keras.callbacks import (
        BaseLogger,
        Callback,
        CSVLogger,
        ModelCheckpoint,
        ProgbarLogger,
        TensorBoard,
    )
else:
    from keras.callbacks import (
        BaseLogger,
        Callback,
        CSVLogger,
        ModelCheckpoint,
        ProgbarLogger,
        TensorBoard,
    )
```

guy1992l (changes requested):
I've read more and since TensorFlow 2.6, Keras got separated and being released as a pip package (like its used to be before TensorFlow 1.12 I think). So, to get access to the `Callback`, please do as I suggested in the comments.

See here a tutorial from tensorflow: https://www.tensorflow.org/guide/keras/writing_your_own_callbacks?hl=en

IMO, we can say we only support TF 2.6 and on to skip the version checking... But lets keep it for now I guess (I'll ask Nir if we have anyone who use TF with lower versions).

guy1992l (approved):

assaf758 (approved):

PR #5380, opened 2024-04-03
[Docs] Add class documentation for KafkaTarget

george0st:


mlrun/datastore/targets.py
```diff
@@ -1413,6 +1413,27 @@ def as_df(self, columns=None, df_module=None, **kwargs):
...
+            path='my_topic',
+            bootstrap_servers='localhost:9092'
+        )
+
+        # ingest
+        stocks_set.ingest(stocks, [kafka_target])
+
+    :param name:        optional, target name.
+    :param path:        optional, topic name e.g. 'my_topic' or 'v3io:///{topic}'.
+    :param bootstrap_servers:   optional, kafka broker e.g. 'localhost:9092'.
```
gtopper: This parameter was just deprecated, and `brokers` was added to replace it.

mlrun/datastore/targets.py
```diff
@@ -1413,6 +1413,27 @@ def as_df(self, columns=None, df_module=None, **kwargs):
...
+            "kafka",
+            path='my_topic',
+            bootstrap_servers='localhost:9092'
+        )
+
+        # ingest
+        stocks_set.ingest(stocks, [kafka_target])
+
+    :param name:        optional, target name.
+    :param path:        optional, topic name e.g. 'my_topic' or 'v3io:///{topic}'.
```
gtopper: It's not really optional though, despite being technically optional in the constructor.

mlrun/datastore/targets.py
```diff
@@ -1413,6 +1413,27 @@ def as_df(self, columns=None, df_module=None, **kwargs):
...
+    """Kafka target storage driver, used to materialize feature set
+     data into kafka topics.
+
+    example::
+
+        # define target
+        kafka_target = KafkaTarget(
+            "kafka",
+            path='my_topic',
+            bootstrap_servers='localhost:9092'
```
gtopper:
```suggestion
            name="kafka",
            path="my_topic",
            bootstrap_servers="localhost:9092",
```

mlrun/datastore/targets.py
```diff
@@ -1413,6 +1413,27 @@ def as_df(self, columns=None, df_module=None, **kwargs):
...
+        )
+
+        # ingest
+        stocks_set.ingest(stocks, [kafka_target])
```
gtopper, line 1530: Maybe it's better not to mention feature store here, since datastore can be used in other contexts as well, e.g. in serving functions. This applies to the docstring topic line as well.

mlrun/datastore/targets.py
```diff
@@ -1413,6 +1413,27 @@ def as_df(self, columns=None, df_module=None, **kwargs):
...
+
+    :param name:        optional, target name.
+    :param path:        optional, topic name e.g. 'my_topic' or 'v3io:///{topic}'.
+    :param bootstrap_servers:   optional, kafka broker e.g. 'localhost:9092'.
+    :param producer_options:    optional, addition options/settings for producer.
```
gtopper:
```suggestion
    :param producer_options:    optional, additional options/settings for kafka producer.
```

gtopper (commented):
Thank you for the addition! I have a couple of suggestions.

mlrun/datastore/targets.py
george0st: I removed that from docstring, but I see that example show only one possible use case from a lot. Is it fine? (NOTE: similar sample is on different places).

gtopper:
Opened #5463 based on this PR.

PR #5381, opened 2024-04-03
[Artifacts] Fix retained producer overwriting artifact spec

alonmr:
Producer of kind "run" are used to resolve the artifact project, target_path and db_key.
However, when we retain the producer, all of the above should not be resolved by the run metadata as the artifact now belongs to another project.
Also, fixed export/import artifacts to keep the target path of the exported artifacts.


theSaarco (commented):
@alonmr - as discussed, let's instead pass `is_retained` as parameter to `log_artifact` in the import flow, and avoid the entire `generate_target_path` in that case - the target path should be retained as-is in that flow.
Also, please add a check to the test that verifies that the target path of the imported artifact is the same as the original target path.

TomerShor (approved):

mlrun/projects/project.py
```diff
@@ -1778,14 +1786,16 @@ def get_artifact(spec):
...
                 # if the dataitem is not a file, it means we downloaded it from a remote source to a temp file,
                 # so we need to remove it after we're done with it
                 dataitem.remove_local()
 
                 return self.log_artifact(
-                    artifact, local_path=temp_dir, artifact_path=artifact_path
+                    artifact,
+                    local_path=temp_dir,
+                    artifact_path=artifact_path,
+                    upload=True,
```
theSaarco, line 1798: Why is this needed? Looks like it's unrelated to this PR? Is this an opportunistic fix for another issue?

theSaarco (approved):
LGTM. Have a question about one of the changes, but otherwise looks great.

PR #5383, opened 2024-04-04
[Requirements] Allow for newer versions of fsspec

gtopper:
[ML-5934](https://iguazio.atlassian.net/browse/ML-5934)

[ML-5934]: https://iguazio.atlassian.net/browse/ML-5934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,7 +30,7 @@ v3io-frames~=0.10.12
 semver~=3.0
 dependency-injector~=4.41
 # should be identical to gcs and s3fs.
-fsspec==2023.9.2
+fsspec>=2023.9.2, <2024.4
 v3iofs~=0.1.17
 storey~=1.7.6
 inflection~=0.5.0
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -30,7 +30,7 @@ v3io-frames~=0.10.12
 semver~=3.0
 dependency-injector~=4.41
 # should be identical to gcs and s3fs.
-fsspec==2023.9.2
+fsspec>=2023.9.2, <2024.4
```
assaf758, line 33: does it actually select the newer fsspec when adlfs, ossfs are behind? Or they support the newer fsspec?

gtopper: * adlfs 2023.9.0 [requires](https://github.com/fsspec/adlfs/blob/2023.9.0/setup.py#L36) `fsspec>=2023.9.0`
* ossfs 2023.12.0 [requires](https://github.com/fsspec/ossfs/blob/2023.12.0/setup.cfg#L29) `fsspec>=2023.9.2`

So in both cases, there is no upper bound that would restrict fsspec.

assaf758 (approved):

PR #5387, opened 2024-04-04
[Pagination] Phase IV.5 - Fix No Pagination on authenticated request and Refactor Config

quaark:
- Found a bug in the filtered paginate request when not supplying a page_size, fixed it so it can handle that scenario.
- Added default_page_size to httpdb config, and added a pagination section to the config


server/api/crud/pagination_cache.py
```diff
@@ -78,8 +78,8 @@ def monitor_pagination_cache(session: sqlalchemy.orm.Session):
         """
 
         # Using cache TTL + 1 to make sure a zero TTL won't remove records that were just created
-        cache_ttl = mlrun.config.config.httpdb.pagination_cache.ttl + 1
-        table_max_size = mlrun.config.config.httpdb.pagination_cache.max_size
+        cache_ttl = mlrun.config.config.httpdb.pagination.pagination_cache.ttl + 1
+        table_max_size = mlrun.config.config.httpdb.pagination.pagination_cache.max_size
```
alonmr: nitpicking but use `mlconf` since its going to be a convention

server/api/utils/pagination.py
```diff
@@ -54,7 +55,7 @@ async def paginate_permission_filtered_request(
         auth_info: typing.Optional[mlrun.common.schemas.AuthInfo] = None,
         token: typing.Optional[str] = None,
         page: typing.Optional[int] = None,
-        page_size: typing.Optional[int] = None,
+        page_size: typing.Optional[int] = mlconf.httpdb.pagination.default_page_size,
```
alonmr: I understand its ok here since `default_page_size` is not mutable but I think it's better to not use variables as param defaults for the sake of avoiding "gotchas".
```
if page_size is None:
    page_size = mlconf.httpdb.pagination.default_page_size
```

server/api/utils/pagination.py
```diff
@@ -78,13 +79,16 @@ async def paginate_permission_filtered_request(
...
             if not pagination_info:
                 # no more results
                 break
+
             last_pagination_info = pagination_info
-            new_result = await filter_(new_result)
-            result.extend(new_result)
-            current_page = pagination_info["page"] + 1
+            current_page = last_pagination_info["page"] + 1
+            page_size = last_pagination_info["page_size"]
```
alonmr, line 91: if you don't have a page_size (by human mistake?) you will enter an infinite loop. I understand its unlikely but we should have sanity validation for it

alonmr (commented):
Small flavor comments

server/api/utils/pagination.py
quaark: no, this is the whole point of the token, which is why I added the second test, notice when you request with the token you don't need to supply a page size at all and we take it from cache. That's what I'm doing here

alonmr (approved):

liranbg (approved):

PR #5393, opened 2024-04-07
[Model Monitoring] Fix security issue - delete pipelines access key from serving env 

davesh0812:
https://iguazio.atlassian.net/browse/ML-6043, https://iguazio.atlassian.net/browse/ML-6040

### DIFF
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -1093,7 +1093,8 @@ class Config:
                                 relative artifact path will be taken from the global MLRun artifact path.
         :param function_name:    Application name, None for model_monitoring_stream.
 
-        :return:                Full configured path for the provided kind.
+        :return:                Full configured path for the provided kind. Can be either a single path
+                                or a list of paths in the case of the online model monitoring stream path.
         """
 
         if target != "offline":
@@ -1115,7 +1116,7 @@ class Config:
                     if function_name is None
                     else f"{kind}-{function_name.lower()}",
                 )
-            elif kind == "stream":  # ML-6043
+            elif kind == "stream":  # return list for mlrun<1.6.3 BC
                 return [
                     mlrun.mlconf.model_endpoint_monitoring.store_prefixes.default.format(
                         project=project,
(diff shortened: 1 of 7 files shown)


mlrun/config.py
```diff
@@ -1115,10 +1115,22 @@ def get_model_monitoring_file_target_path(
                     if function_name is None
                     else f"{kind}-{function_name.lower()}",
                 )
-            return mlrun.mlconf.model_endpoint_monitoring.store_prefixes.default.format(
-                project=project,
-                kind=kind,
-            )
+            elif kind == "stream":  # ML-6043
```
Eyal-Danieli:
```suggestion
            elif kind == "stream":  # ML-6043
```
I suggest to remove the `#ML-6043` notes and explain here that we return 2 stream paths in order to handle backwards compatibility of old mlrun versions (<1.7)

mlrun/config.py
```diff
@@ -1077,7 +1077,7 @@ def get_model_monitoring_file_target_path(
         target: str = "online",
         artifact_path: str = None,
         function_name: str = None,
-    ) -> str:
+    ) -> typing.Union[str, list[str]]:
```
Eyal-Danieli, line 1080: cool, I suggest to update the `return` docstring as well:
`Full configured path for the provided kind. Can be either a single path or a list of paths in the case of the online model monitoring stream path.`

server/api/crud/model_monitoring/deployment.py
```diff
@@ -117,34 +117,42 @@ def deploy_model_monitoring_stream_processing(
...
-                db_session=self.db_session, project=self.project
+            overwrite=overwrite,
+        ):
+            # Get parquet target value for model monitoring stream function
+            parquet_target = (
+                server.api.crud.model_monitoring.helpers.get_monitoring_parquet_path(
+                    db_session=self.db_session, project=self.project
+                )
+            )
+            if overwrite and stream_image.startswith("mlrun/mlrun"):
```
Eyal-Danieli: As we talked, the image doesn't really matter. We just need to add the new trigger in this case.

Eyal-Danieli (commented):
great job
few comments

Eyal-Danieli (approved):

PR #5399, opened 2024-04-08
[Pagination] Phase V.1 - Runs Pagination Endpoint Support

quaark:
https://iguazio.atlassian.net/browse/ML-6112
https://iguazio.atlassian.net/browse/ML-5804

Contains general pagination alterations:
- Support both async and sync filters and methods. If the paginated method is sync, run it in a thread pool.
- Pagination dynamically creates Pydantic schemas to use when serializing/deserializing the paginated method kwargs to/from the pagination cache table

Runs pagination:
- Add page, page-size and page-token params to `list_runs` endpoint.
- Implement db `list_runs` pagination with limit + offset
- Use pagination mechanism in `list_runs` endpoint for caching paginated requests
- Add sanity UT for paginated list runs using all new params - page, page-size and token (extensive pagination testing is in test_pagination.py)


server/api/utils/pagination.py
```diff
@@ -190,3 +201,13 @@ def _create_or_update_pagination_cache_record(
             kwargs=method_kwargs,
         )
         return token, page, page_size, method, method_kwargs
+
+    @staticmethod
+    async def _call_or_await_method(
```
alonmr: IMO can be moved to `server.api.utils.asyncio`

server/api/utils/pagination.py
```diff
@@ -79,7 +80,7 @@ async def paginate_permission_filtered_request(
                 page_size,
                 **method_kwargs,
             )
-            new_result = await filter_(new_result)
+            new_result = await self._call_or_await_method(False, filter_, new_result)
```
alonmr: why False?

server/api/utils/pagination.py
```diff
@@ -132,15 +135,23 @@ def paginate_request(
...
                 method=method.__name__,
             )
-            return method(**method_kwargs, page=page, page_size=page_size), {
+            return await self._call_or_await_method(
+                True, method, session, **method_kwargs, page=page, page_size=page_size
+            ), {
                 "token": token,
                 "page": page,
-                "page_size": page_size,
+                "page-size": page_size,
```
alonmr: I think we can create a class to encapsulate this instead of returning a dict

alonmr (commented):
Small comments / questions

server/api/utils/pagination.py
quaark: Because I didn't believe there'd be a sync filter I need to run in a thread pool but I guess better be safe than sorry

liranbg (approved):

liranbg (approved):
since it mutates openapi schema - lets have a note on that
overall we can tackle it in parallel to the pagination effort of course

PR #5416, opened 2024-04-11
[Application] Forward query params to sidecar

alonmr:
https://iguazio.atlassian.net/browse/ML-6150


liranbg (changes requested):
Use url query builder to rebuild the path and then dump it to string

liranbg (approved):

PR #5417, opened 2024-04-14
[Pipelines] Isolate version-specific KFP code on its own modules

laurybueno:
https://iguazio.atlassian.net/browse/ML-4717

The work compiled here moves away from MLRun most of the knowledge about Kubeflow Pipelines and its APIs. All version-specific KFP code, in particular, is now fully managed outside the main MLRun package.

To achieve this, we introduce a new package to the MLRun ecosystem: `mlrun_pipelines`, which is dynamically composed as a [namespaced entity](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) by multiple [distribution packages](https://packaging.python.org/en/latest/discussions/distribution-package-vs-import-package/). 

These new distribution packages are called `mlrun-pipelines-kfp-common` and `mlrun-pipelines-kfp-v1-8`. The former has code that any known KFP version can share, while the latter manages everything specific to v1.8.

The following PR for the ML-4717 epic will introduce `mlrun-pipelines-kfp-v2` and leverage the namespaced nature of `mlrun_pipelies` to switch all version-specific code while requiring no changes on MLRun. 

Note that in this PR, the distribution packages have a `-experiment` suffix to their names. This is a temporary measure since we are not using Iguazio keys to send them to PyPI at the moment. Refer to [this discussion for details](https://iguazio.slack.com/archives/C067YLKLZG9/p1712834134105309?thread_ts=1712825975.258139&cid=C067YLKLZG9).

[ML-4717]: https://iguazio.atlassian.net/browse/ML-4717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/helpers.py
```diff
@@ -0,0 +1,144 @@
...
+
+        Args:
+            key: the key to look up.
+
+        Returns:
+            the value associated with the key.
+
+        Raises:
+            KeyError: if the key is not found
+        """
```
TomerShor: Use ReStructured format for docstrings as in the rest of mlrun (see https://iguazio.atlassian.net/wiki/spaces/TLE/pages/2457650/MLRun+Python+-+Coding+Conventions#:~:text=Docstring%20format%2C%20for%20all%20public%20API%2Dlevel%20functions%3A).
Same goes for the rest of the methods in this file

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -594,78 +379,58 @@ def build_op(
...
-    is_v3io = function.spec.build.source and function.spec.build.source.startswith(
-        "v3io"
-    )
-    if "V3IO_ACCESS_KEY" in os.environ and is_v3io:
-        cop.container.add_env_variable(
-            k8s_client.V1EnvVar(
-                name="V3IO_ACCESS_KEY", value=os.environ.get("V3IO_ACCESS_KEY")
-            )
-        )
+    generate_image_builder_pipeline_node(name, function, func_url, cmd)
```
TomerShor: missing `return`?

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -594,78 +379,58 @@ def build_op(
...
 
-    auth_env_var = mlrun.runtimes.constants.FunctionEnvironmentVariables.auth_session
-    if auth_env_var in os.environ or "V3IO_ACCESS_KEY" in os.environ:
-        cop.container.add_env_variable(
-            k8s_client.V1EnvVar(
-                name=auth_env_var,
-                value=os.environ.get(auth_env_var) or os.environ.get("V3IO_ACCESS_KEY"),
-            )
-        )
+    generate_deployer_pipeline_node(
```
TomerShor: missing `return`?

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/models.py
```diff
@@ -0,0 +1,117 @@
...
+            main_manifest = json.loads(workflow_manifest)
+        except TypeError:
+            main_manifest = workflow_manifest
+        if pipeline_manifest:
+            pipeline_manifest = json.loads(pipeline_manifest)
+            main_manifest["status"] = pipeline_manifest.get("status", {})
+        super().__init__(main_manifest)
+
+
+class PipelineRun(FlexibleMapper):
```
TomerShor, line 41: Is this class planned to be different in KFP 2?
If not, it might reside in the common package

TomerShor (commented):
Really nice work!
I like the namespaced packages approach.
Have some really minor comments that I noticed, but otherwise LGTM 👌

mlrun/runtimes/pod.py
```diff
@@ -19,8 +19,9 @@
 from enum import Enum
 
 import dotenv
-import kfp.dsl
 import kubernetes.client as k8s_client
+import mlrun_pipelines.mounts
```
theSaarco, line 23: The fact that modifiers' import path has shifted to another location will break backwards compatibility to many users, which `import` these modifiers and use them in their code.
We must still support importing those from their previous location - `mlrun.v3io_cred` etc. We can consider deprecating the "old" import path, but we cannot simply remove it like this.

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -335,6 +210,7 @@ def mlrun_pipeline(
             image = image or function.spec.kfp_image or config.dask_kfp_image
 
     image = image or config.kfp_image
+    image = "laury/mlrun:kfp2"
```
theSaarco: Should not keep this image.

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/models.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @finished_at.setter
+    def finished_at(self, finished_at):
+        self._external_data["finished_at"] = finished_at
+
+
+class PipelineExperiment(FlexibleMapper):
+    @property
+    def id(self):
+        return self._external_data["id"]
```
theSaarco, line 113: Do we need a setter for this? Or do we expect this to only be set on creation?

theSaarco (changes requested):
Looks pretty solid. My main concern is about the BC issue with the mount modifiers.

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/helpers.py
laurybueno: Great shout! I fixed it now. Thanks!

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
laurybueno: great catch! I added it back. Thanks!

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
laurybueno: another awesome catch! Thanks!

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/models.py
laurybueno: Yeah, the implementation code will be very different, but the class name, methods, and arguments will be identical. So MLRun will call the v2 version the same way it does with this one. 

Neat stuff if I do say so myself 😅

mlrun/runtimes/pod.py
laurybueno: Indeed, but I think this PR accounts for this case by [adding import aliases to the main mlrun module](https://github.com/laurybueno/mlrun/blob/c0bfab9e7f4a3563e7dcbac67607251cb6aa1998/mlrun/__init__.py#L72-L75), doesn't it? 

I also added the aliases to the `__all__` list to include wildcard imports (stuff like `from mlrun import *`).

The original import paths seemed to work in my tests, but let me know if I'm missing something.

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
laurybueno: Great catch! I removed it now. Thanks, Saar!

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/models.py
laurybueno: That's a good call. I removed this setter now. It's not being used, nor it should.

pipeline-adapters/mlrun-pipelines-kfp-common/setup.py
```diff
@@ -0,0 +1,42 @@
...
+
+from setuptools import find_namespace_packages, setup
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger("mlrun-kfp-setup")
+
+setup(
+    name="mlrun-pipelines-kfp-common-experiment",
+    version="0.1.3",
+    description="MLRun Pipelines package for providing KFP 1.8 compatibility",
```
liranbg, line 25: for provide common kfp functionality

liranbg (approved):
Really small and def not a blocker

PR #5420, opened 2024-04-14
[Formatting] Run the Ruff Formatter on docstrings

jond01:
Enable with a flag in `pyproject.toml`'s `[tool.ruff.format]` table.

All of the changes were automatic, except for:
* Fix `-` -> `=` in `AWS_ACCESS_KEY_ID` example in `project.py`.

PR #5427, opened 2024-04-16
[Docs] Fix text block formatting

jond01:
Fix the spaces `ENV_VAR = VAL` --> `ENV_VAR=VAL` in the docstrings introduced in https://github.com/mlrun/mlrun/pull/5420.
Remove the spaces in the text code block by specifying the language.

The API reference after this change:
image

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -2669,17 +2669,20 @@ class MlrunProject(ModelObj):
         file_path: str = None,
         provider: typing.Union[str, mlrun.common.schemas.SecretProviderName] = None,
     ):
-        """set project secrets from dict or secrets env file
+        """
+        Set project secrets from dict or secrets env file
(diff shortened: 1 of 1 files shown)


mlrun/projects/project.py
```diff
@@ -2669,17 +2669,20 @@ def set_secrets(
...
-        """set project secrets from dict or secrets env file
+        """
+        Set project secrets from dict or secrets env file
         when using a secrets file it should have lines in the form KEY=VALUE, comment line start with "#"
         V3IO paths/credentials and MLrun service API address are dropped from the secrets
 
-        example secrets file::
+        example secrets file:
+
+        .. code-block:: text
```
gtopper: How about this?
```suggestion
        .. code-block:: shell
```

jond01: Done :)
Indeed, it looks better:
image

gtopper (approved):

PR #5430, opened 2024-04-16
[Kafka] Fix handling of new `brokers` parameter in `KafkaTarget`

gtopper:
[ML-6211](https://iguazio.atlassian.net/browse/ML-6211)

[ML-6211]: https://iguazio.atlassian.net/browse/ML-6211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/datastore/targets.py
+++ b/mlrun/datastore/targets.py
@@ -1532,6 +1532,8 @@ class KafkaTarget(BaseStoreTarget):
         **kwargs,
     ):
         attrs = {}
+
+        # TODO: Remove this in 1.9.0
         if bootstrap_servers:
             if brokers:
                 raise mlrun.errors.MLRunInvalidArgumentError(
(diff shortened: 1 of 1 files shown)


mlrun/datastore/targets.py
```diff
@@ -1539,8 +1539,9 @@ def __init__(
                 # TODO: Remove this in 1.9.0
                 FutureWarning,
             )
-        if bootstrap_servers is not None:
-            attrs["brokers"] = brokers or bootstrap_servers
+            brokers = brokers or bootstrap_servers
```
assaf758: I would actually assert early on in the function that brokers and bootstrap_servers are not provided in the same constructor call - and raise an exception if they do.
After that, we can simply do `brokers = bootstrap_servers`

gtopper: Done.

assaf758 (approved):

PR #5432, opened 2024-04-16
[Datastore] Change v3io upload method to count on buffer instead of mmap

tomerm-iguazio:
1) Change v3io upload process, using buffer instead of mmap.
2) Change default chunk size to 100MB instead of 1 GB.

[ML-6103](https://iguazio.atlassian.net/browse/ML-6103)

The bug can be reproduced with any file of any size within the range of:
2097152< file_size < 10475524

Hotfix:
We may want to explore using mmap for improved performance in the future.

[ML-6103]: https://iguazio.atlassian.net/browse/ML-6103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/datastore/v3io.py
```diff
@@ -94,13 +93,13 @@ def get_storage_options(self):
...
 
-    def _upload(self, key: str, src_path: str, max_chunk_size: int = ONE_GB):
+    def _upload(self, key: str, src_path: str, max_chunk_size: int = MAX_CHUNK_SIZE):
         """helper function for upload method, allows for controlling max_chunk_size in testing"""
         container, path = split_path(self._join(key))
         file_size = os.path.getsize(src_path)  # in bytes
         if file_size <= ONE_MB:
             with open(src_path, "rb") as source_file:
-                data = source_file.read()
+                data = memoryview(source_file.read())
```
alxtkr77: I don't see a reason to keep the 'if file_size <= ONE_MB' case. The second case does exactly the same now

tests/system/datastore/test_v3io.py
```diff
@@ -89,13 +89,15 @@ def _setup_df_dir(self, first_file_path, second_file_path, file_extension):
     @pytest.mark.skip(
         reason="Skipping this test as it hangs when running against the CI system. ML-5598"
     )
-    def test_v3io_large_object_upload(self, tmp_path):
+    @pytest.mark.parametrize(
+        "file_size", [4 * 1024 * 1024, 20 * 1024 * 1024]
```
alxtkr77, line 93: Why those sizes? See no logic. Maybe 99Kb and 10Mb since the max_chunk_size is set to 100Kb?

mlrun/datastore/helpers.py
```diff
@@ -14,5 +14,5 @@
 #
 
 
-ONE_GB = 1024 * 1024 * 1024
 ONE_MB = 1024 * 1024
```
alxtkr77: ONE_MB can be removed probably as well

mlrun/datastore/helpers.py
```diff
@@ -14,5 +14,5 @@
...
+MAX_CHUNK_SIZE = 1024 * 1024 * 100
```
alxtkr77: I would name it somehow more clear: V3IO_UPLOAD_CHUNK_SIZE

tests/system/datastore/test_v3io.py
tomerm-iguazio: 4MB is in the range of this bug (as I mentioned in the pr description).
99Kb and 100MB or not in this range.

20MB was the previous size of the file of this test.

mlrun/datastore/v3io.py
```diff
@@ -94,13 +93,15 @@ def get_storage_options(self):
...
 
-    def _upload(self, key: str, src_path: str, max_chunk_size: int = ONE_GB):
+    def _upload(
+        self, key: str, src_path: str, max_chunk_size: int = V3IO_UPLOAD_CHUNK_SIZE
+    ):
         """helper function for upload method, allows for controlling max_chunk_size in testing"""
         container, path = split_path(self._join(key))
         file_size = os.path.getsize(src_path)  # in bytes
-        if file_size <= ONE_MB:
+        if file_size <= V3IO_UPLOAD_CHUNK_SIZE:
```
alxtkr77: Like I said all under 'if file_size <= V3IO_UPLOAD_CHUNK_SIZE' should be removed

mlrun/datastore/v3io.py
```diff
@@ -148,11 +139,13 @@ def get(self, key, size=None, offset=0):
...
 
-    def _put(self, key, data, append=False, max_chunk_size: int = ONE_GB):
+    def _put(
+        self, key, data, append=False, max_chunk_size: int = V3IO_UPLOAD_CHUNK_SIZE
+    ):
         """helper function for put method, allows for controlling max_chunk_size in testing"""
         container, path = split_path(self._join(key))
         buffer_size = len(data)  # in bytes
-        if buffer_size <= ONE_MB:
+        if buffer_size <= V3IO_UPLOAD_CHUNK_SIZE:
```
alxtkr77: Lines 155-163 should be removed , since no need for special case

alxtkr77 (approved):

PR #5436, opened 2024-04-17
[Pagination] Phase VI - SDK Uses Pagination

quaark:
https://iguazio.atlassian.net/browse/ML-5807

SDK now uses pagination to gradually load the items from the API instead of requesting them all at once.
The UX stays exactly the same and the sdk methods return all the items together with the exact same types, but the implementation makes it so requests to the API are chunked and alleviate stress on the API.


liranbg (approved):
NEAT

PR #5437, opened 2024-04-17
[Serving] Add new `set_flow()` API

gtopper:
[ML-6220](https://iguazio.atlassian.net/browse/ML-6220)

[ML-6220]: https://iguazio.atlassian.net/browse/ML-6220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/serving/states.py
```diff
@@ -346,6 +346,53 @@ def to(
         parent._last_added = step
         return step
 
+    def set_flow(
+        self,
+        steps: list[Union[str, StepToDict, dict[str, Any]]],
+        force: bool = False,
+    ):
```
theSaarco, line 356: Mention that you `:return:` the last step in the added flow.
Could even add an example where you're doing a `step1.set_flow(...).to()` or something similar.

mlrun/serving/states.py
```diff
@@ -346,6 +346,53 @@ def to(
...
+                    dict(name="step2", handler="step2_handler"),
+                    dict(name="step3", class_name="Step3Class"),
+                ],
+                force=True,
+            )
+        """
+        if hasattr(self, "steps"):
+            parent = self
+        elif self._parent:
+            parent = self._parent
```
theSaarco: I know this code is used also in the `to()` method, but what does it mean in this context? The `steps` attribute will exist in Flow steps, so what's happening in this part? Should this really apply to this scenario which should only work for Flow steps?

theSaarco (changes requested):
Almost there, but I think the logic should be more strict and **only** apply to flow steps.

mlrun/serving/states.py
gtopper: According to the requirement
> this should be enabled for any step in the graph

mlrun/serving/states.py
```diff
@@ -349,6 +349,54 @@ def to(
...
+        """set list of steps as downstream from this step, in the order specified. This will overwrite any existing
+        downstream steps.
+
+        :param steps: list of steps to follow this one
+        :param force: whether to overwrite existing downstream steps. If False, this method will fail if any downstream
+        steps have already been defined. Defaults to False.
+        :return: the last step added to the flow
+
+        example:
+            a 3-step pipeline (step1 -> step2 -> step3) to overwrite existing downstream steps:
```
assaf758: Since you are giving an example where `force` is used, I would demonstrate its effect:
```suggestion
            The below code sets the downstream nodes of step1 by using a list of steps (provided to `set_flow()`) and a single step (provided to `to()`), resulting in the graph (step1 -> step2 -> step3 -> step4).
            Notice that using `force=True` is required in case step1 already had downstream nodes (say the existing graph was step1->step2_old) and that following the execution of this code the existing downstreams nodes are removed. If the intention is to split the graph (and not to overwrite), please use `.to()`
```

gtopper: Updated with a few touch ups.

mlrun/serving/states.py
```diff
@@ -349,6 +349,59 @@ def to(
...
+        :param steps: list of steps to follow this one
+        :param force: whether to overwrite existing downstream steps. If False, this method will fail if any downstream
+        steps have already been defined. Defaults to False.
+        :return: the last step added to the flow
+
+        example:
+            The below code sets the downstream nodes of step1 by using a list of steps (provided to `set_flow()`) and a
+            single step (provided to `to()`), resulting in the graph (step1 -> step2 -> step3 -> step4).
+            Notice that using `force=True` is required in case step1 already had downstream nodes (e.g. if the existing
+            graph were step1 -> step2_old) and that following the execution of this code the existing downstream steps
```
assaf758:
```suggestion
            graph was step1 -> step2_old) and that following the execution of this code the existing downstream steps
```

assaf758 (approved):

mlrun/serving/states.py
```diff
@@ -349,6 +349,59 @@ def to(
...
+                f"step {self.name} parent is not set or it's not part of a graph"
+            )
+
+        if not force and parent._steps:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "'set_flow' called on a step that already has downstream steps. "
+                "If you want to overwrite existing steps, set force=True"
+            )
+
+        parent.steps = None
```
theSaarco: This only works if `parent` is the same as `step`. If you did `parent = self._parent`, then you actually removed `step` from the parent, and you're left with a detached step that you add the flow to.

theSaarco (changes requested):
We should finish our discussion on the parent topic. Note that a new issue exists now.

theSaarco (approved):
LGTM.

PR #5439, opened 2024-04-17
[Datastore] Rely on buffer instead of mmap

tomerm-iguazio:
1) Change v3io upload process, using buffer instead of mmap.
2) Change default chunk size to 100MB instead of 1 GB.

[ML-6103](https://iguazio.atlassian.net/browse/ML-6103)

The bug can be reproduced with any file of any size within the range of:
2097152< file_size < 10475524

Hotfix:
We may want to explore using mmap for improved performance in the future.

original pr:
https://github.com/mlrun/mlrun/pull/5432

benchmark tests:

Total benchmark check - of 6 GB file.

Buffer + memoryview based upload  VS mmap - in seconds:

**Buffer + memoryview**
100 MB chunks - 60
50 MB chunks - 57.11
10 MB chunks - 35.711
5 MB chunks - 33.04
1 MB chunks - 50.89

**Mmap**
100 MB chunks - 86.46
10 MB chunks - 83.25
1 MB chunks - 53.37
100 KB - 126.10

[ML-6103]: https://iguazio.atlassian.net/browse/ML-6103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/datastore/v3io.py
```diff
@@ -32,6 +30,7 @@
 )
 
 V3IO_LOCAL_ROOT = "v3io"
+V3IO_UPLOAD_CHUNK_SIZE = 1024 * 1024 * 100
```
gtopper: Should mention "default" somewhere in the name.

gtopper: Isn't a 100MB a bit high? This PR would increase the chunk size by a factor of 100.

mlrun/datastore/v3io.py
```diff
@@ -98,46 +97,26 @@ def get_storage_options(self):
...
-                    file_offset += chunk_size
+                append = file_offset != 0
+                self._do_object_request(
+                    self.object.put,
+                    container=container,
+                    path=path,
+                    body=file_obj.read(chunk_size),
+                    append=append,
+                )
+                file_offset += chunk_size
```
gtopper: I think this loop can be simplified now (no need to keep offsets)
```python
        with open(src_path, "rb") as file_obj:
            append = False
            while True:
                data = file_obj.read(max_chunk_size)
                if not data:
                    break
                self._do_object_request(
                    self.object.put,
                    container=container,
                    path=path,
                    body=data,
                    append=append,
                )
                append = True
```
should be safer as it makes fewer assumptions (e.g. about concurrent writes).

gtopper (approved):
@tomerm-iguazio, this would be a good place to put your benchmark results.

PR #5443, opened 2024-04-18
[Notifications] Pipeline Notifications from Workflow Manifest

quaark:
https://iguazio.atlassian.net/browse/ML-2426

- Pipeline notifications now look at the kubeflow workflow manifest in order to build the notification data with all steps, in addition to runs and jobs, also function builds and nuclio function deployments.
- Moved `mlrun.runtimes.constants` to `mlrun.common.runtime.constants` to resolve import loop
- Refactor Alert System test to make the nuclio notification receiver common
- Add system test for pipeline notification that ensures all the steps were included in the notification data

Screenshot 2024-05-02 at 14 39 48


mlrun/utils/notifications/notification_pusher.py
```diff
@@ -395,6 +388,109 @@ def _update_notification_status(
...
+        if not kfp_run:
+            return None
+
+        kfp_run = kfp_run.to_dict()
+        try:
+            return json.loads(kfp_run["pipeline_runtime"]["workflow_manifest"])
+        except Exception:
+            return None
+
+    def _extract_function_uri(self, function_uri):
```
liranbg: give a short example of docstring with example of inpiut and output

mlrun/utils/notifications/notification_pusher.py
```diff
@@ -395,6 +388,109 @@ def _update_notification_status(
...
+                project=run.metadata.project,
+                labels=f"workflow={workflow_id}",
+            )
+
+    @staticmethod
+    def _get_workflow_manifest(workflow_id: str) -> typing.Optional[dict]:
+        kfp_client = kfp.Client(namespace=mlrun.config.config.namespace)
+
+        # arbitrary timeout of 5 seconds, the workflow should be done by now
+        kfp_run = kfp_client.wait_for_run_completion(workflow_id, 5)
```
liranbg, line 487: and if not? not sure why do you assume it is done by now? because the notifications are sent once workflow is done?

mlrun/utils/notifications/notification_pusher.py
```diff
@@ -395,6 +388,109 @@ def _update_notification_status(
...
+        if not kfp_run:
+            return None
+
+        kfp_run = kfp_run.to_dict()
+        try:
+            return json.loads(kfp_run["pipeline_runtime"]["workflow_manifest"])
+        except Exception:
+            return None
```
liranbg, line 495: when would that happen?

mlrun/utils/notifications/notification_pusher.py
```diff
@@ -395,6 +388,109 @@ def _update_notification_status(
...
+                )
+                step_type = node_template["metadata"]["annotations"].get(
+                    "mlrun/pipeline-step-type"
+                )
+                step_method = step_methods.get(step_type)
+                if step_method:
+                    step_method(node_name, node_template)
+            return steps
+        except Exception:
+            # If we fail to read the pipeline steps, we will return the list of runs that have the same workflow id
```
liranbg, line 470: if running on api - I suggest logging the error

liranbg (changes requested):

mlrun/utils/notifications/notification_pusher.py
quaark: Exactly

mlrun/utils/notifications/notification_pusher.py
quaark: It shouldn't, but it is a string that we did not create (kubeflow did). If the string is malformed for some reason, I'd rather it still send a partial notification than explode, no?

tests/system/common/helpers/notifications.py
```diff
@@ -0,0 +1,50 @@
...
+) -> typing.Generator[dict, None, None]:
+    response = requests.post(nuclio_function_url, json={"operation": "list"})
+    response_data = json.loads(response.text)
+
+    # Extract notification data from the response
+    notifications = response_data["data_list"]
+
+    yield from notifications
+
+    requests.post(nuclio_function_url, json={"operation": "reset"})
```
TomerShor, line 50: it is not obvious from the name that it is also resets the cache, perhaps add is as a docstring at least?

tests/system/common/assets/notification_nuclio_function.py
```diff
@@ -31,6 +31,9 @@ def handler(context, event):
     if operation == "add":
         context.user_data.data_list.append(data)
 
+    if runs := event_data.get("runs"):
+        context.user_data.data_list.append(runs)
```
TomerShor: is this used at all?..

quaark: Yes of course, That's how I check at the end that all the runs arrived

tests/system/common/helpers/notifications.py
quaark: You are right will do

tests/system/common/assets/notification_nuclio_function.py
TomerShor: It just seems that `get_notifications_from_nuclio_and_reset_notification_cache` only does the `list` operation. are you using a different function to check this?

quaark: Added some clarifications to eliminate confusions there :)

liranbg (approved):

PR #5456, opened 2024-04-23
[Auth] Authenticate with OAuth client-credentials grant

theSaarco:
Allow MLRun client to work in an IDE when service is behind an ingress using OAuth authentication. This is mostly targeting MLRun CE support, where IDE work would be done with client-credentials (or service account) support.
This implements the following:
* Expand the current token support of `httpdb` - generate token-provider classes that allow for a static token (existing implementation) and also oauth client-id tokens.
* User can configure client-id support through the `auth_with_client_id` config parameter, and provide the rest of parameters (token-endpoint url, client-id and client-secret) through env variables.
* When issuing calls to BE, `httpdb` will use the token endpoint to retrieve a token, and will plant it in the REST call headers.
* The token will be refreshed as needed (after 1/2 of the expiry time), retrieving a new token before issuing calls.
* Not part of the main use-case, but jobs and runs executed through MLRun can also utilize oauth tokens to authenticate - this is not usually needed since those will talk to MLRun service directly (not through ingress), but if the need arises this can be done through project secrets.

Implements [ML-6229](https://iguazio.atlassian.net/browse/ML-6229).

[ML-6229]: https://iguazio.atlassian.net/browse/ML-6229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/rundb/test_httpdb.py
```diff
@@ -323,10 +326,92 @@ def test_bearer_auth(create_server):
...
+    requests_mock.reset_mock()
+
+    db.trigger_migrations()
+    assert len(requests_mock.request_history) == 2
+    assert "Authorization" not in requests_mock.last_request.headers
+    assert db.token_provider.token is None
+
+    mlrun.mlconf.auth_with_client_id = False
+    for key in test_env:
+        del os.environ[key]
```
gtopper: Should be done in teardown. Otherwise, if the test fails, the environment will stay modified.

mlrun/config.py
```diff
@@ -691,6 +691,7 @@
         # supported modes: "enabled", "disabled".
         "mode": "disabled"
     },
+    "auth_with_client_id": False,
```
alonmr: can be under `httpdb.authentication` no?

mlrun/db/auth_utils.py
```diff
@@ -0,0 +1,142 @@
...
+# limitations under the License.
+
+from abc import ABC, abstractmethod
+from datetime import datetime, timedelta
+
+import requests
+
+import mlrun.errors
+
+from ..utils import logger
```
alonmr: don't use relative imports, they are prone to circular imports issues

mlrun/db/auth_utils.py
```diff
@@ -0,0 +1,142 @@
...
+    def __init__(self, token_endpoint: str, client_id: str, client_secret: str):
+        if not token_endpoint or not client_id or not client_secret:
+            raise mlrun.errors.MLRunValueError(
+                "Invalid client_id configuration for authentication. Must provide token endpoint, client-id and secret"
+            )
+        self.token_endpoint = token_endpoint
+        self.client_id = client_id
+        self.client_secret = client_secret
+
+        self._cleanup()
```
alonmr, line 65: is this to force refresh? maybe add a comment

mlrun/db/auth_utils.py
```diff
@@ -0,0 +1,142 @@
...
+        session = requests.Session()
+        try:
+            headers = {"Content-Type": "application/x-www-form-urlencoded"}
+            request_body = {
+                "grant_type": "client_credentials",
+                "client_id": self.client_id,
+                "client_secret": self.client_secret,
+            }
+            response = session.request(
+                "POST",
```
alonmr, line 101: will it always be POST?

mlrun/db/auth_utils.py
```diff
@@ -0,0 +1,142 @@
...
+                self.token_endpoint,
+                timeout=5,
```
alonmr: suggest making the timeout configurable and maybe add a retry as well see https://iguazio.atlassian.net/browse/ML-5810

mlrun/db/httpdb.py
```diff
@@ -54,6 +54,7 @@
     normalize_name,
     version,
 )
+from .auth_utils import OAuthClientIDTokenProvider, StaticTokenProvider
```
alonmr: same relative import

mlrun/db/httpdb.py
```diff
@@ -138,17 +139,27 @@ def _enrich_and_validate(self, url):
...
-        self.base_url = base_url
         self.user = username
         self.password = password
-        self.token = token
+        self.token_provider = None
+
+        if config.auth_with_client_id:
+            token_endpoint = mlrun.get_secret_or_env("MLRUN_AUTH_TOKEN_ENDPOINT")
+            client_id = mlrun.get_secret_or_env("MLRUN_AUTH_CLIENT_ID")
+            client_secret = mlrun.get_secret_or_env("MLRUN_AUTH_CLIENT_SECRET")
```
alonmr: I would move this to the constructor e.g.
```
self.token_endpoint = token_endpoint or mlrun.get_secret_or_env("MLRUN_AUTH_TOKEN_ENDPOINT")
```

alonmr (commented):
Looks great! Some small suggestions

liranbg:
Adding on top of @alonmr review - worth using the returyable http client to ensure retries upon transient connection issues

mlrun/config.py
theSaarco: I thought about it, but as a matter of fact the `httpdb` part in the config is server-side configuration (with a terrible name), and is mostly used by the server side code. I didn't want to break that convention, though the temptation was there.

mlrun/db/auth_utils.py
theSaarco: No, this is just to initialize the class - members need to be set or else they cannot be accessed later (the method to refresh the token checks the value of `self.token`) - using this is simply a shortcut to set these members to `None`.

mlrun/db/auth_utils.py
theSaarco: yes

mlrun/db/auth_utils.py
```diff
@@ -0,0 +1,142 @@
...
+                return
+
+        if not response.ok:
+            error = "No error available"
+            if response.content:
+                try:
+                    data = response.json()
+                    error = data.get("error")
+                except Exception:
+                    pass
```
gtopper, line 122: Or we could just include the error as is? It might not be json.

mlrun/db/auth_utils.py
```diff
@@ -0,0 +1,142 @@
...
+            logger.warning("Retrieving token failed", error=error)
```
gtopper: Status code might be worth mentioning too?

tests/rundb/test_httpdb.py
```diff
@@ -323,10 +326,92 @@ def test_bearer_auth(create_server):
...
+
+    token_url = "https://mock/token_endpoint/protocol/openid-connect/token"
+    test_env = {
+        "MLRUN_AUTH_TOKEN_ENDPOINT": token_url,
+        "MLRUN_AUTH_CLIENT_ID": "some-client-id",
+        "MLRUN_AUTH_CLIENT_SECRET": "some-client-secret",
+    }
+
+    mlrun.mlconf.auth_with_client_id = True
+    os.environ.update(test_env)
```
jond01: In the same spirit of Gal's comment, I suggest using pytest's  `monkeypatch` fixture:
https://docs.pytest.org/en/latest/reference/reference.html#pytest.MonkeyPatch.setenv
Proper teardown guaranteed 😉

mlrun/db/auth_utils.py
theSaarco: I'm ok with configurable timeout, I'm not really sure about retries, since in most cases we don't retry on `POST` anyway, so it is possible to have an exception here, but I didn't see much value in that.

mlrun/db/httpdb.py
theSaarco: You mean within the provider? I deliberately didn't want to have the provider understand what env variables are used to provide its configuration - it should do the OAuth work, not MLRun configuration.

mlrun/db/auth_utils.py
theSaarco: This is what we do in all other clients, I guess if it works for everyone else...

mlrun/db/auth_utils.py
```diff
@@ -0,0 +1,142 @@
...
+            # We only cleanup if token was really expired - even if we fail in refreshing the token, we can still
+            # use the existing one given that it's not expired.
+            if now >= self.token_expiry_time:
+                self._cleanup()
+
+        self._issue_token_request()
+        return self.token
+
+    def _issue_token_request(self, raise_on_error=False):
+        session = requests.Session()
```
gtopper: This session is created and never close. Instead, either
1. keep it in `self`, or
2. replace `session.request(` with `requests.request(` (in which case it won't reuse the connection as (1) would, but also won't leak connections).

mlrun/db/auth_utils.py
theSaarco: Sure it is.

tests/rundb/test_httpdb.py
theSaarco: I actually looked for ways to do that and `monkeypatch` was an option I considered - I seem to recall seeing some discussions talking about how `monkeypatch` persists beyond a single test and causing "interesting" issues. I'll check again and if it works I'll indeed use it.

tests/rundb/test_httpdb.py
theSaarco: yeah, I was looking for ways to do that in a consistent way (and not having to generate a fixture just for that), and as I mentioned to @jond01 on his comment - 

> I actually looked for ways to do that and `monkeypatch` was an option I considered - I seem to recall seeing some discussions talking about how `monkeypatch` persists beyond a single test and causing "interesting" issues. I'll check again and if it works I'll indeed use it.

Eventually I just forgot about it. I'll see what I can do here.

mlrun/db/auth_utils.py
```diff
@@ -0,0 +1,142 @@
...
+        try:
+            headers = {"Content-Type": "application/x-www-form-urlencoded"}
```
gtopper, line 94: Shouldn't it be `application/json`? Removing this and replacing
```
data=request_body,
```
with 
```
json=request_body,
```
would do that.

theSaarco: The spec of this API calls for `urlencoded` content-type. The requests lib encodes the dictionary as needed. Seemed to me it's worthwhile to be specific here.

mlrun/db/auth_utils.py
theSaarco: I'm moving to a retry-able session anyway.

tests/rundb/test_httpdb.py
theSaarco: Anyway, decided to use `monkeypatch` - seems ok.

liranbg (approved):

mlrun/db/httpdb.py
```diff
@@ -218,17 +229,19 @@ def api_call(
...
+                # Iguazio auth doesn't support passing token through bearer, so use cookie instead
+                if self.token_provider.is_iguazio_session():
+                    session_cookie = f'j:{{"sid": "{token}"}}'
+                    cookies = {
+                        "session": session_cookie,
+                    }
+                    kw["cookies"] = cookies
+                else:
+                    if "Authorization" not in kw.setdefault("headers", {}):
+                        kw["headers"].update({"Authorization": "Bearer " + token})
```
quaark, line 244: How about having each token provider class support a `generate_token_headers` method? So we don't have this `if/else` that might grow in the future maybe

quaark (changes requested):
Looks awesome!
Very minor suggestion

quaark (approved):

PR #5460, opened 2024-04-24
[Alerts] Convert event value to be dict to allow multiple values

yanburman:
Add more values to the model monitoring event that would be of interest for the user

https://iguazio.atlassian.net/browse/ML-6233

### DIFF
--- a/mlrun/model_monitoring/writer.py
+++ b/mlrun/model_monitoring/writer.py
@@ -117,6 +117,7 @@ class ModelMonitoringWriter(StepToDict):
             notification_types=[NotificationKind.slack]
         )
         self._create_tsdb_table()
+        self._endpoints_records = {}
 
     @staticmethod
     def get_v3io_container(project_name: str) -> str:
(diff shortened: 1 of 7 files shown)


mlrun/model_monitoring/writer.py
```diff
@@ -227,10 +232,20 @@ def do(self, event: _RawEvent) -> None:
         _Notifier(event=event, notification_pusher=self._custom_notifier).notify()
 
         if mlrun.mlconf.alerts.mode == mlrun.common.schemas.alert.AlertsModes.enabled:
+            endpoint_record = get_endpoint_record(
```
davesh0812: I think it is nice to have the model name in the message, but I think we don't want to access the db in each writer run. Try to save all the model_endpoint you already get from the db like we are doing in the `EventStreamProcessor::ProcessEndpointEvent.resume_state`.

davesh0812 (commented):
Looks good!
I had one comment on the implementation. 
About the testing maybe we can had in the system test a check to ensure the message is has we expected, if it is possible.

davesh0812 (approved):
LGTM

theSaarco (approved):
LGTM.

liranbg (approved):

PR #5462, opened 2024-04-25
[Nuclio] Delete Nuclio function when removing function of Nuclio runtime

rokatyy:
This PR addresses a bug where deleting an MLRun function from the Nuclio runtime only removed the function from the database, leaving the associated Nuclio pod running.

Since an MLRun function can be mapped to multiple Nuclio functions with different tags, the deletion process first checks if the function is associated with Nuclio. Then, it retrieves all tags belonging to this function, generates Nuclio function names, and sends requests to the Nuclio API to delete each Nuclio function mapped to the given MLRun function.

Jira - https://iguazio.atlassian.net/browse/ML-3432
https://iguazio.atlassian.net/browse/ML-6293

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -1149,7 +1149,7 @@ class HTTPRunDB(RunDBInterface):
             )
             background_task = mlrun.common.schemas.BackgroundTask(**response.json())
             background_task = self._wait_for_background_task_to_reach_terminal_state(
-                background_task.metadata.name
+                background_task.metadata.name, project=project
             )
             if (
                 background_task.status.state
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -1142,7 +1142,23 @@ class HTTPRunDB(RunDBInterface):
         project = project or config.default_project
         path = f"projects/{project}/functions/{name}"
         error_message = f"Failed deleting function {project}/{name}"
-        self.api_call("DELETE", path, error_message)
+        response = self.api_call("DELETE", path, error_message, version="v2")
+        if response.status_code == http.HTTPStatus.ACCEPTED:
+            logger.info(
+                "Function is being deleted", project_name=project, function_name=name
+            )
+            background_task = mlrun.common.schemas.BackgroundTask(**response.json())
+            background_task = self._wait_for_background_task_to_reach_terminal_state(
+                background_task.metadata.name
+            )
+            if (
+                background_task.status.state
+                == mlrun.common.schemas.BackgroundTaskState.succeeded
+            ):
+                logger.info(
+                    "Function deleted", project_name=project, function_name=name
+                )
+                return
 
     def list_functions(self, name=None, project=None, tag=None, labels=None):
         """Retrieve a list of functions, filtered by specific criteria.
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -1158,7 +1158,13 @@ class HTTPRunDB(RunDBInterface):
                 logger.info(
                     "Function deleted", project_name=project, function_name=name
                 )
-                return
+            elif (
+                background_task.status.state
+                == mlrun.common.schemas.BackgroundTaskState.failed
+            ):
+                logger.info(
+                    "Function deletion failed", project_name=project, function_name=name
+                )
 
     def list_functions(self, name=None, project=None, tag=None, labels=None):
         """Retrieve a list of functions, filtered by specific criteria.
--- a/server/api/api/endpoints/functions.py
+++ b/server/api/api/endpoints/functions.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import asyncio
 import traceback
 from distutils.util import strtobool
 from http import HTTPStatus
@@ -190,18 +191,22 @@ async def delete_function(
     fn = await run_in_threadpool(
         server.api.crud.Functions().get_function, db_session, name, project
     )
-    if fn.get("kind") in RuntimeKinds.pure_nuclio_deployed_runtimes():
+    if fn.get("kind") in RuntimeKinds.nuclio_runtimes():
         # getting all function tags, so we can delete all nuclio functions mapped to given function
-
         all_nuclio_functions_mapped_to_function = await run_in_threadpool(
             server.api.crud.Functions().list_functions, db_session, project, name
         )
+
         async with server.api.utils.clients.async_nuclio.Client(auth_info) as client:
+            tasks = []
             for function in all_nuclio_functions_mapped_to_function:
                 nuclio_name = mlrun.runtimes.nuclio.function.get_fullname(
                     name, project, function.get("metadata", {}).get("tag")
                 )
-                await client.delete_function(name=nuclio_name, project_name=project)
+                tasks.append(
+                    client.delete_function(name=nuclio_name, project_name=project)
+                )
+            await asyncio.gather(*tasks)
 
     await run_in_threadpool(
         server.api.crud.Functions().delete_function, db_session, project, name
--- a/server/api/api/endpoints/functions.py
+++ b/server/api/api/endpoints/functions.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import asyncio
 import traceback
 from distutils.util import strtobool
 from http import HTTPStatus
@@ -41,7 +40,6 @@ import server.api.db.session
 import server.api.launcher
 import server.api.utils.auth.verifier
 import server.api.utils.background_tasks
-import server.api.utils.clients.async_nuclio
 import server.api.utils.clients.chief
 import server.api.utils.functions
 import server.api.utils.pagination
@@ -138,8 +136,13 @@ async def get_function(
     }
 
 
+# TODO: Remove in 1.9.0
 @router.delete(
-    "/projects/{project}/functions/{name}", status_code=HTTPStatus.NO_CONTENT.value
+    "/projects/{project}/functions/{name}",
+    status_code=HTTPStatus.NO_CONTENT.value,
+    deprecated=True,
+    description="'/projects/{project}/functions/{name}' will be removed in 1.9.0, "
+    "use '/v2/projects/{project}/functions/{name}' instead.",
 )
 async def delete_function(
     request: Request,
@@ -187,26 +190,6 @@ async def delete_function(
             await run_in_threadpool(
                 get_scheduler().delete_schedule, db_session, project, name
             )
-
-    # getting all function tags
-    functions = await run_in_threadpool(
-        server.api.crud.Functions().list_functions, db_session, project, name
-    )
-    if len(functions) > 0:
-        if functions[0].get("kind") in RuntimeKinds.nuclio_runtimes():
-            async with server.api.utils.clients.async_nuclio.Client(
-                auth_info
-            ) as client:
-                tasks = []
-                for function in functions:
-                    nuclio_name = mlrun.runtimes.nuclio.function.get_fullname(
-                        name, project, function.get("metadata", {}).get("tag")
-                    )
-                    tasks.append(
-                        client.delete_function(name=nuclio_name, project_name=project)
-                    )
-                await asyncio.gather(*tasks)
-
     await run_in_threadpool(
         server.api.crud.Functions().delete_function, db_session, project, name
     )
--- a/server/api/api/endpoints/functions.py
+++ b/server/api/api/endpoints/functions.py
@@ -40,6 +40,7 @@ import server.api.db.session
 import server.api.launcher
 import server.api.utils.auth.verifier
 import server.api.utils.background_tasks
+import server.api.utils.clients.async_nuclio
 import server.api.utils.clients.chief
 import server.api.utils.functions
 import server.api.utils.pagination
@@ -185,6 +186,23 @@ async def delete_function(
             await run_in_threadpool(
                 get_scheduler().delete_schedule, db_session, project, name
             )
+
+    fn = await run_in_threadpool(
+        server.api.crud.Functions().get_function, db_session, name, project
+    )
+    if fn.get("kind") in RuntimeKinds.pure_nuclio_deployed_runtimes():
+        # getting all function tags, so we can delete all nuclio functions mapped to given function
+
+        all_nuclio_functions_mapped_to_function = await run_in_threadpool(
+            server.api.crud.Functions().list_functions, db_session, project, name
+        )
+        async with server.api.utils.clients.async_nuclio.Client(auth_info) as client:
+            for function in all_nuclio_functions_mapped_to_function:
+                nuclio_name = mlrun.runtimes.nuclio.function.get_fullname(
+                    name, project, function.get("metadata", {}).get("tag")
+                )
+                await client.delete_function(name=nuclio_name, project_name=project)
+
     await run_in_threadpool(
         server.api.crud.Functions().delete_function, db_session, project, name
     )
--- a/server/api/api/endpoints/functions.py
+++ b/server/api/api/endpoints/functions.py
@@ -188,25 +188,24 @@ async def delete_function(
                 get_scheduler().delete_schedule, db_session, project, name
             )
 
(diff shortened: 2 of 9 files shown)


server/api/api/endpoints/functions.py
```diff
@@ -185,6 +186,23 @@ async def delete_function(
             await run_in_threadpool(
                 get_scheduler().delete_schedule, db_session, project, name
             )
+
+    fn = await run_in_threadpool(
+        server.api.crud.Functions().get_function, db_session, name, project
+    )
+    if fn.get("kind") in RuntimeKinds.pure_nuclio_deployed_runtimes():
```
alonmr: Should apply for application runtime as well
```suggestion
    if fn.get("kind") in RuntimeKinds.nuclio_runtimes():
```

server/api/api/endpoints/functions.py
```diff
@@ -185,6 +186,23 @@ async def delete_function(
...
+
+        all_nuclio_functions_mapped_to_function = await run_in_threadpool(
+            server.api.crud.Functions().list_functions, db_session, project, name
+        )
+        async with server.api.utils.clients.async_nuclio.Client(auth_info) as client:
+            for function in all_nuclio_functions_mapped_to_function:
+                nuclio_name = mlrun.runtimes.nuclio.function.get_fullname(
+                    name, project, function.get("metadata", {}).get("tag")
+                )
+                await client.delete_function(name=nuclio_name, project_name=project)
```
alonmr: use asyncio.gather()

alonmr (changes requested):
Nice job! I agree we can start with this and discuss if we want to tackle to monitoring flow as well.
Also, please add a couple of unit tests.

server/api/api/endpoints/functions.py
```diff
@@ -185,6 +186,23 @@ async def delete_function(
...
                 get_scheduler().delete_schedule, db_session, project, name
             )
+
+    fn = await run_in_threadpool(
+        server.api.crud.Functions().get_function, db_session, name, project
+    )
+    if fn.get("kind") in RuntimeKinds.pure_nuclio_deployed_runtimes():
+        # getting all function tags, so we can delete all nuclio functions mapped to given function
+
+        all_nuclio_functions_mapped_to_function = await run_in_threadpool(
```
TomerShor:
```suggestion
    if fn.get("kind") in RuntimeKinds.pure_nuclio_deployed_runtimes():
    
        # getting all function tags, so we can delete all nuclio functions mapped to given function
        all_nuclio_functions_mapped_to_function = await run_in_threadpool(
```

server/api/api/endpoints/functions.py
TomerShor, line 1: I think the entire new functionality should be in `crud/functions.py` instead of endpoints/functions.py` as the CRUD is the one doing the actual deletion of the resources.

TomerShor (changes requested):

server/api/api/endpoints/functions.py
rokatyy: @TomerShor crud is not async

TomerShor: I thought you'll say it, but there is no reason it cannot be async - you can change it.

rokatyy: @TomerShor Yes, it's possible to do so, but I've checked all the places where it is used and I am not sure if it worth it.

server/api/api/endpoints/functions.py
rokatyy: @alonmr great idea!

server/api/api/endpoints/functions.py
TomerShor: And don't we want to delete the nuclio function from the other flows `CRUD.Functions.delete_functions()` is called?

rokatyy: @TomerShor on API side that's the only place we delete function (except periodic).

server/api/api/endpoints/functions.py
```diff
@@ -185,6 +187,26 @@ async def delete_function(
...
+            ) as client:
+                tasks = []
+                for function in functions:
+                    nuclio_name = mlrun.runtimes.nuclio.function.get_fullname(
+                        name, project, function.get("metadata", {}).get("tag")
+                    )
+                    tasks.append(
+                        client.delete_function(name=nuclio_name, project_name=project)
+                    )
+                await asyncio.gather(*tasks)
```
liranbg: few things here are not very clear
1. the heuristic if checking the first function is nuclio and assuming rest are nuclio - perhaps worth a comment why do oyu assume it
2. gather on tasks - if you have 500 functions, do you want to send 500 requests in parallel? why not group those requests by 10ens and send them all (remember not to fail on first but send all requests, log what fails and if one failed, then at the end - fail the operation
3. will you wait for functions deletion? Id assume that is a long-process - thus, needs to be a background job.
4. as stated on (2) - gather will fail if one deletion has failed. I suggest to wait for all before bailing

---
since (3) requires a change on response, I suggest moving this operation to functions_v2.py and have this function return background job
then you update 
httpdb (sdk) to expect background job (and use v2)
open ticket for ui to use /v2/... endpoint and expect background job

liranbg (changes requested):

server/api/api/endpoints/functions.py
TomerShor: `crud/functions` is also on API side, so I don't see a reason not to move it there.
The reasons I insist on this is:
1. This is the way MLRun is designed - `Endpoints` handle the requests / validations, `CRUD` handles the resources.
2. We don't want to to have any flow that won't delete the function in Nuclio.

rokatyy: @TomerShor okay, then I think before making it async we need to confirm one thing. `Crud.delete_function()` is used here https://github.com/mlrun/mlrun/blob/57e09417293d03bfff4377d06acd54178600afaa/server/api/rundb/sqldb.py#L276, `SQLDb` is an implementation of Base interface which has also other implementation of `delete_function` in httpdb an nopDB. Httpdb is used on user side. So, do we really want to do it async?

server/api/utils/background_tasks/kinds.py
```diff
@@ -18,3 +18,4 @@ class BackgroundTaskKinds:
     db_migrations = "db.migrations"
     project_deletion = "project.deletion.{0}"
     project_deletion_wrapper = "project.deletion.wrapper.{0}"
+    function_deletion = "project.deletion.{0}"
```
TomerShor:
```suggestion
    function_deletion = "function.deletion.{0}"
```

server/api/api/endpoints/functions.py
```diff
@@ -136,8 +136,13 @@ async def get_function(
     }
 
 
+# TODO: Remove in 1.9.0
 @router.delete(
-    "/projects/{project}/functions/{name}", status_code=HTTPStatus.NO_CONTENT.value
+    "/projects/{project}/functions/{name}",
+    status_code=HTTPStatus.NO_CONTENT.value,
+    deprecated=True,
+    description="'/projects/{project}/functions/{name}' will be removed in 1.9.0, "
```
TomerShor:
```suggestion
    description="'v1/projects/{project}/functions/{name}' will be removed in 1.9.0, "
```

server/api/api/utils.py
```diff
@@ -1276,3 +1277,111 @@ def _verify_project_is_deleted():
...
+async def delete_nuclio_functions_in_batches(auth_info, project_name, function_names):
+    async def delete_function(
+        nuclio_client: server.api.utils.clients.iguazio.AsyncClient,
+        project,
+        function,
+    ):
+        try:
+            await nuclio_client.delete_function(name=function, project_name=project)
+            return None
+        except Exception as e:
```
```suggestion
        except Exception as exc:
```

server/api/api/utils.py
```diff
@@ -1276,3 +1277,111 @@ def _verify_project_is_deleted():
...
+            # return tuple with failure info
+            return function, str(e)
+
+    # batch size for deleting functions in parallel
+    batch_size = 10
```
TomerShor: make it configurable in `config.py`

server/api/api/utils.py
```diff
@@ -1276,3 +1277,111 @@ def _verify_project_is_deleted():
...
+    # delete the function from the database
+    await run_in_threadpool(
+        server.api.crud.Functions().delete_function,
+        db_session,
+        project_name,
+        function_name,
+    )
+
+
+async def delete_nuclio_functions_in_batches(auth_info, project_name, function_names):
```
TomerShor: add type hints

tests/api/api/test_functions.py
```diff
@@ -228,12 +231,31 @@ async def test_list_functions_with_hash_key_versioned(
...
+@pytest.mark.parametrize("use_v2_deletion_endpoint", [True, False])
+@unittest.mock.patch.object(server.api.utils.clients.async_nuclio, "Client")
+@unittest.mock.patch.object(
+    server.api.utils.clients.async_nuclio.Client, "delete_function"
+)
+def test_delete_function(
+    patched_nuclio_client,
+    patched_delete_nuclio_function,
     db: sqlalchemy.orm.Session,
     client: fastapi.testclient.TestClient,
```
TomerShor: Use `unversioned_client` instead, as `client` uses `v1` prefix by default

tests/api/api/test_functions.py
```diff
@@ -228,12 +231,31 @@ async def test_list_functions_with_hash_key_versioned(
...
-def test_delete_function_with_schedule(
+@pytest.mark.parametrize(
+    "post_schedule, kind",
+    [
+        (True, "job"),
+        (False, "job"),
+        (False, "remote"),
+    ],
+)
+@pytest.mark.parametrize("use_v2_deletion_endpoint", [True, False])
```
TomerShor:
```suggestion
@pytest.mark.parametrize("endpoint_prefix, expected_status", 
        [
            ("v1", HTTPStatus.NO_CONTENT.value),
            ("v2", HTTPStatus.ACCEPTED.value)
        ]
)
```

server/api/api/endpoints/functions.py
rokatyy: As discussed, not relevant anymore

TomerShor (changes requested):

server/api/api/utils.py
rokatyy: @TomerShor why?

tests/api/api/test_functions.py
rokatyy: @TomerShor default value is with /v1/ (it is configured in fastapi.base_url), I'm not sure if there is a sense to do all logic with `v1` explicitly since it's already there

server/api/api/utils.py
TomerShor: Wanted to align with the rest of the code, although I see that there are different uses.
Also, I don't like single letter variables

server/api/api/utils.py
```diff
@@ -1276,3 +1277,111 @@ def _verify_project_is_deleted():
...
+    auth_info: mlrun.common.schemas.AuthInfo,
+):
+    background_task_kind_format = (
+        server.api.utils.background_tasks.BackgroundTaskKinds.function_deletion
+    )
+    background_task_kind = background_task_kind_format.format(function_name)
+    background_task_name = str(uuid.uuid4())
+
+    # create the background task for function deletion
+    return server.api.utils.background_tasks.InternalBackgroundTasksHandler().create_background_task(
```
quaark: We shouldn't use the `InternalBackgroundTasksHandler` but `ProjectBackgroundTasksHandler` instead. 
- `InternalBackgroundTasksHandler` is managed in memory and run ONLY on the chief. 
- While `ProjectBackgroundTasksHandler` saves the tasks in the DB (related to some project) and the tasks can be run on any worker - This is what you want in your case.

I'm guessing you were inspired by the `delete_project` task, however the ONLY reason we use `InternalBackgroundTasksHandler` there is that the project deletion will delete the Background Task resource as well while it is still running. But this won't happen in your case.

quaark (changes requested):
Looks great!
One comment though I suggest we use the other Background Task handler

server/api/api/utils.py
rokatyy: @quaark Make sense! Thank you for the clarification, I was indeed inspired by `delete_project`

quaark: No problem 😄

server/api/api/endpoints/functions.py
rokatyy: @liranbg reworked

mlrun/db/httpdb.py
```diff
@@ -1142,7 +1142,23 @@ def delete_function(self, name: str, project: str = ""):
...
-        self.api_call("DELETE", path, error_message)
+        response = self.api_call("DELETE", path, error_message, version="v2")
+        if response.status_code == http.HTTPStatus.ACCEPTED:
+            logger.info(
+                "Function is being deleted", project_name=project, function_name=name
+            )
+            background_task = mlrun.common.schemas.BackgroundTask(**response.json())
+            background_task = self._wait_for_background_task_to_reach_terminal_state(
+                background_task.metadata.name
+            )
```
quaark:
```suggestion
            background_task = self._wait_for_background_task_to_reach_terminal_state(
                background_task.metadata.name, project=project
            )
```
So it will get the project background task and not the internal one

quaark (changes requested):
Looks good! One minor fix for SDK side

quaark (approved):
LGTM 🚀

mlrun/db/httpdb.py
```diff
@@ -1142,7 +1142,23 @@ def delete_function(self, name: str, project: str = ""):
...
+                background_task.metadata.name, project=project
+            )
+            if (
+                background_task.status.state
+                == mlrun.common.schemas.BackgroundTaskState.succeeded
+            ):
+                logger.info(
+                    "Function deleted", project_name=project, function_name=name
+                )
+                return
```
TomerShor: `failed` is also a terminal state, add a log if the task failed

liranbg: redundant return

server/api/api/utils.py
```diff
@@ -1276,3 +1277,109 @@ def _verify_project_is_deleted():
...
+    # delete the function from the database
+    await run_in_threadpool(
+        server.api.crud.Functions().delete_function,
+        db_session,
+        project,
+        function_name,
+    )
+
+
+async def delete_nuclio_functions_in_batches(
```
```suggestion
async def _delete_nuclio_functions_in_batches(
```

TomerShor (changes requested):

server/api/api/utils.py
```diff
@@ -1276,3 +1277,109 @@ def _verify_project_is_deleted():
...
+
+            results = await asyncio.gather(*tasks, return_exceptions=True)
+
+            # process results to identify failed deletion requests
+            for result in results:
+                if isinstance(result, tuple):
+                    nuclio_name, error_message = result
+                    failed_requests.append(
+                        f"Failed to delete nuclio function {nuclio_name}: {error_message}"
+                    )
```
liranbg: that's some sorcery, you can simply use semaphore 
```
semaphore = asyncio.Semaphore(N)

async def delete_function(function):
  async with semaphore:
    await self._delete_it()

tasks = [delete_function(function) for function in functions ]
results = await asyncio.gather(*tasks, return_exceptions=True)
for result in results:
  ...
```

liranbg (changes requested):

liranbg (approved):

server/api/api/endpoints/functions_v2.py
```diff
@@ -0,0 +1,114 @@
...
+from server.api.api import deps
+from server.api.utils.singletons.scheduler import get_scheduler
+
+router = APIRouter()
+
+
+@router.delete(
+    "/projects/{project}/functions/{name}",
+    responses={
+        http.HTTPStatus.NO_CONTENT.value: {},
```
TomerShor: In what flow is `NO_CONTENT` returned?

rokatyy: @TomerShor the only flow we have is creating a background task. So, in any case

rokatyy: @TomerShor good point! removed it

TomerShor (approved):

PR #5463, opened 2024-04-25
[Docs] Add docstring for `KafkaTarget`

gtopper:
Based on #5380.

### DIFF
--- a/mlrun/datastore/targets.py
+++ b/mlrun/datastore/targets.py
@@ -1529,8 +1529,8 @@ class KafkaTarget(BaseStoreTarget):
     :param path:                topic name e.g. "my_topic"
     :param bootstrap_servers:   Deprecated. Use the brokers parameter instead.
     :param producer_options:    additional configurations for kafka producer
-    :param brokers:             kafka broker, or list of kafka brokers, e.g. "localhost:9092", or
-        ["kafka-broker-1:9092", "kafka-broker-2:9092"]
+    :param brokers:             kafka broker as represented by a host:port pair, or a list of kafka brokers, e.g.
+        "localhost:9092", or ["kafka-broker-1:9092", "kafka-broker-2:9092"]
     """
 
     kind = TargetTypes.kafka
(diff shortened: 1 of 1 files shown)


mlrun/datastore/targets.py
```diff
@@ -1516,6 +1516,23 @@ def as_df(self, columns=None, df_module=None, **kwargs):
...
+    example::
+        # define target
+        kafka_target = KafkaTarget(
+            name="kafka", path="my_topic", brokers="localhost:9092"
+        )
+        # ingest
+        stocks_set.ingest(stocks, [kafka_target])
+    :param name:                target name.
+    :param path:                topic name e.g. "my_topic"
+    :param bootstrap_servers:   Deprecated. Use the brokers parameter instead.
```
assaf758: Since it's deprecated, to ease readability, please move below "brokers"

mlrun/datastore/targets.py
```diff
@@ -1516,6 +1516,23 @@ def as_df(self, columns=None, df_module=None, **kwargs):
...
+    :param producer_options:    additional configurations for kafka producer
+    :param brokers:             kafka broker, or list of kafka brokers, e.g. "localhost:9092", or
+        ["kafka-broker-1:9092", "kafka-broker-2:9092"]
```
assaf758:
```suggestion
    :param brokers:             kafka broker url (e.g. "127.0.0.1:9092"), or a list of kafka broker urls, 
       (e.g ["kafka-broker-1:9092", "kafka-broker-2:9092"])
```

assaf758 (changes requested):

mlrun/datastore/targets.py
gtopper: The docstring order should match the actual parameter order... Which can't be changed for bwc.

mlrun/datastore/targets.py
gtopper: Those are not URLs.

assaf758 (approved):

PR #5464, opened 2024-04-25
[Nuclio] Refactor API Gateway class to inherit from `ModelObj`

TomerShor:
Align API Gateway object with the rest of the MLRun objects by inheriting from `ModelObj` class.
This organizes the API Gateway properties under `metadata` and `spec`, as well as  allows using the known methods such as `to_dict` / `to_yaml`.


liranbg (approved):

mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -134,21 +202,26 @@ def __init__(
...
+            path=path,
+            authentication=authentication,
+            functions=None,
+            canary=canary,
+        )
+
+        self._validate(
+            functions=functions,
             canary=canary,
         )
```
rokatyy: Let's move validate_function and canary to APIGatewaySpec

TomerShor: makes sense 👍

mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -134,21 +295,21 @@ def __init__(
...
         :param canary: The canary percents for the API gateway of type list[int]; for instance: [20,80]
         """
-        self.functions = None
-        self._validate(
+        self.metadata = APIGatewayMetadata(
+            name=name,
             project=project,
+        )
+
+        self.spec = APIGatewaySpec(
```
alonmr: see the following example on how to define meta and spec. constructor should only have meta and spec params
```
    @property
    def metadata(self) -> ProjectMetadata:
        return self._metadata

    @metadata.setter
    def metadata(self, metadata):
        self._metadata = self._verify_dict(metadata, "metadata", ProjectMetadata)

    @property
    def spec(self) -> ProjectSpec:
        return self._spec

    @spec.setter
    def spec(self, spec):
        self._spec = self._verify_dict(spec, "spec", ProjectSpec)
```

mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -92,7 +93,167 @@ def to_scheme(
...
+            Union[
+                list[
+                    Union[
+                        RemoteRuntime,
+                        ServingRuntime,
+                    ]
+                ],
+                Union[RemoteRuntime, ServingRuntime],
+            ],
+        ],
```
alonmr, line 238: I feel like this can be simpler. Can't union accept more than 2 options?

alonmr (changes requested):

liranbg (approved):

PR #5465, opened 2024-04-25
[Feature Store] Read optimisation on partitioned target

tomerm-iguazio:
1) Add support for additional_filters:
  a) In local_merger + dask_merger.
  b) In ParquetSource.
  c) In ParquetTarget.
  d) get_offline_features and more.
2) add warning for non-supporting source/targets/mergers.
3) add tests for dask and local engines.
4) bump storey.

[ML-6161](https://iguazio.atlassian.net/browse/ML-6161)

[ML-6161]: https://iguazio.atlassian.net/browse/ML-6161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/datastore/base.py
```diff
@@ -193,11 +205,15 @@ def set_filters(
...
+            if filters and filters_inner:
+                filters[0] += filters_inner
+
             kwargs["filters"] = filters
 
         def reader(*args, **kwargs):
-            if start_time or end_time:
-                if time_column is None:
+            time_limit = start_time or end_time
+            if time_limit or additional_filters:
```
gtopper: This if should start after the first check which is unrelated to `additional_filters`.

mlrun/datastore/base.py
```diff
@@ -193,11 +205,15 @@ def set_filters(
...
             )
+            if filters and filters_inner:
+                filters[0] += filters_inner
+
             kwargs["filters"] = filters
 
         def reader(*args, **kwargs):
-            if start_time or end_time:
-                if time_column is None:
+            time_limit = start_time or end_time
```
gtopper: Bad name. Better just inline it.

mlrun/datastore/base.py
```diff
@@ -217,6 +234,7 @@ def reader(*args, **kwargs):
                     ):
                         raise ex
 
+                    #  TODO fix this code, if end_time has tz and start_time is None?
```
gtopper: Looks like you're right. Please add a test to confirm the issue, and fix it. We're not just going to leave a TODO in the code about an obvious bug.

mlrun/datastore/base.py
```diff
@@ -550,6 +577,10 @@ def as_df(
         :param end_time:    filters out data after this time
         :param time_column: Store timestamp_key will be used if None.
                             The results will be filtered by this column and start_time & end_time.
+        :param additional_filters: (list of tuples, optional): List of filters conditions as tuples.
+                                                    Each tuple should be in the format (column_name, operator, value).
+                                                    Supported operators: '=', '>=', '<=', '>', '<'.
+                                                    Example: ('Product', '=', 'Computer')]
```
gtopper: Same as in storey, this needs a reference to the relevant pyarrow docs page.

gtopper: Unmatched square bracket.

gtopper: We use double quotes, even in the examples.

mlrun/datastore/inmem.py
```diff
@@ -80,8 +80,8 @@ def as_df(self, url, subpath, columns=None, df_module=None, format="", **kwargs)
             reader = df_module.read_json
         else:
             raise mlrun.errors.MLRunInvalidArgumentError(f"file type unhandled {url}")
-        # InMemoryStore store do not filter on time
-        for field in ["time_column", "start_time", "end_time"]:
+        # InMemoryStore store do not use filters
```
gtopper:
```suggestion
        # InMemoryStore store – don't pass filters
```

mlrun/utils/helpers.py
```diff
@@ -1596,3 +1596,13 @@ def get_serving_spec():
             )
     spec = json.loads(data)
     return spec
+
+
+def additional_filters_warning(additional_filters, class_name):
+    if additional_filters and any(
+        additional_filter for additional_filter in additional_filters
```
gtopper: That's a redundant for expression.
```suggestion
        additional_filter
```

tests/system/feature_store/test_feature_store.py
```diff
@@ -4813,6 +4813,77 @@ def test_merge_different_number_of_entities(self):
...
         assert_frame_equal(expected_all, df, check_dtype=False)
 
+    @staticmethod
+    def _get_sorted_df(df: pd.DataFrame, sort_column: str):
+        result = (
+            df.reindex(sorted(df.columns), axis=1)
+            .sort_values(by=sort_column)
+            .reset_index(drop=True)
+        )
+        return result
```
gtopper:
```suggestion
    def _sort_df(df: pd.DataFrame, sort_column: str):
        return (
            df.reindex(sorted(df.columns), axis=1)
            .sort_values(by=sort_column)
            .reset_index(drop=True)
        )
```

tests/system/feature_store/test_feature_store.py
```diff
@@ -4813,6 +4813,77 @@ def test_merge_different_number_of_entities(self):
...
+            .to_dataframe()
+            .reset_index(drop=False)
+        )
+        expected = self._get_sorted_df(filtered_df.query("bad == 95"), "patient_id")
+        result = self._get_sorted_df(result, "patient_id")
+        result["department"] = result["department"].astype("str")
+        if engine == "dask":
+            # Because Dask changes this column to "string[pyarrow]" type.
+            result["patient_id"] = result["patient_id"].astype("str")
+        assert_frame_equal(result, expected)
```
gtopper:
```suggestion
        assert_frame_equal(result, expected, check_dtype=False)
```

tests/system/feature_store/test_feature_store.py
```diff
@@ -4813,6 +4813,77 @@ def test_merge_different_number_of_entities(self):
...
+            name="department_based_target",
+            path=v3io_parquet_target_path,
+            partitioned=True,
+            partition_cols=["department"],
+        )
+        feature_set.ingest(source=parquet_source, targets=[target])
+        result = target.as_df(additional_filters=("room", "=", 1)).reset_index(
+            drop=False
+        )
+        # We want to include patient_id in the comparison and to sort both the columns and the values.
```
gtopper: Sort both the columns and the values? That doesn't really make sense.

tests/system/feature_store/test_feature_store.py
```diff
@@ -4813,6 +4813,77 @@ def test_merge_different_number_of_entities(self):
...
+        result = self._get_sorted_df(result, "patient_id")
+        expected = self._get_sorted_df(filtered_df.query("room == 1"), "patient_id")
+        #  reset category type to string:
+        result["department"] = result["department"].astype("str")
+        assert_frame_equal(result, expected)
```
gtopper:
```suggestion
        assert_frame_equal(result, expected, check_dtype=False)
```

tests/system/feature_store/test_feature_store.py
```diff
@@ -4813,6 +4813,77 @@ def test_merge_different_number_of_entities(self):
...
+
+        target = ParquetTarget(
+            name="department_based_target",
+            path=v3io_parquet_target_path,
+            partitioned=True,
+            partition_cols=["department"],
+        )
+        feature_set.ingest(source=parquet_source, targets=[target])
+        result = target.as_df(additional_filters=("room", "=", 1)).reset_index(
+            drop=False
```
gtopper:
```suggestion
```

tests/system/feature_store/test_feature_store.py
```diff
@@ -4813,6 +4813,77 @@ def test_merge_different_number_of_entities(self):
...
+        )
+        result = (
+            fstore.get_offline_features(
+                feature_vector=vec,
+                additional_filters=[("bad", "=", 95)],
+                with_indexes=True,
+                engine=engine,
+            )
+            .to_dataframe()
+            .reset_index(drop=False)
```
gtopper:
```suggestion
            .reset_index()
```

gtopper (changes requested):
Looks good in principle but I do have some comments. See below.

mlrun/datastore/base.py
tomerm-iguazio: I started to do so, but it seems to be more complicated than it looks. I will open an issue and add the right issue number to the TODO.

tests/system/feature_store/test_feature_store.py
tomerm-iguazio: I prefer to check all types that I can, and for the one which changed, to check their expected result.

tests/system/feature_store/test_feature_store.py
tomerm-iguazio: The entity column after reset index go to position number 0. That's why we need to order by column name as well.

tests/system/feature_store/test_feature_store.py
tomerm-iguazio: want to check the types that are not changing, for those that do change, test if I get the expected result.

mlrun/utils/helpers.py
tomerm-iguazio: what about [()]? the condition will return True, that's why I need to check every tuple in the list.

tomerm-iguazio: right, tnx!

mlrun/datastore/base.py
```diff
@@ -217,6 +232,7 @@ def reader(*args, **kwargs):
                     ):
                         raise ex
 
+                    #  TODO ML-6308 fix.
```
```suggestion
                    # TODO: fix ... (ML-6308)
```
and replace ... with an explanation.

I'm not sure I understand what ML-6308 says though. Is it even a proven bug?

tests/system/feature_store/test_feature_store.py
gtopper: Please notice that my suggestion also renames the function (looks like you only inlined `result`).

tests/system/feature_store/test_feature_store.py
gtopper: Comparing the exact dtype is almost always useless though.

tests/system/feature_store/test_feature_store.py
gtopper: Do you mean that we sort the columns alphabetically and we sort the rows by the index? If so, that's what it should say.

tests/system/feature_store/test_feature_store.py
```diff
@@ -4813,6 +4813,74 @@ def test_merge_different_number_of_entities(self):
...
+                engine=engine,
+            )
+            .to_dataframe()
+            .reset_index()
+        )
+        expected = self._get_sorted_df(filtered_df.query("bad == 95"), "patient_id")
+        result = self._get_sorted_df(result, "patient_id")
+        result["department"] = result["department"].astype("str")
+        if engine == "dask":
+            # Because Dask changes this column to "string[pyarrow]" type.
```
gtopper:
```suggestion
            # because Dask changes this column to "string[pyarrow]" type
```

mlrun/datastore/sources.py
```diff
@@ -281,6 +289,10 @@ class ParquetSource(BaseSourceDriver):
     :parameter start_time: filters out data before this time
     :parameter end_time: filters out data after this time
     :parameter attributes: additional parameters to pass to storey.
+    :param additional_filters: (list of tuples, optional): List of additional_filters conditions as tuples.
+
+                                For all supported filters, please see:
+                                https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetDataset.html
```
gtopper: No example here in particular?

mlrun/feature_store/api.py
```diff
@@ -175,6 +176,14 @@ def get_offline_features(
                                     By default, the filter executes on the timestamp_key of each feature set.
                                     Note: the time filtering is performed on each feature set before the
                                     merge process using start_time and end_time params.
+    :param additional_filters: (list of tuples, optional): List of additional_filters conditions as tuples.
```
gtopper: Would be nice to follow the convention of how we specify type and optionality in the rest of the code base. I guess there isn't exactly one way that we do this, but better not introduce one more way. Docs should be consistent.

gtopper (changes requested):

tests/system/feature_store/test_feature_store.py
tomerm-iguazio: fixed, tnx

tests/system/feature_store/test_feature_store.py
tomerm-iguazio: fixed comment:)

tests/system/feature_store/test_feature_store.py
tomerm-iguazio: The problem here is that the type was changed to category, if I use check_dtype=False I will still get an error.
I can use check_categorical=False, but in this way it will not check the values of this column.

So I think that the current code is the best option:)

tomerm-iguazio: added also check_categorical=False

tests/system/feature_store/test_feature_store.py
tomerm-iguazio: added check_categorical=False

mlrun/datastore/base.py
```diff
@@ -550,6 +575,12 @@ def as_df(
         :param end_time:    filters out data after this time
         :param time_column: Store timestamp_key will be used if None.
                             The results will be filtered by this column and start_time & end_time.
+        :param additional_filters: List of additional_filters conditions as tuples.
```
```suggestion
        :param additional_filters: List of additional filter conditions as tuples.
```

gtopper (approved):
LGTM otherwise.

PR #5467, opened 2024-04-28
[Tests] Change v3io test chunks size

tomerm-iguazio:
Change Test chunks size for better test performance.

[ML-5598](https://iguazio.atlassian.net/browse/ML-5598)

[ML-5598]: https://iguazio.atlassian.net/browse/ML-5598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/datastore/v3io.py
+++ b/mlrun/datastore/v3io.py
@@ -29,7 +29,7 @@ from .base import (
 )
 
 V3IO_LOCAL_ROOT = "v3io"
-V3IO_DEFAULT_UPLOAD_CHUNK_SIZE = 1024 * 1024 * 100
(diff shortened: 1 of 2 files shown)


gtopper (approved):
I'll take your word for it regarding the performance.

assaf758:
So just clarifying in pr #5468 the default chunk size was set to 10mb because it has optimal performance…
Why use 500KB chunks in the test? From your tests, it seems it would be much slower than the 10mb optimum…

PR #5469, opened 2024-04-29
[RemoteWorkflows] Fail fast when remote runner is failing [1.6.x]

liranbg:
When running a remote workflow, the configuration might be incorrect causing the remote runner (the pod that spins the remote workflow) to fail before it spins the workflow at all.
This would lead to needlessly wait for the workflow to come up the entire timeout (300 seconds by default)

This PR introduces a fail-fast mechanism for remote workflows where it detects the remote runner status and once it fails, it returns 412 to client (a-la SDK) which stops immediately waiting for the remote workflow to be finished

https://iguazio.atlassian.net/browse/ML-6188

### DIFF
--- a/mlrun/model.py
+++ b/mlrun/model.py
@@ -624,6 +624,11 @@ class RunMetadata(ModelObj):
     def iteration(self, iteration):
         self._iteration = iteration
 
+    def is_workflow_runner(self):
+        if not self.labels:
+            return False
+        return self.labels.get("job-type", "") == "workflow-runner"
+
 
 class HyperParamStrategies:
     grid = "grid"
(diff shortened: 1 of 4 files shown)


mlrun/model.py
```diff
@@ -1068,6 +1073,14 @@ def __init__(
         self.reason = reason
         self.notifications = notifications or {}
 
+    def is_failed(self):
+        if not self.state:
+            return None
```
TomerShor: Add a docstring / type hints that specify that the possible return values are Bool or None

TomerShor (approved):
Approved with a really minor comment

mlrun/model.py
liranbg: you got it

PR #5474, opened 2024-04-30
[Model Monitoring] Improvements to mm-apps do_tracking API

davesh0812:
Implements - [ML-6259](https://iguazio.atlassian.net/browse/ML-6259).

1. Implements `ModelMonitoringApplicationBaseV2` : 
    a.  with new `do_tracing` signature.
    b. delete `self.context` and uses the `monitoring_context` param as mlrun context.
2. Implements `MonitoringApplicationContext` - The monitoring context holds all the relevant information for the monitoring application, and also it can be used for logging artifacts and results.
3.  Implements `EvidentlyModelMonitoringApplicationBaseV2`.
4. Add `__PrepareMonitoringEvent` step the application serving `graph` - preparing the application event for the application step.
5. Update `_PushToWriter` step.
6. Update `HistogramDataDriftApplication` to v2.
7. Change `ModelMonitoringWriter` and `ModelMonitoringWriter._reconstruct_event` to work with 2 kinds of events - `metric` and `result`, for now the `ModelMonitoringWriter` writes only the results to the databases.
8. Change the applications (regular and evidently) and the result object directories. 

[ML-6259]: https://iguazio.atlassian.net/browse/ML-6259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/model_monitoring/applications/application_steps.py
```diff
@@ -0,0 +1,158 @@
+# Copyright 2024 Iguazio
```
jond01: This module should be private - prepend the file name with `_`

mlrun/model_monitoring/applications/base.py
```diff
@@ -0,0 +1,284 @@
...
+                    )
+                ]
+
+
+        # mlrun: end-code
+    """
+
+    kind = "monitoring_application"
+
+    def do(
```
jond01, line 57: Since this method is essentially private (we do not expect user overrides) but it has to be named `do` - I suggest renaming it to `_do` and adding in the class `do = _do`.
I think that it will hide this method in the API reference, which also needs to be updated: `docs/api/mlrun.model_monitoring.rst`

mlrun/model_monitoring/applications/histogram_data_drift.py
```diff
@@ -264,45 +270,54 @@ def _log_drift_artifacts(
...
         )
-        self.context.logger.debug("Computing average per metric")
-        results = self._get_results(metrics_per_feature)
-        self.context.logger.debug("Finished running the application", results=results)
-        return results
+        monitoring_context.logger.debug("Computing average per metric")
+        results_and_metrics = self._get_metrics(metrics_per_feature)
+        self._add_general_drift_result(
+            metrics=results_and_metrics,
+        )
```
jond01: 1. I think that `metrics_and_result` is a more accurate name
2. You actually cast the list type here. A decent static-type analyzer should emit a warning here. The solution is to declare the return type of `_get_metrics` as a union.

mlrun/model_monitoring/applications/results.py
```diff
@@ -0,0 +1,97 @@
...
+    kind: mm_constant.ResultKindApp
+    status: mm_constant.ResultStatusApp
+    extra_data: dict = dataclasses.field(default_factory=dict)
+
+    def __post_init__(self):
+        pat = re.compile(r"[a-zA-Z_][a-zA-Z0-9_]*")
+        if not re.fullmatch(pat, self.name):
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "Attribute name must be of the format [a-zA-Z_][a-zA-Z0-9_]*"
+            )
```
jond01: 1. This function is duplicated.
2. How about result inheriting from metric:
   ```py
   class ModelMonitoringApplicationResult(ModelMonitoringApplicationMetric):
       pass
   ```
   I believe it's correct to say that a result is a sub-type of a metric. It also solves natural duplications, and simplifies the type annotations.

tests/model_monitoring/test_applications/test_histogram_data_drift.py
```diff
@@ -180,27 +192,34 @@ def test(
...
+            ):
+                metrics.append(res)
+        assert len(metrics) == 3, "Expected three metrics"
 
 
 @pytest.mark.parametrize(
     ("sample_df_stats", "feature_stats"),
     [
-        pytest.param(pd.DataFrame(), pd.DataFrame(), id="empty-dfs"),
+        # pytest.param(pd.DataFrame(), pd.DataFrame(), id="empty-dfs"),
```
jond01: Why?

tests/model_monitoring/test_application_steps.py
```diff
@@ -0,0 +1,125 @@
...
+from mlrun.model_monitoring.applications import (
+    ModelMonitoringApplicationMetric,
+    ModelMonitoringApplicationResult,
+)
+from mlrun.model_monitoring.applications.application_steps import (
+    _PushToMonitoringWriter,
+)
+from mlrun.utils import Logger
+
+STREAM_PATH = "./test_stream.json"
```
jond01, line 35: Please do not add this empty resource file. Use pytest's temp file fixture utility instead.

tests/system/model_monitoring/assets/application.py
```diff
@@ -71,5 +73,38 @@ def do_tracking(
         ]
 
 
+class DemoMonitoringAppV2(ModelMonitoringApplicationBaseV2):
```
jond01, line 76: Why not replace the "V1"?

tests/system/model_monitoring/test_app.py
```diff
@@ -212,6 +223,21 @@ def custom_setup_class(cls) -> None:
                     "with_training_set": True,
                 },
                 results={"data_drift_test"},
+            ),  # TODO : delete in 1.9.0
+            _AppData(
+                class_=DemoMonitoringAppV2,
+                rel_path="assets/application.py",
+                results={"data_drift_test", "model_perf"},
+            ),
```
jond01: Don't you think we have too many apps for this test with V1 + V2?

jond01 (commented):
Nice work!

I had a few comments about the docs, tests, and implementation.

tests/system/model_monitoring/assets/application.py
davesh0812: We are still need to support V1 until 1.9.0

tests/system/model_monitoring/test_app.py
davesh0812: Because we still need to support V1 until 1.9.0 I think we should test V1 as well.. what is your suggestion? To split it to V1 and V2 tests?

jond01: Ok, testing also V1 is a valid point.
I think you can make this class slightly more abstract and split V1 from V2, without duplicating the class

mlrun/model_monitoring/applications/application_steps.py
davesh0812: It not enough that the classes are prepend  with `_`?

jond01: It's great :)
But if this module is entirely internal and will not include parts that users should interact with, it's safer to make it private, in addition to the classes.

mlrun/model_monitoring/applications/base.py
davesh0812: I can also delete the do at all by adding in the class `do = do_tracking` sounds better? because the `do` function isn't doing something important now...

davesh0812: lets use the `:exclude-members: do`  in the docs for this

mlrun/model_monitoring/applications/results.py
davesh0812: I am not sure I agree on that, for now I just created an abstract class they are both inheriting from.

mlrun/model_monitoring/applications/_application_steps.py
```diff
@@ -0,0 +1,158 @@
...
+                writer_event[mm_constant.WriterEvent.DATA] = json.dumps(data)
+            else:
+                writer_event[mm_constant.WriterEvent.EVENT_KIND] = (
+                    mm_constant.WriterEventKind.METRIC
+                )
+                writer_event[mm_constant.WriterEvent.DATA] = json.dumps(data)
+
+            writer_event[mm_constant.WriterEvent.EVENT_KIND] = (
+                mm_constant.WriterEventKind.RESULT
+                if type(result) == ModelMonitoringApplicationResult
```
Eyal-Danieli: best practice to check types is by using `isinstance()`:
```suggestion
                if isinstance(result, ModelMonitoringApplicationResult):
```

mlrun/model_monitoring/applications/base.py
```diff
@@ -0,0 +1,273 @@
...
+        pd.Timestamp,
+        pd.Timestamp,
+        str,
+        str,
+    ]:
+        """
+        Converting the event into a single tuple that will be used for passing the event arguments to the running
+        application
+
+        :param event: dictionary with all the incoming data
```
Eyal-Danieli: update the docstring param
```suggestion
        :param monitoring_context: ...
```

mlrun/model_monitoring/applications/base.py
```diff
@@ -0,0 +1,273 @@
...
+                mm_results.ModelMonitoringApplicationResult,
+                mm_results.ModelMonitoringApplicationMetric,
+            ]
+        ],
+        mm_context.MonitoringApplicationContext,
+    ]:
+        """
+        Process the monitoring event and return application results & metrics.
+        :param monitoring_context:   (MonitoringApplicationContext) The monitoring application context.
+        :returns:                    (list[ModelMonitoringApplicationResult], dict) The application results
```
Eyal-Danieli: just a recommendation (not mandatory) - when the return value is a tuple it is nicer to explain in with the following format: 
```suggestion
        :returns:  A tuple of:
             [0] = list of application results that can be either from type `ModelMonitoringApplicationResult` or from type `ModelMonitoringApplicationResult`.
             [1] = the original application event, wrapped in `MonitoringApplicationContext` object
```

Eyal-Danieli: + note that in your docstrings the return value is` list[ModelMonitoringApplicationResult], dict` while in the function signature is actually 
```
tuple[
        list[
            Union[
                mm_results.ModelMonitoringApplicationResult,
                mm_results.ModelMonitoringApplicationMetric,
            ]
        ],
        mm_context.MonitoringApplicationContext,
    ]:
```

mlrun/model_monitoring/applications/evidently_base.py
```diff
@@ -0,0 +1,211 @@
...
+        # TODO : more then one project (mep -> project)
+        if not _HAS_EVIDENTLY:
+            raise ModuleNotFoundError("Evidently is not installed - the app cannot run")
+        self.evidently_workspace = Workspace.create(evidently_workspace_path)
+        self.evidently_project_id = evidently_project_id
+        self.evidently_project = self.evidently_workspace.get_project(
+            evidently_project_id
+        )
+
+    def log_evidently_object(
```
Eyal-Danieli, line 162: Can be staticmethod :) 
```suggestion
    @staticmethod
    def log_evidently_object(
```

mlrun/model_monitoring/applications/histogram_data_drift.py
```diff
@@ -264,45 +270,59 @@ def _log_drift_artifacts(
...
-        self.context.logger.debug("Computing average per metric")
-        results = self._get_results(metrics_per_feature)
-        self.context.logger.debug("Finished running the application", results=results)
-        return results
+        monitoring_context.logger.debug("Computing average per metric")
+        metrics_and_result: list[
+            typing.Union[
+                mm_results.ModelMonitoringApplicationMetric,
+                mm_results.ModelMonitoringApplicationResult,
+            ]
```
Eyal-Danieli, line 315: `_get_metrics` returns only a list of `ModelMonitoringApplicationMetric`. therefore the typehint here should be:
`list[mm_results.ModelMonitoringApplicationMetric]`

Eyal-Danieli: oh I see that in the next line you are adding mm_results.ModelMonitoringApplicationResult as well. Just a recommendation: maybe wrap these 2 operations in a different function `_add_metrics_and_general_drift` and then you can keep the same typehints: 
```
metrics_and_result: list[
            typing.Union[
                mm_results.ModelMonitoringApplicationMetric,
                mm_results.ModelMonitoringApplicationResult,
            ] = self._add_metrics_and_general_drift(metrics_per_feature)

```

mlrun/model_monitoring/applications/results.py
```diff
@@ -0,0 +1,99 @@
...
+            mm_constant.ResultData.RESULT_KIND: self.kind.value,
+            mm_constant.ResultData.RESULT_STATUS: self.status.value,
+            mm_constant.ResultData.RESULT_EXTRA_DATA: json.dumps(self.extra_data),
+        }
+
+
+@dataclasses.dataclass
+class ModelMonitoringApplicationMetric(_ModelMonitoringApplicationDataRes):
+    """
+    Class representing the result of a custom model monitoring application.
```
Eyal-Danieli:
```suggestion
    Class representing a single metric of a custom model monitoring application.
```

Eyal-Danieli (commented):
LGTM.
Few comments and suggestions, mostly docstrings and typehints :)

mlrun/model_monitoring/applications/histogram_data_drift.py
davesh0812: I think I will leave it like it is now, it is the only way the typing isn't yeling on me.

Eyal-Danieli (approved):

PR #5478, opened 2024-05-01
[Datastore] When necessary, save the bucket as part of the data source profile and use the default path if the path is not specified

alxtkr77:
[ML-6326](https://iguazio.atlassian.net/browse/ML-6326)

[ML-6326]: https://iguazio.atlassian.net/browse/ML-6326?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/store/datastore.md
+++ b/docs/store/datastore.md
@@ -179,7 +179,8 @@ The equivalent to this parameter in environment authentication is "AZURE_STORAGE
 Credential authentication:
 - `credential` &mdash; TokenCredential or SAS token. The credentials with which to authenticate.
 This variable is sensitive information and is kept confidential.
-- `bucket` &mdash; A string representing the bucket. When specified, it is automatically appended to the object path, and thus, it should not be manually included in the target path by the user. 
+- `bucket` &mdash; A string representing the bucket. When specified, it is automatically prepended to the object path, and thus, it should not be manually included in the target path by the user.
+This parameter will become mandatory starting with version 1.9.
 
 ## Databricks file system 
 ### DBFS credentials and parameters
@@ -230,7 +231,8 @@ The equivalent to this parameter in environment authentication is "GOOGLE_APPLIC
 - `gcp_credentials` &mdash; A JSON in a string format representing the authentication parameters required by GCS API. 
 For privacy reasons, it's tagged as a private attribute, and its default value is `None`.
 The equivalent to this parameter in environment authentication is "GCP_CREDENTIALS".
-- `bucket` &mdash; A string representing the bucket. When specified, it is automatically appended to the object path, and thus, it should not be manually included in the target path by the user. 
+- `bucket` &mdash; A string representing the bucket. When specified, it is automatically prepended to the object path, and thus, it should not be manually included in the target path by the user. 
+This parameter will become mandatory starting with version 1.9.
 
 The code prioritizes `gcp_credentials` over `credentials_path`.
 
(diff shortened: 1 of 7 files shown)


docs/store/datastore.md
```diff
@@ -179,6 +179,7 @@ The equivalent to this parameter in environment authentication is "AZURE_STORAGE
 Credential authentication:
 - `credential` &mdash; TokenCredential or SAS token. The credentials with which to authenticate.
 This variable is sensitive information and is kept confidential.
+- `bucket` &mdash; A string representing the bucket. When specified, it is automatically appended to the object path, and thus, it should not be manually included in the target path by the user.
```
gtopper:
```suggestion
- `bucket` &mdash; A string representing the bucket. When specified, it is automatically prepended to the object path, and thus, it should not be manually included in the target path by the user. 
```
(all instances)

mlrun/datastore/datastore_profile.py
```diff
@@ -185,6 +185,17 @@ class DatastoreProfileS3(DatastoreProfile):
     assume_role_arn: typing.Optional[str] = None
     access_key_id: typing.Optional[str] = None
     secret_key: typing.Optional[str] = None
+    bucket: typing.Optional[str] = None
+
+    @pydantic.validator("bucket")
+    def check_bucket(cls, v):
+        if v is None:
```
gtopper:
```suggestion
        if not v:
```
I'm assuming an empty string won't cut it either.
(all instances)

mlrun/datastore/datastore_profile.py
```diff
@@ -283,7 +308,10 @@ def url(self, subpath) -> str:
         if subpath.startswith("/"):
             #  in gcs the path after schema is starts with bucket, wherefore it should not start with "/".
             subpath = subpath[1:]
-        return f"gcs://{subpath}"
+        if self.bucket:
+            return f"gcs://{self.bucket}/{subpath}"
+        else:
+            return f"gcs://{subpath}"
```
gtopper, line 319: Looks like this is an existing issue, but there's something wrong with the subpath being assumed to not start with a slash here, but the opposite assumption is made in S3.

mlrun/datastore/targets.py
```diff
@@ -656,6 +656,41 @@ def get_target_templated_path(self):
...
+                    _get_target_path(
+                        self,
+                        self._resource,
+                        self.run_id is not None,
+                        netloc=parsed_url.netloc,
+                        scheme=parsed_url.scheme,
+                    ),
+                    self.run_id,
+                    is_single_file,
+                )
```
gtopper: Looks like a coding accident.
```suggestion
```

tests/system/datastore/test_aws_s3.py
```diff
@@ -140,3 +152,40 @@ def test_ingest_with_parquet_source(self, use_datastore_profile):
...
+        df = pd.DataFrame(data)
+        source_path = param["parquet_url"]
+        with tempfile.NamedTemporaryFile(mode="w+", delete=True) as temp_file:
+            df.to_parquet(temp_file.name)
+            path_only = self.s3["s3"]["parquet_url"]
+            s3_fs.put_file(temp_file.name, path_only)
+
+        parquet_source = ParquetSource(name="test", path=source_path)
+
+        # ingest
```
gtopper: Ingest is several lines down, and doesn't really need to be announced.
```suggestion
```

tests/system/datastore/test_aws_s3.py
```diff
@@ -140,3 +152,40 @@ def test_ingest_with_parquet_source(self, use_datastore_profile):
...
+        expected_default_ds_data_prefix = get_default_prefix_for_target(
+            "dsnosql"
+        ).format(
+            ds_profile_name="s3ds_profile_with_bucket",
+            project=fset.metadata.project,
+            kind=targets[0].kind,
+            name=fset.metadata.name,
+        )
+
+        assert fset.get_target_path().startswith(expected_default_ds_data_prefix)
```
gtopper, line 189: Shouldn't we check that it actually wrote the data also?

docs/store/datastore.md
gtopper: Also, need to add the fact that the user needs to specify this, and will be required to do so from 1.9.0 onward.

tests/system/datastore/test_aws_s3.py
```diff
@@ -140,3 +152,40 @@ def test_ingest_with_parquet_source(self, use_datastore_profile):
         assert_frame_equal(
             df.sort_index(axis=1), result.sort_index(axis=1), check_like=True
         )
+
+    def test_ingest_ds_default_target(self):
+        #  create source
```
gtopper: The source is created 14 lines below, and doesn't really need to be announced.
```suggestion
```

mlrun/datastore/datastore_profile.py
alxtkr77: Like you said - its a separate issue.

gtopper: I guess at least add a TODO?

mlrun/datastore/datastore_profile.py
```diff
@@ -311,12 +339,26 @@ class DatastoreProfileAzureBlob(DatastoreProfile):
...
     sas_token: typing.Optional[str] = None
     credential: typing.Optional[str] = None
+    bucket: typing.Optional[str] = None
+
+    @pydantic.validator("bucket")
+    def check_bucket(cls, v):
+        if not v:
+            warnings.warn(
+                "The 'bucket' attribute will be mandatory starting from version 1.9",
+                DeprecationWarning,
```
gtopper: It turns out that we use `FutureWarning` everywhere else in the code base, so we should probably change this (all instances in this file).
```suggestion
                FutureWarning,
```

mlrun/datastore/datastore_profile.py
gtopper: I guess in S3 though.

gtopper (approved):

PR #5479, opened 2024-05-01
[Alerts] Add templates functionality

yanburman:
implements:
https://iguazio.atlassian.net/browse/ML-6175
https://iguazio.atlassian.net/browse/ML-6032
https://iguazio.atlassian.net/browse/ML-6031

Create/delete are provided, but not exposed to the user via SDK yet


server/api/initial_data.py
```diff
@@ -821,6 +824,49 @@ def _delete_state_file():
...
+        ),
+        mlrun.common.schemas.AlertTemplate(
+            template_name="DriftSuspected",
+            template_description="Generic template for drift suspected alerts",
+            system_generated=True,
+            description="A model drift is suspected",
+            severity=mlrun.common.schemas.alert.AlertSeverity.MEDIUM,
+            trigger={"events": [mlrun.common.schemas.alert.EventKind.DRIFT_SUSPECTED]},
+            reset_policy=mlrun.common.schemas.alert.ResetPolicy.MANUAL,
+        ),
```
quaark: I'm not sure I like using the data migration mechanism in order to ensure predefined data in the DB. That isn't what data migrations is meant to do. This means in order to change them in the future (if needed) you'll need to add more data migrations and this will become messy fast.
Wondering if maybe worth saving these predefined templates as part of mlrun config (Good for customizing them easily) and when api initializes in `initial_data` but not as part of a "migration" we can check the DB and add the default templates to the table overriding them (only override the predefined templates)

quaark (changes requested):
Looks good in general, but one thing bothers me with creating the predefined templates

server/api/initial_data.py
theSaarco: @quaark - I half-agree - I certainly agree the default templates should indeed be in the configuration.
These should be added to the DB as part of the `update_default_configuration_data` method (which resides in `initial_data.py`) - this is the same thing we do for the common hub source.
Note that `update_default_configuration_data` is actually not called as part of the data-migration, but rather called from `move_api_to_online`, so I actually think that's pretty much aligned with what you're looking for.

mlrun/alerts/alert.py
```diff
@@ -0,0 +1,104 @@
...
+            self.entity.dict() if not isinstance(self.entity, dict) else self.entity
+        )
+        data["notifications"] = [
+            notification.dict() if not isinstance(notification, dict) else notification
+            for notification in self.notifications
+        ]
+        data["trigger"] = (
+            self.trigger.dict() if not isinstance(self.trigger, dict) else self.trigger
+        )
+        return data
```
theSaarco, line 88: IIUC, this is done so you can use (and serialize) pydantic-based objects (entity, notification, trigger) within a client-side `ModelObj`-based class. Question is - why not create client-side schemas for these as well? 
Also, if you do this, you'll need to implement `from_dict` as well using the same technique - otherwise when de-serializing the object all you get are the fields in `_dict_fields`.

mlrun/alerts/alert.py
```diff
@@ -0,0 +1,104 @@
...
+        "project",
+        "name",
+        "description",
+        "summary",
+        "severity",
+        "criteria",
+        "reset_policy",
+    ]
+
+    def __init__(
```
theSaarco, line 35: Please add type hints. In many cases these are simply plain strings, but there are many fields that are classes. We should help the user with this.

mlrun/alerts/alert.py
```diff
@@ -0,0 +1,104 @@
...
+        name=None,
+        description=None,
+        summary=None,
+        severity=None,
+        trigger=None,
+        criteria=None,
+        reset_policy=None,
+        notifications=None,
+        entity=None,
+        template=None,
```
theSaarco: Does the `template` have to be a template object? User should be able to generate the `AlertConfig` using a template name as well (in which case we go and retrieve the template from the DB).
Also, please move the `template` to be right after the mandatory params (`project` + `name`).

mlrun/alerts/alert.py
```diff
@@ -0,0 +1,104 @@
...
+        "summary",
+        "severity",
+        "criteria",
+        "reset_policy",
+    ]
+
+    def __init__(
+        self,
+        project=None,
+        name=None,
```
theSaarco: Both `project` and `name` should be mandatory.

mlrun/alerts/alert.py
```diff
@@ -0,0 +1,104 @@
...
+            self.notifications.append(notification)
+        return self
+
+    def with_entity(self, entity: EventEntity):
+        if not isinstance(entity, EventEntity):
+            raise ValueError("entity parameter must be of type: EventEntity")
+        self.entity = entity
+        return self
+
+    def apply_template(self, template):
```
theSaarco: This should be a private function, unless we want to enable the user to call `apply_template` on an already existing alert-config, which I think is not needed and just confusing.

mlrun/db/httpdb.py
```diff
@@ -41,6 +42,7 @@
 from mlrun.db.auth_utils import OAuthClientIDTokenProvider, StaticTokenProvider
 from mlrun.errors import MLRunInvalidArgumentError, err_to_str
 
+from ..alerts.alert import AlertConfig
```
theSaarco: Channeling @alonmr 😄 - don't use relative imports.

mlrun/db/httpdb.py
```diff
@@ -3943,7 +3945,7 @@ def generate_event(
     def store_alert_config(
         self,
         alert_name: str,
-        alert_data: Union[dict, mlrun.common.schemas.AlertConfig],
+        alert_data: Union[dict, AlertConfig],
         project="",
     ):
```
theSaarco:
```suggestion
    ) -> AlertConfig:
```

mlrun/db/httpdb.py
```diff
@@ -3956,11 +3958,21 @@ def store_alert_config(
         project = project or config.default_project
         endpoint_path = f"projects/{project}/alerts/{alert_name}"
         error_message = f"put alert {project}/alerts/{alert_name}"
-        if isinstance(alert_data, mlrun.common.schemas.AlertConfig):
-            alert_data = alert_data.dict()
+        if isinstance(alert_data, AlertConfig):
+            alert_data = alert_data.to_dict()
+        self._validate_alert(alert_data)
```
theSaarco: Won't this be done anyway in the server side as part of the pydantic de-serialization? It's a bit faster to check it here rather than round-trip to the server, but since this is only useful if the user provided a dict then I'm not sure it's really needed.

mlrun/projects/project.py
```diff
@@ -3899,7 +3899,9 @@ def store_alert_config(self, alert_data: AlertConfig, alert_name=None):
         db = mlrun.db.get_run_db(secrets=self._secrets)
         if alert_name is None:
             alert_name = alert_data.name
-        return db.store_alert_config(alert_name, alert_data.dict(), self.metadata.name)
+        return db.store_alert_config(
+            alert_name, alert_data.to_dict(), self.metadata.name
```
```suggestion
            alert_name, alert_data, project=self.metadata.name
```
Don't really need the `to_dict()` here now, right?
Also, since you're passing 2 names here it would be helpful to explicitly state which parameter is what (at least for the project case).

mlrun/projects/project.py
```diff
@@ -3956,6 +3958,23 @@ def reset_alert_config(
             alert_name = alert_data.name
         db.reset_alert_config(alert_name, self.metadata.name)
 
+    def get_alert_template(self, template_name: str):
```
```suggestion
    def get_alert_template(self, template_name: str) -> AlertTemplate:
```

mlrun/projects/project.py
```diff
@@ -3956,6 +3958,23 @@ def reset_alert_config(
...
+        """
+        Retrieve a specific alert template.
+        :param template_name: The name of the template to retrieve.
+        :return: The template object.
+        """
+        db = mlrun.db.get_run_db(secrets=self._secrets)
+        return db.get_alert_template(template_name)
+
+    def list_alert_templates(self):
```
```suggestion
    def list_alert_templates(self) -> list[AlertTemplate]:
```

mlrun/template_config.py
```diff
@@ -0,0 +1,46 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import mlrun.common.schemas
+
```
theSaarco: Please add a comment explaining where these are used and when it makes sense to change them.

server/api/api/api.py
```diff
@@ -192,6 +193,11 @@
     tags=["artifacts"],
     dependencies=[Depends(deps.authenticate_request)],
 )
+api_v2_router.include_router(
+    alert_template.router,
+    tags=["alert-templates"],
+    dependencies=[Depends(deps.authenticate_request)],
+)
```
theSaarco: Why does this need to be `v2`? 
The logic is that we still add endpoints to `v1` as long as we don't introduce a new version to an existing endpoint. Since these are brand-new endpoints, there's no real reason to use `v2` here.

server/api/crud/alert_template.py
```diff
@@ -0,0 +1,96 @@
...
+
+import sqlalchemy.orm
+
+import mlrun.utils.singleton
+import server.api.api.utils
+import server.api.utils.helpers
+import server.api.utils.singletons.db
+
+
+class AlertTemplate(
```
theSaarco:
```suggestion
class AlertTemplates(
```

CRUDs are named in plural when dealing with a specific object-type. Exceptions are usually when we deal with functionality (such as `FeatureStore`) that has multiple objects.

server/api/crud/alert_template.py
```diff
@@ -0,0 +1,96 @@
...
+        template = server.api.utils.singletons.db.get_db().get_alert_template(
+            session, name
+        )
+
+        if template is None:
+            return
+
+        if template.system_generated:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"Alert template {name} can not be deleted as it is a system template"
```
theSaarco: Please ask @jillnogold to take a look at the relevant strings as these will be user-visible. This is part of our emerging process for handling system messages.

server/api/db/sqldb/db.py
```diff
@@ -4214,6 +4215,42 @@ def create_data_version(self, session, version):
...
+    def list_alert_templates(self, session) -> list[mlrun.common.schemas.AlertTemplate]:
+        query = self._query(session, AlertTemplate)
+        return list(map(self._transform_alert_template_record_to_schema, query.all()))
+
+    def get_alert_template(
+        self, session, name: str
+    ) -> mlrun.common.schemas.AlertTemplate:
+        return self._transform_alert_template_record_to_schema(
+            self._get_alert_template_record(session, name)
+        )
```
theSaarco, line 4252: This will throw an ugly error that is not clear to the user in case `_get_alert_template_record` returns `None`. Suggest to separate it - first `_get`, and explode with a meaningful `mlrun.errors.MLRunNotFoundError` error if not found, then package and return.

server/api/db/sqldb/models/models_mysql.py
```diff
@@ -702,6 +702,29 @@ def full_object(self):
...
+    class AlertTemplate(Base, mlrun.utils.db.BaseModel):
+        __tablename__ = "alert_templates"
+        __table_args__ = (UniqueConstraint("name", name="_alert_templates_uc"),)
+
+        id = Column(Integer, primary_key=True)
+        name = Column(
+            String(255, collation=SQLCollationUtil.collation()), nullable=False
+        )
+
+        _full_object = Column("object", JSON)
```
theSaarco, line 714: Don't know if you want to be the one volunteering for this, but I think it's about time to implement a base-class (or a mixin, for python lovers) for classes that have `JSON` body, like the `mlrun.utils.db.HasStruct` class. This line along with the `full_object` property definition are simple boilerplate code that just gets repeated over and over.

server/api/initial_data.py
```diff
@@ -821,6 +825,21 @@ def _delete_state_file():
...
+
+def _migrate_alert_templates_table(
+    db: server.api.db.sqldb.db.SQLDB, db_session: sqlalchemy.orm.Session
+):
+    for template in mlrun.template_config.pre_defined_templates:
+        record = db.get_alert_template(db_session, template.template_name)
+        if record is None:
+            db.store_alert_template(db_session, template)
+
+
```
theSaarco: As @quaark noted correctly, this should not happen in a data-migration context. See my comment in reply to his for where this should be.

theSaarco (changes requested):
Overall looks good - please see comments.
A related question - are you going to introduce the change to multiple entities in the alert config here, or will this be in a follow-up PR?

server/api/crud/alert_template.py
jillnogold: f"Cannot delete the Alert template {name}: it is a system template"

server/api/db/sqldb/db.py
yanburman: _transform_alert_template_record_to_schema() handles None, so it will return None

server/api/db/sqldb/models/models_mysql.py
yanburman: Will look into it, but not part of this PR

mlrun/common/schemas/alert.py
```diff
@@ -120,3 +120,25 @@ class AlertConfig(pydantic.BaseModel):
...
+class AlertTemplate(
+    pydantic.BaseModel
+):  # Template fields that are not shared with created configs
+    template_id: int = None
+    template_name: str
+    template_description: Optional[str] = (
+        "String explaining the purpose of this template"
+    )
+
+    # A property identifying templates that were created by the system and cannot be modified/deleted by the user
```
jillnogold: # A property that identifies templates that were created by the system and cannot be modified/deleted by the user

mlrun/common/schemas/alert.py
```diff
@@ -120,3 +120,25 @@ class AlertConfig(pydantic.BaseModel):
...
+    system_generated: bool = False
+
+    # AlertConfig fields that are pre-defined
+    description: Optional[str] = (
+        "String to be sent in the notifications generated e.g. 'Model {{ $project }}/{{ $entity }} is drifting.'"
```
jillnogold: "String to be sent in the generated notifications e.g. 'Model {{ $project }}/{{ $entity }} is drifting.'"

mlrun/db/httpdb.py
```diff
@@ -4012,6 +4024,32 @@ def reset_alert_config(self, alert_name: str, project=""):
...
+        """
+        endpoint_path = f"alert-templates/{template_name}"
+        error_message = f"get template alert-templates/{template_name}"
+        response = self.api_call("GET", endpoint_path, error_message, version="v2")
+        return mlrun.common.schemas.AlertTemplate(**response.json())
+
+    def list_alert_templates(self):
+        """
+        Retrieve list of all alert templates.
+        :return: All the alert templates objects in the database.
```
jillnogold: :return: All the alert template objects in the database.

mlrun/projects/project.py
```diff
@@ -3956,6 +3958,23 @@ def reset_alert_config(
...
+        """
+        Retrieve list of all alert templates.
+        :return: All the alert templates objects in the database.
```
jillnogold: :return: All the alert template objects in the database.

mlrun/template_config.py
```diff
@@ -0,0 +1,46 @@
...
+        description="A job has failed",
+        severity=mlrun.common.schemas.alert.AlertSeverity.MEDIUM,
+        trigger={"events": [mlrun.common.schemas.alert.EventKind.FAILED]},
+        reset_policy=mlrun.common.schemas.alert.ResetPolicy.MANUAL,
+    ),
+    mlrun.common.schemas.AlertTemplate(
+        template_name="DriftDetected",
+        template_description="Generic template for drift detected alerts",
+        system_generated=True,
+        description="A model drift has been detected",
```
jillnogold: description="Model drift has been detected",

mlrun/template_config.py
```diff
@@ -0,0 +1,46 @@
...
+        severity=mlrun.common.schemas.alert.AlertSeverity.HIGH,
+        trigger={"events": [mlrun.common.schemas.alert.EventKind.DRIFT_DETECTED]},
+        reset_policy=mlrun.common.schemas.alert.ResetPolicy.MANUAL,
+    ),
+    mlrun.common.schemas.AlertTemplate(
+        template_name="DriftSuspected",
+        template_description="Generic template for drift suspected alerts",
+        system_generated=True,
+        description="A model drift is suspected",
```
jillnogold: description="Model drift is suspected",

server/api/api/api.py
yanburman: you defined it to be v2: https://iguazio.atlassian.net/wiki/spaces/ARC/pages/1867801/SDK+and+alert+usage

tests/integration/sdk_api/alerts/test_alerts.py
```diff
@@ -176,6 +176,82 @@ def test_alert_after_project_deletion(self):
...
+        project_name = "my-project"
+        project = mlrun.new_project(project_name)
+
+        # one of the pre-defined system templates
+        drift_system_template = mlrun.common.schemas.AlertTemplate(
+            template_id=2,
+            template_name="DriftDetected",
+            template_description="Generic template for drift detected alerts",
+            system_generated=True,
+            description="A model drift has been detected",
```
jillnogold: description="Model drift has been detected",

server/api/db/sqldb/db.py
theSaarco: Yes, I was looking at `_transform_alert_template_schema_to_record` which doesn't do this check. 
Usually we raise already in the db code, but I do see that you check in the endpoint code and raise if `None` is returned, so that works.

server/api/api/api.py
theSaarco: I wonder what my 6-month ago self was thinking 😰

mlrun/db/httpdb.py
```diff
@@ -3956,13 +3957,13 @@ def store_alert_config(
...
         endpoint_path = f"projects/{project}/alerts/{alert_name}"
         error_message = f"put alert {project}/alerts/{alert_name}"
-        if isinstance(alert_data, mlrun.common.schemas.AlertConfig):
-            alert_data = alert_data.dict()
+        if isinstance(alert_data, AlertConfig):
+            alert_data = alert_data.to_dict()
         body = _as_json(alert_data)
         response = self.api_call("PUT", endpoint_path, error_message, body=body)
-        return mlrun.common.schemas.AlertConfig(**response.json())
+        return AlertConfig(**response.json())
```
theSaarco: For `ModelObj` classes it's better to use `from_dict` here.

mlrun/db/httpdb.py
```diff
@@ -3973,9 +3974,9 @@ def get_alert_config(self, alert_name: str, project=""):
         endpoint_path = f"projects/{project}/alerts/{alert_name}"
         error_message = f"get alert {project}/alerts/{alert_name}"
         response = self.api_call("GET", endpoint_path, error_message)
-        return mlrun.common.schemas.AlertConfig(**response.json())
+        return AlertConfig(**response.json())
```
theSaarco: Same here, and in every other method.

server/api/initial_data.py
```diff
@@ -821,6 +823,26 @@ def _delete_state_file():
...
+def _alert_templates_differ(template1, template2):
+    return (
+        template1.template_description != template2.template_description
+        or template1.description != template2.description
+        or template1.severity != template2.severity
+        or template1.trigger != template2.trigger
+        or template1.reset_policy != template2.reset_policy
+        or template1.criteria != template2.criteria
+    )
```
theSaarco: Since you added `__eq__` to sub-classes of the template (probably so you can do this comparison), why not go all the way and put this code as `__eq__` in `AlertTemplate` and here just do `record != template`?

tests/integration/sdk_api/alerts/test_alerts.py
```diff
@@ -176,6 +176,82 @@ def test_alert_after_project_deletion(self):
...
+        drift_system_template = mlrun.common.schemas.AlertTemplate(
+            template_id=2,
+            template_name="DriftDetected",
+            template_description="Generic template for drift detected alerts",
+            system_generated=True,
+            description="Model drift has been detected",
+            severity=alert_constants.AlertSeverity.HIGH,
+            trigger={"events": [mlrun.common.schemas.alert.EventKind.DRIFT_DETECTED]},
+            reset_policy=mlrun.common.schemas.alert.ResetPolicy.MANUAL,
+        )
```
theSaarco: Why not import the `pre_defined_templates` directly, and take the definitions from the source instead of re-defining it here? This way you can verify that you get what is defined, even if we modify the pre-defined templates at some point.

theSaarco (approved):
Looks good. Have some comments that I think it's better to fix prior to merging, but I won't block the PR on them.

server/api/initial_data.py
yanburman: because I didn't want to do extra comparisons that are not needed during mlrun-api startup

mlrun/template_config.py
```diff
@@ -0,0 +1,49 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import mlrun.common.schemas
+
+# These are alert templates that come built-in with the system and pre-populated on system start
+# If we define new system templates, those should be added here
+
+pre_defined_templates = [
```
quaark: When I suggested adding them to the config, I did mean the actual mlconf. As the templates can be represented by a simple json, we can allow the API to be configured with templates before deployment and not necessarily hard-coded.

Even if that isn't what we want, I'd prefer this be part of the Config class and not in a global `templates_config.py` file only for alert templates.

quaark (changes requested):
Other than Saar's comments, just one more thing regarding the template config

mlrun/alerts/alert.py
```diff
@@ -0,0 +1,133 @@
...
+            isinstance(item, Notification) for item in notifications
+        ):
+            raise ValueError("Notifications parameter must be a list of notifications")
+        for notification in notifications:
+            self.notifications.append(notification)
+        return self
+
+    def with_entity(self, entity: alert_constants.EventEntity):
+        if not isinstance(entity, alert_constants.EventEntity):
+            raise ValueError("entity parameter must be of type: EventEntity")
```
liranbg: capitalize the errors
```suggestion
            raise ValueError("Entity parameter must be of type: EventEntity")
```

mlrun/db/httpdb.py
```diff
@@ -4012,6 +4013,32 @@ def reset_alert_config(self, alert_name: str, project=""):
...
+        response = self.api_call("GET", endpoint_path, error_message)
+        return mlrun.common.schemas.AlertTemplate(**response.json())
+
+    def list_alert_templates(self) -> list[mlrun.common.schemas.AlertTemplate]:
+        """
+        Retrieve list of all alert templates.
+        :return: All the alert template objects in the database.
+        """
+        endpoint_path = "alert-templates"
+        error_message = "get alert templates"
```
liranbg: not consistent with https://github.com/mlrun/mlrun/pull/5479/files#diff-d9433d3acce3d91c2f6fc2779c332ac73b99675a1f056078fa17158e3b09bd5eR4025

mlrun/db/httpdb.py
liranbg: ☝🏼

server/api/api/endpoints/alert_template.py
```diff
@@ -0,0 +1,134 @@
...
+    if (
+        mlrun.mlconf.httpdb.clusterization.role
+        != mlrun.common.schemas.ClusterizationRole.chief
+    ):
+        chief_client = server.api.utils.clients.chief.Client()
+        return await chief_client.delete_alert_template(name=name, request=request)
+
+    await run_in_threadpool(
+        server.api.crud.AlertTemplates().delete_alert_template, db_session, name
+    )
```
liranbg, line 136: what about a log that you are about to delete?

quaark (approved):
LGTM

liranbg (changes requested):
minor suggestions

liranbg (approved):

PR #5484, opened 2024-05-01
[API] Add model endpoint list metrics/results endpoint

jond01:
Resolves [ML-6287](https://iguazio.atlassian.net/browse/ML-6287).

Currently, support only:
* results - metrics are not reported and saved yet,
* V3IO - to be extended to MySQL later.

The core functionality is unit-tested here and was checked e2e by me. For example:
image

[ML-6287]: https://iguazio.atlassian.net/browse/ML-6287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -703,3 +702,57 @@ def get_v3io_monitoring_apps_container(project_name: str) -> str:
...
+        """Get model monitoring results and metrics on the endpoint"""
+        metrics: list[
+            mlrun.common.schemas.model_monitoring.ModelEndpointMonitoringMetric
+        ] = []
+        container = self.get_v3io_monitoring_apps_container(self.project)
+        try:
+            response = self.client.kv.scan(container=container, table_path=endpoint_id)
+        except v3io.dataplane.response.HttpResponseError:
+            logger.info(
+                "Attempt getting metrics and results - no data. Check the "
```
gtopper: How do you know there's no data? It could be another error response.

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -703,3 +702,57 @@ def get_v3io_monitoring_apps_container(project_name: str) -> str:
...
+                "project name, endpoint, or wait for the applications to start.",
+                container=container,
+                table_path=endpoint_id,
+            )
+            return []
+
+        while response.output.items:
```
gtopper:
```suggestion
        while True:
```
As we discussed, you can't make this assumption.

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -703,3 +702,57 @@ def get_v3io_monitoring_apps_container(project_name: str) -> str:
...
+        try:
+            response = self.client.kv.scan(container=container, table_path=endpoint_id)
+        except v3io.dataplane.response.HttpResponseError:
+            logger.info(
+                "Attempt getting metrics and results - no data. Check the "
+                "project name, endpoint, or wait for the applications to start.",
+                container=container,
+                table_path=endpoint_id,
+            )
+            return []
```
gtopper: This has the strange effect that if the first call fails, we log an info message and return an empty list, but if subsequent calls fail, we raise an error.

mlrun/model_monitoring/writer.py
```diff
@@ -230,6 +232,8 @@ def do(self, event: _RawEvent) -> None:
         logger.info("Starting to write event", event=event)
         self._update_tsdb(event)
         self._update_kv_db(event)
+        logger.info("Completed event DB writes")
```
gtopper, line 235: Not sure what the expected traffic is, but logging anything on a per request basis is very verbose.

server/api/api/endpoints/model_monitoring.py
```diff
@@ -35,6 +35,29 @@ class _CommonParams:
     project: str
     auth_info: mlrun.common.schemas.AuthInfo
     db_session: Session
+    model_monitoring_access_key: Optional[str] = None
+
+    def __post_init__(self) -> None:
+        if not mlrun.mlconf.is_ce_mode():
+            # Generate V3IO Access Key
```
gtopper: Probably "generate" is the wrong word. Seems like it just gets it from the project secrets.

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -703,3 +702,57 @@ def get_v3io_monitoring_apps_container(project_name: str) -> str:
...
         return f"users/pipelines/{project_name}/monitoring-schedules/functions"
+
+    @staticmethod
+    def _extract_metrics_from_items(
+        items: list[dict[str, str]],
+    ) -> list[mlrun.common.schemas.model_monitoring.ModelEndpointMonitoringMetric]:
+        metrics: list[
+            mlrun.common.schemas.model_monitoring.ModelEndpointMonitoringMetric
+        ] = []
+        for app_item in items:
```
gtopper: I guess pick one name or the other?

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
jond01: IIRC we talked about the `.last` attribute.
I assume that if the items are empty, there is nothing left. Am I wrong?
I also tried to avoid `while True`, as it's a bit scary

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
jond01: Do you mean just `app` or `item`?

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
jond01: That's right.
However, if we pass that try-except block, then the table for sure exists, and the error is caused by a different source.

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
jond01: In theory - right, but I'm not sure there's a simple way to differentiate with just one error class.
This is the error I get:
```
HttpResponseError: Request failed with status 404: b'{\n\t"ErrorCode": -2,\n\t"ErrorMessage": "No such file or directory"\n}'
```

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
gtopper: I mean `app_item in app_items` or `item in items`.

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
gtopper: Sure (if you assume no one will delete the table concurrently), but this block will supress all errors.

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
gtopper: And I said you can't make this assumption 😉

jond01: Ok, `while True` it is

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
gtopper: `HttpResponseError` has a `status_code` attribute.

And even if you couldn't easily do it, that doesn't mean you can just assume that it's a 404...

jond01: > `HttpResponseError` has a `status_code` attribute.

Nice! It's a relatively new addition, so I wasn't fully aware of that:
https://github.com/v3io/v3io-py/blame/6e1942ae481ecf4694d8065a37526e2e6ff10d91/v3io/dataplane/response.py#L27
I will use it.

server/api/api/endpoints/model_monitoring.py
jond01: 4c70829

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
jond01: 76aa677

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
jond01: 4744fca

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
jond01: Fixed to suppressing only 404 in 4744fca

mlrun/model_monitoring/writer.py
jond01: I believe it's still ok at this point, maybe we should remove it in 1.8

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -703,3 +702,67 @@ def get_v3io_monitoring_apps_container(project_name: str) -> str:
...
+        """Get model monitoring results and metrics on the endpoint"""
+        metrics: list[
+            mlrun.common.schemas.model_monitoring.ModelEndpointMonitoringMetric
+        ] = []
+        container = self.get_v3io_monitoring_apps_container(self.project)
+        try:
+            response = self.client.kv.scan(container=container, table_path=endpoint_id)
+        except v3io.dataplane.response.HttpResponseError as err:
+            if err.status_code == HTTPStatus.NOT_FOUND:
+                logger.info(
```
gtopper: Consider making this a warning.

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -703,3 +702,67 @@ def get_v3io_monitoring_apps_container(project_name: str) -> str:
...
+                    table_path=endpoint_id,
+                )
+                return []
+            raise
+
+        while True:
+            metrics.extend(self._extract_metrics_from_items(response.output.items))
+            if response.output.last:
+                break
+            # TODO: Use AIO client: `v3io.aio.dataplane.client.Client`
```
gtopper, line 761: That's a good idea, but it's missing the reason, which I imagine is so that `get_model_endpoint_metrics` doesn't have to be run in a separate thread (via thread pool).

Since this is new code, why not do this from the start (now)?

gtopper (commented):
Looks good, only a couple of comments.

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
jond01: > so that `get_model_endpoint_metrics` doesn't have to be run in a separate thread (via thread pool).

Exactly.

> Since this is new code, why not do this from the start (now)?

It's a good question.
The class already uses the sync client, and I wanted to migrate all the relevant usages to async, not only the new one here.
What do you think?

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
jond01: Done in 0fa9e0b

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
jond01: We agreed on deferring that :)

gtopper (approved):

PR #5486, opened 2024-05-01
[Dask] Enhance extending env vars to avoid memory leak [1.6.x]

liranbg:
Function spec may already include some of the envvars generated by `generate_runtime_k8s_env` causing the extend list items to include them again. every scheduled cycle of such dask function would extend its env var list to infinity

https://iguazio.atlassian.net/browse/ML-6317

### DIFF
--- a/server/api/runtime_handlers/daskjob.py
+++ b/server/api/runtime_handlers/daskjob.py
@@ -328,8 +328,16 @@ def enrich_dask_cluster(
     # leaving no room for human mistakes
     env.extend(
         filter(
-            lambda spec_env: not any(
-                [True for _env in env if _env["name"] == spec_env["name"]]
+            lambda spec_env: isinstance(spec_env, dict)
+            and not any(
+                [
+                    True
+                    for _env in env
+                    # spec_env might be V1EnvVar or a dict
+                    # _env is just a dict
+                    if getattr(spec_env, "name", spec_env.get("name", ""))
+                    == _env["name"]
+                ]
             ),
             spec.env,
         )
--- a/server/api/runtime_handlers/daskjob.py
+++ b/server/api/runtime_handlers/daskjob.py
@@ -326,17 +326,20 @@ def enrich_dask_cluster(
     # in other words, dont let spec.env override env (or not even duplicate it)
     # we dont want to override env to ensure k8s runtime envs are enforced and correct
     # leaving no room for human mistakes
+    def get_env_name(env_: Union[client.V1EnvVar, Dict]) -> str:
+        if isinstance(env_, client.V1EnvVar):
+            return env_.name
+        return env_.get("name", "")
+
     env.extend(
         filter(
-            lambda spec_env: isinstance(spec_env, dict)
-            and not any(
+            lambda spec_env: not any(
                 [
                     True
                     for _env in env
                     # spec_env might be V1EnvVar or a dict
                     # _env is just a dict
-                    if getattr(spec_env, "name", spec_env.get("name", ""))
-                    == _env["name"]
+                    if get_env_name(spec_env) == get_env_name(_env)
                 ]
             ),
             spec.env,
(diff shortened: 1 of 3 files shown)


server/api/runtime_handlers/daskjob.py
```diff
@@ -321,7 +321,20 @@ def enrich_dask_cluster(
...
+                # gc
+                del env[j]
+
+                # prioritize the function runtime env var over the given one
+                env[j] = function_env
+                found = True
+                break
+        if not found:
+            env.append(runtime_env)
+
```
quaark: I suggest changing this code to something a bit simpler with less deletes and breaks.
I will say however the code comment `# prioritize the function runtime env var over the given one` doesn't really describe what's happening, you called the spec.env function env and you are prioritizing that spec.env over the runtime env generated in k8s. So my suggestion has the same logic but I called `function_env` -> `spec_env`. If your comment meant we should prioritize the runtime env, then we can filter the spec env instead and then add the runtime env.
```suggestion
    def _not_in_spec_env(_runtime_env):
        for _env in env:
            if _runtime_env["name"] == _env["name"]:
                return False
        return True
    
    runtime_env = filter(function.generate_runtime_k8s_env(), _not_in_spec_env)
    env.extend(runtime_env)
```

server/api/runtime_handlers/daskjob.py
```diff
@@ -321,7 +321,20 @@ def enrich_dask_cluster(
...
     env = spec.env
-    env.extend(function.generate_runtime_k8s_env())
+    for i, runtime_env in enumerate(function.generate_runtime_k8s_env()):
+        found = False
+        for j, function_env in enumerate(env):
+            if runtime_env["name"] == function_env["name"]:
+                # gc
+                del env[j]
+
+                # prioritize the function runtime env var over the given one
```
TomerShor:
```suggestion
                # prioritize the function runtime env var over the existing one
```

tests/api/runtimes/test_dask.py
```diff
@@ -432,6 +433,89 @@ def test_dask_with_security_context(self, db: Session, client: TestClient):
         _ = runtime.client
         self.assert_security_context(other_security_context)
 
+    def test_enrich_dask_cluster(self):
```
TomerShor, line 437: Define `expected_env` and `expected_labels` and reuse
```
expected_env = [
            {"name": "MLRUN_DEFAULT_PROJECT", "value": "project"},
            {"name": "MLRUN_NAMESPACE", "value": "test-namespace"},
        ]
expected_labels = {
            "mlrun/project": "project",
            "mlrun/class": "dask",
            "mlrun/function": "test",
            "label1": "val1",
            "mlrun/scrape-metrics": "True",
            "mlrun/tag": "latest",
        }
```

TomerShor (changes requested):

quaark (changes requested):
Awesome catch!
I do however suggest making the logic slightly simpler and the code comment is really confusing actually, so I suggest making it clearer or aligning the code to it.

server/api/runtime_handlers/daskjob.py
```diff
@@ -320,8 +320,29 @@ def enrich_dask_cluster(
...
+    env = function.generate_runtime_k8s_env()
+
+    # filter any spec.env that already exists in env
+    # in other words, dont let spec.env override env (or not even duplicate it)
+    # we dont want to override env to ensure k8s runtime envs are enforced and correct
+    # leaving no room for human mistakes
+    env.extend(
+        filter(
+            lambda spec_env: isinstance(spec_env, dict)
+            and not any(
```
quaark: You are excluding all env vars that are `V1EnvVar` with this `isinstance(spec_env, dict) and ...`. Looks like a mistake

quaark (changes requested):

server/api/runtime_handlers/daskjob.py
liranbg: ahh right

PR #5489, opened 2024-05-02
[Notifications] Fix Slack Notifications With Long Messages

quaark:
https://iguazio.atlassian.net/browse/ML-6324

Slack notifications now use headers for making the notification message bold. However these header sections are limited to 150 characters. Any message longer than that resulted in a 400 Bad Request from slack servers.

Mostly this arose from Pipeline start notifications which contain a long url in them. 
So now we truncate the header to the nearest `,` or 150 characters, whichever comes first, and the rest is set in another section below the header.

Attached are images showing before (only section, no header):
Screenshot 2024-05-02 at 11 40 16

And after (section and header):
Screenshot 2024-05-02 at 11 40 35


liranbg (approved):
💯

PR #5492, opened 2024-05-02
[Linting] Upgrade Ruff

jond01:
Upgrade Ruff to the latest version - 0.4.2, which is [even faster](https://astral.sh/blog/ruff-v0.4.0).
In addition, set the minimal required version so that devs get an informative error message when running with an older version locally.

### DIFF
--- a/mlrun/data_types/to_pandas.py
+++ b/mlrun/data_types/to_pandas.py
@@ -65,10 +65,10 @@ def toPandas(spark_df):
                 msg = (
                     "toPandas attempted Arrow optimization because "
                     "'spark.sql.execution.arrow.pyspark.enabled' is set to true; however, "
-                    "failed by the reason below:\n  %s\n"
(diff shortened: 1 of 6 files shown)


mlrun/data_types/to_pandas.py
```diff
@@ -65,10 +65,10 @@ def toPandas(spark_df):
                 msg = (
                     "toPandas attempted Arrow optimization because "
                     "'spark.sql.execution.arrow.pyspark.enabled' is set to true; however, "
-                    "failed by the reason below:\n  %s\n"
+                    f"failed by the reason below:\n  {str(e)}\n"
```
gtopper: `str()` is redundant here.
```suggestion
                    f"failed by the reason below:\n  {e}\n"
```
This applies to all instances of course.

jond01: Done in 95cec63.

gtopper (approved):
LGTM

PR #5496, opened 2024-05-02
[ProjectSummaries] Make `_calculate_pipelines_counters` use pagination

roei3000b:
Fix bug in function `_calculate_pipelines_counters` - it ignored the return value of `page_token` from `_list_pipelines` and because of that if there were some pages, we only return the statistics for the first page.
Fixing it by using pagination in `list_pipelines` function.

https://iguazio.atlassian.net/browse/ML-6121

### DIFF
--- a/server/api/crud/projects.py
+++ b/server/api/crud/projects.py
@@ -399,9 +399,8 @@ class Projects(
             return project_to_running_pipelines_count
 
         try:
-            should_continue_next_page = True
             next_page_token = ""
-            while should_continue_next_page:
+            while True:
                 (
                     _,
                     next_page_token,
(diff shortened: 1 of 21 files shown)


server/api/crud/projects.py
```diff
@@ -396,10 +399,26 @@ async def _calculate_pipelines_counters(
...
+                    page_token=next_page_token,
+                )
+                should_continue_next_page = bool(next_page_token)
+                for pipeline in pipelines:
+                    if (
+                        pipeline["status"]
+                        not in mlrun.run.RunStatuses.stable_statuses()
+                    ):
+                        project_to_running_pipelines_count[pipeline["project"]] += 1
+
```
rokatyy: Could be simplified a bit
```suggestion
            next_page_token = ""
            while True:
                (
                    _,
                    next_page_token,
                    pipelines,
                ) = await fastapi.concurrency.run_in_threadpool(
                    server.api.db.session.run_function_with_new_db_session,
                    self._list_pipelines,
                    page_token=next_page_token,
                )
                for pipeline in pipelines:
                    if (
                        pipeline["status"]
                        not in mlrun.run.RunStatuses.stable_statuses()
                    ):
                        project_to_running_pipelines_count[pipeline["project"]] += 1
               if not next_page_token:
                   break     

```

rokatyy (approved):
LGTM! Added just one minor suggestion

liranbg (changes requested):
Missing a unit test

liranbg (approved):

PR #5509, opened 2024-05-05
[API Gateway] Support access key authentication

TomerShor:
Add support for access key authentication mode, that uses the Cookie header with Iguazio access key for authentication.

Implements https://iguazio.atlassian.net/browse/ML-6359


liranbg (changes requested):
condition that on minimum iguazio version (where you have the fix to support control planes

liranbg (approved):

mlrun/platforms/iguazio.py
```diff
@@ -582,3 +582,48 @@ def sanitize_username(username: str):
...
+
+def min_iguazio_versions(*versions):
+    def decorator(function):
+        def wrapper(*args, **kwargs):
+            if validate_iguazio_version_compatibility(*versions):
+                return function(*args, **kwargs)
+
+            message = (
+                f"{function.__name__} is supported since Iguazio {' or '.join(versions)}, currently using "
+                f"Iguazio {mlconf.nuclio_version}."
```
alonmr: you are using `mlconf.nuclio_version` in the message

mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -17,18 +17,21 @@
...
 
 import requests
+from nuclio.auth import AuthInfo as NuclioAuthInfo
 from requests.auth import HTTPBasicAuth
 
 import mlrun
 import mlrun.common.schemas
 
 from ...model import ModelObj
+from ...platforms.iguazio import min_iguazio_versions
```
alonmr: dont add relative imports

mlrun/platforms/iguazio.py
```diff
@@ -582,3 +582,48 @@ def sanitize_username(username: str):
...
+            )
+            raise mlrun.errors.MLRunIncompatibleVersionError(message)
+
+        return wrapper
+
+    return decorator
+
+
+def validate_iguazio_version_compatibility(*min_versions):
```
alonmr: I feel like this is code duplication maybe we can merge it with `validate_nuclio_version_compatibility`

tests/platforms/test_iguazio.py
```diff
@@ -227,3 +229,23 @@ def test_is_iguazio_session_cookie():
...
+    mlconf.igz_version = "3.5.4"
+
+    for case in [
+        ["3.5.5"],
+        ["3.6.0"],
+        ["3.5.5-b25.20231224135202"],
+        ["4.0.0"],
+        ["3.2.0", "3.6.0"],
+    ]:
+        print(f"Testing case: {case}")
```
alonmr: change print to log and use pytest parameterize because it will continue for the other cases if one of them fails

alonmr (changes requested):
Looks good! A few small comments

alonmr (approved):

PR #5511, opened 2024-05-05
[Alerts] Modify alert entity to have list of ids per alert

yanburman:
https://iguazio.atlassian.net/browse/ML-6369
Note that we break backward compatibility here, but the original endpoint was added in 1.7 and not released yet, so nobody is using it
We add an ability to have list of ids, as in the future (1.8 ?) we want to have 2 more modes of operation where we allow multiple ids as either a single group or as a helper to manage alert logic per id


mlrun/alerts/alert.py
```diff
@@ -120,10 +122,10 @@ def with_notifications(self, notifications: list[Notification]):
             self.notifications.append(notification)
         return self
 
-    def with_entity(self, entity: alert_constants.EventEntity):
-        if not isinstance(entity, alert_constants.EventEntity):
-            raise ValueError("Entity parameter must be of type: EventEntity")
-        self.entity = entity
+    def with_entity(self, entity: alert_constants.EventEntities):
```
theSaarco:
```suggestion
    def with_entities(self, entities: alert_constants.EventEntities):
```

If we're renaming, let's adjust the sdk accordingly.

server/api/crud/alerts.py
```diff
@@ -190,6 +190,17 @@ def process_event(
                 active=active,
             )
 
+    @staticmethod
+    def _event_entity_matches(alert_entity, event_entity):
+        if "*" in alert_entity.ids:
+            return True
+
+        for id in event_entity.ids:
+            if id in alert_entity.ids:
```
theSaarco: Why double iteration? I assume an event will only have a single entity ID (that's why I also let the field still be called `entity` in the Event class rather than `entities`) - you're even validating this in the CRUD.

theSaarco (changes requested):
Some pretty minor comments, but let's make sure our naming and terminology is consistent.

theSaarco (approved):
LGTM

quaark (changes requested):
I don't really understand this PR, and I feel it is adding instability to the code without adding any functionality at all. 
Why would we change the id/entity to a list of ids/entities which we validate to always be a length of 1, and any time we access it we always access `ids[0]`? 
Let's either actually extend the functionality to work with lists or stay with single entity, this "preparation" adds complexity and instability in my honest opinion. (And I'll also add that the PR should explain the rationality of the change as well)

@theSaarco as you approved this PR, maybe you could comment here as well?

theSaarco:
This is to prepare for future stages where we will as support for multiple entities in a single alert. We want to be ready without having to modify the data structures and go with BC changes. 
The sdk currently allows adding a single entity, until such time where we remove the limitation.

quaark (approved):
Spoke with @theSaarco and @yanburman offline and understood the magnitude of implementing multiple entity support is out of scope at the current release, so we will keep the preparation for the next release.
@yanburman please add some info on the plans of the multiple entity support in the PR description.

PR #5514, opened 2024-05-05
[FeatureStore] Spark read optimization

tomerm-iguazio:
1) Added support for additional_filters in:
    a)spark_merger
    b) ParquetSource.to_spark_df.
2) solved remote ingest bug.
3) Added system and unit tests.

[ML-6289](https://iguazio.atlassian.net/browse/ML-6289)

[ML-6289]: https://iguazio.atlassian.net/browse/ML-6289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/datastore/sources.py
```diff
@@ -412,6 +413,51 @@ def to_dataframe(
             **reader_args,
         )
 
+    def _get_spark_additional_filters(self):
+        if not self.additional_filters:
+            return None
+        none_values = [None, np.nan, np.NaN]
+        additional_filters = deepcopy(self.additional_filters)
```
gtopper: What is this deepcopy for? It doesn't seem to be necessary.

tests/system/base.py
```diff
@@ -348,3 +349,11 @@ def _upload_code_to_cluster(self):
...
                 )
         self.uploaded_code = True
+
+    @staticmethod
+    def _sort_df(df: pd.DataFrame, sort_column: str):
+        return (
+            df.reindex(sorted(df.columns), axis=1)
+            .sort_values(by=sort_column)
+            .reset_index(drop=True)
+        )
```
gtopper: Nope, this can't go in the base class for all system tests. It's just a specific utility for a few feature store tests.

mlrun/datastore/sources.py
```diff
@@ -412,6 +413,51 @@ def to_dataframe(
...
+        for filter_tuple in additional_filters:
+            if not filter_tuple:
+                continue
+            if not is_first:
+                spark_filter += " AND "
+            else:
+                is_first = False
+            col_name, op, value = filter_tuple
+            if isinstance(value, (str, datetime)):
+                value = rf"'{value}'"
```
gtopper:
```suggestion
                value = f"'{value}'"
```

mlrun/datastore/sources.py
```diff
@@ -412,6 +413,51 @@ def to_dataframe(
             **reader_args,
         )
 
+    def _get_spark_additional_filters(self):
+        if not self.additional_filters:
+            return None
+        none_values = [None, np.nan, np.NaN]
```
gtopper: Need to add a comment to clarify what this is about.

gtopper: Also not really sure why both `np.nan` and `np.NaN` are here? Aren't these aliases for the same thing?

mlrun/datastore/sources.py
```diff
@@ -412,6 +413,51 @@ def to_dataframe(
...
+        additional_filters = deepcopy(self.additional_filters)
+        is_first = True
+        spark_filter = ""
+        for filter_tuple in additional_filters:
+            if not filter_tuple:
+                continue
+            if not is_first:
```
gtopper:
```suggestion
            if spark_filter:
```
No need for the extra variable.

mlrun/datastore/sources.py
```diff
@@ -412,6 +413,51 @@ def to_dataframe(
...
+                spark_filter += " AND "
+            else:
+                is_first = False
+            col_name, op, value = filter_tuple
+            if isinstance(value, (str, datetime)):
+                value = rf"'{value}'"
+            if isinstance(value, (list, tuple, set, np.ndarray)) and (
```
gtopper: Why `np.ndarray`? According to [pyarrow docs](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetDataset.html), it must be
> list, a set or a tuple

mlrun/datastore/sources.py
```diff
@@ -412,6 +413,51 @@ def to_dataframe(
...
+                "in" in op.lower()
```
gtopper: Maybe clearer and safer to make this more specific. i.e. `op.lower() in ["in", "not in"]`.

mlrun/datastore/sources.py
```diff
@@ -412,6 +413,51 @@ def to_dataframe(
...
+                if len(value) == 1:
+                    value = f"({value[0]})"
+                if none_exists:
+                    if value:
+                        spark_filter += f"({str(col_name)} {str(op)} {str(value)} OR {str(col_name)} IS NULL)"
+                    else:
+                        spark_filter += f"{str(col_name)} IS NULL"
+                    continue
+
+            spark_filter += f"{str(col_name)} {str(op)} {str(value)}"
```
gtopper: Redundant `str`.
```suggestion
            spark_filter += f"{col_name} {op} {value}"
```
(other occurrences as well)

mlrun/datastore/sources.py
```diff
@@ -412,6 +413,51 @@ def to_dataframe(
...
+                for none_value in none_values:
+                    if none_value in value:
+                        value.remove(none_value)
+                        none_exists = True
+                value = tuple(value)
+                if len(value) == 1:
+                    value = f"({value[0]})"
+                if none_exists:
+                    if value:
+                        spark_filter += f"({str(col_name)} {str(op)} {str(value)} OR {str(col_name)} IS NULL)"
```
gtopper: This is actually not really safe... It opens us up to (probably accidental rather than malicious) SQL injection. It would be better to use the structured API – https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.filter.html (see "Filter by Column instances").

gtopper (changes requested):
Good work!

mlrun/datastore/sources.py
tomerm-iguazio: Nice

mlrun/datastore/sources.py
tomerm-iguazio: nice!

mlrun/datastore/sources.py
tomerm-iguazio: lets talk

tests/system/base.py
tomerm-iguazio: I need it in `TestFeatureStoreSparkEngine` and `TestFeatureStore`, and they bot inherit from `TestMLRunSystem`.
Why do we need to implement it twice?

mlrun/datastore/sources.py
tomerm-iguazio: you're right, in  value = list(value) it created a new project with the same vars. so it considered safe.

tests/system/base.py
gtopper: False dichotomy. You can put it in a utility module of some sort.

mlrun/datastore/sources.py
assaf758: `a`

mlrun/datastore/sources.py
```diff
@@ -313,6 +315,13 @@ def __init__(
         end_time: Optional[Union[datetime, str]] = None,
         additional_filters: Optional[list[tuple]] = None,
     ):
+        if additional_filters:
+            additional_filters_dict = {"additional_filters": additional_filters}
+            if attributes:
+                attributes = dict(attributes)
+                attributes.update(additional_filters_dict)
+            else:
+                attributes = additional_filters_dict
```
gtopper: Anyway, this is much simpler
```suggestion
            attributes = attributes or {}
            attributes["additional_filters"] = additional_filters
```

mlrun/datastore/sources.py
```diff
@@ -313,6 +315,13 @@ def __init__(
         end_time: Optional[Union[datetime, str]] = None,
         additional_filters: Optional[list[tuple]] = None,
     ):
+        if additional_filters:
+            additional_filters_dict = {"additional_filters": additional_filters}
+            if attributes:
+                attributes = dict(attributes)
```
gtopper: I think you meant `additional_filters = dict(additional_filters)`. Attributes is already a dict.

mlrun/datastore/sources.py
```diff
@@ -313,6 +315,13 @@ def __init__(
...
+                attributes.update(additional_filters_dict)
```
gtopper: The types don't match. `attributes` is `dict[str, str]` and `additional_filters` is `Optional[list[tuple]]` and I think you meant to convert it to a dictionary above, but either way it's not a string.

mlrun/datastore/sources.py
```diff
@@ -412,6 +424,92 @@ def to_dataframe(
...
+            "==": operator.eq,
+            "=": operator.eq,
+            ">": operator.gt,
+            "<": operator.lt,
+            ">=": operator.ge,
+            "<=": operator.le,
+            "!=": operator.ne,
+        }
+
+        none_values = [None, np.nan, np.NaN, float("nan")]
```
gtopper: Again, `np.nan` and `np.NaN` are the same thing.

mlrun/datastore/sources.py
```diff
@@ -412,6 +424,92 @@ def to_dataframe(
...
+        new_filter = lit(True)
+        for filter_tuple in self.additional_filters:
+            if not filter_tuple:
+                continue
+            col_name, op, value = filter_tuple
+            if op.lower() in ("in", "not in") and isinstance(value, (list, tuple, set)):
+                none_exists = False
+                value = list(value)
+                for none_value in none_values:
+                    if none_value in value:
```
gtopper: This doesn't work. Need to update test to cover this and fix. The correct way to check for nan is generally with `math.isnan()`
```python
>>> float("nan") in [float("nan")]
False
```

mlrun/datastore/sources.py
```diff
@@ -412,6 +424,92 @@ def to_dataframe(
...
+                                if filter_nan
+                                else new_filter
+                            )
+                else:
+                    if op.lower() == "in":
+                        new_filter = col(col_name).isin(value)
+                    elif op.lower() == "not in":
+                        new_filter = ~col(col_name).isin(value)
+            else:
+                new_filter = operators[op](col(col_name), value)
```
gtopper, line 501: Should probably raise a better error on invalid operator.

mlrun/datastore/sources.py
```diff
@@ -412,6 +424,92 @@ def to_dataframe(
...
+                            )
+                        else:
+                            new_filter = (
+                                ~col(col_name).isin(value) & ~col(col_name).isNull()
+                            )
+                            new_filter = (
+                                new_filter & ~isnan(col(col_name))
+                                if filter_nan
+                                else new_filter
+                            )
```
gtopper: Simplify
```suggestion
                            if filter_nan:
                                new_filter = new_filter & ~isnan(col(col_name))
```
(all occurrences)

tests/system/feature_store/test_spark_engine.py
```diff
@@ -332,6 +333,69 @@ def test_basic_remote_spark_ingest(self):
...
+            source=parquet_source,
+            targets=[target],
+            spark_context=self.spark_service,
+            run_config=run_config,
+        )
+        result = mlrun.datastore.utils.sort_df(
+            pd.read_parquet(feature_set.get_target_path()), "patient_id"
+        )
+        expected = mlrun.datastore.utils.sort_df(filtered_df, "patient_id")
+        assert_frame_equal(result, expected, check_dtype=False, check_categorical=False)
```
gtopper: It's not good for tests to reuse a part of the implementation, because it allows for changing behaviour without having to update the test, among other things.

mlrun/datastore/sources.py
```diff
@@ -412,6 +424,92 @@ def to_dataframe(
             **reader_args,
         )
 
+    def _get_spark_additional_filters(self, column_types: dict):
```
gtopper: get -> build or something like that. This is not a getter.

mlrun/datastore/sources.py
```diff
@@ -412,6 +424,92 @@ def to_dataframe(
...
+                spark_filter = spark_filter & new_filter
+            else:
+                spark_filter = new_filter
+        return spark_filter
+
+    def _filter_spark_df(self, df, time_field=None, columns=None):
+        spark_additional_filters = self._get_spark_additional_filters(
+            column_types=dict(df.dtypes)
+        )
+        if spark_additional_filters is not None:
```
gtopper, line 516: I think here and in a few other places, it's possible to use the more idiomatic
```suggestion
        if spark_additional_filters:
```
dropping the `is not None`.

mlrun/datastore/sources.py
```diff
@@ -341,6 +349,10 @@ def end_time(self):
     def end_time(self, end_time):
         self._end_time = self._convert_to_datetime(end_time)
 
+    @property
+    def additional_filters(self):
+        return self.attributes.get("additional_filters", None)
```
gtopper: Redundant `None`
```suggestion
        return self.attributes.get("additional_filters")
```

mlrun/datastore/sources.py
```diff
@@ -412,6 +424,92 @@ def to_dataframe(
...
+                            new_filter = col(col_name).isNull()
+                            new_filter = (
+                                new_filter & isnan(col(col_name))
+                                if filter_nan
+                                else new_filter
+                            )
+                        else:
+                            new_filter = ~col(col_name).isNull() & ~isnan(col(col_name))
+                            new_filter = (
+                                new_filter & ~isnan(col(col_name))
```
gtopper: Something here doesn't make sense... `& ~isnan(col(col_name))` is applied twice.

mlrun/datastore/sources.py
```diff
@@ -412,6 +424,92 @@ def to_dataframe(
...
+            col_name, op, value = filter_tuple
+            if op.lower() in ("in", "not in") and isinstance(value, (list, tuple, set)):
+                none_exists = False
+                value = list(value)
+                for none_value in none_values:
+                    if none_value in value:
+                        value.remove(none_value)
+                        none_exists = True
+                if none_exists:
+                    filter_nan = column_types[col_name] != "timestamp"
```
gtopper: What about other types that can't be NaN, like strings?

tests/system/feature_store/test_spark_engine.py
```diff
@@ -332,6 +333,69 @@ def test_basic_remote_spark_ingest(self):
...
+            name="test-fs-vec", features=["parquet-filters-fs.*"]
+        )
+        vec.save()
+        get_offline_target = ParquetTarget(
+            "mytarget", path=f"{self.output_dir()}-get_offline_features"
+        )
+        kind = None if self.run_local else "remote-spark"
+        resp = fstore.get_offline_features(
+            feature_vector=vec,
+            additional_filters=[("bad", "=", 95)],
```
gtopper: I guess better test a more complex one, to really test the filter logic. I'm thinking at least two filters, some thing with `not in` and something with NaN (which so much of the logic is dedicated to).

gtopper (changes requested):
Nice work replacing the string query with structured dataframe operations.

mlrun/datastore/sources.py
tomerm-iguazio: you can not use if condition with spark Column object.
That is why I have not use not None...

mlrun/datastore/sources.py
tomerm-iguazio: fixed tnx!

mlrun/datastore/sources.py
tomerm-iguazio: It will not cause any issue, unlike timestamp (or date that I just added) types that will raise an error.

mlrun/datastore/sources.py
tomerm-iguazio: I want to create a copy of attributes, that is why I used dict(attributes). I do not want to update original attributes.

tomerm-iguazio: I changed it to something simpler:

```suggestion
            attributes = deepcopy(attributes) or {}
            attributes["additional_filters"] = additional_filters
```

mlrun/datastore/sources.py
tomerm-iguazio: it was in order to create another object, I change it to use deepcopy

mlrun/datastore/sources.py
tomerm-iguazio: fixed

mlrun/datastore/sources.py
```diff
@@ -361,13 +370,14 @@ def to_step(
...
         import storey
 
-        attributes = self.attributes or {}
+        attributes = self.attributes.copy() or {}
+        print(f"attributes1: {attributes}")
+        attributes.pop("additional_filters", None)
         if context:
             attributes["context"] = context
-
+        print(f"attributes2: {attributes}")
```
gtopper: ❗

mlrun/datastore/sources.py
```diff
@@ -361,13 +370,14 @@ def to_step(
     ):
         import storey
 
-        attributes = self.attributes or {}
+        attributes = self.attributes.copy() or {}
+        print(f"attributes1: {attributes}")
```
gtopper: ❗

mlrun/datastore/sources.py
```diff
@@ -361,13 +370,14 @@ def to_step(
     ):
         import storey
 
-        attributes = self.attributes or {}
+        attributes = self.attributes.copy() or {}
```
gtopper: Why deepcopy in some places and copy in others?

gtopper: Also, it's not clear what's the point of falling back on `{}`? It seems redundant here.

mlrun/datastore/sources.py
gtopper: This issue is still outstanding.

mlrun/datastore/sources.py
gtopper: This issue is still outstanding.

tests/system/feature_store/utils.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2023 Iguazio
```
gtopper:
```suggestion
# Copyright 2024 Iguazio
```

tests/system/feature_store/test_spark_engine.py
gtopper: Still missing a NaN value in the test data afaict. Not to be confused with `None`.

gtopper (changes requested):

mlrun/datastore/sources.py
tomerm-iguazio: please look at the next loop I did: 


```
for sub_value in value:
    if isinstance(sub_value, float) and math.isnan(sub_value):
        value.remove(sub_value)
        none_exists = True
```

mlrun/datastore/sources.py
tomerm-iguazio: plz look at else case:

```
  elif op in operators:
      new_filter = operators[op](col(col_name), value)
  else:
          f"additional filter operator is not supported,"
          f"given Operator: {op} ."
      )
```

mlrun/datastore/sources.py
tomerm-iguazio: You are right, in the builder method of the base class if attributes is None, it sets it to empty dict.

mlrun/datastore/sources.py
gtopper: Still need to remove the incorrect membership test for NaN, and replace
```python
for none_value in none_values:
    if none_value in value:
```
with a simple
```python
if value is None:
```

mlrun/datastore/sources.py
```diff
@@ -412,6 +420,90 @@ def to_dataframe(
...
+                    if op.lower() == "in":
+                        new_filter = col(col_name).isin(value)
+                    elif op.lower() == "not in":
+                        new_filter = ~col(col_name).isin(value)
+            elif op in operators:
+                new_filter = operators[op](col(col_name), value)
+            else:
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    f"additional filter operator is not supported,"
+                    f"given Operator: {op} ."
```
gtopper:
```suggestion
                    f"unsupported filter operator: {op}"
```

mlrun/datastore/sources.py
gtopper: It cannot be `None` anyway because the constructor already replaces `None` with `{}`.

mlrun/datastore/sources.py
```diff
@@ -313,6 +316,9 @@ def __init__(
         end_time: Optional[Union[datetime, str]] = None,
         additional_filters: Optional[list[tuple]] = None,
     ):
+        if additional_filters:
+            attributes = deepcopy(attributes) or {}
+            attributes["additional_filters"] = additional_filters
```
gtopper, line 320: Type annotations still don't match
```
Unexpected type(s): (str, list[tuple]) Possible type(s): (str, str) (str, str)
```

mlrun/datastore/sources.py
```diff
@@ -350,6 +359,25 @@ def _convert_to_datetime(time):
...
+            col_name, op, value = filter_tuple
+            if isinstance(value, float) and math.isnan(value):
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    "using nan in additional_filters is not supported"
+                )
+            elif isinstance(value, (list, tuple, set)):
+                for sub_value in value:
+                    if isinstance(sub_value, float) and math.isnan(sub_value):
+                        raise mlrun.errors.MLRunInvalidArgumentError(
+                            "using nan in additional_filters is not supported"
```
gtopper:
```suggestion
                            "using NaN in additional_filters is not supported"
```

mlrun/datastore/sources.py
```diff
@@ -350,6 +359,25 @@ def _convert_to_datetime(time):
...
+    def validate_additional_filters(additional_filters):
+        if not additional_filters:
+            return
+        for filter_tuple in additional_filters:
+            if not filter_tuple:
+                continue
+            col_name, op, value = filter_tuple
+            if isinstance(value, float) and math.isnan(value):
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    "using nan in additional_filters is not supported"
```
gtopper:
```suggestion
                    "using NaN in additional_filters is not supported"
```

mlrun/datastore/sources.py
```diff
@@ -361,13 +389,12 @@ def to_step(
     ):
         import storey
 
-        attributes = self.attributes or {}
+        attributes = deepcopy(self.attributes)
```
gtopper: Isn't a regular copy enough?

mlrun/datastore/sources.py
```diff
@@ -412,6 +439,84 @@ def to_dataframe(
...
+                        value.remove(sub_value)
+                        none_exists = True
+                if none_exists:
+                    filter_nan = column_types[col_name] not in ("timestamp", "date")
+                    if value:
+                        if op.lower() == "in":
+                            new_filter = (
+                                col(col_name).isin(value) | col(col_name).isNull()
+                            )
+                            if filter_nan:
```
gtopper, line 477: I thought we dropped NaN support?

mlrun/datastore/sources.py
tomerm-iguazio: no, if you have "in" operator it can function as a nested dict.

mlrun/datastore/sources.py
tomerm-iguazio: I still filter if nan exists in the df, and you enter None.

mlrun/datastore/sources.py
```diff
@@ -305,14 +307,18 @@ def __init__(
...
+        attributes: dict[str, object] = None,
         key_field: str = None,
         time_field: str = None,
         schedule: str = None,
         start_time: Optional[Union[datetime, str]] = None,
         end_time: Optional[Union[datetime, str]] = None,
         additional_filters: Optional[list[tuple]] = None,
     ):
+        if additional_filters:
+            attributes = deepcopy(attributes) or {}
```
gtopper:
```suggestion
            attributes = copy(attributes) or {}
```

gtopper (approved):

PR #5515, opened 2024-05-05
[Application] Create API Gateway by default

TomerShor:
To better secure access to application runtimes, we want to utilize the authentication features of API Gateways, and use it as the default gateway to the application.

Thus, in this PR we create an API Gateway by default when creating an application runtime, with access key authentication mode, and set it on the runtime's status.
Furthermore, when invoking the application using `invoke`, we prioritize invoking the api gateway, rather then the function's external invocation url.

This PR is dependent on https://github.com/mlrun/mlrun/pull/5509 and https://github.com/mlrun/mlrun/pull/5512, as it is using the features added there.

Resolves https://iguazio.atlassian.net/browse/ML-6364 and https://iguazio.atlassian.net/browse/ML-6365


mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -389,10 +392,10 @@ def invoke(
         ):
             # inject access key from env
             auth = NuclioAuthInfo().from_envvar().to_requests_auth()
-
+        url = urljoin(self.invoke_url, path or "")
```
rokatyy, line 409: This is already done in `self.invoke_url`

TomerShor: the `ApplicationRuntime`'s `invoke` function accepts a required `path` argument.
This is due to the way `invoke` is defined the parent class `RemoteRuntime`.

So, we could potentially have a sub-path to invoke with, regardless of the path in the API Gateway's spec.

rokatyy: @TomerShor gotcha, however `self.invoke_url` already does `urljoin(host, self.spec.path)`. Does it mean that we will do urljoin twice?

mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -389,10 +392,10 @@ def invoke(
         ):
             # inject access key from env
             auth = NuclioAuthInfo().from_envvar().to_requests_auth()
```
rokatyy: What if user doesn't have this envvar set? I think it would improve UX if we warn a user about it.

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -269,6 +292,9 @@ def deploy(
             builder_env,
         )
 
+        ports = self.spec.internal_application_port if direct_port_access else []
+        self.create_api_gateway(skip_access_key_auth=skip_access_key_auth, ports=ports)
+
```
rokatyy: More a question than a comment. Do we want to provide a way to create an api gateway with skip_access_key only? What if user wants to create it with BasicAuth?

mlrun/runtimes/nuclio/api_gateway.py
TomerShor: Yes, in the following example case:
- Host: "my.domain.com"
- API Gateway path: "my-path"
- Call `invoke("/some-sub-path")`

The result URL to be called should be `"my.domain.com/my-path/some-sub-path"`

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -290,6 +316,77 @@ def with_source_archive(
...
+            )
+
+        auth = (auth_info.username, auth_info.password) if auth_info else None
+
+        return self.status.api_gateway.invoke(
+            method=method or "POST",
+            headers=headers,
+            auth=auth,
+            path=path,
+        )
```
rokatyy: The way we work here with `auth` looks a bit confusing. Due to introduction of `accessKey` auth, it makes sense from user perspective that `auth` param represents any authentication if it's needed. However, in this implementation `auth` represents only `basicAuth` and has nothing to do with `accessKey`. I think we should either rename `auth` to `basic_auth` or smth like this or rework the logic

mlrun/runtimes/nuclio/api_gateway.py
TomerShor: 👍 will raise exception in that case

mlrun/runtimes/nuclio/application/application.py
TomerShor: That a valid question.
Users can still manually create api gateways with different configuration that will direct to the application runtime.
The question is what should be the default behavior.
Thinking about it, we might want to accept the `mode` as an argument, and perhaps have some default mode in the mlrun config.

rokatyy: @TomerShor as an option, we can do the same way we do in api gateway init. Which means passing `APIGatewayAuthenticator` when create api gateway. But in case of application runtime we can create an api gateway with `accessKey` by default.

mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -385,17 +392,21 @@ def invoke(
...
             self.spec.authentication.authentication_mode
-            == NUCLIO_API_GATEWAY_AUTHENTICATION_MODE_ACCESS_KEY
+            == APIGatewayAuthenticationMode.ACCESS_KEY.value
         ):
             # inject access key from env
             auth = NuclioAuthInfo().from_envvar().to_requests_auth()
-
+            if not auth:
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    "API Gateway invocation requires authentication. Please set V3IO_ACCESS_KEY env var"
```
liranbg, line 407: Remember that this should be invocable not only from jupyter but other. if user can pass it explicitly that would be great.
thus I suggest setting the `auth` object as either
(username, password) or (username?, v3io_access_key | read_from_envvar)

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -290,6 +338,86 @@ def with_source_archive(
...
+            ),
+        )
+
+        authentication_mode = (
+            authentication_mode
+            or mlrun.mlconf.function.application.default_authentication_mode
+        )
+        if authentication_mode == APIGatewayAuthenticationMode.ACCESS_KEY:
+            api_gateway.with_access_key_auth()
+        if authentication_mode == "basic":
```
liranbg:
```suggestion
        elif authentication_mode == APIGatewayAuthenticationMode.BASIC:
```

mlrun/runtimes/nuclio/application/application.py
TomerShor: I prefer using a `mode` to allow changing defaults in mlrun config, and use the `with_XXX` methods for the API Gateway, instead of forcing the user to pass a `APIGatewayAuthenticator` object (a bit more complex usage IMO).

liranbg (commented):
minor

mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -18,33 +18,36 @@
...
-NUCLIO_API_GATEWAY_AUTHENTICATION_MODE_BASIC_AUTH = "basicAuth"
-NUCLIO_API_GATEWAY_AUTHENTICATION_MODE_NONE = "none"
-NUCLIO_API_GATEWAY_AUTHENTICATION_MODE_ACCESS_KEY = "accessKey"
 PROJECT_NAME_LABEL = "nuclio.io/project-name"
 
 
+class APIGatewayAuthenticationMode(mlrun.common.types.StrEnum):
+    NONE = "none"
+    BASIC = "basicAuth"
+    ACCESS_KEY = "accessKey"
```
rokatyy: Why not using https://github.com/mlrun/mlrun/blob/87244882c16090977448d7b1500842d3732921fe/mlrun/common/schemas/api_gateway.py#L23?

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -290,6 +337,89 @@ def with_source_archive(
...
+    def invoke(
+        self,
+        path: str,
+        body: typing.Union[str, bytes, dict] = None,
+        method: str = None,
+        headers: dict = None,
+        dashboard: str = "",
+        force_external_address: bool = False,
+        auth_info: AuthInfo = None,
+        mock: bool = None,
```
rokatyy: `auth_info` should be initialised and passed by user, right? Why not aligning this logic with how it is done in API Gateway class which means just passing the credentials.

rokatyy (approved):
Cool staff!
Having just 2 comments which are minor imo.

mlrun/runtimes/nuclio/application/application.py
TomerShor: Because `invoke` is inherited from the base class `RemoteRuntime`.
Changing it will break backwards compatibility.

mlrun/runtimes/nuclio/api_gateway.py
TomerShor: No actual reason :|
removing it

liranbg (approved):

PR #5519, opened 2024-05-06
[API] Add read model monitoring metrics values endpoint

jond01:
Implements [ML-6360](https://iguazio.atlassian.net/browse/ML-6360).

Read the data from V3IO TSDB "app-results" table.

* Enforce timezone in the start and end times.
* Require metrics names.
* Add API part to the primary system test - the whole test passed.

Usage example:

(1) Request the model endpoints:
```
GET api/v1
/projects/test-app-flow-api-t8
/model-endpoints
```
Receive the single model endpoint UID: `ea2225f9a1bf7f986beeaa59632cfab44f0c244b`.

(2) Request the results of this model endpoint (implemented in #5484):
```
GET api/v1
/projects/test-app-flow-api-t8
/model-endpoints/ea2225f9a1bf7f986beeaa59632cfab44f0c244b
/metrics?type=results
```
Response:
image

(3) Request the results values using the wanted `full_name`s: 
```
GET api/v1
/projects/test-app-flow-api-t8
/model-endpoints/ea2225f9a1bf7f986beeaa59632cfab44f0c244b
/metrics-values
?name=test-app-flow-api-t8.histogram-data-drift.result.general_drift
&name=test-app-flow-api-t8.monitoring-test-v2.result.model_perf
&name=test-app-flow-api-t8.no-app.result.no-res
```
Response:
image

Start and end times can be added in (3):
```
&start=2024-04-01T18:00:28Z
&end=2024-04-04T18:00:28Z
```
Without them, the last 24 hours of data is fetched.

[ML-6360]: https://iguazio.atlassian.net/browse/ML-6360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/model_monitoring/db/v3io_tsdb_reader.py
```diff
@@ -0,0 +1,132 @@
...
+    ModelEndpointMonitoringResultValues,
+    _compose_full_name,
+    _ModelEndpointMonitoringResultValuesBase,
+)
+from mlrun.utils import logger
+
+
+def _get_sql_query(endpoint_id: str, names: list[tuple[str, str]]) -> str:
+    query = (
+        f"SELECT * FROM '{mm_writer._TSDB_TABLE}' "
```
gtopper: Maybe it shouldn't be private if we're going to access it from other modules.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
```diff
@@ -0,0 +1,132 @@
...
+                f"({mm_writer.WriterEvent.APPLICATION_NAME}='{app_name}' "
+                f"AND {mm_writer.ResultData.RESULT_NAME}='{result_name}')"
+            )
+            if i != 0:  # not first sub condition
+                query += " OR "
+            query += sub_cond
+
+        query += ")"
+
+    query += ";"
```
gtopper: Maybe use `StringIO` to keep allocations linear?

server/api/api/endpoints/model_endpoints.py
```diff
@@ -343,16 +357,113 @@ async def get_model_endpoint_monitoring_metrics(
...
+    )
+
+
+@router.get(
+    "/{endpoint_id}/metrics-values",
+    response_model=list[
+        mlrun.common.schemas.model_monitoring.model_endpoints._ModelEndpointMonitoringResultValuesBase
+    ],
+)
+async def get_model_endpoint_monitoring_metrics_values(
```
gtopper, line 451: Shouldn't we add an integration test for this new API?

mlrun/model_monitoring/db/v3io_tsdb_reader.py
jond01: I get what you say, but I'm not familiar with it. Do you have an example?

mlrun/model_monitoring/db/v3io_tsdb_reader.py
jond01: Ok, but I think that the db subpackage should be private instead (`mlrun.model_monitoring._db`)

jond01: This function isn't accessed from other modules at the moment.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
jond01: Done in c6b794e.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
```diff
@@ -0,0 +1,135 @@
...
+from mlrun.common.schemas.model_monitoring.model_endpoints import (
+    ModelEndpointMonitoringMetric,
+    ModelEndpointMonitoringMetricType,
+    ModelEndpointMonitoringResultNoData,
+    ModelEndpointMonitoringResultValues,
+    _compose_full_name,
+    _ModelEndpointMonitoringResultValuesBase,
+)
+from mlrun.model_monitoring.db.stores.v3io_kv.kv_store import KVStoreBase
+from mlrun.model_monitoring.db.tsdb.v3io.v3io_connector import _TSDB_BE
```
gtopper, line 35: I guess here too, if we need to import it here, maybe it shouldn't be private.

server/api/api/endpoints/model_endpoints.py
```diff
@@ -343,16 +357,113 @@ async def get_model_endpoint_monitoring_metrics(
...
+            raise mlrun.errors.MLRunInvalidArgumentTypeError(
+                "Custom start and end times must contain the timezone."
+            )
+        if inputs.start > inputs.end:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "The start time must precede the end time."
+            )
+    else:
+        raise mlrun.errors.MLRunInvalidArgumentError(
+            "Provided only one of start time, end time. Provide both or none of them"
```
gtopper:
```suggestion
            "Provided only one of start time, end time. Please provide both or neither."
```

mlrun/model_monitoring/db/v3io_tsdb_reader.py
jond01: You're right, but this module should move entirely to `mlrun.model_monitoring.db.tsdb.v3io.v3io_connector`, so this import won't happen :)

server/api/api/endpoints/model_endpoints.py
jond01: Done in 24778da.

gtopper (approved):

PR #5521, opened 2024-05-06
[Datastore] Rename `DatastoreProfileAzureBlob` 'bucket' parameter to 'container'

alxtkr77:
https://iguazio.atlassian.net/browse/ML-6326

### DIFF
--- a/mlrun/datastore/datastore_profile.py
+++ b/mlrun/datastore/datastore_profile.py
@@ -188,6 +188,7 @@ class DatastoreProfileS3(DatastoreProfile):
     bucket: typing.Optional[str] = None
 
     @pydantic.validator("bucket")
+    @classmethod
     def check_bucket(cls, v):
         if not v:
             warnings.warn(
@@ -292,6 +293,7 @@ class DatastoreProfileGCS(DatastoreProfile):
     bucket: typing.Optional[str] = None
 
     @pydantic.validator("bucket")
+    @classmethod
     def check_bucket(cls, v):
         if not v:
             warnings.warn(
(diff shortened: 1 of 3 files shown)


gtopper (approved):

mlrun/datastore/datastore_profile.py
```diff
@@ -344,24 +344,24 @@ class DatastoreProfileAzureBlob(DatastoreProfile):
     client_secret: typing.Optional[str] = None
     sas_token: typing.Optional[str] = None
     credential: typing.Optional[str] = None
-    bucket: typing.Optional[str] = None
+    container: typing.Optional[str] = None
 
-    @pydantic.validator("bucket")
-    def check_bucket(cls, v):
+    @pydantic.validator("container")
+    def check_container(cls, v):
```
gtopper: Need to add `@classmethod` or change `cls` -> `self`.
(all occurrences)

alxtkr77: Shouldn't we enforce it in lint though?

gtopper: @jond01

gtopper (approved):

mlrun/datastore/datastore_profile.py
jond01: The relevant rules are:
* N804: https://docs.astral.sh/ruff/rules/invalid-first-argument-name-for-class-method/
* N805: https://docs.astral.sh/ruff/rules/invalid-first-argument-name-for-method/

They are a part of the pep8-naming rules (`N`), which are not selected in mlrun. I encourage you to add it 🙂

PR #5523, opened 2024-05-06
[ProjectSummaries] Add pipelines_completed_recent_count, pipelines_failed_recent_count

roei3000b:
Add pipelines_completed_recent_count, pipelines_failed_recent_count statistics to the project summaries endpoint.
`pipelines_completed_recent_count` - Succeeded within the last 24h
`pipelines_failed_recent_count` -  Failed within the last 24h

https://iguazio.atlassian.net/browse/ML-6122


server/api/crud/projects.py
```diff
@@ -417,7 +443,24 @@ async def _calculate_pipelines_counters(
...
                     ):
                         project_to_running_pipelines_count[pipeline["project"]] += 1
-
+                    elif "finished_at" in pipeline:
+                        finished_at = datetime.datetime.strptime(
+                            pipeline["finished_at"], "%Y-%m-%d %H:%M:%S%z"
+                        )
+                        if finished_at > datetime.datetime.now().astimezone(
+                            tz=datetime.timezone.utc
+                        ) - datetime.timedelta(days=1):
```
theSaarco, line 452: We need to be performance-minded, and this query is extremely inefficient, looking for all pipelines from all projects, just so we can filter pipelines from last day here.
For our purposes this should be done with 2 filtered-by-kfp `list_pipelines` - one looking for running pipelines (by filtering on status not being a stable status) and another looking for pipelines completed in the last day. Should check to see if kfp allows such a filter (not all fields can be used to filter) - if it does, then this query should definitely be filtered already in the kfp side.

roei3000b: Hey, agree with you but our version doesn't support filtering by `finished_at`.
They added it only in this pr:
https://github.com/kubeflow/pipelines/pull/8662

theSaarco: Got it. Thanks for checking.

server/api/crud/projects.py
```diff
@@ -427,9 +470,13 @@ async def _calculate_pipelines_counters(
                 "Failed to list pipelines. Pipelines counters will be set to None",
                 exc=mlrun.errors.err_to_str(exc),
             )
-            return collections.defaultdict(lambda: None)
+            return [collections.defaultdict(lambda: None)] * 3
```
liranbg, line 477: why * 3? comment maybe

liranbg (approved):

PR #5534, opened 2024-05-08
[Logger] Enhance formatter creation [1.6.x]

liranbg:
Potentially reduce the creation of formatter instance creation every time configmap is loaded.
Now, create the instance only if really needed

In addition, ensured that dask-job has no duplicate envvar by name to mitigate cases where function was prone to https://github.com/mlrun/mlrun/commit/dbb53dc1e528cfa04c02960e9de29fcad530c9ed issue

### DIFF
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -1383,7 +1383,7 @@ def read_env(env=None, prefix=env_prefix):
         current_formatter_name = current_handler.formatter.__class__.__name__
         desired_formatter_name = log_formatter.__name__
         if current_formatter_name != desired_formatter_name:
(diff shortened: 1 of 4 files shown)


mlrun/config.py
```diff
@@ -1376,12 +1376,12 @@ def read_env(env=None, prefix=env_prefix):
     if log_formatter_name := config.get("log_formatter"):
         import mlrun.utils.logger
 
-        log_formatter = mlrun.utils.create_formatter_instance(
+        log_formatter = mlrun.utils.resolve_formatter_by_kind(
```
TomerShor, line 1379: don't you need to instantiate this formatter?

PR #5536, opened 2024-05-09
[Functions] Update function with deletion task id

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-6422

This task is a small part of  https://iguazio.atlassian.net/browse/ML-3432 during which function deletion api of v2 was implemented. In the new implementation, we've introduced a background task that handles function deletion and returns a task name to the user.

During the initial implementation, we overlooked how the UI would recognize that a function is being deleted. This task addresses that oversight by ensuring that function["status"]["deletion_task_id"] is set to deletion_task_id before the function is deleted. This way, the UI can accurately indicate that the function is in the process of being deleted.

### DIFF
--- a/server/api/api/utils.py
+++ b/server/api/api/utils.py
@@ -1351,13 +1351,13 @@ async def _delete_function(
                 error_message = f"Failed to delete function {function_name}. Errors: {' '.join(failed_requests)}"
                 raise mlrun.errors.MLRunInternalServerError(error_message)
 
-    # delete the function from the database
-    await run_in_threadpool(
-        server.api.crud.Functions().delete_function,
-        db_session,
-        project,
-        function_name,
-    )
+        # delete the function from the database
+        await run_in_threadpool(
+            server.api.crud.Functions().delete_function,
+            db_session,
+            project,
+            function_name,
+        )
 
 
 async def _update_functions_with_deletion_task_ids(
--- a/server/api/api/utils.py
+++ b/server/api/api/utils.py
@@ -1318,49 +1318,49 @@ async def _delete_function(
         project,
         function_name,
     )
-    if len(functions) > 0:
-        logger.debug(
-            "Updating functions with deletion task id",
-            function_name=function_name,
-            functions_count=len(functions),
-            project=project,
-        )
-
-        # update functions with deletion task id
-        await _update_functions_with_deletion_task_ids(
-            db_session, functions, project, background_task_name
-        )
-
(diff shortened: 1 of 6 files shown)


server/api/api/utils.py
```diff
@@ -1343,6 +1352,27 @@ async def _delete_function(
     )
 
 
+async def _update_functions_with_deletion_task_ids(
+    db_session, functions, project, background_task_name
+):
+    semaphore = asyncio.Semaphore(
```
liranbg, line 1369: I think it worth adding a log line with how many functions being updated, project name

liranbg (commented):
well done 🎖️ 
minor comment here

server/api/api/utils.py
```diff
@@ -1314,6 +1318,19 @@ async def _delete_function(
...
+        functions_count=len(functions),
+        project=project,
+    )
+
+    # update functions with deletion task id
+    await _update_functions_with_deletion_task_ids(
+        db_session, functions, project, background_task_name
+    )
+
     if len(functions) > 0:
```
TomerShor: can `len(functions)` be Zero?
If so we can skip everything, no?

TomerShor (commented):
Nicely done!

server/api/api/utils.py
rokatyy: @TomerShor technically, yes, it's possible, because user can send a deletion request of the function which doesn't exist

rokatyy: @TomerShor moved function deletion under the condition 👍

TomerShor: Then let's move the `len` check right after getting the functions.
If there are not function - log that there is nothing to do and return

server/api/api/utils.py
```diff
@@ -1318,20 +1318,19 @@ async def _delete_function(
         project,
         function_name,
     )
+    if len(functions) > 0:
+        logger.debug(
+            "Updating functions with deletion task id",
+            function_name=function_name,
+            functions_count=len(functions),
+            project=project,
+        )
```
```suggestion
    if len(functions) == 0:
        logger.debug(
            "No functions found, skipping deletion", function_name=function_name, project=project
        )
        return
        
    logger.debug(
        "Updating functions with deletion task id",
        function_name=function_name,
        functions_count=len(functions),
        project=project,
    )
    
    ...
```

TomerShor (changes requested):

TomerShor (approved):

liranbg (approved):

PR #5543, opened 2024-05-09
[FrontendSpec] Adding TTLCache to `try_get_grafana_service_url` function

roei3000b:
Adding TTLCache to `try_get_grafana_service_url` function.

https://iguazio.atlassian.net/browse/ML-5874

### DIFF
--- a/server/api/utils/clients/iguazio.py
+++ b/server/api/utils/clients/iguazio.py
@@ -21,10 +21,10 @@ import http
 import json
 import threading
 import typing
-import cachetools
 import urllib.parse
 
 import aiohttp
+import cachetools
 import fastapi
 import humanfriendly
 import igz_mgmt.schemas.manual_events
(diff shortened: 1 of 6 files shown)


server/api/utils/clients/iguazio.py
```diff
@@ -340,6 +341,7 @@ def is_sync(self):
         """
         return True
 
+    @cachetools.cached(cache=cachetools.TTLCache(maxsize=1, ttl=120))
```
liranbg: I assume LRU right? then if we you 100k different sessions - how many records would you save? 
I think the approach you have taken is a plaster. the reason it needs to be cached is because its callee resolves the jupyter url twice which results in 2 api calls to igz.
so the idea is to make a single call to igz and use its results once.

liranbg (changes requested):

server/api/utils/clients/iguazio.py
roei3000b: As we discussed, because we want to cache the result per **session**, and because the ttl is only 2 minutes it's ok with our users scale.

liranbg (approved):

PR #5544, opened 2024-05-09
[Runs] Store artifact URIs instead of full artifacts in body [1.6.x]

alonmr:
https://iguazio.atlassian.net/browse/ML-6397

Artifacts stats and preview can bloat the DB when stored in run body.
Instead we now only store the artifact URIs while in memory the run object still holds the artifacts (client side) to maintain easy access.

In the API we strip the artifacts if given and enrich the artifact URIs when storing / updating the runs.

Example from run yaml:
Screenshot 2024-05-12 at 18 27 17

Example from run output html:
Screenshot 2024-05-12 at 18 28 45

### DIFF
--- a/mlrun/model.py
+++ b/mlrun/model.py
@@ -1073,7 +1073,8 @@ class RunStatus(ModelObj):
         self.ui_url = ui_url
         self.reason = reason
         self.notifications = notifications or {}
-        self.artifact_uris = artifact_uris or []
+        # Artifact key -> URI mapping, since the full artifacts are not stored in the runs DB table
+        self.artifact_uris = artifact_uris or {}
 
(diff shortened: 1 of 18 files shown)


mlrun/model.py
```diff
@@ -1058,6 +1058,7 @@ def __init__(
         ui_url=None,
         reason: str = None,
         notifications: Dict[str, Notification] = None,
+        artifact_uris: list[str] = None,
```
theSaarco: This needs to be a dict now, doesn't it?

theSaarco (changes requested):
Looks good - one small change needed.

theSaarco (approved):
Cool. Looks good.

PR #5545, opened 2024-05-12
[API] Check the artifact kind while attempting to remove artifact data

moranbental:
Data artifacts that are `ModelArtifact`, `DirArtifact`, or `DatasetArtifact` must not be removed because we do not yet support the deletion of artifacts that contain multiple files.

If the `deletion strategy` is `"data-force"` and one of these artifact kinds is attempted to be deleted, the error `MLRunNotImplementServerError` will be raised.

https://iguazio.atlassian.net/browse/ML-6432

### DIFF
--- a/mlrun/errors.py
+++ b/mlrun/errors.py
@@ -183,7 +183,7 @@ class MLRunInternalServerError(MLRunHTTPStatusError):
     error_status_code = HTTPStatus.INTERNAL_SERVER_ERROR.value
 
 
-class MLRunNotImplementServerError(MLRunHTTPStatusError):
+class MLRunNotImplementedServerError(MLRunHTTPStatusError):
     error_status_code = HTTPStatus.NOT_IMPLEMENTED.value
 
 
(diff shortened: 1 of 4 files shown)


server/api/crud/artifacts.py
```diff
@@ -271,6 +271,11 @@ def _delete_artifact_data(
                 producer_id=producer_id,
                 object_uid=object_uid,
             )
+            artifact_kind = artifact["kind"]
+            if artifact_kind in ["model", "dataset", "dir"]:
+                raise mlrun.errors.MLRunNotImplementServerError(
+                    f"Deleting artifact data kind: {artifact_kind} is currently not supported"
+                )
```
theSaarco: Please add a comment explaining why we do this now, and a `TODO` to remove once we add support.

theSaarco (approved):
LGTM.

mlrun/errors.py
```diff
@@ -183,6 +183,10 @@ class MLRunInternalServerError(MLRunHTTPStatusError):
     error_status_code = HTTPStatus.INTERNAL_SERVER_ERROR.value
 
 
+class MLRunNotImplementServerError(MLRunHTTPStatusError):
```
alonmr:
```suggestion
class MLRunNotImplementedServerError(MLRunHTTPStatusError):
```

server/api/crud/artifacts.py
```diff
@@ -288,6 +297,9 @@ def _delete_artifact_data(
...
             ):
-                raise mlrun.errors.MLRunInternalServerError(
-                    "Failed to delete artifact data"
-                ) from exc
+                if isinstance(exc, mlrun.errors.MLRunNotImplementServerError):
+                    raise
+                else:
+                    raise mlrun.errors.MLRunInternalServerError(
+                        "Failed to delete artifact data"
+                    ) from exc
```
alonmr: Just raise the exception you got, what if you get 404 or 400, you wouldn't want to wrap it with internal error

alonmr (changes requested):
Looks good! Small touch ups

alonmr (approved):

PR #5546, opened 2024-05-12
[CLI] Add `pod` to kinds of get helpstring

roei3000b:
We already support mlrun get pod, so here we added it to the helpstring.

https://iguazio.atlassian.net/issues/ML-2015


liranbg (changes requested):
how do you conclude it supports get pods?

PR #5547, opened 2024-05-12
[Alerts] Add cooldown_period to notification config in alert

yanburman:
https://iguazio.atlassian.net/browse/ML-6400

### DIFF
--- a/mlrun/alerts/alert.py
+++ b/mlrun/alerts/alert.py
@@ -16,7 +16,6 @@ from typing import Union
 
 import mlrun
 import mlrun.common.schemas.alert as alert_constants
-from mlrun.common.schemas.notification import Notification
 from mlrun.model import ModelObj
 
 
@@ -43,7 +42,7 @@ class AlertConfig(ModelObj):
         trigger: alert_constants.AlertTrigger = None,
         criteria: alert_constants.AlertCriteria = None,
         reset_policy: alert_constants.ResetPolicy = None,
-        notifications: list[Notification] = None,
+        notifications: list[alert_constants.AlertNotification] = None,
         entities: alert_constants.EventEntities = None,
         id: int = None,
         state: alert_constants.AlertActiveState = None,
@@ -81,8 +80,10 @@ class AlertConfig(ModelObj):
             else self.entities
         )
         data["notifications"] = [
-            notification.dict() if not isinstance(notification, dict) else notification
-            for notification in self.notifications
+            notification_data.dict()
+            if not isinstance(notification_data, dict)
+            else notification_data
+            for notification_data in self.notifications
         ]
         data["trigger"] = (
             self.trigger.dict() if not isinstance(self.trigger, dict) else self.trigger
@@ -101,8 +102,8 @@ class AlertConfig(ModelObj):
         notifications_data = struct.get("notifications")
         if notifications_data:
             notifications_objs = [
-                Notification.parse_obj(notification_data)
-                for notification_data in notifications_data
+                alert_constants.AlertNotification.parse_obj(notification)
+                for notification in notifications_data
             ]
             new_obj.notifications = notifications_objs
 
@@ -113,13 +114,18 @@ class AlertConfig(ModelObj):
 
         return new_obj
 
-    def with_notifications(self, notifications: list[Notification]):
+    def with_notifications(
+        self, notifications: list[alert_constants.AlertNotification]
+    ):
         if not isinstance(notifications, list) or not all(
-            isinstance(item, Notification) for item in notifications
+            isinstance(item, alert_constants.AlertNotification)
+            for item in notifications
         ):
-            raise ValueError("Notifications parameter must be a list of notifications")
-        for notification in notifications:
-            self.notifications.append(notification)
+            raise ValueError(
+                "Notifications parameter must be a list of AlertNotification"
+            )
+        for notification_data in notifications:
+            self.notifications.append(notification_data)
         return self
 
     def with_entities(self, entities: alert_constants.EventEntities):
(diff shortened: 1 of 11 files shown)


mlrun/alerts/alert.py
```diff
@@ -16,7 +16,6 @@
 
 import mlrun
 import mlrun.common.schemas.alert as alert_constants
```
theSaarco: Out of curiosity - why do you call this `alert_constants`? There are the schemas of the alert types, isn't it? Why not call it `alert_types` or `alert_objects` or something like this?
It looks strange when you use `alert_constants.AlertNotification` and I think "but this is not a constant!".

mlrun/common/schemas/alert.py
```diff
@@ -101,6 +101,17 @@ class ResetPolicy(StrEnum):
     AUTO = "auto"
 
 
+class AlertNotification(pydantic.BaseModel):
+    notification: Notification
+    cooldown_period: Annotated[
+        str,
```
theSaarco, line 107: What would be the format of the cooldown period string? The description here doesn't hint on it, and you don't do any validation anywhere on the format. 
If you keep this as string (for example, since date-time objects cannot serialize very well), then I would expect at least some validation and parsing of these strings, and also a documentation about it.

mlrun/common/schemas/alert.py
```diff
@@ -121,10 +132,15 @@ class AlertConfig(pydantic.BaseModel):
...
     reset_policy: ResetPolicy = ResetPolicy.MANUAL
-    notifications: pydantic.conlist(Notification, min_items=1)
+    notifications: pydantic.conlist(AlertNotification, min_items=1)
     state: AlertActiveState = AlertActiveState.INACTIVE
     count: Optional[int] = 0
 
+    def get_notifications(self):
+        return [
+            alert_notification.notification for alert_notification in self.notifications
+        ]
```
theSaarco: Why are you stripping away the cooldown period here? Is this because this method is used by someone expecting to get the raw notification? It would help to add a hint here (`-> list[Notification]`) and also call it differently, since now the object has a `notifications` property and `get_notifications` accessor which return different objects.

server/api/crud/alerts.py
```diff
@@ -266,5 +262,23 @@ def reset_alert(self, session: sqlalchemy.orm.Session, project: str, name: str):
     def _delete_notifications(self, alert: mlrun.common.schemas.AlertConfig):
         for notification in alert.notifications:
             server.api.api.utils.delete_notification_params_secret(
-                alert.project, notification
+                alert.project, notification.notification
+            )
+
+    def _validate_and_mask_notificatsions(self, alert_data):
```
theSaarco:
```suggestion
    def _validate_and_mask_notifications(self, alert_data):
```
typo.

server/api/crud/alerts.py
```diff
@@ -266,5 +262,23 @@ def reset_alert(self, session: sqlalchemy.orm.Session, project: str, name: str):
...
+            )
+        ]
+        cooldowns = [
+            notification.cooldown_period for notification in alert_data.notifications
+        ]
+
+        alert_data.notifications = [
+            mlrun.common.schemas.alert.AlertNotification(
+                **{"cooldown_period": cooldown, "notification": notification}
             )
```
theSaarco:
```suggestion
            mlrun.common.schemas.alert.AlertNotification(
                cooldown_period=cooldown, notification=notification
            )
```
No need to create a dict and expand it here.

server/api/crud/alerts.py
```diff
@@ -266,5 +262,23 @@ def reset_alert(self, session: sqlalchemy.orm.Session, project: str, name: str):
     def _delete_notifications(self, alert: mlrun.common.schemas.AlertConfig):
         for notification in alert.notifications:
             server.api.api.utils.delete_notification_params_secret(
-                alert.project, notification
+                alert.project, notification.notification
+            )
+
+    def _validate_and_mask_notificatsions(self, alert_data):
+        notifications = [
+            mlrun.common.schemas.notification.Notification(**notification.to_dict())
```
theSaarco, line 282: Looking at this PR, I'm starting to wonder if it wasn't better to use the client-side `Notification` object in your structures. You're very busy transforming between those objects all the time going back-and-forth their dict representation.
It's worth considering, maybe will save a lot of this dirty work.

theSaarco (changes requested):
Overall looks good. It's a bit cumbersome with all the transformations and packing/unpacking of objects, but perhaps there's no better way to do it. I wonder if using client-side schemas might help (I'm not sure, maybe worth to think about it).
I do think the `cooldown_period` needs to be better documented, in terms of allowed formats and everything. Or are you planning to add that when the functionality is actually added?

mlrun/common/schemas/alert.py
yanburman: Yes, some code needs the raw notifications

server/api/crud/alerts.py
yanburman: I need the server side notifications to use existing APIs (that are used elsewhere as well)

quaark (approved):

server/api/db/sqldb/db.py
```diff
@@ -4429,6 +4437,17 @@ def _enrich_notification(_notification):
             )
         ]
 
+        cooldowns = [
+            notification.cooldown_period for notification in alert.notifications
+        ]
+
+        alert.notifications = [
+            mlrun.common.schemas.alert.AlertNotification(
+                **{"cooldown_period": cooldown, "notification": notification}
```
```suggestion
                cooldown_period=cooldown, notification=notification
```

theSaarco (approved):
Looks good - a minor suggestion (same suggestion as already done in another place).

server/api/db/sqldb/db.py
yanburman: right. missed it

PR #5558, opened 2024-05-13
[Helpers] Ignore pre-release in `min_iguazio_version`

TomerShor:
Iguazio versions are named `<major>.<minor>.<patch>-bXX-YYYYY`, even when it is a GA release.
According to semver, this version is marked as a "pre-release", although it is not necessarily the case.

Thus, in the `min_iguazio_version` wrapper we can safely ignore the pre-release and build version when comparing versions.

### DIFF
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -1572,13 +1572,19 @@ def validate_component_version_compatibility(
     component_current_version = None
     try:
         if component_name == "iguazio":
-            parsed_current_version = mlrun.mlconf.get_parsed_igz_version()
             component_current_version = mlrun.mlconf.igz_version
+            parsed_current_version = mlrun.mlconf.get_parsed_igz_version()
+
+            # ignore pre-release versions and build metadata, as iguazio version is not semver compliant
+            # (we will always have a "pre-release" version even on GA releases)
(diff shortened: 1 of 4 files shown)


quaark (approved):

mlrun/utils/helpers.py
```diff
@@ -1572,13 +1572,19 @@ def validate_component_version_compatibility(
     component_current_version = None
     try:
         if component_name == "iguazio":
-            parsed_current_version = mlrun.mlconf.get_parsed_igz_version()
             component_current_version = mlrun.mlconf.igz_version
+            parsed_current_version = mlrun.mlconf.get_parsed_igz_version()
+
+            # ignore pre-release versions and build metadata, as iguazio version is not semver compliant
```
liranbg: what is not semver compliant here?

liranbg (changes requested):

mlrun/utils/helpers.py
quaark: @liranbg `3.5.5-b687-202401182200` is smaller than `3.5.5` as `-` denotes pre-release. If we were to use `+` as `3.5.5+b687-202401182200` then the `b687-202401182200` would be semver compliant build metadata. According to https://semver.org/#spec-item-10

PR #5560, opened 2024-05-13
[Runs] Enrich run with artifacts when getting a single run [1.6.x]

alonmr:
When we run a remote job we get the run response from the server. Since we removed the artifacts from the run body, the client will not be able to use the runs' outputs and thus it breaks BC and in particular it breaks pipelines that are dependent on previous runs artifacts.
The fix is solely for getting a single run, there we will enrich it with its artifacts.

Jira - https://iguazio.atlassian.net/browse/ML-6441

### DIFF
--- a/server/api/crud/runs.py
+++ b/server/api/crud/runs.py
@@ -148,11 +148,14 @@ class Runs(
             db_session, uid, project, iter
         )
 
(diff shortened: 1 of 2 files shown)


theSaarco (approved):
LGTM 👍

PR #5573, opened 2024-05-15
[API] Support multiple state filter on runs api

roei3000b:
Before vs pr we can filter runs by only one specific state.
In this pr we introduce option to filter by multiple states.

https://iguazio.atlassian.net/browse/ML-6450

### DIFF
--- a/mlrun/db/base.py
+++ b/mlrun/db/base.py
@@ -17,8 +17,8 @@ from abc import ABC, abstractmethod
 from typing import Optional, Union
 
 import mlrun.alerts
-import mlrun.common.schemas
 import mlrun.common.runtimes.constants
+import mlrun.common.schemas
 import mlrun.model_monitoring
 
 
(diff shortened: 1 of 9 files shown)


liranbg (approved):

alonmr (commented):
Do we want to keep the state param forever? why not deprecate it?
To deprecate you need 2 things:
1. Deprecate it from HTTPRunDB
2. Deprecate the query param - see https://fastapi.tiangolo.com/reference/parameters/#fastapi.Query

liranbg:
> Do we want to keep the state param forever? why not deprecate it? To deprecate you need 2 things:
> 
> 1. Deprecate it from HTTPRunDB
> 2. Deprecate the query param - see https://fastapi.tiangolo.com/reference/parameters/#fastapi.Query

@roei3000b  - this should be a followup pr

PR #5574, opened 2024-05-15
[Nuclio] Fix nuclio-name when mlrun name passed as a string

rokatyy:
Generates proper nuclio name if MLRun function name is given as a string


liranbg (approved):

liranbg (commented):
thinking about it - instead of letting client decide - which could be cumbersome - perhaps we can understand it from the function names (if they have mlrun representation) or just nuclio
I am not sure what flow does this pr try to solve?

PR #5577, opened 2024-05-16
[Config] YAML dump convention

moranbental:
using `yaml.safe_dump` instead of using `yaml.dump`
https://iguazio.atlassian.net/browse/ML-6452


alonmr (approved):
lgtm

PR #5581, opened 2024-05-16
[Model Monitoring] Add invocations metric

gtopper:
[ML-6349](https://iguazio.atlassian.net/browse/ML-6349)

[ML-6349]: https://iguazio.atlassian.net/browse/ML-6349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/model_monitoring/test_stores/test_v3io.py
```diff
@@ -338,3 +373,27 @@ def test_read_data() -> None:
...
+        project="fictitious-one",
+        endpoint_id="70450e1ef7cc9506d42369aeeb056eaaaa0bb8bd",
+        start="0",
+        end="now",
+        aggregation_window="1m",
+    )
+    assert result.full_name == "fictitious-one.mlrun-infra.result.invocations"
+    assert result.values == [
+        _ResultPoint(
+            timestamp=pd.Timestamp("2024-04-02 18:00:00+0000", tz="UTC"),
```
jond01: Specifying the two: `+0000"`, `tz="UTC"` appears to be superfluous, and confusing as when there's an incompatibility between the two, seems like there's no error and the `tz` argument rules.
Same above and below.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
```diff
@@ -132,3 +133,87 @@ def read_data(
...
+            type=ModelEndpointMonitoringMetricType.METRIC,
+        )
+
+    rows = df.reset_index().to_dict(orient="records")
+    values = []
+    for row in rows:
+        values.append(
+            [
+                row["time"],
+                row["count(latency)"],
```
jond01: Why is this called "count(latency)"?

mlrun/model_monitoring/db/v3io_tsdb_reader.py
```diff
@@ -132,3 +133,87 @@ def read_data(
...
+
+def prediction_metric_for_endpoint(
+    *,
+    project: str,
+    endpoint_id: str,
+) -> Optional[ModelEndpointMonitoringMetric]:
+    predictions = read_predictions(
+        project=project,
+        endpoint_id=endpoint_id,
+        limit=1,
```
jond01: The meaning of `limit=1` is unclear, please add a comment or in the docstrings.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
```diff
@@ -132,3 +133,87 @@ def read_data(
...
+    )
+    if predictions:
+        return ModelEndpointMonitoringMetric(
+            project=project,
+            app="mlrun-infra",
+            type=ModelEndpointMonitoringMetricType.METRIC,
+            name="invocations",
```
jond01: These literals should be in `mlrun/common/schemas/model_monitoring/constants.py`.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
```diff
@@ -132,3 +133,87 @@ def read_data(
...
+        )
+    return ModelEndpointMonitoringResultValues(
+        full_name=full_name,
+        type=ModelEndpointMonitoringMetricType.RESULT,
+        result_kind=mm_constants.ResultKindApp.system_performance,
+        values=values,
+    )
+
+
+def prediction_metric_for_endpoint(
```
jond01: I think that `check_invocations_metric_existence` is more comprehensible.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
```diff
@@ -132,3 +133,87 @@ def read_data(
...
+
+    if df.empty:
+        return ModelEndpointMonitoringResultNoData(
+            full_name=full_name,
+            type=ModelEndpointMonitoringMetricType.METRIC,
+        )
+
+    rows = df.reset_index().to_dict(orient="records")
+    values = []
+    for row in rows:
```
jond01: I believe it can be done without a for loop in a vectorized way.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
```diff
@@ -132,3 +133,87 @@ def read_data(
...
+        frames_client_kwargs["limit"] = limit
+    df: pd.DataFrame = client.read(
+        backend=_TSDB_BE,
+        table=f"pipelines/{project}/model-endpoints/predictions",
+        filter=f"endpoint_id=='{endpoint_id}'",
+        start=start,
+        end=end,
+        aggregators="count",
+        **frames_client_kwargs,
+    )
```
jond01: Why not request only the relevant columns with the `columns` parameter?

tests/system/model_monitoring/test_app.py
```diff
@@ -148,6 +148,14 @@ def _test_tsdb_record(cls, ep_id: str) -> None:
                     set(tsdb_metrics[app_name]) == app_metrics
                 ), "The TSDB saved metrics are different than expected"
 
+        predictions_df: pd.DataFrame = cls._tsdb_storage.get_records(
+            table=mm_constants.FileTargetKind.PREDICTIONS,
+        )
+        assert not predictions_df.empty, "No TSDB predictions data"
+        assert (
+            predictions_df.endpoint_id == ep_id
+        ).all(), "The endpoint IDs are different than expected"
```
jond01: This should be in another sub-test method, as it checks a non-app metric.

server/api/api/endpoints/model_endpoints.py
```diff
@@ -389,21 +403,23 @@ async def _get_metrics_values_data(
     ],
     start: Optional[datetime] = None,
     end: Optional[datetime] = None,
+    aggregation_window: Optional[str] = None,
```
jond01: Don't expose this in the API, as
1. We don't implement it for other metrics/results.
2. It's not a part of the agreed API.

server/api/api/endpoints/model_endpoints.py
```diff
@@ -389,21 +403,23 @@ async def _get_metrics_values_data(
...
+    :param project:            The name of the project.
+    :param endpoint_id:        The unique id of the model endpoint.
+    :param name:               The full names of the requested results. At least one is required.
+    :param start:              Start and end times are optional, and must be timezone aware.
+    :param end:                See the `start` parameter.
+    :param aggregation_window: Time window to group the results by.
+    :param auth_info:          The auth info of the request.
 
-    :return:            _MetricsValuesData object with the validated data.
+    :return: _MetricsValuesData object with the validated data.
```
jond01:
```suggestion
    :return: _MetricsValuesParams object with the validated data.
```

server/api/api/endpoints/model_endpoints.py
```diff
@@ -461,11 +478,30 @@ async def get_model_endpoint_monitoring_metrics_values(
...
+
+    if any([metric.name == "invocations" for metric in params.metrics]):
+        data, predictions = asyncio.gather(
+            read_data_coroutine,
+            run_in_threadpool(
+                mlrun.model_monitoring.db.v3io_tsdb_reader.read_predictions,
+                project=params.project,
+                endpoint_id=params.endpoint_id,
+                start=str(params.start.timestamp()),
+                end=str(params.end.timestamp()),
```
jond01: Why? I think it's unnecessary.

server/api/api/endpoints/model_endpoints.py
```diff
@@ -461,11 +478,30 @@ async def get_model_endpoint_monitoring_metrics_values(
...
+                aggregation_window=params.aggregation_window or "1m",
```
jond01: This constant should be more visible. Please add a constant for it in this module.

jond01 (commented):
Nice work :)

mlrun/model_monitoring/db/v3io_tsdb_reader.py
gtopper: It's what tsdb calls it when aggregating over a column. In this case, the only column saved to the table is latency, so it's the only one we can count.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
gtopper: But it actually returns a `ModelEndpointMonitoringMetric`. The name you suggest implies that the return type is boolean.

server/api/api/endpoints/model_endpoints.py
gtopper: What do you mean? We want the values within the specified time range.

server/api/api/endpoints/model_endpoints.py
gtopper: Alright.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
jond01: I see. So there's no actual "count" column? I think it's worth a comment.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
jond01: Right, but if you think about it - the nature of this function is boolean, and it maps false to [] and true to a constant object.
Maybe `get_invocations_metric_if_exists`?

mlrun/model_monitoring/db/v3io_tsdb_reader.py
gtopper: Right, there is no count column.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
gtopper: I called it `read_prediction_metric_for_endpoint_if_exists()`. Hopefully that works...

mlrun/common/schemas/model_monitoring/constants.py
```diff
@@ -234,6 +234,7 @@ class MonitoringTSDBTables(MonitoringStrEnum):
     APP_RESULTS = "app-results"
     METRICS = "metrics"
     EVENTS = "events"
+    INVOCATIONS = "invocations"
```
jond01: It doesn't belong to `MonitoringTSDBTables`.

Add it to a new place, including the `"mlrun-infra"` app name.

server/api/api/endpoints/model_endpoints.py
```diff
@@ -461,11 +475,36 @@ async def get_model_endpoint_monitoring_metrics_values(
...
         endpoint_id=params.endpoint_id,
         metrics=params.metrics,
         start=params.start,
         end=params.end,
     )
+
+    if any(
+        [
+            metric.name
+            == mlrun.common.schemas.model_monitoring.ModelMonitoringStoreKinds.EVENTS
```
jond01: It's not the proper condition and not the correct name... It must match the returned FQN.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
```diff
@@ -132,3 +133,91 @@ def read_data(
...
+        return ModelEndpointMonitoringMetric(
+            project=project,
+            app="mlrun-infra",
+            type=ModelEndpointMonitoringMetricType.METRIC,
+            name=mlrun.common.schemas.model_monitoring.MonitoringTSDBTables.INVOCATIONS,
+            full_name=mlrun.common.schemas.model_monitoring.model_endpoints._compose_full_name(
+                project=project,
+                app="mlrun-infra",
+                name=mlrun.common.schemas.model_monitoring.MonitoringTSDBTables.INVOCATIONS,
+            ),
```
jond01: Here you have the FQN, but it's also somewhat wrong:
1. It should be a metric (through the `type` parameter). Ok to keep for now, I'll change it later.
2. The app should be saved as a constant, see the previous comment.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
```diff
@@ -132,3 +133,91 @@ def read_data(
...
+
+    rows = df.reset_index().to_dict(orient="records")
+    values = [
+        [
+            row["time"],
+            row["count(latency)"],  # event count for the time window
+            mm_constants.ResultStatusApp.irrelevant,
+        ]
+        for row in rows
+    ]
```
jond01: This is not vectorized

jond01 (changes requested):
Please address the FQN comment in the `/metrics-values` endpoint.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
jond01: It's better, but instead of describing the function internals, I think the name should reflect its intent

tests/system/model_monitoring/test_app.py
```diff
@@ -178,6 +187,7 @@ def _test_v3io_records(
         cls._test_apps_parquet(ep_id, inputs, outputs)
         cls._test_kv_record(ep_id)
         cls._test_tsdb_record(ep_id)
+        cls._test_predictions_table(ep_id)
 
     @classmethod
     def _test_api_get_results(
```
jond01: I would also like to see tests for this metric in the `_test_api` method.
It's not covered despite the central role it has

mlrun/model_monitoring/db/v3io_tsdb_reader.py
gtopper: Whether a function performs I/O or not is important information. Please reach out if you want to discuss this further.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
gtopper: Please clarify the suggested change.

tests/system/model_monitoring/test_app.py
gtopper: Still working on it, due to the long testing cycle.

mlrun/common/schemas/model_monitoring/constants.py
gtopper: I created a constant for `"mlrun-infra"`, but I don't know where you'd like to move `INVOCATIONS` (it's a TSDB table for monitoring, so I'm not sure why it wouldn't belong in `MonitoringTSDBTables`).

mlrun/model_monitoring/db/v3io_tsdb_reader.py
```diff
@@ -250,3 +253,83 @@ def df_to_metrics_values(
...
+        project=project,
+        endpoint_id=endpoint_id,
+        start="0",
+        end="now",
+        limit=1,  # Read just one record, because we just want to check if there is any data for this endpoint_id
+    )
+    if predictions:
+        return ModelEndpointMonitoringMetric(
+            project=project,
+            app="mlrun-infra",
```
jond01: Use the constant here

mlrun/common/schemas/model_monitoring/constants.py
jond01: > I created a constant for `"mlrun-infra"`,

Ok, please use it in other places (see above)

> but I don't know where you'd like to move `INVOCATIONS` (it's a TSDB table for monitoring, so I'm not sure why it wouldn't belong in `MonitoringTSDBTables`).

The table you use is:
```py
f"pipelines/{project}/model-endpoints/predictions"
```
"invocations" is just the metric name.

server/api/api/endpoints/model_endpoints.py
```diff
@@ -14,6 +14,7 @@
 
 import asyncio
 import json
+import typing
```
jond01: It can be nicer to use `from typing import ...` and combine with this line below

server/api/api/endpoints/model_endpoints.py
```diff
@@ -475,25 +494,45 @@ async def get_model_endpoint_monitoring_metrics_values(
 
     :returns:      A list of the results values for this model endpoint.
     """
-    tasks: list[asyncio.Task] = []
+    coroutines: list[typing.Coroutine] = []
```
jond01: Do you have any reason for this change? I believe it's ok, but tasks may be somewhat faster as they start upon creation

tests/system/model_monitoring/test_app.py
```diff
@@ -241,14 +270,38 @@ def _test_api_get_values(
...
                 "data"
             ], f"No data for result {result_values['full_name']}"
             assert result_values[
                 "values"
             ], f"The values list is empty for result {result_values['full_name']}"
 
+        result = results[0]
+        assert (
+            result["full_name"]
+            == "test-app-flow-v2.histogram-data-drift.result.general_drift"
```
jond01: This for sure need not be here. You hard code the project name, but this test is shared between different system tests, and one may choose a different project name.

tests/system/model_monitoring/test_app.py
```diff
@@ -241,14 +270,38 @@ def _test_api_get_values(
...
+
+        result = results[1]
+
+        assert result["type"] == "metric"
+        assert result["data"], f"No data for result {result['full_name']}"
+        assert result[
+            "values"
+        ], f"The values list is empty for result {result['full_name']}"
+
+        assert len(results) == 5, "Expected 5 results from metric-values API"
```
jond01: Also that, too specific and not generally correct,

tests/system/model_monitoring/test_app.py
```diff
@@ -241,14 +270,38 @@ def _test_api_get_values(
...
+            == "test-app-flow-v2.histogram-data-drift.result.general_drift"
+        )
+        assert result["type"] == "result"
+        assert result["data"], f"No data for result {result['full_name']}"
+        assert result["result_kind"] == 0
+        assert result[
+            "values"
+        ], f"The values list is empty for result {result['full_name']}"
+
+        result = results[1]
```
jond01: The results return order is not guaranteed.

tests/system/model_monitoring/test_app.py
```diff
@@ -219,12 +231,29 @@ def _test_api_get_metrics(
...
+
+        invocations_in_response = {
+            "project": cls.project_name,
+            "app": "mlrun-infra",
+            "type": "metric",
+            "name": "invocations",
+            "full_name": f"{cls.project_name}.mlrun-infra.metric.invocations",
+        } in parsed_response
+
+        assert invocations_in_response == (type == "metrics")
```
jond01:
```suggestion
    if type == "metrics":
        assert {
            "project": cls.project_name,
            "app": "mlrun-infra",
            "type": "metric",
            "name": "invocations",
            "full_name": f"{cls.project_name}.mlrun-infra.metric.invocations",
        } in parsed_response
```

tests/system/model_monitoring/test_app.py
```diff
@@ -219,12 +231,29 @@ def _test_api_get_metrics(
...
+
+        for result in parsed_response:
+            if result["app"] in [app_data.class_.NAME, "mlrun-infra"]:
```
jond01, line 273: On second thought, this can be implemented more elegantly if you pass another `app_data` to `_test_api`.

server/api/api/endpoints/model_endpoints.py
gtopper: Alright, but `make fmt` changes that to `from collections.abc import Coroutine`.

jond01: https://docs.python.org/3/library/typing.html#typing.Coroutine

> Deprecated since version 3.9: [collections.abc.Coroutine](https://docs.python.org/3/library/collections.abc.html#collections.abc.Coroutine) now supports subscripting ([]). See [PEP 585](https://peps.python.org/pep-0585/) and [Generic Alias Type](https://docs.python.org/3/library/stdtypes.html#types-genericalias).

server/api/api/endpoints/model_endpoints.py
gtopper: [Yup](https://docs.micropython.org/en/latest/library/asyncio.html#asyncio.Task),
> Tasks should not be created directly, rather use create_task to create them.

Could have changed it to `create_task()`, but `asyncio.gather()` does the same thing as the pattern that was there, but is more idiomatic.

It's not correct that tasks start on creation btw. They can only start once the function that created them `await`s or yields in some other way (e.g. with `asyncio.sleep()`).

i.e. if tasks are created in an infinite loop, none of them will ever run.

tests/system/model_monitoring/test_app.py
gtopper: I changed to `cls.project_name`, but it sounds like you would like this assert removed entirely?

gtopper: I'll just delete all the added checks, which I think is what you're asking for.

tests/system/model_monitoring/test_app.py
gtopper: That would remove the check that invocations did not somehow show up in the results.

server/api/api/endpoints/model_endpoints.py
jond01: Ok

tests/system/model_monitoring/test_app.py
gtopper: I'm not sure what exactly you have in mind here, but at this stage I prefer not to refactor the test for elegance. In this PR anyway.

tests/system/model_monitoring/test_app.py
jond01: It's checked below in
```py
assert get_app_results == getattr(app_data, type)
```

server/api/api/endpoints/model_endpoints.py
gtopper: Ok, but then I need to change it back, because ruff insists on changing `from typing import Coroutine` to this.

jond01:
```py
from collections.abc import Coroutine
```
Should be fine...
`typing.Coroutine` is deprecated, as I quoted above.

jond01 (approved):

PR #5582, opened 2024-05-16
[Datastore] Fix delete method in filestore and s3

tomerm-iguazio:
1) Fix s3 path issue - in custom `rm` implementation.
2) Fix the usage of `file://` schema.
3) Add tests to each fix.

[ML-6393](https://iguazio.atlassian.net/browse/ML-6393)

[ML-6393]: https://iguazio.atlassian.net/browse/ML-6393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/datastore/datastore.py
+++ b/mlrun/datastore/datastore.py
@@ -223,12 +223,9 @@ class StoreManager:
             subpath = url[len("memory://") :]
             return in_memory_store, subpath, url
 
-        elif schema == "":
+        elif schema in get_local_file_schema():
             # parse_url() will drop the windows drive-letter from the path for url like "c:\a\b".
             # As a workaround, we set subpath to the url.
-            subpath = url
-
-        elif schema == "file://":
             subpath = url.replace("file://", "", 1)
 
         if not schema and endpoint:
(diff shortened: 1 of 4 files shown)


liranbg (approved):

mlrun/datastore/datastore.py
```diff
@@ -242,7 +242,9 @@ def get_or_create_store(
         if not secrets and not mlrun.config.is_running_as_api():
             self._stores[store_key] = store
         # in file stores in windows path like c:\a\b the drive letter is dropped from the path, so we return the url
-        return store, url if store.kind == "file" else subpath, url
+        if store.kind == "file" and isinstance(url, str):
+            subpath = url.replace("file://", "", 1)
```
assaf758: This fix is needed for when there is no schema in the URL (file://c:\a\b does not seem to work in windows)
I suggest you move the special-handling code above, and 
```
        if schema == "memory":
            subpath = url[len("memory://") :]
            return in_memory_store, subpath, url
        elif schema == "":
            # parse_url() will drop the windows drive-letter from the path for url like "c:\a\b".
            # As a workaround, we set subpath to the url. 
            subpath = url
```

assaf758 (approved):

assaf758 (approved):

PR #5583, opened 2024-05-16
[Notifications & KFP] Move Workflow Step Analysis to Pipelines-Adapters 

quaark:
- Move all the kfp1.8 specific code to be part of pipelines-adapters so that we can add the functionality to kfp2 in the future.
- Display Skipped steps in pipeline notification as well:
Screenshot 2024-05-20 at 10 42 42


mlrun/utils/notifications/notification_pusher.py
```diff
@@ -392,17 +392,27 @@ def get_workflow_steps(self, run: mlrun.model.RunObject) -> list:
         steps = []
         db = mlrun.get_run_db()
 
-        def _add_run_step(_node_name, _node_template, _step_kind):
-            _run = db.list_runs(
-                project=run.metadata.project,
-                labels=f"mlrun/runner-pod={_node_name}",
-            )[0]
-            _run["step_kind"] = _step_kind
+        def _add_run_step(_step):
```
alonmr: please add type hints

mlrun/utils/notifications/notification_pusher.py
```diff
@@ -392,17 +392,27 @@ def get_workflow_steps(self, run: mlrun.model.RunObject) -> list:
...
+            except IndexError:
+                _run = {
+                    "metadata": {
+                        "name": _step.display_name,
+                        "project": run.metadata.project,
+                    },
+                }
+            _run["step_kind"] = _step.step_type
+            if _step.skipped:
+                _run.setdefault("status", {})["state"] = "skipped"
```
alonmr: Use `mlrun.common.runtimes.constants.RunStates`. Looks like you need to add skipped there.
Also we might want to check with UI how they show skipped state. I'm not sure if it's something we already have or new here.

mlrun/utils/notifications/notification_pusher.py
```diff
@@ -392,17 +392,27 @@ def get_workflow_steps(self, run: mlrun.model.RunObject) -> list:
...
             steps.append(_run)
 
-        def _add_deploy_function_step(_, _node_template, _step_kind):
+        def _add_deploy_function_step(_step):
```
alonmr: type hints :)

mlrun/utils/notifications/notification_pusher.py
```diff
@@ -419,16 +429,18 @@ def _add_deploy_function_step(_, _node_template, _step_kind):
                             "hash_key": hash_key,
                         },
                     }
-                function["status"] = {
-                    "state": mlrun.common.runtimes.constants.PodPhases.pod_phase_to_run_state(
-                        node["phase"]
-                    ),
-                }
+                if _step.skipped:
+                    state = "skipped"
```
alonmr: can use `mlrun.common.schemas.function.FunctionState` ?
Same comment as above regarding UI

alonmr (commented):
LGTM some small comments though

laurybueno (approved):
Nice work!

mlrun/common/runtimes/constants.py
```diff
@@ -26,12 +26,14 @@ class PodPhases:
     pending = "Pending"
     running = "Running"
     unknown = "Unknown"
+    skipped = "Skipped"
```
alonmr: I don't think you need a pod phase because if the run is skipped than there is no pod.
The PodPhases enum is for k8s states https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase
I think you should use `RunStates` for _add_run_step and `FunctionStates` for  _add_deploy_function_step

alonmr (approved):
👑

PR #5584, opened 2024-05-16
[API] Fix session usage when deleting functions

rokatyy:
Since ORM is not async, we can't use the same db session to perform parallel actions

### DIFF
--- a/server/api/api/utils.py
+++ b/server/api/api/utils.py
@@ -1295,8 +1295,8 @@ async def _delete_function(
     )
 
     # update functions with deletion task id
-    await _update_functions_with_deletion_task_ids(
-        functions, project, background_task_name
+    await _update_functions_with_deletion_info(
+        functions, project, {"status.deletion_task_id": background_task_name}
     )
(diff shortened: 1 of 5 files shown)


server/api/api/utils.py
```diff
@@ -1327,27 +1333,6 @@ async def _delete_function(
...
-
-    async def update_function_with_task_id(function):
-        async with semaphore:
-            await run_in_threadpool(
-                server.api.crud.Functions().set_function_deletion_task_id,
-                db_session,
-                function,
-                project,
-                background_task_name,
-            )
```
alonmr: you can keep this but use `server.api.db.session.run_function_with_new_db_session` instead.

alonmr: see `server.api.crud.runs.Runs.delete_runs` for reference

rokatyy: @alonmr thanks for saying! Reworked it.

liranbg (approved):

PR #5585, opened 2024-05-16
[Model Monitoring] Add application metrics to V3IO tables and endpoints

jond01:
Implements [ML-6430](https://iguazio.atlassian.net/browse/ML-6430).

Write the application metrics to V3IO KV and TSDB.
The KV store is per model endpoint, as done for results. Each key is per application + metric name, creating a schema that Grafana can consume.
The TSDB is implemented in its destined "metrics" table.

Support reading the metrics through the existing `/metrics` and `/metrics-values` API endpoints.

The testing is through some new unit tests and an extended test part in the primary system test.

[ML-6430]: https://iguazio.atlassian.net/browse/ML-6430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/common/schemas/model_monitoring/grafana.py
+++ b/mlrun/common/schemas/model_monitoring/grafana.py
@@ -11,11 +11,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-#
 
 from typing import Optional, Union
 
 from pydantic import BaseModel
+
+import mlrun.common.types
+
+
+class GrafanaColumnType(mlrun.common.types.StrEnum):
+    NUMBER = "number"
+    STRING = "string"
 
 
 class GrafanaColumn(BaseModel):
@@ -24,13 +30,11 @@ class GrafanaColumn(BaseModel):
 
 
 class GrafanaNumberColumn(GrafanaColumn):
-    text: str
-    type: str = "number"
+    type: str = GrafanaColumnType.NUMBER
 
 
 class GrafanaStringColumn(GrafanaColumn):
-    text: str
-    type: str = "string"
+    type: str = GrafanaColumnType.STRING
 
 
 class GrafanaTable(BaseModel):
--- a/mlrun/common/schemas/model_monitoring/model_endpoints.py
+++ b/mlrun/common/schemas/model_monitoring/model_endpoints.py
@@ -298,6 +298,7 @@ class ModelEndpointList(BaseModel):
 
 class ModelEndpointMonitoringMetricType(mlrun.common.types.StrEnum):
     RESULT = "result"
+    METRIC = "metric"
 
 
 class ModelEndpointMonitoringMetric(BaseModel):
@@ -322,7 +323,7 @@ _FQN_PART_PATTERN = r"[a-zA-Z0-9_-]+"
 _FQN_PATTERN = (
     rf"^(?P<project>{_FQN_PART_PATTERN})\."
     rf"(?P<app>{_FQN_PART_PATTERN})\."
-    rf"(?P<type>{_FQN_PART_PATTERN})\."
+    rf"(?P<type>{ModelEndpointMonitoringMetricType.RESULT}|{ModelEndpointMonitoringMetricType.METRIC})\."
     rf"(?P<name>{_FQN_PART_PATTERN})$"
 )
 _FQN_REGEX = re.compile(_FQN_PATTERN)
--- a/mlrun/common/schemas/model_monitoring/model_endpoints.py
+++ b/mlrun/common/schemas/model_monitoring/model_endpoints.py
@@ -338,27 +338,37 @@ def _parse_metric_fqn_to_monitoring_metric(fqn: str) -> ModelEndpointMonitoringM
     )
 
 
+class _MetricPoint(NamedTuple):
+    timestamp: datetime
+    value: float
+
+
 class _ResultPoint(NamedTuple):
     timestamp: datetime
     value: float
     status: ResultStatusApp
 
 
-class _ModelEndpointMonitoringResultValuesBase(BaseModel):
+class _ModelEndpointMonitoringMetricValuesBase(BaseModel):
     full_name: str
     type: ModelEndpointMonitoringMetricType
     data: bool
 
 
-class ModelEndpointMonitoringResultValues(_ModelEndpointMonitoringResultValuesBase):
-    full_name: str
-    type: ModelEndpointMonitoringMetricType
+class ModelEndpointMonitoringMetricValues(_ModelEndpointMonitoringMetricValuesBase):
+    type: ModelEndpointMonitoringMetricType = ModelEndpointMonitoringMetricType.METRIC
+    values: list[_MetricPoint]
+    data: bool = True
+
+
+class ModelEndpointMonitoringResultValues(_ModelEndpointMonitoringMetricValuesBase):
+    type: ModelEndpointMonitoringMetricType = ModelEndpointMonitoringMetricType.RESULT
     result_kind: ResultKindApp
     values: list[_ResultPoint]
     data: bool = True
 
 
-class ModelEndpointMonitoringResultNoData(_ModelEndpointMonitoringResultValuesBase):
+class ModelEndpointMonitoringMetricNoData(_ModelEndpointMonitoringMetricValuesBase):
     full_name: str
     type: ModelEndpointMonitoringMetricType
     data: bool = False
--- a/mlrun/model_monitoring/db/stores/base/store.py
+++ b/mlrun/model_monitoring/db/stores/base/store.py
@@ -116,7 +116,7 @@ class StoreBase(ABC):
         self,
         event: dict[str, typing.Any],
         kind: mm_constants.WriterEventKind = mm_constants.WriterEventKind.RESULT,
-    ):
+    ) -> None:
         """
         Write a new event in the target table.
 
--- a/mlrun/model_monitoring/db/stores/base/store.py
+++ b/mlrun/model_monitoring/db/stores/base/store.py
@@ -15,7 +15,7 @@
 import typing
 from abc import ABC, abstractmethod
 
-import mlrun.common.schemas.model_monitoring.constants as mm_constants
+import mlrun.common.schemas.model_monitoring as mm_schemas
 
 
 class StoreBase(ABC):
@@ -115,7 +115,7 @@ class StoreBase(ABC):
     def write_application_event(
         self,
         event: dict[str, typing.Any],
-        kind: mm_constants.WriterEventKind = mm_constants.WriterEventKind.RESULT,
+        kind: mm_schemas.WriterEventKind = mm_schemas.WriterEventKind.RESULT,
     ) -> None:
         """
         Write a new event in the target table.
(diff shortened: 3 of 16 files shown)


mlrun/common/schemas/model_monitoring/grafana.py
```diff
@@ -11,26 +11,30 @@
...
 class GrafanaNumberColumn(GrafanaColumn):
-    text: str
-    type: str = "number"
+    type: str = GrafanaColumnType.NUMBER
 
 
 class GrafanaStringColumn(GrafanaColumn):
-    text: str
-    type: str = "string"
+    type: str = GrafanaColumnType.STRING
```
Eyal-Danieli, line 37: I'm fine with that refactoring (though I would prefer to do that in a different PR) but just make sure that nothing breaks in Grafana.

mlrun/common/schemas/model_monitoring/model_endpoints.py
```diff
@@ -337,27 +338,37 @@ def _parse_metric_fqn_to_monitoring_metric(fqn: str) -> ModelEndpointMonitoringM
...
+
+class ModelEndpointMonitoringResultValues(_ModelEndpointMonitoringMetricValuesBase):
+    type: ModelEndpointMonitoringMetricType = ModelEndpointMonitoringMetricType.RESULT
     result_kind: ResultKindApp
     values: list[_ResultPoint]
     data: bool = True
 
 
-class ModelEndpointMonitoringResultNoData(_ModelEndpointMonitoringResultValuesBase):
+class ModelEndpointMonitoringMetricNoData(_ModelEndpointMonitoringMetricValuesBase):
```
Eyal-Danieli, line 371: what is it `ModelEndpointMonitoringMetricNoData`, in which use cases it's relevant?

mlrun/model_monitoring/db/stores/base/store.py
```diff
@@ -157,3 +157,16 @@ def update_last_analyzed(
 
         """
         pass
+
+    @abstractmethod
+    def get_model_endpoint_metrics(
```
Eyal-Danieli, line 162: nice, I guess that we will need to support more filters and not just `endpoint_id` which basically can also be an optional filter along side other filters. Anyway, let's keep it as you did for now because it's enough for the current requirement.

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -15,23 +15,79 @@
...
 
 # Fields to encode before storing in the KV table or to decode after retrieving
 fields_to_encode_decode = [
-    mm_constants.EventFieldType.FEATURE_STATS,
-    mm_constants.EventFieldType.CURRENT_STATS,
+    mm_schemas.EventFieldType.FEATURE_STATS,
+    mm_schemas.EventFieldType.CURRENT_STATS,
 ]
 
+_METRIC_FIELDS: list[str] = [
```
Eyal-Danieli, line 37: I would place it in `constants.py` because anyway we are going to use this schema for other DBs as well (such as SQL/TDENGINE)

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -268,11 +324,19 @@ def delete_model_endpoints_resources(self):
                 raise_for_status=v3io.dataplane.RaiseForStatus.never,
             )
 
+    @staticmethod
+    def _get_results_table_path(endpoint_id: str) -> str:
+        return endpoint_id
```
Eyal-Danieli, line 329: lol

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -281,66 +345,63 @@ def write_application_event(
...
-        metric_name = event.pop(mm_constants.ResultData.RESULT_NAME)
-        attributes = {metric_name: json.dumps(event)}
 
-        v3io_monitoring_apps_container = self.get_v3io_monitoring_apps_container(
-            project_name=self.project
-        )
+        container = self.get_v3io_monitoring_apps_container(project_name=self.project)
+        endpoint_id = event.pop(mm_schemas.WriterEvent.ENDPOINT_ID)
+
+        if kind == mm_schemas.WriterEventKind.METRIC:
```
Eyal-Danieli, line 352: just a recommendation, I think it's better to wrap all of this if clause in a different function and keep the `write_application_event` more "clean". I mean, adding something like
`table_path, key, attributes = self._generate_application_event_attributes(endpoint_id, event, kind)`...

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -112,25 +112,24 @@ def _init_tables_path(self):
...
+            mm_schemas.MonitoringTSDBTables.METRICS,
         ]
-        for table in application_tables:
-            logger.info("Creating table in V3IO TSDB", table=table)
+        for table_name in application_tables:
+            logger.info("Creating table in V3IO TSDB", table_name=table_name)
+            table = self.tables[table_name]
             self._frames_client.create(
                 backend=_TSDB_BE,
-                table=self.tables[table],
```
Eyal-Danieli, line 133: Can you explain why is better to pass a different "new" variable in this case?

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -248,45 +247,44 @@ def apply_storey_filter():
...
+        index_cols_base = [
+            mm_schemas.WriterEvent.END_INFER_TIME,
+            mm_schemas.WriterEvent.ENDPOINT_ID,
+            mm_schemas.WriterEvent.APPLICATION_NAME,
+        ]
 
-        if kind == mm_constants.WriterEventKind.METRIC:
-            # TODO : Implement the logic for writing metrics to V3IO TSDB
-            return
+        if kind == mm_schemas.WriterEventKind.METRIC:
```
Eyal-Danieli, line 263: As I wrote above, I would suggest to wrap this part in a different func as well (including the initialization of the `index_cols_base`.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
```diff
@@ -125,10 +179,59 @@ def read_data(
...
+            ModelEndpointMonitoringMetricNoData(
                 full_name=metric.full_name,
                 type=ModelEndpointMonitoringMetricType.RESULT,
             )
         )
 
     return metrics_values
+
+
+def df_to_metrics_values(
```
Eyal-Danieli, line 201: I don't want to delay this PR because of it but in general next time I suggest to add more description. All of the functions in this file are missing documentation. It's more about why and when each function is important and less about what it's doing. As a person who didn't wrote the code it takes me time to understand and explore why each function is needed.

server/api/api/endpoints/model_endpoints.py
```diff
@@ -335,45 +325,64 @@ async def get_model_endpoint(
...
     """
     :param project:     The name of the project.
     :param endpoint_id: The unique id of the model endpoint.
     :param auth_info:   The auth info of the request.
-    :param type:        The type of the metrics to return. Currently, only "results"
-                        is supported.
+    :param type:        The type of the metrics to return. "all" means "results"
+                        and "metrics".
 
     :returns:           A list of the application results for this model endpoint.
```
Eyal-Danieli:
```suggestion
    :returns:           A list of the application metrics or/and results for this model endpoint.
```

Eyal-Danieli (commented):
LGTM, I didn't run the tests.
I had few suggestions but in general 2 important comments:
1 - try to add more documentation. Even 1-2 lines can be super useful to understand the importance of a certain function. 
2 - TBH I still not sure why not implementing these changes on top of the V3IO Connector in this PR. As long as we are using the v3io_tsdb_reader, we have to import v3io resources into the model endpoints API file, which in other words means that it is probably breaks the CE flow + anyway we will have to "rewrite" this logic on top of the abstraction, so why not now?

mlrun/common/schemas/model_monitoring/model_endpoints.py
jond01: Just to note, this object isn't new, it's only the name that changed.
For example:
A result was recorded to the DB (can be V3IO KV) a day ago.
We call the `/metrics` endpoint and receive this result's FQN, say `project1.app1.result.res1`.
We call the `/metrics-values?name=project1.app1.result.res1&start=...&end=...` endpoint with start and end times of the last two hours. There's no data for this metric, but we still return this object to signify we didn't ignore this part.

mlrun/common/schemas/model_monitoring/grafana.py
jond01: I believe that's quite a not risky refactoring as `StrEnum` can replace `str` without needing to check too much.
How would you check it, though?

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
jond01: Why TDEngine? This is only KV here.
Once we support MySQL, we can move it to a shared location. At the moment, I'm not sure about this location.
I'm not sure that the constants module is the best place for this

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
jond01: I agree with you that the implementation can be cleaner.
The main goal here was to do it fast.
How about deferring that? Once we think about all the cases, we may have a more holistic approach, not only for V3IO KV.

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
jond01: It's the same, but I think that the previous version is more confusing compared to
```py
table = self.tables[table_name]
```

server/api/api/endpoints/model_endpoints.py
jond01: Done in 824e83e.

mlrun/model_monitoring/db/v3io_tsdb_reader.py
jond01: Ok. Improved a bit in 1ff9e6c.

jond01:
> 2 - TBH I still not sure why not implementing these changes on top of the V3IO Connector in this PR. As long as we are using the v3io_tsdb_reader, we have to import v3io resources into the model endpoints API file, which in other words means that it is probably breaks the CE flow + anyway we will have to "rewrite" this logic on top of the abstraction, so why not now?

1. It should indeed move there soon. I updated the comment at the top of this module in this PR.
2. In what way do you mean it breaks the CE **at the moment**?
3. Why not now - time and priorities... the top priority now is fully supporting one permutation for our consumers (UI, QA).

mlrun/common/schemas/model_monitoring/grafana.py
Eyal-Danieli: going into grafana and check that everything is available as expected :)

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
Eyal-Danieli: TDEngine supertable is based on schemas of tags and values. Anyway, let's keep it here and if needed will replace it later.

Eyal-Danieli (approved):

mlrun/common/schemas/model_monitoring/grafana.py
jond01: image
Done, thanks for the help!

PR #5588, opened 2024-05-17
[Frameworks] Bump lightgbm

liranbg:
Making MLRun better arm64-compatible.


guy1992l (approved):
I would like to say I'm in favor of removing `lightgbm` from any req file. It is a user framework and only if the user uses it, we should have a "defense mechanism" that will check that `lightgbm` is indeed `~=4.3` in code during runtime.

liranbg:
ill add a ticket @guy1992l for that. in the meantime, any objections to bump req to 4.3?

PR #5590, opened 2024-05-19
[Alerts] Add support for summary message

yanburman:
https://iguazio.atlassian.net/browse/ML-6404
https://iguazio.atlassian.net/browse/ML-6401

### DIFF
--- a/mlrun/common/schemas/alert.py
+++ b/mlrun/common/schemas/alert.py
@@ -123,7 +123,8 @@ class AlertConfig(pydantic.BaseModel):
         pydantic.Field(
             description=(
                 "String to be sent in the notifications generated."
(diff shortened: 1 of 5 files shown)


theSaarco (changes requested):
Looks good, but what about unit tests?
Also, there's another location in `alert.py` where you're using the `$project` and `$entity` notation - should be modified as well.

theSaarco (commented):
ok, but I thought to add a test to verify that the place-holders are indeed being replaced when formatting events. 
You can add a simple unit test that calls `format_alert_summary` directly for now, but we should have a test that the functionality is working as expected. Of course, this should also be verified with more complex tests later.

theSaarco (approved):
Sababa

quaark (approved):

PR #5592, opened 2024-05-19
[Model Monitoring] Add TDEngine Connector 

Eyal-Danieli:
Following #5210 , this PR implements TDEngine connector as a supported TSDB solution for storing model monitoring real time data.

**How to define a new tdengine connection**
To define a new TDEngine connection, the user can pass `tsdb_connection` through `project.set_model_monitoring_credentials(tsdb_connection="taosws://...")`.

Using one of the above suggested approaches, the TDEngine connection will be stored as a k8s secret under the project and the relevant workers (stream, writer). 

**How the TDEngine is implemented**
- The new connector is defined under `mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py`. The main connector object called `TDEngineConnector` and it implements the required functionality as defined in the connector class `TSDBConnector`.
- The implementation is based on the recommended [TDEngine websocket connection](https://docs.tdengine.com/develop/connect/). We used the TDEngine official connector `taosws`. 
- The supertables schemas and the queries are defined under `mlrun/model_monitoring/db/tsdb/tdengine/schemas.py`. Each supertable should include a name, columns, tags, and a database. At the moment we have 3 different tables under the same default database - `mlrun_model_monitoring`. The schema design of each supertable is described below.
- The supertables are created on the 1st api call to establish the model monitoring infrastructure. Note that it only happens once because the supertables are similar accross different projects. 
 
tdengine_predictions_170

tdengine_app_result_170_V2

tdengine_metrics_170_V2

A related JIRA: https://iguazio.atlassian.net/browse/ML-6261


mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
```diff
@@ -0,0 +1,199 @@
...
+        columns_to_filter: list[str] = None,
+        filter_query: str = "",
+        start: str = datetime.datetime.now().astimezone() - datetime.timedelta(hours=1),
+        end: str = datetime.datetime.now().astimezone(),
+        timestamp_column: str = "time",
+        database: str = _MODEL_MONITORING_DATABASE,
+    ) -> str:
+        full_query = "select "
+        if columns_to_filter:
+            full_query += ", ".join(columns_to_filter)
```
jond01: 1. The defaults of the `start` and `end` times are not what you probably expect. They are the "now" time when running this class/signature the first time, which is the import time.
2. If "now" datetime is needed, please use `mlrun.utils.datetime_now()`.
3. Don't use `+=` on strings, as it's not performant.
4. This method can (and should) be unit-tested easily.

dependencies.py
```diff
@@ -76,6 +76,7 @@ def extra_requirements() -> dict[str, list[str]]:
             "distributed~=2023.9.0",
         ],
         "alibaba-oss": ["ossfs==2023.12.0", "oss2==2.18.1"],
+        "taosws": ["taospy[ws]>=2,<3"],
```
jond01:
```suggestion
        "tdengine": ["taospy[ws]~=2.7.15"],
```
If you use only `taosws` package - why not require `taos-ws-py` instead of `taospy[ws]`?
See:
https://github.com/taosdata/taos-connector-python?tab=readme-ov-file#install-taos-ws-py-only

jond01 (dismissed):
Requesting changes because of the confusing defaults

dependencies.py
Eyal-Danieli: great catch :) https://github.com/mlrun/mlrun/pull/5592/commits/3e508dee5f194eec59df7574af8a321ab1ba5db6

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
Eyal-Danieli: As we agreed:
-> default time values were removed
-> the query build is not using `StringIO()` instead of `+=`
-> unittests were added

jond01: > -> default time values were removed

Ok, I dismissed my changes request, but I still see the defaults in this method - so please make sure to remove it.

server/api/crud/model_monitoring/deployment.py
```diff
@@ -586,14 +599,23 @@ def is_monitoring_stream_has_the_new_stream_trigger(self) -> bool:
...
-    def _create_tsdb_application_tables(self) -> None:
-        """
-        Each project writer service writes the application results into a single TSDB
-        table and therefore the target table is created during the writer deployment
-        """
-        mlrun.model_monitoring.get_tsdb_connector(
-            project=self.project,
-        ).create_tsdb_application_tables()
+    @staticmethod
+    def _create_tsdb_tables(project: str):
```
jond01: Passing the project is redundant. It's a part of the object.
See the previous version, where this method is not static

mlrun/projects/project.py
```diff
@@ -3102,14 +3102,15 @@ def set_model_monitoring_credentials(
         access_key: str = None,
         endpoint_store_connection: str = None,
         stream_path: str = None,
+        tsdb_connection: str = None,
```
jond01:
```py
x: str = None
```
This type-hint is inherently inaccurate, as the default value is `None`.
It should be instead:
```py
x: Optional[str] = None
```
This appears in other changes you made, e.g. `apply_monitoring_serving_graph`.

Eyal-Danieli: I'll update it though across mlrun this behaviour (: str = None) repeats itself 100+ times.

Eyal-Danieli: updated for the new changes :)

server/api/crud/model_monitoring/deployment.py
Eyal-Danieli: https://github.com/mlrun/mlrun/pull/5592/commits/f0ea2d602c946f0cd9d765666427072cf4c0aa02

assaf758 (approved):

PR #5593, opened 2024-05-19
[Pipelines] Support watch remote:local engine

alonmr:
Add support to watch a remote pipeline with local inner engine.
The pipeline runs a kubejob runner which runs the pipeline without kfp inside the job (local to the job).
After we watch the job we enrich with the run state and error.

https://iguazio.atlassian.net/browse/ML-5334

tested with `test_remote_pipeline_with_local_engine_from_github`


quaark (approved):
Awesome!

mlrun/projects/pipelines.py
```diff
@@ -742,6 +746,9 @@ class _RemoteRunner(_PipelineRunner):
 
     engine = "remote"
 
+    def __init__(self):
+        self.inner_runner: type[_PipelineRunner] = _KFPRunner
```
TomerShor: Where is this being used?
the inner engine is passed as an argument to `get_run_status`, doesn't look like it should be a property

alonmr: right its a leftover from solution. removed

PR #5595, opened 2024-05-20
[Alerts] Update entity id to take specific id & fix slack job overview

Yacouby:


mlrun/utils/notifications/notification/slack.py
```diff
@@ -144,10 +144,19 @@ def _get_alert_fields(
             data_text = "\n".join(data_lines)
             line.append(self._get_slack_row(f"*Event data:*\n{data_text}"))
 
-        if url := mlrun.utils.helpers.get_ui_url(
-            alert.project, event_data.entity.ids[0]
-        ):
-            line.append(self._get_slack_row(f"*Overview:*\n<{url}|*Job overview*>"))
+        uid = event_data.value_dict.get("uid")
+        if uid:  # JOB entity
```
theSaarco: You should use `event_data.entity.kind` to decide what type of entity it is, don't rely on value dict fields which are not mandatory and can change.

theSaarco (approved):
Looks good, except for a single comment I have.
Please note that this change will break upgrades between RCs (since `AlertTemplate` structure has changed), so developers will need to be aware of the change.

server/api/crud/alerts.py
```diff
@@ -184,6 +184,8 @@ def process_event(
                 obj=state_obj,
                 active=active,
             )
+        else:
+            logger.debug("The entity of the alert does not match the one in event")
```
alonmr, line 188: Is this expected? in your next PR please add some parameters kwargs to log the entities or some other alert/event information

alonmr (approved):

PR #5598, opened 2024-05-20
[NuclioRuntime] Always pass serving graph to pod using config-map

alxtkr77:
https://iguazio.atlassian.net/browse/ML-6223

### DIFF
--- a/server/api/crud/runtimes/nuclio/function.py
+++ b/server/api/crud/runtimes/nuclio/function.py
@@ -216,6 +216,7 @@ def _compile_function_config(
     if serving_spec is not None:
         # To keep backward comatability, allow passing service spec
         # via Config Map only for client version higher then 1.7.0
+        # TODO: remove in 1.9.0.
         can_pass_via_cm = not client_version or (
             semver.Version.parse(client_version) >= semver.Version.parse("1.7.0")
         )
--- a/server/api/crud/runtimes/nuclio/function.py
+++ b/server/api/crud/runtimes/nuclio/function.py
@@ -19,6 +19,7 @@ import shlex
 import nuclio
 import nuclio.utils
 import requests
+import semver
 
 import mlrun
 import mlrun.common.constants
@@ -213,9 +214,17 @@ def _compile_function_config(
     serving_spec_volume = None
     serving_spec = function._get_serving_spec()
     if serving_spec is not None:
+        # To keep backward comatability, allow passing service spec
+        # via Config Map only for client version higher then 1.7.0
+        can_pass_via_cm = not client_version or (
+            semver.Version.parse(client_version) >= semver.Version.parse("1.7.0")
+        )
         # since environment variables have a limited size,
         # large serving specs are stored in config maps that are mounted to the pod
-        if len(serving_spec) >= mlrun.mlconf.httpdb.nuclio.serving_spec_env_cutoff:
+        if (
+            can_pass_via_cm
+            and len(serving_spec) >= mlrun.mlconf.httpdb.nuclio.serving_spec_env_cutoff
+        ):
             function_name = mlrun.runtimes.nuclio.function.get_fullname(
                 function.metadata.name, project, tag
             )
--- a/tests/api/api/test_functions.py
+++ b/tests/api/api/test_functions.py
@@ -481,6 +481,13 @@ def test_tracking_on_serving(
     Validate that `.set_tracking()` configurations are applied to
     a serving function for model monitoring.
     """
+    server.api.utils.singletons.k8s.get_k8s_helper().v1api = unittest.mock.Mock()
+
+    config_map = unittest.mock.Mock()
+    config_map.items = []
+    server.api.utils.singletons.k8s.get_k8s_helper().v1api.list_namespaced_config_map = unittest.mock.Mock(
+        return_value=config_map
+    )
 
     # Generate a test project
     tests.api.api.utils.create_project(client, PROJECT)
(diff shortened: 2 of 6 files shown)


mlrun/utils/helpers.py
```diff
@@ -1550,12 +1550,6 @@ def get_serving_spec():
     except FileNotFoundError:
         pass
 
-    if data is None:
-        data = os.environ.get("SERVING_SPEC_ENV", "")
```
gtopper: Can we keep this fallback? I use it quite extensively for running functions locally, and having to create a file does complicate things a bit, especially when its path is hard coded.

tests/serving/test_serving.py
```diff
@@ -36,6 +37,16 @@
 from mlrun.utils import logger
 
 
+def setUpModule():
+    with tempfile.NamedTemporaryFile(mode="w+", delete=False) as temp_file:
+        mlrun.common.constants.MLRUN_SERVING_SPEC_PATH = temp_file.name
```
gtopper: I wonder if the path should be properly overridable via an environment variable. This avoids the hacks in the tests/benchmark and also can be helpful in debugging.

tests/serving/test_serving.py
```diff
@@ -36,6 +37,16 @@
...
+
+
+# Global teardown function
+def tearDownModule():
```
gtopper: Redundant comment that's inconsistent with `setUpModule()`.
```suggestion
def tearDownModule():
```

gtopper (commented):
Looks good, but I'm wondering if we can keep the flexibility so we're not married to a hard coded path under `/tmp`.

mlrun/utils/helpers.py
alxtkr77: While I understand the reasoning, I personally oppose having two methods for doing the same thing. Regarding testing, you can always override `mlrun.common.constants.MLRUN_SERVING_SPEC_PATH` with a temporary file. I will add your comment to the case and see what others have to say.

gtopper: I don't see a problem with keeping the override, but anyway, see my other comment regarding overwriting the global constant. As far as I can tell, it's not done anywhere else in the code base.

theSaarco (changes requested):
@alxtkr77 - looking at this, I actually think there is an issue here with BC. If the user is running a function that runs with an older mlrun client version that doesn't support passing the spec as CM (i.e. reading it from file), with this change there is no way to make it work except upgrading the client version.
Before making this change constant, most serving graphs would pass as-is in env variable, and the issue may happen only for large serving graphs (which arguable wouldn't work anyway due to their size). Now we're introducing a breaking change that will happen for every serving function deployed using an older client.
We need to give the user an option to keep using env variables for the serving spec using a configuration on the runtime or something similar. Trying to automatically deduce whether the client is old or not is problematic since we don't always know what version of mlrun is included in the image used.

server/api/crud/runtimes/nuclio/function.py
```diff
@@ -212,9 +213,17 @@ def _compile_function_config(
     serving_spec_volume = None
     serving_spec = function._get_serving_spec()
     if serving_spec is not None:
+        # To keep backward comatability, allow passing service spec
+        # via Config Map only for client version higher then 1.7.0
```
gtopper, line 218:
```suggestion
        # via Config Map only for client version 1.7.0 or higher
```

server/api/crud/runtimes/nuclio/function.py
```diff
@@ -213,9 +214,17 @@ def _compile_function_config(
     serving_spec_volume = None
     serving_spec = function._get_serving_spec()
     if serving_spec is not None:
+        # To keep backward comatability, allow passing service spec
+        # via Config Map only for client version 1.7.0 or higher
```
gtopper: I guess this needs a TODO about removal in 1.9.0?

tests/api/runtimes/test_serving.py
```diff
@@ -248,6 +248,8 @@ def _setup_serving_spec_in_config_map(self):
     def test_remote_deploy_with_secrets(self, use_config_map):
         if use_config_map:
             self._setup_serving_spec_in_config_map()
+        else:
+            mlrun.mlconf.httpdb.nuclio.serving_spec_env_cutoff = 4096
```
gtopper, line 252: Isn't this sticky, potentially affecting tests that run after it?

tests/api/api/test_functions.py
```diff
@@ -481,6 +481,13 @@ def test_tracking_on_serving(
     Validate that `.set_tracking()` configurations are applied to
     a serving function for model monitoring.
     """
+    server.api.utils.singletons.k8s.get_k8s_helper().v1api = unittest.mock.Mock()
+
+    config_map = unittest.mock.Mock()
+    config_map.items = []
+    server.api.utils.singletons.k8s.get_k8s_helper().v1api.list_namespaced_config_map = unittest.mock.Mock(
+        return_value=config_map
+    )
```
theSaarco: You need to revert these changes at the end of the test. Since you have `monkeypatch` available, best to use it to perform these patching, as it will take care of reverting automatically.

theSaarco (approved):
LGTM - one small fix needed in the test.

tests/api/runtimes/test_serving.py
alxtkr77: It was set to 0 in _setup_serving_spec_in_config_map() before, so I guess not

server/api/crud/runtimes/nuclio/function.py
alxtkr77: Are we committed for BC only 2 versions back?

gtopper: Yup

gtopper (approved):

PR #5604, opened 2024-05-21
[Runtimes] Reload function code when running relative handler on local mode

moranbental:
When we run a function in local mode, we need to reload the function code again if the code has changed. 
To do this, we create a new argument of `reset_on_run` on `run_function` method

**reloading the module as best effort**
```
%%writefile ./functions/func.py
def func():
    return "dummy value"

project.set_function(name='func',handler='functions.func.func',kind='job',image='mlrun/mlrun')
project.run_function('func',local=True)

%%writefile ./functions/func.py
def func():
    return "dummy value updated"

project.run_function('func',local=True, reset_on_run=True)
```

also adding `reload_max_recursion_depth` to `mlconf` with 100 as the default value

https://iguazio.atlassian.net/browse/ML-3943
https://iguazio.atlassian.net/browse/ML-6514


mlrun/utils/helpers.py
```diff
@@ -1015,7 +1015,16 @@ def create_function(pkg_func: str):
     splits = pkg_func.split(".")
     pkg_module = ".".join(splits[:-1])
     cb_fname = splits[-1]
+    pkg_module_exists_in_modules = pkg_module in sys.modules
     pkg_module = __import__(pkg_module, fromlist=[cb_fname])
+
+    # if the function appears in the modules list, we need to reload the code again because it may have changed
+    if pkg_module_exists_in_modules:
+        logger.warning(
+            "Reloading module - be aware that not all modules can reload again"
```
TomerShor:
```suggestion
            "Reloading module - be aware that not all modules can be reloaded again"
```
Also - what should the user do if the modules aren't reloaded?

mlrun/utils/helpers.py
```diff
@@ -1015,7 +1015,16 @@ def create_function(pkg_func: str):
...
     cb_fname = splits[-1]
+    pkg_module_exists_in_modules = pkg_module in sys.modules
     pkg_module = __import__(pkg_module, fromlist=[cb_fname])
+
+    # if the function appears in the modules list, we need to reload the code again because it may have changed
+    if pkg_module_exists_in_modules:
+        logger.warning(
+            "Reloading module - be aware that not all modules can be reloaded again"
+        )
+        reload(pkg_module)
```
liranbg: wrap in try/except to ensure that it wont fail. reload will recompilepython code and might fail so we want to avoid.
`also, worth doing it *only* when local=True`

mlrun/utils/helpers.py
moranbental: The user should import the modules manually, or if he is using Jupyter, he can restart his Python kernel.
This warning will be fixed by - https://iguazio.atlassian.net/browse/ML-6514

mlrun/utils/helpers.py
```diff
@@ -1015,16 +1015,32 @@ def create_class(pkg_class: str):
     return class_
 
 
-def create_function(pkg_func: str):
+def create_function(pkg_func: str, need_to_reload: bool = False):
```
TomerShor:
```suggestion
def create_function(pkg_func: str, reload: bool = False):
```

mlrun/utils/helpers.py
```diff
@@ -1015,16 +1015,32 @@ def create_class(pkg_class: str):
...
+        # if the function appears in the modules list, we need to reload the code again because it may have changed
+        if pkg_module_exists_in_modules:
+            logger.warning(
+                "Reloading module - be aware that not all modules can be reloaded again"
+            )
+            try:
+                reload(pkg_module)
+            except Exception as exc:
+                logger.debug(
+                    "Failed to reload module", err=mlrun.errors.err_to_str(exc)
```
TomerShor:
```suggestion
                    "Failed to reload module", module=pkg_module, err=mlrun.errors.err_to_str(exc)
```

TomerShor: why is this a debug and no warning too?

mlrun/utils/helpers.py
```diff
@@ -1082,8 +1098,14 @@ def get_class(class_name, namespace=None):
     return class_object
 
 
-def get_function(function, namespace):
-    """return function callable object from function name string"""
+def get_function(function, namespaces, need_to_reload: bool = False):
```
TomerShor:
```suggestion
def get_function(function, namespaces, reload: bool = False):
```

mlrun/utils/helpers.py
```diff
@@ -1015,16 +1015,32 @@ def create_class(pkg_class: str):
     return class_
 
 
-def create_function(pkg_func: str):
+def create_function(pkg_func: str, need_to_reload: bool = False):
     """Create a function from a package.module.function string
 
     :param pkg_func:  full function location,
                       e.g. "sklearn.feature_selection.f_classif"
+    :param need_to_reload: reload the function if it displays in the modules list.
```
TomerShor:
```suggestion
    :param need_to_reload: reload the function if it is in the modules list.
```

mlrun/utils/helpers.py
```diff
@@ -1082,8 +1098,14 @@ def get_class(class_name, namespace=None):
...
 
 
-def get_function(function, namespace):
-    """return function callable object from function name string"""
+def get_function(function, namespaces, need_to_reload: bool = False):
+    """return function callable object from function name string
+
+    :param function: path to the function ([class_name::]function)
+    :param namespaces: one or list of namespaces/modules to search the function in
+    :param need_to_reload: reload the function if it displays in the modules list
```
TomerShor:
```suggestion
    :param need_to_reload: reload the function if it is in the modules list
```

mlrun/utils/helpers.py
```diff
@@ -1106,18 +1128,23 @@ def get_function(function, namespace):
 
 
 def get_handler_extended(
-    handler_path: str, context=None, class_args: dict = {}, namespaces=None
+    handler_path: str,
+    context=None,
+    class_args: dict = {},
+    namespaces=None,
+    need_to_reload: bool = False,
```
TomerShor:
```suggestion
    reload: bool = False,
```

mlrun/utils/helpers.py
```diff
@@ -1106,18 +1128,23 @@ def get_function(function, namespace):
...
 ):
     """get function handler from [class_name::]handler string
 
     :param handler_path:  path to the function ([class_name::]handler)
     :param context:       MLRun function/job client context
     :param class_args:    optional dict of class init kwargs
     :param namespaces:    one or list of namespaces/modules to search the handler in
+    :param need_to_reload: reload the function if it displays in the modules list
```
TomerShor:
```suggestion
    :param need_to_reload: reload the function if it is in the modules list
```

TomerShor (changes requested):
looks good! just some naming changes

mlrun/utils/helpers.py
moranbental: `reload` name is already exists when `from importlib import import_module, reload`

TomerShor: `reload_modules` then?
I just don't like the "need to" part

liranbg (approved):

TomerShor (approved):
🤘🏼

tests/launcher/test_local.py
```diff
@@ -116,3 +118,35 @@ def test_validate_runtime_success():
...
     launcher._validate_runtime(runtime, run)
+
+
+def test_launch_local_reload_module():
+    """This test ensures that the function code is updated when running a relative handler in local
+    mode when the code changes during execution"""
+    sys.path.append(str(pathlib.Path(__file__).parent))
+    file_path = f"{assets_path}/temp_function.py"
+    if os.path.exists(file_path):
+        os.remove(file_path)
```
alonmr: Use a temp file with context manager to ensure no leftover files

mlrun/utils/helpers.py
```diff
@@ -1015,16 +1015,35 @@ def create_class(pkg_class: str):
...
+    pkg_module_exists_in_modules = pkg_module in sys.modules
     pkg_module = __import__(pkg_module, fromlist=[cb_fname])
+
+    if reload_modules:
+        # if the function appears in the modules list, we need to reload the code again because it may have changed
+        if pkg_module_exists_in_modules:
+            logger.warning(
+                "Reloading module. Not all associated modules can be reloaded, import them manually."
+                "Or, with Jupyter, restart the Python kernel."
+            )
```
alonmr: what kind of modules can't be reloaded? feels like some inside knowledge may be worth adding as a comment.

mlrun/utils/helpers.py
```diff
@@ -1082,8 +1101,14 @@ def get_class(class_name, namespace=None):
     return class_object
 
 
-def get_function(function, namespace):
-    """return function callable object from function name string"""
+def get_function(function, namespaces, reload_modules: bool = False):
+    """return function callable object from function name string
```
alonmr: As a general note please use capital letters for comments and docstring even though it is not really enforced yet

mlrun/utils/helpers.py
moranbental: I was added a recursive method, so now it will reload the module recursively and it should fix this issue - https://iguazio.atlassian.net/browse/ML-3943#:~:text=Proposition%3A%C2%A0In,will%20be%20partial.

mlrun/utils/helpers.py
```diff
@@ -1015,16 +1015,33 @@ def create_class(pkg_class: str):
...
     """
     splits = pkg_func.split(".")
     pkg_module = ".".join(splits[:-1])
     cb_fname = splits[-1]
+    pkg_module_exists_in_modules = pkg_module in sys.modules
     pkg_module = __import__(pkg_module, fromlist=[cb_fname])
+
+    if reload_modules:
+        # if the function appears in the modules list, we need to reload the code again because it may have changed
+        if pkg_module_exists_in_modules:
```
liranbg: since we have opt-in `reload_modules` feature. I think we can remove this condition and execute it regardless to whether it apeear on modules or not.

liranbg (commented):
last one

mlrun/utils/helpers.py
```diff
@@ -1628,3 +1655,12 @@ def format_alert_summary(
     result = result.replace("{{name}}", alert.name)
     result = result.replace("{{entity}}", event_data.entity.ids[0])
     return result
+
+
+def _reload(module):
```
quaark: I suggest adding a `max_recursion_depth` parameter here.
I don't think the user should supply it though, so let's put a high enough (but not infinite) threshold in the `mlconf` and use that as the value. Maybe `100`?

quaark (changes requested):
Looks great! 🚀 
A small suggestion

mlrun/projects/operations.py
```diff
@@ -167,6 +168,8 @@ def my_pipe(url=""):
                               artifact type can be given there. The artifact key must appear in the dictionary as
                               "key": "the_key".
     :param builder_env:     env vars dict for source archive config/credentials e.g. builder_env={"GIT_TOKEN": token}
+    :param reset_on_run:    when set to True and local=True it is required to reload the code again because
+                            it may have changed
```
liranbg:
```suggestion
    :param reset_on_run:    When True, function python modules would reload prior to code execution. This ensures latest code changes are executed. This argument must be used in conjunction with the local=True argument.
```

mlrun/projects/project.py
```diff
@@ -3245,6 +3246,8 @@ def run_function(
                                   artifact type can be given there. The artifact key must appear in the dictionary as
                                   "key": "the_key".
         :param builder_env: env vars dict for source archive config/credentials e.g. builder_env={"GIT_TOKEN": token}
+        :param reset_on_run:    when set to True and local=True it is required to reload the code again because
+                                it may have changed
```
liranbg:
```suggestion
    :param reset_on_run:    When True, function python modules would reload prior to code execution. This ensures latest code changes are executed. This argument must be used in conjunction with the local=True argument.
```

mlrun/runtimes/base.py
```diff
@@ -390,6 +392,8 @@ def run(
                 standards and is at least 1 minute (-1 for infinite).
                 If the phase is active for longer than the threshold, the run will be aborted.
                 See mlconf.function.spec.state_thresholds for the state options and default values.
+        :param reset_on_run: when set to True and local=True it is required to reload the code again because
+                             it may have changed
```
liranbg:
```suggestion
    :param reset_on_run:    When True, function python modules would reload prior to code execution. This ensures latest code changes are executed. This argument must be used in conjunction with the local=True argument.
```

liranbg (commented):
comment update

PR #5605, opened 2024-05-21
[Runs] Deprecate `state` arg

roei3000b:
Deprecate `state` field. The user should use `states` instead.


alonmr (commented):
~See if you can also deprecate in the API with https://fastapi.tiangolo.com/reference/parameters/#fastapi.Query
it should be visible in swagger~ discussed offline
Add it to - https://iguazio.atlassian.net/browse/ML-5948
And change the remove version to 1.9

alonmr (approved):

PR #5606, opened 2024-05-21
[Alerts] Rename mm related entity and alerts + add new events

yanburman:
https://iguazio.atlassian.net/browse/ML-6499

Add new events + rename existing ones to make more sense.
Also, change mm entity to be per result and not per entire model

We are breaking backwards compatibility here, but this has not been released yet, so the SDK breakage is only between RCs


assaf758 (approved):
Looks good.
Please just make sure that TestMonitoringAppFlow pass.
I assume the failure in BC test can be ignored?

PR #5607, opened 2024-05-21
[Model Monitoring] Reuse monitoring function images

TomerShor:
# Background:
The model monitoring infra deploys multiple nuclio functions per project. These functions are basically identical in every project. 
Up until now, each model monitoring enablement through MLRun triggered multiple simultaneous nuclio image builds, since MLRun used `code` as entry type, and provided a base-image, which resulted in nuclio building the images, and a major load on large-scale systems, as the kaniko build pods took a lot of resources - causing starvation to other kubernetes resources.

# Solution:
To improve this, we will reuse the built images for each of the monitoring functions across projects.
We are utilizing Nuclio's ability to deploy a function from an existing image, and skipping the build stage.

The images we want to reuse are for these monitoring functions: `Controller` , `Writer`, `Stream` and `Histogram`.
These 4 functions have different codebase, so each image will need to be built once so they can be reused.
In a later stage we will discuss if and how to handle other application images.

## Function image cache:
The images are cached in the DB in the Function Image Cache Table.
This table holds the following fields:
```python
class FunctionImageCache:
    id
    function_name
    mlrun_version
    nuclio_version
    base_image
    image
```

With the unique constraint of  `function_name`, `mlrun_version`, `nuclio_version` and `base_image`, so we will know to rebuild the image if one of these values changes.

Resolves https://iguazio.atlassian.net/browse/ML-6408


server/api/crud/model_monitoring/deployment.py
```diff
@@ -516,84 +679,150 @@ def _check_if_already_deployed(
...
-            )
-            is None
-        ):
-            logger.info(
-                "The stream function needs to be updated with the new stream trigger",
-                project=self.project,
-            )
-            return False
-        return True
+    def _update_image_manifest(
```
quaark: I'd like this method to work parallel. Writing to the manifest file parallel is a bad practice. But you can wait for the nuclio deployments in parallel. Wait for all the statuses to arrive and then write to the file. (Unless we make these rows in a DB table in which case everything can be parallel)

quaark (changes requested):
I feel like using a file here is a bad practice and will cause issues in the future. In my opinion we should create a new SQL table here and save the cache persistently there.
Otherwise looks good!

server/api/crud/model_monitoring/deployment.py
```diff
@@ -83,32 +85,69 @@ def __init__(
         self._parquet_batching_max_events = parquet_batching_max_events
         self._max_parquet_save_interval = max_parquet_save_interval
 
+        # the image manifest acts as a cache for deployed model monitoring function images.
+        # this allows us to reuse these images for future deployments across projects, improving efficiency.
+        # the manifest stores a unique image for each combination of three factors:
+        # mlrun version, nuclio version, and base image.
+        self._image_manifest = {}
+        self._read_image_manifest()
```
liranbg: this does not survive restarts. why not sql table?

server/api/crud/model_monitoring/deployment.py
```diff
@@ -516,84 +679,150 @@ def _check_if_already_deployed(
...
+        the key is a combination of the mlrun version and the nuclio version"""
+        return f"{config.version}-{config.nuclio_version}".encode().hex()
+
+    @staticmethod
+    def _set_nuclio_image_config(function, nuclio_image):
+        function.set_config("spec.image", nuclio_image)
+        function.set_config("spec.build.codeEntryType", "image")
+
+        # make sure the image won't be built in nuclio by clearing up build values
+        # TODO: remove this once the api will support deploying functions from an existing image directly
```
liranbg, line 876: maybe elaborate here / add jira reference to what needs to be done here

server/api/crud/model_monitoring/deployment.py
```diff
@@ -516,84 +679,150 @@ def _check_if_already_deployed(
...
+                "containerImage"
+            ]
+            save = True
+
+        # write the updated image manifest
+        if save:
+            manifest_path = self._get_manifest_path()
+            mlrun.utils.ensure_file_path_exists(manifest_path)
+            with open(manifest_path, "w") as f:
+                json.dump(self._image_manifest, f)
```
liranbg: note that this is not threadsafe

liranbg (changes requested):

server/api/crud/model_monitoring/deployment.py
```diff
@@ -309,17 +405,53 @@ def apply_and_create_stream_trigger(
...
+
+        try:
+            function = server.api.crud.Functions().get_function(
+                name=mm_constants.MonitoringFunctionNames.STREAM,
+                db_session=self.db_session,
+                project=self.project,
+            )
+        except mlrun.errors.MLRunNotFoundError:
+            logger.info(
+                "The stream function is not deployed yet when the user will run `enable_model_monitoring` "
```
gtopper:
```suggestion
                "The stream function is not deployed yet. When the user runs `enable_model_monitoring` "
```

PR #5610, opened 2024-05-22
[Notifications] Redact Secret Params from Notification Data in Webhook Notification

quaark:
https://iguazio.atlassian.net/browse/ML-6471

Created server side implementation of `NotificationPusher` that inherits the original `NotificationPusher` and adds masking the runs' notification configs just before being sent in the notification. This is mainly relevant for webhook notifications which include the entire body of the run and not a pretty display with cherry-picked data.


liranbg (approved):
🚀

PR #5613, opened 2024-05-22
[APIGateway] Keep mlrun function names in API gateway object on client side

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-6491

There is a confusion between nuclio function names and mlrun function names in MLRun’s api gateway implementation. The problem is that now when functions are passed as MLRun object, mlrun generates a nuclio name and keeps it in `api_gateway.functions` (not an MLRun name). It can be a bit confusing for a user, so this is how the problem is solved:

* We will keep MLRun function URI in `api_gateway.function` and generate nuclio function name just on MLRun api side

* When sending api gateway request to nuclio api, we also need to populate an annotation which contains MLRun function URI (this functionality will be used in the future for https://iguazio.atlassian.net/browse/ML-6363)

* When receive api gateway object from nuclio api, we want to decode a URI from annotation back to mlrun function name on MLRun api side, so when we get it on MLRun client side user won’t even have an idea what nuclio function name is.


mlrun/common/constants.py
```diff
@@ -14,6 +14,7 @@
 #
 IMAGE_NAME_ENRICH_REGISTRY_PREFIX = "."  # prefix for image name to enrich with registry
 MLRUN_CREATED_LABEL = "mlrun-created"
+MLRUN_FUNCTIONS_LABEL = "mlrun-functions"
```
TomerShor: 1. This is used for an annotation, not a label - so `MLRUN_FUNCTIONS_ANNOTATION`
2. In nuclio we usually annotate internal stuff with a `nuclio.io/` prefix. In mlrun we don't really do it but we do label internal stuff with `mlrun/` prefix so I think we should go with that in the annotation as well - `mlrun/function-names`

mlrun/common/schemas/api_gateway.py
```diff
@@ -91,6 +93,54 @@ class APIGateway(_APIGatewayBaseModel):
...
+            )
+
+            upstream_with_nuclio_names.append(upstream)
+            mlrun_function_uris.append(uri)
+
+        self.spec.upstreams = upstream_with_nuclio_names
+        if len(mlrun_function_uris) == 1:
+            self.metadata.annotations[MLRUN_FUNCTIONS_LABEL] = mlrun_function_uris[0]
+        elif len(mlrun_function_uris) == 2:
+            self.metadata.annotations[MLRUN_FUNCTIONS_LABEL] = "&".join(
```
TomerShor: nice

mlrun/common/schemas/api_gateway.py
```diff
@@ -91,6 +93,54 @@ class APIGateway(_APIGatewayBaseModel):
...
+        if mlrun_functions:
+            mlrun_function_uris = (
+                mlrun_functions.split("&")
+                if "&" in mlrun_functions
+                else [mlrun_functions]
+            )
+            if len(mlrun_function_uris) != len(self.spec.upstreams):
+                raise mlrun.errors.MLRunValueError(
+                    "Error when translating nuclio names to mlrun names in api gateway:"
+                    " upstream length doesn't match with mlrun functions label"
```
TomerShor: Maybe less nuclio-lingo:
```suggestion
                    " number of functions doesn't match the mlrun functions in annotation"
```

Also - is this user facing? if so we might need to think of a better error message as it is pretty much internal stuff.

mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -284,7 +284,21 @@ def _validate_functions(
         function_names = []
         for func in functions:
             if isinstance(func, str):
-                function_names.append(func)
+                # check whether the function was passed as a URI or just a name
+                parsed_project, function_name, tag, hash_key = (
```
TomerShor:
```suggestion
                parsed_project, function_name, _, _ = (
```

mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -284,7 +284,21 @@ def _validate_functions(
...
+                    mlrun.common.helpers.parse_versioned_object_uri(func)
+                )
+
+                if parsed_project and function_name:
```
TomerShor, line 291: if function_name is not None that it wasn't a URI?
If so - explain this in the comment above (L287)

tests/api/api/test_nuclio.py
```diff
@@ -153,3 +156,48 @@ def test_store_api_gateway(
...
+            ["test-project-test-func"],
+            "test-project/test-func",
+        ),
+        (
+            ["test-func1", "test-func2"],
+            ["test-project-test-func1", "test-project-test-func2"],
+            "test-project/test-func1&test-project/test-func2",
+        ),
+    ],
+)
```
TomerShor: Add test case with `tag`

TomerShor (changes requested):

mlrun/common/schemas/api_gateway.py
rokatyy: @TomerShor theoretically, yes - use may face this issue. I like your suggestion with changing comment, but I don't see a way to fully avoid a mentioning of internal specific here.

mlrun/runtimes/nuclio/api_gateway.py
rokatyy: @TomerShor I have a comment above

TomerShor (approved):

PR #5618, opened 2024-05-23
[Application] Validate the version of Nuclio before deploying

moranbental:
Functions with `kind="application"` are deployed on non-supported nuclio versions. The function was generated, and then the error was raised: `__init__ is supported since nuclio 1.13.1, currently using nuclio 1.12.14, please upgrade`
The validation was done after the base deploy when calling API Gateways that were only supported by Nuclio 1.13.1.
We can fail the request early by doing validation in the `ApplicationRuntime` constructor. 
https://iguazio.atlassian.net/browse/ML-6540

Also fixing the error message 
https://iguazio.atlassian.net/browse/ML-6539


alonmr (changes requested):
There is already a check in the constructor
```
    @min_nuclio_versions("1.12.7")
    def __init__(self, spec=None, metadata=None):
```
just change the version

alonmr (approved):

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -172,7 +172,7 @@ def __init__(
 class ApplicationRuntime(RemoteRuntime):
     kind = "application"
 
-    @min_nuclio_versions("1.12.7")
+    @min_nuclio_versions("1.13.1")
```
liranbg, line 175: it worth changing `min_nuclio_versions` implementation that when using it on __init__ it will log the Class name (ApplicationRuntime) and not __init__ as it says nothing.
when using specific function (.deploy) it will log Class.deploy functionality not supported.

liranbg (changes requested):

mlrun/runtimes/nuclio/function.py
```diff
@@ -66,7 +66,7 @@ def wrapper(*args, **kwargs):
                 return function(*args, **kwargs)
 
             message = (
-                f"{function.__name__} is supported since nuclio {' or '.join(versions)}, currently using "
+                f"'{function.__qualname__}' is supported since nuclio {' or '.join(versions)}, currently using "
```
liranbg: TIL 🚀

liranbg (approved):

PR #5621, opened 2024-05-23
[Model Monitoring] Support alerting with different result types 

Eyal-Danieli:
Following https://github.com/mlrun/mlrun/pull/5606, in this PR we add the support of sending an alert using the writer with different types of result kind (and not just `drift_detected`). In addition, we add a new result kind called `custom` which at the moment will be represented as `mm_app_anomaly` in the alert system. 

Related JIRA: https://iguazio.atlassian.net/browse/ML-6515


mlrun/model_monitoring/writer.py
```diff
@@ -117,26 +118,49 @@ def __init__(self, project: str, tsdb_secret_provider=None) -> None:
...
         )
         mlrun.get_run_db().generate_event(event_kind, event_data)
 
+    @staticmethod
+    def _generate_alert_event_kind(
+        result_kind: int, result_status: int
+    ) -> alert_objects.EventKind:
+        """Generate the required Event Kind format for the alerting system"""
+        if result_kind == ResultKindApp.custom:
+            # Right now the custom kind is represented as an anomaly detection
```
assaf758:
```suggestion
            # Custom kind is represented as an anomaly detection
```

assaf758 (approved):

PR #5631, opened 2024-05-26
[Feature Store] Fix remote ingest parquet filters

tomerm-iguazio:
Json serialization changes additional filters by transforming tuples into lists.
This issue cause remote run to fail.

1) Fix this issue by transform lists back to tuples.
2) Added system and unit tests.
3) Implement this solution on ParquetTarget and ParquetSource.
4) As a following Idea -> In order to properly check our test, we should perform the same process of encoding and decoding from JSON in a local ingestion as well.

[ML-6563](https://iguazio.atlassian.net/browse/ML-6563)

[ML-6563]: https://iguazio.atlassian.net/browse/ML-6563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/datastore/test_targets.py
```diff
@@ -111,3 +112,25 @@ def test_write_with_too_many_partitions():
         match="Maximum number of partitions exceeded. To resolve this.*",
     ):
         parquet_target.write_dataframe(df)
+
+
+def test_revert_list_filters_to_tuple():
+    #  Target-specific to test_revert_list_filters_to_tuple test, which checks all the cases.
```
gtopper: I don't understand this comment.

mlrun/datastore/targets.py
```diff
@@ -999,7 +1000,7 @@ def as_df(
             start_time=start_time,
             end_time=end_time,
             time_column=time_column,
-            additional_filters=additional_filters,
+            additional_filters=revert_list_filters_to_tuple(additional_filters),
```
gtopper: Why is this needed in the target?

mlrun/datastore/utils.py
```diff
@@ -180,3 +181,55 @@ def get_kafka_brokers_from_dict(options: dict, pop=False) -> typing.Optional[str
             FutureWarning,
         )
     return kafka_bootstrap_servers
+
+
+def revert_list_filters_to_tuple(additional_filters):
```
gtopper: Revert is a bit confusing here. Maybe
```suggestion
def transform_list_filters_to_tuple(additional_filters):
```

tests/datastore/test_sources.py
```diff
@@ -107,14 +108,43 @@ def test_timestamp_format_inference(rundb_mock):
...
+        ),
+    ],
 )
-def test_nan_additional_filters(additional_filters):
-    with pytest.raises(
-        mlrun.errors.MLRunInvalidArgumentError,
-        match="using NaN in additional_filters is not supported",
-    ):
+def test_revert_list_filters_to_tuple(additional_filters, message):
+    #  Source-specific to test_revert_list_filters_to_tuple test, which checks all the cases.
```
gtopper: I don't understand this comment.

tests/datastore/test_targets.py
```diff
@@ -111,3 +112,25 @@ def test_write_with_too_many_partitions():
...
+    def json_change(filters):
+        json_data = json.dumps(filters)
+        return json.loads(json_data)
+
+    additional_filters = [[("x", "=", 3), ("x", "=", 4), ("x", "=", 5)]]
+    parquet_target = ParquetTarget("parquet_target", path="path/to/file")
+    after_json_change_filters = json_change(additional_filters)
```
gtopper: Just inline it... "json change" doesn't make it any more readable.
```suggestion
    back_from_json_serialization = json.loads(json.dumps(additional_filters))
```

mlrun/datastore/utils.py
```diff
@@ -180,3 +181,55 @@ def get_kafka_brokers_from_dict(options: dict, pop=False) -> typing.Optional[str
...
+
+def validate_additional_filters(additional_filters):
+    if not additional_filters:
+        return
+    for filter_tuple in additional_filters:
+        if not filter_tuple:
+            continue
+        if not isinstance(filter_tuple, (list, tuple)):
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"mlrun supports additional_filters only as a list of tuples."
```
gtopper, line 205: This error message is duplicated in 3 places. It seems to me like more specific error messages are in order.

mlrun/datastore/utils.py
```diff
@@ -180,3 +181,55 @@ def get_kafka_brokers_from_dict(options: dict, pop=False) -> typing.Optional[str
...
+                f" Current additional_filters: {additional_filters}"
+            )
+        col_name, op, value = filter_tuple
+        if isinstance(value, float) and math.isnan(value):
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "using NaN in additional_filters is not supported"
+            )
+        elif isinstance(value, (list, tuple, set)):
```
gtopper: Not sure why all other validations refer to `list` or `tuple`, but this one checks for `set` as well.

mlrun/datastore/utils.py
```diff
@@ -180,3 +181,55 @@ def get_kafka_brokers_from_dict(options: dict, pop=False) -> typing.Optional[str
...
+            for sub_value in value:
+                if isinstance(sub_value, float) and math.isnan(sub_value):
+                    raise mlrun.errors.MLRunInvalidArgumentError(
+                        "using NaN in additional_filters is not supported"
```
gtopper: Maybe it's fine to duplicate this one, but in that case better extract it to a local variable.

tests/datastore/test_base.py
```diff
@@ -214,3 +216,47 @@ def test_schema_to_store(schemas, expected_class, expected):
...
+        ),
+        (
+            [("age", "in", [10, float("nan")])],
+            "using NaN in additional_filters is not supported",
+        ),
+        ([("x", "=", "=", 3), ("y", "in", [3, 4])], "illegal filter tuple length"),
+        ([()], ""),
+    ],
+)
+def test_revert_list_filters_to_tuple(additional_filters, message):
```
gtopper: If the method being tested is in utils (which maybe it shouldn't be), then the test should be in `test_utils.py`.

tests/system/feature_store/test_feature_store.py
```diff
@@ -4860,12 +4865,19 @@ def test_parquet_filters(self, engine):
         vec = fstore.FeatureVector(
             name="test-fs-vec", features=["parquet-filters-fs.*"]
         )
+        vec.save()
+        get_offline_target = ParquetTarget(
+            "get_offline_target",
```
gtopper: Not sure how it ended up bring called `get_offline_target`.
```suggestion
        target = ParquetTarget(
```

mlrun/datastore/sources.py
```diff
@@ -313,12 +315,13 @@ def __init__(
         schedule: str = None,
         start_time: Optional[Union[datetime, str]] = None,
         end_time: Optional[Union[datetime, str]] = None,
-        additional_filters: Optional[list[tuple]] = None,
+        additional_filters: Optional[list[tuple, list]] = None,
```
gtopper:
```suggestion
        additional_filters: Optional[list[Union[tuple, list]]] = None,
```

mlrun/datastore/targets.py
tomerm-iguazio: to_dataframe can call this function after additional filters passed 
json encode-decode.

tests/datastore/test_targets.py
tomerm-iguazio: tnx!

mlrun/datastore/utils.py
tomerm-iguazio: It is just a nested type check, it is making sense that type error can have multiple appearances.

tests/system/feature_store/test_feature_store.py
```diff
@@ -4860,12 +4865,18 @@ def test_parquet_filters(self, engine):
         vec = fstore.FeatureVector(
             name="test-fs-vec", features=["parquet-filters-fs.*"]
         )
+        vec.save()
+        get_offline_target = ParquetTarget(
```
gtopper: Same comment as before:
```suggestion
        target = ParquetTarget(
```

mlrun/datastore/utils.py
gtopper: To the extent that it makes sense, it should be extracted like the other error.

mlrun/datastore/utils.py
```diff
@@ -180,3 +181,53 @@ def get_kafka_brokers_from_dict(options: dict, pop=False) -> typing.Optional[str
...
+                f"mlrun supports additional_filters only as a list of tuples."
+                f" Current additional_filters: {additional_filters}"
+            )
+        if len(filter_tuple) != 3:
+            if all(isinstance(element, (list, tuple)) for element in filter_tuple):
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    f"mlrun supports additional_filters only as a list of tuples."
+                    f" Current additional_filters: {additional_filters}"
+                )
+            else:
```
gtopper: This doesn't seem to make sense. If all elements are lists or tuples, then we raise an error that "mlrun supports additional_filters only as a list of tuples"? It seems like this whole case can be dropped.
```suggestion
```

mlrun/datastore/utils.py
```diff
@@ -180,3 +181,53 @@ def get_kafka_brokers_from_dict(options: dict, pop=False) -> typing.Optional[str
...
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    f"illegal filter tuple length, {filter_tuple} in additional filters:"
+                    f" {additional_filters}"
+                )
+        if isinstance(filter_tuple[0], (list, tuple)) or isinstance(
+            filter_tuple[1], (list, tuple)
+        ):
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"mlrun supports additional_filters only as a list of tuples."
+                f" Current additional_filters: {additional_filters}"
```
gtopper: Here too, it doesn't seem to make sense. If `filter_tuple[0]` is a list or a tuple, and `filter_tuple[1]` is a list or a tuple, then we raise an error that "mlrun supports additional_filters only as a list of tuples"?

tests/datastore/test_sources.py
```diff
@@ -107,14 +108,38 @@ def test_timestamp_format_inference(rundb_mock):
...
+        ),
+    ],
 )
-def test_nan_additional_filters(additional_filters):
-    with pytest.raises(
-        mlrun.errors.MLRunInvalidArgumentError,
-        match="using NaN in additional_filters is not supported",
-    ):
+def test_transform_list_filters_to_tuple(additional_filters, message):
+    after_json_change_filters = json.loads(json.dumps(additional_filters))
```
gtopper: Same suggestion again
```suggestion
    back_from_json_serialization = json.loads(json.dumps(additional_filters))
```

mlrun/datastore/utils.py
```diff
@@ -180,3 +181,53 @@ def get_kafka_brokers_from_dict(options: dict, pop=False) -> typing.Optional[str
             FutureWarning,
         )
     return kafka_bootstrap_servers
+
+
+def transform_list_filters_to_tuple(additional_filters):
+    tuple_filters = []
+    if not additional_filters:
+        return tuple_filters
+    validate_additional_filters(additional_filters=additional_filters)
```
gtopper: Minor style thing, but I think this is unnecessary boilerplate.
```suggestion
    validate_additional_filters(additional_filters)
```

mlrun/datastore/sources.py
```diff
@@ -393,6 +377,7 @@ def to_step(
         attributes.pop("additional_filters", None)
         if context:
             attributes["context"] = context
+        additional_filters = transform_list_filters_to_tuple(additional_filters)
```
gtopper, line 378: 10 lines down, we have `additional_filters=self.additional_filters or additional_filters`. What if we end up getting the additional filters from self and not from the parameter?

mlrun/datastore/sources.py
```diff
@@ -428,6 +423,7 @@ def to_dataframe(
         additional_filters=None,
     ):
         reader_args = self.attributes.get("reader_args", {})
+        additional_filters = transform_list_filters_to_tuple(additional_filters)
```
gtopper, line 424: 8 lines down, we have `additional_filters=additional_filters or self.additional_filters`. What if we end up getting the additional filters from self and not from the parameter?

Also, this is preexisting, but I wonder why here it's `additional_filters or self.additional_filters` and in `to_step()` it's `self.additional_filters or additional_filters`.

mlrun/datastore/sources.py
```diff
@@ -31,6 +30,9 @@
 import mlrun
 from mlrun.config import config
 from mlrun.datastore.snowflake_utils import get_snowflake_spark_options
+from mlrun.datastore.utils import (
+    transform_list_filters_to_tuple,
+)
```
gtopper: Ruff seems to be lazy, not changing this either way, so I'll take it up on myself to comment :)
```suggestion
from mlrun.datastore.utils import transform_list_filters_to_tuple
```

mlrun/datastore/utils.py
```diff
@@ -180,3 +181,53 @@ def get_kafka_brokers_from_dict(options: dict, pop=False) -> typing.Optional[str
...
+
+
+def validate_additional_filters(additional_filters):
+    nan_error_message = "using NaN in additional_filters is not supported"
+
+    if not additional_filters:
+        return
+    for filter_tuple in additional_filters:
+        if not filter_tuple:
+            continue
```
gtopper: Probably not good enough? Sure, you might have an empty list there for some reason, but you could also have other stuff that just happens to evaluate to boolean false, e.g. zero, empty string, `None`...

mlrun/datastore/utils.py
```diff
@@ -180,3 +181,53 @@ def get_kafka_brokers_from_dict(options: dict, pop=False) -> typing.Optional[str
...
+    for additional_filter in additional_filters:
+        tuple_filters.append(tuple(additional_filter))
+    return tuple_filters
+
+
+def validate_additional_filters(additional_filters):
+    nan_error_message = "using NaN in additional_filters is not supported"
+
+    if not additional_filters:
+        return
```
gtopper: Here too, it could be an empty list or `None`, or it could be something unexpected like zero or an empty string.

mlrun/datastore/utils.py
tomerm-iguazio: if every element in the inner tuple is also list or tuple...

instead of [('x','in',[1,2,3])]

we list of list of tuples:
[[('x','in',[1,2,3])]]

mlrun/datastore/sources.py
tomerm-iguazio: in self additional filters saved as tuple, you can see the transformation I did in `from_dict` function and `__init__`

mlrun/datastore/sources.py
tomerm-iguazio: about 1, same explanation as my comment before.
2 - you are absolutely right, but this are the convention of other parameters in these function, If I will change them I can break the behavior.

mlrun/datastore/utils.py
tomerm-iguazio: nice catch!

tomerm-iguazio: + added unit test case

mlrun/datastore/utils.py
tomerm-iguazio: this error here is about what inside the list of tuples :)

mlrun/datastore/utils.py
tomerm-iguazio: additional filters does not support nested list inside filter tuples.  Current filter_tuple: {filter_tuple}.

mlrun/datastore/utils.py
```diff
@@ -180,3 +181,48 @@ def get_kafka_brokers_from_dict(options: dict, pop=False) -> typing.Optional[str
...
+        tuple_filters.append(tuple(additional_filter))
+    return tuple_filters
+
+
+def validate_additional_filters(additional_filters):
+    nan_error_message = "using NaN in additional_filters is not supported"
+    if (
+        additional_filters is None
+        or additional_filters == ()
+        or additional_filters == []
```
gtopper: How about the shorter
```suggestion
        additional_filters in [None, [], ()]
```
?

gtopper (approved):

PR #5635, opened 2024-05-27
[Serving] Rename `model-conf` to `serving-conf`

gtopper:
Since it has nothing to do with models.


mlrun/common/constants.py
```diff
@@ -14,7 +14,7 @@
 #
 
 IMAGE_NAME_ENRICH_REGISTRY_PREFIX = "."  # prefix for image name to enrich with registry
-MLRUN_MODEL_CONF = "model-conf"
+MLRUN_MODEL_CONF = "serving-conf"
```
alxtkr77: If model-conf is misleading I would change the MLRUN_MODEL_CONF to MLRUN_SERVING_CONF as well

gtopper: Right, missed that...

alxtkr77 (approved):

PR #5638, opened 2024-05-27
[ApiGateway] Make authenticators serializable with `to_dict()`

alonmr:
Since the authenticators are not serializable, it was not possible to send the application object over http when enriched the api gateway authenticator objects. This makes them serializable with only minimal information as it is not relevant to deploying or saving the function, only for metadata. When we actually create the API gateway we enrich with authentication information.


liranbg (approved):

mlrun/runtimes/nuclio/api_gateway.py
TomerShor, line 1: Just to simplify the inheritance **a bit**, I would suggest something of this sort:
```python
class Authenticator(typing.Protocol):
    # same implementation as the current APIGatewayAuthenticator...

class APIGatewayAuthenticator(Authenticator, ModelObj):
    _dict_fields = ["authentication_mode"]

class class NoneAuth(APIGatewayAuthenticator):
    # same thing...
```

PR #5640, opened 2024-05-27
[Deprecations] Deprecate SQLite Migrations

quaark:
https://iguazio.atlassian.net/browse/ML-5884

We still support SQLite as a DB, but no longer do we support upgrading the SQLite DB with alembic migrations. Now migrations are only for MySQL:

- Removed sqlite migrations and models file
- API only runs migrations if DB engine is mysql
- Move migration tests to test against mysql and not sqlite (removed 3 tests that are no longer relevant as those migrations are before the init migration of mysql)
- Added migration with missing changes from mysql db
- Update mysql collation from `utf8_bin` to `utf8mb3_bin`. `utf8_bin` is deprecated in MySQL and now uses `utf8mb3_bin` when `utf8_bin` is set. Alembic tries to migrate the db back to `utf8_bin` but shouldn't. Now the collation in code matches MySQL collation. TODO: upgrade to `utf8mb4_bin` in the future as described in the MySQL warning.
```
2024-05-28 11:29:05 2024-05-28T08:29:05.147254Z 0 [Warning] [MY-013242] [Server] --character-set-server: 'utf8' is currently an alias for the character set UTF8MB3, but will be an alias for UTF8MB4 in a future release. Please consider using UTF8MB4 in order to be unambiguous.
2024-05-28 11:29:05 2024-05-28T08:29:05.147580Z 0 [Warning] [MY-013244] [Server] --collation-server: 'utf8mb3_bin' is a collation of the deprecated character set UTF8MB3. Please consider using UTF8MB4 with an appropriate collation instead.
```


automation/scripts/test_migration_mysql.sh
```diff
@@ -16,24 +16,47 @@
...
+docker run \
+	--name=migration-db \
+	--rm \
+	-v "${ROOT_DIR}:/mlrun" \
+	-p 3306:3306 \
+	-e MYSQL_ROOT_PASSWORD="pass" \
+	-e MYSQL_ROOT_HOST="%" \
+	-e MYSQL_DATABASE="mlrun" \
+	-d \
+	mysql/mysql-server:8.0 \
```
liranbg:
```suggestion
	mysql/mysql-server:8.0 \
```
take from gcr.io/iguazio

server/api/db/mlrun.db
liranbg, line 1: delete. add to git ignore

automation/scripts/create_migration_mysql.sh
```diff
@@ -57,6 +57,6 @@ done
 
 export PYTHONPATH=$ROOT_DIR
 
-alembic -c "${ROOT_DIR}/server/api/alembic_mysql.ini" upgrade head
-alembic -c "${ROOT_DIR}/server/api/alembic_mysql.ini" revision --autogenerate -m "${MLRUN_MIGRATION_MESSAGE}"
-
+cd ${ROOT_DIR}/server/api
```
alonmr, line 59: IIRC if you cd you change the cwd which may cause errors when running this as part of some make command

server/api/db/mlrun.db
alonmr, line 1: should this be removed?

automation/scripts/test_migration_mysql.sh
```diff
@@ -16,24 +16,47 @@
...
+		exit 1
+	fi
+	times=$(( times + 1 ))
+done
 
 export PYTHONPATH=$ROOT_DIR
 
-alembic -c "${ROOT_DIR}/server/api/alembic.ini" upgrade head
-alembic -c "${ROOT_DIR}/server/api/alembic.ini" revision --autogenerate -m "${MLRUN_MIGRATION_MESSAGE}"
+cd ${ROOT_DIR}/server/api
```
alonmr: same about cd, think you may be able to cd without changing the cwd like in `make compile-schemas`

alonmr (commented):
Looks good! small suggestions

PR #5641, opened 2024-05-27
[Model Monitoring] Add MySQL support for saving and querying metrics

jond01:
Implements [ML-6268](https://iguazio.atlassian.net/browse/ML-6268).

I tested the `/metrics` endpoint on a system configured to use MySQL in model monitoring, and it works:

<details>
  <summary>
    Example
  </summary>

```json
[
  {
    "project": "test-app-flow-sql-3",
    "app": "evidently-app-test-v2",
    "type": "result",
    "name": "data_drift_test",
    "full_name": "test-app-flow-sql-3.evidently-app-test-v2.result.data_drift_test"
  },
  {
    "project": "test-app-flow-sql-3",
    "app": "histogram-data-drift",
    "type": "result",
    "name": "general_drift",
    "full_name": "test-app-flow-sql-3.histogram-data-drift.result.general_drift"
  },
  {
    "project": "test-app-flow-sql-3",
    "app": "monitoring-test-v2",
    "type": "result",
    "name": "data_drift_test",
    "full_name": "test-app-flow-sql-3.monitoring-test-v2.result.data_drift_test"
  },
  {
    "project": "test-app-flow-sql-3",
    "app": "monitoring-test-v2",
    "type": "result",
    "name": "model_perf",
    "full_name": "test-app-flow-sql-3.monitoring-test-v2.result.model_perf"
  },
  {
    "project": "test-app-flow-sql-3",
    "app": "histogram-data-drift",
    "type": "metric",
    "name": "hellinger_mean",
    "full_name": "test-app-flow-sql-3.histogram-data-drift.metric.hellinger_mean"
  },
  {
    "project": "test-app-flow-sql-3",
    "app": "histogram-data-drift",
    "type": "metric",
    "name": "kld_mean",
    "full_name": "test-app-flow-sql-3.histogram-data-drift.metric.kld_mean"
  },
  {
    "project": "test-app-flow-sql-3",
    "app": "histogram-data-drift",
    "type": "metric",
    "name": "tvd_mean",
    "full_name": "test-app-flow-sql-3.histogram-data-drift.metric.tvd_mean"
  },
  {
    "project": "test-app-flow-sql-3",
    "app": "mlrun-infra",
    "type": "metric",
    "name": "invocations",
    "full_name": "test-app-flow-sql-3.mlrun-infra.metric.invocations"
  }
]
```

</details>

[ML-6268]: https://iguazio.atlassian.net/browse/ML-6268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/model_monitoring/db/stores/base/store.py
```diff
@@ -125,7 +125,6 @@ def write_application_event(
                       object.
         :param kind: The type of the event, can be either "result" or "metric".
         """
-        pass
```
davesh0812, line 128: please revert it in order to be consistent with mlrun, and add a pass to `get_model_endpoint_metrics`.

mlrun/model_monitoring/db/stores/sqldb/models/base.py
```diff
@@ -89,11 +90,11 @@ class ModelEndpointsBaseTable(BaseModel):
     metrics = Column(EventFieldType.METRICS, Text)
     first_request = Column(
         EventFieldType.FIRST_REQUEST,
-        TIMESTAMP,
+        TIMESTAMP(timezone=True),
```
davesh0812, line 93: Are you sure this change is safe for existing  users?

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -500,22 +514,32 @@ def _delete_application_result(
     ) -> None:
         self._init_application_results_table()
         criteria = self._get_filter_criteria(
-            table=self.ApplicationResultsTable,
+            table=self.application_results_table,
+            endpoint_id=endpoint_id,
+            application_name=application_name,
+        )
+        # Delete the table
```
davesh0812: Are you sure it deletes the table? I think it deletes only the relevant entries. 
The same for line 533.

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -605,13 +629,56 @@ def delete_model_endpoints_resources(self) -> None:
...
+        # results, not only `one_or_none`.
+        with sqlalchemy.orm.Session(self._engine) as session:
+            metric_rows = (
+                session.query(table)  # pyright: ignore[reportOptionalCall]
+                .filter(table.endpoint_id == endpoint_id)
+                .all()
+            )
+
+        return [
+            mm_schemas.ModelEndpointMonitoringMetric(
```
davesh0812, line 671: I’m not sure if this is the right PR, but yesterday you mentioned that the result kind is being returned to the UI. I don't see it here. I noticed it’s being returned from the TSDB, but it seems more appropriate to return it from the `/metrics` endpoint since it’s metadata.

davesh0812 (commented):
Overall looks good.
I have few comments.

Just saying we will have a lot of BC fun in 1.8.0 :)

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
jond01: "kind" isn't designed to be in this response, see section 2-c in the description of [ML-6119](https://iguazio.atlassian.net/browse/ML-6119).

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
jond01: Right, I'll update the comment.

mlrun/model_monitoring/db/stores/sqldb/models/base.py
jond01: This change is mainly to signify that timezones are important and should be thought of by the devs.
In practice, it [isn't supported in MySQL](https://docs.sqlalchemy.org/en/14/dialects/mysql.html#sqlalchemy.dialects.mysql.TIMESTAMP.__init__), so it won't cause BC issues:
> timezone – not used by the MySQL dialect.

See also:
https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.TIMESTAMP.params.timezone
> If the target database does not support timezones, this flag is ignored.

jond01:
> Just saying we will have a lot of BC fun in 1.8.0 :)

I believe it might not be that big here 😉
Once we support retention, we may dismiss these metrics/results tables in MySQL/V3IO altogether and instead read from the TSDB.

mlrun/model_monitoring/db/stores/base/store.py
jond01: We may talk about it, but `pass` is just a placeholder when a syntactic construct is needed.
The docstring fulfills that need in this abstract method. Therefore, `pass` is redundant.

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
davesh0812: The design can be changed. Don't you think it's more accurate the way I mentioned?

davesh0812 (approved):

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
jond01: I think that's a valid point. I would also like to send the latest status in case of a result, to indicate which results should be inspected.
However, we discussed showing the status in the dropdown, and we deferred it (to >=1.8), so I think we shouldn't reopen now the API.

PR #5620, opened 2024-05-23
[Model Monitoring] Employ the TSDB abstraction in the metrics endpoints

jond01:
Implements [ML-6429](https://iguazio.atlassian.net/browse/ML-6429).

Add three abstract methods to the base TSDB connector class:

* `read_metrics_data` - get application results/metrics values.
* `read_predictions` - get the invocations metric values.
* `read_prediction_metric_for_endpoint_if_exists` - get invocations metric existence.

Remove `v3io_tsdb_reader.py`. Move almost all the functions to `tsdb/v3io/v3io_connector.py`, except for `get_invocations_fqn` that went to the helpers module.

The main system test passed.

[ML-6429]: https://iguazio.atlassian.net/browse/ML-6429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5642, opened 2024-05-27
[Model Monitoring] Implement TDEngine read metrics API

Eyal-Danieli:
Following https://github.com/mlrun/mlrun/pull/5620, this PR implements the metrics endpoints on top of the TDEngine.

In general, the implementation is quite similar to the V3IO Connector. For example, `read_metrics` endpoint includes: 
- Type validate (`metrics` or `results`)
- Prepare SQL query with the provided metrics
- Query the DB and get a `DataFrame` result
- Parse the `DataFrame` into a list of metrics values per distinct results (using  `df_to_results_values` & `df_to_metrics_values` that are now stored under the base class).

In addition, to keep consistency with the V3IO TSDB schema, we updated the application result table and added a new tag - `result_kind`. 

Related JIRA: https://iguazio.atlassian.net/browse/ML-6557


mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -260,13 +326,65 @@ def read_predictions(
...
+            end=mlrun.utils.now_date(),
+            limit=1,
+        )
+        if predictions:
+            return mm_schemas.ModelEndpointMonitoringMetric(
+                project=self.project,
+                app=mm_schemas.SpecialApps.MLRUN_INFRA,
+                type=mm_schemas.ModelEndpointMonitoringMetricType.METRIC,
+                name=mm_schemas.PredictionsQueryConstants.INVOCATIONS,
+                full_name=mlrun.model_monitoring.helpers.get_invocations_fqn(
```
gtopper: I think better import it directly... Too much boilerplate.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -251,7 +274,50 @@ def read_metrics_data(
...
+            timestamp_column=mm_schemas.WriterEvent.END_INFER_TIME,
+        )
+
+        df[mm_schemas.WriterEvent.END_INFER_TIME] = pd.to_datetime(
+            df[mm_schemas.WriterEvent.END_INFER_TIME]
+        )
+        df.set_index(mm_schemas.WriterEvent.END_INFER_TIME, inplace=True)
+
+        logger.debug(
+            "Read a data-frame",
```
gtopper: Not sure we need this log, but if we want to keep it, better improve the message and add the table path.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -186,34 +188,55 @@ def get_model_endpoint_real_time_metrics(
...
         :raise:  MLRunInvalidArgumentError if query the provided table failed.
         """
-
-        filter_query += f" project = '{self.project}'"
+        with StringIO() as query:
+            if filter_query:
+                query.write(filter_query)
+                query.write(" and ")
+            query.write(f"project = '{self.project}'")
+            filter_query = query.getvalue()
```
gtopper: Overkill to use `StringIO` here, this is only a few short strings being concatenated. Can be something like:
```suggestions
        project_condition = f"project = '{self.project}'"
        filter_query = (
            f"{filter_query} and {project_condition}"
            if filter_query
            else project_condition
        )
```

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
```diff
@@ -148,10 +173,21 @@ def _get_records_query(
                     query.write(f"{timestamp_column} >= '{start}'" + " and ")
                 if end:
                     query.write(f"{timestamp_column} <= '{end}'")
+            full_query = query.getvalue()
+            if full_query.endswith(" and "):
+                full_query = full_query[:-5]
+        if any([limit, interval]):
+            with StringIO() as query:
```
gtopper: Here too it's an overkill to use `StringIO`.

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
```diff
@@ -148,10 +173,21 @@ def _get_records_query(
                     query.write(f"{timestamp_column} >= '{start}'" + " and ")
                 if end:
                     query.write(f"{timestamp_column} <= '{end}'")
+            full_query = query.getvalue()
+            if full_query.endswith(" and "):
+                full_query = full_query[:-5]
+        if any([limit, interval]):
```
gtopper: That's just
```suggestion
        if limit or interval:
```

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -260,13 +326,65 @@ def read_predictions(
...
+                    df["count(latency)"],
+                )
+            ),  # pyright: ignore[reportArgumentType]
+        )
 
     def read_prediction_metric_for_endpoint_if_exists(
         self, endpoint_id: str
     ) -> typing.Optional[mm_schemas.ModelEndpointMonitoringMetric]:
-        raise NotImplementedError
+        # Read just one record, because we just want to check if there is any data for this endpoint_id
```
gtopper, line 383: I wonder if we really need to duplicate this method... It's basically the same in both engines.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -186,34 +188,55 @@ def get_model_endpoint_real_time_metrics(
...
+        :param columns:          Columns to include in the result.
+        :param filter_query:     Optional filter expression as a string. TDengine supports SQL-like syntax.
+        :param interval:         The interval to aggregate the data by. Note that if interval is provided,
+                                 agg must bg provided as well. Provided as a string in the format of
+                                 '1m', '1h', etc.
+        :param agg:              The aggregation functions to apply on the columns. Note that if agg is provided,
+                                 interval must bg provided as well. Provided as a list of strings in the format of
+                                 ['sum', 'avg', 'count', ...].
+        :param limit:            The maximum number of records to return.
+        :param sliding_window:   The time step for which the time window moves forward. Note that if sliding_window is
```
gtopper: The name `sliding_window` isn't informative enough. Best guess would be that it's the sliding window's size, but even that's wrong according to the description. Maybe `sliding_window_step`?

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -251,7 +274,50 @@ def read_metrics_data(
...
+
+        list_of_metrics = [
+            f"({mm_schemas.WriterEvent.APPLICATION_NAME} = '{metric.app}' AND {name} = '{metric.name}')"
+            for metric in metrics
+        ]
+        with StringIO() as query:
+            query.write(f"endpoint_id='{endpoint_id}' ")
+            query.write("AND ")
+            query.write(" OR ".join(list_of_metrics))
+            filter_query = query.getvalue()
```
gtopper: Here too, using `StringIO` is overkill, but it also seems to be wrong, because `AND` takes precedence over `OR`.

Maybe something like
```suggestion
        metrics_condition = " OR ".join([
            f"({mm_schemas.WriterEvent.APPLICATION_NAME} = '{metric.app}' AND {name} = '{metric.name}')"
            for metric in metrics
        ])
        filter_query = f"endpoint_id='{endpoint_id}' AND ({metrics_condition})"
```

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -260,13 +326,65 @@ def read_predictions(
...
+            )
+
+        df["_wend"] = pd.to_datetime(df["_wend"])
+        df.set_index("_wend", inplace=True)
+        return mm_schemas.ModelEndpointMonitoringMetricValues(
+            full_name=full_name,
+            values=list(
+                zip(
+                    df.index,
+                    df["count(latency)"],
```
gtopper: Is this correct also in TDEngine? Is that what the query actually returns?

tests/model_monitoring/test_tdengine.py
```diff
@@ -237,3 +238,127 @@ def test_get_records_query(
...
+            assert "interval must be provided when using sliding window" in str(
+                err.value
+            )
+        columns_to_select = ", ".join(
+            [f"{a}({col})" for a in agg for col in columns_to_filter]
+        )
+        expected_query = (
+            f""
+            f"SELECT _wstart, _wend, {columns_to_select} from {_MODEL_MONITORING_DATABASE}.{subtable} "
+            f"where {timestamp_column} >= '{start}' and {timestamp_column} <= '{end}' "
```
gtopper: Not sure why `SELECT` and `LIMIT` are uppercase but `from`, `where` and `and` are lowercase? Better keep them consistent.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -186,34 +188,55 @@ def get_model_endpoint_real_time_metrics(
...
-        start: str,
-        end: str,
+        start: datetime,
+        end: datetime,
         columns: typing.Optional[list[str]] = None,
         filter_query: str = "",
+        interval: str = "",
+        agg: typing.Optional[list] = None,
+        limit: int = 0,
+        sliding_window: str = "",
```
gtopper: Whatever parameters are added here should be added in the base class as well. And the types should match too.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -260,13 +326,65 @@ def read_predictions(
...
         aggregation_window: typing.Optional[str] = None,
+        limit: typing.Optional[int] = None,
     ) -> typing.Union[
         mm_schemas.ModelEndpointMonitoringMetricValues,
         mm_schemas.ModelEndpointMonitoringMetricNoData,
     ]:
-        raise NotImplementedError
+        if not aggregation_window:
+            logger.warning(
+                "Aggregation window is not provided, defaulting to 10 minute."
```
gtopper:
```suggestion
                "Aggregation window is not provided, defaulting to 10 minutes."
```

gtopper: Also, that's inconsistent with the behvaiour of the TSDB implementation, isn't it?

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
Eyal-Danieli: yes

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
Eyal-Danieli: amm it's not 100% similar. In V3IO we need to use strings for the `start` and `end` fields (`0` and `now`) while in TDEngine we need to use datetime format for these params (`datetime.min` and `mlrun.utils.now_date()`)

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
Eyal-Danieli: Even though it's not enforced yet, we are trying to avoid direct internal imports as much as we can (following platform instructions)

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -103,8 +106,8 @@ def get_model_endpoint_real_time_metrics(
     def get_records(
         self,
         table: str,
-        start: str,
-        end: str,
+        start: typing.Union[datetime, str],
+        end: typing.Union[datetime, str],
```
jond01:
```suggestion
        start: datetime,
        end: datetime,
```
Since you support only `datetime` and not `str` in TDEngine. It's ok to extend the supported types in subclasses, but not to narrow them.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
Eyal-Danieli: renamed to `sliding_window_step`

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
Eyal-Danieli: good catch!

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -209,3 +212,141 @@ def read_prediction_metric_for_endpoint_if_exists(
...
+    ) -> list[
+        typing.Union[
+            mm_schemas.ModelEndpointMonitoringMetricValues,
+            mm_schemas.ModelEndpointMonitoringMetricNoData,
+        ]
+    ]:
+        """
+        Parse a time-indexed data-frame of metrics from the TSDB into a list of
+        metrics values per distinct results.
+        When a metric is not found in the data-frame, it is represented in no-data object.
```
jond01:
```suggestion
        When a metric is not found in the data-frame, it is represented in a no-data object.
```

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -209,3 +212,141 @@ def read_prediction_metric_for_endpoint_if_exists(
...
+    ) -> list[
+        typing.Union[
+            mm_schemas.ModelEndpointMonitoringResultValues,
+            mm_schemas.ModelEndpointMonitoringMetricNoData,
+        ]
+    ]:
+        """
+        Parse a time-indexed data-frame of results from the TSDB into a list of
+        results values per distinct results.
+        When a result is not found in the data-frame, it is represented in no-data object.
```
jond01:
```suggestion
        When a result is not found in the data-frame, it is represented in a no-data object.
```

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
```diff
@@ -125,16 +125,41 @@ def _get_subtables_query(
...
     ) -> str:
+        if agg and not columns_to_filter:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "columns_to_filter must be provided when using aggregate functions"
+            )
+
+        # if aggregate function or interval is provided, the other must be provided as well
+        if (agg and not interval) or (interval and not agg):
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "both interval and aggregate function must be provided or neither"
```
jond01:
```suggestion
                "Both `interval` and `agg` must be provided or neither"
```

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
```diff
@@ -125,16 +125,41 @@ def _get_subtables_query(
...
         table: str,
-        start: str,
-        end: str,
+        start: datetime,
+        end: datetime,
         columns_to_filter: list[str] = None,
         filter_query: str = "",
+        interval: str = "",
+        limit: int = 0,
+        agg: Optional[list] = None,
```
jond01:
```suggestion
        aggregations: Optional[list] = None,
```

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
```diff
@@ -148,10 +173,21 @@ def _get_records_query(
                     query.write(f"{timestamp_column} >= '{start}'" + " and ")
                 if end:
                     query.write(f"{timestamp_column} <= '{end}'")
+            full_query = query.getvalue()
+            if full_query.endswith(" and "):
+                full_query = full_query[:-5]
```
jond01: It's ugly - please avoid it

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
```diff
@@ -148,10 +173,21 @@ def _get_records_query(
...
+        if any([limit, interval]):
+            with StringIO() as query:
+                query.write(full_query)
```
jond01: It doesn't make sense - just write to the same `StringIO` object and call `getvalue` upon completion

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
```diff
@@ -148,10 +173,21 @@ def _get_records_query(
...
+                if sliding_window:
+                    query.write(f" SLIDING({sliding_window})")
+                if limit:
+                    query.write(f" LIMIT {limit}")
                 full_query = query.getvalue()
-                if full_query.endswith(" and "):
-                    full_query = full_query[:-5]
-            return full_query + ";"
+
+        return full_query + ";"
```
jond01: Same comment as above. The whole point of using `StringIO` is to avoid string concatenation

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -260,13 +325,65 @@ def read_predictions(
...
+            )
+            aggregation_window = "10m"
```
jond01: It is better to keep it as a constant in the module

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -508,7 +494,8 @@ def read_metrics_data(
         )
 
         logger.debug(
-            "Read a data-frame",
+            "Convert dataframe to a list of metrics or results values.",
```
jond01:
```suggestion
            "Converting a data-frame to a list of metrics or results values",
```

server/api/api/endpoints/model_endpoints.py
```diff
@@ -377,7 +377,10 @@ async def get_model_endpoint_monitoring_metrics(
                 _wrap_coroutine_in_list(
                     run_in_threadpool(
                         mlrun.model_monitoring.get_tsdb_connector(
-                            project=project
+                            project=project,
+                            secret_provider=server.api.crud.secrets.get_project_secret_provider(
+                                project=project
```
jond01, line 382: Can't it be handled inside the function by default?

tests/model_monitoring/test_tdengine.py
```diff
@@ -237,3 +238,127 @@ def test_get_records_query(
...
+            "limit",
+            "agg",
+            "sliding_window_step",
+        ),
+        [
+            (
+                "subtable_1",
+                ["column2"],
+                datetime.datetime.now().astimezone() - datetime.timedelta(hours=2),
+                datetime.datetime.now().astimezone() - datetime.timedelta(hours=1),
```
jond01: In tests it's OK, but note that the difference between the start and the end time here is a bit longer than 1hr.
It's always better to use `end = mlrun.utils.datetime_now()` and then subtract a `timedelta` to get the start time.

jond01 (approved):
Looks good, I had some minor comments.

My major question is about documentation: I think that installing TDEngine and setting the env vars has to be documented. If I want to test it myself, it would be hard to understand what I need to configure.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
Eyal-Danieli: I'll remove it

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
Eyal-Danieli: changed to `agg_func`, I think it's a good indication of aggregated functions. LMK if you think it's better to use `aggregations`.

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
Eyal-Danieli: beauty is in the eye of the beholder ;) 
anyway as `end` and `start` are actually mandatory then we don't need this clause

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
Eyal-Danieli: yeah as Gal mentioned it's an overkill to use StringIO here, so anyway it has been fixed.

server/api/api/endpoints/model_endpoints.py
Eyal-Danieli: but in this case we will have to check each time if we are on API or not. The best practice for handling these secrets is using the secret provider (already went through it with Saar).

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
jond01: Maybe `agg_funcs` to hint for the multi-func option.

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
jond01: > beauty is in the eye of the beholder ;)

I disagree in this case 😆 For sure it's a "magic" number

tests/system/model_monitoring/test_app.py
```diff
@@ -99,6 +99,10 @@ class _V3IORecordsChecker:
 
     @classmethod
     def custom_setup(cls, project_name: str) -> None:
+        # By default, the TSDB connection is based on V3IO TSDB
+        # To use TDEngine, set the `TSDB_CONNECTION` environment variable to the TDEngine connection string,
+        # e.g. "taosws://user:password@host:port"
```
jond01, line 104: That's good once the DB is available.
What do you think about docs for setting this DB up?

Eyal-Danieli: amm we can add it to internal docs (such as `dev-docs`) but I don't think we should add it to MLRun docs. TDEngine has its own docs on how to deploy it with different approaches (docker, k8s, ...).

jond01: Ok, internal documentation is acceptable if we provide it out of the box in CE 👍

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
gtopper: I'm not sure who "instructed" this, but you'll see that this is not how things are done in mlrun, e.g. https://github.com/mlrun/mlrun/pull/5600/files#diff-e62fc7d3375a15b22fd77c96b2fdaeadca2127a0e47facbdbb3e63843a936adbR27.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -251,7 +274,49 @@ def read_metrics_data(
...
+            timestamp_column=mm_schemas.WriterEvent.END_INFER_TIME,
+        )
+
+        df[mm_schemas.WriterEvent.END_INFER_TIME] = pd.to_datetime(
+            df[mm_schemas.WriterEvent.END_INFER_TIME]
+        )
+        df.set_index(mm_schemas.WriterEvent.END_INFER_TIME, inplace=True)
+
+        logger.debug(
+            "Converting a data-frame to a list of metrics or results values",
```
gtopper:
```suggestion
            "Converting a DataFrame to a list of metrics or results values",
```

gtopper: I see it's referred to as a data-frame in a bunch of places. As far as I know, that's not the correct nomenclature.

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
```diff
@@ -125,33 +125,66 @@ def _get_subtables_query(
...
-                if full_query.endswith(" and "):
-                    full_query = full_query[:-5]
-            return full_query + ";"
+            if limit or interval:
+                if interval:
+                    query.write(f" INTERVAL({interval})")
+                if sliding_window_step:
+                    query.write(f" SLIDING({sliding_window_step})")
+                if limit:
+                    query.write(f" LIMIT {limit}")
```
gtopper: Hm, do we need the outer if?

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
gtopper: Those two could be extracted to the class though, rather than duplicating the whole thing.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -186,34 +187,56 @@ def get_model_endpoint_real_time_metrics(
...
+        start: datetime,
+        end: datetime,
         columns: typing.Optional[list[str]] = None,
-        filter_query: str = "",
+        filter_query: typing.Optional[str] = None,
+        interval: typing.Optional[str] = None,
+        agg_funcs: typing.Optional[list] = None,
+        limit: typing.Optional[int] = None,
+        sliding_window_step: typing.Optional[str] = None,
         timestamp_column: str = mm_schemas.EventFieldType.TIME,
```
gtopper, line 198: This one is also missing from the base class.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -260,13 +325,65 @@ def read_predictions(
...
+
+        full_name = mlrun.model_monitoring.helpers.get_invocations_fqn(self.project)
+
+        if df.empty:
+            return mm_schemas.ModelEndpointMonitoringMetricNoData(
+                full_name=full_name,
+                type=mm_schemas.ModelEndpointMonitoringMetricType.METRIC,
+            )
+
+        if not aggregation_window:
```
gtopper: I still don't think it's consistent with TSDB... Maybe just make it mandatory?

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -413,6 +413,10 @@ def get_records(
                 f"Table '{table}' does not exist in the tables list of the TSDB connector. "
                 f"Available tables: {list(self.tables.keys())}"
             )
+
+        if agg_funcs:
+            # Frame client expects the aggregators to be a string, separated by commas
```
```suggestion
            # Frames client expects the aggregators to be a comma-separated string
```

tests/model_monitoring/test_tdengine.py
```diff
@@ -186,16 +186,16 @@ def test_get_subtables(
                 "subtable_1",
                 [],
                 "",
-                datetime.datetime.now().astimezone() - datetime.timedelta(hours=1),
-                datetime.datetime.now().astimezone(),
+                mlrun.utils.datetime_now() - datetime.timedelta(hours=1),
```
gtopper, line 189: Here `astimezone()` is removed, but new lines below are added with `datetime.datetime.now().astimezone()`. Does this inconsistency make sense? Do we need the local time zone or naive timestamps?

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
Eyal-Danieli: I'll change it back to DataFrame and that's it 
https://github.com/mlrun/mlrun/pull/5642#discussion_r1618537660

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
Eyal-Danieli: good catch, I guess not because we are already handling "invalid" use cases before that

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
Eyal-Danieli: it's not mandatory with V3IO so that's why I kept the same behaviour here as well. I think it's OK because in that case when you call `read_predictions()` and get just a single value which is the total amount of predictions in the provided time range.

tests/model_monitoring/test_tdengine.py
Eyal-Danieli: yeah it doesn't really matter in that test scenario but as you mentioned I'll remove `astimezone()` from the new lines as well to keep consistency

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -209,3 +236,141 @@ def read_prediction_metric_for_endpoint_if_exists(
...
+    ) -> list[
+        typing.Union[
+            mm_schemas.ModelEndpointMonitoringMetricValues,
+            mm_schemas.ModelEndpointMonitoringMetricNoData,
+        ]
+    ]:
+        """
+        Parse a time-indexed DataFrame of metrics from the TSDB into a list of
+        metrics values per distinct results.
+        When a metric is not found in the data-frame, it is represented in a no-data object.
```
gtopper: Here too, data-frame -> DataFrame.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -260,13 +325,69 @@ def read_predictions(
...
     ) -> typing.Union[
         mm_schemas.ModelEndpointMonitoringMetricValues,
         mm_schemas.ModelEndpointMonitoringMetricNoData,
     ]:
-        raise NotImplementedError
+        if (agg_funcs and not aggregation_window) or (
+            aggregation_window and not agg_funcs
+        ):
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "Both `aggregation_window` and `agg_funcs` must be provided or neither of them."
```
gtopper: I think the convention is no capitalization and no period at the end of the sentence
```suggestion
                "both or neither of `aggregation_window` and `agg_funcs` must be provided"
```

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -413,17 +413,30 @@ def get_records(
...
             filter=filter_query,
+            aggregation_window=interval,
+            aggregators=agg_funcs,
+            limit=limit,
+            step=sliding_window_step,
             **kwargs,
         )
 
+        if limit:
+            # Todo: remove this when the issue is fixed in the frames client
```
gtopper: * Todo -> TODO
* This is not an issue in frames client. It's not implemented in the server. Which probably means it won't be fixed any time soon. And if it does get fix, you won't be able to rely on it because it will only be in new enterprise versions.

Eyal-Danieli: I removed the comment because it's meaningless if we are not going to fix it anyway

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
Eyal-Danieli: now that `_get_records` it's a private function, we don't have to add this `timestamp_column ` to the other connector.

gtopper (approved):

PR #5401, opened 2024-04-08
[Docs] Add/update model monitoring

jillnogold:

PR #5644, opened 2024-05-28
[Docs] Add/update model monitoring [1.6.x]

assaf758:
[Docs] Add/update model monitoring (#5401) [1.6.x]


docs/monitoring/model-monitoring.md
```diff
@@ -0,0 +1,170 @@
...
+```
+
+You can use the MLRun built-in class, `EvidentlyModelMonitoringApplicationBase`, to integrate [Evidently](https://github.com/evidentlyai/evidently) as an MLRun function and create MLRun artifacts.
+See the [Model monitoring and drift detection tutorial](../tutorials/05-model-monitoring).
+
+## Invoke the model again
+
+The controller checks for new datasets every `base_period` to send to the app. Invoking the model a second time ensures that the previous 
+window closed and therefore the data contains the full monitoring window. From this point on, the applications are triggered by the controller. 
+From this point on, the applications are triggered by the controller. The controller checks the Parquet DB every 10 minutes (or non-default
```
Eyal-Danieli: Duplicated...
```suggestion
The controller checks the Parquet DB every 10 minutes (or non-default 
```

Eyal-Danieli (approved):

PR #5645, opened 2024-05-28
[Docs] Fix docstring format/publish

jillnogold:


mlrun/db/httpdb.py
```diff
@@ -3989,10 +3996,11 @@ def store_alert_config(
     ) -> AlertConfig:
         """
         Create/modify an alert.
+
         :param alert_name: The name of the alert.
         :param alert_data: The data of the alert.
-        :param project: the project that the alert belongs to.
-        :return: The created/modified alert.
+        :param project:    The project that the alert belongs to.
+        :return:           The created/modified alert.
```
TomerShor: I see the `param`s and `return` sometimes has a blank line between them and sometimes not.
What is the correct format and is it displayed differently?

mlrun/db/httpdb.py
```diff
@@ -4012,9 +4020,11 @@ def store_alert_config(
     def get_alert_config(self, alert_name: str, project="") -> AlertConfig:
         """
         Retrieve an alert.
+
         :param alert_name: The name of the alert to retrieve.
-        :param project: The project that the alert belongs to.
-        :return: The alert object.
+        :param project:    The project that the alert belongs to.
+
+        :returns:           The alert object.
```
TomerShor, line 4030: Sometime you use `return` and sometimes `returns`. What is the correct format and are both options displayed differently?

mlrun/db/httpdb.py
jillnogold: The blank line (or not) does not affect the display.

mlrun/db/httpdb.py
jillnogold: Updated all to [returns](url)

TomerShor (approved):

TomerShor (approved):

PR #5646, opened 2024-05-28
[Function] Optimise an error when function deletion is failed

rokatyy:
Currently, the error string can be too long, so I have removed the additional string about failure for every Nuclio function. There will still be a message indicating that the deletion failed, including the MLRun function name.

So, instead of this 

```
Failed to delete function func-nuclio. Errors: Failed to delete nuclio function run-nuclio-hello-world-func-nuclio: 412, message='Precondition Failed', url=URL('http://nuclio-dashboard.default-tenant.svc.cluster.local:8070/api/functions/'): : Function is "
```

we will have:

```
Failed to delete function func-nuclio. Errors: Function is used by api gateways"
```

Jira - https://iguazio.atlassian.net/browse/ML-6552

### DIFF
--- a/server/api/api/utils.py
+++ b/server/api/api/utils.py
@@ -1397,7 +1397,7 @@ async def _delete_nuclio_functions_in_batches(
                 nuclio_name, error_message = result
                 if error_message:
                     failed_requests.append(
-                        f"Failed to delete nuclio function {nuclio_name}: {error_message}"
+                        error_message
                     )
 
     return failed_requests
--- a/server/api/api/utils.py
+++ b/server/api/api/utils.py
@@ -1396,8 +1396,6 @@ async def _delete_nuclio_functions_in_batches(
             if isinstance(result, tuple):
                 nuclio_name, error_message = result
                 if error_message:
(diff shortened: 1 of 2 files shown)


server/api/api/utils.py
```diff
@@ -1396,8 +1396,6 @@ async def delete_function(
             if isinstance(result, tuple):
                 nuclio_name, error_message = result
                 if error_message:
-                    failed_requests.append(
-                        f"Failed to delete nuclio function {nuclio_name}: {error_message}"
-                    )
+                    failed_requests.append(error_message)
```
TomerShor, line 1399: See [my comment on JIRA](https://iguazio.atlassian.net/browse/ML-6552?focusedCommentId=151269), and try to extract the relevant info from the exception

TomerShor (changes requested):

TomerShor:
@rokatyy can you updated the description with the new error after your latest change? I'd like to see how it look

rokatyy:
@TomerShor updated

TomerShor (approved):

PR #5647, opened 2024-05-28
[Docker] Update docker ignore - clouds ymls

tomerm-iguazio:
[ML-6548](https://iguazio.atlassian.net/browse/ML-6548)

[ML-6548]: https://iguazio.atlassian.net/browse/ML-6548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.dockerignore
```diff
@@ -17,3 +17,17 @@ kubeconfig
...
+**/*.coverage
+**/.aws-secrets
+**/credentials.txt
+**/credentials.json
+**/test-alibaba-oss.yml
+**/test-azure-blob.yml
+**/test-aws-s3.yml
+**/test-google-cloud-storage.yml
+**/test-dbfs-store.yml
+**/*.pyc
```
assaf758: why do you remove pyc?

assaf758 (commented):
a lot of tests fail in 2 suits - please go over it

.dockerignore
tomerm-iguazio: *.pyc is compiled files (in binary) from python scripts, we do not need them in the image.

.gitignore
```diff
@@ -51,3 +51,4 @@ mlrun_bc_*.json
 
 # used for local development.
 playground
+patch-env.yml
```
liranbg: newline

.dockerignore
assaf758: Since it's not related to security (correct me if I'm wrong) there is no reason to add it here "on the go".
I'm not sure if there is no implications to pyc removal (performance? storage space?) - which is yet another reason not to include it in this PR.

.gitignore
alonmr: `automation/patch_igz/patch_env.yml` is in line 41

tomerm-iguazio: what I wrote is patch-env.yml with '-' in the middle instead of existing ignore with '_'.
Currently the file is as I described, but I do not know if it had another name before that, so let's keep them both.

assaf758: you can replace the two entries with `patch[_-]env.yml`

.dockerignore
```diff
@@ -18,5 +18,18 @@ kubeconfig
...
+**/.aws-secrets
+**/credentials.txt
+**/credentials.json
+**/test-alibaba-oss.yml
+**/test-azure-blob.yml
+**/test-aws-s3.yml
+**/test-google-cloud-storage.yml
+**/test-dbfs-store.yml
+
+**/patch-env.yml
```
assaf758: can join the two entries to `**/patch[-_]env.yml`

tomerm-iguazio: changed to **/patch[_\-]env.yml because "-" is  a special character.

assaf758 (approved):

PR #5648, opened 2024-05-28
[Artifacts] Improve list artifact tags db query

TomerShor:
When listing artifact tags, we first loaded all of the artifacts in the project to the API's memory, then iterated over them and looked for tags in their metadata.
This resulted with a major load on the API and in systems with a large number of artifacts, requests to `GET /artifact-tags` were stuck and timing out.

To fix it, instead of loading the entire artifacts table then iterating over it, we use a simple join on the artifact tags table and query only the artifact key (needed for permission filtering) and the tag name.
This results with a much quicker query and simpler code.

Resolves https://iguazio.atlassian.net/browse/ML-6611

### DIFF
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -836,32 +836,28 @@ class SQLDB(DBInterface):
 
     def list_artifact_tags(
         self, session, project, category: mlrun.common.schemas.ArtifactCategories = None
-    ) -> list[tuple[str, str, str]]:
+    ) -> list[str]:
         """
-        :return: a list of Tuple of (project, artifact.key, tag)
-        """
+        List all tags for artifacts in the DB
 
-        # query artifact key and tag from the db
+        :param session: DB session
+        :param project: Project name
+        :param category: Artifact category to filter by
+
+        :return: a list of distinct tags
+        """
         query = (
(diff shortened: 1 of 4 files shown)


liranbg (approved):

server/api/db/sqldb/db.py
```diff
@@ -840,18 +840,26 @@ def list_artifact_tags(
...
         :return: a list of Tuple of (project, artifact.key, tag)
         """
-        artifacts = self.list_artifacts(session, project=project, category=category)
+
+        # query artifact key and tag from the db
+        query = (
+            self._query(session, ArtifactV2, project=project)
+            .join(ArtifactV2.Tag, ArtifactV2.Tag.obj_id == ArtifactV2.id)
+            .with_entities(ArtifactV2.key, ArtifactV2.Tag.name)
+        )
```
liranbg: will that ensure distinct tag names?
also, why not query over tags and join over artifacts? because then the join will start with less entities to cross over

liranbg (commented):
post approve thoughts

server/api/db/sqldb/db.py
TomerShor: updated the query

liranbg (approved):

PR #5651, opened 2024-05-29
[CI] Automate release pipeline-adapters

liranbg:
MLRun's pipeline adapters was developed under a private pypi account now moved under Iguazio's official account.
In addition, This PR introduces a Github workflow to trigger manual releases upon need.

The current proposed flow is
1. Upon MLRun's release - the minor version would be bumped, merged and triggered for release
2. During MLRun's release cycle - the patch version would be bumped on PR, merged and only then triggered for release

Any specific need for release would be handled manually.

### DIFF
--- a/.github/workflows/release-pipeline-adapters.yaml
+++ b/.github/workflows/release-pipeline-adapters.yaml
@@ -1,0 +1,85 @@
+# Copyright 2023 Iguazio
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name: Release Pipeline Adapters
+on:
+  workflow_dispatch:
+    inputs:
+      packages:
+        description: 'Comma-separated list of packages to release'
+        default: '*'
+        options:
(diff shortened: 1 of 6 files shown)


.github/workflows/release-pipeline-adapters.yaml
```diff
@@ -0,0 +1,85 @@
...
+  workflow_dispatch:
+    inputs:
+      packages:
+        description: 'Comma-separated list of packages to release'
+        default: '*'
+        options:
+          - '*'
+          - 'mlrun-pipelines-kfp-common'
+          - 'mlrun-pipelines-kfp-v1-8'
+          - 'mlrun-pipelines-kfp-v2'
```
quaark: Will this give us a dropdown of options to select?
If so, the description isn't accurate as it's not a Comma-separated list of packages to release

.github/workflows/release-pipeline-adapters.yaml
```diff
@@ -0,0 +1,85 @@
...
+        with:
+          python-version: "3.9"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip build
+      - name: Build package distributions
+        run: |
+          cd pipeline-adapters
+          make build
```
quaark, line 76: Set the package here for make build, You are currently building all packages and releasing one of them in each of the matrix strategies

quaark (changes requested):
Looks good, minor workflow comments

.github/workflows/release-pipeline-adapters.yaml
liranbg: will update description

.github/workflows/release-pipeline-adapters.yaml
liranbg: I override the the env with `PACKAGES: ${{ matrix.name }}` so it will build only one adapter

quaark (approved):

PR #5653, opened 2024-05-29
[Coding Conventions] Enforce sending self as first arg of instance methods

Yacouby:
https://iguazio.atlassian.net/browse/ML-6626

This PR addresses  the 3rd python coding convention rule of the conventions that we defined:
"Use self as the first parameter in instance methods."

### DIFF
--- a/mlrun/common/schemas/model_monitoring/model_endpoints.py
+++ b/mlrun/common/schemas/model_monitoring/model_endpoints.py
@@ -104,6 +104,7 @@ class ModelEndpointSpec(ObjectSpec):
         )
 
     @validator("monitor_configuration")
+    @classmethod
     def set_name(cls, monitor_configuration):
         return monitor_configuration or {
             EventFieldType.DRIFT_DETECTED_THRESHOLD: (
@@ -115,6 +116,7 @@ class ModelEndpointSpec(ObjectSpec):
         }
 
     @validator("model_uri")
+    @classmethod
     def validate_model_uri(cls, model_uri):
         """Validate that the model uri includes the required prefix"""
         prefix, uri = mlrun.datastore.parse_store_uri(model_uri)
--- a/mlrun/model_monitoring/db/stores/sqldb/models/mysql.py
+++ b/mlrun/model_monitoring/db/stores/sqldb/models/mysql.py
@@ -60,7 +60,7 @@ class _ApplicationResultOrMetric:
     )
 
     @declared_attr
-    def endpoint_id(cls):
+    def endpoint_id(self):
         return Column(
             String(40),
             ForeignKey(f"{EventFieldType.MODEL_ENDPOINTS}.{EventFieldType.UID}"),
@@ -81,7 +81,7 @@ class ApplicationMetricsTable(
 
 class MonitoringSchedulesTable(Base, MonitoringSchedulesBaseTable):
     @declared_attr
-    def endpoint_id(cls):
+    def endpoint_id(self):
         return Column(
             String(40),
             ForeignKey(f"{EventFieldType.MODEL_ENDPOINTS}.{EventFieldType.UID}"),
--- a/tests/model_monitoring/test_stores/test_sql.py
+++ b/tests/model_monitoring/test_stores/test_sql.py
@@ -49,6 +49,7 @@ class TestSQLStore:
     def store_connection(tmp_path: Path) -> str:
         return f"sqlite:///{tmp_path / 'test.db'}"
 
+    @classmethod
     @pytest.fixture()
     def _mock_random_endpoint(
         cls,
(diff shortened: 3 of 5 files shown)


mlrun/model_monitoring/db/stores/sqldb/models/mysql.py
```diff
@@ -52,7 +52,7 @@ class ApplicationResultTable(Base, ApplicationResultBaseTable):
     )
 
     @declared_attr
-    def endpoint_id(cls):
+    def endpoint_id(self):
```
jond01, line 63: To the defense of the original author, the official SQLAlchemy examples are confusing:
https://docs.sqlalchemy.org/en/20/orm/mapping_api.html#sqlalchemy.orm.declared_attr

But they all use `cls`, and in one correct example, they specify the `@classmethod` decorator.
However, looks like it's supported only from version 2.0, and we use 1.4 :(

In short - there's no good solution here.

tests/model_monitoring/test_stores/test_sql.py
```diff
@@ -47,7 +47,7 @@ def store_connection(tmp_path: Path) -> str:
 
     @pytest.fixture()
     def _mock_random_endpoint(
-        cls,
+        self,
```
jond01: Can be a class method

tests/model_monitoring/test_stores/test_sql.py
```diff
@@ -131,7 +131,7 @@ def new_sql_store(cls, store_connection: str) -> Iterator[SQLStoreBase]:
             assert (len(list_of_endpoints)) == 0
 
     def test_sql_target_list_model_endpoints(
-        cls,
+        self,
```
jond01: Can be a `@staticmethod`. I updated it and the methods below in #5641

tests/model_monitoring/test_stores/test_sql.py
```diff
@@ -163,7 +163,7 @@ def test_sql_target_list_model_endpoints(
         assert len(filtered_list_of_endpoints) == 1
 
     def test_sql_target_patch_endpoint(
-        cls,
+        self,
```
jond01: Can be a `@staticmethod`

tests/model_monitoring/test_stores/test_sql.py
```diff
@@ -192,7 +192,7 @@ def test_sql_target_patch_endpoint(
         assert endpoint_dict["error_count"] == 2
 
     def test_sql_write_application_event(
-        cls,
+        self,
```
jond01: Can be a `@classmethod`

mlrun/common/schemas/model_monitoring/model_endpoints.py
```diff
@@ -104,6 +104,7 @@ def from_flat_dict(cls, endpoint_dict: dict, json_parse_values: list = None):
         )
 
     @validator("monitor_configuration")
+    @classmethod
     def set_name(cls, monitor_configuration):
```
jond01, line 108: Again on another dependency.
The official documentation of the legacy version we use contains this mistake:
https://docs.pydantic.dev/1.10/usage/validators/

I assume it works even in version 1.10 as I see it in other places.

jond01 (commented):
Looks good! Please address the comments in `test_sql.py`.

tests/model_monitoring/test_stores/test_sql.py
Yacouby: this method uses the decorator: `pytest.fixture`...
`pytest.fixture` and `classmethod ` are not designed to work together.

jond01: It used to not work together, but it is fixed in the current version of pytest that we use.
For example:
```py
import pytest


class Test:
    @classmethod
    @pytest.fixture()
    def my_fixt(cls) -> int:
        return 1

    @staticmethod
    def test(my_fixt: int) -> None:
        assert my_fixt == 1
```

tests/model_monitoring/test_stores/test_sql.py
jond01: Why is it marked as resolved without any comment or change?
Anyway, the PR I mentioned above is in dev. You can remove the changes to this file except for the fixture.

tests/model_monitoring/test_stores/test_sql.py
```diff
@@ -50,6 +50,7 @@ def store_connection(tmp_path: Path) -> str:
         return f"sqlite:///{tmp_path / 'test.db'}"
 
     @pytest.fixture()
+    @classmethod
```
jond01: Only this order works:
```suggestion
    @classmethod
    @pytest.fixture()
```

jond01 (approved):
Awesome 😎

PR #5655, opened 2024-05-29
[Artifacts] Improve list artifacts querying

TomerShor:
Previously, when listing artifacts we did several intensive operations against the DB:

1. When asking a specific tag, we ran and SQL query on the tags, than filtered the artifacts on their UIDs from the returned tags.
2. For adding tags to the artifact object, we:
    1. Queried the db for all artifacts with the given filters
    2. Iterated over the returned artifacts 
    3. For each artifact, query the DB for its tags and add them to the struct.

All of the above resulted in a lot of overhead and too many queries.

Now we:

1. Query all artifacts, joined* on the artifact tags, filtered by tag name. (This will already return the permutations)
2. For each artifact add the tag to its metadata.
3. Moved the `producer_uri` filtering to the same for loop as (2) to not iterate over all artifacts again.

*outer join if no specific tag was requested so we return artifacts that are tag-less, and a simpler join when querying on a specific tag as in that case tag-less functions are irrelevant.

https://iguazio.atlassian.net/browse/ML-6642


server/api/db/sqldb/db.py
```diff
@@ -793,7 +807,7 @@ def del_artifacts(
         project = project or config.default_project
         distinct_keys = {
             artifact.key
-            for artifact in self._find_artifacts(
+            for artifact, _ in self._find_artifacts(
```
quaark: dedup here as well I think

server/api/db/sqldb/db.py
```diff
@@ -2577,7 +2560,8 @@ def _get_project_record(
         return project_record
 
     def verify_project_has_no_related_resources(self, session: Session, name: str):
-        artifacts = self._find_artifacts(session, name, "*")
+        artifact_records = self._find_artifacts(session, name, "*")
+        artifacts = [artifact for artifact, _ in artifact_records]
```
quaark: Since you do this many times, maybe it should be a facility of find_artifacts?
`attatch_tags=True/False` or something

quaark (changes requested):
Looks good! Couple of things

server/api/db/sqldb/db.py
TomerShor: not relevant anymore

server/api/db/sqldb/db.py
```diff
@@ -683,20 +684,34 @@ def list_artifacts(
             uid=uid,
             producer_id=producer_id,
             best_iteration=best_iteration,
-            producer_uri=producer_uri,
+            most_recent=most_recent,
+            attach_tags=True,
```
quaark:
```suggestion
            attach_tags=not as_records,
```

Then you can:
```
if as_records:
    return artifact_records
```

quaark (changes requested):
One more minor thing

quaark (approved):
👑

PR #5663, opened 2024-05-30
[Builder] Log build pod events

alonmr:
Log build pod events to reflect pod initialization errors such as mount errors.
This PR is coupled with https://github.com/v3io/helm-charts/pull/1028

Output example:
Screenshot 2024-10-21 at 19 41 29

https://iguazio.atlassian.net/browse/ML-6723
https://iguazio.atlassian.net/browse/ML-6527


mlrun/db/base.py
```diff
@@ -776,6 +776,7 @@ def get_builder_status(
         self,
         func: "mlrun.runtimes.BaseRuntime",
         offset: int = 0,
+        events_offset: int = 0,
```
TomerShor: move to last argument, in case there are calls with ordered args instead of named args.

mlrun/runtimes/pod.py
```diff
@@ -1412,20 +1413,29 @@ def _build_watch(
     ):
         db = self._get_db()
         offset = 0
+        events_offset = 0
         try:
-            text, _ = db.get_builder_status(self, 0, logs=logs)
+            text, _, deploy_status_text_kind = db.get_builder_status(
+                self, 0, 0, logs=logs
+            )
```
TomerShor:
```suggestion
            text, _, deploy_status_text_kind = db.get_builder_status(
                self, 
                offset=0, 
                events_offset=0, 
                logs=logs,
            )
```

mlrun/runtimes/pod.py
```diff
@@ -1434,14 +1444,24 @@ def print_log(text):
                 time.sleep(2)
                 if show_on_failure:
                     text = ""
-                    db.get_builder_status(self, 0, logs=False)
+                    db.get_builder_status(self, 0, 0, logs=False)
```
TomerShor:
```suggestion
                    db.get_builder_status(
                        self, 
                        offset=0, 
                        events_offset=0, 
                        logs=logs,
                    )
```

server/api/api/endpoints/functions.py
```diff
@@ -580,6 +592,40 @@ def _handle_job_deploy_status(
...
+            function_name=name,
+            pod=pod,
+            events_len=len(build_pod_events),
+        )
+        resp = ""
+        for event in build_pod_events:
+            event: kubernetes.client.CoreV1Event
+            if event.type == "Normal":
+                # Do not spam with normal events
+                continue
```
TomerShor, line 618: Is this the only OK event type?
So we want just print events that are not interesting.

server/api/utils/helpers.py
```diff
@@ -258,3 +258,36 @@ def merge_schedule_and_schedule_object_labels(
...
+            client_version=client_version,
+            min_versions=min_versions,
+        )
+        return False
+
+    parsed_min_versions.sort(reverse=True)
+    for parsed_min_version in parsed_min_versions:
+        if parsed_current_version < parsed_min_version:
+            return False
+    return True
```
TomerShor: Isn't this a dup for IGZ and Nuclio versions?

server/api/utils/singletons/k8s.py
```diff
@@ -832,7 +816,22 @@ def get_secret_data(self, secret_name, namespace=""):
...
+        return self._list_events(
+            namespace=namespace, field_selector=f"involvedObject.name={object_name}"
+        )
+
+    @raise_for_status_code
+    def _list_events(self, namespace=None, field_selector=""):
+        resp = self.v1api.list_namespaced_event(
+            self.resolve_namespace(namespace), field_selector=field_selector
+        )
+        return resp.items
```
TomerShor, line 908: Why a wrapper for a private method?

TomerShor (changes requested):
Overall looking good!
Left some questions

server/api/api/endpoints/functions.py
alonmr: according to https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/event-v1/
> type is the type of this event (Normal, Warning), new types could be added in the future. It is machine-readable. This field cannot be empty for new Events.

server/api/utils/singletons/k8s.py
alonmr: you mean decorator? It can be moved to `list_object_events` though I don't see a reason at this time

TomerShor: I meant wrapper, as `list_object_events` doesn't do anything other than calling `_list_events`.
So are two functions needed here and not just have the implementation in `list_object_events`?

server/api/api/endpoints/functions.py
```diff
@@ -580,6 +592,40 @@ def _handle_job_deploy_status(
...
+
+            resp += f"""
+Event: {event.metadata.name}
+Type: {event.type}
+Reason: {event.reason}
+Message: {event.message}
+"""
```
liranbg: I think it is worth adding timestamp so we can understand time gaps between events. wdyt? maybe it is redundant if it is being logged using logger

server/api/utils/singletons/k8s.py
```diff
@@ -935,12 +935,58 @@ def _resolve_secret_labels(self, username):
...
         return "".join(
             random.choice(string.ascii_lowercase + string.digits) for _ in range(length)
         )
 
+    def _init_k8s_config(self, log=True):
+        try:
+            config.load_incluster_config()
+            self.running_inside_kubernetes_cluster = True
+            if log:
+                logger.info("Using in-cluster config.")
```
liranbg, line 972:
```suggestion
                logger.info("Using in-cluster config")
```

server/api/utils/singletons/k8s.py
```diff
@@ -935,12 +935,58 @@ def _resolve_secret_labels(self, username):
...
+        except Exception:
+            try:
+                config.load_kube_config(self.config_file)
+                self.running_inside_kubernetes_cluster = True
+                if log:
+                    logger.info("Using local kubernetes config.")
```
liranbg, line 978:
```suggestion
                    logger.info("Using local kubernetes config")
```

liranbg (changes requested):
nicely done. you're missing ut coverage for listing the pod events. I suggest doing a small coverage for the flow

server/api/utils/singletons/k8s.py
alonmr: when logging without kwargs I always use a period I think it looks better. Think we should align this convention?

server/api/api/endpoints/functions.py
alonmr: good point, we print the text without the logger

server/api/utils/singletons/k8s.py
alonmr: It's not needed but it's to allow reusing `_list_events` without forcing the field selector.
There are other jira tickets that will use this in the future.

TomerShor (approved):

liranbg (approved):

PR #5668, opened 2024-05-30
[CI] Optimize test integration runtime

liranbg:
Reducing the time to spin up and down mlrun db for each test case allowing them to be reused each test suite.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -576,17 +576,16 @@ run-test-db:
 	docker rm test-db --force || true
 	docker run \
 		--name=test-db \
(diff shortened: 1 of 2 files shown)


TomerShor (commented):
Code looks good, but aren't there tests that run in parallel at all?
If so, we might encounter issues when trying to reset the db

liranbg:
> Code looks good, but aren't there tests that run in parallel at all? If so, we might encounter issues when trying to reset the db

no parallel for integration.

TomerShor (approved):

PR #5670, opened 2024-06-02
[K8s] Fix validating label keys

liranbg:
Previously we have validated for max length of the entire key which is wrong according to k8s spec https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set whereas the key can be split into 2 segments in case of a "/" delimiter. prefix and name. the name is indeed limited to 64 chars while the prefix can be longer and up to 254 chars. 

This PR also indicating the key has been truncated in error in case it exceeds the default 64 chars.

https://iguazio.atlassian.net/browse/ML-5872

### DIFF
--- a/mlrun/k8s_utils.py
+++ b/mlrun/k8s_utils.py
@@ -165,6 +165,11 @@ def verify_label_key(key: str):
     mlrun.utils.helpers.verify_field_regex(
         f"project.metadata.labels.'{key}'",
         name,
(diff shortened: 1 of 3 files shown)


mlrun/k8s_utils.py
TomerShor, line 1: Where do you validate the `name` part (after the `/`) is <= 63 characters?

TomerShor (approved):

PR #5671, opened 2024-06-02
[Application] Populate resource configuration to sidecar

TomerShor:
When setting requests / limits on an application runtime, we expect these resources to be applied on the sidecar container as well.
In this PR we populate the sidecar with the application resources before deploying.

Resolves https://iguazio.atlassian.net/browse/ML-6526

### DIFF
--- a/tests/runtimes/test_application.py
+++ b/tests/runtimes/test_application.py
@@ -190,6 +190,28 @@ def test_application_api_gateway(rundb_mock):
     assert function_name in api_gateway.spec.functions[0]
 
 
+def test_application_runtime_resources(rundb_mock):
+    mlrun.mlconf.igz_version = "3.6.0"
+    fn: mlrun.runtimes.ApplicationRuntime = mlrun.code_to_function(
+        "application-test",
+        kind="application",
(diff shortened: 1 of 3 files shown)


tests/runtimes/test_application.py
```diff
@@ -190,6 +190,43 @@ def test_application_api_gateway(rundb_mock):
     assert function_name in api_gateway.spec.functions[0]
 
 
+def test_application_runtime_resources(rundb_mock):
+    mlrun.mlconf.igz_version = "3.6.0"
```
liranbg: why is it needed? also, will it not corrupt consecutive tests?

TomerShor: It is due to this: https://github.com/mlrun/mlrun/blob/5d8c6d707d4711c02ae52ee4a714cf2ec0102a6e/mlrun/runtimes/nuclio/api_gateway.py#L493-L494 

I will add a setup/teardown with resetting the version to the original

TomerShor: @liranbg added fixture to avoid corruption

liranbg (approved):

PR #5683, opened 2024-06-02
[Alerts] Populate event cache upon api restart

yanburman:
We need to handle a case where alert was created and mlrun-api was restarted after that. Need to populate event cache from DB in such case

https://iguazio.atlassian.net/issues/ML-6616

### DIFF
--- a/server/api/crud/alerts.py
+++ b/server/api/crud/alerts.py
@@ -17,10 +17,12 @@ import datetime
 
 import sqlalchemy.orm
 
+import mlrun.common.schemas
 import mlrun.utils.singleton
 import server.api.api.utils
 import server.api.utils.helpers
 import server.api.utils.singletons.db
+from mlrun.config import config as mlconfig
 from mlrun.utils import logger
 from server.api.utils.notification_pusher import AlertNotificationPusher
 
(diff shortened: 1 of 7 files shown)


server/api/initial_data.py
```diff
@@ -139,6 +139,7 @@ def update_default_configuration_data():
         db = server.api.db.sqldb.db.SQLDB()
         _add_default_hub_source_if_needed(db, db_session)
         _add_default_alert_templates(db, db_session)
+        server.api.crud.Alerts().populate_event_cache(db_session)
```
liranbg: avoid using critical path which would block (or extend) mlrun startup time.

server/api/db/sqldb/db.py
```diff
@@ -4353,6 +4353,10 @@ def get_alert_template(
             self._get_alert_template_record(session, name)
         )
 
+    def get_all_alerts(self, session) -> list[mlrun.common.schemas.AlertConfig]:
+        query = self._query(session, AlertConfig)
+        return list(map(self._transform_alert_config_record_to_schema, query.all()))
```
liranbg, line 4399: since you are retrieving all db records and assume some alerts configs are bound to some heuristic amount of records, please make the query limited by that calculation (or any config value) - eg 1000 records
that way, you are protecting mlrun from overflowing its memory

server/api/crud/alerts.py
```diff
@@ -191,6 +191,14 @@ def process_event(
                 event=event_data.entity.ids[0],
             )
 
+    @staticmethod
+    def populate_event_cache(session: sqlalchemy.orm.Session):
+        for alert in server.api.utils.singletons.db.get_db().get_all_alerts(session):
+            for event_name in alert.trigger.events:
+                server.api.crud.Events().add_event_configuration(
+                    alert.project, event_name, alert.name
+                )
```
liranbg, line 226: this will result in N*M access to DB where N = alerts, M = events.
I suggest to revise the query and make it more efficient wrt accessing DB

liranbg (commented):
Some suggestions to ensure mlrun startup critical path is not compromised. its memory is managed correctly by not loading all records without putting a rough limit. and access to db is not naive

server/api/crud/alerts.py
```diff
@@ -191,6 +203,41 @@ def process_event(
                 event=event_data.entity.ids[0],
             )
 
+    def populate_event_cache(self, session: sqlalchemy.orm.Session):
+        try:
+            self._try_populate_event_cache(session)
+        except Exception as exc:
+            logger.error(
+                "Error populating event cache for alerts. Transitioning state to offline!",
+                exc=exc,
```
liranbg: `mlrun.errors.err_to_str(exc)`

liranbg (commented):
minor

liranbg (approved):

PR #5690, opened 2024-06-03
[Makefile] Specify docker repository explicitly

gtopper:
Fixes the following errors when signed into business account with Docker Desktop:
```
Error response from daemon: Head "https://registry-1.docker.io/v2/mlrun/mlrun/manifests/unstable": denied: image access restricted: mlrun/mlrun blocked for organization X
```
and
```
Error response from daemon: Head "https://registry-1.docker.io/v2/mysql/mysql-server/manifests/8.0": denied: image access restricted: mysql/mysql-server blocked for organization X
```
where X is the name of the organization.


liranbg (commented):
Mind adding to pr description the reason for this change?

assaf758 (approved):

PR #5697, opened 2024-06-03
[Notifications] Improve Remote Workflow Start Notification Functionality

quaark:

1. If a notification with the `when` being `running` is set in the notification, this notification's parameters will be used for the pipeline start notification. (Otherwise we take the params of the last notification in the list)
2. Added an option to opt out of the start notification without BC breakages. 
**TODO**: discuss if we want to change the behavior to opt in to the start notification by using a notification with `running` and not send it if it doesn't exist.


liranbg (approved):

mlrun/projects/pipelines.py
```diff
@@ -758,13 +763,20 @@ def run(
...
-        for notification in notifications or []:
-            project.notifiers.add_notification(notification.kind, notification.params)
+        # if a notification with `when=running` is provided, it will be used explicitly and others will be ignored
+        if send_start_notification:
+            for notification in notifications or []:
+                project.notifiers.add_notification(
+                    notification.kind, notification.params
+                )
+                if "running" in notification.when:
+                    break
```
rokatyy: nitpicking, but maybe it is better to move the comment below
```suggestion
        if send_start_notification:
            for notification in notifications or []:
                project.notifiers.add_notification(
                    notification.kind, notification.params
                )
                 # if a notification with `when=running` is provided, it will be used explicitly and others will be ignored
                if "running" in notification.when:
                    break
```

PR #5698, opened 2024-06-03
[Functions] Add deletion failure details

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-6553

Add deletion failure details when function is being deleted via SDK

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -1223,10 +1223,9 @@ class HTTPRunDB(RunDBInterface):
                 background_task.status.state
                 == mlrun.common.schemas.BackgroundTaskState.failed
             ):
-                reason = background_task.status.error
                 logger.info(
                     "Function deletion failed",
-                    reason=reason,
+                    reason=background_task.status.error,
                     project_name=project,
                     function_name=name,
                 )
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -1223,8 +1223,11 @@ class HTTPRunDB(RunDBInterface):
                 background_task.status.state
                 == mlrun.common.schemas.BackgroundTaskState.failed
             ):
+                reason = background_task.status.error
                 logger.info(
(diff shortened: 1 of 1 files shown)


mlrun/db/httpdb.py
```diff
@@ -1223,8 +1223,11 @@ def delete_function(self, name: str, project: str = ""):
                 background_task.status.state
                 == mlrun.common.schemas.BackgroundTaskState.failed
             ):
+                reason = background_task.status.error
                 logger.info(
-                    "Function deletion failed", project_name=project, function_name=name
+                    f"Function deletion failed. Reason: {reason}",
+                    project_name=project,
+                    function_name=name,
```
quaark:
```suggestion
                    f"Function deletion failed",
                    reason=background_task.status.error,
                    project_name=project,
                    function_name=name,
```

mlrun/db/httpdb.py
```diff
@@ -1223,8 +1223,11 @@ def delete_function(self, name: str, project: str = ""):
                 background_task.status.state
                 == mlrun.common.schemas.BackgroundTaskState.failed
             ):
+                reason = background_task.status.error
                 logger.info(
-                    "Function deletion failed", project_name=project, function_name=name
+                    f"Function deletion failed. Reason: {reason}",
```
moranbental: I think it is better to include the `reason` as an argument, similar to `project_name`

moranbental (changes requested):

quaark (commented):
We usually put parameters in the logger parameters. Not a must if there's a specific reason to put it in the log message itself though

mlrun/db/httpdb.py
rokatyy: @moranbental make sense! fixed

quaark (approved):
Great!

liranbg (approved):

moranbental (approved):
LGTM

PR #5354, opened 2024-03-31
[Docs] Update change log for V1.6.3

jillnogold:
https://iguazio.atlassian.net/browse/ML-6414

PR #5699, opened 2024-06-03
Revert "[Docs] Update change log for V1.6.3"

liranbg:
Reverts mlrun/mlrun#5354


liranbg:
revert because it should first go to 1.6.x branch

PR #5703, opened 2024-06-03
[API] Support Memray profiling

alonmr:
## Memory profiling MLRun API with Memray
This PR basically ports #5570 but in bash instead of python because we want to run the API straight from uvicorn CLI instead of a wapper python process.
Also added support to specify extra flags as some may be useful e.g. `--trace-python-allocators` to help finding leaks.

### DIFF
--- a/dockerfiles/mlrun-api/Dockerfile
+++ b/dockerfiles/mlrun-api/Dockerfile
@@ -84,10 +84,4 @@ VOLUME /mlrun/db
 # and avoid zombie processes
 ENTRYPOINT ["tini", "--"]
 
-CMD [ \
-    "uvicorn", \
-    "server.api.main:app", \
-    "--proxy-headers", \
-    "--host", "0.0.0.0", \
-    "--log-config", "server/api/uvicorn_log_config.yaml" \
-]
+CMD ["bash", "./dockerfiles/mlrun-api/start_api.sh"]
--- a/dockerfiles/mlrun-api/start_api.sh
+++ b/dockerfiles/mlrun-api/start_api.sh
@@ -12,6 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+:'       _\|/_
+         (o o)
+ +----oOO-{_}-OOo----------------------------------------------------------------------------------------------------+
+ |                                                                                                                   |
+ |  This script is run over tini to ensure capturing zombie processes and signal handling.                           |
+ |  It is important to run the API with exec so that the bash process will be replaced with                          |
+ |   the API process and for zombie processes reaping.                                                               |
+ |                                                                                                                   |
+ +-------------------------------------------------------------------------------------------------------------------+
+ '
+
 # Lower case the MLRUN_MEMRAY env var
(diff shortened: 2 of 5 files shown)


dockerfiles/mlrun-api/start_api.sh
```diff
@@ -0,0 +1,34 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Lower case the MLRUN_MEMRAY env var
+MLRUN_MEMRAY_LOWER=$(echo "$MLRUN_MEMRAY" | tr '[:upper:]' '[:lower:]')
+# Ensure 1 leading space
+MLRUN_MEMRAY_EXTRA_FLAGS=$(echo "${MLRUN_MEMRAY_EXTRA_FLAGS# }" | sed 's/^/ /')
+
+# Check if the mlrun memray is set to a true value
+if [[ -n "$MLRUN_MEMRAY_LOWER"  && ( "$MLRUN_MEMRAY_LOWER" == "1" || "$MLRUN_MEMRAY_LOWER" == "true" || "$MLRUN_MEMRAY_LOWER" == "yes" || "$MLRUN_MEMRAY_LOWER" == "on" )]]; then
```
TomerShor, line 32: A bit more readable
```suggestion
# Check if the mlrun memray is set to a true value
mlrun_memray_enabled() {
  [[ -n "$MLRUN_MEMRAY_LOWER" && ( "$MLRUN_MEMRAY_LOWER" == "1" || "$MLRUN_MEMRAY_LOWER" =~ ^(true|yes|on)$ ) ]]
}

if mlrun_memray_enabled; then
```

TomerShor (changes requested):

dockerfiles/mlrun-api/Dockerfile
```diff
@@ -84,10 +84,4 @@ VOLUME /mlrun/db
...
 ENTRYPOINT ["tini", "--"]
 
-CMD [ \
-    "uvicorn", \
-    "server.api.main:app", \
-    "--proxy-headers", \
-    "--host", "0.0.0.0", \
-    "--log-config", "server/api/uvicorn_log_config.yaml" \
-]
+CMD ["bash", "./dockerfiles/mlrun-api/start_api.sh"]
```
liranbg, line 87: since you have moved from running uvicorn under tini to bash, you may now not use tini at all and remove it altogether (see line 40).

since you are using `exec` it automatically replaces the bash with your program meaning the signals would automatically be forwarded to the process (as it replaces the bash with the program you `exec`ing on).

that being said, I would suggest is comment on bash script that `exec` is used to replace the bash as program so that events would be delegated correctly.

once you have all in, make sure that delete/delete --force works as expected and it doesnt take too much time before it goes down (==signals are effectively handled)

liranbg: scrape everything out. I think that leaving in tini and bash with exec would make perfect sense.

just leave a comment on the bash script to ensure the `exec` remains there for zombie process reaping purpose :)

PR #5705, opened 2024-06-03
[Function] Validate tag name before deploying

TomerShor:
Validate tag adheres to k8s limitations before we get to the DB.

Resolves https://iguazio.atlassian.net/browse/ML-6567

### DIFF
--- a/mlrun/runtimes/nuclio/function.py
+++ b/mlrun/runtimes/nuclio/function.py
@@ -544,6 +544,8 @@ class RemoteRuntime(KubeResource):
         if tag:
             self.metadata.tag = tag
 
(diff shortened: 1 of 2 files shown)


mlrun/runtimes/nuclio/function.py
```diff
@@ -544,6 +544,8 @@ def deploy(
         if tag:
             self.metadata.tag = tag
 
+        mlrun.utils.validate_tag_name(self.metadata.tag, "function.metadata.tag")
```
alonmr: please move to `pre_deploy_validation` - remember validations should be on server side

alonmr (approved):

PR #5706, opened 2024-06-04
[Model Monitoring] Avoid failing on missing predictions TSDB table

gtopper:
[ML-6628](https://iguazio.atlassian.net/browse/ML-6628)

Return an empty dataframe instead.

Fixes system test: `TestRecordResults::test_inference_feature_set`

[ML-6628]: https://iguazio.atlassian.net/browse/ML-6628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/system/model_monitoring/test_app.py
```diff
@@ -207,7 +207,7 @@ def _test_tsdb_record(cls, ep_id: str) -> None:
                 ), "The TSDB saved metrics are different than expected"
 
     @classmethod
-    def _test_predictions_table(cls, ep_id: str) -> None:
+    def _test_predictions_table(cls, ep_id: str, should_be_empty=False) -> None:
```
jond01: minor:
```suggestion
    def _test_predictions_table(cls, ep_id: str, should_be_empty: bool = False) -> None:
```

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -418,18 +418,24 @@ def _get_records(
...
+                end=end,
+                columns=columns,
+                filter=filter_query,
+                aggregation_window=interval,
+                aggregators=agg_funcs,
+                step=sliding_window_step,
+                **kwargs,
+            )
+        except v3io_frames.ReadError as err:
+            if "No TSDB schema file found" in str(err):
```
jond01, line 435: I believe this can be checked more directly with a V3IO KV looking for the `.#schema` file. What do you think?

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -418,18 +418,24 @@ def _get_records(
...
+                return pd.DataFrame()
```
jond01, line 436: What is the `/metrics` output on a new model endpoint?
If I follow correctly, it will always show the "invocations" metric, which is incorrect.

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
gtopper: That would add a redundant, non-transactional I/O operation without good reason.

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
gtopper: It will return an empty dataframe, and hence the expected `ModelEndpointMonitoringMetricNoData`.

jond01: I refer to the `read_prediction_metric_for_endpoint_if_exists` method, which is used in the `/metrics` endpoint.
What is the `/metrics` (not `/metrics-values`) output in the mentioned case?

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
jond01: Ok, it just feels a bit sketchy to rely on the error string.

gtopper: Not ideal, but we do it in some places, and we can't rely on a change to frames (because it's versioned as part of the enterprise system).

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
gtopper: You're right, that's an issue in development. I pushed a 1-liner commit to fix that.

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -620,7 +626,7 @@ def read_prediction_metric_for_endpoint_if_exists(
         predictions = self.read_predictions(
             endpoint_id=endpoint_id, start="0", end="now", limit=1
         )
-        if predictions:
+        if predictions.data:
```
jond01, line 629: From what I see, if this condition is False, the function returns `None`, and the metrics list will contain a `null`.

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
jond01: > What is the `/metrics` (not `/metrics-values`) output in the mentioned case?

👆

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
gtopper: Right, I see that you already added a TODO about that. I pushed a fix.

jond01 (approved):

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
jond01: This question remains unanswered - it was the primary motivation for this fix.
It should be ok now. Addressing it could reduce the review time for both sides.

PR #5709, opened 2024-06-04
[Run] Set the default run kind to local

moranbental:
When a user creates a context in local (rather than a run) from `get_or_create_context` , it creates a local run that cannot be aborted or deleted. The issue is that you can't delete the run because it doesn't have a kind.
For that, I changed the default run kind to local.

https://iguazio.atlassian.net/browse/ML-4636


mlrun/execution.py
```diff
@@ -1041,7 +1041,11 @@ def set_if_not_none(_struct, key, val):
 
         # completion of runs is not decided by the execution as there may be
         # multiple executions for a single run (e.g. mpi)
-        if self._state != "completed":
+        run_kind = self.labels.get(mlrun_constants.MLRunInternalLabels.kind, "")
```
liranbg, line 1045: update the comment above it, explaining the reasoning to allow changing run state for local

liranbg (commented):
very well, minor comment

tests/test_execution.py
```diff
@@ -52,7 +52,7 @@ def test_local_context(rundb_mock):
     run = db.read_run(context._uid, project=project_name)
 
     # run state should not be updated by the context
-    assert run["status"]["state"] == "running", "run status was updated in db"
+    assert run["status"]["state"] == "completed", "run status was updated in db"
```
liranbg: read the comment. the assertion is wrong now.

liranbg (changes requested):

mlrun/execution.py
```diff
@@ -1041,7 +1041,12 @@ def set_if_not_none(_struct, key, val):
 
         # completion of runs is not decided by the execution as there may be
         # multiple executions for a single run (e.g. mpi)
-        if self._state != "completed":
+        # If the run is local, we can change the state to completed, so it doesn't stay in running mode forever.
```
alonmr: can you make the comment a bit more coherent? something like
> Completion of runs is decided by the API runs monitoring as there may be multiple executions for a single run (e.g. mpi). For kinds that are not monitored by the API (local) we allow changing the state.

tests/integration/sdk_api/run/test_run.py
```diff
@@ -68,6 +68,10 @@ def test_ctx_run_labels(self):
         # remove the host label
         assert mlrun_constants.MLRunInternalLabels.host in runs[0]["metadata"]["labels"]
         del runs[0]["metadata"]["labels"][mlrun_constants.MLRunInternalLabels.host]
+
+        # remove the kind label
+        del runs[0]["metadata"]["labels"][mlrun_constants.MLRunInternalLabels.kind]
+
```
alonmr:
```suggestion
        # remove internal labels
        assert mlrun_constants.MLRunInternalLabels.host in runs[0]["metadata"]["labels"]
        del runs[0]["metadata"]["labels"][mlrun_constants.MLRunInternalLabels.host]
        assert mlrun_constants.MLRunInternalLabels.kind in runs[0]["metadata"]["labels"]
        del runs[0]["metadata"]["labels"][mlrun_constants.MLRunInternalLabels.kind]

```

tests/run/assets/simple.py
```diff
@@ -12,5 +12,13 @@
...
 # limitations under the License.
 #
+import mlrun
+
+
 def handler(context, x=1):
     context.log_result("y", x * 2)
+
+
+def func():
```
alonmr: can make the handler name a bit more meaningful
```suggestion
def get_ctx_kind_label():
```

alonmr (commented):
Looks nice! Minor touch up requests

PR #5711, opened 2024-06-04
[Alerts] Verify entity and notifications exist before storing alert

Yacouby:
provide more user-friendly error messages in case the user tries to store an alert without providing entity or at least one  notification.

resolves:
https://iguazio.atlassian.net/browse/ML-6676

### DIFF
--- a/mlrun/alerts/alert.py
+++ b/mlrun/alerts/alert.py
@@ -30,6 +30,11 @@ class AlertConfig(ModelObj):
         "reset_policy",
         "state",
     ]
+    _fields_to_serialize = ModelObj._fields_to_serialize + [
+        "entities",
+        "notifications",
+        "trigger",
+    ]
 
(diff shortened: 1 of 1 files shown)


yanburman (approved):

mlrun/alerts/alert.py
```diff
@@ -74,11 +74,17 @@ def validate_required_fields(self):
     def to_dict(self, fields: list = None, exclude: list = None, strip: bool = False):
         data = super().to_dict(self._dict_fields)
 
+        if self.entities is None:
```
alonmr, line 110: I don't see an issue with the error but the way you use `to_dict` to parse the entities and notifications is not really the mlrun way. you should use `_serialize_field` see `ModelObj._fields_to_serialize`

alonmr (approved):

PR #5714, opened 2024-06-04
[Docs] Add links for both api gateways schema

rokatyy:
Add `mlrun.api_gateways` page to docs and also api gateway schema page to `mlrun.common.schemas`

### DIFF
--- a/docs/api/mlrun.api_gateways.rst
+++ b/docs/api/mlrun.api_gateways.rst
@@ -1,7 +1,0 @@
-mlrun.api_gateway
-==============
-
-.. automodule:: mlrun.runtimes.nuclio.api_gateway
-   :members:
-   :show-inheritance:
-   :undoc-members:
--- a/docs/api/mlrun.api_gateways.rst
+++ b/docs/api/mlrun.api_gateways.rst
@@ -1,0 +1,7 @@
+mlrun.api_gateway
+==============
+
+.. automodule:: mlrun.runtimes.nuclio.api_gateway
+   :members:
+   :show-inheritance:
+   :undoc-members:
--- a/docs/api/mlrun.runtimes/index.rst
+++ b/docs/api/mlrun.runtimes/index.rst
@@ -1,0 +1,10 @@
+.. _mlrun.runtimes:
+
+mlrun.runtimes
+================
+
+.. toctree::
+  :maxdepth: 1
+  :glob:
+
+  mlrun.runtimes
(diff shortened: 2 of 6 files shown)


docs/api/mlrun.api_gateways.rst
alonmr, line 1: shouldn't it be `mlrun.runtimes/mlrun.runtimes.api_gateways.rst` ?

rokatyy: @alonmr why would it? the docs structure has nothing to do with code structure, api_gateway is not a sub-entity of runtime, so I'd add it as a separate entity in MLRun docs.

moranbental (changes requested):

docs/api/mlrun.api_gateways.rst
moranbental: I think you should create new folder of `docs/api/mlrun.runtimes` and move the `docs/api/mlrun.runtimes.rst` inside this folder and rename it to `index.rst` and then create new folder `docs\api\mlrun.runtimes\nuclio` with `index.rst` inside that will calls to `docs\api\mlrun.runtimes\nuclio\api_gatway.rst`
so then in the future we can auto module all the docstrings

rokatyy: @moranbental could you please explain how it will change the behaviour of docs generation? APIGateway-related code is in api_gateway.py file, so it will be always generated properly and be updated. Also, as I said above api_gateway is not a runtime, so it would not be correct from UX perspective to place it under the mlrun.runtimes.

moranbental: If it is not related to runtime, why is `api_gateway.py` in the runtime module in the first place?
When you import APIGateway you do this - `import mlrun.runtimes.nuclio.api_gateway.APIGateway`. So, I believe it should be in the same directory, but under /docs, because each is a module.  Then it will be easier to auto-modulate other classes/functions in the other module.

moranbental (approved):

docs/api/mlrun.runtimes/index.rst
TomerShor, line 1: Is this still needed following https://github.com/mlrun/mlrun/pull/5725?

docs/api/mlrun.runtimes/mlrun.runtimes.rst
TomerShor, line 1: same question?

docs/api/mlrun.runtimes/index.rst
rokatyy: it will be reverted https://github.com/mlrun/mlrun/pull/5730

TomerShor (approved):

PR #5715, opened 2024-06-04
[API] Improve Formatter Filter Utility + Docs

quaark:
- Improve the formatter filter utility to accept dot-delimited key strings instead of lists of keys.
- Add method and class docs


mlrun/common/formatters/base.py
```diff
@@ -54,10 +73,19 @@ def format_obj(
         return format_method(obj)
 
     @staticmethod
-    def filter_obj_method(_filter: list[list[str]]) -> typing.Callable:
+    def filter_obj_method(_filter: list[str]) -> typing.Callable:
+        """
+        Returns a method that filters the object based on the provided list of keys.
+        The keys should be in a dot-separated format, denoting the path within the dictionary to the desired key.
```
TomerShor, line 79: Maybe add:
```
Example:
    [
        "kind",
        "metadata.name",
        "spec.something",
    ]
```

TomerShor (approved):

PR #5717, opened 2024-06-04
[Artifacts] Parse producer URI when retaining producer

TomerShor:
When an artifact is logged in a run, its producer doesn't necessarily contain a `project` and `tag` properties, but it does contain a `uri` (as in the context object).

When importing artifacts that were created in a run, we retain the run as the artifact producer.
However, the URI was ignored and we only looked for a `project` and a `tag`, causing the retained producer to not have a UID.

In this PR we first look for the project and tag on the producer itself, and if not found we try to parse its URI.

Resolves https://iguazio.atlassian.net/browse/ML-4134


mlrun/artifacts/manager.py
```diff
@@ -76,6 +77,23 @@ def get_meta(self) -> dict:
     def uid(self):
         return None
 
+    @staticmethod
+    def parse_uri(uri):
+        """parse artifact producer's uri to project, uid, iteration"""
+        uri_pattern = r"^((?P<project>.*)/)?(?P<uid>.*?)(\-(?P<iteration>.*?))?$"
```
liranbg: move to the piled-on `regex.py`

mlrun/artifacts/manager.py
```diff
@@ -76,6 +77,23 @@ def get_meta(self) -> dict:
     def uid(self):
         return None
 
+    @staticmethod
+    def parse_uri(uri):
```
liranbg:
```suggestion
    def parse_uri(uri) -> typing.Tuple[str, str, str]:
```

liranbg (commented):
minor syntax suggestions

mlrun/artifacts/manager.py
```diff
@@ -76,6 +77,23 @@ def get_meta(self) -> dict:
...
+        """parse artifact producer's uri to project, uid, iteration"""
+        uri_pattern = r"^((?P<project>.*)/)?(?P<uid>.*?)(\-(?P<iteration>.*?))?$"
+        match = re.match(uri_pattern, uri)
+        if not match:
+            raise ValueError(
+                "Uri not in supported format [<project>/]<key>[#<iteration>][:<tag>][@<tree>]"
```
alonmr: Update the error message

alonmr (commented):
~please also change in `server.api.crud.runs.Runs.get_run` to use the new utility~

alonmr (approved):

PR #5721, opened 2024-06-05
[Docs] 1.6.3 change log and other updates [1.6.x]

jillnogold:


liranbg (approved):

docs/store/log-artifacts.md
```diff
@@ -0,0 +1,85 @@
...
+
+**Log with body**
+```python
+    project.log_artifact(
+        "some-data",
+        body=b"abc is 123",
+        format="txt",
+        labels={"Test": "label-test"},
+    )
+```
```
guy1992l: I think @theSaarco said once it is really not recommended to use `body` in `log_artifact`.

theSaarco: That is correct, though for demo purposes it's ok to use a small body. The main issue was with `inline` artifacts, where the body was stored in the artifact metadata in the DB, causing size and scale issues. Nowadays we almost always upload the body as a file, unless the user specifically asks to make the artifact `inline`. 
Still, if we maintain this demo I would warn that it's just a demo and when using a fairly large body it's recommended to not use the `body` property but instead use a file.

jillnogold: added comment

docs/change-log/index.md
```diff
@@ -13,6 +13,52 @@
...
+|NA     | New page describing {ref}`log-artifacts`.                                 |
+|NA     | New page describing {ref}`parallel-workflows`.                                 |
+|NA     | New page describing {ref}`conditional-workflow`.                                 |
+|NA     | New page describing {ref}`exithandler-workflow`.                                 |
+
+### Breaking change
+| ID          |Description                                                               |
+|-------|---------------------------------------------------------------------------|
+|ML-6098|The `prediction` and `named_predictions` columns (list of all predictions) were removed from the model monitoring parquet files. Each prediction is still available in a column of its own.|
+ML-6397|`list runs`  API does not return artifacts in the run body. Instead you get a dictionary of artifact_uris with path to the artifacts in the DB. You can get the artifacts only if you use the `get run`API (i.e. when getting a single run). If an artifact is deleted, it will not be returned in the run body.|
```
alonmr:
```suggestion
ML-6397|`list runs`  API does not return artifacts in the run body. Instead you get a dictionary of artifact_uris with store path to the artifacts which can be retrieved via `mlrun.store_manager`. You can get the artifacts in the run status, only if you use the `get run`API (i.e. when getting a single run). If an artifact is deleted, it will not be returned in the run body.|
```

alonmr (approved):

docs/change-log/index.md
```diff
@@ -1,7 +1,7 @@
 (change-log)=
 # Change log
 
-- [v1.6.3](#v1-6-3-4-june-2024)  | [v1.6.2](#v1-6-2-29-march-2024) | [v1.6.1](#v1-6-1-29-february-2024) | [v1.6.0](#v1-6-0-22-february-2024)
+-  [v1.6.4](#v1-6-4-15-june-2024) | [v1.6.2](#v1-6-2-29-march-2024) | [v1.6.1](#v1-6-1-29-february-2024) | [v1.6.0](#v1-6-0-22-february-2024)
```
alonmr: you are missing 1.6.3 here

PR #5731, opened 2024-06-06
[KFP] Enforce kfp is imported from pipeline adapters only

liranbg:
Blocking mlrun package from directly importing `kfp`
This PR is the last step before introducing "nop kfp" kinda adapter which basically makes mlrun package to not relay on any kfp


mlrun/projects/pipelines.py
```diff
@@ -220,9 +219,6 @@ def is_run_local(self, local=None):
         force_run_local = mlrun.mlconf.force_run_local
         if force_run_local is None or force_run_local == "auto":
             force_run_local = not mlrun.mlconf.is_api_running_on_k8s()
-            kfp_url = mlrun.mlconf.resolve_kfp_url()
-            if not kfp_url:
-                force_run_local = True
```
quaark: This check is still needed and important, just use `mlrun.mlconf.kfp_url` instead of `mlrun.mlconf.resolve_kfp_url`.
For example running CE on k8s in docker desktop on Apple Silicon forces the state of k8s but no kfp.

quaark (changes requested):
Looks good, one thing shouldn't be removed though

quaark (approved):
LGTM

PR #5733, opened 2024-06-06
[CLI] Update workflow help instruction

liranbg:
Avoid confusion when users use the `--workflow` to trigger a specific workflow

https://iguazio.atlassian.net/browse/ML-5420

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -204,7 +204,7 @@ Examples:
 | -s, --secrets TEXT           | Secrets, either as `file=<filename>` or `env=<ENVAR>,...`; for example, `-s file=secrets.txt`           |
 | \--uid TEXT                  | Unique run ID                                                                                           |
 | \--name TEXT                 | Run name                                                                                                |
-| \--workflow TEXT             | Workflow name/id                                                                                        |
+| \--workflow TEXT             | Sets run metadata pipelines related labels to match given workflow name/id                              |
 | \--project TEXT              | Project name or ID                                                                                      |
 | \--db TEXT                   | Save run results to DB url                                                                              |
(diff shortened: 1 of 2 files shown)


mlrun/__main__.py
```diff
@@ -102,7 +102,9 @@ def main():
 )
 @click.option("--uid", help="unique run ID")
 @click.option("--name", help="run name")
-@click.option("--workflow", help="workflow name/id")
+@click.option(
+    "--workflow", help="sets the run pipeline labels to match given workflow name/id"
```
jillnogold: "--workflow", help="sets the pipeline labels to match the given workflow name/id"

docs/cli.md
```diff
@@ -204,7 +204,7 @@ Examples:
 | -s, --secrets TEXT           | Secrets, either as `file=<filename>` or `env=<ENVAR>,...`; for example, `-s file=secrets.txt`           |
 | \--uid TEXT                  | Unique run ID                                                                                           |
 | \--name TEXT                 | Run name                                                                                                |
-| \--workflow TEXT             | Workflow name/id                                                                                        |
+| \--workflow TEXT             | Sets run metadata pipelines related labels to match given workflow name/id                              |
```
jillnogold: Sets the pipeline labels to match the given workflow name/id

mlrun/__main__.py
jillnogold: I think the word "run" is superfluous, yes?

liranbg: not really as what it actually do is spinning a run that its label matches the given value of `--workflow`

jillnogold (approved):

PR #5735, opened 2024-06-06
[Automation] Fix patch remote script docker registry

TomerShor:
Following https://github.com/mlrun/mlrun/pull/5690, the `MLRUN_DOCKER_REGISTRY` env var defaults to `registry.hub.docker.com`.
The patch script only sets the `MLRUN_DOCKER_REPO` env var, causing weird behavior such as resolving the image name to:
```
registry.hub.docker.com/<datanode-ip>:8009/mlrun/mlrun-api:1.7.0-rc21
```

This PR fixes this by:
1. Populating `MLRUN_DOCKER_REGISTRY` with the `DOCKER_REGISTRY` config value, ensuring it ends with a `/` .
2. Setting `MLRUN_DOCKER_REPO=mlrun` or to an empty string, conditioned on the docker registry. This is done because the makefile concatenates "mlrun" to `MLRUN_DOCKER_REGISTRY` value by default.

### DIFF
--- a/automation/patch_igz/patch_env_template.yml
+++ b/automation/patch_igz/patch_env_template.yml
@@ -36,7 +36,7 @@ DOCKER_REPO:
 
 # Optional - Use when your registry is connected to a DNS which is accessible from the cluster
 # For iguazio datanode registry use:
-# datanode-registry.iguazio-platform.app.<system-name>.<lab.iguazeng/cloud-cd>.com:80/mlrun
+# datanode-registry.iguazio-platform.app.<system-name>.<lab.iguazeng/cloud-cd>.com:80
 OVERWRITE_IMAGE_REGISTRY:
 
 # Optional - will attempt docker login if defined, need to login manually otherwise
--- a/automation/patch_igz/patch_env_template.yml
+++ b/automation/patch_igz/patch_env_template.yml
@@ -27,8 +27,12 @@ SSH_PASSWORD:
 
 # Docker registry you can push to. Make sure you are logged into it before running the script (required)
 # e.g. docker.io/my_user
-# For iguazio datanode registry use <data-node-ip>:8009/mlrun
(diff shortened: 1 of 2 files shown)


alonmr (commented):
I think `DOCKER_REPO` should be in the env yml and the user should be responsible on setting the registry and repo correctly.
Here is how I did it ;}
```
        mlrun_docker_repo = self._config["DOCKER_REPO"]
        mlrun_docker_registry = self._config["DOCKER_REGISTRY"]
        if not mlrun_docker_registry.endswith("/"):
            mlrun_docker_registry += "/"
        env = {
            "MLRUN_VERSION": image_tag,
            "MLRUN_DOCKER_REGISTRY": mlrun_docker_registry,
            "MLRUN_DOCKER_REPO": mlrun_docker_repo,
        }
        cmd = ["make", target]
        self._exec_local(cmd, live=True, env=env)
        return f"{mlrun_docker_registry}{mlrun_docker_repo}/{image_name}:{image_tag}"

```

automation/patch_igz/patch_env_template.yml
```diff
@@ -27,9 +27,13 @@ SSH_PASSWORD:
...
+# For iguazio datanode registry use <data-node-ip>:8009
 DOCKER_REGISTRY:
 
+# Docker repository to push to
+# If you are using the iguazio datanode registry, use "mlrun"
+DOCKER_REPO:
+
 # Optional - Use when your registry is connected to a DNS which is accessible from the cluster
 # For iguazio datanode registry use:
 # datanode-registry.iguazio-platform.app.<system-name>.<lab.iguazeng/cloud-cd>.com:80/mlrun
```
alonmr:
```suggestion
# datanode-registry.iguazio-platform.app.<system-name>.<lab.iguazeng/cloud-cd>.com:80
```

PR #5737, opened 2024-06-06
[Model Monitoring] Support endpoint connection string based on project secret

Eyal-Danieli:
Until now, if a user wished to define a SQL target for storing the model monitoring data instead of V3IO KV, he had to apply configmap to update MLRun configurations, particularly the `model_endpoint_monitoring.store_type` value. Afterwards he had to provide a SQL connection string either through configmap or by using the  `set_model_monitoring_credentials` API.

In this PR, we improve this behaviour so that users can provide the connection string using the   `set_model_monitoring_credentials` method, without applying any configmap to update MLRun model monitoring configurations. It's important to note that this API is project=specific, meaning that the user will have to apply this API to each project. For example:

`project.set_model_monitoring_credentials(endpoint_store_connection="mysql+pymysql://root@my-db:3306/mlrun_model_monitoring")`
JIRA: https://iguazio.atlassian.net/browse/ML-6747

### DIFF
--- a/mlrun/model_monitoring/db/stores/__init__.py
+++ b/mlrun/model_monitoring/db/stores/__init__.py
@@ -31,17 +31,12 @@ class ObjectStoreFactory(enum.Enum):
     def to_object_store(
         self,
         project: str,
-        access_key: str = None,
-        secret_provider: typing.Callable = None,
+        **kwargs,
     ) -> StoreBase:
         """
         Return a StoreBase object based on the provided enum value.
 
         :param project:                   The name of the project.
-        :param access_key:                Access key with permission to the DB table. Note that if access key is None
-                                          and the endpoint target is from type KV then the access key will be
-                                          retrieved from the environment variable.
-        :param secret_provider:           An optional secret provider to get the connection string secret.
 
         :return: `StoreBase` object.
 
@@ -50,10 +45,7 @@ class ObjectStoreFactory(enum.Enum):
         if self == self.v3io_nosql:
             from mlrun.model_monitoring.db.stores.v3io_kv.kv_store import KVStoreBase
 
-            # Get V3IO access key from env
-            access_key = access_key or mlrun.mlconf.get_v3io_access_key()
-
-            return KVStoreBase(project=project, access_key=access_key)
+            return KVStoreBase(project=project)
 
         # Assuming SQL store target if store type is not KV.
         # Update these lines once there are more than two store target types.
@@ -62,7 +54,7 @@ class ObjectStoreFactory(enum.Enum):
 
         return SQLStoreBase(
             project=project,
-            secret_provider=secret_provider,
+            **kwargs,
         )
 
     @classmethod
@@ -95,24 +87,40 @@ def get_model_endpoint_store(
 
 def get_store_object(
     project: str,
-    access_key: str = None,
+    store_type: typing.Optional[str] = None,
     secret_provider: typing.Callable = None,
+    **kwargs,
 ) -> StoreBase:
     """
     Getting the DB target type based on mlrun.config.model_endpoint_monitoring.store_type.
 
     :param project:         The name of the project.
-    :param access_key:      Access key with permission to the DB table.
+    :param store_type:      The type of the store target. See mlrun.model_monitoring.db.stores.ObjectStoreFactory
+                            for available store types.
     :param secret_provider: An optional secret provider to get the connection string secret.
 
     :return: `StoreBase` object. Using this object, the user can apply different operations on the
              model monitoring record such as write, update, get and delete a model endpoint.
     """
 
+    store_connection_string = mlrun.model_monitoring.helpers.get_connection_string(
+        secret_provider=secret_provider
+    )
+
+    if store_connection_string and (
+        store_connection_string.startswith("mysql")
+        or store_connection_string.startswith("sqlite")
+    ):
+        store_type = mlrun.common.schemas.model_monitoring.ModelEndpointTarget.SQL
+        kwargs["store_connection_string"] = store_connection_string
+
+    # Set the default store type if no connection has been set
+    store_type = store_type or mlrun.mlconf.model_endpoint_monitoring.store_type
+
     # Get store type value from ObjectStoreFactory enum class
-    store_type = ObjectStoreFactory(mlrun.mlconf.model_endpoint_monitoring.store_type)
+    store_type = ObjectStoreFactory(store_type)
 
     # Convert into store target object
     return store_type.to_object_store(
-        project=project, access_key=access_key, secret_provider=secret_provider
+        project=project, secret_provider=secret_provider, **kwargs
     )
--- a/mlrun/model_monitoring/db/stores/__init__.py
+++ b/mlrun/model_monitoring/db/stores/__init__.py
@@ -71,7 +71,7 @@ class ObjectStoreFactory(enum.Enum):
 def get_model_endpoint_store(
     project: str,
     access_key: str = None,
-    secret_provider: typing.Callable = None,
+    secret_provider: typing.Optional[typing.Callable[[str], str]] = None,
 ) -> StoreBase:
     # Leaving here for backwards compatibility
     warnings.warn(
@@ -87,21 +87,18 @@ def get_model_endpoint_store(
 
 def get_store_object(
     project: str,
-    store_type: typing.Optional[str] = None,
-    secret_provider: typing.Callable = None,
+    secret_provider: typing.Optional[typing.Callable[[str], str]] = None,
     **kwargs,
 ) -> StoreBase:
     """
-    Generate a store object based on the provided store type. If a connection string is provided, the store type will
-    be updated according to the connection string.
+    Generate a store object. If a connection string is provided, the store type will be updated according to the
+    connection string. Currently, the supported store types are SQL and v3io-nosql.
 
     :param project:         The name of the project.
-    :param store_type:      The type of the store target. See mlrun.model_monitoring.db.stores.ObjectStoreFactory
-                            for available store types.
     :param secret_provider: An optional secret provider to get the connection string secret.
 
-    :return: `StoreBase` object. Using this object, the user can apply different operations on the
-             model monitoring record such as write, update, get and delete a model endpoint.
+    :return: `StoreBase` object. Using this object, the user can apply different operations such as write, update, get
+    and delete a model endpoint record.
     """
 
     store_connection_string = mlrun.model_monitoring.helpers.get_connection_string(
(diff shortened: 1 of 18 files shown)


mlrun/model_monitoring/db/stores/__init__.py
```diff
@@ -95,24 +87,40 @@ def get_model_endpoint_store(
 
 def get_store_object(
     project: str,
-    access_key: str = None,
+    store_type: typing.Optional[str] = None,
     secret_provider: typing.Callable = None,
+    **kwargs,
 ) -> StoreBase:
     """
     Getting the DB target type based on mlrun.config.model_endpoint_monitoring.store_type.
```
jond01: Please update this

server/api/api/endpoints/model_endpoints.py
```diff
@@ -349,7 +349,10 @@ async def get_model_endpoint_monitoring_metrics(
     )
 
     get_model_endpoint_metrics = mlrun.model_monitoring.get_store_object(
-        project=project
+        project=project,
+        secret_provider=server.api.crud.secrets.get_project_secret_provider(
+            project=project
+        ),
```
jond01: I see this repeats in multiple places, how about a helper function?
E.g., in `server/api/crud/model_monitoring/model_endpoints.py` L143:
```py
        model_endpoint_store = mlrun.model_monitoring.get_store_object(
            project=model_endpoint.metadata.project,
            secret_provider=server.api.crud.secrets.get_project_secret_provider(
                project=model_endpoint.metadata.project
            ),
        )
```

mlrun/model_monitoring/writer.py
```diff
@@ -102,7 +102,7 @@ class ModelMonitoringWriter(StepToDict):
 
     kind = "monitoring_application_stream_pusher"
 
-    def __init__(self, project: str, tsdb_secret_provider=None) -> None:
+    def __init__(self, project: str, secret_provider=None) -> None:
```
jond01: A type hint could be nice

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -89,16 +89,20 @@ class SchemaParams:
...
 
-    def __init__(self, project: str, access_key: typing.Optional[str] = None) -> None:
+    def __init__(
+        self,
+        project: str,
+    ) -> None:
         super().__init__(project=project)
         # Initialize a V3IO client instance
-        self.access_key = access_key or os.environ.get("V3IO_ACCESS_KEY")
+        self.access_key = os.environ.get("V3IO_ACCESS_KEY")
```
jond01: This can probably be safely removed in all places inside this class. This env var is taken by default by the clients.

mlrun/model_monitoring/helpers.py
```diff
@@ -278,9 +278,13 @@ def calculate_inputs_statistics(
     return inputs_statistics
 
 
-def get_endpoint_record(project: str, endpoint_id: str):
+def get_endpoint_record(
+    project: str,
+    endpoint_id: str,
+    secret_provider: typing.Optional[typing.Callable] = None,
```
jond01: Only if you want to be extra with the type hints - what are the input and output types of this callable?

server/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -13,7 +13,6 @@
 # limitations under the License.
```
jond01, line 13: In this module:
Please update the docstrings of `ModelEndpoints.create_model_endpoint` - it still mentions only `mlrun.config.model_endpoint_monitoring.store_type` as the store type selector.

server/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -534,26 +522,24 @@ def delete_model_endpoints_resources(project_name: str):
...
-        auth_info = mlrun.common.schemas.AuthInfo(
-            data_session=os.getenv("V3IO_ACCESS_KEY")
+
+        # Delete model monitoring store resources
+        endpoint_store = mlrun.model_monitoring.get_store_object(
+            project=project_name,
+            secret_provider=server.api.crud.secrets.get_project_secret_provider(
+                project=project_name
+            ),
         )
```
jond01: One more occurrence :)

server/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -617,7 +603,6 @@ def _adjust_stats(
 
     @staticmethod
     def _add_real_time_metrics(
-        model_endpoint_store: mlrun.model_monitoring.db.StoreBase,
```
jond01, line 620: Nice catch! This argument was not used.

mlrun/model_monitoring/db/stores/base/store.py
```diff
@@ -19,6 +19,7 @@
 
 
 class StoreBase(ABC):
+    type: str = ""
```
jond01: I think that's more accurate + you don't need a default value:
```suggestion
    type: typing.ClassVar[str]
```

mlrun/model_monitoring/stream_processing.py
```diff
@@ -310,7 +306,10 @@ def apply_infer_schema():
                 table=self.kv_path,
             )
 
-        if self.model_endpoint_store_target == ModelEndpointTarget.V3IO_NOSQL:
+        store_object = mlrun.model_monitoring.get_store_object(
+            project=self.project, secret_provider=secret_provider
+        )
+        if store_object.type == ModelEndpointTarget.V3IO_NOSQL:
             apply_infer_schema()
```
jond01, line 313: Unrelated: do you know why we limit this to V3IO only?

mlrun/model_monitoring/db/stores/__init__.py
```diff
@@ -95,24 +87,40 @@ def get_model_endpoint_store(
...
+
     # Get store type value from ObjectStoreFactory enum class
-    store_type = ObjectStoreFactory(mlrun.mlconf.model_endpoint_monitoring.store_type)
+    store_type = ObjectStoreFactory(store_type)
 
     # Convert into store target object
     return store_type.to_object_store(
-        project=project, access_key=access_key, secret_provider=secret_provider
+        project=project, secret_provider=secret_provider, **kwargs
     )
```
jond01: If I follow correctly, it fails if there's a connection string in the secrets, and we try to get a V3IO-based store.

mlrun/model_monitoring/db/stores/__init__.py
```diff
@@ -95,24 +87,40 @@ def get_model_endpoint_store(
...
     """
 
+    store_connection_string = mlrun.model_monitoring.helpers.get_connection_string(
+        secret_provider=secret_provider
+    )
+
+    if store_connection_string and (
+        store_connection_string.startswith("mysql")
+        or store_connection_string.startswith("sqlite")
+    ):
```
jond01, line 111: To resolve the previous comment do this only if the `store_type` is `None` or SQL.

tests/system/model_monitoring/test_model_monitoring.py
```diff
@@ -246,16 +248,23 @@ class TestBasicModelMonitoring(TestMLRunSystem):
     # Set image to "<repo>/mlrun:<tag>" for local testing
     image: Optional[str] = None
 
-    @pytest.mark.timeout(270)
-    def test_basic_model_monitoring(self) -> None:
+    @pytest.mark.timeout(540)
+    @pytest.mark.parametrize("with_sql_target", [True, False])
+    def test_basic_model_monitoring(self, with_sql_target: bool) -> None:
```
jond01, line 253: Does it pass now? I remember it failed. If you fixed it here - awesome!
Was it only the timeout?

jond01 (approved):
Great job! 🐬

mlrun/model_monitoring/db/stores/base/store.py
Eyal-Danieli: nice, I've updated `TSDBConnector` as well

mlrun/model_monitoring/stream_processing.py
Eyal-Danieli: yes, because we have to generate V3IO KV Schema for Grafana. In general, these steps should be wrapped within the KV/SQL abstraction (similar to what we did with the TSDB connector) but anyway for now we use that approach.

mlrun/model_monitoring/db/stores/__init__.py
Eyal-Danieli: but if you have connection string, we update the `store_type` accordingly

mlrun/model_monitoring/db/stores/__init__.py
Eyal-Danieli: but that's the thing - the store type should be decided based on the connection string

tests/system/model_monitoring/test_model_monitoring.py
Eyal-Danieli: it passes :) didn't change anything in the logic. I raised the timeout 2x because we are running it "twice", actually I wasn't sure if pytest timeout resets on each pytest param running.

mlrun/model_monitoring/stream_processing.py
jond01: I see, thx

tests/system/model_monitoring/test_model_monitoring.py
jond01: The timeout applies to each test item (params set) separately.
But since it solves the issue, let's keep it as is.

mlrun/model_monitoring/db/stores/__init__.py
jond01: I see your point.
Generally, we would like to minimize surprises in the code.
If one called
```py
v3io_store = get_store_object(..., store_type="v3io-nosql", secret_provider=...)
```
They would be surprised to get a MySQL store instead of a V3IO KV store - and that situation might happen here.

So, I think the `store_type` argument should determine the returned type if explicitly provided.

mlrun/model_monitoring/db/stores/__init__.py
```diff
@@ -95,24 +87,41 @@ def get_model_endpoint_store(
 
 def get_store_object(
     project: str,
-    access_key: str = None,
+    store_type: typing.Optional[str] = None,
     secret_provider: typing.Callable = None,
```
jond01:
```suggestion
    secret_provider: typing.Optional[typing.Callable] = None,
```

mlrun/model_monitoring/db/stores/__init__.py
Eyal-Danieli: Actually we don't really need this param here. I mean we shouldn't pass it to `get_store_object` because anyway it should be decided based on the connection string (otherwise take it from the default config). Therefore I removed it. 
https://github.com/mlrun/mlrun/pull/5737/commits/922ee45ce586395d45ca97c58d8daadbfb667d6a#diff-7cf9046981c4e232dd7484cd0651eb603ecc22a53310b05e4d737bb190399df6:~:text=%3A%20str%2C-,store_type,-%3A%20typing.

jond01 (approved):

PR #5738, opened 2024-06-06
[Iguazio] Modify the message that appears when a project cannot be deleted

moranbental:
When Zebo returns an exception, the error message now shows `Project deletion job failed. Failed to delete a project in mlrun`
This error appears in the UI when MLRun generates it, so the error `Fail to delete project in mlrun` is sufficient

Before the change:
image

After the change:
image

https://iguazio.atlassian.net/browse/ML-6625


alonmr (commented):
Generally looks good though I would love to see some pre and post fix screenshots

TomerShor:
@moranbental it's a good opportunity to change `"Fail to delete project in MLRun"` to `"Failed to delete project in MLRun"`

moranbental:
> @moranbental it's a good opportunity to change `"Fail to delete project in MLRun"` to `"Failed to delete project in MLRun"`

@TomerShor This message comes from zebo

server/api/utils/clients/iguazio.py
```diff
@@ -956,6 +961,17 @@ def _is_job_terminated(self, session: str, job_id: str) -> bool:
...
+    @staticmethod
+    def _resolve_final_error_message(error_message, result_message):
+        if error_message and result_message:  # If both are not empty
+            return f"{error_message}: {result_message}"
+        if not result_message:  # If result_message is empty
+            return error_message
+        if not error_message:  # If error_message is empty
+            return result_message
+        # If both are empty
+        return ""
```
alonmr: More concise, 
```suggestion
        result_message = result_message or ""
        error_message = error_message or ""
        return result_message or error_message
```

server/api/utils/clients/iguazio.py
```diff
@@ -956,6 +961,17 @@ def _is_job_terminated(self, session: str, job_id: str) -> bool:
                 )
             raise
 
+    @staticmethod
+    def _resolve_final_error_message(error_message, result_message):
```
alonmr: add type hints and default values

server/api/utils/clients/iguazio.py
moranbental: This is not right because if `result_message` is empty and `error_message` is **not** empty it will return empty message instead of the `error_message` value

alonmr (approved):

liranbg (approved):

PR #5740, opened 2024-06-07
[Nuclio] List only functions with status when deleting Nuclio functions

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-6744

The problem with duplicated error happened because `list_functions()` returns 2 copies of the same functions. One of tag `latest` and another one of an empty tag. The one with empty tag is the same with latest, but without status, so them both point to the same Nuclio function. In order to avoid duplicated deletion of nuclio functions, we need to filter out those without status.

### DIFF
--- a/server/api/api/utils.py
+++ b/server/api/api/utils.py
@@ -1303,12 +1303,17 @@ async def _delete_function(
     # in MLRun terminology, they are all just versions of the same function
     # therefore, it's enough to check the kind of the first one only
     if functions[0].get("kind") in mlrun.runtimes.RuntimeKinds.nuclio_runtimes():
+
+        # filter functions which doesn't have status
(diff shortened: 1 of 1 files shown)


server/api/api/utils.py
```diff
@@ -1303,12 +1303,16 @@ async def _delete_function(
     # in MLRun terminology, they are all just versions of the same function
     # therefore, it's enough to check the kind of the first one only
     if functions[0].get("kind") in mlrun.runtimes.RuntimeKinds.nuclio_runtimes():
+        # filter functions which doesn't have status
```
```suggestion
        # filter functions which don't have a status
```

TomerShor (approved):
Minor typo - otherwise approved 👍

PR #5744, opened 2024-06-10
[Runs] Get artifacts from `run.status.artifact_uris` for workflow runs [1.6.x]

alonmr:
We saw that for workflow runs where the workflow has many runs, listing the artifacts to enrich the run status takes too long. Instead we will get the runs from the artifact uris that are stored in the run status.
In case one of the artifacts is missing, it will be ignored.
We also add DB index for artifacts_v2 (project, producer_id, best_iteration) because this filter is frequently used and the index will also be used by the DB when querying for specific iteration.

https://iguazio.atlassian.net/browse/ML-6770


server/api/crud/runs.py
```diff
@@ -150,29 +150,14 @@ def get_run(
...
-        artifacts = server.api.crud.Artifacts().list_artifacts(
-            db_session,
-            iter=iter,
-            best_iteration=best_iteration,
-            producer_id=producer_id,
-            producer_uri=producer_uri,
-            project=project,
-        )
+        else:
+            artifacts = self._get_artifacts_from_uris(db_session, project, run)
```
TomerShor: Add a comment why we need a different flow

server/api/crud/runs.py
```diff
@@ -468,3 +453,48 @@ async def _post_delete_run(project, uid):
                 project,
                 uid,
             )
+
+    @staticmethod
+    def _get_artifacts_from_uris(db_session, project, run):
+        artifacts = []
+        artifact_uris = run.get("status", {}).get("artifact_uris", {})
+        for key, uri in artifact_uris.items():
```
TomerShor: Not needed since its parsed from the URI
```suggestion
        for _, uri in artifact_uris.items():
```

server/api/crud/runs.py
```diff
@@ -468,3 +453,48 @@ async def _post_delete_run(project, uid):
                 project,
                 uid,
             )
+
+    @staticmethod
+    def _get_artifacts_from_uris(db_session, project, run):
```
TomerShor: docstring + typehints

server/api/crud/runs.py
```diff
@@ -468,3 +453,48 @@ async def _post_delete_run(project, uid):
...
+                    iteration=iteration,
+                )
+        return artifacts
+
+    @staticmethod
+    def _list_run_artifacts(db_session, iteration: int, producer_id: str, project: str):
+        best_iteration = False
+        if not iteration:
+            iteration = None
+            best_iteration = True
```
TomerShor: explain that `iter == 0` means best iteration

TomerShor (changes requested):
Minor comments

server/api/db/sqldb/models/models_sqlite.py
```diff
@@ -206,10 +206,14 @@ class ArtifactV2(Base, mlrun.utils.db.BaseModel):
 
         id = Column(Integer, primary_key=True)
         key = Column(String(255, collation=SQLCollationUtil.collation()), index=True)
-        project = Column(String(255, collation=SQLCollationUtil.collation()))
+        project = Column(
```
liranbg: You need to index the combination of columns and not separately

server/api/crud/runs.py
roei3000b: I'm not sure, but should we use TTLcache for this function?
IIUC, in case we have lots of artifacts and there are multiple users that try to get them, we will cause heavy load on the db.

alonmr: Interesting.. though the artifacts are per run so i don't think caching will help here since we don't expect many requests for the same run.

server/api/crud/runs.py
```diff
@@ -150,29 +150,16 @@ def get_run(
...
-            db_session,
-            iter=iter,
-            best_iteration=best_iteration,
-            producer_id=producer_id,
-            producer_uri=producer_uri,
-            project=project,
-        )
+        else:
+            # For workflow runs, we fetch the artifacts one by one since listing them may be too heavy with the
+            # workflow id as the producer_id
```
TomerShor:
```suggestion
            # For workflow runs, we fetch the artifacts one by one since listing them with the workflow_id as the producer_id
            # may return too many artifacts
```

server/api/crud/runs.py
```diff
@@ -468,3 +455,59 @@ async def _post_delete_run(project, uid):
...
+    def _list_run_artifacts(
+        db_session: sqlalchemy.orm.Session,
+        iteration: int,
+        producer_id: str,
+        project: str,
+    ):
+        best_iteration = False
+        # If the iteration is 0, we mark the artifacts as best iteration.
+        # Specifically for hyper runs, iteration 0 is the parent run, so it needs the get the artifacts
+        # of the real best iteration which has some other iteration number.
```
TomerShor: Changed it a bit but it is still not super-understood.
```suggestion
        # Specifically for hyper runs, iteration 0 is the parent run, so we need to get the artifacts
        # of the real best iteration run, which can be from a different iteration.
```

server/api/crud/runs.py
alonmr: too many artifacts is not accurate IMO so I elaborated instead

server/api/crud/runs.py
```diff
@@ -498,8 +499,8 @@ def _list_run_artifacts(
     ):
         best_iteration = False
         # If the iteration is 0, we mark the artifacts as best iteration.
-        # Specifically for hyper runs, iteration 0 is the parent run, so it needs the get the artifacts
-        # of the real best iteration which has some other iteration number.
+        # Specifically for hyper runs, iteration 0 is the parent run, so we needs the get the artifacts
```
TomerShor:
```suggestion
        # Specifically for hyper runs, iteration 0 is the parent run, so we need to get the artifacts
```

server/api/crud/runs.py
```diff
@@ -157,8 +157,9 @@ def get_run(
             )
 
         else:
-            # For workflow runs, we fetch the artifacts one by one since listing them may be too heavy with the
-            # workflow id as the producer_id
+            # For workflow runs, we fetch the artifacts one by one since listing them with the workflow_id as
+            # the producer_id may be too heavy with the as it fetches all the artifacts of the workflow and then
```
TomerShor:
```suggestion
            # the producer_id may be too heavy as it fetches all the artifacts of the workflow and then
```

TomerShor (changes requested):
last typos

theSaarco (approved):
Looks good. Are you still checking if the index creation can be "create or replace"?

server/api/crud/runs.py
```diff
@@ -468,3 +456,59 @@ async def _post_delete_run(project, uid):
...
+            )
+            artifact = server.api.crud.Artifacts().get_artifact(
+                db_session,
+                key=key,
+                tag=tag,
+                iter=iteration,
+                project=project,
+                producer_id=producer_id,
+                raise_on_not_found=False,
+            )
```
liranbg: so you have the key, tag, iteration, producer_id, etc. why not performing one API request to get all artifacts matching the query? performing single DB-call vs N-calls of GET is much better

server/api/db/sqldb/db.py
```diff
@@ -4721,6 +4724,48 @@ def delete_table_records_by_name(
...
+        sanitized_table_name = re.sub(r"[^a-zA-Z0-9_]", "", table_name)
+
+        # checking if the index exists can also help prevent SQL injection
+        if self._is_index_exists(
+            index_name=sanitized_index_name, table_name=sanitized_table_name
+        ):
+            drop_index_statement = text(
+                f"DROP INDEX {sanitized_index_name} ON {sanitized_table_name}"
+            )
+            session.execute(drop_index_statement)
```
liranbg:
```suggestion
                from sqlalchemy import text, ...
                query = text("DROP INDEX :index_name ON :table_name")
                params = {
                    "index_name": "...",
                    "table_name": "...",
                }
                session.execute(query, params)
```
and now you dont need to sanitize anything as execute will perform sql-injection validation

liranbg (changes requested):

server/api/crud/runs.py
alonmr: the producer id is the workflow id, I don't see how can I build a list query that ensures I get the correct artifacts

server/api/db/sqldb/db.py
alonmr: Turns out this is not possible because query params are used for values and not identifiers/names (they will get quoted)

server/api/crud/runs.py
```diff
@@ -468,3 +458,60 @@ async def _post_delete_run(project, uid):
...
+    @staticmethod
+    def _get_artifacts_from_uris(
+        db_session: sqlalchemy.orm.Session, project: str, producer_id: str, run: dict
+    ):
+        """Fetch run artifacts by their artifact URIs in the run status"""
+        artifact_uris = run.get("status", {}).get("artifact_uris", {})
+        key_tag_iteration_pairs = []
+        for _, uri in artifact_uris.items():
+            _, uri = mlrun.datastore.parse_store_uri(uri)
+            project, key, iteration, tag, _ = mlrun.utils.parse_artifact_uri(
```
theSaarco: I think it makes sense here to verify that the last return value (which is the `tree`), if returned, is the same as the `producer_id` that you got as parameter - just as sanity check.

server/api/db/base.py
```diff
@@ -213,6 +213,16 @@ def list_artifacts(
     ):
         pass
 
+    @abstractmethod
+    def list_grouped_artifacts(
```
theSaarco: I'm not a big fan of the name, since the grouping is specific for artifacts with the same producer id. So, why not call it something like `list_artifacts_for_producer_id`. I'm also not sure why you allow passing `None` as `producer_id` - what's the use-case? I would think that `producer_id` needs to be a mandatory parameter here.

theSaarco (changes requested):
Some comments inside. 
TBH, I'm not sure we want to automatically remove indexes that were added manually - I think we should have this done by the support team whenever an index was manually added, so they manually remove it. It seems to me that this adds a lot of complexity to the code that should be handled by whoever added the indexes in the first place.

server/api/db/sqldb/db.py
liranbg: you can use `execute("DROP INDEX %s on %s", (index_name, table_name))` for selection iirc

server/api/initial_data.py
```diff
@@ -179,6 +187,42 @@ def _create_alembic_util() -> server.api.utils.db.alembic.AlembicUtil:
...
+        index_name="ix_artifacts_v2_project_producer_best_iter",
+        table_name=server.api.db.sqldb.models.ArtifactV2.__tablename__,
+        raise_on_not_exists=False,
+    )
+    db.delete_index_by_name(
+        session=db_session,
+        index_name="idx_artifacts_producer_id_best_iteration_and_project",
+        table_name=server.api.db.sqldb.models.ArtifactV2.__tablename__,
+        raise_on_not_exists=False,
+    )
```
liranbg: this can be deleted. any manual index would be deleted manually as well

liranbg (commented):
last 2 comments

server/api/db/sqldb/db.py
alonmr: if it's decided that we delete manually then I'm removing this as well

theSaarco (approved):
LGTM

PR #5745, opened 2024-06-10
[Nuclio] Add api gateway URLs to function's external_invocation_urls list 

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-6363 and also fixes this bug https://iguazio.atlassian.net/issues/ML-6485

This PR updates a function's `external_invocation_urls` whenever an API gateway is created or deleted


server/api/api/endpoints/nuclio.py
```diff
@@ -160,9 +177,24 @@ async def delete_api_gateway(
...
+                        server.api.crud.Functions().delete_function_external_invocation_url,
+                        function_uri=function,
+                        project=project,
+                        invocation_url=api_gateway.spec.host,
+                    )
+                )
+                for function in api_gateway.get_function_names()
+            ]
+            await asyncio.gather(*tasks)
+            return await client.delete_api_gateway(project_name=project, name=gateway)
```
TomerShor: Can this be also in the concurrent tasks?
Or is there a dependency of deleting the AGW only after removing its URL from the functions?

server/api/api/endpoints/nuclio.py
```diff
@@ -504,7 +539,15 @@ def _deploy_nuclio_runtime(
...
-    db_session, auth_info, fn, name, project, tag, last_log_timestamp, verbose
+    db_session,
+    auth_info,
+    fn,
+    name,
+    project,
+    tag,
+    last_log_timestamp,
+    verbose,
+    api_gateway_hosts,
```
TomerShor: Add type hints

server/api/api/endpoints/nuclio.py
```diff
@@ -587,6 +634,25 @@ def _handle_nuclio_deploy_status(
...
+        # extract the hosts from the API gateway specifications and return them as a list
+        return (
+            [
+                api_gateway.spec.host
+                for api_gateway in api_gateways.values()
+                if function_uri in api_gateway.get_function_names()
+            ]
+            if api_gateways
+            else []
+        )
```
TomerShor: Is this filtering possible in the `list` request (selector in Nuclio or something)?
It will be better then filtering in memory.

server/api/api/endpoints/nuclio.py
```diff
@@ -587,6 +634,25 @@ def _handle_nuclio_deploy_status(
     )
 
 
+async def _get_api_gateways_hosts_for_function(
```
TomerShor: I think the meaning is clearer with:
```suggestion
async def _get_api_gateways_urls_for_function(
```

server/api/db/sqldb/db.py
```diff
@@ -1741,6 +1741,66 @@ def update_function(
             self._upsert(session, [function])
             return function.struct
 
+    def update_function_external_invocation_url(
+        self,
+        session,
+        name,
+        external_invocation_url: str,
```
```suggestion
        url: str,
```

server/api/db/sqldb/db.py
```diff
@@ -1741,6 +1741,66 @@ def update_function(
...
+        project="",
+        tag="",
+        hash_key="",
+        operation="add",
+    ):
```
TomerShor: Add type hints

server/api/db/sqldb/db.py
```diff
@@ -1741,6 +1741,66 @@ def update_function(
...
 
+    def update_function_external_invocation_url(
+        self,
+        session,
+        name,
+        external_invocation_url: str,
+        project="",
+        tag="",
+        hash_key="",
+        operation="add",
```
TomerShor: I would force a set of supported operations (using pydantic).
Also, a docstring for the function will help understand the options.

server/api/db/sqldb/db.py
```diff
@@ -1741,6 +1741,66 @@ def update_function(
...
+            tag=tag or "latest",
+            hash_key=hash_key,
+            project=project,
+        )
+        if uid:
+            query = query.filter(Function.uid == uid)
+        function = query.one_or_none()
+        if not function:
+            logger.debug(
+                "Function is not found",
```
TomerShor: "Not found" can be in too many flows
```suggestion
                "Function is not found, skipping external invocation urls update",
```

TomerShor (changes requested):
Nice! Left some questions and suggestions

server/api/api/endpoints/nuclio.py
rokatyy: @TomerShor I made it like this to avoid a case where we don't have an api gateway anymore, but still have a link in the function to it.

server/api/api/endpoints/nuclio.py
rokatyy: @TomerShor I agree that it would be much better than filtering on mlrun side, but the only possible way to do it on Nuclio will require changes from Nuclio as well meaning either filtering by function name or by annotation contain given string. Currently we support only filtering by labels in Nuclio.

server/api/api/endpoints/functions.py
```diff
@@ -454,6 +455,9 @@ async def build_status(
 
     # nuclio deploy status
     if fn.get("kind") in RuntimeKinds.pure_nuclio_deployed_runtimes():
+        api_gateways_hosts = await _get_api_gateways_urls_for_function(
```
TomerShor:
```suggestion
        api_gateways_urls = await _get_api_gateways_urls_for_function(
```

server/api/api/endpoints/nuclio.py
```diff
@@ -274,7 +306,9 @@ async def deploy_status(
             HTTPStatus.BAD_REQUEST.value,
             reason=f"Runtime kind {fn.kind} is not a nuclio runtime",
         )
-
+    api_gateways_hosts = await _get_api_gateways_urls_for_function(
```
TomerShor:
```suggestion
    api_gateways_urls = await _get_api_gateways_urls_for_function(
```

server/api/api/endpoints/nuclio.py
```diff
@@ -504,7 +539,15 @@ def _deploy_nuclio_runtime(
...
-    db_session, auth_info, fn, name, project, tag, last_log_timestamp, verbose
+    db_session,
+    auth_info,
+    fn,
+    name: str,
+    project: str,
+    tag: str,
+    last_log_timestamp: int,
+    verbose: bool,
+    api_gateway_hosts: list[str],
```
TomerShor:
```suggestion
    api_gateway_urls: list[str],
```

server/api/api/endpoints/nuclio.py
TomerShor: Let's add a TODO then because we would want the optimized way.

server/api/db/sqldb/db.py
```diff
@@ -42,6 +42,7 @@
 import server.api.db.session
 import server.api.utils.helpers
 from mlrun.artifacts.base import fill_artifact_object_hash
+from mlrun.common.types import Operation
```
TomerShor: in general - prefer full imports over `from`: see [this section](https://github.com/mlrun/mlrun/blob/development/CONTRIBUTING.md#:~:text=When%20importing%20from%20local%20code%2C%20NEVER%20use%3A%20from%20X%20import%20Y%20instead%20use%3A%20import%20X%20or%20import%20X%20as%20Y.%20For%20external%20packages%2C%20it%27s%20acceptable%20to%20use%20from%20X%20import%20Y%20since%20they%20won%27t%20try%20to%20import%20back%20to%20our%20code.)
```suggestion
import mlrun.common.types
```

server/api/db/sqldb/db.py
```diff
@@ -1741,6 +1742,69 @@ def update_function(
...
 
+    def update_function_external_invocation_url(
+        self,
+        session,
+        name: str,
+        url: str,
+        project: str = "",
+        tag: str = "",
+        hash_key: str = "",
+        operation=Operation.ADD,
```
TomerShor: add type hint

server/api/crud/functions.py
```diff
@@ -153,3 +154,42 @@ def update_function(
...
+            mlrun.common.helpers.parse_versioned_object_uri(function_uri)
+        )
+        server.api.utils.singletons.db.get_db().update_function_external_invocation_url(
+            session=db_session,
+            name=function_name,
+            url=invocation_url,
+            project=project,
+            tag=tag,
+            hash_key=hash_key,
+        )
```
TomerShor: I know it's the default, but explicitness is worth it here to be resilient to changes
```suggestion
            hash_key=hash_key,
            operation=Operation.ADD,
        )
```

server/api/db/base.py
```diff
@@ -333,6 +334,19 @@ def update_function(
...
+    @abstractmethod
+    def update_function_external_invocation_url(
+        self,
+        session,
+        name: str,
+        url: str,
+        project: str = "",
+        tag: str = "",
+        hash_key: str = "",
+        operation=Operation.ADD,
```
TomerShor: type hint

server/api/db/sqldb/db.py
```diff
@@ -1741,6 +1742,69 @@ def update_function(
...
+        )
+        uid = self._get_function_uid(
+            session=session,
+            name=normalized_function_name,
+            tag=tag or "latest",
+            hash_key=hash_key,
+            project=project,
+        )
+        if uid:
+            query = query.filter(Function.uid == uid)
```
TomerShor: use `self._find_functions` instead. it returns a `query` object you can then `one_or_none`

TomerShor (changes requested):
A few more comments, and also missing a test for the full flow - create a function, create api gateway, check urls, delete etc. (system test perhaps?)

rokatyy:
@TomerShor we have a system test for the whole api gateway flow (create/update/delete) - https://github.com/mlrun/mlrun/blob/6fa0bfba84687c4bbc9952e4fe38ebeed0c77335/tests/system/runtimes/test_nuclio.py#L589

TomerShor:
> @TomerShor we have a system test for the whole api gateway flow (create/update/delete) -
> 
> https://github.com/mlrun/mlrun/blob/6fa0bfba84687c4bbc9952e4fe38ebeed0c77335/tests/system/runtimes/test_nuclio.py#L589

But it is without the checks on the function's external invocation urls

server/api/db/sqldb/db.py
rokatyy: @TomerShor I've just tried it and for some reason this method doesn't find a correct function for me...

server/api/api/endpoints/nuclio.py
```diff
@@ -123,10 +126,29 @@ async def store_api_gateway(
         auth_info,
     )
     async with server.api.utils.clients.async_nuclio.Client(auth_info) as client:
-        create = not await client.api_gateway_exists(
-            name=gateway,
-            project_name=project,
-        )
+        # get if api gateway exists
```
TomerShor:
```suggestion
```

server/api/api/endpoints/nuclio.py
```diff
@@ -123,10 +126,29 @@ async def store_api_gateway(
...
+        create = False
+        try:
+            existing_api_gateway = await client.get_api_gateway(
+                project_name=project, name=gateway
+            )
+            # check if there are any functions that are not used in api gateway anymore
```
TomerShor:
```suggestion
            # check if any functions were removed from the api gateway
```

server/api/api/endpoints/nuclio.py
```diff
@@ -285,9 +322,46 @@ async def deploy_status(
         tag,
         last_log_timestamp,
         verbose,
+        api_gateways_urls,
     )
 
 
+async def delete_functions_external_invocation_url(
```
TomerShor: can be a private function

server/api/api/endpoints/nuclio.py
```diff
@@ -285,9 +322,46 @@ async def deploy_status(
...
+                project=project,
+                invocation_url=url,
+            )
+        )
+        for function in function_names
+    ]
+    await asyncio.gather(*tasks)
+
+
+async def add_functions_external_invocation_url(
```
TomerShor: can be a private function

server/api/db/sqldb/db.py
TomerShor: That's odd.
Using `_get_function_uid` this way results with 2 db queries - first fetch the tag and get the uid, then attach the uid to the query and get the function.
`_find_functions` optimizes this by doing an outer join on the tags, so we end up with only a single query.
In large scale systems this matters. Also - It's best reusing the existing methods then implementing queries differently for different flows.

I would give it another minute to check why it doesn't yield the same results.

TomerShor (changes requested):
A few more comments below

server/api/db/sqldb/db.py
rokatyy: @TomerShor `_find_functions` performs a request itself, so there is not way to do `one_or_none`. And also in `get_function` method we don't use _find_functions btw

rokatyy: This function doesn't provide an interface for getting a specific function by it's name. This method returns all the functions which names start with given `name`, so that's why we can't use this method.

TomerShor (approved):
Nicely done

PR #5748, opened 2024-06-10
[Feature Store] Avoid unnecessary copy in feature store schema

gtopper:
This can be expensive, especially when transforming the results from a join operation.

Also updates `copy_on_model_validation` from boolean to string, as setting this configuration to a boolean has been deprecated.

### DIFF
--- a/mlrun/common/schemas/feature_store.py
+++ b/mlrun/common/schemas/feature_store.py
@@ -14,7 +14,7 @@
 #
 from typing import Optional
 
-from pydantic import BaseModel, Extra, Field
+import pydantic
 
 from .auth import AuthorizationResourceTypes, Credentials
 from .object import (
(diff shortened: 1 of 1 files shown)


mlrun/common/schemas/feature_store.py
```diff
@@ -27,42 +27,52 @@
...
+
+
+class Feature(FeatureStoreBaseModel):
     name: str
     value_type: str
     labels: Optional[dict] = {}
 
     class Config:
-        extra = Extra.allow
+        extra = pydantic.Extra.allow
```
theSaarco, line 46: Does this apply on top of the base-class? I mean, doesn't the internal class Config override the same class defined in the parent class? The example you point at have the class Config in the parent class, but not in the inherited classes. Possibly pydantic is doing some black magic allowing this, but have you verified this works?

theSaarco (approved):
Given the config no-override is tested and verified working, then this looks good to me.

PR #5749, opened 2024-06-10
[Application] Remove unused argument from `deploy`

TomerShor:
`skip_access_key_auth` was replaced with `authentication_mode` so it is not needed anymore.
BC is not broken because the application runtime feature is still unreleased in GA.


alonmr (approved):
thanks 👍

PR #5729, opened 2024-06-06
[Feature Store] Fix quadratic result size of `/entities` API

gtopper:
in a new, v2 implementation.

Also eliminates duplication of feature set digest objects to further improve performance.

[ML-6249](https://iguazio.atlassian.net/browse/ML-6249)

[ML-6249]: https://iguazio.atlassian.net/browse/ML-6249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5751, opened 2024-06-11
[Feature Store] Fix quadratic result size of `/features` API

gtopper:
in a new, v2 implementation.

Also eliminates duplication of feature set digest objects to further improve performance.

[ML-6765](https://iguazio.atlassian.net/browse/ML-6765)

Based on the same changes for `/entities` in https://github.com/mlrun/mlrun/pull/5729.

Contains the changes in https://github.com/mlrun/mlrun/pull/5748 which as of now is still open.

[ML-6765]: https://iguazio.atlassian.net/browse/ML-6765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/common/schemas/__init__.py
+++ b/mlrun/common/schemas/__init__.py
@@ -99,6 +99,7 @@ from .feature_store import (
     FeatureSetSpec,
     FeatureSetsTagsOutput,
     FeaturesOutput,
+    FeaturesOutputV2,
     FeatureVector,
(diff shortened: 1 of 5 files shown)


theSaarco (approved):
Looks good.

PR #5752, opened 2024-06-11
[Model Monitoring] Fix redeploy a model with monitoring

davesh0812:
We will create MEP only when tracking is enabled. This is because MEP creation requires running set_model_monitoring_credentials(), and doing so beforehand could cause data store conflicts. 

Tracking can be enabled or disabled using set_tracking(enable_tracking=True/False), and redeployed accordingly.

https://iguazio.atlassian.net/browse/ML-4442

### DIFF
--- a/mlrun/runtimes/nuclio/serving.py
+++ b/mlrun/runtimes/nuclio/serving.py
@@ -312,15 +312,18 @@ class ServingRuntime(RemoteRuntime):
         sample: Optional[int] = None,
         stream_args: Optional[dict] = None,
         tracking_policy: Optional[Union["TrackingPolicy", dict]] = None,
+        enable_tracking: bool = True,
     ) -> None:
         """apply on your serving function to monitor a deployed model, including real-time dashboards to detect drift
            and analyze performance.
 
-        :param stream_path:     Path/url of the tracking stream e.g. v3io:///users/mike/mystream
-                                you can use the "dummy://" path for test/simulation.
-        :param batch:           Micro batch size (send micro batches of N records at a time).
-        :param sample:          Sample size (send only one of N records).
-        :param stream_args:     Stream initialization parameters, e.g. shards, retention_in_hours, ..
+        :param stream_path:         Path/url of the tracking stream e.g. v3io:///users/mike/mystream
+                                    you can use the "dummy://" path for test/simulation.
+        :param batch:               Micro batch size (send micro batches of N records at a time).
+        :param sample:              Sample size (send only one of N records).
+        :param stream_args:         Stream initialization parameters, e.g. shards, retention_in_hours, ..
+        :param enable_tracking:     Enabled/ Disable tracking on current serving runtime.
+                                    Default True (enabling tracking).
 
                                 example::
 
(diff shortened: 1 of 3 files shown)


mlrun/runtimes/nuclio/serving.py
```diff
@@ -312,15 +312,18 @@ def set_tracking(
...
-                                you can use the "dummy://" path for test/simulation.
-        :param batch:           Micro batch size (send micro batches of N records at a time).
-        :param sample:          Sample size (send only one of N records).
-        :param stream_args:     Stream initialization parameters, e.g. shards, retention_in_hours, ..
+        :param stream_path:         Path/url of the tracking stream e.g. v3io:///users/mike/mystream
+                                    you can use the "dummy://" path for test/simulation.
+        :param batch:               Micro batch size (send micro batches of N records at a time).
+        :param sample:              Sample size (send only one of N records).
+        :param stream_args:         Stream initialization parameters, e.g. shards, retention_in_hours, ..
+        :param enable_tracking:     Enabled/ Disable tracking on current serving runtime.
```
assaf758: > on current serving runtime
This is confusing.
```suggestion
        :param enable_tracking:     Enabled/Disable model-monitoring tracking.
```

mlrun/runtimes/nuclio/serving.py
```diff
@@ -312,15 +312,18 @@ def set_tracking(
...
+                                    Default True (enabling tracking).
```
assaf758:
```suggestion
                                    Default True (tracking enabled).
```

mlrun/serving/v2_serving.py
```diff
@@ -542,48 +542,67 @@ def _init_endpoint_record(
         function_uri=graph_server.function_uri, versioned_model=versioned_model_name
     ).uid
 
+    model_ep = None
     # If model endpoint object was found in DB, skip the creation process.
     try:
-        mlrun.get_run_db().get_model_endpoint(project=project, endpoint_id=uid)
-
+        model_ep = mlrun.get_run_db().get_model_endpoint(
```
assaf758, line 546: seems you are trying get the mep even if tracking is not enabled, why?

mlrun/serving/v2_serving.py
```diff
@@ -542,48 +542,67 @@ def _init_endpoint_record(
...
+            project=project, endpoint_id=uid
+        )
     except mlrun.errors.MLRunNotFoundError:
         logger.info("Creating a new model endpoint record", endpoint_id=uid)
```
assaf758: this log will appear even if tracking is not enabled, which is wrong in this case.
also, I think having the mep creation logic inside exception handling is confusing - better to raise a flag and then do the mep creation based on the flag value.

tests/system/model_monitoring/test_app.py
```diff
@@ -999,3 +1004,85 @@ def test_all(self) -> None:
             assert res_dict[
                 "has_all_the_events"
             ], f"For {res_dict['model_name']} Not all the events were saved"
+
+
+class TestTracking(TestAllKindOfServing):
+    project_name = "test-tracking"
+    # Set image to "<repo>/mlrun:<tag>" for local testing
+    image: typing.Optional[str] = "docker.io/davesh0812/mlrun:1.7.0"
```
assaf758: please remove your image

mlrun/serving/v2_serving.py
davesh0812: It is  for the case that the tacking were enabled and now the user is disabling it, you can see I patch the exist MEP in the end of this function.

assaf758 (approved):

PR #5753, opened 2024-06-11
[Nuclio] Allow having api gateways with the same name in different projects

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-6784

The goal of this PR is to unblock creation of api gateways with the same names, but in different project. I implemented this by populating a project name to an api gateway name on server side. 

BC tests failed because of changing the `gateway` to `name`, but it doesn't really break anything, because that's the 1st release when we introduce this feature.

### DIFF
--- a/mlrun/common/helpers.py
+++ b/mlrun/common/helpers.py
@@ -42,6 +42,6 @@ def generate_api_gateway_name(project: str, name: str) -> str:
     :param project: project name
     :param name: api gateway name
 
-    :return: "{project}-{name}" if project name is given else name
+    :return: the resolved api gateway name
     """
     return f"{project}-{name}" if project else name
--- a/mlrun/common/helpers.py
+++ b/mlrun/common/helpers.py
@@ -34,3 +34,14 @@ def parse_versioned_object_uri(
         uri = uri[:loc]
 
     return project, uri, tag, hash_key
+
+
+def generate_api_gateway_name(project: str, name: str) -> str:
+    """
+    Generate a unique (within project) api gateway name
+    :param project: project name
+    :param name: api gateway name
+
+    :return: "{project}-{name}" if project name is given else name
+    """
+    return f"{project}-{name}" if project else name
--- a/mlrun/common/schemas/api_gateway.py
+++ b/mlrun/common/schemas/api_gateway.py
@@ -17,6 +17,7 @@ from typing import Optional
 
 import pydantic
 
+import mlrun.common.constants as mlrun_constants
 import mlrun.common.types
 from mlrun.common.constants import MLRUN_FUNCTIONS_ANNOTATION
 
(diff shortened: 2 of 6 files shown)


mlrun/common/schemas/api_gateway.py
```diff
@@ -100,7 +101,49 @@ def get_function_names(self):
...
+        # replace api gateway name
+        # in Nuclio, api gateways are named as `<project>-<mlrun-api-gateway-name>`
+        # remove the project prefix from the name if it exists
+        project_name = self.metadata.labels.get(
+            mlrun_constants.MLRunInternalLabels.nuclio_project_name
+        )
+        if project_name and self.spec.name.startswith(f"{project_name}-"):
+            self.spec.name = self.spec.name[len(project_name) + 1 :]
+            self.metadata.name = self.spec.name
+        return self
```
TomerShor, line 145: Why do you need to return `self` in every method?

mlrun/common/schemas/api_gateway.py
```diff
@@ -100,7 +101,49 @@ def get_function_names(self):
...
+        self._enrich_mlrun_function_names()
+        return self
+
+    def replace_nuclio_names_with_mlrun_names(self):
+        self._replace_nuclio_api_gateway_name_with_mlrun_name()
+        self._replace_nuclio_function_names_with_mlrun_names()
+        return self
+
+    def _replace_nuclio_function_names_with_mlrun_names(self):
+        # replace function names from nuclio names to mlrun names
```
TomerShor, line 116: this comment is a duplicate of the function name, so it doesn't add much.
A better comment will say something about the use of URIs in the annotations etc..

server/api/utils/clients/async_nuclio.py
```diff
@@ -65,15 +65,23 @@ async def list_api_gateways(
...
             parsed_api_gateways[name] = mlrun.common.schemas.APIGateway.parse_obj(
                 gw
-            ).replace_nuclio_names_with_mlrun_uri()
+            ).replace_nuclio_names_with_mlrun_names()
         return parsed_api_gateways
 
     async def api_gateway_exists(self, name: str, project_name: str = None):
+        # enrich api gateway name with project prefix
+        if project_name:
+            name = f"{project_name}-{name}"
```
TomerShor: duplicated across several functions - move to a common method

server/api/api/endpoints/nuclio.py
TomerShor, line 1: A general note about **all** of the `api-gateways` endpoints:
I would change all occurrences of `{gateway}` to `{name}`, as to not be confused with the name of the resources vs the resource itself.

E.g: in `store_api_gateway`:
```python
@router.put(
    "/projects/{project}/api-gateways/{name}",
    response_model=mlrun.common.schemas.APIGateway,
    response_model_exclude_none=True,
)
async def store_api_gateway(
    project: str,
    name: str,
    api_gateway: mlrun.common.schemas.APIGateway,
):
``` 


Similar to the function endpoints:
```
@router.get("/projects/{project}/nuclio/{name}/deploy")
```

TomerShor (changes requested):
Looking good, added some comments below

mlrun/common/schemas/api_gateway.py
rokatyy: @TomerShor to be able to return api gw right from the function

server/api/api/endpoints/nuclio.py
rokatyy: @TomerShor changed

mlrun/common/helpers.py
```diff
@@ -34,3 +34,14 @@ def parse_versioned_object_uri(
...
     return project, uri, tag, hash_key
+
+
+def generate_api_gateway_name(project: str, name: str) -> str:
+    """
+    Generate a unique (within project) api gateway name
+    :param project: project name
+    :param name: api gateway name
+
+    :return: "{project}-{name}" if project name is given else name
```
TomerShor:
```suggestion
    :return: the resolved api gateway name
```

TomerShor (changes requested):
really minor, otherwise awesome

TomerShor (approved):

TomerShor:
BC tests fail due to the change in the url param name. 
However - this is only a naming change in the BE but the URL stays exactly the same, hence BC is kept.

liranbg (approved):

PR #5754, opened 2024-06-12
[Error handling] Do not populate exception string when error message is given

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-6636

Old message example:
image

New format:
image


TomerShor (approved):
Nice 🚀

PR #5755, opened 2024-06-12
[API] Fix comment being logged

alonmr:
For some unclear reason the comment was being outputted in the api logs. This clears it.


TomerShor (approved):
I believe the "unclear reason" was indeed the space  between `:` and `'`

PR #5760, opened 2024-06-13
[API] Improve log collection on start-up

TomerShor:
In a large-scale systems, MLRun API can possibly not move to online due to sending start_logs requests to the log collector on runs that weren’t requested.

To avoid this, we do the following changes:
1. When listing runs that have `requested_logs=False`, we check the amount of runs that we need to start logs for.
If we get more than 150 runs (configurable) - skip start logs and mark all of them as `requested_logs=True`.
2. Increase the default concurrency from 15 to 50 runs
3. First query only the run uids from the DB (faster query), then if we’re under the limit query the actual runs.
4. Start log collection before cleanups + monitoring.

Resolves https://iguazio.atlassian.net/browse/ML-6806


server/api/main.py
```diff
@@ -211,10 +211,12 @@ async def move_api_to_online():
...
         # runs cleanup/monitoring is not needed if we're not inside kubernetes cluster
         if get_k8s_helper(silent=True).is_running_inside_kubernetes_cluster():
-            _start_periodic_cleanup()
-            _start_periodic_runs_monitoring()
             await _start_periodic_logs_collection()
             await _start_periodic_stop_logs()
+            _start_periodic_cleanup()
+            _start_periodic_runs_monitoring()
+
+    logger.info("API is online")
```
alonmr: There is already (I believe uvicorn internal log) log for this 
```
[info] Application startup complete
```

server/api/main.py
```diff
@@ -289,18 +292,35 @@ async def _verify_log_collection_started_on_startup(
             states=mlrun.runtimes.constants.RunStates.terminal_states(),
         )
     )
-    if runs:
-        logger.debug(
-            "Found runs which require logs collection on startup",
-            runs_uids=[run.get("metadata", {}).get("uid", None) for run in runs],
-        )
+    if runs_uids:
+        if len(runs_uids) <= config.log_collector.start_logs_startup_run_limit:
```
alonmr: You may need to cast it. Also mlconf as convention
```suggestion
        if len(runs_uids) <= int(mlrun.mlconf.log_collector.start_logs_startup_run_limit):
```

server/api/main.py
```diff
@@ -289,18 +292,35 @@ async def _verify_log_collection_started_on_startup(
...
+            await _start_log_and_update_runs(
+                start_logs_limit=start_logs_limit,
+                db_session=db_session,
+                runs_uids=runs_uids,
+                best_effort=True,
+            )
+        else:
+            logger.warning(
+                "Found too many runs which require logs collection on startup, "
+                "marking them as requested logs collection",
```
alonmr:
```suggestion
                "Amount of runs requiring logs collection on startup exceeds configured limit, "
                "marking them as requested logs collection",
```

server/api/main.py
```diff
@@ -289,18 +292,35 @@ async def _verify_log_collection_started_on_startup(
...
+                runs_count=len(runs_uids),
+                limit=config.log_collector.start_logs_startup_run_limit,
```
alonmr:
```suggestion
                start_logs_startup_run_limit =mlrun.mlconf.log_collector.start_logs_startup_run_limit,
```

alonmr (commented):
A few comments but I feel like this is very destructive and not sure if I would merge to 1.6.4 unless the pain is big and I'm not aware.
Point to ponder - if you have more the 150 runs, take the first 150 and request logs for them, about the other ones you can mark them as requested

server/api/main.py
```diff
@@ -232,11 +233,11 @@ async def move_api_to_online():
         server.api.initial_data.update_default_configuration_data()
         # runs cleanup/monitoring is not needed if we're not inside kubernetes cluster
         if get_k8s_helper(silent=True).is_running_inside_kubernetes_cluster():
-            _start_periodic_cleanup()
-            _start_periodic_runs_monitoring()
             _start_periodic_pagination_cache_monitoring()
             await _start_periodic_logs_collection()
             await _start_periodic_stop_logs()
+            _start_periodic_cleanup()
```
alonmr: why move below?

server/api/main.py
```diff
@@ -342,19 +374,23 @@ async def _initiate_logs_collection(start_logs_limit: asyncio.Semaphore):
...
+        if run_uids:
             logger.debug(
                 "Found runs which require logs collection",
-                runs_uids=[run.get("metadata", {}).get("uid", None) for run in runs],
+                runs_uids=run_uids,
+            )
+            await _start_log_and_update_runs(
+                start_logs_limit=start_logs_limit,
+                db_session=db_session,
+                runs_uids=run_uids,
```
alonmr: maybe align the variable name to `runs_uids`

alonmr (commented):
minor

server/api/main.py
TomerShor: What variable?
There is only `run_uids`

server/api/main.py
TomerShor: Because we don't ensure run logs, we want to have a chance to collect logs for runs before they are cleaned-up

server/api/main.py
alonmr: `runs_uids=run_uids` should be `runs_uids=runs_uids` (missing 's' in the rhs)

server/api/main.py
alonmr: so why runs monitoring as well? I think this is not a trivial change and we should at least discuss it. cleanup should only delete pods that were last updated over 4 hours ago so their logs should have been collected before the API starts unless something else very bad happened.

TomerShor: OK, I'll leave it out of this PR, as this is not the actual improvement here.

liranbg (approved):

alonmr (approved):

PR #5761, opened 2024-06-13
[Runs] Do not ensure run logs collected when runs are terminated [1.6.x]

TomerShor:
When runs reach terminal state, the monitoring function ensures logs were collected for this run by checking that a log file exists for this run. If not - it requests the run logs straight from k8s, bypassing the log collector as it assumes something was wrong with it.

To decrease the IO load on the chief, we do not ensure logs were collected at all, as logs are a best-effort operation.

Resolves https://iguazio.atlassian.net/browse/ML-6807


alonmr (approved):
Though I'm not sure how would we get to this situation and if we expect the log collector to kick in for these runs. I agree this is redundant and should be removed.

TomerShor:
Implemented in https://github.com/mlrun/mlrun/pull/5907 for development

PR #5767, opened 2024-06-16
[Runs] Do not enrich runs in server side launcher [1.6.x]

alonmr:
https://iguazio.atlassian.net/browse/ML-6817

Artifacts are not required on the server side.
To decide whether enrichment is necessary, we have two formats for runs:

* Standard - no enrichment; data is pulled as-is from the database.
* Full - includes enriched artifacts.

Clients must use the Full format for backward compatibility.

### DIFF
--- a/mlrun/common/schemas/runs.py
+++ b/mlrun/common/schemas/runs.py
@@ -27,6 +27,7 @@ class RunIdentifier(pydantic.BaseModel):
 
 # In 1.7 should be moved to mlrun.common.formatters.run.py
 class RunsFormat(mlrun.common.types.StrEnum):
+    # No enrichment, data is pulled as-is from the database.
     standard = "standard"
 
     # Performs run enrichment, including the run's artifacts. Only available for the `get` run API.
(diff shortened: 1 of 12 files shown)


mlrun/common/schemas/runs.py
```diff
@@ -16,8 +16,18 @@
...
 
 class RunIdentifier(pydantic.BaseModel):
     kind: typing.Literal["run"] = "run"
     uid: typing.Optional[str]
     iter: typing.Optional[int]
+
+
+# In 1.7 should be moved to mlrun.common.formatters.run.py
+class RunsFormat(mlrun.common.types.StrEnum):
+    standard = "standard"
```
TomerShor, line 31: specify here in a comment what this format means

server/api/crud/runs.py
```diff
@@ -505,6 +489,33 @@ def _get_artifacts_from_uris(
             )
         return artifacts
 
+    def _enrich_run(
```
TomerShor:
```suggestion
    def _enrich_run_artifacts(
```

TomerShor (changes requested):

liranbg (approved):

TomerShor (approved):

PR #5773, opened 2024-06-17
[Project] ensure_project: instead of list all projects, get project by name

roei3000b:
In the ensure_project function, instead of getting all project names and then check programmatically if the project name exists,  add it to the sql query.

https://iguazio.atlassian.net/browse/ML-6840

### DIFF
--- a/server/api/utils/projects/member.py
+++ b/server/api/utils/projects/member.py
@@ -44,6 +44,7 @@ class Member(abc.ABC):
             db_session,
             format_=mlrun.common.formatters.ProjectFormat.name_only,
             leader_session=auth_info.session,
+            names=[name],
         )
         if name not in project_names.projects:
             raise mlrun.errors.MLRunNotFoundError(f"Project {name} does not exist")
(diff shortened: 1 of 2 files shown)


server/api/utils/projects/member.py
```diff
@@ -44,6 +44,7 @@ def ensure_project(
             db_session,
             format_=mlrun.common.formatters.ProjectFormat.name_only,
             leader_session=auth_info.session,
+            names=[name],
```
alonmr: why not do get project instead?

server/api/utils/projects/member.py
```diff
@@ -40,12 +40,14 @@ def ensure_project(
...
+        project = self.get_project(
             db_session,
             format_=mlrun.common.formatters.ProjectFormat.name_only,
             leader_session=auth_info.session,
+            from_leader=False,
+            name=name,
         )
-        if name not in project_names.projects:
+        if not project:
             raise mlrun.errors.MLRunNotFoundError(f"Project {name} does not exist")
```
alonmr, line 56: I agree we should keep this but add a comment that `get_project` should raise 404 but we keep this for sanity

server/api/utils/projects/remotes/nop_follower.py
```diff
@@ -71,6 +71,8 @@ def get_project(
         self, session: sqlalchemy.orm.Session, name: str
     ) -> mlrun.common.schemas.Project:
         # deep copy so we won't accidentally get changes from tests
```
liranbg: move comment down

liranbg (approved):

PR #5775, opened 2024-06-17
[Runtimes] Refactor node selector handling in function run

yaelgen:
Jira ticket: [ML-3874](https://iguazio.atlassian.net/browse/ML-3874) 
This PR refactors the default function node selector task to improve how node selectors are managed and applied. Previously, the project-level node selector was merged into the runtime object, which can cause issues when functions being moved between projects, as project information became embedded in the function object. 
  
  - The new implementation places the merged value of the node selector (combining function-level and project-level) on the run object.
  - The `RunObject` spec now includes a node selector field, enriched during the function run. Since `RunObject` is namespaced to a single project, it is acceptable to contain project-specific information.
  - Added functionality to handle the removal of labels from the project by using keys with empty values, while k8s typically ignores labels with empty values by default, from python it accepts both empty strings and None as values, and does not ignore the label. 
  
  Handling by Runtime Kind:
  - `Kubejob` - Enriched on pod creation.
  - `Databricks` - Same behavior as kubejob, enriched on pod creation.
  - `Nuclio` -  No Run object, not enriching, letting nuclio side handle it
  - `Spark` -  Enriching the nodeSelector on the run, and enriching the driverNodeSelector and executorNodeSelector from the project on job creation.
  - `MPI` -  Enriching on the run, and adding to the CRD from the run instead of the runtime.
  - `Dask` -  Will be handled in a follow-up PR  

[ML-3874]: https://iguazio.atlassian.net/browse/ML-3874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Fixes
https://iguazio.atlassian.net/browse/ML-6612
https://iguazio.atlassian.net/browse/ML-6529
https://iguazio.atlassian.net/browse/ML-6633


mlrun/runtimes/daskjob.py
```diff
@@ -554,3 +554,35 @@ def _run(self, runobj: RunObject, execution):
...
+        Example:
+        Suppose self.spec.node_selector = {"type": "gpu", "zone": "us-east-1"}
+        and project_node_selector = {"type": "cpu", "environment": "production"}.
+        After the merge, the resulting node_selector will be:
+        {"type": "gpu", "zone": "us-east-1", "environment": "production"}
+
+        Note:
+        - The merge uses the ** operator, also known as the "unpacking" operator in Python,
+          combining key-value pairs from each dictionary. Later dictionaries take precedence
+          when there are conflicting keys.
```
TomerShor: This docstring is better suited for `select_non_empty_fields` than dask's `enrich_runrime_spec`, as it is just describing the functionality.

mlrun/runtimes/nuclio/function.py
```diff
@@ -670,6 +670,37 @@ def with_service_type(
...
+        {"type": "gpu", "zone": "us-east-1", "environment": "production"}
+
+        Note:
+        - The merge uses the ** operator, also known as the "unpacking" operator in Python,
+          combining key-value pairs from each dictionary. Later dictionaries take precedence
+          when there are conflicting keys.
+        """
+        self.spec.node_selector = mlrun.utils.helpers.select_non_empty_fields(
+            {**project_node_selector, **self.spec.node_selector}
+        )
```
TomerShor: 1. This is a duplicate of the dask function
2. Same note about the docstring as above
3. Discussed offline - can possibly be removed from Nuclio runtime as the enrichment is done in the Nuclio Dashboard.

server/api/launcher.py
```diff
@@ -175,6 +175,60 @@ def launch(
...
+            artifact_path=artifact_path,
+            workdir=workdir,
+            notifications=notifications,
+            state_thresholds=state_thresholds,
+        )
+        if runtime._get_db():
+            project = runtime._get_db().get_project(run.metadata.project)
+            if project.spec.default_function_node_selector:
+                mlrun.utils.logger.info(
+                    "Enriching node selector from project",
```
TomerShor:
```suggestion
                    "Enriching run node selector from project",
```

server/api/runtime_handlers/sparkjob/spark3job.py
```diff
@@ -694,5 +702,35 @@ def _enrich_job(
             )
         return
 
+    def _enrich_node_selectors_with_project_level(self, runtime, job, project_name):
```
TomerShor:
```suggestion
    def _enrich_node_selectors_from_project(self, runtime, job, project_name):
```

server/api/runtime_handlers/sparkjob/spark3job.py
```diff
@@ -694,5 +702,35 @@ def _enrich_job(
...
+        # After adding the runtime executor and driver node selectors to the job object,
+        # we enrich the node selectors with any additional selectors from the project-level configuration, if exists.
+        if runtime._get_db() and project_name:
```
TomerShor: Why do you need to check `if runtime._get_db()`?

server/api/runtime_handlers/sparkjob/spark3job.py
```diff
@@ -694,5 +702,35 @@ def _enrich_job(
...
+                    update_in(
+                        job,
+                        "spec.executor.nodeSelector",
+                        mlrun.utils.helpers.select_non_empty_fields(
+                            {
+                                **project.spec.default_function_node_selector,
+                                **runtime.spec.executor_node_selector,
+                            }
+                        ),
+                    )
```
TomerShor: And what if we don't have neither of them (executor/driver selectors)?
don't we still want to enrich with project node selectors?

tests/projects/test_project.py
```diff
@@ -1259,7 +1259,7 @@ def test_function_receives_project_default_image():
     assert enriched_function.spec.image == new_default_image
 
 
-def test_function_receives_project_default_function_node_selector():
+def test_function_does_not_receives_project_default_function_node_selector():
```
TomerShor:
```suggestion
def test_function_not_enriched_with_project_default_function_node_selector():
```

TomerShor (changes requested):

server/api/runtime_handlers/sparkjob/spark3job.py
yaelgen: it is being enriched prior to this function

mlrun/utils/helpers.py
```diff
@@ -1606,6 +1610,30 @@ def additional_filters_warning(additional_filters, class_name):
         )
 
 
+def merge_with_precedence(first_dict, second_dict):
```
TomerShor: add type hints

server/api/crud/runtimes/nuclio/function.py
```diff
@@ -453,13 +458,18 @@ def _set_build_params(function, nuclio_spec, builder_env, project, auth_info=Non
...
         "1.5.20", "1.6.10"
     ):
         if function.spec.node_selector:
-            nuclio_spec.set_config("spec.nodeSelector", function.spec.node_selector)
+            nuclio_spec.set_config(
+                "spec.nodeSelector",
+                _enrich_function_node_selector_with_project(
+                    db_session, function.metadata.project, function.spec.node_selector
+                ),
+            )
```
TomerShor: what if `function.spec.node_selector` is empty but `project.spec.default_function_node_selector` is not empty?
resolve the node selector first, then check:
```suggestion
        if node_selector := _enrich_function_node_selector_with_project(
                    db_session, function.metadata.project, function.spec.node_selector
                ):
                nuclio_spec.set_config(
                "spec.nodeSelector",
                node_selector,
            )
```

server/api/launcher.py
```diff
@@ -175,6 +175,60 @@ def launch(
...
+            out_path=out_path,
+            artifact_path=artifact_path,
+            workdir=workdir,
+            notifications=notifications,
+            state_thresholds=state_thresholds,
+        )
+        if runtime._get_db():
+            project = runtime._get_db().get_project(run.metadata.project)
+            if project.spec.default_function_node_selector:
+                mlrun.utils.logger.info(
```
TomerShor: not THAT important
```suggestion
                mlrun.utils.logger.debug(
```

tests/api/runtimes/test_nuclio.py
```diff
@@ -402,6 +403,26 @@ def test_compile_function_config_with_special_character_labels(
         ) = server.api.crud.runtimes.nuclio.function._compile_function_config(function)
         assert config["metadata"]["labels"].get(key) == val
 
+    def test_compile_function_config_with_node_selector(
```
TomerShor: this is what you are actually testing
```suggestion
    def test_compile_function_config_node_selector_enriched_from_project(
```

TomerShor (commented):
Looking good!
Added a few more comments

mlrun/execution.py
```diff
@@ -207,6 +208,11 @@ def labels(self):
         """Dictionary with labels (read-only)"""
         return deepcopy(self._labels)
 
+    @property
+    def node_selector(self):
+        """Dictionary with labels (read-only)"""
```
liranbg:
```suggestion
        """Dictionary with node selectors (read-only)"""
```

mlrun/projects/operations.py
```diff
@@ -194,6 +194,7 @@ def my_pipe(url=""):
         project = project_object or pipeline_context.project
         local = pipeline_context.is_run_local(local)
         task.metadata.labels = task.metadata.labels or labels or {}
+        project_name = project.metadata.name if project else ""
```
liranbg: what's this fix about? can you elaborate?

mlrun/utils/helpers.py
```diff
@@ -1259,6 +1259,10 @@ def _fill_project_path_template(artifact_path, project):
     return artifact_path
 
 
+def select_non_empty_fields(input_dict: dict) -> dict:
```
```suggestion
def to_non_empty_values_dict(input_dict: dict) -> dict:
```
perhaps says more about what it does?

server/api/crud/runtimes/nuclio/function.py
```diff
@@ -486,6 +496,21 @@ def _set_function_scheduling_params(function, nuclio_spec):
...
+    if db_session and project_name:
+        project = server.api.utils.singletons.db.get_db().get_project(
+            session=db_session, name=project_name
+        )
+        if project.spec.default_function_node_selector:
+            return mlrun.utils.helpers.merge_with_precedence(
+                project.spec.default_function_node_selector,
+                function_node_selector,
+            )
+    return function_node_selector
```
liranbg: uncertain why it has been decided to carry the db_session all the way here and get the project from db while
server launcher enriches the runtime for us @ https://github.com/mlrun/mlrun/blob/5883a9ea1d4dc8035ca6112cfc21a556f285b8a3/server/api/launcher.py#L213

you might have wanted to enrich the function config before sending it to nuclio, right? but the outcome is that you have nuclio CRD with the node selector and not the mlrun-function itself so I think this is yet to accomplish the desire of "export/import" project without making it dirty with per-ssytem node selectors.

perhaps what we can do here is, instead of carry the db session to get the project, rather pass the project object that we have already resolved. wdyt?

server/api/launcher.py
```diff
@@ -175,6 +175,60 @@ def launch(
...
+            hyper_param_options=hyper_param_options,
+            verbose=verbose,
+            scrape_metrics=scrape_metrics,
+            out_path=out_path,
+            artifact_path=artifact_path,
+            workdir=workdir,
+            notifications=notifications,
+            state_thresholds=state_thresholds,
+        )
+        if runtime._get_db():
```
liranbg, line 226: why do we need this explicit check?

server/api/launcher.py
```diff
@@ -175,6 +175,60 @@ def launch(
...
+            if project.spec.default_function_node_selector:
+                mlrun.utils.logger.debug(
+                    "Enriching run node selector from project",
+                    project_name=run.metadata.project,
+                    project_node_selector=project.spec.default_function_node_selector,
+                )
+                run.spec.node_selector = mlrun.utils.helpers.merge_with_precedence(
+                    project.spec.default_function_node_selector,
+                    runtime.spec.node_selector,
+                )
```
liranbg, line 237: can add a "pre_run_enrichment" and explain in a sentence that we have decided to do the enrichment on run to avoid dirty-ing the runtime itself as the label selector is per-system

liranbg (changes requested):
some food for thoughts

server/api/crud/runtimes/nuclio/function.py
yaelgen: If I understand correctly, the launcher flow doesn't apply to the deploying of Nuclio functions, which is why I chose to handle it this way. Regarding your suggestion to pass the project object instead of the db_session,
I agree it's cleaner. However, could you clarify how this change would specifically address the import/export issue? not sure I got what you meant there

yaelgen: To document here as well after follow up discussion - this change creates a discrepancy between the behavior of nuclio functions and mlrun functions. The nuclio function will be affected by per-system node selectors, making it "dirty," while the mlrun function will not. Typically, we let mlrun dictate the behavior of its runtimes, but in this case, we have decided to let nuclio handle the merge logic on it's own. Consequently, we will remove all extra behavior related to nuclio functions from mlrun in this context.

mlrun/projects/operations.py
yaelgen: removed

liranbg (approved):

TomerShor (approved):

PR #5777, opened 2024-06-17
[Artifacts] Ensure metadata dictionary when migrating legacy artifacts to v2

TomerShor:
If an old artifact is invalid in the db (doesn't have the expected fields in its `struct`), the `metadata` field can be empty even after converting it from legacy artifact.

Since we defaulted the metadata to `None` if doesn't exist, this can cause later `.get` calls to explode.
Defaulting to an empty dictionary fixes it.

Related to https://iguazio.atlassian.net/browse/ML-6795

### DIFF
--- a/server/api/initial_data.py
+++ b/server/api/initial_data.py
@@ -577,7 +577,7 @@ def _migrate_artifacts_batch(
             link_artifact_ids.append(artifact.id)
             continue
 
-        artifact_metadata = artifact_dict.get("metadata", {})
+        artifact_metadata = artifact_dict.get("metadata", None) or {}
 
         # producer_id - the current uid value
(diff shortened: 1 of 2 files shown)


server/api/initial_data.py
```diff
@@ -577,7 +577,7 @@ def _migrate_artifacts_batch(
             link_artifact_ids.append(artifact.id)
             continue
 
-        artifact_metadata = artifact_dict.get("metadata", None)
+        artifact_metadata = artifact_dict.get("metadata", {})
```
alonmr: Just to be extra safe I think you should do
```
artifact_dict.get("metadata") or {}
```
because `{"metadata": None}.get("metadata", {})` returns None

alonmr (approved):

PR #5782, opened 2024-06-18
[Run] Set the default run kind to local only if it does not already exist

moranbental:
Creating a Spark function called `get_or_create_context` and running it as a scheduler task. The scheduler specifies `spark` as the run type, however when the run is started, the spec changes to `local`, with no owner field and no other labels.
This occurred due to this pr: https://github.com/mlrun/mlrun/pull/5709 when executing `get_or_create_context`, we set the default run kind to local but we actually override all labels, even if they already exist in the spec (all spec config can be found in the `MLRUN_EXEC_CONFIG` env var). 

https://iguazio.atlassian.net/browse/ML-6750


mlrun/run.py
```diff
@@ -293,10 +293,13 @@ def get_or_create_ctx(
     newspec["metadata"]["project"] = (
         newspec["metadata"].get("project") or project or mlconf.default_project
     )
+
     newspec["metadata"].setdefault("labels", {})
-    newspec["metadata"]["labels"] = {
-        mlrun_constants.MLRunInternalLabels.kind: RuntimeKinds.local
-    }
+
+    # Only if kind doesn't exist in the spec, add the default runtime kind to local
```
roei3000b: I suggest instead of directly explain the setdefault functionality, explain here why by default we want to set the runtime kind to local.

liranbg: ⬆️

tests/run/test_run.py
```diff
@@ -320,3 +321,11 @@ def test_get_or_create_ctx_run_kind_local_from_function():
     )
     assert run.state() == "completed"
     assert run.output("return") == "local"
+
+
+def test_get_or_create_ctx_run_kind_exists_in_mlrun_exec_config():
+    os.environ["MLRUN_EXEC_CONFIG"] = (
+        '{"spec":{},"metadata":{"uid":"123411", "name":"tst", "labels": {"kind": "spark"}}}'
+    )
```
liranbg: use pytest setenv and not directly on `os` as it carries to other tests as well

liranbg (changes requested):

roei3000b (approved):

alonmr (approved):

liranbg (approved):

PR #5785, opened 2024-06-18
[FeatureStore] Fix saving attributes in targets

tomerm-iguazio:
1) Fixed saving attributes in remote + local ingest.
2) Added tests for both cases.

[ML-6876](https://iguazio.atlassian.net/browse/ML-6876)

[ML-6876]: https://iguazio.atlassian.net/browse/ML-6876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/system/feature_store/test_feature_store.py
```diff
@@ -4818,6 +4820,45 @@ def test_merge_different_number_of_entities(self):
         ).to_dataframe()
         assert_frame_equal(expected_all, df, check_dtype=False)
 
+    @pytest.mark.parametrize("local", [True, False])
+    def test_attributes_in_offline_target(self, local):
```
gtopper: It's not specific to offline target. Also, do we need to add a test or can we add this check elsewhere?

tests/system/feature_store/test_feature_store.py
```diff
@@ -4818,6 +4820,45 @@ def test_merge_different_number_of_entities(self):
...
+            "no_sql_target", attributes={"test_key_online": "test_value_online"}
+        )
+        source = ParquetSource("test_source", path=v3io_parquet_source_path)
+        feature_set.ingest(
+            source=source,
+            targets=[offline_target, online_target],
+            run_config=run_config,
+        )
+        result_offline_target = get_offline_target(feature_set)
+        result_online_target = get_online_target(feature_set)
```
gtopper, line 4858: This just demonstrates that the attributes are still in the feature set object. What we actually want to test is that they are stored in the database and retrieved successfully.

From a quick debugger session, it actually seems like that's not happening.

tomerm-iguazio: It is saved in `fs.status.targets`, not in `fs.spec.targets.`
 I also added a part to load it from the run_db to test.

Does this answer your concern?

tests/system/feature_store/test_feature_store.py
tomerm-iguazio: rename test.

I still think it should be a separated test that target this specific problem:)

tests/system/feature_store/test_feature_store.py
```diff
@@ -4818,6 +4820,50 @@ def test_merge_different_number_of_entities(self):
...
+
+        assert result_offline_target.attributes == offline_target.attributes
+        assert result_online_target.attributes == online_target.attributes
+
+        fs = self._run_db.get_feature_set("attributes_fs")
```
gtopper: I guess just name it better
```suggestion
        read_back_feature_set = self._run_db.get_feature_set("attributes_fs")
```
or something like that.

tomerm-iguazio: fixed

gtopper (approved):

PR #5787, opened 2024-06-18
[Artifacts] Fix tag filtering when reading artifact

TomerShor:
When getting a single artifact with a `tag` filter, we used to query the tag first, get the object uid from it then add the `uid` filter to the artifact query.
The problem was when we have multiple iterations - each iteration has the same tag (e.g. `latest`), but when fetching the tag from the db, only the first hit got returned (see https://github.com/mlrun/mlrun/blob/development/server/api/db/sqldb/db.py#L3856-L3861). This means that getting the specific iteration which is not 1 returned the uid connected to iteration 1, so the artifact was not found.

This PR fixes it by joining the tag table in the same query instead of querying the tags first for the UID.

Fixes https://iguazio.atlassian.net/browse/ML-6878


server/api/db/sqldb/db.py
```diff
@@ -781,8 +774,8 @@ def read_artifact(
         artifact = db_artifact.full_object
 
         # If connected to a tag add it to metadata
-        if artifact_tag_uid:
-            self._set_tag_in_artifact_struct(artifact, computed_tag)
+        if tag:
+            self._set_tag_in_artifact_struct(artifact, tag)
```
theSaarco: Looks like there's a change in behavior here - we'll always add tag to the object, while in the previous behavior a tag was added to the object only if the user explicitly asked for it.

theSaarco (approved):
Sababa.

liranbg (approved):

PR #5788, opened 2024-06-18
[Artifacts] Adding configurable limit for list_artifacts [1.6.x]

liranbg:
⚠️ Change

This pr introduces a `limit` for when fetching artifacts from DB. This ensures that when ui/sdk lists artifacts, it can limit the amount of fetched artifacts.
The reason for this change is mainly because when UI screens requests all artifacts, it does so without any real boundary nonetheless without being able to process the response as the response, on large-scale would contain millions of artifacts.

The limit is configurable for clients needs more than that.

in addition to that:
1. changed test s3:// to memory:// to speed up "trial-and-error" of s3 when trying to stat on file to find its size.
2. fixed logged headers to be json-friendly
3. added pipeline-adapters to gitignore as it is relevant for 1.7+ only (to easy on switching branches)

for the sake of developers, the new sql query, with the changes would look like:

```SQL
SELECT artifacts_v2.object, ...., anon_1.name 
FROM artifacts_v2
JOIN (SELECT artifacts_v2.id AS id, artifacts_v2_tags.name AS name 
        FROM artifacts_v2 
        JOIN artifacts_v2_tags
         ON artifacts_v2_tags.obj_id = artifacts_v2.id 
         WHERE artifacts_v2.project = 'load' 
         AND artifacts_v2.iteration = 1 
         AND (artifacts_v2.kind NOT IN ('dataset', 'model')) 
         AND artifacts_v2_tags.name = 'latest'
         LIMIT 1000
)
AS anon_1 ON artifacts_v2.id = anon_1.id
```

as opposed to what we had before:
```
SELECT artifact .... , .... , ....artifacts_v2_tags.name AS artifacts_v2_tags_name
FROM artifacts_v2
INNER JOIN artifacts_v2_tags ON artifacts_v2_tags.obj_id = artifacts_v2.id 
WHERE artifacts_v2_tags.name = 'latest'
 AND artifacts_v2.project = 'default' 
 AND artifacts_v2.iteration = 1 
 AND artifacts_v2.kind NOT IN ('dataset', 'model');
```

note that the limit is within the inner join, to ensure that the DB first loads the ids, returns the limit and only then we load the rest of the artifact fields. if we were applying the limit without having a nested query, the DB would first load the entire table and only then provide the limit.
tl;dr - this is required to ensure LIMIT is optimized

https://iguazio.atlassian.net/browse/ML-6867


server/api/db/sqldb/db.py
```diff
@@ -1321,8 +1316,31 @@ def _find_artifacts(
             query = self._add_artifact_category_query(category, query)
         if most_recent:
             query = self._attach_most_recent_artifact_query(session, query)
+        if limit:
+            query = query.limit(limit)
+
+        # compile the outer query which returns the tuple of artifact and its tag name
+        # it will join on the subquery matching the artifact IDs
+        # this way, out limit will take effect and optimized
```
TomerShor: ?
```suggestion
        # this way, the limit will take effect and optimized
```

tests/api/api/test_artifacts.py
```diff
@@ -264,6 +264,57 @@ def test_list_artifacts(db: Session, client: TestClient) -> None:
...
+
+    default_endpoint_limit = 2
+
+    # override client's default limit
+    for route in unversioned_client.app.routes:
+        if route.path.endswith(LIST_API_ARTIFACTS_V2_PATH):
+            for qp in route.dependant.query_params:
+                if qp.name == "limit":
+                    qp.default = default_endpoint_limit
+                    break
```
TomerShor: is this "sticky" for all tests?

tests/api/api/test_artifacts.py
```diff
@@ -264,6 +264,57 @@ def test_list_artifacts(db: Session, client: TestClient) -> None:
...
+            STORE_API_ARTIFACTS_V2_PATH.format(project=PROJECT),
+            json=data,
+        )
+        assert resp.status_code == HTTPStatus.CREATED.value
+
+    artifact_path = LIST_API_ARTIFACTS_V2_PATH.format(project=PROJECT)
+    resp = unversioned_client.get(f"{artifact_path}?limit={default_endpoint_limit-1}")
+    assert resp.status_code == HTTPStatus.OK.value
+    artifacts = resp.json()["artifacts"]
+    assert len(artifacts) == 1
```
TomerShor:
```suggestion
    assert len(artifacts) == default_endpoint_limit-1
```

TomerShor (commented):
Looks good, really minor comments

mlrun/db/httpdb.py
```diff
@@ -987,6 +988,7 @@ def list_artifacts(
         :param producer_uri:    Return artifacts produced by the requested producer URI. Producer URI usually
             points to a run and is used to filter artifacts by the run that produced them when the artifact producer id
             is a workflow id (artifact was created as part of a workflow).
+        :param limit:           Maximum number of artifacts to return. Default is 1000.
```
alonmr: IMO we should document that we do the tag filtering after the limit and you might get less than 1000 artifacts even if you have more with the specific tag.

alonmr (commented):
Minor documentation suggestion

mlrun/db/httpdb.py
TomerShor: Limit means Maximum - less is accepted 🤷‍♂️

tests/api/api/test_artifacts.py
liranbg: yeah, ill fix it

tests/api/api/test_artifacts.py
```diff
@@ -264,6 +264,60 @@ def test_list_artifacts(db: Session, client: TestClient) -> None:
...
+    artifacts = resp.json()["artifacts"]
+    assert len(artifacts) == default_endpoint_limit - 1
+
+    # Get all artifacts
+    resp = unversioned_client.get(artifact_path)
+    assert resp.status_code == HTTPStatus.OK.value
+    artifacts = resp.json()["artifacts"]
+    assert len(artifacts) == default_endpoint_limit
+
+    ensure_endpoint_limit(1000)
```
TomerShor: this won't run if an `assert` fails.
Use a fixture to achieve `defer`-like behavior

TomerShor (changes requested):
one last thing

mlrun/db/httpdb.py
alonmr: what? if you have 2000 artifacts and the 1st 1000 are artifacts with tag v1 and the next 1000 are with v2 and you list with v2 and limit 1000 you will get 0 artifacts. how is that accepted?

server/api/db/sqldb/db.py
```diff
@@ -1321,8 +1316,31 @@ def _find_artifacts(
             query = self._add_artifact_category_query(category, query)
         if most_recent:
             query = self._attach_most_recent_artifact_query(session, query)
+        if limit:
+            query = query.limit(limit)
+
+        # compile the outer query which returns the tuple of artifact and its tag name
+        # it will join on the subquery matching the artifact IDs
+        # this way, the limit will take effect and optimized
```
quaark: I feel this comment doesn't really explain what exactly we are doing and why we're doing it
```suggestion
        # compile the above query as a sub query only for filtering out the relevant ids so the 
        # limit operation which sadly needs to load all the results before limiting will work on a slimmer data set
        # and will be much more optimized. Then join the outer query on the subquery to select the correct 
        # columns of the table
```

quaark (changes requested):
Looks good!
I suggest a different comment for the explanation of the subquery

server/api/db/sqldb/db.py
```diff
@@ -1316,30 +1319,29 @@ def _find_artifacts(
...
-            outer_query = outer_query.outerjoin(
+            query = query.outerjoin(
                 ArtifactV2.Tag, ArtifactV2.Tag.obj_id == ArtifactV2.id
             )
 
+        if limit:
+            query = query.limit(limit)
+
+        # limit operation loads all the results before performing the actual limiting,
+        # therefor, we compile the above query as a sub query only for filtering out the relevant ids,
```
alonmr:
```suggestion
        # therefore, we compile the above query as a sub query only for filtering out the relevant ids,
```

alonmr (approved):
Found a typo, other than that LGTM

PR #5794, opened 2024-06-19
[Serving] Fix creation of function which name is close to 63 chars

yanburman:
K8S does not allow label values over 63 chars, so we split the name into 2 separate fields
https://iguazio.atlassian.net/browse/ML-6896

### DIFF
--- a/server/api/utils/singletons/k8s.py
+++ b/server/api/utils/singletons/k8s.py
@@ -552,7 +552,9 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
     ):
         namespace = self.resolve_namespace(namespace)
         have_confmap = False
-        label_name = mlrun_constants.MLRunInternalLabels.resource_name
+        label_resource_name = mlrun_constants.MLRunInternalLabels.resource_name
+        label_resource_type = mlrun_constants.MLRunInternalLabels.resource_type
+        resource_name = name
         full_name = f"{resource}-{name}"
 
         configmap_with_label = self.get_configmap(name, resource, namespace)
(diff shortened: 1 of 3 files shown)


server/api/utils/singletons/k8s.py
```diff
@@ -567,9 +569,10 @@ def ensure_configmap(
             )
 
         if labels is None:
-            labels = {label_name: full_name}
+            labels = {label_resource_name: resource_name, label_resource_type: resource}
         else:
-            labels[label_name] = full_name
+            labels[label_resource_name] = resource_name
+            labels[label_resource_type] = resource
```
TomerShor, line 575:
```python
if labels is None:
    labels = {}
labels[label_resource_name] = resource_name
labels[label_resource_type] = resource
```

server/api/utils/singletons/k8s.py
```diff
@@ -604,14 +607,15 @@ def ensure_configmap(
...
+        label_resource_type = mlrun_constants.MLRunInternalLabels.resource_type
         configmaps_with_label = self.v1api.list_namespaced_config_map(
-            namespace=namespace, label_selector=f"{label_name}={full_name}"
+            namespace=namespace,
+            label_selector=f"{label_resource_name}={name},{label_resource_type}={resource}",
         )
         if len(configmaps_with_label.items) > 1:
             raise mlrun.errors.MLRunInternalServerError(
-                f"Received more than one config map for label: {full_name}"
+                f"Received more than one config map for labels: {name}, {resource}"
```
TomerShor, line 618: this will result in the following string with only the values, and it is not clear what are the labels
```
Received more than one config map for labels: my-function, serving-conf
```

TomerShor (changes requested):
Looking good, added minor comments.

Also in `ensure_configmap`, it is a bit unclear what is the `resource` argrument, since it accepts a string `serving-conf`. A docstring will be helpful in this case.

PR #5796, opened 2024-06-19
[Alerts] Set criteria to defaults if not provided

yanburman:
https://iguazio.atlassian.net/issues/ML-6857

When user does not provide criteria, set it to defaults in order to be more explicit when user gets the object back

### DIFF
--- a/mlrun/common/schemas/alert.py
+++ b/mlrun/common/schemas/alert.py
@@ -104,7 +104,7 @@ class AlertCriteria(pydantic.BaseModel):
         pydantic.Field(
             description="Number of events to wait until notification is sent"
         ),
-    ] = 0
+    ] = 1
     period: Annotated[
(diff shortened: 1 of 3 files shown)


alonmr (commented):
Comment and qustion:
1. In `mlrun.common.schemas.alert.AlertCriteria` the default count is 0. might want to change that.
2. You also have count in `mlrun.common.schemas.alert.AlertConfig` whats the difference?

yanburman:
> Comment and qustion:
> 
> 1. In `mlrun.common.schemas.alert.AlertCriteria` the default count is 0. might want to change that.
Agree
> 2. You also have count in `mlrun.common.schemas.alert.AlertConfig` whats the difference?
That is count of how many times the alert fired

alonmr (approved):

PR #5797, opened 2024-06-19
[Scheduler] Enrich kind label

roei3000b:
If kind label doesn't exists, set it.

https://iguazio.atlassian.net/browse/ML-6695

### DIFF
--- a/server/api/utils/scheduler.py
+++ b/server/api/utils/scheduler.py
@@ -302,7 +302,7 @@ class Scheduler:
             self._remove_schedule_scheduler_resources(
                 db_session, schedule.project, schedule.name
             )
-        get_db().delete_schedules(db_session, project)
+        get_db().delete_project_schedules(db_session, project)
 
     @server.api.utils.helpers.ensure_running_on_chief
     def store_schedule(
--- a/server/api/db/base.py
+++ b/server/api/db/base.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import datetime
+import typing
 from abc import ABC, abstractmethod
 from typing import Any, Optional, Union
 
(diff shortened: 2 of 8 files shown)


moranbental (changes requested):
Maybe create unit tests to check that the kind is really enriched

server/api/utils/scheduler.py
```diff
@@ -347,9 +349,9 @@ def store_schedule(
 
         self._ensure_auth_info_has_access_key(auth_info, kind)
         secret_name = self._store_schedule_secrets_using_auth_secret(auth_info)
-        labels = self._append_access_key_secret_to_labels(labels, secret_name)
+        labels = self._enrich_labels(labels, secret_name)
+        labels.setdefault(mlrun.common.constants.MLRunInternalLabels.kind, kind)
```
rokatyy: why not like this? Do we need explicit replacement in that case?

```
labels = self._enrich_labels(labels, secret_name, kind)
```

roei3000b: OOPS, I forgot to delete this line.

rokatyy (commented):
I've found this test https://github.com/mlrun/mlrun/blob/109ca7d115b8790c165726fc1136e241e42774c7/tests/api/utils/test_scheduler.py#L1466, maybe just worth checking that it covers this part and we're done :)

alonmr (changes requested):
~You are confusing schedule kinds with runtime kinds 😅 . This is not the correct label value.~ - scratch that

alonmr (commented):
I'm not sure I'm following the fix. Please explain the issue and how this solves it. 
When the scheduler triggers the job, you want it to have the runtime kind label value not the schedule kind. So what is this for? I don't think you should use the labels to determine the schedule kind if you have it in the schedule object

roei3000b:
> I'm not sure I'm following the fix. Please explain the issue and how this solves it. When the scheduler triggers the job, you want it to have the runtime kind label value not the schedule kind. So what is this for? I don't think you should use the labels to determine the schedule kind if you have it in the schedule object

Oh, you right, this pr is wrong.
I will open another one.

PR #5802, opened 2024-06-19
[Formatters] Reintroduce old Schemas with Deprecation Warnings for 1.9.0

quaark:
Can't remove the schemas as was done without the proper deprecation procedure as these are client facing classes.


alonmr (approved):
thanks ❤️

alonmr:
We also need to make sure to update https://iguazio.atlassian.net/browse/ML-5948

PR #5803, opened 2024-06-19
[Linting] Check syntax and format code blocks in markdown with blacken-docs 

jond01:
https://github.com/adamchainz/blacken-docs

This change introduces the following benefits:

1. Ensure code blocks in the docs are coherent syntax-wise.
2. Format them according to Black conventions.

Add a new phony target, `make fmt-docs`, and enforce it as part of `make lint`, which runs in the CI workflow.

Some examples of wrong syntax in the docs before this PR:
1. Missing closing parenthesis `)`: 
   https://github.com/mlrun/mlrun/blob/f72280de3594af30e85ea6d344fc5c4a34b11ec8/docs/monitoring/model-monitoring.md?plain=1#L38-L45
2. Missing trailing comma `,` in L61 and L62:
   https://github.com/mlrun/mlrun/blob/f72280de3594af30e85ea6d344fc5c4a34b11ec8/docs/data-prep/logging_datasets.md?plain=1#L60-L64
3. Not a Python code block: 
   https://github.com/mlrun/mlrun/blob/f72280de3594af30e85ea6d344fc5c4a34b11ec8/docs/install/kubernetes.md?plain=1#L378-L380
4. Function body (`...` placeholder) unindented:
   https://github.com/mlrun/mlrun/blob/f72280de3594af30e85ea6d344fc5c4a34b11ec8/docs/concepts/decorators-and-auto-logging.md?plain=1#L80-L88

There are many more wrong syntax examples, but they are all fixed here.

Note: I previously configured Ruff to check and format code blocks inside Python/Jupyter Notebooks, such as code blocks in docstrings.
However, it's not designed to run on markdown files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -601,13 +601,18 @@ fmt: ## Format the code using Ruff
 	python -m ruff check --fix-only
 	python -m ruff format
 
+.PHONY: fmt-docs
+fmt-docs: ## Format the code blocks in markdown files
+	@echo "Formatting the docs"
+	git ls-files -z -- '*.md' | xargs -0 blacken-docs -t=py39
+
 .PHONY: lint-imports
 lint-imports: ## Validates import dependencies
 	@echo "Running import linter"
 	lint-imports
 
 .PHONY: lint
-lint: lint-check lint-imports ## Run lint on the code
+lint: lint-check lint-imports fmt-docs ## Run lint on the code
 
 .PHONY: lint-check
 lint-check: ## Check the code (using ruff)
--- a/Makefile
+++ b/Makefile
@@ -596,14 +596,17 @@ html-docs-dockerized: build-test ## Build html docs dockerized
 		make html-docs
 
 .PHONY: fmt
-fmt: ## Format the code using Ruff
+fmt: ## Format the code using Ruff and blacken-docs
 	@echo "Running ruff checks and fixes..."
 	python -m ruff check --fix-only
 	python -m ruff format
+	@echo "Formatting the code blocks with blacken-docs..."
+	git ls-files -z -- '*.md' | xargs -0 blacken-docs -t=py39
 
-.PHONY: fmt-docs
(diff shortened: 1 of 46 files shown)


Makefile
```diff
@@ -601,13 +601,18 @@ fmt: ## Format the code using Ruff
...
+	git ls-files -z -- '*.md' | xargs -0 blacken-docs -t=py39
+
 .PHONY: lint-imports
 lint-imports: ## Validates import dependencies
 	@echo "Running import linter"
 	lint-imports
 
 .PHONY: lint
-lint: lint-check lint-imports ## Run lint on the code
+lint: lint-check lint-imports fmt-docs ## Run lint on the code
```
moranbental: you should run `fmt-docs` in `make fmt` and not in `make lint`
and in `make lint` you should lint the code with `blacken-docs` using the `--check` flag

moranbental (changes requested):
It looks very good, just one comment

Makefile
jond01: > you should run `fmt-docs` in `make fmt` and not in `make lint`

I thought it's better to separate `fmt-docs` from `fmt` that runs on actual code, but I'll change it :)

> and in `make lint` you should lint the code with `blacken-docs` using the `--check` flag

I wanted to use it - but the package is yet to be released with this relatively new flag.
I opened https://github.com/adamchainz/blacken-docs/issues/346 to request a new release.

moranbental: you can separate `make fmt-docs` from `make fmt` but I don't think it should be in `make lint`

jond01: I changed it now to be ready for the next release, but I kept it in `make lint` as only that is enforced in the CI

moranbental (approved):

jillnogold (approved):

assaf758 (approved):

PR #5805, opened 2024-06-19
[API] Enrich job kind during submit job

roei3000b:
During submit job there could be use cases where the kind label of the job is dismissed.
In this PR we enrich it from the function we resolved.

https://iguazio.atlassian.net/browse/ML-6695


server/api/api/utils.py
```diff
@@ -211,9 +212,12 @@ async def submit_run(
     return response
 
 
-def apply_enrichment_and_validation_on_task(task):
+def apply_enrichment_and_validation_on_task(task, kind):
     # Conceal notification config params from the task object with secrets
     mask_notification_params_on_task(task, server.api.constants.MaskOperations.CONCEAL)
+    task.setdefault("metadata", {}).setdefault("labels", {}).setdefault(
+        mlrun_constants.MLRunInternalLabels.kind, kind
+    )
```
alonmr: Enriching the label here is making the task enrichment even more sparse than it already is. Also this enriches non scheduled jobs that I don't think we want to. We should have a consolidated place that enriches schedules. It should be in `server.api.utils.scheduler.Scheduler` and all of the common enrichments for store/create/update should be there.

alonmr (commented):
Looks much better! but I do think we should move the enrichment. See comment below.

server/api/utils/scheduler.py
```diff
@@ -94,7 +94,6 @@ def _append_access_key_secret_to_labels(self, labels, secret_name):
         if secret_name:
             labels = labels or {}
             labels[self._db_record_auth_label] = secret_name
-        return labels
```
alonmr: if labels is `None` this will not work. the labels attr in the inner function scope will point to a new dict but the outer function's labels will still be `None`

server/api/utils/scheduler.py
```diff
@@ -994,6 +981,16 @@ def _enrich_schedule_notifications(
                 )
             ]
 
+    def _enrich(self, auth_info, kind, labels, name, project, scheduled_object, fn_kind=None):
```
alonmr:
```suggestion
    def _enrich_schedule(self, auth_info, kind, labels, name, project, scheduled_object, fn_kind=None):
```

alonmr (commented):
Very well! minor comment

liranbg (approved):

PR #5807, opened 2024-06-20
[Feature Store] Add v2 of `list_entities` and `list_features` SDK APIs

gtopper:
Building on #5729 and #5751 and decoupling the v2 implementation from the v1 implementation.

[ML-6870](https://iguazio.atlassian.net/browse/ML-6870)

[ML-6870]: https://iguazio.atlassian.net/browse/ML-6870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/api/api/endpoints/feature_store_v2.py
```diff
@@ -67,39 +66,21 @@ async def list_entities(
     auth_info: mlrun.common.schemas.AuthInfo = Depends(deps.authenticate_request),
     db_session: Session = Depends(deps.get_db_session),
 ):
-    entities = await server.api.api.endpoints.feature_store.list_entities(
+    await server.api.utils.auth.verifier.AuthVerifier().query_project_permissions(
+        project,
+        mlrun.common.schemas.AuthorizationAction.read,
+        auth_info,
+    )
```
theSaarco: Since we've removed the fine-grained check later, this check is now not good enough. `query_project_permissions` will only check if the user has general read permissions on the project itself, not on its objects. For example, the project-reader user which has the project-read-only policy can list all projects, but cannot see their objects. This check will allow that user to list entities.
This needs to change to `query_project_resource_permissions` with the `mlrun.common.schemas.AuthorizationResourceTypes.feature_set` resource type (since we don't have specific resource types for entities/features).

server/api/api/endpoints/feature_store_v2.py
```diff
@@ -112,37 +93,19 @@ async def list_features(
...
-        name=name,
-        tag=tag,
-        entities=entities,
-        labels=labels,
-        auth_info=auth_info,
-        db_session=db_session,
+    await server.api.utils.auth.verifier.AuthVerifier().query_project_permissions(
+        project,
+        mlrun.common.schemas.AuthorizationAction.read,
+        auth_info,
```
theSaarco: Same here.

server/api/crud/feature_store.py
```diff
@@ -164,6 +165,12 @@ def delete_feature_set(
             uid,
         )
 
+    # TODO: remove in 1.9.0
+    @deprecated(
+        version="1.9.0",
+        reason="'list_features' will be removed in 1.9.0, use 'list_entities_v2' instead",
+        category=FutureWarning,
+    )
```
theSaarco: Not really needed, since the CRUD is internal and is not called by the user directly. Maybe just keep the `TODO` to remind us.

server/api/db/sqldb/db.py
```diff
@@ -3170,6 +3176,102 @@ def list_features(
...
+                # Get the feature from the feature-set full structure, as it may contain extra fields (which are not
+                # in the DB)
+                feature = next(
+                    (
+                        feature
+                        for feature in feature_set.spec.features
+                        if feature.name == feature_name
+                    ),
+                    None,
+                )
```
theSaarco: I'm thinking this block here can also gain from using memoization, by checking the `feature_set.spec.features` and the feature-name as params, saving us repeated iteration on features. WDYT?

theSaarco (changes requested):
Looks pretty good. Need a small fix in the permissions checking, and I'm thinking we can maybe optimize a bit more.

mlrun/db/base.py
```diff
@@ -329,6 +337,23 @@ def list_features(
...
 
+    @abstractmethod
+    def list_features_v2(
+        self,
+        project: str,
+        name: str = None,
+        tag: str = None,
+        entities: list[str] = None,
+        labels: list[str] = None,
+    ) -> mlrun.common.schemas.FeaturesOutput:
```
theSaarco:
```suggestion
    ) -> mlrun.common.schemas.FeaturesOutputV2:
```

theSaarco (approved):
Looks very good. Minor typo to fix.

PR #5823, opened 2024-06-20
[Serving] Fix env config map generation and other veggies

alonmr:
1. Config map resource name label value cannot be longer than 64 chars so use the function name without prepending the serving_conf prefix - fixes `TestNuclioRuntime::test_workaround_for_nuclio_bug`.
2. Use `conda-libmamba-solver` in ml-base for faster dependencies solving (like in mlrun image).
3. Set default project if not specified for artifact endpoints in httpdb.
4. `with_repo` docstring clarification asked by @yonishelach and @ZeevRispler


dockerfiles/base/Dockerfile
```diff
@@ -61,9 +61,13 @@ RUN wget --quiet https://repo.anaconda.com/miniconda/Miniconda3${MLRUN_ANACONDA_
 
 ENV PATH /opt/conda/bin:$PATH
 
+RUN conda install -y -n base conda-libmamba-solver && \
+    conda config --set solver libmamba && \
+    conda clean -aqy
+
```
liranbg, line 67: Can we update miniconda for python 3.9? to get latest python 3.9 version

server/api/api/utils.py
```diff
@@ -32,6 +32,7 @@
 from fastapi.concurrency import run_in_threadpool
 from sqlalchemy.orm import Session
 
+import mlrun.common.constants
```
TomerShor: not needed if you removed the constant

server/api/utils/singletons/k8s.py
```diff
@@ -553,24 +553,23 @@ def ensure_configmap(
         namespace = self.resolve_namespace(namespace)
         have_confmap = False
         label_name = mlrun_constants.MLRunInternalLabels.resource_name
-        full_name = f"{resource}-{name}"
+        if labels is None:
+            labels = {label_name: name}
+        else:
+            labels[label_name] = name
```
TomerShor:
```suggestion
        if labels is None:
            labels = {}  
        labels[label_name] = name
```

TomerShor (changes requested):

dockerfiles/base/Dockerfile
alonmr: The base image is not latest python 3.9 either. Its python 3.9.7 (iguazio python slim). May need to upgrade that 1st

server/api/utils/singletons/k8s.py
```diff
@@ -553,24 +553,23 @@ def ensure_configmap(
...
 
-        configmap_with_label = self.get_configmap(name, resource, namespace)
+        configmap_with_label = self.get_configmap(name, namespace)
```
rokatyy: All this naming of `name` variable is a bit confusing, so what I'd suggest is having `resource_name` and `configmap_name` to reduce this confusion.

yanburman (approved):

liranbg (approved):

TomerShor (approved):

PR #5825, opened 2024-06-22
[Setup] Move packaging configuration to `pyproject.toml` [1.8.x]

alonmr:
## Updating MLRun SDK build process
As `setuptools` is transitioning to a static based project configuration - https://setuptools.pypa.io/en/latest/userguide/quickstart.html#transitioning-from-setup-py-to-declarative-config 
Though the main reason we need this change is this unlocks using tools that do not support `setup.py` based building.
In particular see - [Rye](https://rye.astral.sh/guide/) which syncs and locks requirements, leveraging `uv` as its installation engine which is written in Rust.
This will solve a serious pain point for us which is release of mlrun dependencies which breaks functionality out of no-where.
Either way, since `setup.py` is being deprecated there is no reason to stay behind.

> Note: The reason for the `setup.cfg` file is that `setuptools` is not 100% compatible with `pyproject.toml` yet so I had to define some specific metadata in there.

Things I need to do to complete the pull request:
- [ ] Update the requirements tests
- [ ] Remove requirements.txt files


pyproject.toml
```diff
@@ -1,3 +1,204 @@
...
+    "igz-mgmt~=0.2.0",
+    "humanfriendly~=10.0",
+    "fastapi~=0.110.0",
+    # in sqlalchemy>=2.0 there is breaking changes (such as in Table class autoload argument is removed)
+    "sqlalchemy~=1.4",
+    "pymysql~=1.0",
+    "alembic~=1.9",
+    "timelength~=1.1",
+    "memray~=1.12",
+]
```
liranbg, line 170: isnt better to have `pyproject.toml` for the api for itself?

alonmr:
according to https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata there is an option to specify dependancies from file but it is still in beta

PR #5827, opened 2024-06-23
[Model Monitoring] Enforce user to define credentials

davesh0812:
General notes:

1. Users must define all three mandatory credentials for each project before using any MM functionality, including deploying serving with tracking. These credentials can't be changed once set.
2. Users have two options for setting those creds:
  - Use the set_model_monitoring_credentials API/SDK.
  - Pass fetch_credentials_from_sys_config=True with the enable_model_monitoring API/SDK, setting the system config via MapConfig (no default configuration).
3. For each project, credentials will be saved to the project secret, even when using system configuration, to prevent data loss if the system config changes.
4. For upgraded projects that had MM, we will set any missing credentials to v3io.

regarding the system tests I added a model monitoring marker in order to check that the cred are set to the env.

https://iguazio.atlassian.net/browse/ML-6778


mlrun/common/schemas/model_monitoring/constants.py
```diff
@@ -158,19 +158,38 @@ class EventKeyMetrics:
...
+
+
 class ProjectSecretKeys:
     ENDPOINT_STORE_CONNECTION = "MODEL_MONITORING_ENDPOINT_STORE_CONNECTION"
-    ACCESS_KEY = "MODEL_MONITORING_ACCESS_KEY"
-    PIPELINES_ACCESS_KEY = "MODEL_MONITORING_PIPELINES_ACCESS_KEY"
-    KAFKA_BROKERS = "KAFKA_BROKERS"
+    ACCESS_KEY = "MODEL_MONITORING_ACCESS_KEY"  # ?
+    PIPELINES_ACCESS_KEY = "MODEL_MONITORING_PIPELINES_ACCESS_KEY"  # ?
+    KAFKA_BROKERS = "KAFKA_BROKERS"  # ?
```
Eyal-Danieli: not in use
```suggestion
```

mlrun/config.py
```diff
@@ -523,11 +523,11 @@
         "parquet_batching_max_events": 10_000,
         "parquet_batching_timeout_secs": timedelta(minutes=1).total_seconds(),
         # See mlrun.model_monitoring.db.stores.ObjectStoreFactory for available options
-        "store_type": "v3io-nosql",
         "endpoint_store_connection": "",
         # See mlrun.model_monitoring.db.tsdb.ObjectTSDBFactory for available options
-        "tsdb_connector_type": "v3io-tsdb",
         "tsdb_connection": "",
+        # See mlrun.common.schemas.model_monitoring.constants.ModelMonitoringStreamKind for available options
```
Eyal-Danieli:
```suggestion
        # See mlrun.common.schemas.model_monitoring.constants.StreamKind for available options
```

mlrun/config.py
```diff
@@ -1112,13 +1112,6 @@ def get_model_monitoring_file_target_path(
         """
 
         if target != "offline":
-            store_prefix_dict = (
-                mlrun.mlconf.model_endpoint_monitoring.store_prefixes.to_dict()
-            )
-            if store_prefix_dict.get(kind):
-                # Target exist in store prefix and has a valid string value
-                return store_prefix_dict[kind].format(project=project, **kwargs)
```
Eyal-Danieli, line 1122: remove the `**kwrags` from the func signature

server/api/api/endpoints/model_monitoring.py
```diff
@@ -283,3 +287,53 @@ async def delete_model_monitoring_function(
...
     return tasks
+
+
+@router.post("/set-model-monitoring-credentials")
+def set_model_monitoring_credentials(
+    commons: Annotated[_CommonParams, Depends(_common_parameters)],
+    access_key: Optional[str] = None,
+    endpoint_store_connection: Optional[str] = None,  # v3io-nosql, sql, system-default
+    stream_path: Optional[str] = None,  # v3io-stream, kafka, system-default
+    tsdb_connection: Optional[str] = None,  # v3io-tsdb, TDEngine, system-default
```
Eyal-Danieli: remove the comments

server/api/api/endpoints/model_monitoring.py
```diff
@@ -283,3 +287,53 @@ async def delete_model_monitoring_function(
...
+                                      3. MySQL/SQLite - for SQL endpoint store, please provide full
+                                         connection string, for example
+                                         mysql+pymysql://<username>:<password>@<host>:<port>/<db_name>
+    :param stream_path:               Path to the model monitoring stream. By default, None.
+                                      Options:
+                                      1. None, will be set from the system configuration.
+                                      2. v3io - for v3io stream,
+                                         pass `v3io` and the system will generate the exact path.
+                                      3. Kafka - for Kafka stream, please provide full connection string without
+                                         costume topic, for example kafka://<some_kafka_broker>:<port>.
```
Eyal-Danieli:
```suggestion
                                         custom topic, for example kafka://<some_kafka_broker>:<port>.
```

server/api/crud/model_monitoring/deployment.py
```diff
@@ -875,6 +882,212 @@ async def delete_monitoring_function(
...
+            "or pass fetch_credentials_from_sys_config=True when using enable_model_monitoring API/SDK."
+        )
+
+    def set_credentials(
+        self,
+        access_key: typing.Optional[str] = None,
+        endpoint_store_connection: typing.Optional[str] = None,
+        stream_path: typing.Optional[str] = None,
+        tsdb_connection: typing.Optional[str] = None,
+        default_secrets: typing.Optional[str] = "",  # used only for upgrade case
```
Eyal-Danieli: as for the default_secrets, I recommend to convert it to a `dict` that represents secret key-value. Then, it actually supports multiple optional keys.

mlrun/model_monitoring/db/stores/__init__.py
```diff
@@ -111,10 +111,12 @@ def get_store_object(
...
         kwargs["store_connection_string"] = store_connection_string
+    elif store_connection_string and store_connection_string == "v3io":
+        store_type = (
+            mlrun.common.schemas.model_monitoring.ModelEndpointTarget.V3IO_NOSQL
+        )
     else:
-        # Set the default store type if no connection has been set
-        store_type = mlrun.mlconf.model_endpoint_monitoring.store_type
-
+        store_type = None
```
Eyal-Danieli, line 119: In that case the user will probably get an error when calling any of the model monitoring APIs. The issue is that it can breaks some APIs in the UI in which instead of returning "nothing", the user will get an error (e.g. model endpoints tab before setting mm creds).
We need to handle this issue by catching it on the BE side and return empty values without going into this flow.

mlrun/model_monitoring/db/tsdb/__init__.py
```diff
@@ -86,12 +86,10 @@ def get_tsdb_connector(
...
-
-    # Set the default TSDB connector type if no connection has been set
-    tsdb_connector_type = (
-        tsdb_connector_type
-        or mlrun.mlconf.model_endpoint_monitoring.tsdb_connector_type
-    )
+    elif tsdb_connection_string and tsdb_connection_string == "v3io":
+        tsdb_connector_type = mlrun.common.schemas.model_monitoring.TSDBTarget.V3IO_TSDB
+    else:
+        tsdb_connector_type = None
```
Eyal-Danieli, line 91: remove `tsdb_connector_type` from func signature

mlrun/model_monitoring/db/stores/__init__.py
Eyal-Danieli: never mind I think you already handled it in https://github.com/mlrun/mlrun/pull/5827/files#:~:text=endpoints%20%3D%20schemas.ModelEndpointList(endpoints%3D%5B%5D) , just need to make sure it works without errors

server/api/crud/client_spec.py
```diff
@@ -107,15 +107,16 @@ def get_client_spec(
...
             model_endpoint_monitoring_endpoint_store_connection=self._get_config_value_if_not_default(
                 "model_endpoint_monitoring.endpoint_store_connection"
             ),
             model_monitoring_tsdb_connection=self._get_config_value_if_not_default(
                 "model_endpoint_monitoring.tsdb_connection"
             ),
+            model_monitoring_stream_connection=self._get_config_value_if_not_default(
+                "model_endpoint_monitoring.stream_connection"
+            ),
+            # TODO : ask eyal about this
```
Eyal-Danieli: ;)
```suggestion
```

tests/system/model_monitoring/test_app.py
```diff
@@ -396,6 +396,9 @@ def custom_setup(cls) -> None:
     def _submit_controller_and_deploy_writer(
         self, deploy_histogram_data_drift_app
     ) -> None:
+        self.project.set_model_monitoring_credentials(
+            endpoint_store_connection="v3io", stream_path="v3io", tsdb_connection="v3io"
+        )
```
Eyal-Danieli: Note that this test should pass with other targets such as TDEngine. I guess it will be break with this change.

tests/system/model_monitoring/test_model_monitoring.py
```diff
@@ -283,6 +283,9 @@ def test_basic_model_monitoring(self, with_sql_target: bool) -> None:
         ).apply(mlrun_pipelines.mounts.auto_mount())
         # enable model monitoring
         serving_fn.set_tracking()
+        project.set_model_monitoring_credentials(
```
Eyal-Danieli: add support for mysql as well

Eyal-Danieli (commented):
nice!
several comments

mlrun/model_monitoring/db/stores/__init__.py
davesh0812: so it stays as it is :)

server/api/crud/model_monitoring/deployment.py
```diff
@@ -875,6 +882,213 @@ async def delete_monitoring_function(
...
+        :param tsdb_connection:           Connection string to the time series database. By default, None.
+                                          Options:
+                                          1. None, will be set from the system configuration.
+                                          2. v3io - for v3io stream,
+                                             pass `v3io` and the system will generate the exact path.
+                                          3. TDEngine - for TDEngine tsdb, please provide full websocket connection URL,
+                                             for example taosws://<username>:<password>@<host>:<port>.
+        :param _default_secrets_v3io:     Optional, used only for upgrade case, to set the all the undefined secrets
+                                          to v3io.
+        """
```
```suggestion
        :param _default_secrets_v3io:     Optional parameter for the upgrade process in which the v3io default secret key is set.
        """
```

mlrun/config.py
```diff
@@ -523,11 +523,11 @@
         "parquet_batching_max_events": 10_000,
         "parquet_batching_timeout_secs": timedelta(minutes=1).total_seconds(),
         # See mlrun.model_monitoring.db.stores.ObjectStoreFactory for available options
-        "store_type": "v3io-nosql",
```
Eyal-Danieli, line 526: following our discussion with Saar, maybe leave it for now and add a deprecated note as you did for the `stream` value, at least until we can confirm that it doesn't break anything such as installing CE

Eyal-Danieli (approved):
cool cool cool

PR #5829, opened 2024-06-23
[Alerts] Add criteria field to alert config serialization

Yacouby:
https://iguazio.atlassian.net/browse/ML-6992


alonmr (approved):
Now that I understand better why you need this serialization I think there is a different issue here. Using pydantic model in composition with model obj is not the mlrun way to define a model obj. The model obj attributes should have `to_dict` method. One way to conform is as was done with API Gateway with a dedicated `to_shcema`. I'll not block this for now but we should discuss a convention. Also please add a jira ticket.

PR #5831, opened 2024-06-23
[Alerts] Lower DB access by caching events for inactive alerts

yanburman:
We want to lower performance footprint for alerts that are inactive. This comes with a cost of not persisting state between mlrun-api resets, but that is something we are OK with at this point


liranbg (approved):
Approving. only concern is that its limit is subjected to the fact that if event has configuration of great counter, the `states` dictionary would become gigantic consuming memory.
@theSaarco fyi

PR #5836, opened 2024-06-24
[API] Fix creating pipelines

liranbg:
Fixing a regression created by https://github.com/mlrun/mlrun/pull/5731/files#diff-e8344a65bd3093b98d6cce47bf2ccc79653b1afc64141e7b4ea9076a7b399d62L290-L296

https://iguazio.atlassian.net/browse/ML-6928

### DIFF
--- a/server/api/api/endpoints/pipelines.py
+++ b/server/api/api/endpoints/pipelines.py
@@ -108,15 +108,12 @@ async def list_pipelines(
 async def create_pipeline(
     project: str,
     request: Request,
-    namespace: str = None,
     experiment_name: str = Query("Default", alias="experiment"),
     run_name: str = Query("", alias="run"),
     auth_info: mlrun.common.schemas.AuthInfo = Depends(deps.authenticate_request),
 ):
-    if namespace is None:
-        namespace = config.namespace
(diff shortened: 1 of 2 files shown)


liranbg:
BC test failed because I have removed the query-param "namespace" which wasnt in use really (the param does not work for creating pipeline) as KFP would fail.

server/api/api/endpoints/pipelines.py
```diff
@@ -108,15 +108,12 @@ async def list_pipelines(
...
-    namespace: str = None,
     experiment_name: str = Query("Default", alias="experiment"),
     run_name: str = Query("", alias="run"),
     auth_info: mlrun.common.schemas.AuthInfo = Depends(deps.authenticate_request),
 ):
-    if namespace is None:
-        namespace = config.namespace
     response = await _create_pipeline(
-        auth_info, request, namespace, experiment_name, run_name, project
+        auth_info, request, experiment_name, run_name, project
```
TomerShor: call them with kw args just to make sure
```suggestion
        auth_info=auth_info, 
        request=request,
        ...
```

TomerShor (changes requested):

PR #5842, opened 2024-06-25
[Artifact] Verify that secrets passed by users are not internal when deleting artifact data

moranbental:
Verify that, similar project secrets, secrets passed by users are not internal.
Not allowed to create/update internal secrets (key starts with `mlrun.`)

Internal secrets are internal for the use of MLRun only. If the user may pass in any arbitrary secret to override its value, we allow the user to interfere with how MLRun manages its secrets internally, which can lead to unexpected results.

https://iguazio.atlassian.net/browse/ML-6957

### DIFF
--- a/server/api/crud/files.py
+++ b/server/api/crud/files.py
@@ -48,6 +48,13 @@ class Files(
         secrets: dict = None,
     ):
         secrets = secrets or {}
+
+        # Verify that, similar project secrets, secrets passed by users are not internal.
+        for secret_key in secrets.keys():
+            server.api.crud.Secrets().validate_internal_project_secret_key_allowed(
+                secret_key
+            )
+
         project_secrets = self._verify_and_get_project_secrets(project)
         project_secrets.update(secrets)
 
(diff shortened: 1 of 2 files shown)


server/api/crud/files.py
```diff
@@ -48,6 +48,13 @@ def delete_artifact_data(
         secrets: dict = None,
     ):
         secrets = secrets or {}
+
+        # Verify that, similar project secrets, secrets passed by users are not internal.
```
liranbg, line 52: may you update the comment? it is unclear why it is not allowed

liranbg (changes requested):
up

server/api/crud/files.py
moranbental: @theSaarco why are we not permitting internal secrets when deleting artifact data?

theSaarco: Internal secrets are, as their name implies, internal for the use of MLRun only. If the user can pass here any arbitrary secret to override its value, then we allow the user to interfere with how MLRun manages its secrets internally and can cause unexpected results.
Even if at this flow all we do is go to the storage and delete a file, we should still block that option.

PR #5843, opened 2024-06-25
[Datastore] Fix snowflake get offline features

tomerm-iguazio:
1) Added target.source_attributes for spark_merger usage.
can be used for transform essential attributes from target to source.
2) Added `get_offline_features` to the test, remove `test_source`.
3) Added passthrough=True parameter to the test.
4) Fix `SparkHadoopTestBase` bug - `configure_image_deployment` should build the default `RemoteSparkRuntime` image.

[ML-6301](https://iguazio.atlassian.net/browse/ML-6301)

[ML-6301]: https://iguazio.atlassian.net/browse/ML-6301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/system/feature_store/test_spark_snowflake.py
```diff
@@ -95,6 +94,10 @@ def setup_method(self, method):
         self.current_time = datetime.now().strftime("%Y%m%d_%H%M%S_%f")
         self.source_table = f"source_{self.current_time}"
         self.tables_to_drop = [self.source_table]
+        if self.deployment_type == Deployment.Remote:
+            self.project.set_secrets(
+                {"SNOWFLAKE_PASSWORD": os.environ.get("SNOWFLAKE_PASSWORD")}
```
gtopper:
```suggestion
                {"SNOWFLAKE_PASSWORD": os.environ["SNOWFLAKE_PASSWORD"]}
```

tests/system/feature_store/test_spark_snowflake.py
```diff
@@ -160,40 +165,36 @@ def test_snowflake_source_and_target(self):
...
-            "snowflake_source",
-            query=f"select * from {self.source_table} order by ID limit {number_of_rows}",
-            schema=self.schema,
-            time_field="LICENSE_DATE",
-            **self.snowflake_spark_parameters,
+        run_config = (
+            fstore.RunConfig(local=self.run_local)
+            if self.run_local
+            else fstore.RunConfig(local=self.run_local, kind="remote-spark")
         )
```
gtopper:
```suggestion
        run_config = fstore.RunConfig(
            local=self.run_local, kind=None if self.run_local else "remote-spark"
        )
```

mlrun/datastore/targets.py
```diff
@@ -1259,6 +1262,17 @@ def as_df(
     ):
         raise NotImplementedError()
 
+    @property
+    def source_attributes(self) -> dict:
+        keys = ["url", "user", "database", "db_schema", "warehouse"]
+        if self.attributes:
+            snowflake_dict = {key: self.attributes.get(key) for key in keys}
+        else:
+            snowflake_dict = {key: None for key in keys}
```
gtopper: Redundant safeguard:
```suggestion
        snowflake_dict = {key: self.attributes.get(key) for key in keys}
```

mlrun/feature_store/retrieval/spark_merger.py
```diff
@@ -252,7 +253,7 @@ def _get_engine_df(
                 )
             source_kind = target.kind
             source_path = target.get_target_path()
-
+            source_kwargs = target.source_attributes
```
gtopper: Not sure what to suggest here, but it's a bit confusing that something generic-sounding like `source_attributes` is only used in Spark.

mlrun/datastore/targets.py
```diff
@@ -1208,19 +1212,18 @@ def __init__(
...
+        if url:
+            attrs = {
+                "url": url,
+                "user": user,
+                "database": database,
+                "db_schema": db_schema,
+                "warehouse": warehouse,
+                "table": table_name,
+            }
+            attributes.update(attrs)
```
gtopper: Hm, is this correct? If `url` is set we override all these attrs, otherwise we ignore the parameters? Wouldn't this make more sense?
```suggestion
        attributes = attributes or {}
        if url:
            attributes["url"] = url
        if user:
            attributes["user"] = user
        if database:
            attributes["database"] = database
        if db_schema:
            attributes["db_schema"] = db_schema
        if warehouse:
            attributes["warehouse"] = warehouse
        if table_name:
            attributes["table"] = table_name
```

mlrun/datastore/targets.py
tomerm-iguazio: self.attributes can be None if it is built from dict... lets use the condition

mlrun/feature_store/retrieval/spark_merger.py
tomerm-iguazio: rename to `source_spark_attributes`

mlrun/datastore/targets.py
gtopper: It can't.

gtopper: Ok, so
```suggestion
attributes = self.attributes or {}
snowflake_dict = {key: attributes.get(key) for key in keys}
```

mlrun/datastore/targets.py
```diff
@@ -1259,6 +1264,15 @@ def as_df(
     ):
         raise NotImplementedError()
 
+    @property
+    def source_spark_attributes(self) -> dict:
+        keys = ["url", "user", "database", "db_schema", "warehouse"]
+        attributes = self.attributes or {}
+        snowflake_dict = {key: attributes.get(key) for key in keys}
+        table = self.attributes.get("table")
```
gtopper: If `self.attributes` can really be `None` like you say, this can fail and should be changed to
```suggestion
        table = attributes.get("table")
```

gtopper (approved):

PR #5845, opened 2024-06-25
[Model Monitoring] Fix filter `list_model_monitoring` over function and labels

davesh0812:
https://iguazio.atlassian.net/browse/ML-6825

### DIFF
--- a/mlrun/model_monitoring/db/stores/sqldb/sql_store.py
+++ b/mlrun/model_monitoring/db/stores/sqldb/sql_store.py
@@ -303,6 +303,7 @@ class SQLStoreBase(StoreBase):
 
             # Apply filters
             if model:
+                model = model if ":" in model else f"{model}:latest"
                 query = self._filter_values(
                     query=query,
                     model_endpoints_table=model_endpoints_table,
@@ -310,11 +311,12 @@ class SQLStoreBase(StoreBase):
                     filtered_values=[model],
                 )
             if function:
+                function_uri = f"{self.project}/{function}"
                 query = self._filter_values(
                     query=query,
                     model_endpoints_table=model_endpoints_table,
-                    key_filter=mm_schemas.EventFieldType.FUNCTION,
-                    filtered_values=[function],
+                    key_filter=mm_schemas.EventFieldType.FUNCTION_URI,
+                    filtered_values=[function_uri],
                 )
             if uids:
                 query = self._filter_values(
--- a/mlrun/model_monitoring/db/stores/sqldb/sql_store.py
+++ b/mlrun/model_monitoring/db/stores/sqldb/sql_store.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import datetime
-import json
 import typing
 import uuid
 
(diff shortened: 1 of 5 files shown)


mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -285,7 +285,23 @@ def list_model_endpoints(
...
+                label_in = True
+                endpoint_labels = json.loads(endpoint["labels"])
+                for label in labels:
+                    if "=" in label:
+                        label, value = list(map(lambda x: x.strip(), label.split("=")))
+                    else:
+                        value = None
+                    label_in = label in endpoint_labels and (
+                        (value and str(endpoint_labels[label]) == value) or not value
+                    )
```
gtopper: I think this can be simplified a bit
```suggestion
                    label_in = label in endpoint_labels and (
                        not value or str(endpoint_labels[label]) == value
                    )
```

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -285,7 +285,23 @@ def list_model_endpoints(
...
+                    if not label_in:
+                        break
+                if label_in:
+                    endpoint_list.append(endpoint)
+            else:
+                endpoint_list.append(endpoint)
```
gtopper: Can simplify this a bit (and move the definition of `label_in` one line up, before the if condition):
```suggestion
            if label_in:
                endpoint_list.append(endpoint)
```

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -518,10 +533,7 @@ def _build_kv_cursor_filter_expression(
 
         :param project:    The name of the project.
         :param model:      The name of the model to filter by.
-        :param function:   The name of the function to filter by.
-        :param labels:     A list of labels to filter by. Label filters work by either filtering a specific value of
-                           a label (i.e. list("key=value")) or by looking for the existence of a given
-                           key (i.e. "key").
+        :param function_uri:   The uri of the function to filter by.
```
gtopper: Need to realign the parameters.

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -285,7 +285,23 @@ def list_model_endpoints(
             endpoint = self.get_model_endpoint(
                 endpoint_id=endpoint_id,
             )
-            endpoint_list.append(endpoint)
+            if labels:
+                label_in = True
+                endpoint_labels = json.loads(endpoint["labels"])
+                for label in labels:
+                    if "=" in label:
+                        label, value = list(map(lambda x: x.strip(), label.split("=")))
```
gtopper:
```suggestion
                        label, value = list(map(lambda x: x.strip(), label.split("=", 1)))
```

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -303,18 +303,20 @@ def list_model_endpoints(
...
             if model:
+                model = model if ":" in model else f"{model}:latest"
                 query = self._filter_values(
                     query=query,
                     model_endpoints_table=model_endpoints_table,
                     key_filter=mm_schemas.EventFieldType.MODEL,
                     filtered_values=[model],
                 )
             if function:
+                function_uri = f"{self.project}/{function}"
```
gtopper, line 313: This is not a URI though.

davesh0812: True' but in MEP store it looks like this :)

gtopper: How about calling it something else though?

gtopper: Ah, I see that it's already defined as a URI in MM constants 🤦‍♂️

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -507,22 +524,18 @@ def _get_frames_client(self):
...
-        :param project:    The name of the project.
-        :param model:      The name of the model to filter by.
-        :param function:   The name of the function to filter by.
-        :param labels:     A list of labels to filter by. Label filters work by either filtering a specific value of
-                           a label (i.e. list("key=value")) or by looking for the existence of a given
-                           key (i.e. "key").
-        :param top_level:  If True will return only routers and endpoint that are NOT children of any router.
+        :param project:         The name of the project.
+        :param model:           The name of the model to filter by.
+        :param function_uri:    The uri of the function to filter by (<project_name>/<function_name>).
```
gtopper: At least don't describe it as a URI in the docstring :)

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -285,7 +285,24 @@ def list_model_endpoints(
...
+                    else:
+                        value = None
+                    label_in = label in endpoint_labels and (
+                        not value or str(endpoint_labels[label]) == value
+                    )
+                    if not label_in:
+                        break
+
+            if label_in:
+                endpoint_list.append(endpoint)
```
gtopper: I just realized this is the same logic as in `SQLStoreBase`. Can we eliminate this duplication?

gtopper (commented):
Other than that it looks right 👍

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
davesh0812: Nice catch!

gtopper (approved):

PR #5848, opened 2024-06-26
[Runs] Partition by (project, name) instead of just name

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-6913 (details are in task comments)

### DIFF
--- a/mlrun/common/schemas/constants.py
+++ b/mlrun/common/schemas/constants.py
@@ -120,10 +120,13 @@ class FeatureStorePartitionByField(mlrun.common.types.StrEnum):
 
 class RunPartitionByField(mlrun.common.types.StrEnum):
     name = "name"  # Supported for runs objects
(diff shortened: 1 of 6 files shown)


liranbg (approved):

liranbg:
@rokatyy - we need to ensure UI team query that way too.

theSaarco (approved):
Looks good. It may actually make more sense to keep most (or at least some) tests as-is, as sort of a "regression" test - the impact of the new mode is only expected to be felt when doing a cross-project query (such as the new test you added).

rokatyy:
@theSaarco I changed all the places where we used filtering by `name` to use `project_and_name`, however decided not to delete `name` iself for now to avoid breaking BC(Btw, do you think it's good idea?). So, in reality there is no place left which uses old functionality. So, not sure if we need tests with old functionality.

PR #5849, opened 2024-06-26
[Functions] Remove the explicit status removal

rokatyy:
Part of https://iguazio.atlassian.net/browse/ML-6929 (details are in ticket)

### DIFF
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -1752,7 +1752,6 @@ class SQLDB(DBInterface):
         ):
             function_dict = function.struct
             if not function_tag:
-
                 # function status should be added only to tagged functions
                 # TODO: remove explicit cleaning
                 function_dict["status"] = None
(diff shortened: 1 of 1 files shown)


server/api/db/sqldb/db.py
```diff
@@ -1753,6 +1753,7 @@ def list_functions(
             function_dict = function.struct
             if not function_tag:
                 # function status should be added only to tagged functions
+                # TODO: remove explicit cleaning
```
TomerShor: Comment what you wrote in the Jira, about multiple functions in the UI

TomerShor (changes requested):
Minor

server/api/db/sqldb/db.py
```diff
@@ -1753,6 +1753,9 @@ def list_functions(
             function_dict = function.struct
             if not function_tag:
                 # function status should be added only to tagged functions
+                # TODO: remove explicit cleaning; we also
+                #  will need to understand how to display functions in UI), because if we simply delete this string,
```
```suggestion
                #  will need to understand how to display functions in UI, because if we do not remove the status here,
```

TomerShor (approved):

liranbg (approved):

PR #5851, opened 2024-06-27
[API] Fix warning raised from pydantic for private arguments

liranbg:
Rename _format -> format_ to avoid warning logs of
`2024-06-27T02:26:52.150713007Z stderr F fields may not start with an underscore, ignoring "_format"`

the reason we had `_` before is not to mark it as private but rather avoid overriding python builtins.
instead, the convention should be to put the `_` after the argument name.


quaark (approved):

moranbental (approved):
LGTM

PR #5854, opened 2024-06-27
[CI] Change the enterprise system tests schedule to once a day

jond01:
The previous schedule was once every 8 hours, but since this workflow takes longer to complete - about 18 hrs on dev:
https://github.com/mlrun/mlrun/actions/runs/9629559040

Since this workflow has a single runner and explicitly forbids concurrency, it caused most of the runs to fail as there's no available runner.
Some recent examples:
https://github.com/mlrun/mlrun/actions/runs/9688898397
https://github.com/mlrun/mlrun/actions/runs/9658883699

Changing it to once a day will prevent such situations and reduce the noise in the workflow's runs page:
https://github.com/mlrun/mlrun/actions/workflows/system-tests-enterprise.yml

### DIFF
--- a/.github/workflows/system-tests-enterprise.yml
+++ b/.github/workflows/system-tests-enterprise.yml
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-#
+
 name: System Tests Enterprise
 
 on:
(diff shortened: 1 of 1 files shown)


alonmr (commented):
I think the RCA here should be why it takes so longs and we need to make it shorter.
Running the system tests once a day is what happens in practice but it provides bad coverage and makes it difficult to pinpoint which change broke some tests.

jond01:
> I think the RCA here should be why it takes so longs and we need to make it shorter. Running the system tests once a day is what happens in practice but it provides bad coverage and makes it difficult to pinpoint which change broke some tests.

I generally agree, but I don't think it is related to this PR...
The purpose here is to avoid unnecessary workflow failures in the current situation, and it actually improves the wait time between runs.

PR #5857, opened 2024-06-27
[Model Monitoring] Fix premature credentials resolution

alxtkr77:
https://iguazio.atlassian.net/browse/ML-6745


mlrun/datastore/targets.py
```diff
@@ -1069,6 +1064,47 @@ def prepare_spark_df(self, df, key_columns, timestamp_key=None, spark_options=No
         return df
 
 
+class ParquetTargetStoreyWrapper(storey.ParquetTarget):
```
gtopper, line 1067: What about other targets though?

mlrun/datastore/targets.py
```diff
@@ -1069,6 +1064,47 @@ def prepare_spark_df(self, df, key_columns, timestamp_key=None, spark_options=No
...
+        index_cols: Union[str, Union[list[str], list[tuple[str, str]]], None] = None,
+        columns: Union[str, Union[list[str], list[tuple[str, str]]], None] = None,
+        partition_cols: Union[
+            str, Union[list[str], list[tuple[str, int]]], None
+        ] = None,
+        time_field: Union[None, str, int] = None,
+        time_format: Optional[str] = None,
+        infer_columns_from_data: Optional[bool] = None,
+        max_events: Optional[int] = None,
+        flush_after_seconds: Union[int, float, None] = None,
```
gtopper: Probably better to just forward `*args` and `**kwargs`then repeating all the parameters here.

mlrun/model_monitoring/stream_processing.py
```diff
@@ -367,12 +367,11 @@ def apply_process_before_parquet():
         # Write the Parquet target file, partitioned by key (endpoint_id) and time.
         def apply_parquet_target():
             graph.add_step(
-                "storey.ParquetTarget",
+                "mlrun.datastore.targets.ParquetTargetStoreyWrapper",
```
gtopper, line 405: It seems to me like this problem isn't limited to model monitoring and `ParquetTarget` (even though the ticket is).

alxtkr77: Can you be more specific ?

gtopper: `CSVTarget`, `StreamTarget`, `NoSqlBaseTarget`, maybe others. Just search for `storage_options` in `mlrun/datastore/targets.py`.

alxtkr77: Those targets are not used by the model monitoring. For the mlrun regular graphs we use CM to pass the graph, so there is no credentials leakage

gtopper: What about the possibility of using `ParquetTarget` instead of creating a new wrapper class?

mlrun/datastore/targets.py
```diff
@@ -1069,6 +1064,36 @@ def prepare_spark_df(self, df, key_columns, timestamp_key=None, spark_options=No
...
 
+class ParquetTargetStoreyWrapper(storey.ParquetTarget):
+    def __init__(self, *args, **kwargs):
+        args = list(args)
+
+        if len(args) > 0:
+            path = args[0]
+        else:
+            path = None
+        path = kwargs.get("path", path)
```
gtopper:
```suggestion
            path = kwargs.get("path")
```

mlrun/datastore/targets.py
```diff
@@ -1069,6 +1064,36 @@ def prepare_spark_df(self, df, key_columns, timestamp_key=None, spark_options=No
         return df
 
 
+class ParquetTargetStoreyWrapper(storey.ParquetTarget):
+    def __init__(self, *args, **kwargs):
+        args = list(args)
+
+        if len(args) > 0:
```
gtopper:
```suggestion
        if args:
```

mlrun/datastore/targets.py
```diff
@@ -1069,6 +1064,36 @@ def prepare_spark_df(self, df, key_columns, timestamp_key=None, spark_options=No
...
+            path = args[0]
+        else:
+            path = None
+        path = kwargs.get("path", path)
+
+        external_storage_options = kwargs.get("storage_options", None)
```
gtopper:
```suggestion
        external_storage_options = kwargs.get("storage_options")
```

mlrun/datastore/targets.py
```diff
@@ -1069,6 +1064,36 @@ def prepare_spark_df(self, df, key_columns, timestamp_key=None, spark_options=No
...
+            storage_options = merge(storage_options, external_storage_options)
+        else:
+            storage_options = storage_options or external_storage_options
+
+        if storage_options:
+            kwargs["storage_options"] = storage_options
+
+        if len(args) > 0:
+            args[0] = url
+        if "path" in kwargs:
```
gtopper:
```suggestion

        if "path" in kwargs:
```

mlrun/datastore/targets.py
```diff
@@ -1069,6 +1064,36 @@ def prepare_spark_df(self, df, key_columns, timestamp_key=None, spark_options=No
...
+            kwargs["path"] = url
+
+        args = tuple(args)
```
gtopper:
```suggestion
```

mlrun/datastore/targets.py
```diff
@@ -1069,6 +1064,36 @@ def prepare_spark_df(self, df, key_columns, timestamp_key=None, spark_options=No
         return df
 
 
+class ParquetTargetStoreyWrapper(storey.ParquetTarget):
```
gtopper: @alxtkr77 is going to consider wrapping these targets as well: ParquetTarget, CSVTarget, StreamTarget, KafkaTarget, TDEngineTarget.

alxtkr77: I looked into it - no need

mlrun/datastore/targets.py
```diff
@@ -1069,6 +1070,35 @@ def prepare_spark_df(self, df, key_columns, timestamp_key=None, spark_options=No
...
+        storage_options = store.get_storage_options()
+        if storage_options and external_storage_options:
+            storage_options = merge(storage_options, external_storage_options)
+        else:
+            storage_options = storage_options or external_storage_options
+
+        if storage_options:
+            kwargs["storage_options"] = storage_options
+
+        if len(args) > 0:
```
gtopper: I guess here too
```suggestion
        if args:
```

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -26,6 +27,12 @@
 from mlrun.utils import logger
 
 
+class TDEngineTargetStoreyWrapper(storey.TDEngineTarget):
```
gtopper, line 30: I guess we could use a docstring explanation for why it's needed.

mlrun/datastore/targets.py
```diff
@@ -1069,6 +1070,35 @@ def prepare_spark_df(self, df, key_columns, timestamp_key=None, spark_options=No
         return df
 
 
+class ParquetTargetStoreyWrapper(storey.ParquetTarget):
```
gtopper: I guess we could use a docstring explanation for why it's needed.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -26,6 +27,17 @@
...
+    """
+    TDEngineTarget requires a connection string to the TDEngine server.
```
gtopper:
```suggestion
    TDEngineTarget requires a connection string to the TDEngine server, which may contain credentials.
```

mlrun/model_monitoring/stream_processing.py
```diff
@@ -42,6 +43,40 @@
 from mlrun.utils import logger
 
 
+class ParquetTargetStoreyWrapper(storey.ParquetTarget):
+    """
+    ParquetTarget requires a storage_options to be available.
```
gtopper:
```suggestion
    ParquetTarget may require storage_options, which may contain credentials.
```

gtopper (approved):

PR #5861, opened 2024-06-30
[Function] Validate tag name before save

roei3000b:
Validate tag adheres to k8s limitations before we get to the DB.

Resolves https://iguazio.atlassian.net/browse/ML-6567
image

### DIFF
--- a/server/api/api/endpoints/nuclio.py
+++ b/server/api/api/endpoints/nuclio.py
@@ -456,8 +456,9 @@ def _deploy_function(
         launcher = server.api.launcher.ServerSideLauncher(auth_info=auth_info)
         launcher.enrich_runtime(runtime=fn, full=True)
 
-        fn.save(versioned=False)
         fn.pre_deploy_validation()
+        fn.save(versioned=False)
+
         fn = _deploy_nuclio_runtime(
             auth_info,
             builder_env,
--- a/server/api/api/endpoints/nuclio.py
+++ b/server/api/api/endpoints/nuclio.py
@@ -455,7 +455,7 @@ def _deploy_function(
         # Enrich runtime
         launcher = server.api.launcher.ServerSideLauncher(auth_info=auth_info)
         launcher.enrich_runtime(runtime=fn, full=True)
-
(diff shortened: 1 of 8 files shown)


server/api/api/endpoints/nuclio.py
```diff
@@ -455,7 +455,7 @@ def _deploy_function(
         # Enrich runtime
         launcher = server.api.launcher.ServerSideLauncher(auth_info=auth_info)
         launcher.enrich_runtime(runtime=fn, full=True)
-
+        fn.pre_save_validation()
         fn.save(versioned=False)
         fn.pre_deploy_validation()
```
TomerShor: Why not just move the `pre_deploy_validation` before the `save`?

roei3000b: Because it is validation before saving and not before deploying.
I also thought about renaming `pre_deploy_validation` to `pre_save_validation` but again, it is not good because we call this function before deploying and not before saving.

TomerShor: Spoke offline, tag validation can be moved here: https://github.com/mlrun/mlrun/blob/development/server/api/db/sqldb/db.py#L1687

server/api/api/endpoints/nuclio.py
```diff
@@ -455,7 +455,6 @@ def _deploy_function(
         # Enrich runtime
         launcher = server.api.launcher.ServerSideLauncher(auth_info=auth_info)
         launcher.enrich_runtime(runtime=fn, full=True)
-
```
TomerShor, line 458: leave this empty line, it keeps things organized

TomerShor (approved):
Minor thing but otherwise approved

liranbg (approved):

alonmr (commented):
~why enforce a k8s validation on tag? isn't it relevant only for nuclio runtimes?~ nvm though I do think we need a better place for these kind of validations, we shuoldn't have to copy this line to every tagable object

liranbg (approved):

TomerShor (approved):

PR #5863, opened 2024-06-30
[Alerts] Optimize access to DB

yanburman:
We lower access to DB by re-arranging calls
to avoid getting AlertConfig from db when
not necessary. Also, cache AlertState
and AlertConfig structs

https://iguazio.atlassian.net/browse/ML-6977
https://iguazio.atlassian.net/browse/ML-6993

### DIFF
--- a/server/api/utils/lru_cache.py
+++ b/server/api/utils/lru_cache.py
@@ -1,0 +1,105 @@
+# Copyright 2024 Iguazio
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import collections
+import hashlib
+from copy import deepcopy
+
+
+class LRUCache:
+    """LRU cache which solves some of lru_cache from functools deficiencies.
+    Most importantly, the ability to remove/modify a cache value on top
+    of the ability to clear entire cache.
+    We strive to provide similar API to lru_cache.
+    Note this class is used as is instead of using a decorator for performance reasons.
+    If performance is not as important, a decorator can be added.
+    Note this can be used as a decorator, but only with default parameters i.e. maxsize
+    This code is based on https://pastebin.com/LDwMwtp8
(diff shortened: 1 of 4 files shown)


server/api/utils/lru_cache.py
```diff
@@ -0,0 +1,93 @@
...
+            self.misses = 0
+            self.currsize = 0
+
+    def __init__(self, func, maxsize: int = 128):
+        self.cache = collections.OrderedDict()
+        self.func = func
+        self.maxsize = maxsize
+        self._cache_info = self.CacheInfo(maxsize)
+
+    def __call__(self, *args):
```
quaark: I suggest extending this to support both args and kwargs.
You can take the args and kwargs and hash them to create the cache key. Something like:
```
key = hashlib.sha256(f"{args}/{kwargs}".encode()).hexdigest()
cache[key] = result
```

server/api/utils/lru_cache.py
```diff
@@ -0,0 +1,93 @@
...
+            cache.move_to_end(args)
+            return cache[args]
+        result = self.func(*args)
+        cache[args] = result
+        self._cache_info.misses += 1
+        if len(cache) > self.maxsize:
+            cache.popitem(last=False)
+        return result
+
+    def cache_info(self) -> CacheInfo:
```
quaark, line 62: you can make this a property and then use it elsewhere like:
```
self.cache_info.misses += 1
```

server/api/utils/lru_cache.py
```diff
@@ -0,0 +1,93 @@
...
+        cache = self.cache
+        if args in cache:
+            cache[args] = value
+            cache.move_to_end(args)
+
+    def cache_set(self, value, *args) -> None:
+        """Set a single value in cache"""
+        cache = self.cache
+        cache[args] = value
+        # TODO: Will this be more efficient if we check if args already in cache first?
```
quaark: probably, yes. but if it exists don't you want to update the value anyway?
If not, then you can use `cache.setdefault(key, value)`

quaark (changes requested):
Looks good, a couple of improvement suggestions for the LRU cache class

server/api/utils/lru_cache.py
yanburman: I am returning a deepcopy, so that user doesn't modify my internal structures, so it won't work as is

server/api/utils/lru_cache.py
yanburman: if it exists, I don't need to check if I need to pop a value

server/api/utils/lru_cache.py
quaark: Add a doctoring explaining why we need this please

quaark (approved):

PR #5866, opened 2024-06-30
[Project] Delete serving config maps with project

alonmr:
https://iguazio.atlassian.net/browse/ML-6994

I noticed that on the system tests setup the serving functions config maps were created over a month ago so I checked why they were not deleted and found it was not implemented. Added the implementation here with some minor flavor changes.


server/api/api/utils.py
```diff
@@ -1378,7 +1379,7 @@ async def delete_function(
                 return None
             except Exception as exc:
                 # return tuple with failure info
-                return function, str(exc)
```
liranbg, line 1381: that might be intentional to avoid bloating the failure reason. @rokatyy correct?

liranbg (commented):
minor

server/api/api/utils.py
rokatyy: @liranbg yes, that's correct, I'd not change that

liranbg (approved):

server/api/api/utils.py
alonmr: noted. though I would like to move `_delete_nuclio_functions_in_batches` to `server/api/crud/runtimes/nuclio/function.py`. Its a more fitting place IMO. wdyt?

rokatyy: @alonmr agree 👍

PR #5869, opened 2024-06-30
[KFP] Support no-op adapter

liranbg:
This adapter, gives the client the ability to
1. Remove the default adapter (Currently 1.8) from mlrun's requirement
2. Install the nop adapter to ensure mlrun wont break when default adapter is missing

The idea is to relax mlrun requirements without breaking compatibility of removing the default kfp requirement.


alonmr (commented):
Might be worth to add some warnings like we do with NopDB. At least for the functions we know are not supposed to be called like `compile_pipeline`

PR #5872, opened 2024-07-01
[SDK] Fix Artifact Result Render In Jupyter

quaark:
https://iguazio.atlassian.net/browse/ML-6551

The artifact results where being pulled from Jupyter as files with Jupyter's `/files/...` API.
The issue was we sent a relative path starting from `/files/...` and for some reason either Jupyter or the browser aren't parsing the URL properly and requesting a path which gets redirected to the Jupyter's home page.

The fix is to build the full URL using the origin of the window's address and construct the URL: `https://jupyter.<FQDN>/files/...`

Before:
Snip20240523_6

After:
Screenshot 2024-07-01 at 13 50 27


mlrun/render.py
```diff
@@ -283,9 +283,16 @@ def get_style():
...
-  console.log(el.title);
 
-  document.querySelector(panelName + "-title").innerHTML = el.title
+  // Get the base URL of the current notebook
+  var baseUrl = window.location.origin;
+
+  // Construct the full URL
+  var fullUrl = new URL(el.title, baseUrl).href;
+
+  console.log(fullUrl);
```
TomerShor: not needed IMO

TomerShor (approved):

alonmr (approved):
NIce!!

PR #5873, opened 2024-07-01
[Docs] Add html generation for alerts structures

yanburman:


docs/api/mlrun.alerts.rst
moranbental, line 1: This file is already generated by `mlrun.alerts/index` so you don't need it.

moranbental (changes requested):
It looks good, just one comment

quaark (approved):

PR #5874, opened 2024-07-01
[Model Monitoring] Exclude non-v3io stream from redeployment mechanism 

Eyal-Danieli:
Following the security issue fix in #5393  , the mm stream pod is redeployed with the “new” V3IO stream trigger when the old version is based on the old V3IO stream location. 

However, this fix doesn’t take into account non-v3io streams such as Kafka and as a result each deployment of a serving function with non-v3io stream will lead to redeployment of the mm stream pod.

This PR excludes non-v3io streams from the mm stream pod redeployment process. 

A related JIRA: https://iguazio.atlassian.net/browse/ML-7002

### DIFF
--- a/server/api/api/endpoints/nuclio.py
+++ b/server/api/api/endpoints/nuclio.py
@@ -527,7 +527,15 @@ def _deploy_nuclio_runtime(
                         "On deploy of serving-functions which is based on mlrun image "
                         "('mlrun/') and with set-tracking enabled, client version must be >= 1.6.3"
                     )
-                if not monitoring_deployment.is_monitoring_stream_has_the_new_stream_trigger():
+
+                # Update model monitoring stream trigger if using old V3IO stream path
+                stream_paths = server.api.crud.model_monitoring.get_stream_path(
+                    project=fn.metadata.project,
+                )
+                if (
+                    stream_paths[0].startswith("v3io")
+                    and not monitoring_deployment.is_monitoring_stream_has_the_new_stream_trigger()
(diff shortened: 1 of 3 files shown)


davesh0812 (approved):
The condition looks great, I do  think that the check should be inside `is_monitoring_stream_has_the_new_stream_trigger` and not in `nuclio.py`, but it is not that important. :)

server/api/api/endpoints/nuclio.py
```diff
@@ -527,6 +527,8 @@ def _deploy_nuclio_runtime(
                         "On deploy of serving-functions which is based on mlrun image "
                         "('mlrun/') and with set-tracking enabled, client version must be >= 1.6.3"
                     )
+
+                # Update model monitoring stream trigger if using old V3IO stream path
```
assaf758: i think some refactoring can add to readblity here.
* is_ce_mode() is also checking along these lines
* the check in is_monitoring_stream_has_the_new_stream_trigger() is really checking if the monitoring needs to be redeployed (ex for kafka it does not)

maybe something like:
```
if serving_to_monitor:
   if monitoring_deployment.is_monitoring_stream_redeploy_needed():
        if (fn.spec.image.startswith("mlrun/")...):
            ...    
            raise mlrun.errors.MLRunBadRequestError(..)
                
        monitoring_deployment.deploy_model_monitoring_stream_processing(
                        overwrite=True
                    )
```

assaf758 (approved):

PR #5875, opened 2024-07-01
[Runtimes] Enrich dask runtime node selector from project-level node selector 

yaelgen:
Follow up for: https://github.com/mlrun/mlrun/pull/5775
Resolves: [ML-6479](https://iguazio.atlassian.net/browse/ML-6479) 

[ML-6479]: https://iguazio.atlassian.net/browse/ML-6479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/api/runtime_handlers/daskjob.py
+++ b/server/api/runtime_handlers/daskjob.py
@@ -396,6 +396,11 @@ def enrich_dask_cluster(
     worker_container = client.V1Container(
         resources=spec.worker_resources, args=worker_args, **container_kwargs
     )
+
+    # We query the project to enrich the worker and scheduler pod spec with the project's default node selector.
+    # since the dask runtime operates locally and bypasses the usual server API flow, we do not enrich the run object.
+    # however, the node selector is still relevant for the Dask cluster's workers and scheduler,
+    # this ensures that the cluster pods follow the project's specified node selection.
     project = function._get_db().get_project(function.metadata.project)
     logger.info(
         "Enriching Dask Cluster node selector from project",
--- a/server/api/runtime_handlers/daskjob.py
+++ b/server/api/runtime_handlers/daskjob.py
@@ -398,9 +398,10 @@ def enrich_dask_cluster(
     )
 
     # We query the project to enrich the worker and scheduler pod spec with the project's default node selector.
-    # since the dask runtime operates locally and bypasses the usual server API flow, we do not enrich the run object.
-    # however, the node selector is still relevant for the Dask cluster's workers and scheduler,
-    # this ensures that the cluster pods follow the project's specified node selection.
+    # Since the dask runtime is a local run, and does not run in a dedicated k8s pod, node selectors for that run
+    # are irrelevant, so we do not enrich the run object with the project node selector.
+    # However, the node selector is still relevant for the Dask cluster's workers and scheduler, which do run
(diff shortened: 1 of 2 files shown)


tests/api/runtimes/test_dask.py
```diff
@@ -512,6 +514,12 @@ def test_enrich_dask_cluster(self):
         assert scheduler_pod.spec.containers[0].resources == expected_resources
         assert worker_pod.spec.containers[0].env == expected_env
         assert scheduler_pod.spec.containers[0].env == expected_env
+        assert (
+            scheduler_pod.spec.node_selector
+            == mlrun.utils.helpers.merge_with_precedence(
+                self.project_default_function_node_selector, function.spec.node_selector
+            )
```
alonmr: it is better practice to validate against an expected value and not a dependency of another function.
for example. if someone removes the value for `self.project_default_function_node_selector` the test still passes though it removes the coverage

server/api/runtime_handlers/daskjob.py
```diff
@@ -397,11 +397,24 @@ def get_env_name(env_: Union[client.V1EnvVar, dict]) -> str:
         resources=spec.worker_resources, args=worker_args, **container_kwargs
     )
 
+    # We query the project to enrich the worker and scheduler pod spec with the project's default node selector.
+    # since the dask runtime operates locally and bypasses the usual server API flow, we do not enrich the run object.
+    # however, the node selector is still relevant for the Dask cluster's workers and scheduler,
+    # this ensures that the cluster pods follow the project's specified node selection.
+    project = function._get_db().get_project(function.metadata.project)
+    logger.info(
```
alonmr: IMO the log level should be debug

alonmr (approved):
lgtm, would change the log level but still approved 👍

server/api/runtime_handlers/daskjob.py
```diff
@@ -397,11 +397,24 @@ def get_env_name(env_: Union[client.V1EnvVar, dict]) -> str:
         resources=spec.worker_resources, args=worker_args, **container_kwargs
     )
 
+    # We query the project to enrich the worker and scheduler pod spec with the project's default node selector.
+    # since the dask runtime operates locally and bypasses the usual server API flow, we do not enrich the run object.
+    # however, the node selector is still relevant for the Dask cluster's workers and scheduler,
+    # this ensures that the cluster pods follow the project's specified node selection.
```
quaark:
```suggestion
    # We query the project to enrich the worker and scheduler pod spec with the project's default node selector.
    # Since the dask runtime is a local run, and does not run in a dedicated k8s pod, node selectors for that run
    # are irrelevant, so we do not enrich the run object with the project node selector.
    # However, the node selector is still relevant for the Dask cluster's workers and scheduler, which do run
    # remotely on k8s. This ensures that the cluster pods follow the project's specified node selection.
```
I think the wording and terminology here is quite important, and we should make sure to make it super clear the business/functional reasoning for doing this and not only the technological limitations (bypassing the API)

quaark (commented):
Code looks good!
I suggest slightly different wording on the comment and maybe add it to the PR description as well

quaark (approved):

liranbg (approved):

PR #5881, opened 2024-07-02
[Functions] Do not add empty string to function external urls if function doesn't have any external url

rokatyy:
This PR addresses issue when we had `external_invocation_url` equal to `[""]` when deploying function without external address. After this changes, we won't add empty string to the list.

jira - https://iguazio.atlassian.net/browse/ML-6988


mlrun/runtimes/nuclio/function.py
```diff
@@ -1327,3 +1327,23 @@ def get_nuclio_deploy_status(
     else:
         text = "\n".join(outputs) if outputs else ""
         return state, address, name, last_log_timestamp, text, function_status
+
+
+def enrich_nuclio_function_from_headers(
+    func,
+    headers,
```
alonmr: add type hints please

alonmr (commented):
nice! just a minor nitpicking above

server/api/api/endpoints/nuclio.py
```diff
@@ -587,8 +587,16 @@ def _handle_nuclio_deploy_status(
...
+    internal_invocation_urls = (
+        status.get("internalInvocationUrls")
+        if status.get("internalInvocationUrls")
+        else []
+    )
+    external_invocation_urls = (
+        status.get("externalInvocationUrls")
+        if status.get("externalInvocationUrls")
+        else []
+    )
```
TomerShor, line 599: Why is this different than `status.get("externalInvocationUrls", [])`

mlrun/runtimes/nuclio/function.py
rokatyy: @alonmr I tried to add `Union[RemoteRuntime, ApplicationRuntime]`, but then I got an error about circular imports..

mlrun/db/httpdb.py
```diff
@@ -1685,16 +1677,8 @@ def get_builder_status(
...
-                func.status.internal_invocation_urls = resp.headers.get(
-                    "x-mlrun-internal-invocation-urls", ""
-                ).split(",")
-                func.status.external_invocation_urls = resp.headers.get(
-                    "x-mlrun-external-invocation-urls", ""
-                ).split(",")
-                func.status.container_image = resp.headers.get(
-                    "x-mlrun-container-image", ""
+                mlrun.runtimes.nuclio.function.enrich_nuclio_function_from_headers(
+                    func, resp.headers
```
TomerShor, line 1681: I noticed that in this flow `func.status.state` was not enriched, and now it will be. is this on purpose?

server/api/api/endpoints/nuclio.py
rokatyy: @TomerShor because status["externalInvocationUrls"] can be equal to empty string

mlrun/db/httpdb.py
rokatyy: @TomerShor it's enriched above, so we will just enrich it again. This flow is not only for nuclio functions, so only for nuclio functions it will be enriched twice.

mlrun/runtimes/nuclio/function.py
rokatyy: Added

alonmr (approved):

PR #5883, opened 2024-07-02
[Docs] Speedup html docs generation about 2x

yanburman:
Before:
real    2m12.262s
user    2m2.838s
sys     0m2.296s

After:
real    1m19.884s
user    3m1.104s
sys     0m8.098s


moranbental (approved):
LGTM

liranbg (approved):

PR #5885, opened 2024-07-02
[Model Monitoring] Fix error on empty TSDB table

gtopper:
This was fixed in #5706, but not for the case where framesd already accessed the table, the table was deleted (e.g. if whole project was deleted), and then we try to read the table again. In this case, the error is different.


mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -38,7 +38,8 @@ def _is_no_schema_error(exc: v3io_frames.ReadError) -> bool:
     In case of a nonexistent TSDB table - a `v3io_frames.ReadError` error is raised.
     Check if the error message contains the relevant string to verify the cause.
     """
-    return "No TSDB schema file found" in str(exc)
+    msg = str(exc)
+    return "No TSDB schema file found" in msg or "Failed to read schema at path" in msg
```
jond01, line 44: Could you please add the reference for the source of these error messages?
It will be easier to track in the future.

jond01 (approved):
Great catch :)

PR #5886, opened 2024-07-02
[Notifications] Check if secret key is internal before deleting it

Yacouby:
resolves:
https://iguazio.atlassian.net/browse/ML-6936

when deleting notifications, only internal project secrets should be deleted.

also, this PR introduces a change to skip the validation of notification parameters when the "secret" key is present in the secret parameters.

### DIFF
--- a/mlrun/model.py
+++ b/mlrun/model.py
@@ -737,8 +737,17 @@ class Notification(ModelObj):
             self.kind
         ).get_notification()
 
-        secret_params = self.secret_params
-        params = self.params
+        secret_params = self.secret_params or {}
+        params = self.params or {}
+
+        # if the secret_params are already masked - no need to validate
+        params_secret = secret_params.get("secret", "")
+        if params_secret:
+            if len(secret_params) > 1:
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    "When the 'secret' key is present, 'secret_params' should not contain any other keys."
+                )
+            return
 
         if not secret_params and not params:
             raise mlrun.errors.MLRunInvalidArgumentError(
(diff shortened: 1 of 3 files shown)


server/api/api/utils.py
```diff
@@ -429,7 +430,11 @@ def validate_and_mask_notification_list(
         # validate notification schema
         mlrun.common.schemas.Notification(**notification_object.to_dict())
 
-        notification_object.validate_notification_params()
+        # skip the params validation if the key is "secret"
```
alonmr: Not sure I understand why. Also why not check this in the validation function?

Yacouby: after a conversation with @quaark we decided that we should skip the validation when "secret" is present because it implies that the parameters are securely handled and should not be validated.

mlrun/model.py
```diff
@@ -737,8 +737,14 @@ def validate_notification_params(self):
             self.kind
         ).get_notification()
 
-        secret_params = self.secret_params
-        params = self.params
+        secret_params = self.secret_params or {}
+        params = self.params or {}
+
+        # skip the secret_params validation if it contains the key "secret" since the parameters are securely handled
+        # and should not be validated in this case
```
```suggestion
        # if the secret_params are already masked - no need to validate
```

mlrun/model.py
```diff
@@ -737,8 +737,14 @@ def validate_notification_params(self):
...
+        params_secret = secret_params.get("secret", "")
+        if params_secret:
+            return
```
alonmr: I think the heuristic can be improved. We should also verify that this is the only param in the dict because if there are others than they will be ignored. I think we should be explicit about it. (That is IIUC the behavior @quaark )

quaark: Yes I agree that's a good idea. 
The error message should also be indicative and explain this mutual exclusion

alonmr (approved):

quaark (approved):

PR #5891, opened 2024-07-03
[Artifacts] Revert Legacy Artifact Conversion Deprecation

quaark:
https://iguazio.atlassian.net/browse/ML-6466

We were only supposed to deprecate this feature in 1.8.0, not 1.7.0.
We will remove them when needed.


server/api/crud/artifacts.py
```diff
@@ -58,6 +58,12 @@ def store_artifact(
         # calculate the size of the artifact
         self._resolve_artifact_size(artifact, auth_info)
 
+        # TODO: remove this in 1.8.0
+        if mlrun.utils.helpers.is_legacy_artifact(artifact):
+            artifact = mlrun.artifacts.base.convert_legacy_artifact_to_new_format(
+                artifact
+            ).to_dict()
+
```
alonmr, line 66: I don't think this is needed. IIRC we only need `convert_legacy_artifact_to_new_format` for data migrations. no one should create a legacy artifact

alonmr (approved):

PR #5896, opened 2024-07-03
[API] Efficient Project Summaries

quaark:
https://iguazio.atlassian.net/browse/ML-6752

Previously the project summary cache was only invalidated and recalculated upon user request.
Furthermore, the recalculation was done on all projects (during the request) even though the request might request a specific project's summary.

Now the recalculation is a periodic function triggered from the main startup thread.
The summaries are refreshed in the background on the chief and in each worker (TODO: consider caching the result in the DB and only running on chief) and the results are always pulled in O(1) from the cache. **Only** on startup if a request squeezes in before the background refresh, then the refresh will happen during the request.

## Important Note
This PR also installs `pytest-fork` and runs unit tests now in a separate process from the main process collecting and running all the tests. This achieves 2 things:
- The transient segfaults from the scheduler if they happen are contained and don't crash the entire CI step. (It seems though running these tests in forks remedies the issue completely. My current guess is starting and stopping the scheduler multiple times within the same process isn't properly supported).
- The test outputs in the CI window now only show outputs for tests that failed, so it's much easier to navigate and understand failures

There also existed a test that makes sure certain properties aren't persisted between tests, but this test now fails and anyway isn't relevant as now each test has its own contained memory space thanks to the OS's memory pager.


mlrun/config.py
```diff
@@ -104,7 +104,12 @@
             # max number of parallel abort run jobs in runs monitoring
             "concurrent_abort_stale_runs_workers": 10,
             "list_runs_time_period_in_days": 7,  # days
-        }
+        },
+        "projects": {
+            "summaries": {
+                "cache_interval": "30",
```
roei3000b, line 114: We already have `projects` section that contains `counters_cache_ttl`.
I think you should put the `cache_interval` there instead the `counters_cache_ttl` because it is not relevant anymore.

server/api/crud/projects.py
```diff
@@ -215,18 +214,15 @@ async def list_project_summaries(
...
-            projects_output.projects
-        )
+        if not self._project_resource_counters_cache:
+            await self.refresh_project_resources_counters_cache(session)
+
+        project_summaries = []
+        for project_name in names or self._project_resource_counters_cache.keys():
+            project_summaries.append(
+                self._project_resource_counters_cache.get(project_name)
+            )
```
alonmr: Isn't it the same as?
```suggestion
        project_summaries = self._project_resource_counters_cache.values() or []
```

server/api/db/session.py
```diff
@@ -37,3 +37,17 @@ def run_function_with_new_db_session(func, *args, **kwargs):
         return result
     finally:
         close_session(session)
+
+
+async def run_async_function_with_new_db_session(func, *args, **kwargs):
```
alonmr, line 43: I think this can be merged with the above using `maybe_coroutine`

server/api/main.py
```diff
@@ -238,6 +238,10 @@ async def move_api_to_online():
             await _start_periodic_logs_collection()
             await _start_periodic_stop_logs()
 
+    # periodic project summaries calculation should run on all instances
+    # not just the chief
+    _start_periodic_project_summaries_calculation()
```
alonmr: not sure I understand why run it on workers. that means each worker can hold a different calculation on only 1 of them will have the most up to date one. if the chief does this anyway, its not really offloading cpu usage from it.

alonmr (commented):
Looks nice! some comments / questions below

server/api/crud/projects.py
quaark: no, since you can request specific names.
see 
```
for project_name in names or self._project_resource_counters_cache.keys():
```

server/api/main.py
quaark: This isn't an offloading issue, the issue is the cache is in-memory and the request can arrive anywhere and I don't want to redirect it to the chief. Then they all need to hold a cache

server/api/db/session.py
quaark: That means that `run_function_with_new_db_session` will always be async and you need to await it, whether it awaits the function or not. Prefer not to do that refactor...

server/api/main.py
```diff
@@ -533,6 +537,21 @@ def _start_periodic_pagination_cache_monitoring():
...
+    interval = int(config.monitoring.projects.summaries.cache_interval)
+    if interval > 0:
+        logger.info(
+            "Starting periodic project summaries calculation", interval=interval
+        )
+        run_function_periodically(
+            interval,
+            server.api.crud.projects.Projects().refresh_project_resources_counters_cache.__name__,
+            False,
+            server.api.db.session.run_async_function_with_new_db_session,
```
roei3000b, line 589: Maybe it is a gap with my understanding, but why we need to run this function with new db session each time?

roei3000b (changes requested):
Very nice improvement :)
minor.

server/api/crud/projects.py
alonmr: oh I was confused by the syntax. maybe use `names = names or self._project_resource_counters_cache.keys()` for readability.

server/api/db/session.py
alonmr: you can also use `run_function_with_new_db_session` that will return a future and then await on it

server/api/main.py
alonmr: I think its debatable if its better to calculate on all APIs or to redirect to the chief if it's already cached it should be a quick response. I'm ok with this for now but we might want to stress test it

server/api/db/session.py
```diff
@@ -37,3 +37,17 @@ def run_function_with_new_db_session(func, *args, **kwargs):
...
+    """
+    Run a function with a new db session, useful for concurrent requests where we can't share a single session.
```
TomerShor:
```suggestion
    Run an async function with a new db session, useful for concurrent requests where we can't share a single session.
```

TomerShor (approved):
Nicely done

server/api/db/session.py
```diff
@@ -37,3 +37,17 @@ def run_function_with_new_db_session(func, *args, **kwargs):
...
+    However, any changes made by the new session will not be visible to old sessions until the old sessions commit
+    due to isolation level.
+    """
+    session = create_session()
+    try:
+        result = await func(session, *args, **kwargs)
+        return result
+    finally:
+        close_session(session)
```
liranbg: create and close session is IO blocking, run it in a thread

server/api/api/endpoints/projects.py
```diff
@@ -399,6 +415,17 @@ async def get_project_summary(
...
+        mlrun.mlconf.httpdb.clusterization.role
+        != mlrun.common.schemas.ClusterizationRole.chief
+    ):
+        logger.info(
+            "Requesting project summaries, re-routing to chief",
+            name=name,
+        )
+        chief_client = server.api.utils.clients.chief.Client()
+        return await chief_client.get_project_summary(project=name, request=request)
+
```
liranbg: i wouldnt delegate the responsibility to chief only. why not letting each worker/chief do it same as we have today?
later on, upon separation / shared queue then we can do it in one place and distribute to others

liranbg (changes requested):

server/api/api/endpoints/projects.py
alonmr: Making the workers collect project summaries trades network load for cpu load. IMO since the chief is doing it anyway, the network overhead will be insignificant in comparison to the cpu load on all workers. Also, that means you can make 2 requests to get the project summaries and get 2 different responses because each worker holds something else in cache

liranbg: > Making the workers collect project summaries trades network load for cpu load. IMO since the chief is doing it anyway, the network overhead will be insignificant in comparison to the cpu load on all workers. Also, that means you can make 2 requests to get the project summaries and get 2 different responses because each worker holds something else in cache

I agree the latter is not ideal, but that is what we have now. overall, not optimal at all but fact we need to redirect to chief is another hop/api load on chief we're trying to reduce
if that's a hassle - perhaps simply table of project summaries on DB with single record where  chief updates / workers reads could be ideal

liranbg (approved):

PR #5897, opened 2024-07-03
[Artifacts] Change the artifact target to include the artifact tag

moranbental:
This PR tries to resolve the issue that occurs when an artifact is logged with the v1 tag and then logged with the v2 tag.
Then get the dataitem v1 tag, which fails.
When running log artifact inside a function, the output function returns the artifact uri without the artifact tag, causing the get artifact to fail because if the uri does not have a tag, we will search for the latest tag (instead of v1 tag).

https://iguazio.atlassian.net/browse/ML-6602

### DIFF
--- a/tests/run/test_run.py
+++ b/tests/run/test_run.py
@@ -331,3 +331,15 @@ def test_get_or_create_ctx_run_kind_exists_in_mlrun_exec_config(
     )
     context = mlrun.get_or_create_ctx("ctx")
     assert context.labels.get("kind") == "spark"
+
+
+def test_varify_tag_exists_in_output_uri_on_run():
+    project = mlrun.get_or_create_project("dummy-project")
+    project.set_function(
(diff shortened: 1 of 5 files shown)


roei3000b (approved):

tests/run/test_run.py
```diff
@@ -331,3 +331,15 @@ def test_get_or_create_ctx_run_kind_exists_in_mlrun_exec_config(
     )
     context = mlrun.get_or_create_ctx("ctx")
     assert context.labels.get("kind") == "spark"
+
+
+def test_varify_tag_exists_in_output_uri_on_run():
```
TomerShor:
```suggestion
def test_verify_tag_exists_in_run_output_uri():
```

TomerShor (approved):
Minor typo but approved

liranbg (approved):

PR #5898, opened 2024-07-03
[Scripts] Fix patch remote igz docker registry resolution

TomerShor:
Reverting some of the changes in https://github.com/mlrun/mlrun/pull/5735 to fix the patch igz script.

The script now works with both the datanode docker registry and private registries.

### DIFF
--- a/automation/patch_igz/patch_remote.py
+++ b/automation/patch_igz/patch_remote.py
@@ -152,20 +152,21 @@ class MLRunPatcher:
 
     def _make_mlrun(self, target, image_tag, image_name) -> str:
         logger.info(f"Building mlrun docker image: {target}:{image_tag}")
-        mlrun_docker_registry = self._config["DOCKER_REGISTRY"]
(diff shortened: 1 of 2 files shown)


automation/patch_igz/patch_remote.py
```diff
@@ -152,20 +152,21 @@ def _get_current_version(self) -> str:
...
+        if mlrun_docker_repo:
+            mlrun_docker_registry = (
+                f"{mlrun_docker_registry}/{mlrun_docker_repo.rstrip('/')}"
+            )
 
         env = {
             "MLRUN_VERSION": image_tag,
-            "MLRUN_DOCKER_REGISTRY": mlrun_docker_registry,
-            "MLRUN_DOCKER_REPO": mlrun_docker_repo,
+            "MLRUN_DOCKER_REPO": mlrun_docker_registry,
```
alonmr, line 165: I think this needs a comment to explain it

alonmr (approved):
think a comment is needed to clarify. other than that lgtm

Yacouby (approved):
good job, now running the script is working on my env

PR #5900, opened 2024-07-03
[Schedules] Change default labels value to None and add enrichment with db labels

rokatyy:
This PR addresses an issue where `schedule.labels` were cleaned up when updating a schedule with no labels passed. Additionally, I have updated the `test_update_schedule` test by adding patching functionality to enforce label enrichment. Without this fix, the test fails.

Jira - https://iguazio.atlassian.net/browse/ML-7019


tests/api/utils/test_scheduler.py
```diff
@@ -1089,13 +1089,16 @@ async def test_schedule_convert_from_old_credentials_to_new(
     _assert_schedule_secrets(scheduler, project, schedule_name, None, None)
 
 
+@unittest.mock.patch.object(Scheduler, "_store_schedule_secrets_using_auth_secret")
```
alonmr: I think you can add the return value here no?
```suggestion
@unittest.mock.patch.object(Scheduler, "_store_schedule_secrets_using_auth_secret", return_value="auth-secret")
```

alonmr (commented):
It's important we add a unit test here. We should verify what happens when providing:
1. labels is None
2. labels is {} 
3. labels is not empty
All when we have labels in the DB of course

rokatyy:
@alonmr that sounds exactly like what `test_update_schedule` does :)

tests/api/utils/test_scheduler.py
rokatyy: @alonmr TIL 👍

liranbg (approved):

alonmr (approved):

PR #5901, opened 2024-07-04
[Notifications] Fix failure when runs is None 

GiladShapira94:
https://iguazio.atlassian.net/browse/ML-7032

### DIFF
--- a/mlrun/utils/notifications/notification/webhook.py
+++ b/mlrun/utils/notifications/notification/webhook.py
@@ -69,29 +69,31 @@ class WebhookNotification(NotificationBase):
             request_body["custom_html"] = custom_html
 
         if override_body:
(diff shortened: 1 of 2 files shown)


mlrun/utils/notifications/notification/webhook.py
```diff
@@ -69,29 +69,31 @@ async def push(
             request_body["custom_html"] = custom_html
 
         if override_body:
-            list_edit_runs = []
-            for run in runs:
```
alonmr, line 73: its better to verify that runs is a list with `runs = runs or []` instead of calling a nested function for each value that has {{runs}} in it because that nested function parses the runs each time instead of doing it once.

alonmr (approved):

PR #5904, opened 2024-07-04
[Datastore] Support both v3io:// and v3io:/// prefixes

alxtkr77:
https://iguazio.atlassian.net/browse/ML-7004

### DIFF
--- a/mlrun/datastore/datastore.py
+++ b/mlrun/datastore/datastore.py
@@ -205,9 +205,12 @@ class StoreManager:
     def get_or_create_store(
         self, url, secrets: dict = None, project_name=""
     ) -> (DataStore, str, str):
+        if url and url.startswith("v3io://") and not url.startswith("v3io:///"):
+            url = url.replace("v3io://", "v3io:///", 1)
+
         schema, endpoint, parsed_url = parse_url(url)
         subpath = parsed_url.path
-        store_key = f"{schema}://{endpoint}"
+        store_key = f"{schema}://{endpoint}" if endpoint else f"{schema}://"
 
         if schema == "ds":
             datastore_profile = datastore_profile_read(url, project_name, secrets)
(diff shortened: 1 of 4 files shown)


mlrun/datastore/datastore.py
```diff
@@ -205,9 +205,12 @@ def object(
     def get_or_create_store(
         self, url, secrets: dict = None, project_name=""
     ) -> (DataStore, str, str):
+        if url and url.startswith("v3io://") and not url.startswith("v3io:///"):
+            url = url.replace("v3io://", "v3io:///", 1)
```
gtopper: This means that the endpoint will always be empty, breaking the (possibly unused) feature, without deprecation, and without removing relevant code.

gtopper: Anyway, I'm pretty sure the fix should be [here](https://github.com/mlrun/mlrun/blob/v1.7.0-rc27/mlrun/datastore/sources.py#L222) and [here](https://github.com/mlrun/mlrun/blob/v1.7.0-rc27/mlrun/datastore/sources.py#L407).

assaf758: > This means that the endpoint will always be empty, breaking the (possibly unused) feature, without deprecation, and without removing relevant code.

@gtopper what feature gets broken and why? Alex is just replacing the v3io:// with v3io:///

also, your second comment includes two identical links

gtopper: @assaf758, the reason for the existence of `v3io:///` in the first place is that there is a feature in mlrun that allows for providing the endpoint (webapi host:port) inside the v3io URL. E.g. `v3io://webapi:8080/bigdata/path/to/file`. This is why the code is parsing an "endpoint" out of the URL.

Fixed the second link.

alxtkr77: > Anyway, I'm pretty sure the fix should be [here](https://github.com/mlrun/mlrun/blob/v1.7.0-rc27/mlrun/datastore/sources.py#L222) and [here](https://github.com/mlrun/mlrun/blob/v1.7.0-rc27/mlrun/datastore/sources.py#L407).

In the locations you suggested self.spark_url in v3io.py is set to 'v3io:/', and path always starts with '/' . 
So 'store.spark_url + path' always resolved to `//` - no need to change here. 
The only thing we need to fix is to make v3io store to handle 'v3io://a/b/c/d/' as it handles today 'v3io:///a/b/c/d/'

gtopper (approved):

PR #5905, opened 2024-07-04
[Model Monitoring] Revert tsdb path to origin

davesh0812:
https://iguazio.atlassian.net/browse/ML-7039


Eyal-Danieli (approved):
I suggest to add a relevant unit test to `test_target_path.py`

PR #5906, opened 2024-07-04
[MySQL] Avoid year 2038 problem in new MySQL tables

jond01:
Fixes [ML-7040](https://iguazio.atlassian.net/browse/ML-7040) for new model monitoring tables.
I verified the main MM app flow test with MySQL passed.

Added a new `datetime` method to the `SQLTypesUtil` class - not used yet.

## Background

(Copied from the description of [ML-6921](https://iguazio.atlassian.net/browse/ML-6921).)

https://en.wikipedia.org/wiki/Year_2038_problem

We use `TIMESTAMP` fields in almost all of our tables in the MySQL DBs:

* `mlrun` DB: https://github.com/mlrun/mlrun/blob/acf958f31f6a0e2658d6732680fc971fea2d28f7/server/api/utils/db/sql_types.py#L29
* `mlrun_model_monitoring` DB: https://github.com/mlrun/mlrun/blob/acf958f31f6a0e2658d6732680fc971fea2d28f7/mlrun/model_monitoring/db/stores/sqldb/models/base.py#L93

However, this type is designed to hold int32 values from the Unix epoch (1970-2038). For reference, see:

https://dev.mysql.com/doc/refman/8.0/en/datetime.html 

> `TIMESTAMP` has a range of `'1970-01-01 00:00:01'` UTC to `'2038-01-19 03:14:07'` UTC.

https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-28.html 

> The behavior of the `TIMESTAMP` type is also unaffected by this change; its maximum allowed value remains `'2038-01-19 03:14:07.999999'` UTC, regardless of platform. **For dates futureward of this, use the MySQL `DATETIME` type instead.** (WL #14630)

https://bugs.mysql.com/bug.php?id=12654 

From now on, it’s better to use `DATETIME` instead of `TIMESTAMP`:

> The `DATETIME` type is used for values that contain both date and time parts. MySQL retrieves and displays `DATETIME` values in `'YYYY-MM-DD hh:mm:ss'` format. The supported range is `'1000-01-01 00:00:00'` to `'9999-12-31 23:59:59'`.

[ML-7040]: https://iguazio.atlassian.net/browse/ML-7040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ML-6921]: https://iguazio.atlassian.net/browse/ML-6921?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Backwards Compatibility

The changed tables in this PR did not exist in 1.6, so there is no BC issue for users.
However, for developers working on 1.7 model monitoring with MySQL - the new tables need to be removed.

This can be done easily with:

```py
from sqlalchemy import create_engine

# Change according to the relevant database
connection_uri = "mysql+pymysql://root@<host>:<node-port>/mlrun_model_monitoring"

engine = create_engine(connection_uri)

with engine.connect() as connection:
    connection.execute("DROP TABLE IF EXISTS app_results,app_metrics;")
```

### DIFF
--- a/mlrun/model_monitoring/db/stores/sqldb/models/base.py
+++ b/mlrun/model_monitoring/db/stores/sqldb/models/base.py
@@ -11,9 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 from sqlalchemy import (
     DATETIME,
-    TIMESTAMP,
+    TIMESTAMP,  # TODO: migrate to DATETIME, see ML-6921
     Boolean,
     Column,
     Float,
@@ -91,11 +92,11 @@ class ModelEndpointsBaseTable(BaseModel):
     metrics = Column(EventFieldType.METRICS, Text)
     first_request = Column(
         EventFieldType.FIRST_REQUEST,
-        TIMESTAMP(timezone=True),
+        TIMESTAMP(timezone=True),  # TODO: migrate to DATETIME, see ML-6921
     )
     last_request = Column(
         EventFieldType.LAST_REQUEST,
-        TIMESTAMP(timezone=True),
+        TIMESTAMP(timezone=True),  # TODO: migrate to DATETIME, see ML-6921
     )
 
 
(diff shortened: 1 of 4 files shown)


theSaarco (approved):
Looks good 👍

Eyal-Danieli (approved):
LGTM

PR #5907, opened 2024-07-04
[Runs] Do not ensure run logs collected when runs are terminated

TomerShor:
When runs reach terminal state, the monitoring function ensures logs were collected for this run by checking that a log file exists for this run. If not - it requests the run logs straight from k8s, bypassing the log collector as it assumes something was wrong with it.

To decrease the IO load on the chief, we do not ensure logs were collected at all, as logs are a best-effort operation.

Resolves https://iguazio.atlassian.net/browse/ML-6807

### DIFF
--- a/tests/api/runtime_handlers/base.py
+++ b/tests/api/runtime_handlers/base.py
@@ -31,7 +31,6 @@ import server.api.crud
 import server.api.utils.clients.chief
 from mlrun.common.runtimes.constants import PodPhases, RunStates
 from mlrun.utils import create_test_logger, now_date
-from server.api.constants import LogSources
 from server.api.runtime_handlers import get_runtime_handler
 from server.api.utils.singletons.db import get_db
 from server.api.utils.singletons.k8s import get_k8s_helper
@@ -167,12 +166,6 @@ class TestRuntimeHandlerBase:
         if data is None:
             data = {"key": "value"}
         return client.V1ConfigMap(metadata=metadata, data=data)
-
-    def _generate_get_logger_pods_label_selector(self, runtime_handler):
-        run_label_selector = runtime_handler._get_run_label_selector(
-            self.project, self.run_uid
-        )
-        return f"mlrun/class,{run_label_selector}"
 
     def _assert_runtime_handler_list_resources(
         self,
@@ -514,13 +507,14 @@ class TestRuntimeHandlerBase:
             f"Unexpected number of calls to list_namespaced_pod "
             f"{get_k8s_helper().v1api.list_namespaced_pod.call_count}, expected {expected_number_of_calls}"
         )
-        expected_label_selector = (
-            expected_label_selector or runtime_handler._get_default_label_selector()
-        )
-        get_k8s_helper().v1api.list_namespaced_pod.assert_any_call(
-            get_k8s_helper().resolve_namespace(),
-            label_selector=expected_label_selector,
-        )
+        if expected_label_selector and expected_label_selector != "":
+            expected_label_selector = (
+                expected_label_selector or runtime_handler._get_default_label_selector()
+            )
+            get_k8s_helper().v1api.list_namespaced_pod.assert_any_call(
+                get_k8s_helper().resolve_namespace(),
+                label_selector=expected_label_selector,
+            )
 
     @staticmethod
     def _assert_list_namespaced_crds_calls(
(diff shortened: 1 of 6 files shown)


tests/api/runtime_handlers/base.py
```diff
@@ -514,13 +507,14 @@ def _assert_list_namespaced_pods_calls(
...
-        )
-        get_k8s_helper().v1api.list_namespaced_pod.assert_any_call(
-            get_k8s_helper().resolve_namespace(),
-            label_selector=expected_label_selector,
-        )
+        if expected_label_selector and expected_label_selector != "":
+            expected_label_selector = (
+                expected_label_selector or runtime_handler._get_default_label_selector()
+            )
+            get_k8s_helper().v1api.list_namespaced_pod.assert_any_call(
```
alonmr: shouldn't you assert either way?

tests/api/runtime_handlers/test_mpijob.py
```diff
@@ -664,11 +614,9 @@ def _mock_list_resources_pods(self):
...
         return mocked_responses[0].items
 
-    def _generate_get_logger_pods_label_selector(self, runtime_handler):
-        logger_pods_label_selector = super()._generate_get_logger_pods_label_selector(
-            runtime_handler
-        )
-        return f"{logger_pods_label_selector},{mlrun_constants.MLRunInternalLabels.mpi_job_role}=launcher"
+    @staticmethod
+    def _generate_run_pod_label_selector(run_uid):
+        return f"mlrun/uid={run_uid}"
```
alonmr: use `MLRunInternalLabels`

tests/api/runtime_handlers/test_sparkjob.py
```diff
@@ -170,21 +160,9 @@ async def test_delete_resources_completed_crd(
             self.runtime_handler,
             len(list_namespaced_crds_calls),
         )
-        self._assert_list_namespaced_pods_calls(
```
alonmr, line 173: what about the:
```
            [self.executor_pod, self.driver_pod],
            # additional time for wait for pods deletion - simulate pods gone
            [],
```
should be asserted no?

tests/api/runtime_handlers/test_sparkjob.py
```diff
@@ -493,7 +413,7 @@ def test_state_thresholds(self, db: Session, client: TestClient, threshold_state
         "force",
         (
             True,
-            False,
+            # False,
```
alonmr: comment

alonmr (commented):
Few comments about the unit tests

tests/api/runtime_handlers/test_sparkjob.py
TomerShor: I removed it while I was debugging some stuff but brought it back now.

tests/api/runtime_handlers/test_sparkjob.py
```diff
@@ -257,38 +236,17 @@ async def test_delete_resources_with_force(self, db: Session, client: TestClient
...
-    @pytest.mark.asyncio
-    async def test_monitor_run_completed_crd(self, db: Session, client: TestClient):
+    def test_monitor_run_completed_crd(self, db: Session, client: TestClient):
         list_namespaced_crds_calls = [
             [self.running_crd_dict],
             [self.completed_crd_dict],
         ]
         self._mock_list_namespaced_crds(list_namespaced_crds_calls)
-        # for the get_logger_pods with proper selector
-        list_namespaced_pods_calls = [
```
alonmr, line 284: I believe this test should call list pods at least once for the state threshold check. since you removed the assertion below I think it removes coverage

tests/api/runtime_handlers/test_sparkjob.py
```diff
@@ -298,38 +256,17 @@ async def test_monitor_run_completed_crd(self, db: Session, client: TestClient):
...
-    @pytest.mark.asyncio
-    async def test_monitor_run_failed_crd(self, db: Session, client: TestClient):
+    def test_monitor_run_failed_crd(self, db: Session, client: TestClient):
         list_namespaced_crds_calls = [
             [self.running_crd_dict],
             [self.failed_crd_dict],
         ]
         self._mock_list_namespaced_crds(list_namespaced_crds_calls)
-        # for the get_logger_pods with proper selector
-        list_namespaced_pods_calls = [
```
alonmr, line 325: same here

alonmr (approved):

PR #5911, opened 2024-07-04
[Model Monitoring] Create all tables that don't exist at `SQLStore` init

davesh0812:
https://iguazio.atlassian.net/browse/ML-7049

### DIFF
--- a/mlrun/model_monitoring/db/stores/sqldb/sql_store.py
+++ b/mlrun/model_monitoring/db/stores/sqldb/sql_store.py
@@ -292,61 +292,64 @@ class SQLStoreBase(StoreBase):
         model_endpoints_table = (
             self.ModelEndpointsTable.__table__  # pyright: ignore[reportAttributeAccessIssue]
         )
-
-        # Get the model endpoints records using sqlalchemy ORM
-        with create_session(dsn=self._sql_connection_string) as session:
-            # Generate the list query
-            query = session.query(self.ModelEndpointsTable).filter_by(
-                project=self.project
-            )
-
-            # Apply filters
-            if model:
-                model = model if ":" in model else f"{model}:latest"
-                query = self._filter_values(
-                    query=query,
-                    model_endpoints_table=model_endpoints_table,
-                    key_filter=mm_schemas.EventFieldType.MODEL,
-                    filtered_values=[model],
+        try:
+            # Get the model endpoints records using sqlalchemy ORM
+            with create_session(dsn=self._sql_connection_string) as session:
+                # Generate the list query
+                query = session.query(self.ModelEndpointsTable).filter_by(
+                    project=self.project
                 )
-            if function:
-                function_uri = f"{self.project}/{function}"
-                query = self._filter_values(
-                    query=query,
-                    model_endpoints_table=model_endpoints_table,
-                    key_filter=mm_schemas.EventFieldType.FUNCTION_URI,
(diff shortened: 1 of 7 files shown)


mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -540,7 +523,14 @@ def _create_tables_if_not_exist(self):
         for table in self._tables:
             # Create table if not exist. The `metadata` contains the `ModelEndpointsTable`
             if not self._engine.has_table(table):
+                logger.info(
+                    f"Creating table {table} on {self._sql_connection_string.split('/')[-1]} db."
```
gtopper: I guess it makes sense to extract `self._sql_connection_string.split('/')[-1]` outside the if condition and for loop.

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -539,8 +522,12 @@ def _create_tables_if_not_exist(self):
 
         for table in self._tables:
             # Create table if not exist. The `metadata` contains the `ModelEndpointsTable`
+            db_name = self._sql_connection_string.split("/")[-1]
```
jond01: https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.engine.make_url

I suggest the following more idiomatic parsing:

```py
from sqlalchemy.engine import make_url

db_name = make_url(self._sql_connection_string).database
```

For example, if the connection string includes parameters - `"postgresql://user:pass@host/dbname?alt_host=host1&alt_host=host2&ssl_cipher=%2Fpath%2Fto%2Fcrt"` - it may fail.

jond01 (approved):
Nice!

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -61,6 +61,7 @@ def __init__(
 
         self._sql_connection_string = kwargs.get("store_connection_string")
         self._engine = get_engine(dsn=self._sql_connection_string)
+        self._create_tables_if_not_exist()
```
gtopper, line 67: I don't think `SQLStoreBase.__init__()` is the right place for this. The store is instantiated from a lot of places, including unit tests and `_BatchWindow.__init__()`.

Surely there's a specific point in the flow to do this and not have to repeat it.

jond01: @gtopper these sub-inits don't create a true "resource", they simply populate the current object with the relevant values.
@davesh0812 I would even remove these sub-inits and move them to the main `__init__`.

davesh0812: It does creates the tables @jond01

jond01: Right, my mistake - I thought only about `self._init_tables()`.

The tables creation should move to `deploy_monitoring_functions` as a part of `enable_model_monitoring`.

davesh0812: it is not good enough, the user can have error while trying to run `list_mep`, right after he set the credentials and before he enables MM.

I will put it in the end of `set_cred` in order to avoid this error and also to have some check if the cred are valid.

PR #5914, opened 2024-07-07
[Model Monitoring] Delete stream resources during the delete project API  

Eyal-Danieli:
Delete model monitoring stream resources as part of the delete project API. It includes the stream resources of both mm-stream pod and mm-writer pod along with the existing applications stream resources. 

Note that if the stream resources are based on Kafka, the relevant topics will be removed. 

JIRA: https://iguazio.atlassian.net/browse/ML-6837

### DIFF
--- a/server/api/crud/model_monitoring/deployment.py
+++ b/server/api/crud/model_monitoring/deployment.py
@@ -917,7 +917,7 @@ class MonitoringDeployment:
 
         if not stream_paths:
             # No stream paths to delete
-            pass
+            return
 
         elif stream_paths[0].startswith("v3io"):
             # Delete V3IO stream
--- a/server/api/crud/model_monitoring/deployment.py
+++ b/server/api/crud/model_monitoring/deployment.py
@@ -746,10 +746,12 @@ class MonitoringDeployment:
                     function_name,
                     self.auth_info,
                     delete_app_stream_resources=True
+                    if function_name
                     not in [
                         mm_constants.MonitoringFunctionNames.STREAM,
                         mm_constants.MonitoringFunctionNames.APPLICATION_CONTROLLER,
-                    ],
+                    ]
+                    else False,
                     access_key=self.model_monitoring_access_key,
                 )
                 tasks.append(task)
@@ -913,7 +915,11 @@ class MonitoringDeployment:
                 )
             )
 
-        if stream_paths[0].startswith("v3io"):
+        if not stream_paths:
+            # No stream paths to delete
+            pass
+
+        elif stream_paths[0].startswith("v3io"):
             # Delete V3IO stream
             import v3io.dataplane
             import v3io.dataplane.response
@@ -936,7 +942,7 @@ class MonitoringDeployment:
                     logger.warning(
                         "Can't delete v3io stream",
                         stream_path=stream_path,
-                        error=e,
+                        error=mlrun.errors.err_to_str(e),
                     )
         elif stream_paths[0].startswith("kafka://"):
             # Delete Kafka topics
@@ -955,7 +961,7 @@ class MonitoringDeployment:
             try:
                 kafka_client = kafka.KafkaAdminClient(
                     bootstrap_servers=brokers,
-                    client_id=f"{project}",
+                    client_id=project,
                 )
                 kafka_client.delete_topics(topics)
                 logger.debug("Deleted kafka topics", topics=topics)
@@ -963,8 +969,13 @@ class MonitoringDeployment:
                 logger.warning(
                     "Can't delete kafka topics",
                     topics=topics,
-                    error=e,
+                    error=mlrun.errors.err_to_str(e),
                 )
+        else:
+            logger.warning(
+                "Stream path is not supported and therefore can't be deleted, expected v3io or kafka",
+                stream_path=stream_paths[0],
+            )
 
     def check_if_credentials_are_set(self, only_project_secrets: bool = False):
         """
--- a/server/api/crud/model_monitoring/deployment.py
+++ b/server/api/crud/model_monitoring/deployment.py
@@ -705,9 +705,9 @@ class MonitoringDeployment:
         Disable model monitoring application controller, writer, stream, histogram data drift application
         and the user's applications functions, according to the given params.
 
-        :param delete_resources:                    If True, it would delete the model monitoring controller & writer
-                                                    functions. Default True
-        :param delete_stream_function:              If True, it would delete model monitoring stream function,
+        :param delete_resources:                    If True, delete the model monitoring controller & writer functions.
+                                                    Default True.
+        :param delete_stream_function:              If True, delete model monitoring stream function,
                                                     need to use wisely because if you're deleting this function
                                                     this can cause data loss in case you will want to
                                                     enable the model monitoring capability to the project.
@@ -745,8 +745,7 @@ class MonitoringDeployment:
                     self.project,
                     function_name,
                     self.auth_info,
-                    delete_v3io_stream=not mlrun.mlconf.is_ce_mode()
-                    and function_name
+                    delete_app_stream_resources=True
                     not in [
                         mm_constants.MonitoringFunctionNames.STREAM,
                         mm_constants.MonitoringFunctionNames.APPLICATION_CONTROLLER,
@@ -819,7 +818,7 @@ class MonitoringDeployment:
         project_name: str,
         function_name: str,
         auth_info: mlrun.common.schemas.AuthInfo,
-        delete_v3io_stream: bool,
+        delete_app_stream_resources: bool,
         access_key: str,
     ):
         background_task_name = str(uuid.uuid4())
@@ -837,7 +836,7 @@ class MonitoringDeployment:
             function_name,
             auth_info,
             background_task_name,
-            delete_v3io_stream,
+            delete_app_stream_resources,
             access_key,
         )
 
@@ -848,19 +847,19 @@ class MonitoringDeployment:
         function_name: str,
         auth_info: mlrun.common.schemas.AuthInfo,
         background_task_name: str,
-        delete_v3io_stream: bool,
+        delete_app_stream_resources: bool,
         access_key: str,
-    ):
+    ) -> None:
         """
         Delete the model monitoring function and its resources.
 
-        :param db_session:              A session that manages the current dialog with the database.
-        :param project:                 The name of the project.
-        :param function_name:           The name of the function to delete.
-        :param auth_info:               The auth info of the request.
-        :param background_task_name:    The name of the background task.
-        :param delete_v3io_stream:      If True, delete the V3IO stream.
-        :param access_key:              Model monitoring access key.
+        :param db_session:                  A session that manages the current dialog with the database.
+        :param project:                     The name of the project.
+        :param function_name:               The name of the function to delete.
+        :param auth_info:                   The auth info of the request.
+        :param background_task_name:        The name of the background task.
+        :param delete_app_stream_resources: If True, delete the stream resources (e.g., v3io stream or kafka  topics).
+        :param access_key:                  Model monitoring access key, relevant only for V3IO stream.
         """
         await server.api.api.utils._delete_function(
             db_session=db_session,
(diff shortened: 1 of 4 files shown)


tests/system/model_monitoring/test_app.py
```diff
@@ -516,6 +516,58 @@ def _test_model_endpoint_stats(cls, ep_id: str) -> None:
 
     @pytest.mark.parametrize("with_training_set", [True, False])
     def test_app_flow(self, with_training_set: bool) -> None:
+        if not with_training_set:
+            # validate that stream resources were deleted as expected
```
jond01: I think this sub-test suits better `TestModelMonitoringInitialize` rather than `TestMonitoringAppFlow`.
Even if you keep it here, move this to a separate method. Keeping it here makes it way less tractable.

tests/system/model_monitoring/test_app.py
```diff
@@ -516,6 +516,58 @@ def _test_model_endpoint_stats(cls, ep_id: str) -> None:
...
+                    client.object.get(
+                        container="projects",
+                        path=f"{self.project_name}/model-endpoints/stream-histogram-data-drift/serving-state.json",
+                    )
+
+                with pytest.raises(HttpResponseError):
+                    client.object.get(
+                        container="projects",
+                        path=f"{self.project_name}/model-endpoints/stream-evidently-app-test-v2/serving-state.json",
+                    )
```
jond01: How about a for loop?

tests/system/model_monitoring/test_app.py
jond01: In addition, you test here the proper deletion of the previous test, which is indirect and not very indicative for a single run.
This check should move to:
https://github.com/mlrun/mlrun/blob/276847ec3ed186ec5fefce4294224cbf1b918473/tests/system/base.py#L153-L154

server/api/crud/model_monitoring/deployment.py
```diff
@@ -745,8 +745,7 @@ async def disable_model_monitoring(
                     self.project,
                     function_name,
                     self.auth_info,
-                    delete_v3io_stream=not mlrun.mlconf.is_ce_mode()
-                    and function_name
+                    delete_app_stream_resources=True
                     not in [
                         mm_constants.MonitoringFunctionNames.STREAM,
                         mm_constants.MonitoringFunctionNames.APPLICATION_CONTROLLER,
```
jond01: I don't understand the new condition - `True not in [...]`? 🤔
I guess you forgot to delete L749-L752.

server/api/crud/model_monitoring/deployment.py
```diff
@@ -869,10 +868,27 @@ async def delete_monitoring_function(
             auth_info=auth_info,
             background_task_name=background_task_name,
         )
-        if delete_v3io_stream:
-            import v3io.dataplane
-            import v3io.dataplane.response
+        if delete_app_stream_resources:
+            MonitoringDeployment._delete_model_monitoring_stream_resources(
+                project=project,
```
jond01, line 873: You're inside this class - I would make `_delete_model_monitoring_stream_resources` a standard method and won't pass `project` from method to method, it's in the "self" object.

Edit - after a second thought, I understand this design comes from the background task process.

server/api/crud/model_monitoring/deployment.py
```diff
@@ -837,7 +836,7 @@ def _create_monitoring_function_deletion_background_task(
             function_name,
             auth_info,
             background_task_name,
-            delete_v3io_stream,
+            delete_app_stream_resources,
```
jond01, line 839: I would make `delete_monitoring_function` keyword argument only, as it has a lot of arguments and it's too easy to mess their order.

server/api/crud/model_monitoring/deployment.py
```diff
@@ -891,10 +907,19 @@ async def delete_monitoring_function(
...
-            v3io_client = v3io.dataplane.Client(endpoint=mlrun.mlconf.v3io_api)
-            stream_paths = server.api.crud.model_monitoring.get_stream_path(
-                project=project, function_name=function_name
+            stream_paths.extend(
+                server.api.crud.model_monitoring.get_stream_path(
+                    project=project, function_name=function_name
+                )
             )
+
+        if stream_paths[0].startswith("v3io"):
```
jond01: It's not trivial to me why there's at least one item in `stream_paths`.
Can't we reach this point in the project deletion process without any stream?

Moreover, can't we have some edge cases in which there are both V3IO and Kafka streams? E.g. after an upgrade and setting new Kafka credentials.

server/api/crud/model_monitoring/deployment.py
```diff
@@ -906,12 +931,40 @@ async def delete_monitoring_function(
...
                     )
+                    logger.debug("Deleted v3io stream", stream_path=stream_path)
                 except v3io.dataplane.response.HttpResponseError as e:
                     logger.warning(
-                        f"Can't delete {function_name}'s stream",
+                        "Can't delete v3io stream",
                         stream_path=stream_path,
                         error=e,
                     )
+        elif stream_paths[0].startswith("kafka://"):
```
jond01, line 945: What about the `else` case? Maybe add a log for unexpected behavior.

server/api/crud/model_monitoring/deployment.py
```diff
@@ -906,12 +931,40 @@ async def delete_monitoring_function(
...
+            topics.append(topic)
+
+            for stream_path in stream_paths[1:]:
+                topic, _ = mlrun.datastore.utils.parse_kafka_url(url=stream_path)
+                topics.append(topic)
+
+            try:
+                kafka_client = kafka.KafkaAdminClient(
+                    bootstrap_servers=brokers,
+                    client_id=f"{project}",
```
jond01: This f-string is unnecessary:
```suggestion
                    client_id=project,
```

server/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -547,6 +554,58 @@ def delete_model_endpoints_resources(project_name: str):
...
+            db_session=db_session,
+            model_monitoring_applications=model_monitoring_applications,
+            stream_paths=stream_paths,
+        )
+
+    @staticmethod
+    def _delete_model_monitoring_stream_resources(
+        project_name: str,
+        db_session: sqlalchemy.orm.Session,
+        model_monitoring_applications: typing.Optional[list[str]] = None,
```
jond01: This type hint and default `None` value is misleading - see below

server/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -547,6 +554,58 @@ def delete_model_endpoints_resources(project_name: str):
...
+            model_monitoring_access_key = (
+                server.api.api.endpoints.nuclio.process_model_monitoring_secret(
+                    db_session,
+                    project_name,
+                    mlrun.common.schemas.model_monitoring.ProjectSecretKeys.ACCESS_KEY,
+                )
+            )
+
+        # Add the writer and monitoring stream to the application streams list
+        model_monitoring_applications.append(
```
jond01, line 598: If `model_monitoring_applications` is `None`, you can't call `append` on it

server/api/crud/model_monitoring/deployment.py
```diff
@@ -906,12 +931,40 @@ async def delete_monitoring_function(
...
+                )
+                kafka_client.delete_topics(topics)
+                logger.debug("Deleted kafka topics", topics=topics)
+            except kafka.errors.TopicAuthorizationFailedError as e:
+                logger.warning(
+                    "Can't delete kafka topics",
+                    topics=topics,
+                    error=e,
```
jond01: Use here our `err_to_str` utility

server/api/crud/model_monitoring/deployment.py
```diff
@@ -906,12 +931,40 @@ async def delete_monitoring_function(
                     v3io_client.stream.delete(
                         container, stream_path, access_key=access_key
                     )
+                    logger.debug("Deleted v3io stream", stream_path=stream_path)
                 except v3io.dataplane.response.HttpResponseError as e:
                     logger.warning(
-                        f"Can't delete {function_name}'s stream",
+                        "Can't delete v3io stream",
                         stream_path=stream_path,
                         error=e,
```
jond01: Same as prev. comment

server/api/crud/model_monitoring/deployment.py
```diff
@@ -891,10 +907,19 @@ async def delete_monitoring_function(
...
+            stream_paths.extend(
+                server.api.crud.model_monitoring.get_stream_path(
+                    project=project, function_name=function_name
+                )
             )
+
+        if stream_paths[0].startswith("v3io"):
+            # Delete V3IO stream
+            import v3io.dataplane
+            import v3io.dataplane.response
```
jond01, line 923: IMO top-level imports are better here, same for `import kafka`.

jond01 (changes requested):
Overall looks good, I had some comments and questions :)

tests/system/model_monitoring/test_app.py
Eyal-Danieli: yes but subtest such as `TestModelMonitoringInitialize` validates the `disable_model_monitoring_api()`, and here anyway we call the delete project API so why not use it? I don't think it's necessary to create a new test in which we just deploy everything (adding 5-10 min to system tests) and then call the delete project API. 
I'll try to use the `custom_teardown()` as you suggested.

server/api/crud/model_monitoring/deployment.py
Eyal-Danieli: good catch, actually no need to delete `L749-L752` because we don't want to delete the mm-stream resources on disable_model_monitoring API.

server/api/crud/model_monitoring/deployment.py
Eyal-Danieli: In general it's not the case at the moment but to avoid possible future conflicts I'll check if stream_paths exist. As for your second question right now we don't support multiple stream services under the same project.

server/api/crud/model_monitoring/model_endpoints.py
Eyal-Danieli: good catch

server/api/crud/model_monitoring/deployment.py
Eyal-Danieli: but the whole idea is to avoid importing unnecessary dependencies. There is no reason to import the kafka library if you are using v3io.

server/api/crud/model_monitoring/deployment.py
```diff
@@ -891,10 +909,23 @@ async def delete_monitoring_function(
...
+                server.api.crud.model_monitoring.get_stream_path(
+                    project=project, function_name=function_name
+                )
             )
+
+        if not stream_paths:
+            # No stream paths to delete
+            pass
+
+        elif stream_paths[0].startswith("v3io"):
```
jond01: The block in L918-L920 has no effect, did you intend to return?

tests/system/model_monitoring/test_app.py
```diff
@@ -386,6 +386,58 @@ def custom_setup_class(cls) -> None:
...
+                        container="projects",
+                        path=f"{self.project_name}/model-endpoints/stream-{func}/serving-state.json",
+                    )
+
+            # validate that the monitoring stream was deleted
+            with pytest.raises(HttpResponseError):
+                client.object.get(
+                    container="projects",
+                    path=f"{self.project_name}/model-endpoints/stream/serving-state.json",
+                )
```
jond01, line 421: I would even unify these two checks

server/api/crud/model_monitoring/model_endpoints.py
jond01: I don't see any change here

server/api/crud/model_monitoring/model_endpoints.py
jond01: No change

tests/system/model_monitoring/test_app.py
Eyal-Danieli: the stream of the "stream pod" is represents by empty string so it's kind of an overkill to put it in the same pattern. If we will do so, we will need to rename it afterwards in both v3io (`"{self.project_name}/model-endpoints/stream-/serving-state.json"` -> `{self.project_name}/model-endpoints/stream/serving-state.json"`) and kafka (`monitoring_stream_{self.project_name}_` -> `monitoring_stream_{self.project_name}`). 
Therefore, I prefer to keep it separately.

server/api/crud/model_monitoring/deployment.py
```diff
@@ -745,12 +745,13 @@ async def disable_model_monitoring(
...
-                    delete_v3io_stream=not mlrun.mlconf.is_ce_mode()
-                    and function_name
+                    delete_app_stream_resources=True
+                    if function_name
                     not in [
                         mm_constants.MonitoringFunctionNames.STREAM,
                         mm_constants.MonitoringFunctionNames.APPLICATION_CONTROLLER,
-                    ],
+                    ]
+                    else False,
```
jond01: Why not just `function_name not in [...]`?

jond01 (approved):
Awesome!

PR #5915, opened 2024-07-07
[Artifacts] Support deleting a single iteration artifact

TomerShor:
We do not support deleting an artifact with a specific iteration, although we do have a DELETE button in the UI for each iteration artifact.

In this PR we add support for deleting an artifact with a specific iteration, by adding the `iter` query param to the endpoing, and the `iter` argument to the sdk functions.

Example of a delete artifact URL with the supported query params:
```
DELETE /api/v2/projects/<project>/artifacts/<key>?tag=<tag>&tree=<producer_id>&object-uid=<uid>&iter=<iter>
```

Additionaly, I have improved the query logic in `_delete_tagged_object` to join on the tag in a single flow instead of having 2 different flows for with/without tag.

https://iguazio.atlassian.net/browse/ML-6905


server/api/db/sqldb/db.py
```diff
@@ -4151,28 +4152,32 @@ def _delete_tagged_object(
...
-            object_id = object_record.id
-        elif tag and tag != "*":
-            tag_record = self._query(
-                session, cls.Tag, project=project, name=tag, obj_name=obj_name
-            ).one_or_none()
-            if tag_record is None:
-                return None, None
-            object_id = tag_record.obj_id
+
+        # try to find the object by given arguments
```
theSaarco: I'm concerned since you changed the logic - before this change, if `uid` was given, then it was used to pin-point the object, regardless of tags. Now, you do this join, so it's possible that `uid` and `tag` are passed that are non compatible, resulting in no query results while before it would return the record with the given `uid`.
I'm not saying your logic is not better, I'm just thinking that it may change the results of some client logic, bad as it may be.

server/api/db/sqldb/db.py
```diff
@@ -4151,28 +4152,32 @@ def _delete_tagged_object(
...
+            cls,
+            project=project,
+            uid=uid,
+            name=name,
+            key=key,
+            **kwargs,
+        )
+
+        # join on tags if given
+        if tag and tag != "*":
```
theSaarco: And another thing here - conceptually, if someone asks for `*` tag it may mean that we're looking for all tagged objects (which is strange anyway, since we're looking a `one_or_none` so just a single object) . Conceptually, I would say that a `*` tag should result in a join with tags table, but without filtering on tag name, thus returning all tagged results. 
Have to admit I didn't really check where a `*` tag may come from, and if we accept it at all, just a concern I have based on looking specifically at this code.

server/api/db/sqldb/db.py
```diff
@@ -4151,28 +4152,32 @@ def _delete_tagged_object(
...
+
+        if object_record is None:
+            # object not found, nothing to delete
+            return None, None
+
+        # get the object id from the object record
+        object_id = object_record.id
 
         if object_id:
             if not commit:
```
theSaarco, line 4236: Looking at current code I see that there's a code block a few lines below here (line 4193 now) that handles the case where there isn't `object_id` - I think this is exactly what would happen if I don't use a `uid` and pass `*` as tag. In that case it just deletes all references to the object.
I believe your new code, which changed the logic of the function would never get to that situation, and therefore there's an issue there.

theSaarco (changes requested):
Good change, but I'm concerned with the logic changed you did in the DB code. See my comments to follow my messy thought process...

server/api/db/sqldb/db.py
TomerShor: We do have a validation on this specific case at the start of the function:
```python
        if tag and uid:
                "Both uid and tag specified when deleting an object."
            )
```

server/api/db/sqldb/db.py
TomerShor: Yes I see that now.
I think I will follow what is written in the comment right before deleting all artifacts with the same key:
```
# If we got here, neither tag nor uid were provided - delete all references by name.
```
And will just check that as a specific case.

server/api/db/sqldb/db.py
theSaarco: yeah, my thought process started here but I think the main concern is the last comment which talks about a completely different scenario.

server/api/db/sqldb/db.py
TomerShor: I suppose previously if no uid was passed and `tag=="*"` then we would fall to the last part that deletes all objects with the given key.

> Conceptually, I would say that a * tag should result in a join with tags table, but without filtering on tag name, thus returning all tagged results.

^ While this sounds legit, it does change the behavior

liranbg (approved):

theSaarco (approved):
LGTM 👍

PR #5916, opened 2024-07-07
[Config] Config mlrun with internal python application runtime base image

moranbental:
The application runtime base image is Python 3.9. To enable dark sites, it should be taken from the data node registry.
To allow this, I created a new environment variable `vendor_images_registry`, which states that if the image does not begin with `/mlrun`, we will enrich it with the registry defined in `vendor_images_registry`. 
If the environment variable is empty, the image will be pulled from Dockerhub (index.docker.io). 

https://iguazio.atlassian.net/browse/ML-7014


tests/utils/test_helpers.py
```diff
@@ -579,6 +583,7 @@ def test_enrich_image(case):
     default_images_to_enrich_registry = config.images_to_enrich_registry
     config.images_tag = case.get("images_tag", "0.5.2-unstable-adsf76s")
     config.images_registry = case.get("images_registry", "ghcr.io/")
+    config.vendor_images_registry = case.get("vendor_images_registry", "ghcr.io/")
```
alonmr: Change to a different value than `images_registry` to make sure enriched by the correct config

mlrun/config.py
```diff
@@ -64,11 +64,14 @@
     "api_base_version": "v1",
     "version": "",  # will be set to current version
     "images_tag": "",  # tag to use with mlrun images e.g. mlrun/mlrun (defaults to version)
-    "images_registry": "",  # registry to use with mlrun images e.g. quay.io/ (defaults to empty, for dockerhub)
+    # registry to use with mlrun images that start with "mlrun/" e.g. quay.io/ (defaults to empty, for dockerhub)
+    "images_registry": "",
+    # registry for use with mlrun images that don't start with "mlrun/" (defaults to empty, for dockerhub)
```
alonmr:
```suggestion
    # registry to use with non-mlrun images (don't start with "mlrun/") specified in 'images_to_enrich_registry' defaults to empty, for dockerhub)
```

alonmr (commented):
Looks really good! minor comments

quaark (approved):

alonmr (approved):

PR #5917, opened 2024-07-07
[Docs] Tutorials 06 fix monitoring artifacts

daniels290813:
https://iguazio.atlassian.net/issues/ML-6079


assaf758:
1. drift threshoulds
>By default, the threshold is 0.7 but you can modify it through the batch infer process. As seen above, the drift result in this case was ~0.3. Reduce the threshold value to 0.2 in the following run to generate the drifted result

This is not correct in 1.7 - the user can't change the drift thresholds (this parameter is ignored, as can be seen in the run_functions logs:
```
Custom drift threshold arguments are deprecated since version 1.7.0 and have no effect. They will be removed in version 1.9.0.
``` 
2. there is the below string in the notebook - i guess should be fixed/removed
{admonition}

daniels290813:
@assaf758 please review again

assaf758:
> By default, the threshold is 0.7
This is misleading, since the threshold is hard-coded in the histogram-data-drift app.
I suggest:
```
In the histogram-data-drift application, the "Drift detected" threshold is 0.7 and the "Drift suspected" threshold is 0.3
```

BTW, you might want to "watch" https://iguazio.atlassian.net/browse/ML-7022, as this will include some changes in the relevant screens.

daniels290813:
@assaf758 edited again and fixed the usage of drift metrics and status
please review

assaf758 (approved):

PR #5918, opened 2024-07-07
[Model Monitoring] Fix wrong object links introduced in ML-6745 fix

alxtkr77:


mlrun/model_monitoring/stream_processing.py
```diff
@@ -402,7 +402,7 @@ def apply_process_before_parquet():
         # Write the Parquet target file, partitioned by key (endpoint_id) and time.
         def apply_parquet_target():
             graph.add_step(
-                "mlrun.datastore.targets.ParquetTargetStoreyWrapper",
+                "mlrun.model_monitoring.stream_processing.ParquetTargetStoreyWrapper",
```
Eyal-Danieli, line 405: this should be enough because it's the same file 
```suggestion
                "ParquetTargetStoreyWrapper",
```
similar to `TDEngineTargetStoreyWrapper`

assaf758 (approved):

PR #5922, opened 2024-07-07
[Artifacts] Optimize delete artifacts IO calls

alonmr:
https://iguazio.atlassian.net/browse/ML-6595

Two main changes:
1. Utilize `_delete_multi_objects` to delete multiple objects in a single query instead of using a delete query per artifact/tag.
2. Do not load the full artifacts into memory when only the key and uid are needed


server/api/db/sqldb/db.py
```diff
@@ -1380,16 +1384,24 @@ def _find_artifacts(
         # then join the outer query on the subquery to select the correct columns of the table.
         subquery = query.subquery()
         outer_query = session.query(ArtifactV2, subquery.c.name)
+        if with_entities:
+            outer_query = outer_query.with_entities(*with_entities, subquery.c.name)
```
TomerShor, line 1388: How does this behave with the entities of the original query: https://github.com/mlrun/mlrun/pull/5922/files#diff-b54d4633dcd22d05679a4270bb86619cc560d09e4b68451f01d66969495a58acR1333-R1336

server/api/db/sqldb/db.py
```diff
@@ -1293,10 +1294,8 @@ def _find_artifacts(
         most_recent: bool = False,
         attach_tags: bool = False,
         limit: int = None,
-    ) -> typing.Union[
-        list[tuple[ArtifactV2, str]],
-        list[ArtifactV2],
-    ]:
+        with_entities: list[Any] = None,
```
TomerShor, line 1297: what about using this instead of the `attach_tags` (putting the Tags in `with_entities` ?)

server/api/db/sqldb/db.py
```diff
@@ -2287,17 +2301,24 @@ def _delete_multi_objects(
         main_table: mlrun.utils.db.BaseModel,
         related_tables: list[mlrun.utils.db.BaseModel],
         project: str,
-        names: typing.Union[str, list[str]],
-    ):
-        if not names:
+        main_table_identifier: str = "name",
+        main_table_identifier_values: typing.Union[str, list[str]] = None,
+    ) -> int:
```
TomerShor: Add a docstring that explains how to use this and what does it return

TomerShor (changes requested):

server/api/db/sqldb/db.py
alonmr: wdym? this is on the outer query. you have the full artifact by then

TomerShor: I mean what is returned when the query is run? will we always have the artifact & tag?

server/api/db/sqldb/db.py
alonmr: that's not possible because the tags are joined in the inner query so we need to get the tag names from the subquey

server/api/db/sqldb/db.py
alonmr: the 1st entities filter is for the inner query which returns the art id and the tag name, then we join with the artifacts to get the full ones and if supplied the 2nd with_entitites it filters on the full artifacts and tag name

server/api/db/sqldb/db.py
```diff
@@ -1380,16 +1384,23 @@ def _find_artifacts(
...
+        results = outer_query.all()
         if not attach_tags:
             # we might have duplicate records due to the tagging mechanism, so we need to deduplicate
-            return list({artifact for artifact, _ in artifacts_and_tags})
+            artifacts = set()
+            for *artifact, _ in results:
+                if with_entities:
+                    artifacts.add(tuple(artifact))
+                else:
+                    artifacts.add(artifact[0])
```
```suggestion
            for *artifact, _ in results:
                artifacts.add(tuple(artifact) if with_entities else artifact[0])
```

server/api/db/sqldb/db.py
```diff
@@ -871,26 +870,28 @@ def del_artifacts(
...
+            artifact_column_identifiers.setdefault(
+                artifact_column_identifier, []
+            ).append(column_value)
+
+        failed_deletions_count = 0
+        for (
+            artifact_column_identifier,
+            column_values,
+        ) in artifact_column_identifiers.items():
+            deletions_count = self._delete_multi_objects(
```
moranbental, line 882: In such case, you delete all artifact tags first, then all artifact objects. 
What happens if you delete all of the artifact tags and then try to delete the first artifact object and it fails? Now, all other artifacts remain untagged. 

Since we intend to remove the artifact data (in the future), we must first delete one artifact and its tags and data before moving on to another artifact (since deletion can fail and we need to stop).

So I suggest deleting the artifact tags and then the artifact object without commit, then moving to another artifact and finally committing everything.

moranbental (changes requested):

server/api/db/sqldb/db.py
alonmr: Good point but it should be a transaction anyway so the tags and artifacts should be deleted together (I believe `session.execute` does not auto commit but I'll verify.

> Since we intend to remove the artifact data (in the future), we must first delete one artifact and its tags and data before moving on to another artifact (since deletion can fail and we need to stop).

I'm not sure I understand. You will not know if the deletion failed until you commit so I think either way you have to delete all of the data and then the artifacts. If you want to delete them one by one to be able to abort you have to commit after each artifact.

alonmr: I verified that `session.execute` does not commit. Other issues were discussed offline

moranbental (approved):
Looks very good!

liranbg (approved):

TomerShor (approved):

PR #5923, opened 2024-07-07
[Project] Fix cross project name warning

alonmr:
Look like this:
Screenshot 2024-07-07 at 17 04 59

Instead of looking like this:
Screenshot 2024-07-07 at 17 00 50


GiladShapira94 (commented):
Looks good, one small things i need we need to add information for the user that explain What is the effect of this behavior

roei3000b (approved):

alonmr:
> Looks good, one small things i need we need to add information for the user that explain What is the effect of this behavior

good point 👍 fixed

liranbg (approved):

PR #5924, opened 2024-07-08
[Artifact] Deleting Dataset artifact data only if it includes one file

moranbental:
We want to allow deleting artifact data for dataset artifact only if the data contains one file.

When using log_dataset, the artifact data can be saved in multiple files or in one. This is where we specify how to save the data: https://github.com/mlrun/mlrun/blob/cb29422fe09d67d4c08407e249e9ca0c5a52ae07/mlrun/datastore/targets.py#L552

also create new common function - `is_parquet_file`
https://iguazio.atlassian.net/browse/ML-7042

The error message will be defined by - https://iguazio.atlassian.net/browse/ML-7070


server/api/crud/artifacts.py
```diff
@@ -275,6 +275,12 @@ def _resolve_artifact_size(artifact, auth_info):
         if "spec" in artifact and "inline" in artifact["spec"]:
             validate_inline_artifact_body_size(artifact["spec"]["inline"])
 
+    @staticmethod
+    def _is_single_file(target_path):
+        if target_path:
+            return target_path.endswith(".parquet") or target_path.endswith(".pq")
```
roei3000b: The function name doesn't match to what the function does.
I mean it didn't check if there is single file but also check the extension of this file.

roei3000b (changes requested):
Please add also UT.

server/api/crud/artifacts.py
```diff
@@ -275,6 +275,12 @@ def _resolve_artifact_size(artifact, auth_info):
...
+        return False
```
rokatyy:
```suggestion
       return target_path.endswith((".parquet", ".pq")) if target_path else False
```

mlrun/datastore/targets.py
```diff
@@ -549,9 +549,7 @@ def write_dataframe(
                     os.makedirs(dir, exist_ok=True)
             target_df = df
             partition_cols = None  # single parquet file
-            if not target_path.endswith(".parquet") and not target_path.endswith(
-                ".pq"
-            ):  # directory
+            if not target_path.endswith((".parquet", ".pq")):  # directory
```
rokatyy: Isn't it the same code as in `_is_artifact_data_single_file`? Maybe the method can be relocated to more suitable location so in can be used from both places

server/api/crud/artifacts.py
```diff
@@ -301,15 +305,24 @@ def _delete_artifact_data(
...
                     f"Deleting artifact data kind: {artifact_kind} is currently not supported"
                 )
-            path = artifact["spec"]["target_path"]
+            if artifact_kind == "dataset" and not self._is_artifact_data_single_file(
+                path
+            ):
+                raise mlrun.errors.MLRunNotImplementedServerError(
+                    "Deleting artifact data kind dataset is currently available for a single file"
+                )
+
```
rokatyy:
```suggestion
            ):
                raise mlrun.errors.MLRunNotImplementedServerError(
                    "Deleting artifact data of kind 'dataset' is currently supported for a single file only"
                )

```

tests/api/api/test_artifacts.py
```diff
@@ -294,7 +294,7 @@ def test_fails_deleting_artifact_data_by_artifact_kind(
     artifact_kind, db: Session, unversioned_client: TestClient
 ):
     _create_project(unversioned_client)
-    artifact = artifact_kind(key=KEY, body="123", target_path="dummy-path")
+    artifact = artifact_kind(key=KEY, body="123", target_path="dummy-path/")
```
rokatyy: Why is this "/" needed?

server/api/crud/artifacts.py
moranbental: The error message will be defined by @jillnogold  - https://iguazio.atlassian.net/browse/ML-7070

tests/api/api/test_artifacts.py
moranbental: no need

server/api/crud/artifacts.py
rokatyy: @jillnogold could you please take a look here?

moranbental: @rokatyy She knows that, the ticket is on here.

tests/api/api/test_artifacts.py
```diff
@@ -313,6 +313,41 @@ def test_fails_deleting_artifact_data_by_artifact_kind(
...
 
 
+@pytest.mark.parametrize(
+    "target_path",
+    [
+        "dummy-path.parquet",
+        "dummy-path.pq",
+    ],
+)
+def test_deleting_dataset_artifact_data_includes_one_file(
```
TomerShor, line 323: how about a bad flow as well?

moranbental: Already existed`test_fails_deleting_artifact_data_by_artifact_kind`

mlrun/datastore/targets.py
moranbental: I created new common function - `is_parquet_file`

TomerShor (approved):

rokatyy (approved):

PR #5927, opened 2024-07-08
[GenAI] Align code with controller's api usage and frontend requests

yonishelach:
This pull request is containing 2 commits:
1. Adding all the application code from the original repo.
2. Aligning the code with the controller and the frontend.

In the second commit we have the following updates:
1. Remove all the usage of the direct sql client from code and use the controller's api.
2. Add the application's API in order to receive requests from the frontend.
3. Update `langchain` code to `langchain_community` code.
4. Add schemas from the controller code that were used in the application (temporarily?)


guy1992l (changes requested):
change `langchain_community` to `langchain_openai`

PR #5929, opened 2024-07-09
[Alerts] fix event processing when criteria is defined

Yacouby:
resolves:
https://iguazio.atlassian.net/browse/ML-6829

### DIFF
--- a/server/api/api/endpoints/events.py
+++ b/server/api/api/endpoints/events.py
@@ -61,7 +61,7 @@ async def post_event(
             project=project, name=name, request=request, json=data
         )
 
-    logger.debug("Got event", project=project, name=name)
+    logger.debug("Got event", project=project, name=name, id=event_data.entity.ids[0])
 
     if not server.api.crud.Events().is_valid_event(project, event_data):
         raise HTTPException(status_code=HTTPStatus.BAD_REQUEST.value)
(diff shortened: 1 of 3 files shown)


yanburman (approved):

server/api/crud/alerts.py
```diff
@@ -170,9 +169,10 @@ def process_event(
 
                 if state_obj is None:
                     state_obj = {"events": [event_data.timestamp]}
+                else:
+                    state_obj["events"].append(event_data.timestamp)
```
alonmr: Can be improved a bit
```suggestion
                state_obj = self._states.get(alert.id, None) or {}
                state_obj.setdefault("events", [])
                state_obj["events"].append(event_data.timestamp)
```

server/api/api/endpoints/events.py
```diff
@@ -61,7 +61,7 @@ async def post_event(
             project=project, name=name, request=request, json=data
         )
 
-    logger.debug("Got event", project=project, name=name)
+    logger.debug("Got event", project=project, name=name, id=event_data.entity.ids[0])
```
alonmr, line 64: might be worth to log all ids for the future in case we'll have multiple entities

alonmr (commented):
Looks good minor suggestions

server/api/api/endpoints/events.py
Yacouby: the event will be on a specific id ( or more ), so it will be wrong to log all the ids here.. in the future when we will have multiple entities we will have to add a support for that case and log only the relevant id.

alonmr (approved):

PR #5933, opened 2024-07-09
[Monitoring] Use new db session in each background task when disable function monitoring

rokatyy:
Using the same db session in many Background tasks may lead to race condition, so we will create a new session for each background task.

Jira - https://iguazio.atlassian.net/browse/ML-7095


TomerShor (approved):

alonmr (commented):
good catch, looking at the places that call `disable_model_monitoring` they create a db session that after this change will not be used. Its definitely important that we keep the convention of using a new db session in different threads and we should also not create a db session that will not be used. Perhaps we can add some flag to avoid creating that db session.

liranbg (approved):

davesh0812 (approved):
Nice catch! thanks

PR #5934, opened 2024-07-09
[Model Monitoring] Fix project deletion & catch `MLRunInvalidMMStoreType` when trying to get mm data 

davesh0812:
This pr deals with 3 bug : 
https://iguazio.atlassian.net/browse/ML-7098
https://iguazio.atlassian.net/browse/ML-7100 - when by mistake user set twice the same creds we ignore and dont raise an error
https://iguazio.atlassian.net/browse/ML-7091 - using `verify_project_has_no_model_endpoints` before trying to delete the meps

### DIFF
--- a/server/api/crud/model_monitoring/deployment.py
+++ b/server/api/crud/model_monitoring/deployment.py
@@ -1212,8 +1212,8 @@ class MonitoringDeployment:
         )
 
     def _is_the_same_cred(
-        self, endpoint_store_connection, stream_path, tsdb_connection
-    ):
+        self, endpoint_store_connection: str, stream_path: str, tsdb_connection: str
+    ) -> bool:
         credentials_dict = {
             key: server.api.crud.Secrets().get_project_secret(
                 project=self.project,
@@ -1224,26 +1224,26 @@ class MonitoringDeployment:
             for key in mlrun.common.schemas.model_monitoring.ProjectSecretKeys.mandatory_secrets()
         }
 
-        for key, val in credentials_dict.items():
-            if (
-                key
-                == mlrun.common.schemas.model_monitoring.ProjectSecretKeys.ENDPOINT_STORE_CONNECTION
-            ):
-                if val != endpoint_store_connection:
-                    return False
-            elif (
-                key
-                == mlrun.common.schemas.model_monitoring.ProjectSecretKeys.STREAM_PATH
-            ):
-                val = val if val is not None else "v3io"  # TODO : del in 1.9.0
-                if val != stream_path:
-                    return False
-            elif (
-                key
-                == mlrun.common.schemas.model_monitoring.ProjectSecretKeys.TSDB_CONNECTION
-            ):
-                if val != tsdb_connection:
-                    return False
+        if (
+            credentials_dict[
+                mlrun.common.schemas.model_monitoring.ProjectSecretKeys.ENDPOINT_STORE_CONNECTION
+            ]
+            != endpoint_store_connection
+        ):
+            return False
+        val = credentials_dict[
+            mlrun.common.schemas.model_monitoring.ProjectSecretKeys.STREAM_PATH
+        ]
+        val = val if val is not None else "v3io"  # TODO : del in 1.9.0
+        if val != stream_path:
+            return False
+        if (
+            credentials_dict[
+                mlrun.common.schemas.model_monitoring.ProjectSecretKeys.TSDB_CONNECTION
+            ]
+            == tsdb_connection
+        ):
+            return False
         return True
 
 
--- a/server/api/crud/model_monitoring/deployment.py
+++ b/server/api/crud/model_monitoring/deployment.py
@@ -1122,7 +1122,9 @@ class MonitoringDeployment:
             )
         if endpoint_store_connection:
             if not endpoint_store_connection.startswith(
-                mlrun.common.schemas.model_monitoring.ModelEndpointTargetSchemas.list()
+                tuple(
+                    mlrun.common.schemas.model_monitoring.ModelEndpointTargetSchemas.list()
+                )
             ):
                 raise mlrun.errors.MLRunInvalidMMStoreType(
                     "Currently only MySQL/SQLite connections are supported for non-v3io endpoint store,"
@@ -1231,8 +1233,6 @@ class MonitoringDeployment:
         if old_store != endpoint_store_connection:
             logger.debug(
                 "User provided different endpoint store connection",
-                old_store=old_store,
-                new_store=endpoint_store_connection,
             )
             return False
         old_stream = credentials_dict[
@@ -1244,8 +1244,6 @@ class MonitoringDeployment:
         if old_stream != stream_path:
             logger.debug(
                 "User provided different stream path",
-                old_stream=old_stream,
-                new_stream=stream_path,
             )
             return False
         old_tsdb = credentials_dict[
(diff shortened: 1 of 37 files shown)


server/api/crud/model_monitoring/deployment.py
```diff
@@ -1203,6 +1211,41 @@ def set_credentials(
...
+            ):
+                val = val if val is not None else "v3io"  # TODO : del in 1.9.0
+                if val != stream_path:
+                    return False
+            elif (
+                key
+                == mlrun.common.schemas.model_monitoring.ProjectSecretKeys.TSDB_CONNECTION
+            ):
+                if val != tsdb_connection:
+                    return False
```
gtopper: Can be simplified. Something like this:

```suggestion
            for key in mm_constants.ProjectSecretKeys.mandatory_secrets()
        }

        if credentials_dict[mm_constants.ProjectSecretKeys.ENDPOINT_STORE_CONNECTION] != endpoint_store_connection:
            return False
        
        val = credentials_dict[mm_constants.ProjectSecretKeys.STREAM_PATH]
        val = val if val is not None else "v3io"  # TODO: remove in 1.9.0
        if val != stream_path:
            return False
        
        if credentials_dict[mm_constants.ProjectSecretKeys.TSDB_CONNECTION] != tsdb_connection:
            return False
```

davesh0812: Not sure it really simplified but ok

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1115,15 +1123,15 @@ def set_credentials(
...
-                raise mlrun.errors.MLRunInvalidArgumentError(
+                raise mlrun.errors.MLRunInvalidMMStoreType(
                     "Currently only MySQL/SQLite connections are supported for non-v3io endpoint store,"
                     "please provide a full URL (e.g. mysql+pymysql://<username>:<password>@<host>:<port>/<db_name>)"
                 )
             secrets_dict[
                 mlrun.common.schemas.model_monitoring.ProjectSecretKeys.ENDPOINT_STORE_CONNECTION
             ] = endpoint_store_connection
         else:
-            raise mlrun.errors.MLRunInvalidArgumentError(
```
Eyal-Danieli, line 1126: just a recommendation you can replace lines L1121-L1125
`            if (
                endpoint_store_connection != "v3io"
                and not endpoint_store_connection.startswith("mysql")
                and not endpoint_store_connection.startswith("sqlite")
            ):`

with a simple 

`if not endpoint_store_connection.startswith(("v3io", "mysql", "sqlite")):`

Moreover we should consider having a list that includes all of the current supported store schemas.

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1115,15 +1123,15 @@ def set_credentials(
                 and not endpoint_store_connection.startswith("mysql")
                 and not endpoint_store_connection.startswith("sqlite")
             ):
-                raise mlrun.errors.MLRunInvalidArgumentError(
+                raise mlrun.errors.MLRunInvalidMMStoreType(
```
Eyal-Danieli, line 1129: Cool, just add the `raise mlrun.errors.MLRunInvalidMMStoreType` to docstring

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1203,6 +1211,41 @@ def set_credentials(
...
+            for key in mlrun.common.schemas.model_monitoring.ProjectSecretKeys.mandatory_secrets()
+        }
+
+        if (
+            credentials_dict[
+                mlrun.common.schemas.model_monitoring.ProjectSecretKeys.ENDPOINT_STORE_CONNECTION
+            ]
+            != endpoint_store_connection
+        ):
+            return False
```
Eyal-Danieli, line 1237: another recommendation - add a log before the `False` just to mention the specific secret that has a conflict. It will be useful for debugging as well (e.g. understanding that there is an issue with the store connection secret).

Eyal-Danieli (commented):
nice, few comments

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1110,20 +1121,18 @@ def set_credentials(
                 or _default_secrets_v3io
             )
         if endpoint_store_connection:
-            if (
-                endpoint_store_connection != "v3io"
-                and not endpoint_store_connection.startswith("mysql")
-                and not endpoint_store_connection.startswith("sqlite")
+            if not endpoint_store_connection.startswith(
+                mlrun.common.schemas.model_monitoring.ModelEndpointTargetSchemas.list()
```
Eyal-Danieli: `startswith` only works with str or tuple, not list
```suggestion
                tuple(mlrun.common.schemas.model_monitoring.ModelEndpointTargetSchemas.list())
```

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1203,6 +1212,54 @@ def set_credentials(
...
+
+        old_store = credentials_dict[
+            mlrun.common.schemas.model_monitoring.ProjectSecretKeys.ENDPOINT_STORE_CONNECTION
+        ]
+        if old_store != endpoint_store_connection:
+            logger.debug(
+                "User provided different endpoint store connection",
+                old_store=old_store,
+                new_store=endpoint_store_connection,
+            )
```
Eyal-Danieli: I think that just keeping the message without printing the actual creds is enough (same for the other logs)
```suggestion
        if old_store != endpoint_store_connection:
            logger.debug(
                "User provided different endpoint store connection",
            )
```

Eyal-Danieli (approved):
esh, just 2 tiny comments

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1086,11 +1086,22 @@ def set_credentials(
                                              for example taosws://<username>:<password>@<host>:<port>.
         :param _default_secrets_v3io:     Optional parameter for the upgrade process in which the v3io default secret
                                           key is set.
+        :raise MLRunConflictError:        If the credentials are already set for the project and the user
+                                          provided new creds.
```
assaf758:
```suggestion
                                          provided different creds.
```

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1086,11 +1086,22 @@ def set_credentials(
...
         """
         try:
             self.check_if_credentials_are_set(only_project_secrets=True)
+            if self._is_the_same_cred(
+                endpoint_store_connection, stream_path, tsdb_connection
+            ):
+                logger.warning(
+                    "The same credentials are already set for the project",
+                    project=self.project,
+                )
```
assaf758: Why use `warning` severity?
If anything, I would use `info` severity - but do you think we need any log?

davesh0812: I agree the severity is wrong, maybe I can change it to `debug` because mostly it is for debugging.. I put it there because I think we need to understand somehow that the api didn't really changed anything.

assaf758: `debug` is good as well (assuming that's the default severity). Maybe just to clarify change to
```suggestion
                logger.debug(
                    "The same credentials are already set for the project - aborting with no error",
                    project=self.project,
                )
```

assaf758 (approved):

PR #5937, opened 2024-07-09
[KFP] Add node selector enrichment from project to kfp pods

yaelgen:
resolves: [ML-7107](https://iguazio.atlassian.net/browse/ML-7107) 

[ML-7107]: https://iguazio.atlassian.net/browse/ML-7107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/ops.py
+++ b/pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/ops.py
@@ -30,7 +30,8 @@ import mlrun
 import mlrun.common.constants as mlrun_constants
 import mlrun.common.runtimes.constants
 from mlrun.config import config
-from mlrun.utils import get_in
+from mlrun.utils import get_in, logger
+from mlrun.utils.helpers import merge_with_precedence, to_non_empty_values_dict
 
 # Disable the warning about reusing components
 dsl.ContainerOp._DISABLE_REUSABLE_COMPONENT_WARNING = True
@@ -245,8 +246,10 @@ def add_function_node_selection_attributes(
     function, container_op: dsl.ContainerOp
 ) -> dsl.ContainerOp:
     if not mlrun.runtimes.RuntimeKinds.is_local_runtime(function.kind):
-        if getattr(function.spec, "node_selector"):
-            container_op.node_selector = function.spec.node_selector
+        enriched_node_selector = _enrich_node_selector_from_project(function)
+        logger.info("Enriched !!!!", enriched_node_selector=enriched_node_selector)
+        if getattr(function.spec, "node_selector") or enriched_node_selector:
+            container_op.node_selector = enriched_node_selector
 
         if getattr(function.spec, "tolerations"):
             container_op.tolerations = function.spec.tolerations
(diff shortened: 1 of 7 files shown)


pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/ops.py
```diff
@@ -257,6 +259,22 @@ def add_function_node_selection_attributes(
     return container_op
 
 
+def _enrich_node_selector_from_project(function):
+    function_node_selector = getattr(function.spec, "node_selector") or {}
```
alonmr: why with getattr?

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/ops.py
```diff
@@ -257,6 +259,22 @@ def add_function_node_selection_attributes(
...
+    project = function._get_db().get_project(function.metadata.project)
```
alonmr: I'm contemplating if we should use `get_current_project` instead because I prefer not to bloat  the SDK code with get project requests. On the other hand, we want to enrich with the values from the DB and not with cache or do we 🤔

alonmr (commented):
Looks really good. I do think this also needs to be done for kfp v2

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/ops.py
```diff
@@ -30,7 +30,8 @@
 import mlrun.common.constants as mlrun_constants
 import mlrun.common.runtimes.constants
 from mlrun.config import config
-from mlrun.utils import get_in
+from mlrun.utils import get_in, logger
+from mlrun.utils.helpers import merge_with_precedence, to_non_empty_values_dict
```
moranbental: Use `import X` instead of `from X import Y`  to avoid import loop

moranbental (changes requested):

pipeline-adapters/mlrun-pipelines-kfp-v2/src/mlrun_pipelines/ops.py
```diff
@@ -143,6 +145,24 @@ def add_function_node_selection_attributes(
     return task
 
 
+def _enrich_node_selector_from_project(function):
```
alonmr: can be moved to kfp-common.helpers

alonmr (commented):
Nice! last comemnt

pipeline-adapters/mlrun-pipelines-kfp-v2/src/mlrun_pipelines/ops.py
yaelgen: There is no `import mlrun` in the helpers, and I don't think it's a good idea to add it just for this function. Also, I think it fits better in common.ops anyway, so I have moved it there.

alonmr (approved):

PR #5938, opened 2024-07-10
[CE] Fix `None is not allowed as login value` error in nuclio client

gtopper:
[CEML-173](https://iguazio.atlassian.net/browse/CEML-173)

[CEML-173]: https://iguazio.atlassian.net/browse/CEML-173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/api/utils/clients/async_nuclio.py
+++ b/server/api/utils/clients/async_nuclio.py
@@ -38,7 +38,8 @@ NUCLIO_PROJECT_NAME_HEADER = "X-Nuclio-Project-Name"
 class Client:
     def __init__(self, auth_info: mlrun.common.schemas.AuthInfo):
         self._session = None
-        self._auth = aiohttp.BasicAuth(auth_info.username, auth_info.session)
+        login = auth_info.username
+        self._auth = aiohttp.BasicAuth(login, auth_info.session) if login else None
         self._logger = logger.get_child("nuclio-client")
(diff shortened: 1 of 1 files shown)


server/api/utils/clients/async_nuclio.py
```diff
@@ -38,7 +38,8 @@
 class Client:
     def __init__(self, auth_info: mlrun.common.schemas.AuthInfo):
         self._session = None
-        self._auth = aiohttp.BasicAuth(auth_info.username, auth_info.session)
+        login = auth_info.username
+        self._auth = aiohttp.BasicAuth(login, auth_info.session) if login else None
```
rokatyy, line 42: Why not checking for `auth_info` to be None? 
```suggestion
        self._auth = aiohttp.BasicAuth(auth_info.username, auth_info.session) if auth_info else None
```

gtopper: @rokatyy, the issue is that `auth_info.username` is `None`, not that `auth_info` is `None`.

gtopper: `auth_info` can't be `None` because the server route always creates this object afaict.

rokatyy (approved):

assaf758 (approved):

PR #5945, opened 2024-07-10
[Run] Handling artifact URIs in run outputs

moranbental:
When executing a function remotely and subsequently retrieving its outputs, the outputs are computed after the execution completes, in a separate methods called `outputs` or `output`. 
The `outputs` property returns a dictionary containing result values and artifact URIs.
The `output` method retrieves the value of a specific result or artifact by key.

The issue arises when logging an artifact during remote function execution and then using the `outputs` property to retrieve the artifact URI. The artifact URI returned includes the `latest` tag instead of the intended tag itself. This occurs because when logging artifacts, two tags are always present: the intended tag and `latest`. Therefore, when accessing the `outputs` property and iterating over the artifacts, the `latest` tag overrides the intended tag.

To address this issue,  if there are multiple artifacts for the same key, include only the artifact that does not have the `latest` tag. If no other tag is present, include the "latest" tag as a fallback.
There are cases when status.artifacts does not exist in the run (e.g., when listing runs). This is why we also check the `status.artifact_uris` values.
There are cases when `status.artifact_uris` does not exist in the run (e.g., when running locally).

related pr -  https://github.com/mlrun/mlrun/pull/5897
https://iguazio.atlassian.net/browse/ML-6602
https://iguazio.atlassian.net/browse/ML-7194


mlrun/model.py
```diff
@@ -1518,22 +1518,26 @@ def outputs(self):
         outputs = {}
         self._outputs_wait_for_completion()
         if self.status.results:
-            outputs = {k: v for k, v in self.status.results.items()}
+            outputs.update(self.status.results)
         if self.status.artifacts:
-            for a in self.status.artifacts:
-                key = a["metadata"]["key"]
-                outputs[key] = get_artifact_target(a, self.metadata.project)
+            # Ignore artifact with tag 'latest'
```
liranbg: worth adding the why - "we want to ensure that we return a deterministic tag for the artifact"

mlrun/model.py
```diff
@@ -1551,12 +1555,20 @@ def _outputs_wait_for_completion(
             )
 
     def _artifact(self, key):
-        """return artifact DataItem by key"""
-        if self.status.artifacts:
-            for a in self.status.artifacts:
-                if a["metadata"]["key"] == key:
-                    return a
-        return None
+        """Return artifact DataItem by key."""
```
liranbg: I think it is the callee to "ditch" the returned object if its tag is latest
the purpose of this function is to return an artifact by key, not to ignore "latest" tags

mlrun/model.py
```diff
@@ -1551,12 +1555,20 @@ def _outputs_wait_for_completion(
...
+        if not self.status.artifacts:
+            return None
+
+        # Find the first artifact that matches the given key and tag is not 'latest'
+        # Or None if no such artifact is found
+        return next(
+            (
+                a
+                for a in self.status.artifacts
+                if a["metadata"]["key"] == key and a["metadata"]["tag"] != "latest"
```
liranbg:
```suggestion
                artifact
                for artifact in self.status.artifacts
                if artifact["metadata"]["key"] == key and a["metadata"]["tag"] != "latest"
```
[coding convention - 5](https://github.com/mlrun/mlrun/blob/development/CONTRIBUTING.md#python-code-conventions)

liranbg (changes requested):
Missing some unittests coverage and some suggested fixes

mlrun/model.py
```diff
@@ -1514,26 +1524,45 @@ def ui_url(self) -> str:
...
+            # Exclude artifacts tagged as 'latest' to ensure a deterministic tag is returned for the artifact
+            outputs.update(
+                {
+                    artifact["metadata"]["key"]: get_artifact_target(
+                        artifact, self.metadata.project
+                    )
+                    for artifact in self.status.artifacts
+                    if artifact["metadata"].get("tag", None) != "latest"
+                }
+            )
```
TomerShor: But what if we log artifacts without an explicit tag?
They will automatically be tagged as `latest`, and will be excluded in the output.

alonmr (commented):
I'm also not sure that filtering latest is the correct solution here. Please check if you can use `run.status.artifact_uris` to get the correct artifact.

TomerShor (approved):
LGTM

mlrun/model.py
```diff
@@ -1494,14 +1494,33 @@ def error(self) -> str:
...
         self._outputs_wait_for_completion()
+
+        # Check results first
         if self.status.results and key in self.status.results:
-            return self.status.results.get(key)
+            return self.status.results[key]
+
+        # Check artifact URIs next
+        if self.status.artifact_uris and key in self.status.artifact_uris:
+            return self.status.artifact_uris[key]
```
alonmr: This is not BC. You need to get the artifact. Use the store manager to get it

mlrun/model.py
```diff
@@ -1514,26 +1533,49 @@ def ui_url(self) -> str:
...
+
+        # Add artifact URIs if available
+        if self.status.artifact_uris:
+            outputs.update(self.status.artifact_uris)
+
+        # Add artifacts with targets if available
+        elif self.status.artifacts:
+            for artifact in self.status.artifacts:
+                key = artifact["metadata"]["key"]
+                outputs[key] = get_artifact_target(artifact, self.metadata.project)
```
alonmr: same here. you need to get the artifact target path per key (it might be duplicated with the artifact uris)

alonmr (changes requested):
Some of the changes are not backwards compatible see comments below. This is not only problematic due to BC but it will also cause piplines to fail since a step may be dependent on a previous step outputs and it expects an artifact not an artifact URI.

moranbental:
After speaking with @alonmr  offline, there are some issues that need to be addressed:
1. There are cases when `artifact_uri ` does not exist in the run (e.g., when running locally).
2. There are cases when `status.artifacts` does not exist in the run (e.g., when listing runs).
3. For backward compatibility, when dealing with `datasets`, `models`, or `artifacts`, we need to return the URI.
4. In other cases, we need to return the `target_path`.
5. Ignore the `latest` tag if another tag exists.


For that:
1. Give precedence to `status.artifacts` instead of `status.artifact_uri`.
2. Filter by key in the output at the end (to ignore `latest`).
3. Add a system test for running remotely.
4. Add unit tests for:
      a. Tag not exists
      b. Tag = `latest`
      c. Tag = `v1`

mlrun/model.py
```diff
@@ -1494,14 +1494,34 @@ def error(self) -> str:
...
-    def output(self, key):
-        """return the value of a specific result or artifact by key"""
+    def output(self, key: str):
+        """
+        Return the value of a specific result or artifact by key.
+
+        This method waits for the outputs to complete and retrieves the value corresponding to the provided key.
+        If the key exists in the results, it returns the corresponding result value.
+        If not found in results, it attempts to find and return the artifact URI corresponding to the key.
+        If no artifact or result is found for the key, returns None.
```
TomerShor: The order changed a bit, we should align the comment

mlrun/model.py
```diff
@@ -1514,26 +1534,54 @@ def ui_url(self) -> str:
...
+        # There are cases when `status.artifact_uri` does not exist in the run (e.g., when running locally).
         if self.status.artifacts:
-            for a in self.status.artifacts:
-                key = a["metadata"]["key"]
-                outputs[key] = get_artifact_target(a, self.metadata.project)
+            for artifact in self.status.artifacts:
+                key = artifact["metadata"]["key"]
+                tag = artifact["metadata"].get("tag")
+                if (tag != "latest") or (key not in outputs):
+                    outputs[key] = get_artifact_target(artifact, self.metadata.project)
```
TomerShor: What about a case where in the same run I log the same artifact with multiple tags?
I will only have one of the tags in the output.

Like this for example:
```python
def handler(context):
    # do something first
    context.log_artifact("some-key", tag="v1")
    # do something else
    context.log_artifact("some-key", tag="v2")
    # do something else again
    context.log_artifact("some-key", tag="v3")
```
It might be an edge case but nothing stops me from doing it.

moranbental: @TomerShor Nice catch.
In this method, the `outputs["some-key"]` will include the artifact with tag `v3`. However, the problem is with the `output` function, which returns `v1` because it returns the first artifact it finds. 
This issue only occurs in the remote function, where `status.artifacts` includes 4 artifacts, but in the local function, `status.artifacts` includes only the v3 artifact.

moranbental: I updated the function and added unit and system tests for it

mlrun/model.py
```diff
@@ -1547,12 +1596,39 @@ def _outputs_wait_for_completion(
...
+
+        # Filter out artifacts with 'latest' tag
+        non_latest_artifacts = [
+            artifact
+            for artifact in matching_artifacts
+            if artifact["metadata"].get("tag") != "latest"
+        ]
+
+        # Return the last non-'latest' artifact if available, otherwise return the last artifact
+        return (non_latest_artifacts or matching_artifacts)[-1]
```
liranbg, line 1639: if both are empty list, you will experience `list index out of range` - while it might not be a real-user case - I suggest to defend the code and return the expected value in case list is empty.

liranbg (changes requested):

mlrun/model.py
moranbental: It cannot happen because I am already checking by:
```
        if not matching_artifacts:
            return None
```
that `matching_artifacts` is not an empty list

TomerShor: If there are multiple tags, why do you return the last one?
and how are they sorted?

moranbental: It is not always the last one
In the case of only one tag, `status.artifacts` includes [v1, latest]. In that case, we want to return v1.
In the case of multiple tags, `status.artifacts` includes [v1, latest, v2, v3]. In that case, we need to return the last one (v3).

liranbg: > It is not always the last one In the case of only one tag, `status.artifacts` includes [v1, latest]. In that case, we want to return v1. In the case of multiple tags, `status.artifacts` includes [v1, latest, v2, v3]. In that case, we need to return the last one (v3).

comment the heuristic on the function docs. what ensures you the list is sorted by creation date ("aka recently created is always last")?

mlrun/model.py
```diff
@@ -1490,14 +1490,35 @@ def error(self) -> str:
...
 
-    def output(self, key):
-        """return the value of a specific result or artifact by key"""
+    def output(self, key: str):
+        """
+        Return the value of a specific result or artifact by key.
+
+        This method waits for the outputs to complete and retrieves the value corresponding to the provided key.
+        If the key exists in the results, it returns the corresponding result value.
+        If not found in results, it attempts to fetch the artifact by key.
```
```suggestion
        If not found in results, it attempts to fetch the artifact by key (cached in the run status).
```

mlrun/model.py
```diff
@@ -1490,14 +1490,35 @@ def error(self) -> str:
...
+        # Fetch the artifact for the key if available
+        # There are cases when `status.artifact_uri` does not exist in the run (e.g., when running locally).
         artifact = self._artifact(key)
         if artifact:
             return get_artifact_target(artifact, self.metadata.project)
+
+        # There are cases when `status.artifacts` does not exist in the run (e.g., when listing runs).
+        if self.status.artifact_uris and key in self.status.artifact_uris:
+            return self.status.artifact_uris[key]
+
```
alonmr: The logic is good but I think its better to have one concentrated comment instead of giving bits of information every few lines. Instead of these comments, write something along the lines of:
```
Artifacts are usually cached in the run object under `status.artifacts`. However, the artifacts are not stored in the DB as part of the run. The server may enrich the run with the artifacts or provide `status.artifact_uris` instead. See mlrun.common.formatters.run.RunFormat.

```

mlrun/model.py
alonmr: why not return the latest if you have multiple tags? That way you get the last created one

tests/run/test_run.py
```diff
@@ -342,13 +342,44 @@ def test_get_or_create_ctx_run_kind_exists_in_mlrun_exec_config(
...
+
+def test_verify_tag_in_output_for_relogged_artifact():
+    project = mlrun.get_or_create_project("dummy-project")
+    project.set_function(
+        func=function_path,
+        handler="log_artifact_many_tags",
+        name="test",
+        image="mlrun/mlrun",
+    )
+    run = project.run_function("test", local=True)
```
alonmr: Should work as well
```suggestion
def test_verify_tag_in_output_for_relogged_artifact(setup_project):
    run = project.run_function("test", handler="log_artifact_many_tags", local=True)
```

tests/system/api/test_artifacts.py
```diff
@@ -41,3 +44,45 @@ def test_import_artifact(self):
                 temp_file.name, new_key=f"imported_artifact_key_{uuid.uuid4()}"
             )
         assert artifact.to_dataitem().get().decode() == body
+
+    @pytest.mark.enterprise
```
alonmr: why?

tests/system/api/test_artifacts.py
```diff
@@ -41,3 +44,45 @@ def test_import_artifact(self):
...
+        outputs_uri_from_list_runs = first_run.outputs["file_result"]
+        assert (
+            "v3" in output_uri_from_list_runs
+        ), "Expected 'v3' tag in output_uri_from_list_runs"
+        assert (
+            "v3" in outputs_uri_from_list_runs
+        ), "Expected 'v3' tag in outputs_uri_from_list_runs"
+
+        # run the same function with different tags
+        self.project.set_function(
```
alonmr: you don't need to set the function again as mentioned above

alonmr (commented):
Nice job and coverage! A few comments

mlrun/model.py
moranbental: because this it the bug:
```
func_v1_run = project.run_function("test",params={"tag":"v1"})
df_v1_uri = func_v1_run.outputs["df"]
func_v2_run = project.run_function("test",params={"tag":"v2"})
mlrun.get_dataitem(df_v1_uri)
```

When `df_v1_uri` returns the `latest` artifact, and then the function is run again with v2 (making the second artifact the `latest`), attempting to get the data item from `df_v1_uri` will search for the `latest` artifact instead of v1.

moranbental: I have sorted the artifacts by creation date

liranbg (approved):

mlrun/model.py
```diff
@@ -1620,6 +1618,12 @@ def _artifact(self, key):
         if not matching_artifacts:
             return None
 
+        # Sort matching artifacts by creation date in ascending order.
+        # The last element in the list will be the one created most recently.
+        matching_artifacts.sort(
+            key=lambda artifact: artifact["metadata"].get("created")
```
alonmr: what happens if you compare None values?

moranbental: If the `created` field does not exist in an artifact, that artifact will appear first in the sorted list
But I'll return a default datetime

alonmr (approved):

PR #5948, opened 2024-07-11
[Datastore] Fix handling of start and end time times zones

gtopper:
when they do not match the time zone in the parquet file(s) being read. Raise an error if start time and end time time zones do not match.

[ML-6308](https://iguazio.atlassian.net/browse/ML-6308)

[ML-6308]: https://iguazio.atlassian.net/browse/ML-6308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tomerm-iguazio:
Looking good!

There is one issue that I'm worried about:

What about the representation of a string as start_time or end_time filters? (with or without timezone)

tomerm-iguazio (approved):

PR #5956, opened 2024-07-14
[PipelineAdapters] Deprecate unused rundb function argument

liranbg:
Argument was never used, deprecate it

in addition, fix `as_step` outputs argument as it should be a list (and treated as list in its usages)

### DIFF
--- a/mlrun/runtimes/base.py
+++ b/mlrun/runtimes/base.py
@@ -674,7 +674,7 @@ class BaseRuntime(ModelObj):
         selector="",
         hyper_param_options: HyperParamOptions = None,
         inputs: dict = None,
-        outputs: dict = None,
+        outputs: list = None,
         workdir: str = "",
         artifact_path: str = "",
         image: str = "",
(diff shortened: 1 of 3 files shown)


mlrun/runtimes/base.py
```diff
@@ -674,7 +674,7 @@ def as_step(
         selector="",
         hyper_param_options: HyperParamOptions = None,
         inputs: dict = None,
-        outputs: dict = None,
+        outputs: list = None,
```
TomerShor, line 677: This is a user-facing function, how will this change affect its usage?

TomerShor (approved):
LGTM, just one question

mlrun/runtimes/base.py
liranbg: it won't really change anything as this is type-hinting. if a client passed a dictionary - the execution fails

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -441,9 +445,7 @@ def get_default_reg():
     return ""
 
 
-def format_summary_from_kfp_run(
-    kfp_run, project=None, run_db: "mlrun.db.RunDBInterface" = None
-):
+def format_summary_from_kfp_run(kfp_run, project=None):
```
alonmr, line 448: 😭  this was needed

PR #5959, opened 2024-07-14
[Pipelines] Override PVC mount params only if not given

TomerShor:
In case custom auto mount params are set, but there is an env var `MLRUN_PVC_MOUNT` configured, the custom params will always be overridden by the values in the env var.

Instead, override them only if not set. 

https://iguazio.atlassian.net/browse/CEML-181


liranbg (changes requested):
Two things
1. make the changes on v2 as well
2. bump `setup.py` patch on each adapter

liranbg (approved):

PR #5971, opened 2024-07-15
[Datastore] Fix HDFS datastore and test

gtopper:
[ML-7166](https://iguazio.atlassian.net/browse/ML-7166)

[ML-7166]: https://iguazio.atlassian.net/browse/ML-7166?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/system/feature_store/test_spark_engine.py
```diff
@@ -1444,20 +1444,22 @@ def test_error_is_properly_propagated(self):
...
-                http_port=int(os.getenv("HDFS_HTTP_PORT")),
-            )
+        datastore_profile = DatastoreProfileHdfs(
+            name="my-hdfs",
+            host=os.getenv("HDFS_HOST"),
+            port=int(os.getenv("HDFS_PORT")),
+            http_port=int(os.getenv("HDFS_HTTP_PORT")),
         )
+        register_temporary_client_datastore_profile(datastore_profile)
+        self.project.register_datastore_profile(datastore_profile)
```
gtopper, line 1462: @alxtkr77, it's apparently necessary to call `register_temporary_client_datastore_profile` in addition to `register_datastore_profile`. Is that by design?

tomerm-iguazio (approved):
looks good:)

tests/system/feature_store/test_spark_engine.py
alxtkr77: When running locally (not from pod) the mlrun has no access to secrets, so you have to call register_temporary_client_datastore_profile()

gtopper: If this is really intended usage I think we should probably document it. From what I can tell, you always have to call both before calling a remote `get_offline_features` (e.g. with spark engine). Or we could have `register_datastore_profile` call `register_temporary_client_datastore_profile` internally.

PR #5974, opened 2024-07-15
[Application] Force ssl redirect on API Gateway

TomerShor:
We do not want to allow access to the application's URL via http.
While we still prefix external invocation URLs with `http`, we want to redirect all http requests to https.
This is done using the nginx ingress annotation `force-ssl-redirect`.

Additionally, set the default authentication mode for application runtime to be `None`, as it should be in Open Source MLRun. The default for Iguazio systems will be [set in Provazio](https://github.com/iguazio/provazio/pull/3101).

Resolves https://iguazio.atlassian.net/browse/ML-7108


mlrun/runtimes/nuclio/application/application.py
```diff
@@ -377,6 +377,10 @@ def create_api_gateway(
             ),
         )
 
+        # force ssl redirect so that the application is only accessible via https
+        api_gateway.with_force_ssl_redirect()
```
liranbg: few questions
1. can we opt out?
2. can we enforce it for "production" / "enterprise" deployments?

For an instance, I would not require it for MLRun CE

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -361,13 +365,14 @@ def create_api_gateway(
...
         authentication_creds: tuple[str] = None,
+        ssl_redirect: bool = False,
     ):
         api_gateway = APIGateway(
             APIGatewayMetadata(
                 name=name,
                 namespace=self.metadata.namespace,
                 labels=self.metadata.labels,
-                annotations=self.metadata.annotations,
+                annotations=self.metadata.annotations or {},
```
rokatyy: We already set it to `{}` [here](https://github.com/TomerShor/mlrun/blob/15a82a586d8e680983d73122fcdbeb0c4c7de350/mlrun/runtimes/nuclio/api_gateway.py#L135)

rokatyy (approved):
LGTM

mlrun/runtimes/nuclio/application/application.py
TomerShor: Missed that, fixed.

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -377,6 +382,11 @@ def create_api_gateway(
             ),
         )
 
+        if ssl_redirect or mlrun.mlconf.is_running_on_iguazio():
+            # force ssl redirect so that the application is only accessible via https
+            api_gateway.with_force_ssl_redirect()
```
alonmr: I think this is confusing that the default is false and if in iguazio you set it to true.
If you really want to set the default to false for open source I think it should be in computed in mlrun config and then you can direct the users to check the config instead of looking the the param default value

alonmr (commented):
❤️

mlrun/runtimes/nuclio/application/application.py
TomerShor: Do you mean having a config value / method of wether to enable ssl redirection?
I do however want to allow setting it explicitly per function, with a flag on `deploy`.

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -286,7 +286,8 @@ def deploy(
         :param direct_port_access:      Set True to allow direct port access to the application sidecar
         :param authentication_mode:     API Gateway authentication mode
         :param authentication_creds:    API Gateway authentication credentials as a tuple (username, password)
-        :param ssl_redirect:            Set True to force SSL redirect
+        :param ssl_redirect:            Set True to force SSL redirect, False to disable. Default is determined by
+                                        mlrun configuration
```
alonmr:
```suggestion
        :param ssl_redirect:            Set True to force SSL redirect, False to disable. Defaults to mlrun.mlconf.force_ssl_redirect()
```

mlrun/config.py
```diff
@@ -1040,6 +1040,14 @@ def get_default_function_pod_requirement_resources(
             resource_requirement.pop(gpu)
         return resource_requirement
 
+    def force_ssl_redirect(self):
```
alonmr: wonder if we should name it `force_api_gateway_ssl_redirect` wdyt?

alonmr (commented):
love it! minor suggestions

mlrun/config.py
TomerShor: We can because it is indeed very specific, however it is only used in Application runtime, not for every api gateway by default (this can be enabled using `api_gateway.force_ssl_redirect()` manually).
Do you think we should add this config check on every api gateway creation?

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -285,6 +286,9 @@ def deploy(
         :param direct_port_access:      Set True to allow direct port access to the application sidecar
         :param authentication_mode:     API Gateway authentication mode
         :param authentication_creds:    API Gateway authentication credentials as a tuple (username, password)
+        :param ssl_redirect:            Set True to force SSL redirect, False to disable. Defaults to
+                                        mlrun.mlconf.force_ssl_redirect()
```
alonmr:
```suggestion
                                        mlrun.mlconf. force_api_gateway_ssl_redirect()
```

alonmr (approved):

PR #5978, opened 2024-07-16
[Spark] Remove redundant implementation of `toPandas`

gtopper:
[ML-7153](https://iguazio.atlassian.net/browse/ML-7153)

[ML-7153]: https://iguazio.atlassian.net/browse/ML-7153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/feature_store/retrieval/spark_merger.py
```diff
@@ -24,10 +24,11 @@
 from ...runtimes import RemoteSparkRuntime
 from ...runtimes.sparkjob import Spark3Runtime
 from .base import BaseMerger
-from .conversion import PandasConversionMixin
 
 
 def spark_df_to_pandas(spark_df):
+    from ...data_types.to_pandas import toPandas
```
tomerm-iguazio: I prefer avoiding relative imports

mlrun/feature_store/retrieval/spark_merger.py
```diff
@@ -46,12 +47,12 @@ def spark_df_to_pandas(spark_df):
                 )
                 type_conversion_dict[field.name] = "datetime64[ns]"
 
-        df = PandasConversionMixin.toPandas(spark_df)
+        df = toPandas(spark_df)
```
tomerm-iguazio: It's a bit confusing that some places use the `spark_df_to_pandas` function for special datetime handling, while others use `toPandas` directly without it.

We should decide whether we want to include datetime handling and consistently apply this choice wherever needed.

tomerm-iguazio (changes requested):
Great! I have two comments for you to address. Tnx!

tomerm-iguazio:
I saw another thing that concern me:
in line 115 and 168 in spark.py, I saw that we call the pyspark source function `toPandas` instead of using our own implementation.

Do you think I will be better if we use our implementation also there?

tomerm-iguazio (approved):

PR #5979, opened 2024-07-16
[Model Monitoring] Add force option to set_model_monitoring_credentials

davesh0812:
Add option to set model monitoring credentials after they already exist by `replace_creds` param.
User that using this option will need manually to redeploy:
1. MM infra functions
2. MM functions
3. ModelServer with tracking 

https://iguazio.atlassian.net/browse/ML-7185


assaf758:
Instead of generic/obscure "force" flag I think we should be more explicit with a "replace_creds" flag

mlrun/projects/project.py
```diff
@@ -3244,7 +3246,16 @@ def set_model_monitoring_credentials(
...
             },
+            force=force,
         )
+        if force:
+            logger.info(
+                "Model monitoring credentials were set successfully. "
+                "Please take in mind that if you already had model monitoring functions "
+                "/ model monitoring infra deployed on your project, you will need to redeploy them."
+                "For redeploying the model monitoring infra, please use the `enable_model_monitoring` Api "
+                "and set `rebuild_images=True`"
```
Eyal-Danieli:
```suggestion
                "Model monitoring credentials were set successfully. "
                "Please keep in mind that if you already had model monitoring functions "
                "/ model monitoring infra deployed on your project, you will need to redeploy them."
                "For redeploying the model monitoring infra, please use `enable_model_monitoring` API "
                "and set `rebuild_images=True`"
```

Eyal-Danieli: In addition, I'll add this comment to the docstring as well. I think it's important that the user will understand what `force` does before calling it because it can be a "killing" action that cannot be undone.

server/api/crud/model_monitoring/deployment.py
```diff
@@ -994,6 +985,16 @@ def check_if_credentials_are_set(self, only_project_secrets: bool = False):
             )
             for key in mlrun.common.schemas.model_monitoring.ProjectSecretKeys.mandatory_secrets()
         }
+
+        return credentials_dict
+
+    def check_if_credentials_are_set(self, with_upgrade_case_check: bool = False):
+        """
+        Check if the model monitoring credentials are set. If not, raise an error.
```
Eyal-Danieli, line 993: missing `with_upgrade_case_check` in docstring

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1005,48 +1006,77 @@ def check_if_credentials_are_set(self, only_project_secrets: bool = False):
...
+        """
+        Check and Set the model monitoring credentials for old project that had model monitoring on it before the server
+        upgrade.
+        :return: True if the credentials are set, otherwise False.
+        """
+        # Check if:
+        #   1. There is model endpoints.
+        #   2. Stream pod is on
+        #   3. There is already part of the secrets set.
+        #  if True, set the cred in to the project secret (from exist cred/ from sys config / v3io by default).
```
Eyal-Danieli: Just a recommendation to make it clearer

```suggestion
    def _set_credentials_after_server_upgrade(self) -> bool:
        """
        Check and set the model monitoring credentials for old project that included model monitoring before the server
        upgrade. Will set the credentials only if at least one of the following conditions is met:
        1. There is at least one model endpoint
        2. Model monitoring stream pod is running
        3. Part of the model monitoring credentials are already set

        If True, set the cred in to the project secret (from exist cred/from sys config/v3io by default).

        :return: True if the credentials are set, otherwise False.
        """

```

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1005,48 +1006,77 @@ def check_if_credentials_are_set(self, only_project_secrets: bool = False):
...
+
+        credentials_dict = self._get_monitoring_mandatory_project_secrets()
+        mm_enabled = False
+        store_connector: mlrun.model_monitoring.db.StoreBase = mlrun.model_monitoring.get_store_object(
+            project=self.project,
+            store_connection_string=credentials_dict.get(
+                mlrun.common.schemas.model_monitoring.ProjectSecretKeys.ENDPOINT_STORE_CONNECTION
+            )
+            or "v3io",  # in case they use the default v3io
```
```suggestion
            or "v3io",  # in case the user use the default v3io
```

Eyal-Danieli (commented):
LGTM, just docs comments

Eyal-Danieli (approved):

PR #5985, opened 2024-07-16
[Chief/Worker] Support Opt Out of Chief Features

quaark:
Support deployment of MLRun Chief while opting out of some of the chief features.
This is to allow us to deploy multiple chiefs that will have different responsibility.
For example:
One chief that accepts requests for schedules and project CRUD, and another for all the background cycles.
This way if the background cycles have an issue they can be restarted without affecting the main responsive chief.


roei3000b (approved):
LGTM :)

mlrun/config.py
```diff
@@ -274,6 +274,16 @@
                 "url": "",
                 "service": "mlrun-api-chief",
                 "port": 8080,
+                "features": {
```
alonmr: suggest to rename to `feature_gates`

alonmr (commented):
Nice! minor suggestion

liranbg (approved):

PR #5986, opened 2024-07-16
[API] Paginate list pods/crds in periodic monitoring and cleanup

alonmr:
We are seeing that listing pods in cleanup/monitoring cycle is consuming to much memory which causes the API memory consumption to increase over time and eventually kill the chief API container. Paginating the requests to k8s helps reducing the memory peaks which significantly improves the API consumption over time.

https://iguazio.atlassian.net/browse/ML-6317


server/api/utils/singletons/k8s.py
```diff
@@ -125,6 +125,115 @@ def list_pods(self, namespace=None, selector="", states=None):
...
+                    label_selector=selector,
+                    watch=False,
+                    limit=200,
+                )
+            else:
+                pods_list = self.v1api.list_namespaced_pod(
+                    self.resolve_namespace(namespace),
+                    label_selector=selector,
+                    watch=False,
+                    limit=200,
```
liranbg: Make configurable

server/api/utils/singletons/k8s.py
```diff
@@ -125,6 +125,115 @@ def list_pods(self, namespace=None, selector="", states=None):
...
+        _continue = ""
+        logger.debug(
+            "Listing namespaced pods with pagination",
+            selector=selector,
+            namespace=namespace,
+            states=states,
+        )
+        while True:
+            if not _continue:
+                pods_list = self.v1api.list_namespaced_pod(
```
liranbg: DRY

server/api/utils/singletons/k8s.py
```diff
@@ -125,6 +125,115 @@ def list_pods(self, namespace=None, selector="", states=None):
...
+                    watch=False,
+                    limit=200,
+                    _continue=_continue,
+                )
+
+            for item in pods_list.items:
+                if not states or item.status.phase in states:
+                    yield item
+
+            _continue = pods_list._metadata._continue
```
liranbg: No better way to resolve the continue?

liranbg (commented):
Very well. Minor stuff

server/api/utils/singletons/k8s.py
alonmr: metadata has a public getter but continue doesn't
also still checking how this works for crds

server/api/utils/singletons/k8s.py
```diff
@@ -125,6 +125,99 @@ def list_pods(self, namespace=None, selector="", states=None):
                 items.append(i)
         return items
 
+    @raise_for_status_code
+    def list_pods_paginated(self, namespace=None, selector="", states=None):
```
quaark, line 129: Feels to me like a lot of these 2 functions can be generalized into a `list_paginated` method that receives a list method callback, and can be later reused by `list_pods_paginated`, `list_crds_paginated` and any other resource we want.

quaark (commented):
Looks great!
I have a suggestion, I feel like we can save some code dup

server/api/utils/singletons/k8s.py
alonmr: I'm not sure I agree. There are differences with how to parse the result which needs to be part of the callback and it turns out I'm extracting 90% of the method to the callback. If you have some other Idea lets discuss offline but in general I think that the common part is not big enough to make this refactor.

PR #5987, opened 2024-07-16
[Functions] Do not enrich project image if function base_image is set

rokatyy:
We want to avoid the enrichment of project's `default_image` if `base_image` was set for a function.
 
jira - http://iguazio.atlassian.net/browse/ML-6258

### DIFF
--- a/mlrun/projects/pipelines.py
+++ b/mlrun/projects/pipelines.py
@@ -320,7 +320,7 @@ def enrich_function_object(
     setattr(f, "_enriched", True)
 
     # set project default image if defined and function does not have an image specified
-    if project.spec.default_image and not f.spec.image:
+    if project.spec.default_image and not f.spec.image and not f.spec.build.base_image:
         f._enriched_image = True
         f.spec.image = project.spec.default_image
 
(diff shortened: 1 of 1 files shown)


mlrun/projects/pipelines.py
```diff
@@ -320,7 +320,7 @@ def enrich_function_object(
     setattr(f, "_enriched", True)
 
     # set project default image if defined and function does not have an image specified
```
alonmr:
```suggestion
    # set project default image if defined and function does not have an image/base image specified
```

alonmr (approved):

mlrun/projects/pipelines.py
```diff
@@ -319,8 +319,8 @@ def enrich_function_object(
     f.metadata.project = project.metadata.name
     setattr(f, "_enriched", True)
 
-    # set project default image if defined and function does not have an image specified
-    if project.spec.default_image and not f.spec.image:
+    # set project default image if defined and function does not have an image/base image specified
```
liranbg, line 322: mind you please change the comment to the reasons behind?
the comment itself simply perform human translation of the `if` which, is not really helpful.
perhaps after digging enough you could add explanation of what is the reasons behind

rokatyy:
Decided to close the PR as it breaks some flows

PR #5988, opened 2024-07-17
[Project] Enrich delete project background task with model monitoring access key  

Eyal-Danieli:
Fix for a backwards compatibility issue in the delete project API for users that use previous iguazio versions (`<3.5.5`). As part of this API, we delete some of the model monitoring resources that are associated with the related project. If the user use V3IO, then the user's access key is required for some of these resources, such as the stream resources that are now stored in the user's directory. 

To get this access key (which defined as a project secret), the request to zebo should be made before the project is deleted in iguazio. However in previous iguazio versions, the project was deleted before the request was made. In this PR we make sure to retrieve the model monitoring access key **before** the project is deleted in iguazio. 

A fix for https://iguazio.atlassian.net/browse/ML-7141


server/api/api/utils.py
```diff
@@ -1121,6 +1122,16 @@ def get_or_create_project_deletion_background_task(
...
+        # Following the last comment, the model monitoring access key should be retrieved before the project deletion.
+        # This key will be used to delete the model monitoring resources associated with the project.
+        model_monitoring_access_key = (
+            server.api.api.endpoints.nuclio.process_model_monitoring_secret(
+                db_session,
+                project.metadata.name,
+                mlrun.common.schemas.model_monitoring.ProjectSecretKeys.ACCESS_KEY,
+            )
+        )
+
```
quaark: IIUC, this should be outside the if and should always happen (just one less indentation)

quaark (changes requested):
Looks good!
One minor thing

server/api/api/utils.py
Eyal-Danieli: but this access key is only used for v3io access, so I'm not sure if it can break the delete API for CE users in this case. I'll test it with CE before making this change.

quaark: I understand, So yes it needs to be in the if, but not in the else:
```
if (
    server.api.utils.helpers.is_request_from_leader(auth_info.projects_role)
    or mlrun.mlconf.httpdb.projects.leader == "mlrun"
):
    # this means the request either came from zebo (or we are CE) and we are actually about to delete the project.
    # you need the secret in this case and to pass it on to the job. You might need to add another `if igz_version` here
elif igz_version and igz_version < semver.VersionInfo.parse("3.5.5"):
    # In this case the request came client and we are not CE, so we are going to send a request to igz and no deletion is happening here.

quaark (approved):

PR #5992, opened 2024-07-17
[Notifications] Raise PreconditionFailed when secret doesn't exist

yanburman:
In some cases - like when user provided a secret, but didn't actually create it, we get a mystic exception.
Do not explode inside json, but let upper layers handle the error
https://iguazio.atlassian.net/browse/ML-7188

### DIFF
--- a/server/api/api/utils.py
+++ b/server/api/api/utils.py
@@ -358,15 +358,19 @@ def unmask_notification_params_secret(
             "Not running in k8s environment, cannot load notification params secret"
         )
 
-    notification_object.secret_params = json.loads(
-        server.api.crud.Secrets().get_project_secret(
-            project,
-            mlrun.common.schemas.SecretProviderName.kubernetes,
-            secret_key=params_secret,
-            allow_internal_secrets=True,
-            allow_secrets_from_k8s=True,
-        )
+    secret = server.api.crud.Secrets().get_project_secret(
+        project,
+        mlrun.common.schemas.SecretProviderName.kubernetes,
+        secret_key=params_secret,
+        allow_internal_secrets=True,
+        allow_secrets_from_k8s=True,
     )
+
+    if secret is None:
+        # we don't want to provide a message that is too detailed due to security considerations here
+        raise mlrun.errors.MLRunPreconditionFailedError()
+
+    notification_object.secret_params = json.loads(secret)
 
     return notification_object
 


quaark (changes requested):
I honestly don't agree with the resolution of the issue.

> "when user provided a secret, but didn't actually create it"

This should rightfully make the request fail.
The correct fix would be to have an indicative error message.

yanburman:
> I honestly don't agree with the resolution of the issue.
> 
> > "when user provided a secret, but didn't actually create it"
> 
> This should rightfully make the request fail. The correct fix would be to have an indicative error message.

Not having a secret is not necessarily a problem. The values may be present in the regular params

quaark:
> > I honestly don't agree with the resolution of the issue.
> > > "when user provided a secret, but didn't actually create it"
> > 
> > 
> > This should rightfully make the request fail. The correct fix would be to have an indicative error message.
> 
> Not having a secret is not necessarily a problem. The values may be present in the regular params

I don't necessarily agree. At the end of the day we have a defined API, that API defines that if you supply a secret I must replace your params with the content of the secret.
If the secret doesn't exist I can't just best effort and ignore, that might be a mistake of the user (maybe just a typo on the users end) and you'll fail very late in the flow. This is a fast and relevant failure, even if you did supply regular params

server/api/api/utils.py
```diff
@@ -358,16 +358,19 @@ def unmask_notification_params_secret(
...
+    secret = server.api.crud.Secrets().get_project_secret(
+        project,
+        mlrun.common.schemas.SecretProviderName.kubernetes,
+        secret_key=params_secret,
+        allow_internal_secrets=True,
+        allow_secrets_from_k8s=True,
     )
 
+    if secret is None:
+        raise mlrun.errors.MLRunPreconditionFailedError("Provided secret not found")
```
quaark:
```suggestion
        raise mlrun.errors.MLRunPreconditionFailedError(f"Provided secret not found: {params_secret}")
```

quaark (changes requested):
Looks good, better to show the problem value in the error message

server/api/api/utils.py
yanburman: not sure this is a good idea from security POV

quaark: If that's a concern then we might not want an indicative error message at all. All the user needs to do is remember their input to gain info from the message without the specific secret name. So maybe just return an empty BadRequest exception?

yanburman: It doesn't always happen as a result of specific user request, since it fires when we send the notification. So, perhaps same PreconditionFailes, but empty

quaark (approved):

PR #5996, opened 2024-07-17
[API] Time Window Tracker

quaark:
https://iguazio.atlassian.net/browse/ML-7086

A few use cases in MLRun use a sliding time window for querying recent resources from the DB:
- 2 of the use cases used an in memory sliding window, which causes issues when the api is restarted as the most recent value is `None` and therefore the mechanisms query the entire history of the resources.
- The other use case always queries a constant time window subtracted from now, which means it can miss resources during downtime and always queries a window larger than the cycle length.

The solution here combines both ideas with persistent storage:
A DB managed sliding window system (each table row is a different tracked sliding window) which can always get updated. A max window size can be set on the tracker, so if the time passed is larger than the window we clamp down the window to the max window size.


server/api/main.py
```diff
@@ -282,6 +275,18 @@ async def _verify_log_collection_started_on_startup(
...
+        server.api.utils.time_window_tracker.TimeWindowTracker(
+            key="log_collection_cycle",
+            max_window_size_seconds=min(
+                int(config.log_collector.api_downtime_grace_period),
+                int(config.runtime_resources_deletion_grace_period),
+            ),
+        )
+    )
+    log_collection_cycle_tracker.initialize(db_session)
+    last_update_time = log_collection_cycle_tracker.get_window(db_session)
```
liranbg: these are sync functions atop async functions. blocking main event loop

server/api/main.py
```diff
@@ -359,6 +355,7 @@ async def _verify_log_collection_started_on_startup(
                     requested_logs=True,
                 )
     finally:
+        log_collection_cycle_tracker.update_window(db_session, now)
```
liranbg: same

liranbg (commented):
very well!

minor comment

server/api/main.py
```diff
@@ -282,6 +275,18 @@ async def _verify_log_collection_started_on_startup(
     :param start_logs_limit: Semaphore which limits the number of concurrent log collection tasks
     """
     db_session = await fastapi.concurrency.run_in_threadpool(create_session)
+    log_collection_cycle_tracker = (
+        server.api.utils.time_window_tracker.TimeWindowTracker(
+            key="log_collection_cycle",
```
alonmr: I would move this to StrEnum in the `utils.time_wondow_tracker.py`

server/api/utils/time_window_tracker.py
```diff
@@ -0,0 +1,82 @@
...
+            session,
+            self._key,
+            raise_on_not_found=raise_on_not_found,
+        )
+        if not time_window_tracker_record:
+            return
+
+        self._updated = time_window_tracker_record.updated.replace(
+            tzinfo=datetime.timezone.utc
+        )
```
alonmr: add a comment to explain?

alonmr (commented):
Minor suggestions other than Liran's

server/api/main.py
```diff
@@ -277,15 +277,19 @@ async def _verify_log_collection_started_on_startup(
...
+            key=server.api.utils.time_window_tracker.TimeWindowTrackerKeys.log_collection,
             max_window_size_seconds=min(
                 int(config.log_collector.api_downtime_grace_period),
                 int(config.runtime_resources_deletion_grace_period),
             ),
         )
     )
-    log_collection_cycle_tracker.initialize(db_session)
-    last_update_time = log_collection_cycle_tracker.get_window(db_session)
+    await fastapi.concurrency.run_in_threadpool(
```
alonmr, line 291: you might need to use `run_function_with_new_db_session` when running in different threads

alonmr (approved):

server/api/main.py
```diff
@@ -282,6 +275,22 @@ async def _verify_log_collection_started_on_startup(
     :param start_logs_limit: Semaphore which limits the number of concurrent log collection tasks
     """
     db_session = await fastapi.concurrency.run_in_threadpool(create_session)
+    log_collection_cycle_tracker = (
+        server.api.utils.time_window_tracker.TimeWindowTracker(
+            key=server.api.utils.time_window_tracker.TimeWindowTrackerKeys.log_collection,
+            max_window_size_seconds=min(
+                int(config.log_collector.api_downtime_grace_period),
+                int(config.runtime_resources_deletion_grace_period),
+            ),
```
TomerShor: Don't we need the comment anymore (that was removed)?
```
                # We take the minimum between the api_downtime_grace_period and the
                # runtime_resources_deletion_grace_period because we want to make sure that we don't miss any runs which
                # might have reached terminal state while the API was down, and their runtime resources are not deleted
```

server/api/main.py
```diff
@@ -282,6 +275,20 @@ async def _verify_log_collection_started_on_startup(
     :param start_logs_limit: Semaphore which limits the number of concurrent log collection tasks
     """
     db_session = await fastapi.concurrency.run_in_threadpool(create_session)
+    log_collection_cycle_tracker = server.api.utils.time_window_tracker.TimeWindowTracker(
+        key=server.api.utils.time_window_tracker.TimeWindowTrackerKeys.log_collection,
+        max_window_size_seconds=min(
+            int(config.log_collector.api_downtime_grace_period),
+            int(config.runtime_resources_deletion_grace_period),
+        ),
```
TomerShor, line 289: I think the comment should stay (it was in the `get_runs` call):
```
                # We take the minimum between the api_downtime_grace_period and the
                # runtime_resources_deletion_grace_period because we want to make sure that we don't miss any runs which
                # might have reached terminal state while the API was down, and their runtime resources are not deleted
```

TomerShor (changes requested):

TomerShor (approved):

PR #5998, opened 2024-07-17
[Model Monitoring] Handle exception when using kafka without applying model monitoring 

Eyal-Danieli:
This PR fix a bug in the delete project flow when the user decides to delete his project after he defined model monitoring credentials with `kafka` stream but didn't deployed the model monitoring infrastructure afterwards (meaning he didn't applied `project.enable_model_monitoring()`). In this case, no kafka topic has been generated and therefore no need to try and search for the model monitoring topics. 

A fix to https://iguazio.atlassian.net/browse/ML-7154

### DIFF
--- a/server/api/crud/model_monitoring/deployment.py
+++ b/server/api/crud/model_monitoring/deployment.py
@@ -969,7 +969,7 @@ class MonitoringDeployment:
                     error=mlrun.errors.err_to_str(e),
                 )
             except kafka.errors.UnknownTopicOrPartitionError as e:
-                logger.warning(
+                logger.info(
                     "Kafka model monitoring topics not found, probably not created",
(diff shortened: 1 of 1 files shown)


server/api/crud/model_monitoring/deployment.py
```diff
@@ -968,6 +968,12 @@ def _delete_model_monitoring_stream_resources(
                     topics=topics,
                     error=mlrun.errors.err_to_str(e),
                 )
+            except kafka.errors.UnknownTopicOrPartitionError as e:
+                logger.warning(
```
jond01: Shouldn't it be `INFO` level if expected?

jond01 (approved):

PR #5999, opened 2024-07-17
[GenAI] Add Langchain model server

ZeevRispler:
Add an option to serve model from Langchain using MLRun.

### DIFF
--- a/mlrun/frameworks/langchain/model_server.py
+++ b/mlrun/frameworks/langchain/model_server.py
@@ -1,0 +1,152 @@
+# Copyright 2023 Iguazio
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# flake8: noqa  - this is until we take care of the F401 violations with respect to __all__ & sphinx
+
+from typing import Any, Dict, Union
+
+import langchain_community.llms
+from langchain_core.language_models.llms import LLM
+
+import mlrun
+from mlrun.serving.v2_serving import V2ModelServer
+
+
+class LangChainModelServer(V2ModelServer):
+    """
+    LangChain Model serving class, inheriting the V2ModelServer class for being initialized automatically by the model
+    server and be able to run locally as part of a nuclio serverless function, or as part of a real-time pipeline.
+    For compatibility with langchain models, this class supports the following operations:
+    - invoke
+    - batch
+    - ainvoke
+    - abatch
+    in the future we will add streaming support.
+    Notice: In order to use this serving class, please ensure that the langchain_community package is installed.
+    """
+
+    def __init__(
+        self,
+        context: mlrun.MLClientCtx = None,
+        model_class: str = None,
+        llm: Union[str, LLM] = None,
+        init_kwargs: Dict[str, Any] = None,
+        generation_kwargs: Dict[str, Any] = None,
+        name: str = None,
+        model_path: str = None,
+        **kwargs,
+    ):
+        """
+        Initialize a serving class for general llm usage.
+        :param context:           The mlrun context to use.
+        :param model_class:       The class of the model to use.
+        :param llm:               The llm object itself in case of local usage.
+        :param init_kwargs:       The initialization arguments to use while initializing the llm.
+        :param generation_kwargs: The generation arguments to use while generating text.
+        :param name:              The name of this server to be initialized.
+        :param model_path:        Not in use. When adding a model pass any string value
+        """
+        super().__init__(name=name, context=context, model_path=model_path)
+        self.model_class = model_class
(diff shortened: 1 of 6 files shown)


mlrun/frameworks/langchain/model_server.py
```diff
@@ -0,0 +1,152 @@
...
+    - abatch
+    in the future we will add streaming support.
+    Notice: In order to use this serving class, please ensure that the langchain_community package is installed.
+    """
+
+    def __init__(
+        self,
+        context: mlrun.MLClientCtx = None,
+        model_class: str = None,
+        llm: Union[str, LLM] = None,
```
guy1992l, line 52: I see `llm` can be a string as well, what would be the use case? Doesn't it replace the `model_class` in that case? If so, delete the `model_class`.

mlrun/frameworks/langchain/model_server.py
```diff
@@ -0,0 +1,152 @@
...
+    ):
+        """
+        Initialize a serving class for general llm usage.
+        :param context:           The mlrun context to use.
+        :param model_class:       The class of the model to use.
+        :param llm:               The llm object itself in case of local usage.
+        :param init_kwargs:       The initialization arguments to use while initializing the llm.
+        :param generation_kwargs: The generation arguments to use while generating text.
+        :param name:              The name of this server to be initialized.
+        :param model_path:        Not in use. When adding a model pass any string value
```
guy1992l, line 66: I guess we can "log" a langchain chain as a pickle and then the model path is the pickle file or model artifact. Write it as a TODO

mlrun/frameworks/langchain/model_server.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    LangChain Model serving class, inheriting the V2ModelServer class for being initialized automatically by the model
+    server and be able to run locally as part of a nuclio serverless function, or as part of a real-time pipeline.
+    For compatibility with langchain models, this class supports the following operations:
+    - invoke
+    - batch
+    - ainvoke
+    - abatch
+    in the future we will add streaming support.
+    Notice: In order to use this serving class, please ensure that the langchain_community package is installed.
```
guy1992l:
```suggestion
    LangChain Model serving class, inheriting the V2ModelServer class to be able to run locally or as part of a Nuclio serverless function.
    
    For compatibility with langchain models, this class supports the following methods:
    
    * invoke
    * batch
    * ainvoke
    * abatch
    * stream (TBD)
    * astream (TBD)

    This class can serve a LangChain chain or llm:
    
    * `LLM` - Expose a `langchain.LLM` object as a serverless function. Can work in offline and online modes:
      * Offline: Get an initialized `LLM` object - only in local testing (mock server)
      * Online: Get a class name and initialization keyword arguments to initialize the `LLM` object and serve it.
    * `Chain` (TBD) - Expose an entire `langchain` object as a pickled model logged to MLRun.

    > Notice: In order to use this serving class, please ensure that the `langchain` and `langchain_community` packages are installed.
```

mlrun/frameworks/langchain/model_server.py
```diff
@@ -0,0 +1,152 @@
...
+        self.llm = llm
+        self.init_kwargs = init_kwargs or {}
+        self.generation_kwargs = generation_kwargs
+
+    def load(self):
+        """
+        loads the model.
+        """
+        # If the llm is already an LLM object, use it directly
+        if isinstance(self.llm, LLM):
```
guy1992l: write a TODO to verify about the Chat objects of LangChain.

mlrun/frameworks/langchain/model_server.py
```diff
@@ -0,0 +1,152 @@
...
+            self.model = self.llm
+            return
+        # If the llm is a string (or not given, then we take default model), load the llm from langchain.
```
guy1992l, line 83: how you take a default model?

tests/frameworks/langchain/test_model_server.py
```diff
@@ -0,0 +1,352 @@
...
+# 2. pull the desired model from the ollama repository (for example, in terminal: ollama pull llama3)
+# 3. run ollama with said model in terminal (for example, in terminal: ollama run llama3)
+def ollama_check_skip():
+    """
+    Check if ollama is installed
+    """
+    try:
+        result = subprocess.run(["ollama", "--help"], stdout=PIPE)
+    except Exception as e:
+        print(f"Error checking ollama: {e}")
```
guy1992l, line 49: why would it be raised?

guy1992l (changes requested):
great stuff :)

Some questions and edit for the class docstring.

mlrun/frameworks/langchain/model_server.py
ZeevRispler: you just init without specifing (ex: openai() and then you get gpt-3.5)

tests/frameworks/langchain/test_model_server.py
ZeevRispler: didn't find command ollama (or something similar)

guy1992l: ah so True is to skip, ok!

mlrun/frameworks/langchain/model_server.py
ZeevRispler: you want to do this in the model server?

mlrun/frameworks/langchain/model_server.py
ZeevRispler: in addition to the LLM (which is an langchain class)? like a full chain or something?

PR #6001, opened 2024-07-17
[Pipelines] Fix get run status for local engine

roei3000b:
Moving the `get_run_status` function to the base class: `_PipelineRunner` because we want that function to return tuple of `(state, errors_counter, text)`
`state` - State of the workflow
`errors_counter` - Counter of steps that ends with error state.
`text` - text that describes the workflow.

The flow of `get_run_status`:
1. Waiting for the pipeline to complete
2. Get the status of the run
3. Return the value as we described above

It should fix the `test_remote_pipeline_with_local_engine_from_github` system test(before this cr we returned `None` instead of tuple so we failed in https://github.com/mlrun/mlrun/blob/6f510f67e49ed4ba9d2eca43d0857548359f9085/mlrun/projects/pipelines.py#L1112)


alonmr (changes requested):
I would't use the KFP runner when using local engine since kfp is not even in the equation. Instead, pick the code you want to use and move it to a common place (the base class ;) )

mlrun/projects/pipelines.py
```diff
@@ -65,6 +65,15 @@ def get_workflow_engine(engine_kind, local=False):
     )
 
 
+def generate_text_for_workflow(run_id, errors_counter, state):
```
alonmr: Can be a staticmethod of `_PipelineRunner`
```suggestion
def _generate_workflow_finished_message(run_id, errors_counter, state):
```

mlrun/projects/pipelines.py
```diff
@@ -932,6 +937,13 @@ def get_run_status(
             )
             run._exc = pipeline_runner_run.status.error
 
+            errors_counter = 0
+            if run._state == mlrun_pipelines.common.models.RunStatuses.succeeded:
+                errors_counter = 1
```
alonmr: 1. The if should be `!=`
2. What about the other pipeline steps. This run is just the remote runner. You need to get all of the steps and sum the failed ones.

alonmr (changes requested):
Couple of comments. Also can you please explain why you can't reuse more parts of `_KFPRunner.get_run_status` ? Specifically I wonder about `run.wait_for_completion` and summing the errors parts.

mlrun/projects/pipelines.py
```diff
@@ -458,6 +462,49 @@ def wait_for_completion(run_id, project=None, timeout=None, expected_statuses=No
...
+    def get_run_status(
+        project,
+        run: _PipelineRunStatus,
+        timeout=None,
+        expected_statuses=None,
+        notifiers: mlrun.utils.notifications.CustomNotificationPusher = None,
+        **kwargs,
+    ):
+        if timeout is None:
+            timeout = 60 * 60
```
rokatyy:
```suggestion
timeout = timeout or 60*60
```

mlrun/projects/pipelines.py
```diff
@@ -585,12 +628,12 @@ def run(
         return _PipelineRunStatus(run_id, cls, project=project, workflow=workflow_spec)
 
     @staticmethod
-    def wait_for_completion(run_id, project=None, timeout=None, expected_statuses=None):
-        if timeout is None:
-            timeout = 60 * 60
+    def wait_for_completion(run, project=None, timeout=None, expected_statuses=None):
+        logger.info("Waiting for pipeline run completion")
```
rokatyy: Looks more like a debug log and it would be also beneficial to add some information about run. For instance, project and run_id.

roei3000b: About debug log - agree with you but I just copied the same behavior that was previously.
This log is show to the user so I think I should remain the log level but as you suggest I will add the project and run_id.

rokatyy (approved):
LGTM

mlrun/projects/pipelines.py
```diff
@@ -732,18 +732,8 @@ def get_state(run_id, project=None):
...
-    def get_run_status(
-        project,
-        run,
-        timeout=None,
-        expected_statuses=None,
-        notifiers: mlrun.utils.notifications.CustomNotificationPusher = None,
-        **kwargs,
-    ):
+    def wait_for_completion(run, project=None, timeout=None, expected_statuses=None):
+        # local runner does not have a wait_for_completion method because we already wait in the `run` function
```
alonmr:
```suggestion
        # TODO: local runner blocks for the duration of the pipeline therefore usually there will be nothing to wait for. However, users may run functions with watch=False and then it can be useful to wait for the runs here.
```

mlrun/projects/pipelines.py
```diff
@@ -404,12 +404,16 @@ def exc(self):
...
-        self._state = self._engine.wait_for_completion(
-            self.run_id,
+        returned_state = self._engine.wait_for_completion(
+            self,
             project=self.project,
             timeout=timeout,
             expected_statuses=expected_statuses,
         )
+        # wait_for_completion may return None for some runners(like local) so we want to override
+        # the state only if it returns value
```
alonmr:
```suggestion
# TODO: returning a state is optional until all runners implement wait_for_completion
```

alonmr (commented):
Looks very good! Small comments

liranbg (approved):

alonmr (approved):

PR #6004, opened 2024-07-17
[Nuclio] Add warning that a number of replicas is changed

rokatyy:
Adds warning that a number of replicas is changed

Jira - https://iguazio.atlassian.net/issues/ML-6262


mlrun/runtimes/nuclio/function.py
```diff
@@ -510,8 +509,9 @@ def add_v3io_stream_trigger(
                 **kwargs,
             ),
         )
-        self.spec.min_replicas = shards
-        self.spec.max_replicas = shards
+        if shards:
+            self.spec.min_replicas = shards
+            self.spec.max_replicas = shards
```
TomerShor: What if shards is None? Does min/max replicas get a default value later in the flow? (or in Nuclio?🫣)

mlrun/runtimes/nuclio/function.py
```diff
@@ -462,7 +462,7 @@ def add_v3io_stream_trigger(
         name="stream",
         group="serving",
         seek_to="earliest",
-        shards=1,
+        shards=None,
```
gtopper: Maybe we should require this parameter, or try to check it for the user (i.e. `DescribeStream`), because as I remember, the design is to have one replica per shard, because V3IO streams don't support rebalancing.

gtopper: See https://github.com/mlrun/mlrun/blob/development/docs/serving/graph-ha-cfg.md#consumer-function-configuration

mlrun/runtimes/nuclio/function.py
```diff
@@ -510,6 +510,7 @@ def add_v3io_stream_trigger(
                 **kwargs,
             ),
         )
+        logger.warning(f"Setting function replicas to {shards}")
```
TomerShor: I think you should check if there is some non-zero value first in the min/max replicas, and warn that is being **overridden**.

Otherwise it doesn't matter much.

TomerShor (changes requested):

mlrun/runtimes/nuclio/function.py
rokatyy: @TomerShor agree, fixed

TomerShor (approved):
🚀

PR #6005, opened 2024-07-17
[ProjectSummary] Add a new project summary row to the DB when a project is created

moranbental:
Project summaries are calculated in a separate periodic task that runs every 30 seconds. 
There was a race condition when creating a new project and then immediately trying to get the project summary for this project, but the project summary did not exist in the database. 

To fix this, we create a new project summary row every time a new project is created and also delete the project row from the project summary table when a project is deleted.

https://iguazio.atlassian.net/browse/ML-7182


server/api/db/sqldb/db.py
```diff
@@ -2485,6 +2485,8 @@ def create_project(self, session: Session, project: mlrun.common.schemas.Project
         update_labels(project_record, labels)
         self._upsert(session, [project_record])
```
liranbg: why not upsertting here both project and project summary records?

liranbg (approved):

server/api/db/sqldb/db.py
```diff
@@ -2605,6 +2612,7 @@ def get_project_summary(
                 f"Project summary not found: project={project}"
```
alonmr: not related but nice python feature can be used here
```suggestion
                f"Project summary not found: {project=}"
```

server/api/db/base.py
```diff
@@ -534,6 +534,9 @@ def list_project_summaries(
     ):
         pass
 
+    def create_project_summary(self, session, project: str):
```
alonmr: is this implemented?

PR #6010, opened 2024-07-18
[Functions] Support listing only tagged functions

TomerShor:
In the List Functions API we have a `tag` query params.
This PR adds functionality to the tag params as follows:

- `tag = None` or `tag = ""` - All function are returned (tagged and untagged) - Same as the current behavior.
- `tag="specific-tag"` - Only function with this specific tag are returned.
- `tag="*"` - returns only functions with tags

This allows listing only tagged functions, so we can reduce the number of functions to load in the UI by default.

BC is kept as without the explicit tag filter all function are returned, same as the today.

https://iguazio.atlassian.net/browse/ML-7130

### DIFF
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -4549,7 +4549,7 @@ class SQLDB(DBInterface):
         else:
             # Only get functions that have tags with join (faster than outer join)
             query = query.join(Function.Tag, Function.id == Function.Tag.obj_id)
-            if tag == "*":
+            if tag != "*":
                 # Filter on the specific tag
                 query = query.filter(Function.Tag.name == tag)
 
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -1755,6 +1755,7 @@ class SQLDB(DBInterface):
         format_: str = mlrun.common.formatters.FunctionFormat.full,
         page: typing.Optional[int] = None,
         page_size: typing.Optional[int] = None,
+        untagged: bool = True,
     ) -> list[dict]:
         project = project or mlrun.mlconf.default_project
         functions = []
@@ -1767,6 +1768,7 @@ class SQLDB(DBInterface):
             hash_key,
             page=page,
             page_size=page_size,
+            untagged=untagged,
         ):
             function_dict = function.struct
             if not function_tag:
@@ -4484,6 +4486,7 @@ class SQLDB(DBInterface):
         hash_key: typing.Optional[str] = None,
         page: typing.Optional[int] = None,
         page_size: typing.Optional[int] = None,
+        untagged: bool = True,
     ) -> list[tuple[Function, str]]:
         query = session.query(Function, Function.Tag.name)
         query = query.filter(Function.project == project)
@@ -4495,8 +4498,14 @@ class SQLDB(DBInterface):
             query = query.filter(Function.uid == hash_key)
 
         if not tag:
-            # If no tag is given, we need to outer join to get all functions, even if they don't have tags.
-            query = query.outerjoin(Function.Tag, Function.id == Function.Tag.obj_id)
+            if untagged:
+                # If no tag is given, we need to outer join to get all functions, even if they don't have tags.
+                query = query.outerjoin(
+                    Function.Tag, Function.id == Function.Tag.obj_id
+                )
+            else:
+                # We only want to get functions that have tags
+                query = query.join(Function.Tag, Function.id == Function.Tag.obj_id)
         else:
             # If a tag is given, we can just join (faster than outer join) and filter on the tag.
             query = query.join(Function.Tag, Function.id == Function.Tag.obj_id)
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -1755,16 +1755,20 @@ class SQLDB(DBInterface):
         format_: str = mlrun.common.formatters.FunctionFormat.full,
         page: typing.Optional[int] = None,
         page_size: typing.Optional[int] = None,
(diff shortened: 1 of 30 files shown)


quaark (commented):
The logic is definitely sound and code looks good.
I fear this API is a bit confusing. Why not `only_tagged` that defaults to `False` to maintain BC, and we can request `only_tagged=True` to suggest we only want tagged functions. 
Then:
- `only_tagged=False` with `tag=...` will be exactly like `only_tagged=True` with `tag=...` and just return functions with the requested tag filter.
- Without `tag=...` then as the name suggests `only_tagged=False` will return all functions and `only_tagged=True` will only return tagged functions.

quaark:
> The logic is definitely sound and code looks good. I fear this API is a bit confusing. Why not `only_tagged` that defaults to `False` to maintain BC, and we can request `only_tagged=True` to suggest we only want tagged functions. Then:
> 
> * `only_tagged=False` with `tag=...` will be exactly like `only_tagged=True` with `tag=...` and just return functions with the requested tag filter.
> * Without `tag=...` then as the name suggests `only_tagged=False` will return all functions and `only_tagged=True` will only return tagged functions.

`tagged_only` maybe better

TomerShor:
> The logic is definitely sound and code looks good. I fear this API is a bit confusing. Why not `only_tagged` that defaults to `False` to maintain BC, and we can request `only_tagged=True` to suggest we only want tagged functions. Then:
> 
> * `only_tagged=False` with `tag=...` will be exactly like `only_tagged=True` with `tag=...` and just return functions with the requested tag filter.
> * Without `tag=...` then as the name suggests `only_tagged=False` will return all functions and `only_tagged=True` will only return tagged functions.

I tend to agree with you, going with `False` as the default is more intuitive, and if you want a specific feature them set it to `True`.

roei3000b (approved):

alonmr (commented):
I have an implementation suggestion IIUC should cover all cases:
Instead of the new param allow specifying `tag="*"` which is the equivalent to `untagged=False` and not specifying a tag. `untagged=True` behavior should be done when a `tag` is not specified.
Either way i think there is a loophole in the httpdb impl - if I use mlrun 1.6 and I specify a tag then untagged is None and it will not fail. If I upgrade to 1.7 I have to specify untagged to False or it will explode - that means I have to change the code and its not really BC

TomerShor:
> I have an implementation suggestion IIUC should cover all cases: Instead of the new param allow specifying `tag="*"` which is the equivalent to `untagged=False` and not specifying a tag. `untagged=True` behavior should be done when a `tag` is not specified. 

@alonmr I like your idea 👍 
So if tag is not specified at all it will return all tags, same as previously. 

> Either way i think there is a loophole in the httpdb impl - if I use mlrun 1.6 and I specify a tag then untagged is None and it will not fail. If I upgrade to 1.7 I have to specify untagged to False or it will explode - that means I have to change the code and its not really BC

Nice catch! This could have been avoided by setting the default in http to false, and only enriching the default in the BE, however if I change the logic it's irrelevant.

alonmr (commented):
looks great please add information about the new functionality in httpdb / project api etc.

alonmr (approved):

server/api/db/sqldb/db.py
```diff
@@ -4546,6 +4546,9 @@ def _find_functions(
         if not tag:
             # If no tag is given, we need to outer join to get all functions, even if they don't have tags.
             query = query.outerjoin(Function.Tag, Function.id == Function.Tag.obj_id)
+        elif tag == "*":
+            # Only get functions that have tags
+            query = query.join(Function.Tag, Function.id == Function.Tag.obj_id)
```
```suggestion
        else:
            # Only get functions that have tags
            query = query.join(Function.Tag, Function.id == Function.Tag.obj_id)
            if tag != "*":
                query = query.filter(Function.Tag.name == tag)
```
Could have done this way, but doesn't really matter.

theSaarco (approved):
👍 LGTM

server/api/db/sqldb/db.py
TomerShor: Done 👍

server/api/db/sqldb/db.py
```diff
@@ -4547,9 +4547,11 @@ def _find_functions(
             # If no tag is given, we need to outer join to get all functions, even if they don't have tags.
             query = query.outerjoin(Function.Tag, Function.id == Function.Tag.obj_id)
         else:
-            # If a tag is given, we can just join (faster than outer join) and filter on the tag.
+            # Only get functions that have tags with join (faster than outer join)
             query = query.join(Function.Tag, Function.id == Function.Tag.obj_id)
-            query = query.filter(Function.Tag.name == tag)
+            if tag == "*":
```
```suggestion
            if tag != "*":
```
!!!!!

tests/api/db/test_functions.py
```diff
@@ -451,6 +451,62 @@ def test_list_artifacts_with_time_filters(db: DBInterface, db_session: Session):
...
+    # list only tagged functions
+    tagged_function = db.list_functions(db_session, tag="*")
+
+    assert len(functions) != len(tagged_function)
+
+    all_hashes = [function["metadata"]["hash"] for function in functions]
+    untagged_hashes = [function["metadata"]["hash"] for function in tagged_function]
+
+    assert function_2_hash_key in all_hashes
+    assert function_2_hash_key not in untagged_hashes
```
liranbg, line 507: can add list functions with specific tag to ensure it is covered as well?

liranbg (commented):
minor

tests/api/db/test_functions.py
TomerShor: There was already a test but added the check anyways

PR #6012, opened 2024-07-18
[GenAI] Clean API and Client

yonishelach:
Few things:

1. Fixed the ingest endpoint to enable the controller to reroute ingest requests to the application.
2. Deleted unnecessary endpoints.
3. Deleted Unnecessary Client methods since the controller gets the requests directly from the front end.
4. Deleted the `utils.py` file (was moved into the controller's code - https://github.com/mlrun/genai-factory/pull/6.


mlrun/genai/api.py
```diff
@@ -59,10 +59,19 @@ async def get_auth_user(
...
 
 
-@router.post("/collections/{collection}/{path}/{loader}/ingest")
-async def ingest(collection, path, loader, metadata, version, from_file):
+@router.post("/collections/{collection}/{loader}/ingest")
+async def ingest(
+    collection: str,
+    loader: str,
+    path: str,
+    metadata: Tuple[str, str] = None,
```
guy1992l: why would `metadata` be a tuple and not a dictionary?

guy1992l (approved):

PR #6016, opened 2024-07-19
Revert "[Chief/Worker] Support Opt Out of Chief Features"

liranbg:
Reverts mlrun/mlrun#5985

This one breaks `list_schedules` allowed on workers. we still want scheduler to be able to list schedules on api


liranbg:
being fixed by https://github.com/mlrun/mlrun/pull/6017

PR #6019, opened 2024-07-21
[Docs] Add statement that changes to Nuclio functions should only made in MLRun

jillnogold:
[ML-7190](https://iguazio.atlassian.net/browse/ML-7190)

[ML-7190]: https://iguazio.atlassian.net/browse/ML-7190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/concepts/nuclio-real-time-functions.ipynb
```diff
@@ -26,6 +26,12 @@
     "\n",
     "Read more in the [Nuclio documentation](https://nuclio.io/docs/latest/) and the open-source [MLRun library](https://github.com/mlrun/mlrun).\n",
     "\n",
+    "```{admonition} Important\n",
+    "When working with Nuclio functions (remote / serving / application runtimes) and API Gateways in MLRun, any modifications/deletions should be made in MLRun only. This is also relevant for modifying the spec of a function or API Gateway.\n",
+    "<b>Changes that are made in Nuclio are not guaranteed to be synced to MLRun and can cause unexpected behavior.</b> </br>\n",
+    "For example, if you create a remote function in MLRun (which is created in Nuclio) and then delete it in Nuclio, the funcion continues to run in MLRun.\n",
```
TomerShor: It doesn't actually continue to run in MLRun, as it can only run in Nuclio.
But, it is possible that the function status will be `ready`, but invoking it will fail.

TomerShor (changes requested):

TomerShor (approved):

PR #6020, opened 2024-07-21
[API] Ease project image builder pydantic validation

alonmr:
Validate the image builder only for input and not for output since we may have corrupted values in the DB and we want to be able to return them 
https://iguazio.atlassian.net/browse/ML-7135

### DIFF
--- a/mlrun/common/schemas/__init__.py
+++ b/mlrun/common/schemas/__init__.py
@@ -170,6 +170,7 @@ from .project import (
     Project,
     ProjectDesiredState,
     ProjectMetadata,
+    ProjectOut,
     ProjectOutput,
     ProjectOwner,
     ProjectsOutput,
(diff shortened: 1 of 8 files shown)


TomerShor (commented):
Code looks good!

With this fix we we'll be able to list the projects but we will still encounter errors dealing with the corrupted project. Do we also want to do some kind of migration?

alonmr:
> Code looks good!
> 
> With this fix we we'll be able to list the projects but we will still encounter errors dealing with the corrupted project. Do we also want to do some kind of migration?

Right users will not be able to store this project without fixing the corruption. We prefer to avoid migration because it means losing data (clearing the corrupted attrs)

moranbental (approved):
Looks good!

liranbg (approved):

PR #6007, opened 2024-07-18
[Model Monitoring] Avoid returning stats in `list_model_endpoints`

gtopper:
To drastically reduce the response time.

[ML-6594](https://iguazio.atlassian.net/browse/ML-6594)

[ML-6594]: https://iguazio.atlassian.net/browse/ML-6594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6025, opened 2024-07-21
[Model Monitoring] Enrich model endpoint with the unique `model_uri` reference 

Eyal-Danieli:
Enrich the model endpoint object with the model uri from the model object with a unique identifier. Instead of using the model uri that includes the model object's name and version, we will use a unique reference that includes the `project`, `key`, `iter`, and `tree` values. This will allow us to handle future changes in the model version without updating the model endpoint object.

Related JIRA: https://iguazio.atlassian.net/browse/ML-5675


tests/system/model_monitoring/test_model_monitoring.py
```diff
@@ -338,8 +336,9 @@ def test_basic_model_monitoring(self, with_sql_target: bool) -> None:
 
         endpoint = endpoints_list[0]
 
-        # ML-6594
-        assert not endpoint.status.feature_stats
+        if not with_sql_target:
+            # ML-6594
+            assert not endpoint.status.feature_stats
```
davesh0812: according to https://github.com/mlrun/mlrun/pull/6007 It suppose to be excluded in v3ioKV & SQL no?

Eyal-Danieli: yes it's still a bug in sql

davesh0812: OK so as discussed reopen https://iguazio.atlassian.net/browse/ML-6594 or just open a new bug. 
FYI @gtopper

davesh0812 (approved):
Looks Good, just revert the change in the tests because it already fixed by other pr

PR #6027, opened 2024-07-22
[Alerts] Remove needless log line

liranbg:
too verbose and anyway we have log line per sent notification

also removed unused session parameter


yanburman:
I am fine with the log line removal, but the session removal seems wrong, as there are 2 flavors of this method and one actually has session in it

PR #6034, opened 2024-07-23
[Alerts] Avoid caching session argument

yanburman:
We are using LRU cache for caching objects. The hash is calculated based on the arguments to the function. The session argument will change between different requests, so we must not count it towards the hash, as what we care about are the actual alert/state parameters

### DIFF
--- a/server/api/utils/lru_cache.py
+++ b/server/api/utils/lru_cache.py
@@ -39,11 +39,20 @@ class LRUCache:
             self.misses = 0
             self.currsize = 0
 
-    def __init__(self, func, maxsize: int = 128):
+    def __init__(
+        self, func, maxsize: int = 128, ignore_args_for_hash: list[int] = None
+    ):
+        """
+        Initialize an lru cache instance
+        :param func: The function that gets the actual value
+        :param maxsize: Maximum size of the cache
+        :param ignore_args_for_hash: List of argument indexes to ignore when computing the hash for the cache
+        """
         self.cache = collections.OrderedDict()
         self.func = func
         self.maxsize = maxsize
         self._cache_info = self.CacheInfo(maxsize)
+        self.ignored_args = ignore_args_for_hash or []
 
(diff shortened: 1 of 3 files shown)


server/api/utils/lru_cache.py
```diff
@@ -100,6 +109,10 @@ def cached(self, *args, **kwargs) -> bool:
         key = self._gen_key(args, kwargs)
         return key in self.cache
 
-    @staticmethod
-    def _gen_key(args, kwargs) -> str:
-        return hashlib.sha256(f"{args}/{sorted(kwargs.items())}".encode()).hexdigest()
+    def _gen_key(self, args, kwargs) -> str:
+        args_for_hash = list(args)
+        for idx in self.ignored_args:
+            del args_for_hash[idx]
```
moranbental:
```suggestion
        args_for_hash = [arg for idx, arg in enumerate(args) if idx not in self.ignored_args]
```

moranbental (changes requested):

server/api/utils/lru_cache.py
```diff
@@ -39,11 +39,20 @@ def reset(self):
...
+        Initialize an lru cache instance
+        :param func: The function that gets the actual value
+        :param maxsize: Maximum size of the cache
+        :param ignore_args_for_hash: List of argument indexes to ignore when computing the hash for the cache
+        """
         self.cache = collections.OrderedDict()
         self.func = func
         self.maxsize = maxsize
         self._cache_info = self.CacheInfo(maxsize)
+        self.ignored_args = [] if ignore_args_for_hash is None else ignore_args_for_hash
```
moranbental: I think it is more readable
```
self.ignored_args = ignore_args_for_hash or []
```

moranbental (changes requested):

moranbental (approved):
LGTM

PR #6038, opened 2024-07-23
[Runtimes] Ensure Node Selector is Applied to Spark Runtime properly

yaelgen:
resolves: [ML-7123](https://iguazio.atlassian.net/browse/ML-7123)

This PR addresses an issue where the node selector was not applied to Spark runtime pods when the project lacked a default node selector. The fix ensures the node selector is set on the run object and subsequently applied to the resulting pod. Unit tests were modified to cover this scenario.

Other runtimes were unaffected due to backward compatibility logic that handled both cases where the project had or lacked a node selector (In `resolve_node_selector`, if a project was defined, the node selector on the run was enriched, so it would be used. If no default function node selector was set on the project, the run node selector was not enriched and the node selector would be taken from the runtime object, which is acceptable.
The Spark flow does not reach this function. Consequently, when no project node selector is defined, the run is not enriched, and when creating a run, [the job is enriched from the run object](https://github.com/mlrun/mlrun/blob/8f3bc6f5b2f2dec74a5fba199311e73095f7b945/server/api/runtime_handlers/sparkjob/spark3job.py#L156), resulting in an empty node selector.

[ML-7123]: https://iguazio.atlassian.net/browse/ML-7123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/api/launcher.py
```diff
@@ -224,6 +225,7 @@ def _pre_run_enrichement(self, runtime, run):
         This ensures the node selector is correctly set on the run
         while maintaining the runtime's integrity from system-specific project settings.
         """
+        run.spec.node_selector = deepcopy(runtime.spec.node_selector)
```
alonmr, line 228: I think we need a better differentiation between this function and `_enrich_run` because that also happens "pre run". So when I saw this line I thought it should be in `_enrich_run` but I guess you want it here because it's related to node selector. I think it should be called `_enrich_run_node_selector`

yaelgen: Changed to `_pre_run_node_selector_enrichement`, wdyt?

alonmr:
> (Other runtimes were unaffected due to backward compatibility logic that handled both cases where the project had or lacked a node selector).

Please explain what is that logic and why is it not applied for spark

tests/api/runtimes/test_spark.py
```diff
@@ -465,16 +484,18 @@ def test_with_node_selector(self, db: sqlalchemy.orm.Session, k8s_secrets_mock):
...
-                self.project_default_function_node_selector, driver_node_selector
+                project_node_selector, driver_node_selector
             ),
             mlrun.utils.helpers.merge_with_precedence(
-                self.project_default_function_node_selector, executor_node_selector
+                project_node_selector, executor_node_selector
             ),
         )
 
+        self.project = default_project_name
```
alonmr: this is an issue because if the test fails than this will not be set back. After we found the original issue I believe you should be able to use the same project that is created in the setup

alonmr (approved):

PR #6039, opened 2024-07-23
[Artifacts] Add `uid` query param support in API

TomerShor:
Old clients (httpdb) are sending a `uid` query param when deleting and reading a single artifact.
The backend only supports `object-uid` or `object_uid`, and not `uid`.

In FastAPI+Pydantic there isn't an option to configure 2 aliases to a query param, so added the `uid` param to the relevant endpoints.
Also, aligned the httpdb client to use `object-uid`.

https://iguazio.atlassian.net/browse/ML-7256


server/api/api/endpoints/artifacts_v2.py
```diff
@@ -199,6 +199,7 @@ async def get_artifact(
     tag: str = None,
     iter: int = None,
     object_uid: str = Query(None, alias="object-uid"),
+    uid: str = Query(None),
```
roei3000b, line 206: I think it should work:
```suggestion
    uid: str = Query(None, alias="object-uid"),
```

rokatyy (commented):
I think it is needed to add a comment explaining why do we have `uid` and `object_uid` and to add TODO to get rid of `uid` in the future

server/api/api/endpoints/artifacts_v2.py
```diff
@@ -231,6 +232,7 @@ async def delete_artifact(
     tree: str = None,
     tag: str = None,
     object_uid: str = Query(None, alias="object-uid"),
+    uid: str = Query(None),
```
alonmr: agree with @rokatyy and even further can add `deprecated=True` with description

server/api/api/endpoints/artifacts_v2.py
TomerShor: I wouldn't want to remove `object_uid` as it will remove BC.

rokatyy (approved):

liranbg (approved):

PR #6040, opened 2024-07-23
[Alerts] Change default reset policy to auto

yanburman:
https://iguazio.atlassian.net/browse/ML-7262

### DIFF
--- a/mlrun/common/schemas/alert.py
+++ b/mlrun/common/schemas/alert.py
@@ -149,7 +149,7 @@ class AlertConfig(pydantic.BaseModel):
     entities: EventEntities
     trigger: AlertTrigger
     criteria: Optional[AlertCriteria]
-    reset_policy: ResetPolicy = ResetPolicy.MANUAL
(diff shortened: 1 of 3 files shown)


tests/system/alerts/test_alerts.py
```diff
@@ -343,6 +343,7 @@ def _create_alert_config(
             trigger=alert_objects.AlertTrigger(events=[event_name]),
             criteria=criteria,
             notifications=notifications,
+            reset_policy=alert_objects.ResetPolicy.MANUAL,
```
alonmr: please add a comment why in system tests you use manual

Yacouby: it is not a must to use manual here, this line can be deleted

Yacouby:
```suggestion
```

Yacouby (changes requested):

alonmr (approved):

PR #6041, opened 2024-07-23
[Alerts] Do not require project name if storing alert via project

yanburman:
https://iguazio.atlassian.net/browse/ML-7239

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -4063,6 +4063,12 @@ class MlrunProject(ModelObj):
         db = mlrun.db.get_run_db(secrets=self._secrets)
         if alert_name is None:
             alert_name = alert_data.name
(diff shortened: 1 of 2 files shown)


alonmr (commented):
please add a unit test

mlrun/projects/project.py
```diff
@@ -4063,6 +4063,12 @@ def store_alert_config(
         db = mlrun.db.get_run_db(secrets=self._secrets)
         if alert_name is None:
             alert_name = alert_data.name
+        if alert_data.project is not None and alert_data.project != self.metadata.name:
+            logger.warn(
+                "Project in alert does not match project in operation",
+                project=alert_data.project,
+            )
```
alonmr, line 4070: actually I think we should just overwrite the project. same as we do if you import a function from a different project

alonmr (approved):

PR #6043, opened 2024-07-23
[Runtimes] Update node selector enrichment on spark driver and executor

yaelgen:
resolves: [ML-7148 ](https://iguazio.atlassian.net/browse/ML-7148)

Due to Spark's limitation of not allowing separate node selectors for both the application and driver/executor, we've revised the behavior of with_node_selection.

Changes in this PR  to prevent Spark from failing due to improper node selector configurations:
- The `with_node_selection` method for Spark functions now calls `with_driver_node_selector` and `with_executor_node_selector` instead of modifying the function object directly.
- The CRD creation now sets the spec.nodeSelector to an empty value. If neither `driver_node_selector` nor `executor_node_selector` is defined, the node selector will be inherited from the function, enriched by the project and configuration, similar to other runtimes.
- Added deprecation for the with_node_selection function for Spark, as it is no longer meaningful. Users are directed to use `with_driver_node_selection` and `with_executor_node_selection` instead.

This change is feasible because the application node selector has no real significance since no pod is created for it; only the driver and executor selectors really matter.


mlrun/runtimes/sparkjob/spark3job.py
```diff
@@ -481,6 +482,16 @@ def _get_igz_deps(self):
             "files": ["local:///igz/java/libs/v3io-pyspark.zip"],
         }
 
+    @deprecated(
+        version="1.9.0",
+        reason=(
+            "Due to Spark's restriction that node selectors "
+            "cannot be defined for both the application and driver/executor, "
+            "this method will be deprecated in 1.9.0. Use 'with_executor_node_selection()'"
```
alonmr:
```suggestion
            "this method is deprecated in 1.7.0 and will be removed in 1.9.0. Use 'with_executor_node_selection()'"
```

server/api/runtime_handlers/sparkjob/spark3job.py
```diff
@@ -700,24 +693,52 @@ def _enrich_job(
...
-        # After adding the runtime executor and driver node selectors to the job object,
+    def _enrich_node_selectors(
+        self,
+        run: mlrun.run.RunObject,
+        runtime: mlrun.runtimes.sparkjob.spark3job.Spark3Runtime,
+        job: dict,
+        project_name: str,
+    ):
+        # If `driver_node_selector` is defined, it takes precedence and will be used for the driver node selector.
+        # If not provided or empty, the `node_selector` from the run will be used instead.
```
alonmr: what is the run node selector in this context? project and config?

tests/system/projects/test_project.py
```diff
@@ -1244,6 +1244,42 @@ def test_project_default_function_node_selector_using_igz_mgmt(self):
...
+
+        spark_function.with_executor_limits(cpu="1400m")
+        spark_function.with_executor_requests(cpu=1, mem="512m")
+
+        node_selector = {
+            function_label_name: function_label_val,
+            function_override_label: function_override_val,
+        }
+
+        spark_function.with_node_selection(node_selector=node_selector)
```
alonmr, line 1273: use executer and driver since this is deprecated

tests/system/projects/test_project.py
```diff
@@ -1244,6 +1244,42 @@ def test_project_default_function_node_selector_using_igz_mgmt(self):
...
+
+        assert spark_function.spec.driver_node_selector == node_selector
+        assert spark_function.spec.executor_node_selector == node_selector
+
+        spark_function.with_igz_spark()
+
+        spark_run = spark_function.run(auto_build=True)
+        assert spark_run.status.state == RunStates.completed
```
alonmr, line 1281: what about validating the run / crds?

alonmr (commented):
Great job! Minor comments/questions

mlrun/runtimes/sparkjob/spark3job.py
TomerShor: Why do we deprecate it instead of keeping it as a syntax sugar for setting both executor and driver NS?

server/api/runtime_handlers/sparkjob/spark3job.py
yaelgen: Yes, but in case the user decided to do explicitly edit the function like this: `spark_fn.spec.node_selector = {"key": "val"}`, the value will be what they have defined + project and config.

tests/system/projects/test_project.py
yaelgen: I guess you are right, but I also wanted to test that the changes to `with_node_selection` works as expected now, do you think it's okay not to test it and just  use the driver and executor ones?

mlrun/runtimes/sparkjob/spark3job.py
yaelgen: Saar suggested deprecating it, likely because it's not clear to users what this method actually does, and since in Spark, setting node selection on the function itself has no real significance. What Spark ultimately does is apply the function node selector to the driver and executor if nothing else is set. In this implementation, we're doing this ourselves to prevent users from failing Spark due to poorly defined node selectors.
Note that it also raises errors for all types other than `node_selector`, which eventually call other methods that can set this other types (affinity, tolerations, etc.) for the driver and executor, so this is a bit of a weird behavior to support it like this. I think mostly it is an unclear terminology, since users might expect this method to update the function spec directly.

However, since the method is derived from the base class, we can't simply remove it. So perhaps we could follow your suggestion and add a warning to inform users about its actual behavior. I'm not entirely sure what the best approach is here.
@alonmr, wdyt?

yaelgen: I discussed this with Liran and we agreed to have this as a syntax sugar, and pass also the other parameters to the driver and executor functions, will remove the deprecation.

server/api/runtime_handlers/sparkjob/spark3job.py
alonmr: explain where the `run.spec.node_selector` was enriched from

tests/system/projects/test_project.py
alonmr: use `list_runtime_resource` from httpdb

liranbg (approved):

PR #6047, opened 2024-07-24
[Application] Build flow enhancements

alonmr:
1. Add status to API gateway to align with mlrun models
2. Remove build source before sending to nuclio deploy to make it build from image
3. Do not enrich spec command and do not allow using `.run()` for now.

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -1686,7 +1686,7 @@ class HTTPRunDB(RunDBInterface):
             last_log_timestamp = float(
                 resp.headers.get("x-mlrun-last-timestamp", "0.0")
             )
-            if func.kind in mlrun.runtimes.RuntimeKinds.nuclio_runtimes():
+            if func.kind in mlrun.runtimes.RuntimeKinds.pure_nuclio_deployed_runtimes():
                 mlrun.runtimes.nuclio.function.enrich_nuclio_function_from_headers(
                     func, resp.headers
                 )
--- a/mlrun/runtimes/nuclio/application/application.py
+++ b/mlrun/runtimes/nuclio/application/application.py
@@ -366,6 +366,14 @@ class ApplicationRuntime(RemoteRuntime):
             target_dir=target_dir,
         )
 
+    def from_image(self, image):
+        super().from_image(image)
+        # nuclio implementation detail - when providing the image and emptying out the source code and build source,
+        # nuclio skips rebuilding the image and simply takes the prebuilt image
+        self.spec.build.functionSourceCode = ""
+        self.status.application_source = self.spec.build.source
+        self.spec.build.source = ""
+
     @classmethod
     def get_filename_and_handler(cls) -> (str, str):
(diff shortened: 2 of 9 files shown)


mlrun/runtimes/nuclio/application/application.py
```diff
@@ -506,9 +529,11 @@ def _configure_application_sidecar(self):
 
         if self.status.container_image:
             self.from_image(self.status.container_image)
-            # nuclio implementation detail - when providing the image and emptying out the source code,
+            # nuclio implementation detail - when providing the image and emptying out the source code and build source,
             # nuclio skips rebuilding the image and simply takes the prebuilt image
             self.spec.build.functionSourceCode = ""
+            self.status.application_source = self.spec.build.source
+            self.spec.build.source = ""
```
TomerShor: How about overriding the `from_image` method for application runtime and putting everything there?
Then we can call it independently if we want and it will have the same effect

tests/runtimes/test_application.py
```diff
@@ -265,7 +265,7 @@ def _assert_function_handler(fn):
     ) = mlrun.runtimes.ApplicationRuntime.get_filename_and_handler()
     expected_filename = pathlib.Path(filepath).name
     expected_module = mlrun.utils.normalize_name(expected_filename.split(".")[0])
-    expected_function_handler = f"{expected_module}:{expected_handler}"
+    expected_function_handler = f"{expected_module}-nuclio:{expected_handler}"
```
TomerShor, line 269: Comment that nuclio Jupyter adds the 'nuclio' suffix automatically

TomerShor (approved):
Nicely done 👌

mlrun/db/httpdb.py
```diff
@@ -1686,7 +1686,7 @@ def get_builder_status(
             last_log_timestamp = float(
                 resp.headers.get("x-mlrun-last-timestamp", "0.0")
             )
-            if func.kind in mlrun.runtimes.RuntimeKinds.nuclio_runtimes():
+            if func.kind in mlrun.runtimes.RuntimeKinds.pure_nuclio_deployed_runtimes():
                 mlrun.runtimes.nuclio.function.enrich_nuclio_function_from_headers(
                     func, resp.headers
                 )
```
rokatyy, line 1692: Why only pure_nuclio? Maybe worth addition comment

rokatyy (approved):
Great job! Having just one question :)

mlrun/db/httpdb.py
alonmr: It's an implementation detail because application runtime uses `get_builder_status` when building the image with mlrun and then uses `get_nuclio_deploy_status` when building with nuclio so I don't want `get_builder_status` to override the `status.container_image` because the headers are not expected in that flow for application runtime.
I will add a comment in my next PR because I don't want to block releasing next RC.

rokatyy: @alonmr gotcha, thanks 👍  Comment is indeed required here, but agree that can be done in the next PR.

PR #6048, opened 2024-07-24
[Artifacts] Ensure user-provided secrets take precedence

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-7258

### DIFF
--- a/server/api/crud/files.py
+++ b/server/api/crud/files.py
@@ -77,12 +77,10 @@ class Files(
 
         logger.debug("Got get filestat request", path=path)
 
-        secrets = secrets or {}
-        secrets.update(server.api.api.utils.get_secrets(auth_info))
-
+        enriched_secrets = self._enrich_secrets_with_auth_info(auth_info, secrets)
         stat = None
         try:
-            stat = store_manager.object(url=path, secrets=secrets).stat()
+            stat = store_manager.object(url=path, secrets=enriched_secrets).stat()
         except FileNotFoundError as exc:
             server.api.api.utils.log_and_raise(
                 HTTPStatus.NOT_FOUND.value, path=path, err=err_to_str(exc)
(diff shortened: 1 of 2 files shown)


tests/api/crud/test_files.py
```diff
@@ -22,41 +22,58 @@
 import server.api.crud
 
 
+@pytest.mark.parametrize(
+    "auth_info, user_secrets, expected_secrets",
```
moranbental: The auth_info parameter is the same for both tests, so you can include it in the test itself rather than in the parametrize decorator.

server/api/crud/files.py
TomerShor, line 1: The same exact logic is also done here: https://github.com/mlrun/mlrun/blob/90a665deeadebd123581fe7fa577632b4b4496b4/server/api/crud/files.py#L81

Move it to a common function and align the logic

server/api/crud/files.py
```diff
@@ -110,9 +110,10 @@ def _delete_artifact_data(
         path = self._resolve_obj_path(schema, path, user)
 
         secrets = secrets or {}
-        secrets.update(server.api.api.utils.get_secrets(auth_info))
+        enriched_secrets = server.api.api.utils.get_secrets(auth_info)
```
TomerShor, line 122: add a comment on priorities

TomerShor (changes requested):

moranbental (changes requested):
Just minor comment

server/api/crud/files.py
rokatyy: @TomerShor nice catch! fixed

TomerShor (approved):
LGTM

server/api/crud/files.py
moranbental: But now the logic for get_filestat has changed
I am not sure if it is should be the same logic

server/api/crud/files.py
```diff
@@ -77,12 +77,10 @@ def _get_filestat(
 
         logger.debug("Got get filestat request", path=path)
 
-        secrets = secrets or {}
-        secrets.update(server.api.api.utils.get_secrets(auth_info))
-
+        enriched_secrets = self._enrich_secrets_with_auth_info(auth_info, secrets)
```
moranbental, line 80: But now the logic for get_filestat has changed
I am not sure if it is should be the same logic

moranbental (changes requested):

server/api/crud/files.py
rokatyy: @moranbental I believe that enrichment should follow the common logic across the entire feature. If we change it in artifact deletion, we should change it here as well.

moranbental (approved):
LGTM

server/api/crud/files.py
alonmr: Just to clarify - generally @moranbental is correct that this changes the behavior and we want to avoid breaking BC. However the reason this is ok is because the current behavior is a bug and not because we want to align with other flows. Of course aligned flows is something that we want but not in the cost of breaking BC. A good reason the change behavior is a bug 🙂

alonmr (approved):
Great job!

PR #6049, opened 2024-07-24
[Run] Function name 'handler' is not allowed when the function kind is 'nuclio:mlrun'

moranbental:
When creating a `nuclio:mlrun` function, an MLRun wrapper called 'handler' is generated. Therefore, if the user creates a function with the name 'handler', it overrides the MLRun wrapper's handler function
To fix this, we check that the code does not include a `def handler` function. The only `def handler` in the code should be the MLRun wrapper's handler (the footer).

https://iguazio.atlassian.net/browse/ML-113

### DIFF
--- a/mlrun/launcher/base.py
+++ b/mlrun/launcher/base.py
@@ -248,8 +248,9 @@ class BaseLauncher(abc.ABC):
         # for other runtimes we need to convert the handler to a string
         if run.spec.handler and runtime.kind not in ["handler", "dask"]:
             run.spec.handler = run.spec.handler_name
+            function_kind = getattr(runtime.spec, "function_kind", None)
             mlrun.utils.helpers.validate_handler_name(
-                runtime.spec.function_kind, run.spec.handler
+                function_kind=function_kind, handler=run.spec.handler
             )
 
         def_name = runtime.metadata.name
--- a/mlrun/launcher/base.py
+++ b/mlrun/launcher/base.py
@@ -248,6 +248,12 @@ class BaseLauncher(abc.ABC):
         # for other runtimes we need to convert the handler to a string
         if run.spec.handler and runtime.kind not in ["handler", "dask"]:
             run.spec.handler = run.spec.handler_name
+            # The name of MLRun's wrapper is 'handler', which is why the handler function name cannot be 'handler'
+            # it would override MLRun's wrapper
+            if runtime.spec.function_kind == "mlrun" and run.spec.handler == "handler":
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    "Handler function name cannot be 'handler'"
+                )
 
         def_name = runtime.metadata.name
         if run.spec.handler_name:
--- a/mlrun/launcher/base.py
+++ b/mlrun/launcher/base.py
@@ -248,10 +248,6 @@ class BaseLauncher(abc.ABC):
         # for other runtimes we need to convert the handler to a string
         if run.spec.handler and runtime.kind not in ["handler", "dask"]:
             run.spec.handler = run.spec.handler_name
-            function_kind = getattr(runtime.spec, "function_kind", None)
-            mlrun.utils.helpers.validate_handler_name(
-                function_kind=function_kind, handler=run.spec.handler
-            )
 
         def_name = runtime.metadata.name
         if run.spec.handler_name:
--- a/mlrun/launcher/base.py
+++ b/mlrun/launcher/base.py
@@ -248,12 +248,9 @@ class BaseLauncher(abc.ABC):
         # for other runtimes we need to convert the handler to a string
         if run.spec.handler and runtime.kind not in ["handler", "dask"]:
             run.spec.handler = run.spec.handler_name
-            # The name of MLRun's wrapper is 'handler', which is why the handler function name cannot be 'handler'
-            # it would override MLRun's wrapper
-            if runtime.spec.function_kind == "mlrun" and run.spec.handler == "handler":
-                raise mlrun.errors.MLRunInvalidArgumentError(
-                    "Handler function name cannot be 'handler'"
-                )
+            mlrun.utils.helpers.validate_handler_name(
+                runtime.spec.function_kind, run.spec.handler
+            )
 
         def_name = runtime.metadata.name
         if run.spec.handler_name:
--- a/mlrun/run.py
+++ b/mlrun/run.py
@@ -738,12 +738,8 @@ def code_to_function(
         filename, handler = ApplicationRuntime.get_filename_and_handler()
 
     is_nuclio, sub_kind = RuntimeKinds.resolve_nuclio_sub_kind(kind)
-    # The name of MLRun's wrapper is 'handler', which is why the handler function name cannot be 'handler'
-    # it would override MLRun's wrapper
-    if sub_kind == "mlrun" and handler == "handler":
-        raise mlrun.errors.MLRunInvalidArgumentError(
-            "Handler function name cannot be 'handler'"
-        )
+
+    mlrun.utils.helpers.validate_handler_name(function_kind=sub_kind, handler=handler)
 
     code_origin = add_name(add_code_metadata(filename), name)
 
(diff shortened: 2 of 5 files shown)


mlrun/launcher/base.py
```diff
@@ -248,6 +248,12 @@ def _enrich_run(
         # for other runtimes we need to convert the handler to a string
         if run.spec.handler and runtime.kind not in ["handler", "dask"]:
             run.spec.handler = run.spec.handler_name
+            # The name of MLRun's wrapper is 'handler', which is why the handler function name cannot be 'handler'
+            # it would override MLRun's wrapper
+            if runtime.spec.function_kind == "mlrun" and run.spec.handler == "handler":
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    "Handler function name cannot be 'handler'"
```
liranbg:
```suggestion
                    "Handler function name 'handler' is preserved. Use different name instead."
```

also, looks like code dup, could it be placed on a single place where both base.py and run.py use and validate their runtime spec?

liranbg (changes requested):

mlrun/launcher/base.py
TomerShor: Reserved* (not preserved)

moranbental: Done. Also this error will be fixed (if needed) by - https://iguazio.atlassian.net/browse/ML-7278

mlrun/launcher/base.py
```diff
@@ -248,6 +248,10 @@ def _enrich_run(
         # for other runtimes we need to convert the handler to a string
         if run.spec.handler and runtime.kind not in ["handler", "dask"]:
             run.spec.handler = run.spec.handler_name
+            function_kind = getattr(runtime.spec, "function_kind", None)
+            mlrun.utils.helpers.validate_handler_name(
+                function_kind=function_kind, handler=run.spec.handler
+            )
```
alonmr: at this point the function is already deployed and you fail validation when running it. 
why not also validate `runtime.spec.default_handler` in the deploy? can be validated in `mlrun.runtimes.nuclio.function.RemoteRuntime.pre_deploy_validation`

moranbental: But the handler name/function is being passed through the run

mlrun/run.py
```diff
@@ -738,6 +738,9 @@ def update_common(fn, spec):
         filename, handler = ApplicationRuntime.get_filename_and_handler()
 
     is_nuclio, sub_kind = RuntimeKinds.resolve_nuclio_sub_kind(kind)
+
+    mlrun.utils.helpers.validate_handler_name(function_kind=sub_kind, handler=handler)
+
```
alonmr: 1. This should be below around:
```
    if is_nuclio:
        runtime = RuntimeKinds.resolve_nuclio_runtime(kind, sub_kind)
        # default_handler is only used in :mlrun sub kind, determine the handler to invoke in function.run()
        runtime.spec.default_handler = handler if sub_kind == "mlrun" else ""
```
2. `code_to_function` is not the only way to create a function. What about `new_function`?

in `new_function` you need to validate the `handler` param as you did here but `new_function` is also used when importing a function. i.e. a function can be created from a dict or a yaml and in that case it will bypass this validation. To handle that you need to validate in the server - in `pre_deploy_validation`

mlrun/launcher/base.py
alonmr: I missed that you added validation on `code_to_function` which is a good start but left a comment there for what is missing

mlrun/run.py
moranbental: When creating a function with `new_function`, we cannot set the function kind to `nuclio:mlrun`, It raises an error:

> Unsupported runtime (nuclio:mlrun) or missing command, supported runtimes: remote,nuclio,serving,dask,job,spark,remote-spark,mpijob,local,databricks,application

moranbental: Or we want to enable it and add the `nuclio:mlrun` kind to `RuntimeKinds`

moranbental: I added the validation above if `is_nuclio:` so that we can use this function in the future to check the handler name for other function kinds, not just for Nuclio. Additionally, this allows us to fail earlier, before building the file.

alonmr: walla. no need to support it then if its not already supported. thanks for explaining 😄

alonmr (approved):

tests/run/test_run.py
```diff
@@ -378,3 +382,50 @@ def test_verify_tag_in_output_for_relogged_artifact(setup_project):
...
+        name="nuclio-mlrun",
+        image="mlrun/mlrun",
+        kind="nuclio:mlrun",
+    )
+
+    with pytest.raises(
+        mlrun.errors.MLRunInvalidArgumentError,
+        match=ERROR_MSG_INVALID_HANDLER_NAME,
+    ):
+        function.run(handler=handler)
```
rokatyy: We cannot run Nuclio function - it can only be deployed
```suggestion
        function.deploy()
```

(and same above)

alonmr: it is possible to run nuclio functions when subkind is mlrun. It essentially invokes the function with mlrun context

rokatyy: @alonmr I just tried to run function:

```
function = project.set_function(
    func="func.py",
    name="nuclio-mlrun-2",
    image="mlrun/mlrun",
    kind="nuclio:mlrun",
    handler="func",
)
function.run()
```

and it doesn't work without deployment...

rokatyy: I think the flow is
1. set_function
2. deploy
3. run

Also, handler replacement in `run` function wouldn't have any effect (and it doesn't make sense for Nuclio runtime). If I try to do so, I got an execution error, but it just says `RunError: bad function response` which is not really explanatory

alonmr: Of course you need to deploy 1st because you need a function to be able to invoke it 🙂 
Specifying a handler should work with mlrun subkind. If it doesn't work might be a bug. I doubt anyone tried it lately

rokatyy: We have discussed with @moranbental all the scenarios. As per our discussion, this test doesn't make sense and will be removed and replaced with another one (with 2 functions to invoke - one with correct name and another one called `handler`).

moranbental: @rokatyy I am currently checking the file code to see if there is a function called `handler` aside from the footer function.

mlrun/run.py
```diff
@@ -781,6 +781,13 @@ def update_common(fn, spec):
             raise ValueError("code_output option is only used with notebooks")
 
     if is_nuclio:
+        # The name of MLRun's wrapper is 'handler', which is why the handler function name cannot be 'handler'
+        # it would override MLRun's wrapper
+        if sub_kind == "mlrun" and code.count("def handler(") > 1:
```
liranbg: that's not enough to count it but to ensure the line starts with it.

liranbg (changes requested):

liranbg (approved):

rokatyy (approved):
🚀 Very nice!

tests/utils/test_helpers.py
```diff
@@ -1040,3 +1040,50 @@ def test_get_artifact_target(kind, tag, target_path, expected):
...
+def handler():
+    pass
+def dummy_handler():
+    def handler():
+        pass
+    handler()
+""",
+    ],
+)
+def test_validate_single_def_handler_valid_handler(code):
```
alonmr, line 1089: please add case for commented out "def handler" function

mlrun/utils/helpers.py
```diff
@@ -1705,6 +1705,31 @@ def is_parquet_file(file_path, format_=None):
     )
 
 
+def validate_handler_name(function_kind: str, handler: str):
+    # The name of MLRun's wrapper is 'handler', which is why the handler function name cannot be 'handler'
+    # it would override MLRun's wrapper
+    if function_kind == "mlrun" and handler == "handler":
+        raise mlrun.errors.MLRunInvalidArgumentError(
+            "The name “handler“ is reserved. Use a different name for your function."
+        )
```
alonmr: This validation is no longer needed and I believe it is better to remove since `handler` is the default value and we don't want to fail when specifying the default handler would go to the footer

alonmr (approved):

PR #6050, opened 2024-07-24
[Context] Fix some `MLClientCtx` related stuff

jond01:
This PR makes some small improvements and fixes related to `MLClientCtx` and `MonitoringApplicationContext`.

* Remove an unused private attribute from `MLClientCtx` and initialize the start and end times to be the same.
* In `MonitoringApplicationContext`, don't use `__post_init__` or `self` in a class method.
* In `get_or_create_ctx` - remove the `labels` argument. It was not used anywhere and was set only after the context was created, resulting in off-sync with the DB saved context.

(Partly related to [ML-6836](https://iguazio.atlassian.net/browse/ML-6836).)

[ML-6836]: https://iguazio.atlassian.net/browse/ML-6836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/run.py
+++ b/mlrun/run.py
@@ -21,6 +21,7 @@ import tempfile
 import time
 import typing
 import uuid
+import warnings
 from base64 import b64decode
 from copy import deepcopy
 from os import environ, makedirs, path
@@ -196,11 +197,12 @@ def load_func_code(command="", workdir=None, secrets=None, name="name"):
 def get_or_create_ctx(
     name: str,
     event=None,
-    spec=None,
+    spec: Optional[dict] = None,
     with_env: bool = True,
     rundb: str = "",
     project: str = "",
-    upload_artifacts=False,
+    upload_artifacts: bool = False,
+    labels: Optional[dict] = None,
 ) -> MLClientCtx:
     """
     Called from within the user program to obtain a run context.
@@ -220,6 +222,7 @@ def get_or_create_ctx(
     :param project:  project to initiate the context in (by default `mlrun.mlconf.default_project`)
     :param upload_artifacts:  when using local context (not as part of a job/run), upload artifacts to the
                               system default artifact path location
+    :param labels: (deprecated - use spec instead) dict of the context labels.
     :return: execution context
 
     Examples::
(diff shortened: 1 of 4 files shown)


mlrun/run.py
```diff
@@ -201,13 +201,13 @@ def get_or_create_ctx(
     rundb: str = "",
     project: str = "",
     upload_artifacts=False,
-    labels: Optional[dict] = None,
```
alonmr, line 204: this needs to be properly deprecated before removing though not exactly sure I understand the reason you want remove it

alonmr (commented):
why do you care if we keep this labels param? it may be used by users

mlrun/run.py
jond01: Why I want to remove it:

1. It was added by us by mistake in 1.5 and is not used.
2. The implementation is faulty (as we discussed) - it adds the labels after saving the context in the DB.
Run the following in Jupyter:
```py
import mlrun

ctx = mlrun.get_or_create_ctx("test-label", labels={"kind": "serving"})
print(ctx.to_yaml())
```
The UI shows:
image
The Jupyter client prints (truncated):
image

This is a bug.

jond01: Regarding proper deprecation, since it's relatively new and was not requested, I think no one has used it.
But we can do it... in this case, a fix may be better.

alonmr: I agree probably no one has used it though we do declare 2 minor versions BC. I understand it's annoying and I agree would be best to just delete but I still think we need to uphold our statements.

jond01: Done in 322febb

mlrun/run.py
```diff
@@ -253,6 +255,20 @@ def get_or_create_ctx(
...
+            "spec={'metadata': {'labels': {'key': 'value'}}}",
+            FutureWarning,
+        )
+        if spec is None:
+            spec = {}
+        if "metadata" not in spec:
+            spec["metadata"] = {}
+        if "labels" not in spec["metadata"]:
+            spec["metadata"]["labels"] = {}
+        spec["metadata"]["labels"].update(labels)
```
alonmr, line 271: when deprecating you need to maintain the previous behavior. Also I don't think we need to add this special handling just for the labels. you can say `use 'context.set_label' instead`

alonmr (approved):

PR #6052, opened 2024-07-25
[Alerts] Specify more relevant info in webhook notification

yanburman:
Currently we send the whole alert which isn't interesting, but omit a few important fields such as name, project, severity and summary

https://iguazio.atlassian.net/browse/ML-7254


alonmr (approved):
please see `mlrun.common.formatters.base.ObjectFormat` possibly can be a good fit here for next PRs

PR #6056, opened 2024-07-28
[Nuclio] Fix nuclio version resolution when determining whether to enable explicit ack

gtopper:
[ML-7284](https://iguazio.atlassian.net/browse/ML-7284)

[ML-7284]: https://iguazio.atlassian.net/browse/ML-7284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -1238,18 +1238,10 @@ class Config:
         return storage_options
 
     def is_explicit_ack_enabled(self) -> bool:
-        version = None
-        if is_running_as_api():
-            import server.api.utils.runtimes.nuclio
-
-            version = server.api.utils.runtimes.nuclio.resolve_nuclio_version()
-
-        if not version:
-            version = self.nuclio_version
-
         return self.httpdb.nuclio.explicit_ack == "enabled" and (
-            not version
-            or semver.VersionInfo.parse(version) >= semver.VersionInfo.parse("1.12.10")
+            not self.nuclio_version
+            or semver.VersionInfo.parse(self.nuclio_version)
+            >= semver.VersionInfo.parse("1.12.10")
         )
 
 
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -1237,9 +1237,16 @@ class Config:
 
         return storage_options
 
-    def is_explicit_ack(self, version=None) -> bool:
+    def is_explicit_ack_enabled(self) -> bool:
+        version = None
+        if is_running_as_api():
+            import server.api.utils.runtimes.nuclio
+
+            version = server.api.utils.runtimes.nuclio.resolve_nuclio_version()
+
         if not version:
             version = self.nuclio_version
+
         return self.httpdb.nuclio.explicit_ack == "enabled" and (
             not version
             or semver.VersionInfo.parse(version) >= semver.VersionInfo.parse("1.12.10")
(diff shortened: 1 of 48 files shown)


mlrun/config.py
```diff
@@ -1237,9 +1237,16 @@ def get_s3_storage_options(self) -> dict[str, typing.Any]:
 
         return storage_options
 
-    def is_explicit_ack(self, version=None) -> bool:
+    def is_explicit_ack_enabled(self) -> bool:
+        version = None
+        if is_running_as_api():
+            import server.api.utils.runtimes.nuclio
+
+            version = server.api.utils.runtimes.nuclio.resolve_nuclio_version()
```
alonmr: importing server from the client is not allowed

gtopper: That's why it's inside `if is_running_as_api()`, to ensure it only runs in server. If that's not a good solution, can you briefly explain why and suggest an alternative?

gtopper: I see that it's enforced by the linter, so I guess I'll move `resolve_nuclio_version` to `mlrun`.

gtopper: I see that the linter enforces this. Not 100% sure how to work around it yet.

gtopper: Moved nuclio version resolution to server init, where it writes the nuclio version to the config, as is done with other configuration such as `config.httpdb.state`.

alonmr (commented):
The way to validate the nuclio version in the client is with `min_nuclio_versions` or `is_nuclio_version_in_range` - I'm trying to understand if this is really the 1st time we need the nuclio version in client code that is also used in the server. We have `server.api.utils.runtimes.nuclio.resolve_nuclio_version` why not set the config there?

gtopper:
> The way to validate the nuclio version in the client is with `min_nuclio_versions`

It doesn't seem to be relevant because we don't want to raise an error, just to turn off a feature switch in the config.

> or `is_nuclio_version_in_range`

It doesn't seem to apply because it's limited to server code.

We can set the config from the result value of `resolve_nuclio_version()` or in the function itself, it doesn't really matter, but either way, we need to ensure that it gets called and isn't left empty in some flows as is the case now.

alonmr (commented):
please remove `cached_nuclio_version` and set the config value instead in `resolve_nuclio_version`. If you need it on startup you can call `resolve_nuclio_version` on startup

server/api/main.py
```diff
@@ -184,6 +185,8 @@ async def setup_api():
     initialize_logs_dir()
     initialize_db()
 
+    server.api.utils.runtimes.nuclio.resolve_nuclio_version()
```
alonmr: can you explain why you need it here?
this method is called in the front end spec and client spec which means when a client tries to connect. shouldn't it be enough or you have to have it at start up?

gtopper: Ok, following the last set of changes, it's not necessary anymore in order to resolve [ML-7284](https://iguazio.atlassian.net/browse/ML-7284). I removed it based on that together with your assessment that it should be covered by the other calls.

alonmr (approved):

PR #6057, opened 2024-07-28
[Datastore] Source case handling while using spark.

tomerm-iguazio:
This bug discovered Due to snowflake transformation to uppercase columns.
we decided to fix it in all spark ingestions, with all types of sources:

1) Add an error in spark ingestion if using wrong letter case (in entity, timestamp_key and label_column) + test.
2) Added a test for ingestion with parquet source with lower case columns and snowflake target.
3) Added a note about SnowflakeSource
4) Block DropFeatures from dropping timestamp_key and label_column in addition to entity.

[ML-7223](https://iguazio.atlassian.net/browse/ML-7223)

[ML-7223]: https://iguazio.atlassian.net/browse/ML-7223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/feature-store/sources-targets.md
```diff
@@ -27,6 +27,11 @@ You can also create a custom `source` to access various databases or data source
 | [Kafka source](#kafka-source)                                                  |Event-based. Sets a Kafka source for the flow (supports both Apache and Confluence Kafka).| Y      | N     | N      |
 | {py:meth}`~mlrun.datastore.StreamSource`                                                       |Event-based. Sets the stream source for the flow. If the stream doesn’t exist it creates it. | Y      | N     | N      |
 
+## Snowflake source
+```{admonition} Note
+Snowflake SQL columns are always set to uppercase letters.
+```
```
theSaarco: Maybe be a bit more informative here? Would be great to add an example code where a `SnowflakeSource` class is used as source and an entity name (or timestamp field etc.) is provided with lowercase letter, explaining that this code will fail? Similar to the example we have in [SQL source](https://docs.mlrun.org/en/latest/feature-store/sources-targets.html#sql-source).

mlrun/datastore/sources.py
```diff
@@ -821,6 +822,36 @@ def __init__(
             **kwargs,
         )
 
+    @classmethod
```
theSaarco: Why a `classmethod`? Usually this is done to generate new instances of the given class, here you don't really care what `cls` is. Looks like this should be a static method.

mlrun/datastore/sources.py
```diff
@@ -821,6 +822,36 @@ def __init__(
...
+    def check_upper_case(
+        cls, timestamp_key: str, entities: list[Union[Entity, str]], label_column: str
+    ):
+        if timestamp_key:
+            if not timestamp_key.isupper():
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    f"Snowflake supports timestamp_key as uppercase only during ingestion."
```
theSaarco:
```suggestion
                    f"Snowflake source only supports uppercase timestamp_key during ingestion."
```

mlrun/datastore/sources.py
```diff
@@ -821,6 +822,36 @@ def __init__(
...
+                    f" timestamp_key: {timestamp_key}"
+                )
+        entity_error = (
+            "Snowflake supports entity as uppercase only during ingestion. entity:"
```
theSaarco:
```suggestion
            "Snowflake source only supports uppercase entity names during ingestion. Entity name:"
```

mlrun/datastore/sources.py
```diff
@@ -821,6 +822,36 @@ def __init__(
...
+            elif isinstance(entity, Entity) and not entity.name.isupper():
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    entity_error + f" {entity.name}"
+                )
+            raise mlrun.errors.MLRunInvalidArgumentError()
+
+        if label_column:
+            if not label_column.isupper():
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    f"Snowflake supports label_column as uppercase only during ingestion. label_column: {label_column}"
```
theSaarco:
```suggestion
                    f"Snowflake source only supports uppercase label_column during ingestion. label_column: {label_column}"
```

theSaarco (dismissed):
Overall looks good - I have some small comments around docs/messages.

mlrun/datastore/sources.py
tomerm-iguazio: switch to different method from utils...

docs/feature-store/sources-targets.md
```diff
@@ -27,6 +27,28 @@ You can also create a custom `source` to access various databases or data source
 | [Kafka source](#kafka-source)                                                  |Event-based. Sets a Kafka source for the flow (supports both Apache and Confluence Kafka).| Y      | N     | N      |
 | {py:meth}`~mlrun.datastore.StreamSource`                                                       |Event-based. Sets the stream source for the flow. If the stream doesn’t exist it creates it. | Y      | N     | N      |
 
+## Snowflake source
+```{admonition} Note
+# Snowflake SQL columns are always set to uppercase letters.
+# This code will fail because timestamp_key is in lowercase:
+
+os.environ["SNOWFLAKE_PASSWORD"] = "*****"
```
assaf758, line 34: Better give a description and  working example and then add a note about uppercase.

tests/system/feature_store/test_spark_snowflake.py
```diff
@@ -208,3 +213,145 @@ def test_purge_with_missing_attribute(self):
             match=".*some attributes are missing.*",
         ):
             fake_target.purge()
+
+    def test_feature_set_wrong_letters_case(self):
```
assaf758: since the fix is generic (not specific to source) no need for specific test for snowflake.

assaf758 (changes requested):

docs/feature-store/sources-targets.md
```diff
@@ -27,6 +27,30 @@ You can also create a custom `source` to access various databases or data source
...
+ 
+feature_set = mlrun.feature_store.FeatureSet("my_fs", entities=[fs.Entity('KEY')], engine="spark")
+df = fs.ingest(feature_set, source=source, targets=[ParquetTarget()], \
+  run_config=mlrun.feature_store.RunConfig(local=False),spark_context=spark_context)
+
+# Pay attention!
+# Snowflake SQL columns are always set to uppercase letters.
+# If we replace 'KEY' with the lowercase 'key', this code will fail and raise an MLRunInvalidArgumentError.
+# The same applies to timestamp_key and label_column.
+```
```
assaf758:
```suggestion
# Notice that by default, Snowflake converts to uppercase name of columns ingested to it.
# The feature-set entity, timestamp_key and label_coumnt must have similar case to the source, othewise the ingest will faile with MLRunInvalidArgumentError exception
```

assaf758 (changes requested):

assaf758 (approved):

PR #6060, opened 2024-07-29
[K8s] List crd paginated - remove log line

roei3000b:
Remove the flooding log line.

https://iguazio.atlassian.net/browse/ML-7309

### DIFF
--- a/server/api/utils/singletons/k8s.py
+++ b/server/api/utils/singletons/k8s.py
@@ -199,10 +199,6 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
                 ),
             )
 
-        if _continue:
-            # we want to show this message only if we needed pagination
-            logger.debug("Finished listing crds")
-
     def create_pod(self, pod, max_retry=3, retry_interval=3):
         if "pod" in dir(pod):
             pod = pod.pod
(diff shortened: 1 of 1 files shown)


server/api/utils/singletons/k8s.py
```diff
@@ -198,7 +198,10 @@ def list_crds_paginated(
                     "remaining_item_count"
                 ),
             )
-        logger.debug("Finished listing crds")
+
+        if _continue:
+            # we want to show this message only if we needed pagination
+            logger.debug("Finished listing crds")
```
TomerShor: I think it needs to be above, in line 193:
```python
if not _continue:
    logger.debug("Finished listing crds")
    break
```

Because only then you actually finish.
If `_continue` it means you have more crds to list -> not finished.

liranbg: I think it just needs to be removed, no reason to log that at all.

liranbg (changes requested):

TomerShor (changes requested):

liranbg (approved):

alonmr (commented):
why not remove the log above as well then?

alonmr (approved):

liranbg (approved):

PR #6062, opened 2024-07-29
[Artifacts] Fix delete artifact uid query param

TomerShor:
Query param is `object-uid` and not `object_uid`.

https://iguazio.atlassian.net/browse/ML-7256


moranbental (approved):
LGTM

alonmr (approved):

PR #6064, opened 2024-07-29
[Application] Fix deploying from image

TomerShor:
Deploying an application runtime from image was impossible because the reverse-proxy source code was always populated at first deployment.

Fix is to set the image on the status before the deployment, so the reverse proxy source code won't be set on the function spec.

Furthermore, make sure that if `spec.image` is set, we do not enrich the source code in the backend.

### DIFF
--- a/mlrun/runtimes/nuclio/application/application.py
+++ b/mlrun/runtimes/nuclio/application/application.py
@@ -367,12 +367,20 @@ class ApplicationRuntime(RemoteRuntime):
         )
 
     def from_image(self, image):
+        """
+        Deploy the function from an existing image
+
+        :param image: image name
+        """
         super().from_image(image)
         # nuclio implementation detail - when providing the image and emptying out the source code and build source,
         # nuclio skips rebuilding the image and simply takes the prebuilt image
         self.spec.build.functionSourceCode = ""
         self.status.application_source = self.spec.build.source
         self.spec.build.source = ""
+
+        # save the image in the status, so we won't repopulate the function source code
+        self.status.container_image = image
 
     @classmethod
     def get_filename_and_handler(cls) -> (str, str):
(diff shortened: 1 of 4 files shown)


mlrun/runtimes/nuclio/application/application.py
```diff
@@ -367,13 +367,21 @@ def with_source_archive(
         )
 
     def from_image(self, image):
+        """
+        Deploy the function from an existing image
```
alonmr:
```suggestion
        Deploy the function with an existing nuclio processor image. This does not applies for the reverse proxy and not the application image.
```

server/api/crud/runtimes/nuclio/function.py
```diff
@@ -347,13 +347,15 @@ def _compile_function_config(
...
+    is_nuclio_image_set = function.spec.config.get("spec.image") is not None
+
     # if the user code is given explicitly or from a source, we need to set the handler and relevant attributes
     if (
         function.spec.base_spec
         or function.spec.build.functionSourceCode
         or function.spec.build.source
         or function.kind == mlrun.runtimes.RuntimeKinds.serving  # serving can be empty
-    ):
+    ) and not is_nuclio_image_set:
```
alonmr: can you explain this? The else clause seems to be weird. why would we want to build file if a nuclio image is set?

alonmr (commented):
thanks <3 minor comment and question

mlrun/runtimes/nuclio/application/application.py
TomerShor: This **does** apply for the reverse proxy...

server/api/crud/runtimes/nuclio/function.py
TomerShor: You're right, I don't want to enter the "else" clause.
The actual meaning was to not call `_set_source_code_and_handler`, so I'll move the condition there.

alonmr (approved):

PR #6065, opened 2024-07-29
[Artifacts] Fix artifact identifier resolution

TomerShor:
When tagging artifacts, we create an `ArtifactIdentifier` with values extracted from the artifact objects, and send it to the API.
The key field of the identifier extracted the artifact's `metadata.key` instead of `spec.db_key`, so it failed to find the artifacts to tag unless they both fields were identical (the fields can be identical if the artifact was logged by a project, but not by a context [run]).

Additionally, in the BE we don't need to lock the db if when tagging artifacts, if there are not artifacts to tag.

https://iguazio.atlassian.net/browse/ML-7320

### DIFF
--- a/mlrun/db/base.py
+++ b/mlrun/db/base.py
@@ -242,9 +242,8 @@ class RunDBInterface(ABC):
             )
             artifact_identifiers.append(
                 mlrun.common.schemas.ArtifactIdentifier(
+                    # we pass the db_key and not the key so the API will be able to find the artifact in the db
                     key=mlrun.utils.get_in_artifact(artifact_obj, "db_key"),
-                    # we are passing tree as uid when storing an artifact, so if uid is not defined,
-                    # pass the tree as uid
                     uid=mlrun.utils.get_in_artifact(artifact_obj, "uid"),
                     producer_id=mlrun.utils.get_in_artifact(artifact_obj, "tree"),
                     kind=mlrun.utils.get_in_artifact(artifact_obj, "kind"),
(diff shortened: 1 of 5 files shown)


server/api/db/sqldb/db.py
```diff
@@ -991,6 +991,15 @@ def tag_artifacts(
...
     ):
         artifacts_keys = [artifact.key for artifact in artifacts]
+        if not artifacts_keys:
+            logger.debug(
+                "No artifacts to tag",
+                project=project,
+                tag=tag_name,
+                artifacts=artifacts,
+            )
+            return
```
liranbg: can that be in a separate pr?

mlrun/db/base.py
```diff
@@ -242,7 +242,7 @@ def _resolve_artifacts_to_tag_objects(
             )
             artifact_identifiers.append(
                 mlrun.common.schemas.ArtifactIdentifier(
-                    key=mlrun.utils.get_in_artifact(artifact_obj, "key"),
+                    key=mlrun.utils.get_in_artifact(artifact_obj, "db_key"),
```
liranbg, line 246: worth a comment on code why do you extract the db_key and not the key or even (logged by project vs logged by run)

liranbg (changes requested):

server/api/db/sqldb/db.py
TomerShor: You got it https://github.com/mlrun/mlrun/pull/6066

liranbg (approved):

PR #6067, opened 2024-07-30
[Alerts] Don't overrun state with reset policy AUTO

yanburman:
When reset policy is auto we were effectively
overriding the reset operation thus counting
previous events towards the alert state.

https://iguazio.atlassian.net/browse/ML-7293

### DIFF
--- a/server/api/crud/alerts.py
+++ b/server/api/crud/alerts.py
@@ -182,6 +182,7 @@ class Alerts(
                 send_notification = True
 
             active = False
+            update_state = True
             if send_notification:
                 state["count"] += 1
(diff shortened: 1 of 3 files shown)


Yacouby (approved):

quaark (approved):

server/api/crud/alerts.py
```diff
@@ -206,7 +208,8 @@ def process_event(
                     active=active,
                 )
 
-            self._states[alert.id] = state_obj
+            if update_state:
+                self._states[alert.id] = state_obj
```
rokatyy: Could you add a comment why we don't update the state here?

PR #6069, opened 2024-07-30
[Artifacts] Remove the default tag value when retrieving an artifact

moranbental:
The problem occurs when trying to query an artifact by key, tree, and iteration without a tag. 
This bug happens because in both the SDK and the server, if the request is sent without a tag, the default tag used is "latest."

To get untagged artifacts, use the enum: - `tag=mlrun.common.schemas.artifact.ArtifactTagsTypes.untagged`

For example to `get_artifact`:
```
artifact = self.project.get_artifact(key, tree=tree, iter=iter, tag=mlrun.common.schemas.artifact.ArtifactTagsTypes.untagged)
```

or to `get_store_resource`:
```
uri = self.project.get_artifact_uri(key, iter=iter, tag=mlrun.common.schemas.artifact.ArtifactTagsTypes.untagged)
uri = uri + f"@{tree}"
store_resource = self.project.get_store_resource(uri)
```
https://iguazio.atlassian.net/browse/ML-7270


alonmr (commented):
Removing the default like that is breaking BC and I doubt it will not break other flows. I'm afraid this requires more thought. How about providing a way to specify no tag? As in tag="" would mean no tag

TomerShor:
I agree with Alon, we can't break BC here.
We can add his suggestion and differentiate between `None` (-> enrich to "latest") , `""` (no tag, do not enrich) and a specific tag

liranbg:
> I agree with Alon, we can't break BC here. We can add his suggestion and differentiate between `None` (-> enrich to "latest") , `""` (no tag, do not enrich) and a specific tag

enum it would be more clear as SDK

server/api/db/sqldb/db.py
```diff
@@ -772,14 +772,15 @@ def read_artifact(
     ):
         query = self._query(session, ArtifactV2, key=key, project=project)
 
-        computed_tag = tag or "latest"
+        tag = tag or mlrun.common.schemas.artifact.ArtifactTagsTypes.latest
```
TomerShor: if `tag=""` it will still be enriched with "latest" ("" is also False)...

TomerShor (changes requested):

mlrun/db/httpdb.py
```diff
@@ -1008,7 +1008,6 @@ def read_artifact(
         """
 
         project = project or mlrun.mlconf.default_project
-        tag = tag or "latest"
```
alonmr, line 1011: add to the docstring how to get untagged version

tests/system/api/test_artifacts.py
```diff
@@ -91,3 +92,45 @@ def test_verify_artifact_tag_in_output(self):
...
+        assert (
+            len(non_latest_artifacts) == 1
+        ), "There should be exactly 1 non-latest artifact"
+
+        artifact = non_latest_artifacts[0]
+        key = artifact["metadata"]["key"]
+        tree = artifact["metadata"]["tree"]
+        iter = artifact["metadata"]["iter"]
+
+        retrieved_artifact = self.project.get_artifact(
```
alonmr: Few comments about the test:
1. you need to verify you got the old artifact and not the last one (possibly change the body between versions)
2. you need to have more than 2 artifacts in the DB. what happens if you have 3 artifacts? should you get 1? I believe the current code will explode in ` db_artifact = query.one_or_none()`
3. Why system test? this can be tested with unit test

moranbental: done

mlrun/common/schemas/artifact.py
```diff
@@ -114,3 +114,13 @@ class ArtifactsDeletionStrategies(mlrun.common.types.StrEnum):
     data_force = "data-force"
     """Delete the artifact data, and if cannot delete it fail the deletion
     and don’t delete the artifact db record"""
+
+
+class ArtifactTagsTypes(mlrun.common.types.StrEnum):
```
TomerShor, line 119:
```suggestion
class ArtifactTagTypes(mlrun.common.types.StrEnum):
```

TomerShor (approved):
Lgtm, minor naming suggestion

PR #6072, opened 2024-07-31
[Automation] Support push images to multi data nodes

alonmr:
Relevant to multi data nodes system where the api image must be pushed to all data nodes for the pods to be able to pull the updated image.

### DIFF
--- a/automation/patch_igz/patch_remote.py
+++ b/automation/patch_igz/patch_remote.py
@@ -49,12 +49,13 @@ class MLRunPatcher:
         self._patch_log_collector = bool(log_collector)
         self._validate_config()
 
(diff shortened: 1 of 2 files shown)


automation/patch_igz/patch_remote.py
```diff
@@ -182,6 +184,34 @@ def _connect_to_node(self, node):
...
+                                [
+                                    "docker",
+                                    "tag",
+                                    built_image,
+                                    replaced_built_image,
+                                ],
+                                live=True,
+                            )
+                            resolve_built_images.append(replaced_built_image)
+                    break
```
liranbg: can you comment the idea behind stopping after finding on node in built image

PR #6075, opened 2024-07-31
[Requirements] Add `google-cloud-bigquery-storage` requirement

gtopper:
to avoid running into an issue (https://github.com/pypa/setuptools/issues/4476) with setuptools.

[ML-7273](https://iguazio.atlassian.net/browse/ML-7273)

[ML-7273]: https://iguazio.atlassian.net/browse/ML-7273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dependencies.py
+++ b/dependencies.py
@@ -58,6 +58,10 @@ def extra_requirements() -> dict[str, list[str]]:
         "google-cloud": [
             "google-cloud-storage==2.14.0",
             "google-cloud-bigquery[pandas, bqstorage]==3.14.1",
+            # google-cloud-bigquery[bqstorage] requires google-cloud-bigquery-storage >= 2.6.0, but older (<2.17)
+            # versions of google-cloud-bigquery-storage runs into an issue
+            # (https://github.com/pypa/setuptools/issues/4476) with setuptools (ML-7273)
+            "google-cloud-bigquery-storage~=2.17",
             "google-cloud==0.34",
         ],
         "google-cloud-storage": ["gcsfs>=2023.9.2, <2024.4"],


dependencies.py
```diff
@@ -58,6 +58,10 @@ def extra_requirements() -> dict[str, list[str]]:
         "google-cloud": [
             "google-cloud-storage==2.14.0",
             "google-cloud-bigquery[pandas, bqstorage]==3.14.1",
+            # google-cloud-bigquery[bqstorage] requires google-cloud-bigquery-storage >= 2.6.0, but older (<2.17)
+            # versions of google-cloud-bigquery-storage runs into an issue
+            # (https://github.com/pypa/setuptools/issues/4476) with setuptools (ML-7273)
+            "google-cloud-bigquery-storage~=2.17",
```
assaf758, line 64: if the issue is with <2.17 why not limit with >=2.17?

gtopper: `~=2.17` is equivalent to `>=2.17,<3`.

assaf758: right.
my point is - if the setuptools bug introduced a need for lower limit for google-cloud-bigquery-storage, why add an upper limit as well?

gtopper: You're right, it's not necessary, because `google-cloud-bigquery-storage<3` is already a requirement of `google-cloud-bigquery[bqstorage]`. So it doesn't really matter either way except for purposes of readability.

assaf758 (approved):

PR #6077, opened 2024-07-31
[API] Fix Delete Project w/ Many Schedules Causes Secret.Updated Event Flood

quaark:
https://iguazio.atlassian.net/browse/ML-7331

During project deletion, we delete the project schedules. The schedules delete process calls 3 methods, 2 of which are `_remove_schedule_secrets` and `_remove_schedule_notification_secrets` which remove project secrets related to the schedule.

1. `_remove_schedule_secrets` is no longer relevant since a couple of years ago, the schedules access keys were moved from the project secrets to a dedicated secret per unique access key (https://github.com/mlrun/mlrun/pull/2695, see PR description for detailed reasoning). So we aren't actually removing any project secrets (see note 3)
2. `_remove_schedule_notification_secrets` doesn't need to be called per schedule, as the delete project will delete the entire secret in the end anyway.
3. Delete project secrets, sends an update request and emits an event **even** when no changes to the secret have been made.

The solution is as follows:
1. Remove `_remove_schedule_secrets`
2. Condition `_remove_schedule_notification_secrets` so it doesn't get called during project delete.
3. Delete project secrets now checks if actual changes should be applied and does nothing if there are no changes.


theSaarco (approved):
LGTM.

liranbg (approved):

PR #6083, opened 2024-08-01
[API] Fix storing monitoring access key during delete project API  

Eyal-Danieli:
**Background**
As part of the delete project API, we delete some of the model monitoring resources that are associated with the related project. As part of the delete flow we are trying to get the model monitoring access key secret. If not exist, we generate and store that access key based on the project owner access key. However, this behaviour has exposed a bug in which a user **without** model monitoring resources tries to delete his project and gets an error during the store process. 

**Solution**
In this PR, we fix that issue by skipping the store process when trying to get the access key during the delete process. If the secret has already been stored, then the API will get the secret anyway. If it hasn't been store, there is no need to store it because it mens that there are no model monitoring resources in this case. In addition, in this case it doesn't make sense to store a secret that will be deleted immediately and anyway won't be used. 

Related JIRA: https://iguazio.atlassian.net/browse/ML-7301

### DIFF
--- a/server/api/api/endpoints/nuclio.py
+++ b/server/api/api/endpoints/nuclio.py
@@ -322,7 +322,9 @@ async def deploy_status(
     )
 
 
-def process_model_monitoring_secret(db_session, project_name: str, secret_key: str):
+def process_model_monitoring_secret(
+    db_session, project_name: str, secret_key: str, store: bool = True
+):
     # The expected result of this method is an access-key placed in an internal project-secret.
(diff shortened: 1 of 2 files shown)


quaark (approved):
LGTM!
Note for sake of documentation: Both this PR and https://github.com/mlrun/mlrun/pull/6068 are 2 parallel (but not mutually exclusive) solutions for the bug. But both solutions are logically sound and should be merged :)

PR #6087, opened 2024-08-01
[Pipelines] Add detailed error message from pipeline step

roei3000b:
Add functionality to `_format_run` function to also add error(if there is one) to the pipeline.
Example:
image

https://iguazio.atlassian.net/browse/ML-6991


server/api/crud/pipelines.py
```diff
@@ -363,6 +371,18 @@ def _resolve_project_from_command(
     def resolve_project_from_pipeline(self, pipeline: PipelineRun):
         return self.resolve_project_from_workflow_manifest(pipeline.workflow_manifest())
 
+    def get_error_from_pipeline(self, kfp_client, run: PipelineRun):
+        detail = kfp_client._run_api.get_run(run.id)
```
liranbg: The whole idea of pipeline_adapter is that we leave the logic of retrieving its internal information to it and not performing any kfp related logic on mlrun side
so, accessing onto internals is an anti pattern
create the logic on both v1.8/v2 adapters where you provide a client and get the run according to its relevant api

liranbg (changes requested):

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/mixins.py
```diff
@@ -94,3 +95,15 @@ def resolve_project_from_workflow_manifest(self, workflow_manifest):
                     raise NotImplementedError(f"Unknown action: {action}")
 
         return mlrun.mlconf.default_project
+
+    @staticmethod
+    def resolve_error_from_pipeline(pipeline):
+        if pipeline.run.status in ["Error", "Failed"]:
```
liranbg:
```suggestion
        if pipeline.run.status in ["Error", "Failed"]:
```

use mlrun_pipelines.comon.models.RunStatuses

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/mixins.py
```diff
@@ -94,3 +95,15 @@ def resolve_project_from_workflow_manifest(self, workflow_manifest):
...
+            workflow_status = json.loads(pipeline.pipeline_runtime.workflow_manifest)[
+                "status"
+            ]
+            for node in workflow_status["nodes"].values():
```
liranbg:
```suggestion
            # status might not be available just yet
            workflow_status = json.loads(pipeline.pipeline_runtime.workflow_manifest).get("status", {})
            for node in workflow_status.get("nodes", {}).values():
```

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/mixins.py
```diff
@@ -94,3 +95,15 @@ def resolve_project_from_workflow_manifest(self, workflow_manifest):
...
+                # The "DAG" node is the parent node of the pipeline so we skip it for getting the detailed error
+                if node["type"] != "DAG":
+                    if node["message"]:
+                        return node["message"]
```
liranbg:
```suggestion
                    if message := message.get("message") and message:
                        return message
```

pipeline-adapters/mlrun-pipelines-kfp-v2/src/mlrun_pipelines/mixins.py
```diff
@@ -78,3 +79,16 @@ def resolve_project_from_workflow_manifest(self, workflow_manifest):
                     raise NotImplementedError(f"Unknown action: {action}")
 
         return mlrun.mlconf.default_project
+
+    @staticmethod
+    def resolve_error_from_pipeline(pipeline):
```
liranbg, line 85: make sure same as v1 for now

server/api/crud/pipelines.py
```diff
@@ -293,18 +294,24 @@ def _format_runs(
...
         self,
         run: PipelineRun,
         format_: mlrun.common.formatters.PipelineFormat = mlrun.common.formatters.PipelineFormat.metadata_only,
+        kfp_client=None,
     ) -> dict:
         run.project = self.resolve_project_from_pipeline(run)
+        if kfp_client:
+            error = self.get_error_from_pipeline(kfp_client, run)
+            if error:
+                run.error = error
```
liranbg, line 314: I think this should be moved to monitor runs

liranbg (changes requested):

server/api/crud/pipelines.py
alonmr: I don't know if I would add it to monitor runs, it's already a heavy job and adding a get run request to pipelines for each step sounds not scaleable

server/api/crud/pipelines.py
```diff
@@ -293,18 +294,24 @@ def _format_runs(
...
         for run in runs:
-            formatted_runs.append(self._format_run(run, format_))
+            formatted_runs.append(self._format_run(run, format_, kfp_client))
         return formatted_runs
 
     def _format_run(
         self,
         run: PipelineRun,
         format_: mlrun.common.formatters.PipelineFormat = mlrun.common.formatters.PipelineFormat.metadata_only,
+        kfp_client=None,
```
alonmr: type hint

liranbg (approved):

PR #6093, opened 2024-08-04
[ProjectSummary] Update the project summary rows rather than deleting the entire table

moranbental:
There was a race condition between refreshing project summaries (by deleting all existing summaries and then recreating them) and creating a project in the database. The race condition occurred as follows:

t0: Start refreshing project summaries by listing all projects and beginning calculations.
t1: Create a project, which also creates a corresponding project summary in the database.
t2: Continue refreshing calculations, which now involves deleting all project summaries from the database.
t3: Write the new calculated project summaries to the database.
As a result, the newly created project summary (from step t1) was deleted during the refresh process. When trying to retrieve this specific project summary, it was not found because the refresh operation had overridden the table.

To resolve this issue, instead of deleting the entire table, we now update existing rows. This approach ensures that when a project is created, its corresponding project summary is also created in the database, and when a project is deleted, its project summary is also removed. This prevents the race condition by maintaining existing summaries while updating or adding new ones as needed.

Also I discovered a bug in our tests where some of our models lack the `get_identifier_string` method. The affected models are `ProjectSummary`, `DatastoreProfile`, `PaginationCache`, `AlertState`, and `TimeWindowTracker`.
To resolve this, I've enforced the implementation of this method by creating an abstract method that all subclasses must implement.


TomerShor (approved):
Looks good!

We might still hit a race condition with the **delete** project flow, where the project is deleted after its summery was queried, and before the transaction ended.
However, as we spoke offline, this might not be a real issue if the UI looks for the summery for each project that is returned in `list_projects`.

server/api/db/sqldb/db.py
```diff
@@ -2503,6 +2503,9 @@ def create_project(self, session: Session, project: mlrun.common.schemas.Project
             session, project_record.name, raise_on_not_found=False
         )
         if not project_summary:
+            logger.debug(
+                "Creating project summary in DB", project_name=project.metadata.name
+            )
```
liranbg: if you delete the table upon project deletion - why do we need to perform get or create rather than create? old copy might have stale records. I think better of creating it because the summary should not be there

moranbental:
> Looks good!
> 
> We might still hit a race condition with the **delete** project flow, where the project is deleted after its summery was queried, and before the transaction ended. However, as we spoke offline, this might not be a real issue if the UI looks for the summery for each project that is returned in `list_projects`.

I don’t think it will be an issue anymore because I am checking that the project summary is not None after I query it.

server/api/db/sqldb/db.py
```diff
@@ -2687,14 +2690,16 @@ def refresh_project_summaries(
...
-                    updated=datetime.now(timezone.utc),
+                query = self._query(
+                    session, ProjectSummary, project=project_summary_schema.name
                 )
-                session.add(project_summary)
+                project_summary = query.one_or_none()
+                if project_summary:
+                    project_summary.summary = project_summary_schema.dict()
+                    project_summary.updated = datetime.now(timezone.utc)
+                    session.add(project_summary)
```
liranbg: create the list of objects and then use `upsert` function which does it for you

liranbg (changes requested):

server/api/db/sqldb/db.py
alonmr: also if you use `upsert` than you don't need to disable auto flush

server/api/db/sqldb/db.py
alonmr: IIRC I added it I think because of error handling since we want to get a conflict on the project if it already exists. I agree we should probably change this. @moranbental can you please verify against `tests.api.api.test_projects.test_create_project_failure_already_exists` that you get the expected result?

moranbental: @alonmr The test failed because the `get_identifier_string` method was missing

liranbg (approved):

server/api/db/sqldb/models.py
```diff
@@ -745,6 +745,9 @@ class ProjectSummary(Base, mlrun.utils.db.BaseModel):
         updated = Column(SQLTypesUtil.datetime())
         summary = Column(JSON)
 
+        def get_identifier_string(self) -> str:
```
alonmr, line 757: ❤️

server/api/db/sqldb/db.py
```diff
@@ -2499,22 +2499,16 @@ def create_project(self, session: Session, project: mlrun.common.schemas.Project
...
-                summary=summary.dict(),
-                updated=datetime.now(timezone.utc),
-            )
-            objects_to_upsert.append(project_summary)
-        self._upsert(session, objects_to_upsert)
+        project_summary = ProjectSummary(
+            project=project.metadata.name,
+            summary=summary.dict(),
+            updated=datetime.now(timezone.utc),
+        )
```
alonmr, line 2510: I do want to say that after rethinking this possibly the more robust solution to create the project summary when a project is created is a DB trigger but I guess we should revisit that only if encounter more bugs

server/api/db/sqldb/db.py
```diff
@@ -2687,17 +2678,18 @@ def refresh_project_summaries(
...
-                    summary=project_summary_schema.dict(),
-                    updated=datetime.now(timezone.utc),
-                )
-                session.add(project_summary)
-            session.commit()
+        project_summaries_to_update = []
+        for project_summary_schema in project_summaries:
+            query = self._query(
+                session, ProjectSummary, project=project_summary_schema.name
+            )
```
alonmr, line 2689: why not get them all in a single query? can be done with `ProjectSummary. project.in_(<names>)`

alonmr (commented):
looks good. small suggestion can improve performance

server/api/db/sqldb/db.py
moranbental: @alonmr because it can lead to a race condition, as @TomerShor mentioned earlier:
> We might still hit a race condition with the **delete** project flow, where the project is deleted after its summery was queried, and before the transaction ended. However, as we spoke offline, this might not be a real issue if the UI looks for the summery for each project that is returned in `list_projects`.

by checking that the project summary is not None after querying, this prevents the issue from occurring.
I will add a comment

quaark: I'm wondering if this fix is worth it, the race condition isn't terrible for UX as the list request indeed only returns summaries of projects that exist (with the sub query).
But it is a leak that can happen. So to ensure we don't leak, it might be better to, after the update finished, query all summaries that don't have projects (you can do a left_join on project and query where the project is None) and delete the result of the query.
That brings us from O(n) queries back to O(1)

quaark (changes requested):
Generally looks good, though I'm not sure I like the fix for the race condition in the refresh

liranbg:
> Generally looks good, though I'm not sure I like the fix for the race condition in the refresh

liked your comment about the O(1), and discussed with @moranbental to execute it on a follow-up pr.
merging away for now

PR #6096, opened 2024-08-04
[Nuclio] Add NS enrichment to Nuclio config

yaelgen:
This PR introduces node selector enrichment for Nuclio config within MLRun. MLRun will now incorporate node selector enrichments directly into the Nuclio config, merging settings from both the project level and the MLRun service level. 
In case of any overlap, node selectors from the MLRun service level will take precedence over those specified at the Nuclio service level.

Although this change is not tied to a specific ticket, it aligns with the approach outlined in [this ticket](https://iguazio.atlassian.net/browse/ML-7290). The agreed approach ensures that MLRun effectively manages node selector enrichment when deploying Nuclio functions through MLRun.


server/api/crud/runtimes/nuclio/function.py
```diff
@@ -493,6 +493,24 @@ def _resolve_and_set_build_requirements_and_commands(function, config):
...
+    if run_db and project_name:
+        if project := run_db.get_project(project_name):
+            project_node_selector = project.spec.default_function_node_selector
+
+    # If either project node selector or config node selector is defined, enrich the config node selector
+    if project_node_selector or config_node_selector:
+        return mlrun.utils.helpers.merge_dicts_with_precedence(
+            config_node_selector, project_node_selector, function_node_selector
+        )
+    return function_node_selector
```
TomerShor: Is there a reason we can't use one of the multiple other versions of this function that already exist?

I know some of the implementations have some runtime-specific intricacies, but adding the same functionality in each runtime is not maintainable at all.
See how can you common-ize these functions as much as you can.

tests/api/runtimes/test_nuclio.py
```diff
@@ -923,6 +923,34 @@ def test_deploy_with_node_selection(self, db: Session, client: TestClient):
             tolerations=tolerations,
         )
 
+    def test_compile_function_config_node_selector_enriched_from_project(
```
TomerShor, line 936: do you want to test subsets (only project / only config) or is it already tested in UTs for `merge_dicts_with_precedence` ?

TomerShor (changes requested):

tests/api/runtimes/test_nuclio.py
yaelgen: In my opinion it is already covered with the UTs for `merge_dicts_with_precedence`, but maybe I will add a case for not having any NS defined to make sure nothing is broken.

TomerShor (approved):
👌

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -663,25 +663,13 @@ def _sanitize_ui_metadata(struct):
...
-            config_node_selector=config_node_selector,
-        )
-        return mlrun.utils.helpers.to_non_empty_values_dict(
-            mlrun.utils.helpers.merge_dicts_with_precedence(
-                config_node_selector,
-                project.spec.default_function_node_selector,
-                function_node_selector,
-            )
-        )
+    function_node_selector = mlrun.utils.helpers.enrich_node_selectors(
```
liranbg: Don't forget to bump version on setup py

liranbg (changes requested):

mlrun/utils/helpers.py
```diff
@@ -1687,6 +1687,24 @@ def validate_component_version_compatibility(
     return True
 
 
+def enrich_node_selectors(
```
alonmr: move to `mlrun.runtimes.utils`

mlrun/utils/helpers.py
```diff
@@ -1687,6 +1687,24 @@ def validate_component_version_compatibility(
...
+    project_node_selector: dict, runtime_node_selector: dict
```
alonmr: here it says `runtime_node_selector` but the log says `Enriching run`.
it should be generic so you don't need to specify it its a runtime or a run

server/api/crud/runtimes/nuclio/function.py
```diff
@@ -493,6 +493,20 @@ def _resolve_and_set_build_requirements_and_commands(function, config):
     mlrun.utils.update_in(config, "spec.build.commands", commands)
 
 
+def _enrich_function_node_selector_with_project(
```
alonmr: you are also enriching with service level here
```suggestion
def _enrich_function_node_selector(
```

alonmr (commented):
Great job! Minor naming comments

rokatyy (approved):

alonmr (approved):
🔥

server/api/crud/runtimes/nuclio/function.py
```diff
@@ -493,6 +493,18 @@ def _resolve_and_set_build_requirements_and_commands(function, config):
     mlrun.utils.update_in(config, "spec.build.commands", commands)
 
 
+def _resolve_node_selector(run_db, project_name, function_node_selector):
+    project_node_selector = {}
+
+    if run_db and project_name:
```
liranbg, line 499: when will run_db be none? mind commenting the flow?

liranbg (commented):
really minor

liranbg (approved):

PR #6097, opened 2024-08-04
[Functions] Remove trailing slash when updating external invocation urls

TomerShor:
Sometimes URLs return with a trailing slash and this causes duplicates in the function external invocations url.
Remove them to not cause duplicated.

Also, when updating the external invocation urls and there isn't anything to updated, no need to commit anything to the DB (skip an un-needed DB call).

### DIFF
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -1922,29 +1922,42 @@ class SQLDB(DBInterface):
             )
             return
 
+        # remove trailing slashes from the URL
+        url = url.rstrip("/")
+
         struct = function.struct
         existing_invocation_urls = struct["status"].get("external_invocation_urls", [])
-        if operation == mlrun.common.types.Operation.ADD:
+        updated = False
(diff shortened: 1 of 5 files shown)


liranbg (approved):

liranbg (commented):
Worth adding a ut coverage for the returned url

server/api/db/sqldb/db.py
```diff
@@ -1922,29 +1922,42 @@ def update_function_external_invocation_url(
             )
             return
 
+        # remove trailing slashes from the URL
+        url = url.rstrip("/")
+
```
rokatyy, line 1927: I think this should be done on client side as well when we generate `invoke_url`

rokatyy:
@TomerShor Actually, we didn't commit anything if api URL was already in the function.. We just unnecessary logged that we will try to do so

rokatyy (approved):

liranbg (approved):

PR #6099, opened 2024-08-04
[Application] Support pre-building reverse proxy image

TomerShor:
Add an option to pre-deploy the reverse proxy image for using it in application runtime.

The 2 main reasons for this are:
1. Speed up deployment - if specifying an existing application image, the build time is only for the reverse proxy. If it is pre-built the function can be deploying much quicker.
2. Using application runtime in air-gapped environments - There is a known issue where golang images cannot be built in an air-gapped environment. To overcome this, users can build the reverse proxy image (golang image) in an exposed environment, and save it in the air-gapped docker registry. This can later be used when creation application functions.

### Usage:
```python
import mlrun

# build the reverse proxy image
mlrun.runtimes.nuclio.application.ApplicationRuntime.deploy_reverse_proxy_image()

# the image is saved on the class:
mlrun.runtimes.nuclio.application.ApplicationRuntime.reverse_proxy_image
```

When the image is set on the class, it will automatically be used by following deployments of instances.

For air-gapped systems, we can also use:
```python
application.from_image(reverse_proxy_image)
```

https://iguazio.atlassian.net/browse/ML-7236

### DIFF
--- a/mlrun/runtimes/nuclio/application/application.py
+++ b/mlrun/runtimes/nuclio/application/application.py
@@ -174,6 +174,7 @@ class ApplicationStatus(NuclioStatus):
 
 class ApplicationRuntime(RemoteRuntime):
     kind = "application"
+    reverse_proxy_image = None
 
     @min_nuclio_versions("1.13.1")
     def __init__(self, spec=None, metadata=None):
@@ -306,7 +307,7 @@ class ApplicationRuntime(RemoteRuntime):
                 show_on_failure=show_on_failure,
             )
 
-        self._ensure_reverse_proxy_configurations()
+        self._ensure_reverse_proxy_configurations(self)
         self._configure_application_sidecar()
 
         # we only allow accessing the application via the API Gateway
@@ -488,6 +489,24 @@ class ApplicationRuntime(RemoteRuntime):
             **http_client_kwargs,
         )
 
+    @classmethod
+    def deploy_reverse_proxy_image(cls):
+        # create a function that includes only the reverse proxy, without the application
+        from mlrun import get_run_db
+        from mlrun.run import new_function
+
+        reverse_proxy_func = new_function(name="reverse-proxy-temp", kind="remote")
+        cls._ensure_reverse_proxy_configurations(reverse_proxy_func)
+        reverse_proxy_func.deploy()
+
+        # save the created container image
+        cls.reverse_proxy_image = reverse_proxy_func.status.container_image
+
+        # delete the function to avoid cluttering the project
+        get_run_db().delete_function(
+            reverse_proxy_func.metadata.name, reverse_proxy_func.metadata.project
+        )
+
     def _run(self, runobj: "mlrun.RunObject", execution):
         raise mlrun.runtimes.RunError(
             "Application runtime .run() is not yet supported. Use .invoke() instead."
@@ -527,21 +546,22 @@ class ApplicationRuntime(RemoteRuntime):
             with_mlrun=with_mlrun,
         )
 
-    def _ensure_reverse_proxy_configurations(self):
-        if self.spec.build.functionSourceCode or self.status.container_image:
+    @classmethod
+    def _ensure_reverse_proxy_configurations(cls, function):
+        if function.spec.build.functionSourceCode or function.status.container_image:
             return
 
-        filename, handler = ApplicationRuntime.get_filename_and_handler()
+        filename, handler = cls.get_filename_and_handler()
         name, spec, code = nuclio.build_file(
             filename,
-            name=self.metadata.name,
+            name=function.metadata.name,
             handler=handler,
         )
-        self.spec.function_handler = mlrun.utils.get_in(spec, "spec.handler")
-        self.spec.build.functionSourceCode = mlrun.utils.get_in(
+        function.spec.function_handler = mlrun.utils.get_in(spec, "spec.handler")
+        function.spec.build.functionSourceCode = mlrun.utils.get_in(
             spec, "spec.build.functionSourceCode"
         )
-        self.spec.nuclio_runtime = mlrun.utils.get_in(spec, "spec.runtime")
+        function.spec.nuclio_runtime = mlrun.utils.get_in(spec, "spec.runtime")
 
     def _configure_application_sidecar(self):
         # Save the application image in the status to allow overriding it with the reverse proxy entry point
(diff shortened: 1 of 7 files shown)


mlrun/runtimes/nuclio/application/application.py
```diff
@@ -488,6 +489,39 @@ def invoke(
...
+        """
+        Build the reverse proxy image and save it.
+        The reverse proxy image is used to route requests to the application sidecar.
+        This is useful when you want to decrease build time by building the application image only once.
+
+        :param use_cache:   Use the cache when building the image
+        """
+        # create a function that includes only the reverse proxy, without the application
+        from mlrun import get_run_db
+        from mlrun.run import new_function
```
moranbental: Why is this import being added inside the function instead of at the top of the file? 
Also, use `import X` instead of `from X import Y`

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -527,21 +561,22 @@ def _build_application_image(
             with_mlrun=with_mlrun,
         )
 
-    def _ensure_reverse_proxy_configurations(self):
-        if self.spec.build.functionSourceCode or self.status.container_image:
+    @classmethod
+    def _ensure_reverse_proxy_configurations(cls, function):
```
moranbental: I think this method should be a static method rather than a class method because it does not change any class state or use any class data. 
Also, `get_filename_and_handler` should also be a static method for the same reason.

moranbental (changes requested):

mlrun/runtimes/nuclio/application/application.py
TomerShor: no actual reason, just copy-paste thing.

moranbental (approved):
LGTM

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -527,21 +562,22 @@ def _build_application_image(
             with_mlrun=with_mlrun,
         )
 
-    def _ensure_reverse_proxy_configurations(self):
-        if self.spec.build.functionSourceCode or self.status.container_image:
+    @staticmethod
+    def _ensure_reverse_proxy_configurations(function):
+        if function.spec.build.functionSourceCode or function.status.container_image:
```
rokatyy: What is the reason of making this method static if it changes the `self`?

TomerShor: I wanted to reuse this in the new class method `deploy_reverse_proxy_image`, where we actually create a `RemoteRuntime` instance. to ensure this is done on this instance, I changed the function to get a function instance as an argument, then it can be a non-application runtime function.
Moving this method to the base `RemoteRuntime` didn't feel right, because it is still related to application runtime.

rokatyy: @TomerShor I see. `self._ensure_reverse_proxy_configurations(self)` looks a bit weird, so maybe worth adding a comment explaining the trick. Also, please add type hint for a `function` arg

server/api/launcher.py
```diff
@@ -280,6 +280,7 @@ def enrich_runtime(
             not runtime.spec.image
             and not runtime.requires_build()
             and runtime.kind in mlrun.mlconf.function_defaults.image_by_kind.to_dict()
+            and not runtime.skip_base_image_enrichment()
```
alonmr: the enrichment is for `spec.image` but method says base image its not accurate

mlrun/runtimes/nuclio/function.py
```diff
@@ -707,6 +707,12 @@ def enable_default_http_trigger(
         """
         self.spec.disable_default_http_trigger = False
 
+    def skip_base_image_enrichment(self):
+        # make sure the API does not enrich the base image with mlrun/mlrun if the function is not a python function
```
alonmr: `mlrun/mlrun` is magic. what if we decide to make a dedicated nuclio image?

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -488,6 +491,39 @@ def invoke(
...
+        reverse_proxy_func = mlrun.run.new_function(
+            name="reverse-proxy-temp", kind="remote"
+        )
+        # default max replicas is 4, we only need one replica for the reverse proxy
+        reverse_proxy_func.spec.max_replicas = 1
+
+        # clear out the base image, so it won't be mlrun/mlrun
+        reverse_proxy_func.set_config("spec.build.baseImage", None)
+        reverse_proxy_func.spec.image = ""
+        reverse_proxy_func.spec.build.base_image = ""
```
alonmr: why is this needed as well as `skip_base_image_enrichment` ?
a more fitting comment would be "the reverse proxy image should not be based on another image" (or use whatever magic nuclio does

mlrun/runtimes/nuclio/function.py
```diff
@@ -707,6 +707,12 @@ def enable_default_http_trigger(
...
+        return (
+            self.spec.nuclio_runtime is None or "python" not in self.spec.nuclio_runtime
```
alonmr: when will nuclio_runtime be None?

server/api/launcher.py
TomerShor: Right, its only because Nuclio refers to `spec.image` as the base image.
I'll change the function name

mlrun/runtimes/nuclio/application/application.py
TomerShor: Because of this: https://github.com/mlrun/mlrun/blob/2bd77151e1c1a1051ddfd141cecc191cc79dbcc2/server/api/crud/runtimes/nuclio/function.py#L686-L690

Updated the comment

mlrun/runtimes/nuclio/function.py
TomerShor: In most cases actually.
If it is not `python`, nuclio-jupyter enriches this field.
If it is python, we resolve the runtime in the server from the config:
https://github.com/mlrun/mlrun/blob/2bd77151e1c1a1051ddfd141cecc191cc79dbcc2/server/api/crud/runtimes/nuclio/function.py#L437-L442

mlrun/runtimes/nuclio/application/application.py
alonmr: yes but its just sanity right? because these should be empty anyway

mlrun/runtimes/nuclio/function.py
alonmr: so if it is None and then in the server we resolve it to python. the image should be enriched no?

mlrun/runtimes/nuclio/application/application.py
TomerShor: Generally the should be empty but nothing stops the user from setting them.
So yes, sanity.

mlrun/runtimes/nuclio/function.py
TomerShor: Right I see it now.
I'll change it to:
```python
return self.spec.nuclio_runtime and "python" not in self.spec.nuclio_runtime
```

alonmr (approved):
Nice! 👍

PR #6105, opened 2024-08-05
[Datastore] Fix Azure upload

tomerm-iguazio:
Azure datastore had problem with uploading large files (10GB+-).

Add BlobServiceClient connection, in order to be able to upload large files with configurable `max_block_size`, `max_single_put_size`, and `max_concurrency`.

Large uploads are successful now, and RAM utilization is under control.

[ML-7381](https://iguazio.atlassian.net/browse/ML-7381)

[ML-7381]: https://iguazio.atlassian.net/browse/ML-7381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


assaf758:
You seem to have the S3 fixes (which were part of another PR) in this PR - please remove.

mlrun/datastore/azure_blob.py
```diff
@@ -26,37 +27,58 @@
...
 
 class AzureBlobStore(DataStore):
     using_bucket = True
 
     def __init__(self, parent, schema, name, endpoint="", secrets: dict = None):
         super().__init__(parent, name, schema, endpoint, secrets=secrets)
+        self._service_client = None
+        self.max_concurrency = MAX_CONCURRENCY
+        self.max_blocksize = MAX_BLOCKSIZE
+        self.max_single_put_size = MAX_SINGLE_PUT_SIZE
```
gtopper: Why copy these constants to self if you're not going to allow for overriding them with a parameter?

mlrun/datastore/azure_blob.py
```diff
@@ -26,37 +27,58 @@
 # Azure blobs will be represented with the following URL: az://<container name>. The storage account is already
 # pointed to by the connection string, so the user is not expected to specify it in any way.
 
+MAX_CONCURRENCY = 100
+MAX_BLOCKSIZE = 1024 * 1024 * 4
+MAX_SINGLE_PUT_SIZE = 1024 * 1024 * 8  # for BlobServiceClient only.
```
gtopper:
```suggestion
MAX_SINGLE_PUT_SIZE = 1024 * 1024 * 8  # for BlobServiceClient only
```
aside from the period at the end, this comment doesn't really clarify when this setting has an effect and when it does not.

mlrun/datastore/azure_blob.py
```diff
@@ -26,37 +27,58 @@
...
     def filesystem(self):
         """return fsspec file system object, if supported"""
         if self._filesystem:
             return self._filesystem
+        self.create_filesystem_and_client()
+        return self._filesystem
+
+    @property
+    def service_client(self):
+        if not self._filesystem or self._service_client:
```
gtopper: Looks like an order of operations bug 🐞 that will cause a reconnect to happen every time. Fix is
```suggestion
        if not self._service_client:
```

mlrun/datastore/azure_blob.py
```diff
@@ -26,37 +27,58 @@
...
+            self.create_filesystem_and_client()
+        return self._service_client
+
+    def create_filesystem_and_client(self):
```
gtopper: Not really sure why these would be created together.

mlrun/datastore/azure_blob.py
```diff
@@ -71,6 +93,82 @@ def get_storage_options(self):
         )
         return self._sanitize_storage_options(res)
 
+    def do_connect(self, storage_options):
+        """Connect to the BlobServiceClient, using user-specified connection details.
```
gtopper: Seems like implementation details that shouldn't be in the docstring.

mlrun/datastore/azure_blob.py
```diff
@@ -71,6 +93,82 @@ def get_storage_options(self):
...
+        Tries connection string first, then credential, then account key, SAS token, and finally anonymous login.
+
+        Raises
+        ------
+        ValueError if none of the connection details are available
```
gtopper: I understand this is copied code, but we have a convention of raising mlrun exceptions in mlrun, such as `MLRunInvalidArgumentError`.

mlrun/datastore/azure_blob.py
```diff
@@ -71,6 +93,82 @@ def get_storage_options(self):
...
+        """
+        # based on do_connect in AzureBlobFileSystem.
```
gtopper:
```suggestion
        # based on do_connect in AzureBlobFileSystem
```
Please include a hyperlink.

tests/integration/azure_blob/test_azure_blob.py
```diff
@@ -246,18 +247,82 @@ def test_list_dir_rm(self, use_datastore_profile, auth_method):
...
         )
         upload_data_item = mlrun.run.get_dataitem(self.object_url, self.storage_options)
         upload_data_item.upload(self.test_file)
 
         response = upload_data_item.get()
         assert response.decode() == self.test_string
 
+    @pytest.mark.parametrize("use_datastore_profile", (True, False))
+    def test_large_upload(self, use_datastore_profile):
+        # Time-consuming test, so every authentication method is checked in test_blob_upload, which is faster.
```
gtopper, line 265: Maybe we can also skip the `use_datastore_profile` parameterization then? I find it hard to think of a scenario where that would matter specifically for large files.

gtopper (changes requested):
Looks nice. Can you add information to the description + ticket regarding the speed up and RAM (since you mentioned it)?

mlrun/datastore/azure_blob.py
tomerm-iguazio: fixed to :
`if not self._filesystem or not self._service_client:`

because I want to make sure they both use the same credentials.

mlrun/datastore/azure_blob.py
tomerm-iguazio: I want to make sure they are use the same credentials, even if there is a change (for example, in the env).

tomerm-iguazio: fixed also in filesystem property.

mlrun/datastore/azure_blob.py
tomerm-iguazio: fixed to comments.

mlrun/datastore/azure_blob.py
gtopper: So better save the `storage_options`. There's no need to tie these two things together, so that if one fails, the other becomes unusable as well.

mlrun/datastore/azure_blob.py
```diff
@@ -32,44 +33,143 @@ class AzureBlobStore(DataStore):
 
     def __init__(self, parent, schema, name, endpoint="", secrets: dict = None):
         super().__init__(parent, name, schema, endpoint, secrets=secrets)
+        self._service_client = None
+        self._storage_options = None
+        self.max_concurrency = 100
+        self.max_blocksize = 1024 * 1024 * 4
+        self.max_single_put_size = 1024 * 1024 * 8  # for BlobServiceClient only
```
gtopper: If the intention is not to allow for overriding them, they should be moved to the class level.

mlrun/datastore/azure_blob.py
```diff
@@ -32,44 +33,143 @@ class AzureBlobStore(DataStore):
...
+        if not self._service_client:
+            self._do_connect()
+        return self._service_client
+
+    def _do_connect(self):
+        # Connect to the BlobServiceClient, using user-specified connection details.
+        # Tries connection string first, then credential, then account key, SAS token, and finally anonymous login.
+        # Raises MLRunInvalidArgumentError if none of the connection details are available
+        # based on do_connect in AzureBlobFileSystem:
+        # https://github.com/fsspec/adlfs/blob/7f06dbdd410224047df09d0ef9e5e9913c64bf7d/adlfs/spec.py#L468
```
gtopper: I think it was fine as a docstring.

gtopper: Please use the correct version.
```suggestion
        # Connect to the BlobServiceClient, using user-specified connection details.
        # Tries connection string first, then credential, then account key, SAS token, and finally anonymous login.
        # Raises MLRunInvalidArgumentError if none of the connection details are available
        # based on do_connect in AzureBlobFileSystem:
        # https://github.com/fsspec/adlfs/blob/2023.9.0/adlfs/spec.py#L422
```

mlrun/datastore/azure_blob.py
gtopper: Seems like this one is still relevant. `BlobServiceClient` is an implementation detail. It's also not technically correct to say that we connect to the client.

mlrun/datastore/azure_blob.py
```diff
@@ -29,47 +30,148 @@
...
+                            credential=cred,
+                            _location_mode="primary",
+                            max_block_size=self.max_blocksize,
+                            max_single_put_size=self.max_single_put_size,
+                        )
+                        for cred in creds
+                        if cred is not None
+                    ][0]
+                elif storage_options.get("sas_token") is not None:
+                    sas_token = storage_options.get("sas_token")
```
gtopper: I suggest moving this to the beginning of the method body and using it everywhere, to avoid repeating `storage_options.get("sas_token")` 4 times.

mlrun/datastore/azure_blob.py
```diff
@@ -29,47 +30,148 @@
...
+                    if not sas_token.startswith("?"):
+                        sas_token = f"?{storage_options.get('sas_token')}"
```
gtopper:
```suggestion
                        sas_token = f"?{sas_token}"
```

gtopper: Or 
```suggestion
                        sas_token = "?" + sas_token
```
if you prefer.

mlrun/datastore/azure_blob.py
```diff
@@ -29,47 +30,148 @@
...
+                        BlobServiceClient(
+                            account_url=account_url,
+                            credential=cred,
+                            _location_mode="primary",
+                            max_block_size=self.max_blocksize,
+                            max_single_put_size=self.max_single_put_size,
+                        )
+                        for cred in creds
+                        if cred is not None
+                    ][0]
```
gtopper: Not sure I understand why we would create N clients and then take the first one?

mlrun/datastore/azure_blob.py
```diff
@@ -29,47 +30,148 @@
...
+                elif storage_options.get("sas_token") is not None:
+                    sas_token = storage_options.get("sas_token")
+                    if not sas_token.startswith("?"):
+                        sas_token = f"?{storage_options.get('sas_token')}"
+                    self._service_client = BlobServiceClient(
+                        account_url=account_url + sas_token,
+                        credential=None,
```
gtopper:
```suggestion
```

mlrun/datastore/azure_blob.py
```diff
@@ -29,47 +30,148 @@
...
+                        _location_mode="primary",
```
gtopper:
```suggestion
```

mlrun/datastore/azure_blob.py
tomerm-iguazio: To me - fix to connect to azure without implementation details.

mlrun/datastore/azure_blob.py
tomerm-iguazio: changed to all repetitive usages.

mlrun/datastore/azure_blob.py
tomerm-iguazio: fixed tnx

mlrun/datastore/azure_blob.py
```diff
@@ -29,47 +30,146 @@
...
+                )
+            elif client_name is not None:
+                account_url: str = f"https://{client_name}.blob.core.windows.net"
+
+                creds = [
+                    credential_from_client_id or credential,
+                    account_key,
+                ]
+                if any(creds):
+                    first_cred = [item for item in creds if item is not None][0]
```
gtopper, line 142:
```suggestion
                cred = credential_from_client_id or credential or account_key,
                if cred:
```

mlrun/datastore/azure_blob.py
```diff
@@ -29,47 +30,146 @@
...
+            )
+        try:
+            if connection_string is not None:
+                self._service_client = BlobServiceClient.from_connection_string(
+                    conn_str=connection_string,
+                    max_block_size=self.max_blocksize,
+                    max_single_put_size=self.max_single_put_size,
+                )
+            elif client_name is not None:
+                account_url: str = f"https://{client_name}.blob.core.windows.net"
```
gtopper:
```suggestion
                account_url = f"https://{client_name}.blob.core.windows.net"
```

mlrun/datastore/azure_blob.py
```diff
@@ -29,47 +30,146 @@
...
+                        max_block_size=self.max_blocksize,
+                        max_single_put_size=self.max_single_put_size,
+                    )
+                else:
+                    # Fall back to anonymous login, and assume public container
+                    self._service_client = BlobServiceClient(
+                        account_url=account_url,
+                        max_block_size=self.max_blocksize,
+                        max_single_put_size=self.max_single_put_size,
+                    )
```
gtopper: I think it can be simplified quite a lot
```suggestion
                cred = credential_from_client_id or credential or account_key
                if not cred and sas_token is not None:
                    if not sas_token.startswith("?"):
                        sas_token = f"?{sas_token}"
                    account_url = account_url + sas_token
                self._service_client = BlobServiceClient(
                    account_url=account_url,
                    credential=cred,
                    max_block_size=self.max_blocksize,
                    max_single_put_size=self.max_single_put_size,
                )
```

mlrun/datastore/azure_blob.py
```diff
@@ -29,47 +30,146 @@
...
+            else:
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    "Must provide either a connection_string or account_name with credentials"
+                )
+
```
gtopper:
```suggestion
```

mlrun/datastore/azure_blob.py
```diff
@@ -29,47 +30,146 @@
...
-        # in order to support az and wasbs kinds.
-        filesystem_class = get_filesystem_class(protocol=self.kind)
-        self._filesystem = makeDatastoreSchemaSanitizer(
-            filesystem_class,
-            using_bucket=self.using_bucket,
-            **self.get_storage_options(),
-        )
+
+        if not self._filesystem:
+            # in order to support az and wasbs kinds.
```
gtopper:
```suggestion
            # in order to support az and wasbs kinds
```

gtopper (approved):

PR #6106, opened 2024-08-05
[K8S] Store secret - fix bug when data is None

roei3000b:
When we store an empty secret, the `k8s_secret.data` will be `None`.
The bug was that we tried to call to `copy` function on `k8s_secret.data` but it was None so we got `AttributeError`.

https://iguazio.atlassian.net/browse/ML-7356

### DIFF
--- a/server/api/utils/singletons/k8s.py
+++ b/server/api/utils/singletons/k8s.py
@@ -533,7 +533,7 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
                     )
                 raise exc
 
-        secret_data = k8s_secret.data.copy()
+        secret_data = (k8s_secret.data or {}).copy()
         for key, value in secrets.items():
             secret_data[key] = base64.b64encode(value.encode()).decode("utf-8")
 
--- a/server/api/utils/singletons/k8s.py
+++ b/server/api/utils/singletons/k8s.py
@@ -597,7 +597,7 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
 
         secret_data = {}
         if secrets:
-            secret_data = k8s_secret.data.copy() if k8s_secret.data else {}
+            secret_data = k8s_secret.data.copy()
             for secret in secrets:
                 secret_data.pop(secret, None)
 
--- a/server/api/utils/singletons/k8s.py
+++ b/server/api/utils/singletons/k8s.py
@@ -496,6 +496,10 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
         :param retry_on_conflict:   if True, will retry to create the secret for race conditions
         :return: returns the action if the secret was created or updated, None if nothing changed
         """
+        if not secrets:
+            # Nothing to store
+            return
+
         namespace = self.resolve_namespace(namespace)
         try:
             k8s_secret = self.v1api.read_namespaced_secret(secret_name, namespace)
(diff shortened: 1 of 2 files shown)


server/api/utils/singletons/k8s.py
```diff
@@ -533,7 +533,7 @@ def store_secrets(
                     )
                 raise exc
 
-        secret_data = k8s_secret.data.copy()
+        secret_data = (k8s_secret.data or {}).copy()
```
TomerShor: you don't need to copy an empty dict, so how about this?
```suggestion
        secret_data = k8s_secret.data.copy() of k8s_secret.data else {}
```

roei3000b: You right about no need to copy an empty dict but I think the way I did it is more compact.

TomerShor: yes it's compact, but you created 2 objects instead of one.

alonmr (commented):
can you explain why the secret was created? I believe we shouldn't have empty project secrets or is there a flow I'm not thinking of?
Also if you get a chance please fix the docstring
```
:param secrets:     the secrets to delete
```

roei3000b:
> can you explain why the secret was created? I believe we shouldn't have empty project secrets or is there a flow I'm not thinking of? Also if you get a chance please fix the docstring
> 
> ```
> :param secrets:     the secrets to delete
> ```

`db.create_project_secrets(project="khsfcrscud", provider=SecretProviderName.kubernetes, secrets={})`
(from the [notebook](https://iguazio.atlassian.net/browse/ML-7356))

alonmr (commented):
what about https://github.com/mlrun/mlrun/blob/development/server/api/utils/singletons/k8s.py#L600

moranbental:
> what about https://github.com/mlrun/mlrun/blob/development/server/api/utils/singletons/k8s.py#L600

This is another bug that I am working on - https://github.com/mlrun/mlrun/pull/6111

server/api/utils/singletons/k8s.py
```diff
@@ -533,7 +533,7 @@ def store_secrets(
                     )
                 raise exc
 
-        secret_data = k8s_secret.data.copy()
+        secret_data = k8s_secret.data.copy() if k8s_secret.data else {}
         for key, value in secrets.items():
```
moranbental, line 542: If the `secrets` parameter is `None`, the code will raise an error when attempting to iterate over `secrets.items()`

moranbental (changes requested):

server/api/utils/singletons/k8s.py
```diff
@@ -533,9 +533,10 @@ def store_secrets(
                     )
                 raise exc
 
-        secret_data = k8s_secret.data.copy()
-        for key, value in secrets.items():
-            secret_data[key] = base64.b64encode(value.encode()).decode("utf-8")
+        secret_data = k8s_secret.data.copy() if k8s_secret.data else {}
+        if secrets:
```
moranbental: Now, if `secrets` is `None`, we are still sending an update event, even though no update should be made. The function should first check if the `secrets` parameter is `None` and return `None` if it is. Only if `secrets` is not `None` should the function proceed with `secret_data = k8s_secret.data.copy() if k8s_secret.data else {}` and call `replace_namespaced_secret`.

I also suggest adding some unit tests for this

moranbental (changes requested):

moranbental (approved):
Great!

liranbg (approved):

PR #6110, opened 2024-08-06
[Model Monitoring] Fix storage size limit for both V3IO KV and MySQL 

Eyal-Danieli:
This PR fix data storage limitations within both V3IO KV and MySQL databases. These issues were exposed when testing a model with 200 features. 

The main changes are: 

- **V3IO KV**:  encode application result into binary string (~64KB -> ~1MB).
- **MySQL**: convert metrics and statistics columns `feature_stats`, `current_stats`, `metrics`, and `result_extra_data` from `TEXT` to `MEDIUMTEXT` (~64KB -> ~16MB).

In addition, we add a new test that validates that a model endpoint can be generated from a model with 500 features. 

Related JIRA: https://iguazio.atlassian.net/browse/ML-6963

### DIFF
--- a/mlrun/model_monitoring/db/stores/sqldb/models/base.py
+++ b/mlrun/model_monitoring/db/stores/sqldb/models/base.py
@@ -23,6 +23,8 @@ from sqlalchemy import (
     Text,
 )
 
+from sqlalchemy.dialects.mysql import MEDIUMTEXT
+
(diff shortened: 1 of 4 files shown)


davesh0812 (approved):
Nice, as talked this fix is relevant only for mysql and for sqlite.

PR #6111, opened 2024-08-06
[K8S] Fix deleting project secret flow

moranbental:
Fixed a bug that occurred when trying to delete project secrets with `secret_data` being None. 
When checking the length of `secret_data` using `if len(secret_data) == len(k8s_secret.data):`, an error was raised `object of type 'NoneType' has no len()`

The fix involves copying only the secrets from the `k8s_secret` that do not exist in the `secrets` dictionary, instead of creating a copy of the entire `k8s_secret` dictionary and deleting the `secrets` from it. 
We then update the `k8s_secret` accordingly.

https://iguazio.atlassian.net/browse/ML-7373

### DIFF
--- a/server/api/utils/singletons/k8s.py
+++ b/server/api/utils/singletons/k8s.py
@@ -595,9 +595,9 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
                 )
                 raise exc
 
-        if k8s_secret.data is None:
+        if not k8s_secret.data:
             logger.debug(
-                "Kubernetes secret data not found",
+                "No data found in the Kubernetes secret",
                 secret_name=secret_name,
             )
             return None
--- a/server/api/utils/singletons/k8s.py
+++ b/server/api/utils/singletons/k8s.py
@@ -595,21 +595,28 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
                 )
                 raise exc
 
-        secret_data = {}
+        # Create a copy of the k8s secret data, filtering out specified secrets if any
         if secrets:
+            secret_data = {
+                key: value
+                for key, value in k8s_secret.data.items()
+                if key not in secrets
+            }
+        else:
             secret_data = k8s_secret.data.copy()
-            for secret in secrets:
-                secret_data.pop(secret, None)
 
+        # Check if there were any changes to the secret data
         if len(secret_data) == len(k8s_secret.data):
             # No secrets were deleted
             return None
 
         if secret_data:
+            # Update the existing secret with modified data
             k8s_secret.data = secret_data
             self.v1api.replace_namespaced_secret(secret_name, namespace, k8s_secret)
             return mlrun.common.schemas.SecretEventActions.updated
 
+        # No secrets left, so delete the secret
         self.v1api.delete_namespaced_secret(secret_name, namespace)
         return mlrun.common.schemas.SecretEventActions.deleted
 
--- a/server/api/utils/singletons/k8s.py
+++ b/server/api/utils/singletons/k8s.py
@@ -605,7 +605,8 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
                 "No data found in the Kubernetes secret",
                 secret_name=secret_name,
             )
-            return None
+            self.v1api.delete_namespaced_secret(secret_name, namespace)
+            return mlrun.common.schemas.SecretEventActions.deleted
 
         # Create a copy of the k8s secret data, filtering out specified secrets if any
         if secrets:
--- a/server/api/utils/singletons/k8s.py
+++ b/server/api/utils/singletons/k8s.py
@@ -595,6 +595,13 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
                 )
                 raise exc
 
+        if k8s_secret.data is None:
+            logger.debug(
+                "Kubernetes secret data not found",
+                secret_name=secret_name,
+            )
+            return None
+
         # Create a copy of the k8s secret data, filtering out specified secrets if any
         if secrets:
             secret_data = {
--- a/tests/api/utils/singletons/test_k8s_utils.py
+++ b/tests/api/utils/singletons/test_k8s_utils.py
@@ -80,6 +80,7 @@ def test_get_logger_pods_label_selector(
     "k8s_secret_data,secrets_data",
     [
         ({"key1": "value1", "key2": "value2"}, []),
+        ({"key1": "value1", "key2": "value2"}, None),
         (None, ["key1"]),
     ],
 )
--- a/tests/api/utils/singletons/test_k8s_utils.py
+++ b/tests/api/utils/singletons/test_k8s_utils.py
@@ -82,6 +82,7 @@ def test_get_logger_pods_label_selector(
         ({"key1": "value1", "key2": "value2"}, []),
         ({"key1": "value1", "key2": "value2"}, None),
         (None, ["key1"]),
+        ({}, ["key1"]),
     ],
 )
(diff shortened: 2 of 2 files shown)


server/api/utils/singletons/k8s.py
```diff
@@ -595,21 +595,35 @@ def delete_secrets(
...
                 raise exc
 
-        secret_data = {}
+        if k8s_secret.data is None:
+            logger.debug(
+                "Kubernetes secret data not found",
+                secret_name=secret_name,
+            )
+            return None
+
```
yaelgen, line 610:
```suggestion
if not k8s_secret.data:
    logger.debug(
        "No data found in the Kubernetes secret",
        secret_name=secret_name,
    )
    return
```

tests/api/utils/singletons/test_k8s_utils.py
```diff
@@ -58,8 +68,68 @@ def test_get_logger_pods_label_selector(
...
+        "my-secret", k8s_helper.namespace
+    )
+    k8s_secret_mock.data = {"key2": "value2"}
+    k8s_helper.v1api.replace_namespaced_secret.assert_called_once_with(
+        "my-secret", k8s_helper.namespace, k8s_secret_mock
+    )
+    k8s_helper.v1api.delete_namespaced_secret.assert_not_called()
+
+
+def test_delete_secrets_delete_secret(k8s_helper):
```
yaelgen:
```suggestion
def test_ensure_secret_deletion_with_delete_secrets(k8s_helper):
```

tests/api/utils/singletons/test_k8s_utils.py
```diff
@@ -58,8 +68,68 @@ def test_get_logger_pods_label_selector(
...
+
+@pytest.mark.parametrize(
+    "k8s_secret_data,secrets_data",
+    [
+        ({"key1": "value1", "key2": "value2"}, []),
+        ({"key1": "value1", "key2": "value2"}, None),
+        (None, ["key1"]),
+        ({}, ["key1"]),
+    ],
+)
```
yaelgen: Maybe add a case where they both exists but the keys don't overlap? just to make sure it's not breaking anything

tests/api/utils/singletons/test_k8s_utils.py
```diff
@@ -58,8 +68,68 @@ def test_get_logger_pods_label_selector(
...
+def test_delete_secrets_no_changes(k8s_helper, k8s_secret_data, secrets_data):
```
yaelgen:
```suggestion
def test_no_secret_changes_when_keys_do_not_overlap(k8s_helper, k8s_secret_data, secrets_data):
```

yaelgen (changes requested):
Nice! a few suggestions below :)

tests/api/utils/singletons/test_k8s_utils.py
moranbental: @yaelgen I prefer when test names start with the function being tested, like `test_delete_secrets`
This makes it easier to find tests related to the deletion functionality. How about naming this test `test_delete_secrets_no_secret_changes_when_keys_do_not_overlap`?

tests/api/utils/singletons/test_k8s_utils.py
moranbental: @yaelgen For the same reason below - how about that? 
`test_delete_secrets_confirms_secret_deletion_when_keys_match`

tests/api/utils/singletons/test_k8s_utils.py
yaelgen: Actually, I'm thinking maybe this all can fit into a single combined `test_delete_secrets`  to cover all scenarios, wdyt?

If you prefer separate tests, both suggested names work for me. I would prefer: `test_delete_secrets_no_changes_with_no_key_overlap` and `test_delete_secrets_confirms_deletion_for_matching_keys`
IMO it makes it  a bit cumbersome, but np.

server/api/utils/singletons/k8s.py
```diff
@@ -595,21 +595,35 @@ def delete_secrets(
                 )
                 raise exc
 
-        secret_data = {}
+        if not k8s_secret.data:
+            logger.debug(
+                "No data found in the Kubernetes secret",
+                secret_name=secret_name,
+            )
+            return None
```
yaelgen: No need for the explicit `None` here, you can simply `return`

yaelgen (changes requested):

server/api/utils/singletons/k8s.py
moranbental: Here we check the action returned by the function. I'm unsure if modifying this will raise an error
https://github.com/mlrun/mlrun/blob/30aadb6f3d83b27db15c5704d6c0be0d7503032d/server/api/crud/projects.py#L483

tests/api/utils/singletons/test_k8s_utils.py
```diff
@@ -58,8 +68,71 @@ def test_get_logger_pods_label_selector(
...
+    result = k8s_helper.delete_secrets("my-secret", ["key1"])
+    assert result == mlrun.common.schemas.SecretEventActions.deleted
+
+    k8s_helper.v1api.read_namespaced_secret.assert_called_once_with(
+        "my-secret", k8s_helper.namespace
+    )
+    k8s_helper.v1api.delete_namespaced_secret.assert_called_once_with(
+        "my-secret", k8s_helper.namespace
+    )
+    k8s_helper.v1api.replace_namespaced_secret.assert_not_called()
```
rokatyy: Why using `assert_called` if you can just check output of `delete_secrets` (e.g. updated/deleted/none) and then you won't need 3 tests and will be able to create one small test :)

rokatyy (commented):
LGTM, minor suggestion regarding tests

server/api/utils/singletons/k8s.py
yaelgen: It should work the same in terms of behavior, it will return a `NoneType` value without the explicit None in the return, 
I look at `return` as sort of a "break" of the function, since it has nothing to delete here and no real return value in this case, no action was taken. but since there are other cases with a return value, I guess both is okay.

yaelgen (approved):
LGTM

rokatyy (approved):
Neat!

server/api/utils/singletons/k8s.py
```diff
@@ -600,21 +600,35 @@ def delete_secrets(
...
```
alonmr: if you return here you will not delete the k8s secret 🤔

tests/api/utils/singletons/test_k8s_utils.py
```diff
@@ -68,11 +68,12 @@ def test_get_logger_pods_label_selector(
     if extra_selector:
         selector += f",{extra_selector}"
 
-    k8s_helper = server.api.utils.singletons.k8s.K8sHelper(namespace, silent=True)
     k8s_helper.list_pods = unittest.mock.MagicMock()
```
alonmr, line 71: is this needed after you added the fixture?

alonmr (changes requested):
Seems to me you have a bug

server/api/utils/singletons/k8s.py
moranbental: You’re right!

tests/api/utils/singletons/test_k8s_utils.py
moranbental: @alonmr What do you mean? if we need to mock list_pods? yes

alonmr (approved):

PR #6113, opened 2024-08-06
[ProjectSummary] Improving refresh_project_summaries by ensuring orphaned project summaries are deleted

moranbental:
To optimize `refresh_project_summaries`, we query all project summaries in a single query.
We update only the summaries of projects that have associated projects.
This optimization reduces query complexity from O(n) to O(1).

Additionally, to prevent race conditions where a project might be deleted after its summary is queried but before the transaction completes, we now delete project summaries that have no associated projects.

Also, when deleting a project, first delete the project summary and then the project itself.

related pr - https://github.com/mlrun/mlrun/pull/6093

### DIFF
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -2521,7 +2521,7 @@ class SQLDB(DBInterface):
     @staticmethod
     def _append_project_summary(project, objects_to_store):
         summary = mlrun.common.schemas.ProjectSummary(
-            project=project.metadata.name,
+            name=project.metadata.name,
         )
         project_summary = ProjectSummary(
             project=project.metadata.name,
@@ -2594,7 +2594,7 @@ class SQLDB(DBInterface):
         logger.debug(
             "Deleting project from DB", name=name, deletion_strategy=deletion_strategy
         )
-        self._delete_project_summary(name)
+        self._delete_project_summary(session, name)
         self._delete(session, Project, name=name)
 
     def list_projects(
@@ -2653,7 +2653,7 @@ class SQLDB(DBInterface):
                 f"Project summary not found: {project=}"
             )
 
-        project_summary_record.summary["project"] = project_summary_record.project
+        project_summary_record.summary["name"] = project_summary_record.project
         project_summary_record.summary["updated"] = project_summary_record.updated
         return mlrun.common.schemas.ProjectSummary(**project_summary_record.summary)
 
@@ -2703,9 +2703,7 @@ class SQLDB(DBInterface):
         and removes project summaries that no longer have associated projects.
         """
 
-        summary_dicts = {
-            summary.project: summary.dict() for summary in project_summaries
-        }
+        summary_dicts = {summary.name: summary.dict() for summary in project_summaries}
 
         # Create a query for project summaries with associated projects
         existing_summaries_query = (
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -2594,8 +2594,8 @@ class SQLDB(DBInterface):
         logger.debug(
             "Deleting project from DB", name=name, deletion_strategy=deletion_strategy
         )
+        self._delete_project_summary(name)
         self._delete(session, Project, name=name)
-        self._delete(session, ProjectSummary, project=name)
 
     def list_projects(
         self,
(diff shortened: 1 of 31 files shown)


server/api/db/sqldb/db.py
```diff
@@ -2692,22 +2692,48 @@ def refresh_project_summaries(
...
-            # that the project summary is not `None` after querying.
-            # This is why we cannot query all project summaries outside the loop.
-            query = self._query(
-                session, ProjectSummary, project=project_summary_schema.name
+        """
+        This method updates the summaries of projects that have associated projects in the database
+        and removes project summaries that no longer have associated projects.
+        """
+
+        summary_dict = {summary.name: summary.dict() for summary in project_summaries}
```
```suggestion
        summary_dicts = {summary.name: summary.dict() for summary in project_summaries}
```

TomerShor: Another thing, not super related, but it is a bit confusing that `models.ProjectSummary` has a `project` attribute, but `schemas.ProjectSummary` has a `name` attribute, which both mean the same.
I think we should align them both to `project`

server/api/db/sqldb/db.py
```diff
@@ -2692,22 +2692,48 @@ def refresh_project_summaries(
...
             )
-            project_summary = query.one_or_none()
-            if project_summary:
-                project_summary.summary = project_summary_schema.dict()
-                project_summary.updated = datetime.now(timezone.utc)
-                project_summaries_to_update.append(project_summary)
 
-        self._upsert(session, project_summaries_to_update)
+            for project in orphaned_summaries:
+                session.delete(project)
```
```suggestion
            for summary in orphaned_summaries:
                session.delete(summary)
```

server/api/db/sqldb/db.py
```diff
@@ -2692,22 +2692,48 @@ def refresh_project_summaries(
...
+            project_summary.summary = summary_dict.get(project_summary.project)
+            project_summary.updated = datetime.now(timezone.utc)
+            session.add(project_summary)
+
+        # To avoid race conditions where a project might be deleted after its summary is queried
+        # but before the transaction completes, we delete project summaries that do not have
+        # any associated projects.
+        orphaned_summaries = existing_summaries_query.filter(
+            Project.id.is_(None)
+        ).all()  # No associated project
```
```suggestion
        ).all()
```

server/api/db/sqldb/db.py
```diff
@@ -2692,22 +2692,48 @@ def refresh_project_summaries(
...
+        for project_summary in associated_summaries:
+            project_summary.summary = summary_dict.get(project_summary.project)
+            project_summary.updated = datetime.now(timezone.utc)
+            session.add(project_summary)
+
+        # To avoid race conditions where a project might be deleted after its summary is queried
+        # but before the transaction completes, we delete project summaries that do not have
+        # any associated projects.
+        orphaned_summaries = existing_summaries_query.filter(
+            Project.id.is_(None)
```
TomerShor: In the outer join we filter on `Project.name`. if there is no project, will we still get it even if we don't check for `id.is_(None)`?

TomerShor (changes requested):
Some minor suggestions

server/api/db/sqldb/db.py
moranbental: we need that. also, add a unit test to verify that the flow is working as expected.

server/api/db/sqldb/db.py
moranbental: > Another thing, not super related, but it is a bit confusing that `models.ProjectSummary` has a `project` attribute, but `schemas.ProjectSummary` has a `name` attribute, which both mean the same.
> I think we should align them both to `project`

@TomerShor I can't do that, it is not BC

TomerShor: right, only the model is new...

server/api/db/sqldb/db.py
```diff
@@ -5907,6 +5936,14 @@ def delete_table_records_by_name(
             table_name=sanitized_table_name,
         )
 
+    def _delete_project_summary(
+        self,
+        session: Session,
+        name: str,
+    ):
+        logger.debug("Deleting project summary from DB", name=name)
+        self._delete(session, ProjectSummary, project=name)
```
TomerShor: move this function up, below the rest of the public `project_summary` methods.
The order here is not exactly public then private, we strive to keep the resource-related methods in the same area.

TomerShor (changes requested):

TomerShor (approved):

PR #6116, opened 2024-08-06
[Docs] Add typical error when updating a function in Nuclio, and not in MLRun

jillnogold:
[ML-6613](https://iguazio.atlassian.net/issues/ML-6613)

[ML-6613]: https://iguazio.atlassian.net/browse/ML-6613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/concepts/nuclio-real-time-functions.ipynb
```diff
@@ -29,7 +29,9 @@
     "```{admonition} Important\n",
     "When working with Nuclio functions (remote / serving / application runtimes) and API Gateways in MLRun, any modifications/deletions should be made in MLRun only. This is also relevant for modifying the spec of a function or API Gateway.\n",
     "<b>Changes that are made in Nuclio are not guaranteed to be synced to MLRun and can cause unexpected behavior.</b> </br>\n",
-    "For example, if you create a remote function in MLRun (which is created in Nuclio) and then delete it in Nuclio, the function status might show as ready, but invoking it will fail.\n",
+    "For example, if you create a remote function in MLRun (which is created in Nuclio) and then delete it in Nuclio, the function status might show as ready, but invoking it will fail.<br>\n",
+    "If you get an error similar to: `Error when translating nuclio names to mlrun names in api gateway: number of functions doesn't match the mlrun functions in annotation`\n",
+    "it means that the API gateway was updated at least once manually in Nuclio itself, without using the MLRun SDK.\n",
```
rokatyy: I think we should propose a solution for the problem as well. And the solution in this case is to delete gw and start from scratch

rokatyy (approved):

docs/change-log/index.md
```diff
@@ -11,7 +11,114 @@
...
+|ML-5656|Add Snowflake as offline target store. See [Snowflake target](../feature-store/sources-targets.html#snowflake-target).|
+
+### Feature store
+| ID    |Description                                                                 |
+|-------|----------------------------------------------------------------------------|
+|ML-3303|Optimized the parquet read when the partitioning is on fields other than the timestamp.|
+|ML-5656|Add Snowflake as an offline target store. See [Snowflake target](../feature-store/sources-targets.html#snowflake-target)|
+
+### Model monitoring
+
```
TomerShor: remove empty line

docs/change-log/index.md
```diff
@@ -11,7 +11,114 @@
...
+
+### Projects
+| ID    |Description                                                                 |
+|-------|----------------------------------------------------------------------------|
+|ML-2585|When deleting a project, you can also delete its artifacts. See [Deleting artifacts](..store/artifacts.html##deleting-artifacts).|
+|ML-3874|Node selector can now be defined in the project spec. See [Node selection](../runtimes/configuring-job-resources.html#node-selection).|
+
+### Runtimes
+
```
TomerShor: remove empty line

TomerShor (changes requested):

TomerShor (approved):

PR #6118, opened 2024-08-07
[Artifacts] Error message when attempting to delete DatasetArtifact data including several files

moranbental:
update error message 
https://iguazio.atlassian.net/browse/ML-7070

### DIFF
--- a/server/api/crud/artifacts.py
+++ b/server/api/crud/artifacts.py
@@ -329,7 +329,7 @@ class Artifacts(
                 path
             ):
                 raise mlrun.errors.MLRunNotImplementedServerError(
-                    "Deleting artifact data of kind 'dataset' is currently supported for a single file only"
+                    "You can only delete artifact data of the kind 'dataset' if it is a single file"
                 )
 
             server.api.crud.Files().delete_artifact_data(
--- a/server/api/crud/artifacts.py
+++ b/server/api/crud/artifacts.py
@@ -329,7 +329,7 @@ class Artifacts(
                 path
             ):
                 raise mlrun.errors.MLRunNotImplementedServerError(
-                    "You can only delete artifact data of the kind 'dataset' if it is a single file"
+                    "You can only delete the data of an artifact of kind 'dataset' if it is a single file"
                 )
 
             server.api.crud.Files().delete_artifact_data(


server/api/crud/artifacts.py
```diff
@@ -329,7 +329,7 @@ def _delete_artifact_data(
                 path
             ):
                 raise mlrun.errors.MLRunNotImplementedServerError(
-                    "Deleting artifact data of kind 'dataset' is currently supported for a single file only"
+                    "You can only delete artifact data of the kind 'dataset' if it is a single file"
```
quaark:
```suggestion
                    "You can only delete the data of an artifact of kind 'dataset' if it is a single file"
```

quaark (changes requested):
Slight grammar fix

server/api/crud/artifacts.py
moranbental: The error message was updated by @jillnogold

TomerShor: While grammatically correct, the meaning is a bit different:
- `delete artifact data of the kind 'dataset'` - means the the kind of the **data** is 'dataset'
- `delete the data of an artifact of kind 'dataset'` - means the kind of the **artifact** is 'dataset'

The second one is the correct meaning in this case.

TomerShor (approved):

server/api/crud/artifacts.py
```diff
@@ -329,7 +329,7 @@ def _delete_artifact_data(
                 path
             ):
                 raise mlrun.errors.MLRunNotImplementedServerError(
-                    "Deleting artifact data of kind 'dataset' is currently supported for a single file only"
+                    "You can only delete the data of an artifact of kind 'dataset' if it is a single file"
```
alonmr, line 332: @jillnogold we want code errors to start with a verb, unlike docs it doesn't make sense to start an error message with `You can` or `You can't`. we want to say what is allowed in general without addressing the user

alonmr (changes requested):
Comment for @jillnogold

server/api/crud/artifacts.py
jillnogold, line 332: Aha - so the original message is perfect

jillnogold, line 332: "Deleting artifact data of kind 'dataset' is currently supported for a single file only"

moranbental:
No need to change the error message

PR #6125, opened 2024-08-07
[Notifications] Fix notification schema condition `Optional` type hint

quaark:
https://iguazio.atlassian.net/browse/ML-7253
https://iguazio.atlassian.net/browse/ML-7389

It's not enough to supply pydantic with a `None` default, we need to have the type hint as `Optional` as well.
Added thorough schema description as well.


TomerShor:
> It's not enough to supply pydantic with a `None` default, we need to have the type hint as `Optional` as well.

@quaark isn't that true for the rest of the `None` defaults as well then?

TomerShor (approved):

PR #6127, opened 2024-08-07
[Model Monitoring] Skipping deletion of stream resources when the project does not exist in Iguazio

moranbental:
A new bug occurs when a project is deleted from Nuclio (while MLRun is disabled). In this case, the project is removed from Nuclio and Iguazio, but it still exists in MLRun. When MLRun is re-enabled, the sync process runs and moves the project to the archive. The trying to delete the project from MLRun and all its associated resources, including model monitoring resources.During this process, model monitoring tries to retrieve the project secret. If no secret value is found, it attempts to get the project owner to retrieve the access key. To find the project owner, a request is sent to Iguazio to get the project details. However, the project does not exist in Iguazio, leading to an error.

To fix this, I added a try-except block to handle the `mlrun.errors.MLRunNotFoundError` exception.
In this case, we will not continue with the deletion of the model monitoring stream resources

https://iguazio.atlassian.net/browse/ML-7365

### DIFF
--- a/server/api/api/endpoints/nuclio.py
+++ b/server/api/api/endpoints/nuclio.py
@@ -364,9 +364,16 @@ def process_model_monitoring_secret(
             allow_internal_secrets=True,
         )
         if not secret_value:
-            project_owner = server.api.utils.singletons.project_member.get_project_member().get_project_owner(
-                db_session, project_name
-            )
+            try:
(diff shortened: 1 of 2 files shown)


quaark (approved):

Eyal-Danieli (approved):
LGTM

PR #6128, opened 2024-08-07
[Serving] Fix setting of drain/termination callbacks

gtopper:
Broken in #6044.

[ML-7304](https://iguazio.atlassian.net/browse/ML-7304)

[ML-7304]: https://iguazio.atlassian.net/browse/ML-7304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/serving/states.py
+++ b/mlrun/serving/states.py
@@ -1391,7 +1391,7 @@ class FlowStep(BaseStep):
         return step
 
     def supports_termination(self):
-        return self.engine == "async"
+        return self.engine != "sync"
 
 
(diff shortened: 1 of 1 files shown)


assaf758:
Since the ingestion service is using storey, why not set the "engine" to "async"?
Alternatively (if there are reasons to differentiate between flow-async and ingestion-service) how about defining new flow engine type, instead of using None…

assaf758 (approved):

PR #6129, opened 2024-08-07
[ProjectSummary] Create new data migration for creating project summaries

roei3000b:
When upgrading MLRun from a version that did not have project summaries DB table, to a version that has them, summaries for the existing projects are not created, causing the UI to fail on: `project summary not found for project X`.

This means that we cannot rely on the project creation flow to create the relevant summary.
Here we added new data migration step where we create project summary record for each existing project.
This summaries can be empty as they will be recalculated during the refresh.

### DIFF
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -2496,7 +2496,9 @@ class SQLDB(DBInterface):
         self._upsert(session, tags)
 
     # ---- Projects ----
-    def create_project(self, session: Session, project: mlrun.common.schemas.Project):
+    def create_project(
+        self, session: Session, project: mlrun.common.schemas.Project, with_summary=True
+    ):
         logger.debug("Creating project in DB", project_name=project.metadata.name)
         created = datetime.utcnow()
         project.metadata.created = created
@@ -2514,15 +2516,19 @@ class SQLDB(DBInterface):
         labels = project.metadata.labels or {}
         update_labels(project_record, labels)
 
-        summary = mlrun.common.schemas.ProjectSummary(
-            name=project.metadata.name,
-        )
-        project_summary = ProjectSummary(
-            project=project.metadata.name,
-            summary=summary.dict(),
-            updated=datetime.now(timezone.utc),
-        )
-        self._upsert(session, [project_record, project_summary])
+        objects_to_save = [project_record]
+        if with_summary:
+            summary = mlrun.common.schemas.ProjectSummary(
+                name=project.metadata.name,
+            )
+            project_summary = ProjectSummary(
+                project=project.metadata.name,
+                summary=summary.dict(),
+                updated=datetime.now(timezone.utc),
+            )
+            objects_to_save.append(project_summary)
+
+        self._upsert(session, objects_to_save)
 
     @retry_on_conflict
     def store_project(
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -2496,9 +2496,7 @@ class SQLDB(DBInterface):
         self._upsert(session, tags)
 
     # ---- Projects ----
-    def create_project(
-        self, session: Session, project: mlrun.common.schemas.Project, with_summary=True
-    ):
+    def create_project(self, session: Session, project: mlrun.common.schemas.Project):
         logger.debug("Creating project in DB", project_name=project.metadata.name)
         created = datetime.utcnow()
         project.metadata.created = created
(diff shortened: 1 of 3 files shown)


server/api/initial_data.py
```diff
@@ -867,6 +871,30 @@ def _migrate_model_monitoring_jobs(db, db_session):
...
+    )
+    for project_name in projects.projects:
+        summary = mlrun.common.schemas.ProjectSummary(
+            name=project_name,
+        )
+        project_summary = ProjectSummary(
+            project=project_name,
+            summary=summary.dict(),
+        )
+        db._upsert(db_session, [project_summary], ignore=True)
```
moranbental:
```suggestion
    project_summaries = [
        ProjectSummary(
            project=project_name,
            summary=mlrun.common.schemas.ProjectSummary(name=project_name).dict()
        )
        for project_name in projects.projects
    ]
    
    db._upsert(db_session, project_summaries, ignore=True)
```

This way, you commit all changes together.

tests/api/test_initial_data.py
```diff
@@ -200,6 +208,24 @@ def test_add_default_hub_source_if_needed():
...
 
 
+def test_migrate_project_summaries():
+    db, db_session = _initialize_db_without_migrations()
+
+    # Create a project
+    project = mlrun.common.schemas.Project(
+        metadata=mlrun.common.schemas.ProjectMetadata(name="project-name"),
+    )
+    db.create_project(db_session, project)
```
moranbental: What is the purpose of this test? When you create a project, it also creates the project summary.

moranbental (changes requested):
Nice! few comments

quaark (approved):
Agree with @moranbental's comments. Other than that, LGTM

quaark (changes requested):
Moving the PR to request changes status, to not merge by accident :)

server/api/db/sqldb/db.py
```diff
@@ -2496,7 +2496,9 @@ def tag_objects_v2(
         self._upsert(session, tags)
 
     # ---- Projects ----
-    def create_project(self, session: Session, project: mlrun.common.schemas.Project):
+    def create_project(
+        self, session: Session, project: mlrun.common.schemas.Project, with_summary=True
+    ):
```
quaark: Not sure it's worth adding this complication to the logic, to appease a unit test. We will only ever use this flag for the test but if it gets used by mistake it could cause issues.
Instead, maybe create the project record directly in the test without calling this method. Or separate the creation of the summary to a different method, and the UT can monkey patch it to do nothing.

quaark (changes requested):

server/api/db/sqldb/db.py
roei3000b: Good Idea, thanks!

server/api/initial_data.py
```diff
@@ -867,6 +871,29 @@ def _migrate_model_monitoring_jobs(db, db_session):
     )
 
 
+def _perform_version_7_data_migrations(
+    db: server.api.db.sqldb.db.SQLDB, db_session: sqlalchemy.orm.Session
+):
+    _migrate_project_summaries(db, db_session)
+
+
+def _migrate_project_summaries(db, db_session):
```
TomerShor: We're not actually migrating anything
```suggestion
def _create_project_summaries(db, db_session):
```

server/api/initial_data.py
```diff
@@ -867,6 +871,29 @@ def _migrate_model_monitoring_jobs(db, db_session):
...
+    # Here we want to iterate other all projects and create the related project summary.
+    # We do this because we create ProjectSummary record only when creating project.
+    # Because we upgrade from a version that doesn't have ProjectSummary, we need to create them manually now.
```
TomerShor: Changed the phrasing a bit
```suggestion
    # Create a project summary record for all projects.
    # We need to create them manually because a summary record is created only when a new
    # project is created, so project that existing prior to the upgrade don't have summaries.
```

TomerShor (changes requested):
Some minor changes but overall looks good!

server/api/db/sqldb/db.py
```diff
@@ -2514,6 +2514,16 @@ def create_project(self, session: Session, project: mlrun.common.schemas.Project
         labels = project.metadata.labels or {}
         update_labels(project_record, labels)
 
+        object_to_store = [project_record]
+
+        project_summary = self._generate_project_summary(project)
+        if project_summary:
+            object_to_store.append(project_summary)
```
quaark:
```suggestion
        self._generate_project_summary(project, object_to_store)
```
You can append in the function and then there's no need for the if

quaark (changes requested):

quaark (approved):
🚀

moranbental (approved):
Great!

TomerShor (approved):

PR #6131, opened 2024-08-07
[Model Monitoring] Fix delete pipeline model monitoring stream resources

Eyal-Danieli:
Fix an issue in which the "old" model monitoring stream resources that were stored under the `users` container were not deleted as expected due to invalid v3io access key. In case that the stream is stored under the `users` container, the API will use the `pipeline v3io access key` instead of the `model monitoring access key` which is related to the project resources and doesn't have permission to the `users` container.


davesh0812 (approved):
Nice!

PR #6133, opened 2024-08-07
[Alerts] Add abstract methods and fix imports and type hints

TomerShor:
Some improvements to the alert related db methods that align alerts with the rest of the resources (and improves QOL for developers).

### DIFF
--- a/server/api/db/base.py
+++ b/server/api/db/base.py
@@ -835,11 +835,56 @@ class DBInterface(ABC):
         pass
 
     @abstractmethod
-    def store_alert(self, session, alert: mlrun.common.schemas.AlertConfig):
+    def store_alert(
+        self, session, alert: mlrun.common.schemas.AlertConfig
+    ) -> mlrun.common.schemas.AlertConfig:
         pass
 
     @abstractmethod
     def get_all_alerts(self, session) -> list[mlrun.common.schemas.AlertConfig]:
+        pass
+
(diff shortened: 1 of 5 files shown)


roei3000b (commented):
Looks good, small question:
Why did you choose only create this function on the base class? what about `store_alert_template` for example?

server/api/db/base.py
```diff
@@ -835,13 +835,58 @@ def list_alert_templates(self, session) -> list[mlrun.common.schemas.AlertTempla
...
+        pass
+
+    @abstractmethod
+    def store_alert_state(
+        self,
+        session,
+        project: str,
+        name: str,
+        last_updated: datetime,
+        count: typing.Union[int, None] = None,
```
rokatyy: Why not `Optional[int] = None`?

TomerShor:
> Looks good, small question: Why did you choose only create this function on the base class? what about `store_alert_template` for example?

@roei3000b Some functions already exist in the base class (like `store_alert_template`), I only added the ones that were missing.

server/api/db/base.py
TomerShor: No reason, changed it.

rokatyy (approved):
Looks good!

roei3000b (approved):

Yacouby (approved):

liranbg (approved):

PR #6134, opened 2024-08-07
[Monitoring] Refine list runs filter

alonmr:
This shall help reduce the memory footprint of the runs monitoring job.
The changes are:
1. List runs filtered by kind label
2. List runs with non terminal states. It a run is in terminal state there is no need to update it. If the run does not exist and the pod is in terminal state, then the run was deleted and it will not be recreated (because the run is created when the job is launched).

https://iguazio.atlassian.net/browse/ML-7354

### DIFF
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -112,7 +112,7 @@ default_config = {
             "missing_runtime_resources_debouncing_interval": None,
             # max number of parallel abort run jobs in runs monitoring
             "concurrent_abort_stale_runs_workers": 10,
-            "list_runs_time_period_in_hours": 24 * 1,
+            "list_runs_time_period_in_hours": 24 * 3,  # 3 days
         },
         "projects": {
             "summaries": {
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -112,7 +112,7 @@ default_config = {
             "missing_runtime_resources_debouncing_interval": None,
             # max number of parallel abort run jobs in runs monitoring
             "concurrent_abort_stale_runs_workers": 10,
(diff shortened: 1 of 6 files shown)


server/api/main.py
```diff
@@ -901,7 +901,6 @@ def _generate_event_on_failed_runs(
         event_data = mlrun.common.schemas.Event(
             kind=alert_objects.EventKind.FAILED, entity=entity, value_dict=event_value
         )
-        mlrun.get_run_db().generate_event(alert_objects.EventKind.FAILED, event_data)
```
liranbg, line 904: Why deleting it?

mlrun/config.py
```diff
@@ -112,7 +112,7 @@
             "missing_runtime_resources_debouncing_interval": None,
             # max number of parallel abort run jobs in runs monitoring
             "concurrent_abort_stale_runs_workers": 10,
-            "list_runs_time_period_in_days": 7,  # days
+            "list_runs_time_period_in_hours": 24 * 3,  # 3 days
```
liranbg: Is that a 1.7 feature? If so ok to break if not need to slowly deprecate or so.

liranbg (changes requested):

server/api/main.py
alonmr: This is not doing anything. Verified with @Yacouby

mlrun/config.py
alonmr: I'm not sure if deprecation is really needed because it's only used in server side and is not part of the client spec.
But I think I will revert this change to days because the latest state change allows for more grace

liranbg (approved):

PR #6135, opened 2024-08-08
[KFP] Raise exception if trying to schedule function

roei3000b:
Kubeflow doesn't support submitting  schedule job, so if the user use `kfp` engine and also pass `schedule` arg we want to raise exception about it.
(There was also option to only warn and just not run the function, but we use the approach  of raising exception because there shouldn't be much cases where users used `schedule` on `kfp` engine).

https://iguazio.atlassian.net/browse/ML-7226

### DIFF
--- a/mlrun/projects/operations.py
+++ b/mlrun/projects/operations.py
@@ -190,7 +190,6 @@ def run_function(
         if schedule:
             raise mlrun.errors.MLRunInvalidArgumentError(
                 "Scheduling is not supported in KFP run function."
-                "Use either project.set_workflow() or project.run() to schedule runs."
             )
         return function.as_step(
             name=name, runspec=task, workdir=workdir, outputs=outputs, labels=labels
--- a/mlrun/projects/operations.py
+++ b/mlrun/projects/operations.py
@@ -189,7 +189,7 @@ def run_function(
     if engine == "kfp":
         if schedule:
             raise mlrun.errors.MLRunInvalidArgumentError(
-                "Scheduling is not supported in KFP run function."
+                "Scheduling job is not supported when running a workflow with kfp engine."
             )
         return function.as_step(
             name=name, runspec=task, workdir=workdir, outputs=outputs, labels=labels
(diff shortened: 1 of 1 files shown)


mlrun/projects/operations.py
```diff
@@ -187,6 +187,10 @@ def my_pipe(url=""):
     task.spec.verbose = task.spec.verbose or verbose
 
     if engine == "kfp":
+        if schedule:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "Schedule is not supported in KFP runs"
```
liranbg: maybe suggest what does support (like remote)

mlrun/projects/operations.py
```diff
@@ -187,6 +187,11 @@ def my_pipe(url=""):
     task.spec.verbose = task.spec.verbose or verbose
 
     if engine == "kfp":
+        if schedule:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "Schedule is not supported in KFP run function."
```
jillnogold: "Scheduling is not supported in KFP run function."
                "Use either project.set_schedule() or project.run() to schedule runs."

mlrun/projects/operations.py
```diff
@@ -187,6 +187,11 @@ def my_pipe(url=""):
     task.spec.verbose = task.spec.verbose or verbose
 
     if engine == "kfp":
+        if schedule:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "Scheduling is not supported in KFP run function."
+                "Use either project.set_schedule() or project.run() to schedule runs."
+            )
```
GiladShapira94, line 193: "Scheduling job is not supported when running a workflow with kfp engine."
"Use project.run_function()."

GiladShapira94: on second thought this is better - 
"Scheduling job is not supported when running a workflow with kfp engine."

liranbg (approved):

mlrun/projects/operations.py
```diff
@@ -187,6 +187,10 @@ def my_pipe(url=""):
     task.spec.verbose = task.spec.verbose or verbose
 
     if engine == "kfp":
+        if schedule:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "Scheduling job is not supported when running a workflow with kfp engine."
```
jillnogold, line 192: "Scheduling jobs is not supported when running a workflow with the kfp engine."

PR #6139, opened 2024-08-08
[Application] Force pull source at build time

alonmr:
Pulling source at run time is not supported so there is no reason to allow configuring it.
This PR forces pull at build time and logs appropriate warnings.

https://iguazio.atlassian.net/browse/ML-7429

### DIFF
--- a/mlrun/runtimes/nuclio/application/application.py
+++ b/mlrun/runtimes/nuclio/application/application.py
@@ -375,6 +375,13 @@ class ApplicationRuntime(RemoteRuntime):
         :param pull_at_runtime: currently not supported, source must be loaded into the image during the build process
         :param target_dir:      target dir on runtime pod or repo clone / archive extraction
         """
+        if pull_at_runtime:
+            logger.warning(
+                f"{pull_at_runtime=} is currently not supported for application runtime "
+                "and will be overridden to False",
+                pull_at_runtime=pull_at_runtime,
+            )
+
         self._configure_mlrun_build_with_source(
             source=source,
             workdir=workdir,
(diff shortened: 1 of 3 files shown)


mlrun/runtimes/nuclio/application/application.py
```diff
@@ -349,23 +358,25 @@ def deploy(
         )
 
     def with_source_archive(
-        self, source, workdir=None, pull_at_runtime=True, target_dir=None
+        self,
+        source: str,
+        workdir: str = None,
+        target_dir: str = None,
```
TomerShor: These args are not aligned between the same method in different classes.
Doesn't it cause issues?

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -349,23 +358,25 @@ def deploy(
...
-        """load the code from git/tar/zip archive at runtime or build
+        """load the code from git/tar/zip archive at build
 
         :param source:          valid absolute path or URL to git, zip, or tar file, e.g.
                                 git://github.com/mlrun/something.git
                                 http://some/url/file.zip
                                 note path source must exist on the image or exist locally when run is local
                                 (it is recommended to use 'workdir' when source is a filepath instead)
         :param workdir:         working dir relative to the archive root (e.g. './subdir') or absolute to the image root
-        :param pull_at_runtime: load the archive into the container at job runtime vs on build/deploy
```
liranbg, line 362: you can leave intact and ignore the flag

jillnogold (approved):

mlrun/runtimes/nuclio/application/application.py
alonmr: note this is aligned with KubejobRuntime and not with RemoteRuntime (which is not a KubejobRuntime) but it seems to work

TomerShor (approved):

liranbg (approved):

PR #6143, opened 2024-08-10
[CI] Automate pipelines adapters bump with dependabot

alonmr:
Instead of us forgetting to update every time we release

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -42,3 +42,14 @@ updates:
     commit-message:
       prefix: "[Dependabot-automated] "
     open-pull-requests-limit: 6
+
+  # mlrun pipelines adapters
+  - package-ecosystem: pip
+    directory: /
+    schedule:
+      interval: daily
+    allow:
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -42,3 +42,14 @@ updates:
...
       prefix: "[Dependabot-automated] "
     open-pull-requests-limit: 6
+
+  # mlrun pipelines adapters
+  - package-ecosystem: pip
+    directory: /
+    schedule:
+      interval: daily
+    allow:
+      - dependency-name: "mlrun-pipelines-kfp*"
```
liranbg, line 52: Regex is allowed? Also nicely done!

liranbg (approved):

.github/dependabot.yml
alonmr: I believe only `*` is supported, followed the example here https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow

PR #6149, opened 2024-08-11
[Model Monitoring] Select only the relevant columns in V3IO TSDB query

jond01:
Fixes [ML-7318](https://iguazio.atlassian.net/browse/ML-7318).

Select only the columns required for the data to be returned in the `/metrics-values` API endpoint. This does not include `current_stats`, resulting in a significantly reduced response size from V3IO web-API.
I tested it and it works.

[ML-7318]: https://iguazio.atlassian.net/browse/ML-7318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
+++ b/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
@@ -504,19 +504,29 @@ class V3IOTSDBConnector(TSDBConnector):
         if type == "metrics":
             table_path = self.tables[mm_schemas.V3IOTSDBTables.METRICS]
             name = mm_schemas.MetricData.METRIC_NAME
+            columns = [mm_schemas.MetricData.METRIC_VALUE]
             df_handler = self.df_to_metrics_values
         elif type == "results":
             table_path = self.tables[mm_schemas.V3IOTSDBTables.APP_RESULTS]
             name = mm_schemas.ResultData.RESULT_NAME
+            columns = [
+                mm_schemas.ResultData.RESULT_VALUE,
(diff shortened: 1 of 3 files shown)


mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -504,19 +504,29 @@ def read_metrics_data(
...
+                mm_schemas.ResultData.RESULT_VALUE,
+                mm_schemas.ResultData.RESULT_STATUS,
+                mm_schemas.ResultData.RESULT_KIND,
+            ]
             df_handler = self.df_to_results_values
         else:
             raise ValueError(f"Invalid {type = }")
 
+        groupby = [name, mm_schemas.WriterEvent.APPLICATION_NAME]
+
```
davesh0812: Why you need to group by in this case? You need to return all the data over the required time

jond01: You are right. It's not needed, I removed it.

davesh0812 (dismissed):

davesh0812 (commented):
Lets also do the same in tdengine, to get the same functionality.
In the near future we will add a flag to get the extra_data if the user want to .

jond01:
> Lets also do the same in tdengine, to get the same functionality.

Done

davesh0812 (approved):

PR #6152, opened 2024-08-11
[Datastore] fix gcs large upload

tomerm-iguazio:
Fix gcs large upload, while considering gcs multipart upload limitations as described here:
https://cloud.google.com/storage/docs/multipart-uploads#storage-upload-object-chunks-python

Add tests for large file uploads with and without a datastore profile. Include scenarios for both options: using a credential file and using credentials information directly.

[ML-7410](https://iguazio.atlassian.net/browse/ML-7410)

[ML-7410]: https://iguazio.atlassian.net/browse/ML-7410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

depend on:
https://github.com/mlrun/mlrun/pull/6105


mlrun/datastore/google_cloud_storage.py
```diff
@@ -23,30 +24,54 @@
...
+
 
 class GoogleCloudStorageStore(DataStore):
     using_bucket = True
 
     def __init__(self, parent, schema, name, endpoint="", secrets: dict = None):
+        self._storage_client = None
+        self._transfer_manager = None
+        self.workers = WORKERS
+        self.chunk_size = CHUNK_SIZE
```
gtopper: Same thing we had in Azure PR... Please move these to the class level and remove the module level constants.

mlrun/datastore/google_cloud_storage.py
```diff
@@ -23,30 +24,54 @@
...
+    def create_filesystem_and_client(self):
         try:
             import gcsfs  # noqa
         except ImportError as exc:
             raise ImportError(
                 "Google gcsfs not installed, run pip install gcsfs"
             ) from exc
+        from google.cloud.storage import Client, transfer_manager
+
+        # Creates with the same storage options to avoid credential differences.
```
gtopper:
```suggestion
        # use the same storage options to avoid credential differences
```

mlrun/datastore/google_cloud_storage.py
```diff
@@ -23,30 +24,54 @@
...
+        storage_options = self.get_storage_options()
+        # in order to support az and wasbs kinds.
```
gtopper: What do az and wasbs have to do with gcs?
```suggestion
        # in order to support az and wasbs kinds
```

mlrun/datastore/google_cloud_storage.py
```diff
@@ -103,8 +128,41 @@ def put(self, key, data, append=False):
...
+            self.filesystem.put_file(src_path, united_path, overwrite=True)
+            return
+
+        bucket = self.storage_client.bucket(self.endpoint)
+        bucket_retention_policy = bucket._properties.get("retentionPolicy")
+        if bucket_retention_policy or bucket.object_retention_mode:
+            self.filesystem.put_file(src_path, united_path, overwrite=True)
+            return
+        blob = bucket.blob(key.strip("/"))
+
```
gtopper, line 159:
```suggestion

        blob = bucket.blob(key.strip("/"))
```

mlrun/datastore/google_cloud_storage.py
```diff
@@ -23,30 +24,54 @@
...
         super().__init__(parent, name, schema, endpoint, secrets=secrets)
 
     @property
-    def filesystem(self):
-        """return fsspec file system object, if supported"""
-        if self._filesystem:
-            return self._filesystem
+    def storage_client(self):
+        if not self._filesystem or self._storage_client:
+            self.create_filesystem_and_client()
```
gtopper: Same as in azure, these are better decoupled. We can use google cloud library without relying on gcsfs.

mlrun/datastore/google_cloud_storage.py
```diff
@@ -103,8 +128,41 @@ def put(self, key, data, append=False):
...
     def upload(self, key, src_path):
-        path = self._make_path(key)
-        self.filesystem.put_file(src_path, path, overwrite=True)
+        file_size = os.path.getsize(src_path)
+        united_path = self._make_path(key)
+
+        # Multiple upload limitation recommendations as described in
+        # https://cloud.google.com/storage/docs/multipart-uploads#storage-upload-object-chunks-python
+
+        if file_size < self.chunk_size:
```
gtopper:
```suggestion
        if file_size <= self.chunk_size:
```

mlrun/datastore/google_cloud_storage.py
```diff
@@ -23,30 +24,54 @@
...
             raise ImportError(
                 "Google gcsfs not installed, run pip install gcsfs"
             ) from exc
+        from google.cloud.storage import Client, transfer_manager
+
+        # Creates with the same storage options to avoid credential differences.
+        storage_options = self.get_storage_options()
+        # in order to support az and wasbs kinds.
         filesystem_class = get_filesystem_class(protocol=self.kind)
         self._filesystem = makeDatastoreSchemaSanitizer(
```
gtopper: Pre-existing, but would be nice to rename it to python convention (snakecase).

mlrun/datastore/google_cloud_storage.py
```diff
@@ -23,30 +24,54 @@
...
             filesystem_class,
             using_bucket=self.using_bucket,
-            **self.get_storage_options(),
+            **storage_options,
         )
+        credentials = self._filesystem.credentials.credentials
+        self._storage_client = Client(credentials=credentials)
+        # to avoid importing transfer_manager later:
+        self._transfer_manager = transfer_manager
```
gtopper: Can import at the top.

mlrun/datastore/google_cloud_storage.py
```diff
@@ -23,30 +24,54 @@
...
+        return self._storage_client
+
+    def create_filesystem_and_client(self):
         try:
             import gcsfs  # noqa
         except ImportError as exc:
             raise ImportError(
                 "Google gcsfs not installed, run pip install gcsfs"
             ) from exc
+        from google.cloud.storage import Client, transfer_manager
```
gtopper: Can import at the top.
```suggestion
```

mlrun/datastore/google_cloud_storage.py
```diff
@@ -103,8 +128,41 @@ def put(self, key, data, append=False):
...
+        bucket_retention_policy = bucket._properties.get("retentionPolicy")
+        if bucket_retention_policy or bucket.object_retention_mode:
+            self.filesystem.put_file(src_path, united_path, overwrite=True)
+            return
+        blob = bucket.blob(key.strip("/"))
+
+        if (
+            blob.retention_expiration_time
+            or blob.temporary_hold
+            or blob.event_based_hold
```
gtopper: From the client library code, it doesn't seem like these can be defined immediately after creating the blob, but maybe I'm missing something.

mlrun/datastore/google_cloud_storage.py
tomerm-iguazio: fixed tnx

mlrun/datastore/google_cloud_storage.py
```diff
@@ -28,28 +32,62 @@ class GoogleCloudStorageStore(DataStore):
     using_bucket = True
 
     def __init__(self, parent, schema, name, endpoint="", secrets: dict = None):
+        self._storage_client = None
+        self._storage_options = None
+        self.workers = 8
+        self.chunk_size = 32 * 1024 * 1024
```
gtopper: If you're not allowing for overriding it with a parameter, please move to class level.

mlrun/datastore/google_cloud_storage.py
```diff
@@ -28,28 +32,62 @@ class GoogleCloudStorageStore(DataStore):
...
+                    "gcsfs authentication file not found!"
+                )
+        elif isinstance(token, dict):
+            credentials = service_account.Credentials.from_service_account_info(
+                token, scopes=[access]
+            )
+        elif isinstance(token, Credentials):
+            credentials = token
+        else:
+            raise ValueError("Token format not understood")
```
gtopper:
```suggestion
            raise ValueError(f"Unsupported token type: {type(token)}")
```

mlrun/datastore/google_cloud_storage.py
```diff
@@ -12,44 +12,82 @@
...
 
 class GoogleCloudStorageStore(DataStore):
     using_bucket = True
+    workers = 8
+    chunk_size = 32 * 1024 * 1024
 
     def __init__(self, parent, schema, name, endpoint="", secrets: dict = None):
+        self._storage_client = None
+        self._storage_options = None
         super().__init__(parent, name, schema, endpoint, secrets=secrets)
```
gtopper:
```suggestion
        super().__init__(parent, name, schema, endpoint, secrets=secrets)
        self._storage_client = None
        self._storage_options = None
```
convention is to call `super().__init__` first, unless there's a reason to do otherwise.

gtopper (approved):
LGTM otherwise

assaf758 (approved):

PR #6153, opened 2024-08-11
[Secretes] If secrets is None delete all secrets data

roei3000b:
[Here](https://github.com/mlrun/mlrun/blob/30aadb6f3d83b27db15c5704d6c0be0d7503032d/server/api/crud/projects.py#L477) we expect that when passing None as the `secrets` that we want to delete, we should delete **all** secrets.
Fixing it by checking if `secrets` is None and in that case set the `secret_data` to empty dict.

https://iguazio.atlassian.net/browse/ML-7440

### DIFF
--- a/server/api/utils/singletons/k8s.py
+++ b/server/api/utils/singletons/k8s.py
@@ -615,6 +615,9 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
                 for key, value in k8s_secret.data.items()
                 if key not in secrets
             }
+        elif secrets is None:
+            # Delete all secrets
+            secret_data = {}
         else:
             secret_data = k8s_secret.data.copy()
 
--- a/server/api/utils/singletons/k8s.py
+++ b/server/api/utils/singletons/k8s.py
@@ -577,7 +577,7 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
         """
         Delete secrets from a kubernetes secret object
         :param secret_name: the project secret name
-        :param secrets:     the secrets to delete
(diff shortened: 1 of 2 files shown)


alonmr (changes requested):
Nice! One small change is needed for me to sleep good at night. I know this should never happen but if `k8s_secret.data` is empty than the secret will not be deleted due to if in line 625. If this ever happens we should still delete the secret, otherwise you have to add a secret to be able to completely delete it.

moranbental:
> Nice! One small change is needed for me to sleep good at night. I know this should never happen but if `k8s_secret.data` is empty than the secret will not be deleted due to if in line 625. If this ever happens we should still delete the secret, otherwise you have to add a secret to be able to completely delete it.

@alonmr Why do we need to check that? We’re already checking it on line 603 and deleting the secret in that case. Additionally, we have test coverage for this on line 134

server/api/utils/singletons/k8s.py
```diff
@@ -615,11 +615,14 @@ def delete_secrets(
                 for key, value in k8s_secret.data.items()
                 if key not in secrets
             }
+        elif secrets is None:
+            # Delete all secrets
```
moranbental, line 619: Please also add to the docstring that when `secret=None`, all secrets are deleted

moranbental (changes requested):

liranbg (approved):

moranbental (approved):
Great!

alonmr (approved):

PR #6156, opened 2024-08-12
[Serving] Limit the serving spec length when using ConfigMaps

moranbental:
When trying to deploy a serving function with more than 4500 models, the deployment fails with the error: `Too long: must have at most 1048576 bytes`. This error occurs because large serving specs are stored in config maps mounted to the pod, and config map objects in Kubernetes have a 1 MB size limit.

To fix this issue, we limit the number of models to 4500 and also check the size of the serving spec.

https://iguazio.atlassian.net/browse/ML-7335


server/api/crud/runtimes/nuclio/function.py
```diff
@@ -293,10 +296,16 @@ def _compile_function_config(
         )
         # since environment variables have a limited size,
         # large serving specs are stored in config maps that are mounted to the pod
+        serving_spec_len = len(serving_spec)
         if (
             can_pass_via_cm
-            and len(serving_spec) >= mlrun.mlconf.httpdb.nuclio.serving_spec_env_cutoff
+            and serving_spec_len >= mlrun.mlconf.httpdb.nuclio.serving_spec_env_cutoff
         ):
```
rokatyy: but are we sure that this gives us the number of bytes directly? It is just the len of the object, not the size. It would be more accurate to do `serving_spec.encode('utf-8')`. Or maybe we encode it somewhere in the flow before?

alonmr (commented):
Why not make it configurable? Also let's add a unit test. I believe there are unit tests for serving_spec_env_cutoff you can take as reference

server/api/crud/runtimes/nuclio/function.py
```diff
@@ -293,10 +296,16 @@ def _compile_function_config(
...
+            if serving_spec_len >= SERVING_SPEC_MAX_LENGTH:
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    f"Serving spec length {serving_spec_len} is too large. "
+                    f"The maximum allowed length for the serving spec is {SERVING_SPEC_MAX_LENGTH}."
```
TomerShor: I wonder if that exception is useful from a user perspective.
The serving spec is something that is filled behind the scenes when a user calls `serving_function.add_model()`, so will they know that this is a possible cause of the error?

TomerShor:
> Why not make it configurable? Also let's add a unit test. I believe there are unit tests for serving_spec_env_cutoff you can take as reference

@alonmr this is a k8s limit, you can't make it configurable as larger than the max size anyways

server/api/crud/runtimes/nuclio/function.py
moranbental: we get `serving_spec` from https://github.com/mlrun/mlrun/blob/e0ca3e679d496ffb4cecf511d0b7a24b9927fe53/mlrun/runtimes/nuclio/serving.py#L666
so I think using len() is sufficient

rokatyy: @moranbental i am not sure, could you please share why do you assume that (maybe docs link)?

alonmr:
> > Why not make it configurable? Also let's add a unit test. I believe there are unit tests for serving_spec_env_cutoff you can take as reference
> 
> @alonmr this is a k8s limit, you can't make it configurable as larger than the max size anyways

k8s limit doesn't mean it will never change 🤔

assaf758:
Currently, the fix is focused on identifying that we have reached 1MB and issuing a proper message to the user.
As I see it, this does not help the user too much, since obviously he did want to have serving graph(s) that consume > 1 MB.
I would like to suggest an alternative approach - where instead of config-map we will use the object-store (which does not have 1MB limit). Of course, we would need to save the object in pipeline, so it will not be accessible to normal users.

mlrun/serving/states.py
```diff
@@ -84,6 +84,9 @@ class StepKinds:
 ]
 
 
+MAX_ALLOWED_NUM_OF_STEPS = 4500
```
TomerShor:
```suggestion
MAX_ALLOWED_STEPS = 4500
```

mlrun/serving/states.py
```diff
@@ -733,6 +736,11 @@ def add_route(
         if not route:
             route = TaskStep(class_name, class_args, handler=handler)
         route.function = function or route.function
+
+        if len(self._routes) >= MAX_ALLOWED_NUM_OF_STEPS:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"The maximum allowed number of steps is {MAX_ALLOWED_NUM_OF_STEPS}"
```
TomerShor: A bit better sounding:
```suggestion
                f"The maximum number of steps allowed in the serving graph is {MAX_ALLOWED_NUM_OF_STEPS}"
```

TomerShor (changes requested):
Minor suggestion, otherwise looking good

TomerShor (approved):

server/api/crud/runtimes/nuclio/function.py
alonmr: I agree with @TomerShor that we should change this error message to something the user can work with. Possibly point to `mlrun.runtimes.nuclio.serving.ServingRuntime._get_serving_spec` and tell the user to check it to see what takes so much space. Other than that lgtm

moranbental: Error messages will be defined by @jillnogold in this sub-document https://iguazio.atlassian.net/browse/ML-7694

server/api/crud/runtimes/nuclio/function.py
liranbg: well, since it is a json we format and we dont use symbols or non-utf8 which could increase the underlying size, the usage is most likely be ok. that being said, the correct way is indeed encode and take its length as @rokatyy suggested. `len(serving_spec.encode('utf-8'))` rather than just `len(serving_spec)`

liranbg (approved):

PR #6157, opened 2024-08-13
[Datastore] Block snowflake target in to_dataframe function

tomerm-iguazio:
`to_dataframe` in `job.py` supports targets that `support_pandas=True`.

So we block any target in `get_offline_features` that does not support pandas.
Add a check for this issue while calling `get_offline_features ` function + test with snowflake.

[ML-7434](https://iguazio.atlassian.net/browse/ML-7434)

[ML-7434]: https://iguazio.atlassian.net/browse/ML-7434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/feature_store/retrieval/job.py
```diff
@@ -179,8 +180,15 @@ def to_dataframe(self, columns=None, df_module=None, **kwargs):
...
         file_format = kwargs.get("format")
+        target_kind = self.run.status.results["target"]["kind"]
         if not file_format:
-            file_format = self.run.status.results["target"]["kind"]
+            file_format = target_kind
+
+        target_class = kind_to_driver.get(target_kind)
+        if target_class and not target_class.support_pandas:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"to_dataframe does not support targets that do not support pandas engine. Target kind: {target_kind}"
```
gtopper: Not a fan of the phrasing, but I see a bigger problem. At this point, the job has already completed and written the data. The user has already waited (potentially many) minutes for the result. The place to raise this error is before we start the remote job.

mlrun/datastore/targets.py
```diff
@@ -390,6 +390,7 @@ class BaseStoreTarget(DataTargetBase):
     is_offline = False
     support_spark = False
     support_storey = False
+    support_pandas = False
```
gtopper, line 393: Looks like some targets that support pandas are missing `support_pandas = True`. E.g. `TSDBTarget`.

tests/system/feature_store/test_spark_snowflake.py
```diff
@@ -278,3 +278,63 @@ def test_parquet_source_ingest(self):
         pd.testing.assert_frame_equal(
             sort_df(df, "id"), sort_df(result_df, "id"), check_dtype=False
         )
+
+    def test_snowflake_target_to_dataframe(self):
+        if self.run_local:
+            pytest.skip("test_snowflake_target_to_dataframe run in remote run only")
```
gtopper:
```suggestion
            pytest.skip("local run is not supported")
```

tests/system/feature_store/test_spark_snowflake.py
```diff
@@ -278,3 +278,63 @@ def test_parquet_source_ingest(self):
...
+            spark_context=self.spark_service,
+            run_config=fstore.RunConfig(
+                local=self.run_local,
+            ),
+        )
+        vector = fstore.FeatureVector(
+            "feature_vector_snowflake", ["snowflake_feature_set.*"]
+        )
+        run_config = fstore.RunConfig(
+            local=self.run_local, kind=None if self.run_local else "remote-spark"
```
gtopper: Local run is not supported anyway, so you can remove this boilerplate (including in a few other places).

gtopper (changes requested):

mlrun/datastore/targets.py
tomerm-iguazio: I am confused, so why we got `NotImplementedError` on `as_df` function in `TSDBTarget`?

tests/system/feature_store/test_spark_snowflake.py
tomerm-iguazio: tnx!

mlrun/feature_store/retrieval/job.py
tomerm-iguazio: But we want to let the user write to these targets, we are unable only to do `to_dataframe`.
We can discuss over this topic

gtopper: The whole point of this flow is for the user to get a dataframe in the end.

tests/system/feature_store/test_spark_snowflake.py
gtopper: Again, please change this in all places in the added code.

mlrun/datastore/targets.py
gtopper: Actually you're right. Please add `support_pandas = ...` to all targets, to be consistent. Or only where it needs to be overridden if you prefer, but not a mix of the two.

mlrun/datastore/targets.py
```diff
@@ -1834,6 +1788,7 @@ class CustomTarget(BaseStoreTarget):
     is_online = False
     support_spark = False
     support_storey = True
+    support_pandas = True
```
gtopper, line 1790: This doesn't look right.
```suggestion
    support_pandas = False
```

tests/system/feature_store/test_spark_snowflake.py
tomerm-iguazio: default - local=false if there is not remote spark

tests/system/feature_store/test_spark_snowflake.py
```diff
@@ -278,3 +278,63 @@ def test_parquet_source_ingest(self):
...
+        )
+        self.generate_snowflake_source_table()
+        self.tables_to_drop.append(result_table)
+        feature_set.ingest(
+            source,
+            targets=[target],
+            spark_context=self.spark_service,
+            run_config=fstore.RunConfig(
+                local=self.run_local,
+            ),
```
gtopper:
```suggestion
```

mlrun/datastore/targets.py
tomerm-iguazio: removed from StreamTarget:)

mlrun/datastore/targets.py
tomerm-iguazio: was True before, remove please:)

tests/system/feature_store/test_spark_snowflake.py
tomerm-iguazio: have to mention remote run

mlrun/feature_store/api.py
```diff
@@ -230,6 +230,11 @@ def _get_offline_features(
             "entity_timestamp_column param "
             "can not be specified without entity_rows param"
         )
+    if isinstance(target, BaseStoreTarget) and not target.support_pandas:
```
gtopper, line 233: `isinstance(target, BaseStoreTarget)` as opposed to?

gtopper (approved):
LGTM otherwise (see one comment below). The whole `supports_X` stuff could probably benefit from some refactoring in the future though.

mlrun/feature_store/api.py
tomerm-iguazio: the hint in the target parameter is DataTargetBase.

DataTargetBase does not have support_pandas:)

PR #6160, opened 2024-08-13
[Projects] Enforce project_setup.py returns a project object

liranbg:
Otherwise, the flow fails unexpectedly as the returned project is being used later on for the creation / load

https://iguazio.atlassian.net/browse/ML-7531

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -18,6 +18,7 @@ import glob
 import http
 import importlib.util as imputil
 import json
+import os.path
 import pathlib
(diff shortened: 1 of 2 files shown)


mlrun/projects/project.py
```diff
@@ -18,6 +18,7 @@
 import http
 import importlib.util as imputil
 import json
+import os.path
```
alonmr: redundant can us `path` import below

alonmr (approved):

PR #6161, opened 2024-08-13
[Runs] Abort run from run object

liranbg:
just a syntactic sugar to make run abort-able from the sdk itself

### DIFF
--- a/mlrun/model.py
+++ b/mlrun/model.py
@@ -1789,6 +1789,11 @@ class RunObject(RunTemplate):
 
         return state
 
+    def abort(self):
+        """abort the run"""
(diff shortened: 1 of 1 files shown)


mlrun/model.py
```diff
@@ -1789,6 +1789,11 @@ def wait_for_completion(
 
         return state
 
+    def abort(self):
+        """abort the run"""
+        db = mlrun.get_run_db()
+        db.abort_run(self.metadata.uid, self.metadata.project)
```
alonmr, line 1795: add iteration

alonmr (approved):

PR #6169, opened 2024-08-15
[Model Monitoring] Avoid suppressing errors in `KVStoreBase.get_last_analyzed`

gtopper:
Noticed while investigating an otherwise unrelated issue – https://iguazio.atlassian.net/browse/ML-7160?focusedCommentId=154985


jond01 (approved):
Nice, it's going to be removed soon anyway - see [ML-7457](https://iguazio.atlassian.net/browse/ML-7457), but ok :)

PR #6170, opened 2024-08-15
[Docs] Create a basic example of auto-logging with packager

moranbental:
Update the documentation for auto-logging dataset artifacts with package


docs/cheat-sheet.md
```diff
@@ -590,6 +590,23 @@ log_with_returns_run = my_func.run(
 ```
 
 ### Automatic logging
+# Auto logging dataset basic example
+```python
+import pandas
+
+
+def func(col: list):
```
guy1992l: and just for the good complete type hinted developer experience
```suggestion
def func(col: list) -> pd.DataFrame:
```

docs/cheat-sheet.md
```diff
@@ -590,6 +590,23 @@ log_with_returns_run = my_func.run(
 ```
 
 ### Automatic logging
+# Auto logging dataset basic example
+```python
+import pandas
```
guy1992l: even padnas don't import themselves as `padnas`
```suggestion
import pandas as pd
```

docs/cheat-sheet.md
```diff
@@ -590,6 +590,23 @@ log_with_returns_run = my_func.run(
...
+
+
+def func(col: list):
+    df = pandas.DataFrame({"col": col})
+    return df
+
+
+function.run(
```
guy1992l: function is a builtin in Python
```suggestion
my_func.run(
```

guy1992l (changes requested):
I would add a second function that will receive the now logged DF via `inputs` and return a value to have a full cycle of auto-logging and parsing to show how to move objects between functions.

```python
func_2(df: pd.DataFrame) -> int:
    return int(df.sum()[0])
```

Then:
```python
run_1 = my_func.run(
    handler="func",
    params={"col": [1, 2, 3, 4, 5, 6, 7]},
    returns=["df:dataset"],
    local=True,
)
my_func.run(
    handler="func_2",
    inputs={"df": run_1["df"]},
    returns=["sum"],
    local=True,
)
```

guy1992l (approved):
Amazing :)

docs/cheat-sheet.md
```diff
@@ -590,6 +590,34 @@ log_with_returns_run = my_func.run(
 ```
 
 ### Automatic logging
+# Auto logging dataset basic example
```
jillnogold:
```python
# Auto-logging a dataset: basic example

guy1992l: shouldn't it be a sub-sub-sub-sub-header? `####` instead of `#` if the line above has `###`?

guy1992l (approved):

jillnogold (approved):

alonmr (approved):

PR #6172, opened 2024-08-15
[Alerts] Handle the duplicated attributes between alert and notification

Yacouby:
This PR is intended to handle 3 duplicated attributes between the alert and the notification objects.
the duplicated attributes are :
alert.severity & notification.severity
alert.trigger & notification.when
alert.summary & notification.message

in the current flow the user needed to supply all of the above fields in order to create the alert config, even though we don't use some the attributes that were given in the notification..
so now the user can set these attributes only in the alert config, and will not have to supply them in the notification.
this shortens the process of creating the alert.

resolves:
https://iguazio.atlassian.net/browse/ML-7248

### DIFF
--- a/server/api/utils/notification_pusher.py
+++ b/server/api/utils/notification_pusher.py
@@ -157,7 +157,9 @@ class AlertNotificationPusher(_NotificationPusherBase):
         notification_object: mlrun.common.schemas.Notification,
     ):
         message = (
-            f": {notification_object.message}" if notification_object.message else ""
+            f": {notification_object.message}"
+            if notification_object.message
+            else alert.summary
         )
 
         severity = alert.severity
--- a/mlrun/common/schemas/notification.py
+++ b/mlrun/common/schemas/notification.py
@@ -71,9 +71,9 @@ class Notification(pydantic.BaseModel):
 
     kind: NotificationKind
     name: str
-    message: str
-    severity: NotificationSeverity
-    when: list[str]
+    message: typing.Optional[str] = None
+    severity: typing.Optional[NotificationSeverity] = None
+    when: typing.Optional[list[str]] = None
     condition: typing.Optional[str] = None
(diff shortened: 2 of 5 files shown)


server/api/utils/notification_pusher.py
```diff
@@ -157,7 +157,9 @@ def _prepare_notification_args(
         notification_object: mlrun.common.schemas.Notification,
     ):
         message = (
-            f": {notification_object.message}" if notification_object.message else ""
+            f": {notification_object.message}"
+            if notification_object.message
+            else alert.summary
```
TomerShor, line 162: Is there any chance the alert summary is also None?

TomerShor (commented):
We need to make sure this doesn't cause regression for creating notifications not in the scope.lf alerts (run notifications)

server/api/utils/notification_pusher.py
Yacouby: no, creating an alert without a summary would fail

quaark (approved):

server/api/utils/notification_pusher.py
quaark: @Yacouby what if I created a notification not via alert? Since we are not validating the message anymore

quaark (changes requested):

server/api/utils/notification_pusher.py
Yacouby: this flow is called only in the alerts flow.. there is another method with the same name as this one that is called in the NotificationPusher class, and there it is handled if the message is not supplied

Yacouby: @quaark this flow is called in the alerts flow.. there is another method with the same name as this one, that is called in the NotificationPusher class, and there it is handled if the message is not supplied

quaark (approved):

PR #6174, opened 2024-08-15
[Application] Prevent code to function flow

alonmr:
This flow is not supported yet for application runtime.
Also moved image validation to "new_function" flow because in "code_to_function" we can use the default image.

https://iguazio.atlassian.net/browse/ML-7546


mlrun/projects/project.py
```diff
@@ -4375,18 +4379,17 @@ def _init_function_from_dict(
...
+            # when load_source_on_run is used we allow not providing image as code will be loaded pre-run. ML-4994
+            if (
+                not image
+                and not project.default_image
+                and kind != "local"
+                and not project.spec.load_source_on_run
+            ):
+                raise ValueError(
+                    "image must be provided with py code files which do not run on 'local' engine kind"
+                )
```
theSaarco, line 4403: Why did this check move inside the `if` above? Is this related to some other issue?

mlrun/run.py
```diff
@@ -744,11 +744,10 @@ def update_common(fn, spec):
...
 
     if kind == RuntimeKinds.application:
-        if handler:
-            raise MLRunInvalidArgumentError(
-                "Handler is not supported for application runtime"
-            )
-        filename, handler = ApplicationRuntime.get_filename_and_handler()
+        raise MLRunInvalidArgumentError(
+            "Code to function is not supported for application runtime. "
+            "Code can be specified via project/function source."
```
theSaarco:
```suggestion
            "Embedding a code file is not supported for application runtime. "
            "Code files should be specified via project/function source."
```

I'm not sure this message is really explaining the situation very well (especially now that users are not expected to call `code_to_function` anymore so this sounds strange). TBH my suggestion doesn't sound very good to me either, but it's a bit more explicit.

theSaarco (changes requested):
Overall looks good. I have a question inside, and I'm also not very happy with the error message given (not 100% happy with my suggestion as well).

mlrun/projects/project.py
alonmr: because this should not be validated for the else clause. In the else clause we use code to function which means the code is embedded in the function spec so there is no reason to require an image

tests/runtimes/test_application.py
```diff
@@ -255,6 +258,17 @@ def test_deploy_reverse_proxy_image(rundb_mock, igz_version_mock):
...
 
+def test_application_from_local_file_validation():
+    project = mlrun.get_or_create_project("test-application")
+    func_path = assets_path / "sample_function.py"
+    with pytest.raises(mlrun.errors.MLRunInvalidArgumentError) as exc:
+        project.set_function(func=str(func_path), name="my-app", kind="application")
+    assert str(exc.value) == (
+        "Embedding a code file is not supported for application runtime. "
+        "Code files should be specified via project/function source."
+    )
```
theSaarco:
```suggestion
    with pytest.raises(mlrun.errors.MLRunInvalidArgumentError, match="Embedding a code file ..."):
        project.set_function(func=str(func_path), name="my-app", kind="application")
```

`pytest.raises` already supports matching exception text representation. Not critical, FYI.

theSaarco (approved):
Looks good. Minor suggestion, for your consideration.

theSaarco (approved):
Yep.

PR #6177, opened 2024-08-18
[Application] Default to single replica

alonmr:
https://iguazio.atlassian.net/browse/ML-7555

### DIFF
--- a/mlrun/runtimes/nuclio/application/application.py
+++ b/mlrun/runtimes/nuclio/application/application.py
@@ -122,6 +122,11 @@ class ApplicationSpec(NuclioSpec):
             state_thresholds=state_thresholds,
             disable_default_http_trigger=disable_default_http_trigger,
         )
+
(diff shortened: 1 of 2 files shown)


mlrun/runtimes/nuclio/application/application.py
```diff
@@ -122,6 +122,11 @@ def __init__(
             state_thresholds=state_thresholds,
             disable_default_http_trigger=disable_default_http_trigger,
         )
+
+        # Override default min/max replicas (don't assume application is stateless)
+        self.min_replicas = min_replicas or 1
+        self.max_replicas = max_replicas or 1
```
TomerShor, line 128: you can do it in the call to `__init__` (lines 91-92)

alonmr: I think its more explicit to have it here. If you do it in the super call it can be easy to miss

TomerShor (approved):

PR #6179, opened 2024-08-18
[Model Monitoring] Add `error_handler` step to application graph

Eyal-Danieli:
Add `error_handler` step to handle errors in the model monitoring application. Until now, the user didn't have any details about issues in his application.
In this PR, we generate an event using `error_handler` that provides information about the error. 

Related JIRA: https://iguazio.atlassian.net/browse/ML-7611


mlrun/model_monitoring/applications/_application_steps.py
```diff
@@ -164,3 +165,40 @@ def _create_mlrun_context(app_name: str):
...
+    def __init__(self, project: str, name: Optional[str] = None):
+        self.project = project
+        self.name = name or "ApplicationErrorHandler"
+
+    def do(self, event):
+        """
+        Handle model monitoring application error. This step will generate an event, describing the error.
+
+        :param event: Application event.
+        :return: Application event.
```
gtopper: It doesn't return anything.
```suggestion
```

tests/system/model_monitoring/assets/application.py
```diff
@@ -114,3 +114,14 @@ def do_tracking(
...
+
+class ErrApp(ModelMonitoringApplicationBaseV2):
+    NAME = "err-app"
+
+    def do_tracking(
+        self,
+        monitoring_context: mm_context.MonitoringApplicationContext,
+    ) -> list[mm_results.ModelMonitoringApplicationResult]:
+        monitoring_context.logger.info("Running demo app")
+        raise ValueError(f" This is an ERROR from {self.NAME} app!")
```
gtopper:
```suggestion
        raise ValueError(f"This is an ERROR from {self.NAME} app!")
```

tests/system/model_monitoring/test_app.py
```diff
@@ -423,6 +425,12 @@ def custom_setup_class(cls) -> None:
                 results={"data_drift_test"},
                 artifacts={"evidently_report", "evidently_suite", "dashboard"},
             ),
+            _AppData(
+                class_=ErrApp,
+                rel_path="assets/application.py",
+                results={},
+                artifacts={},
```
jond01: You don't need to specify empty values
```suggestion
```

tests/system/model_monitoring/test_app.py
```diff
@@ -191,11 +193,11 @@ def _test_tsdb_record(
...
 
         assert set(df.application_name) == {
-            app_data.class_.NAME for app_data in cls.apps_data
+            app_data.class_.NAME for app_data in cls.apps_data[:-1]
         }, "The application names are different than expected"
 
         tsdb_metrics = df.groupby("application_name").result_name.unique()
-        for app_data in cls.apps_data:
+        for app_data in cls.apps_data[:-1]:
             if app_metrics := app_data.results:
```
jond01: Unneeded, the slicing is confusing and error prone

tests/system/model_monitoring/test_app.py
```diff
@@ -146,7 +148,7 @@ def _test_results_kv_record(cls, ep_id: str) -> None:
 
     @classmethod
     def _test_metrics_kv_record(cls, ep_id: str) -> None:
-        for app_data in cls.apps_data:
+        for app_data in cls.apps_data[:-1]:
             if not app_data.metrics:
```
jond01: Same as the previous comment.

tests/system/model_monitoring/test_app.py
```diff
@@ -191,11 +193,11 @@ def _test_tsdb_record(
         ).all(), "The endpoint IDs are different than expected"
 
         assert set(df.application_name) == {
-            app_data.class_.NAME for app_data in cls.apps_data
+            app_data.class_.NAME for app_data in cls.apps_data[:-1]
         }, "The application names are different than expected"
```
jond01: Add the relevant condition instead of slicing.

tests/system/model_monitoring/test_app.py
```diff
@@ -127,7 +129,7 @@ def custom_setup(cls, project_name: str) -> None:
 
     @classmethod
     def _test_results_kv_record(cls, ep_id: str) -> None:
-        for app_data in cls.apps_data:
+        for app_data in cls.apps_data[:-1]:
             app_name = app_data.class_.NAME
```
jond01: Same as the previous comment.

jond01 (commented):
I'm afraid four applications in `TestAppFlow` is too much. How much time did it take to run?

tests/system/model_monitoring/assets/application.py
```diff
@@ -114,3 +114,14 @@ def do_tracking(
...
+
+
+class ErrApp(ModelMonitoringApplicationBaseV2):
+    NAME = "err-app"
+
+    def do_tracking(
+        self,
+        monitoring_context: mm_context.MonitoringApplicationContext,
+    ) -> list[mm_results.ModelMonitoringApplicationResult]:
+        monitoring_context.logger.info("Running demo app")
```
jond01: Okay to skip, but I think it's better to differentiate from the other demo app log:
```suggestion
        monitoring_context.logger.info("Running error app")
```

mlrun/model_monitoring/applications/_application_steps.py
```diff
@@ -164,3 +165,39 @@ def _create_mlrun_context(app_name: str):
...
+        self.name = name or "ApplicationErrorHandler"
+
+    def do(self, event):
+        """
+        Handle model monitoring application error. This step will generate an event, describing the error.
+
+        :param event: Application event.
+        """
+
+        logger.error(f"Error in application step: {event}")
```
jond01, line 182: I think this log should be debug/info, an error log has already been raised in the previous MM application step.
Here it's in the normal flow of the step.

jond01 (approved):
Nice!

mlrun/model_monitoring/applications/_application_steps.py
```diff
@@ -164,3 +165,39 @@ def _create_mlrun_context(app_name: str):
...
+
+        logger.info("Generating event for the error")
```
theSaarco: Don't see a lot of value in this log.

mlrun/common/schemas/alert.py
```diff
@@ -57,6 +58,7 @@ class EventKind(StrEnum):
     EventKind.SYSTEM_PERFORMANCE_SUSPECTED: [EventEntityKind.MODEL_ENDPOINT_RESULT],
     EventKind.MM_APP_ANOMALY_DETECTED: [EventEntityKind.MODEL_ENDPOINT_RESULT],
     EventKind.MM_APP_ANOMALY_SUSPECTED: [EventEntityKind.MODEL_ENDPOINT_RESULT],
+    EventKind.MM_APP_FAILED: [EventEntityKind.MODEL_ENDPOINT_RESULT],
```
theSaarco: It would have been nicer to have a new type of `EventEntityKind` that would point at a MM app, rather than a result (which should conceptually be pointing at an app+result, IIUC). 
I see this type was already used for other types of events, so this is not a new thing, but still - is this something that was considered?
As a reference - when generating drift events which use the same type of entity, the `id` that is used is the result fqn, not just the app name.

theSaarco (commented):
Please see my comment, don't know how difficult it would be to change that.

mlrun/common/schemas/alert.py
Eyal-Danieli: We considered it (@Yacouby ) and there were few concerns, mainly about confusing with `MODEL_ENDPOINT_RESULT` but your suggestion is indeed a proper approach. I'll add a new `EventEntityKind` called `MODEL_MONITORING_APPLICATION` which will handle only the new `mm_app_failure` event at the moment but in the future maybe we will use it for more events such as amount of app invocations, app performance indication, etc.

mlrun/model_monitoring/applications/_application_steps.py
Eyal-Danieli: amm you right but the user has no idea that he is in the "error step" so I think `logger.error` provides a good indication for post analysis when searching for the relevant logs.

theSaarco (approved):
LGTM 👍

PR #6181, opened 2024-08-19
[Notification] Fix `send_start_notification` param not propagating to scheduled workflows

quaark:

Passing the `send_start_notification` from SDK all the way to the `load_and_run` function so it affects scheduled pipelines as well.

### DIFF
--- a/mlrun/common/schemas/workflow.py
+++ b/mlrun/common/schemas/workflow.py
@@ -42,6 +42,7 @@ class WorkflowRequest(pydantic.BaseModel):
     run_name: typing.Optional[str] = None
     namespace: typing.Optional[str] = None
     notifications: typing.Optional[list[Notification]] = None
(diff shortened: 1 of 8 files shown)


liranbg (changes requested):
we need to have a better approach rather than sending this conditional arguemnt all over the flow

quaark:
> we need to have a better approach rather than sending this conditional arguemnt all over the flow

I agree, see my comment in https://iguazio.atlassian.net/browse/ML-6644?focusedCommentId=156240

quaark:
> > we need to have a better approach rather than sending this conditional arguemnt all over the flow
> 
> I agree, see my comment in https://iguazio.atlassian.net/browse/ML-6644?focusedCommentId=156240

https://github.com/mlrun/mlrun/pull/6205
Moved to new implementation here

PR #6182, opened 2024-08-19
[Projects] Fix log messages for `get_or_create_project` function

moranbental:
When creating a new project with `user_project=True`, the log messages for the project name were missing the username
https://iguazio.atlassian.net/browse/ML-7495

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -500,6 +500,7 @@ def get_or_create_project(
     :returns: project object
     """
     context = context or "./"
+    project_name = _add_username_to_project_name_if_needed(name, user_project)
     try:
         # load project from the DB.
         # use `name` as `url` as we load the project from the DB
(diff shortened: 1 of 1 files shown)


rokatyy (approved):
LGTM

mlrun/projects/project.py
```diff
@@ -500,6 +500,7 @@ def setup(project):
     :returns: project object
     """
     context = context or "./"
+    project_name = _add_username_to_project_name_if_needed(name, user_project)
```
alonmr: The project name is resolved in `load_project` which is done the same as you did here but now this code is coupled with `load_project` internal logic. If someone changes `load_project` they will need to change this as well. It's true that this logic is unlikely to change but I don't see a reason to duplicate this code. You can use `project.metadata.name` where you have the project OR add the `user_project` flag to the log kwargs where you don't.

moranbental: fixed

liranbg (approved):

PR #6183, opened 2024-08-19
[Alerts] Validate only the alert name in the `store_alert_config` function

moranbental:
When no name is provided in the alert config, the error message the user receives is `MLRunBadRequestError: Project and name must be provided`, although only the alert name should be required 

I also made some changes:
1. Added validation for the `alert_data` argument.
2. Changed the error raised for a missing alert name from `MLRunBadRequestError` to `MLRunInvalidArgumentError` to clarify the error, even though both raise a `BadRequest` (400) error code."

https://iguazio.atlassian.net/browse/ML-7489

### DIFF
--- a/mlrun/alerts/alert.py
+++ b/mlrun/alerts/alert.py
@@ -136,10 +136,6 @@ class AlertConfig(ModelObj):
         if template:
             self._apply_template(template)
 
-    def validate_required_fields(self):
-        if not self.name:
-            raise mlrun.errors.MLRunBadRequestError("Alert name must be provided")
-
     def _serialize_field(
         self, struct: dict, field_name: str = None, strip: bool = False
     ):
--- a/mlrun/alerts/alert.py
+++ b/mlrun/alerts/alert.py
@@ -136,6 +136,10 @@ class AlertConfig(ModelObj):
         if template:
             self._apply_template(template)
 
+    def validate_required_fields(self):
+        if not self.name:
+            raise mlrun.errors.MLRunInvalidArgumentError("Alert name must be provided")
+
     def _serialize_field(
         self, struct: dict, field_name: str = None, strip: bool = False
     ):
--- a/mlrun/alerts/alert.py
+++ b/mlrun/alerts/alert.py
@@ -137,8 +137,8 @@ class AlertConfig(ModelObj):
             self._apply_template(template)
 
     def validate_required_fields(self):
-        if not self.project or not self.name:
-            raise mlrun.errors.MLRunBadRequestError("Project and name must be provided")
+        if not self.name:
+            raise mlrun.errors.MLRunBadRequestError("Alert name must be provided")
 
     def _serialize_field(
         self, struct: dict, field_name: str = None, strip: bool = False
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -4224,7 +4224,6 @@ class HTTPRunDB(RunDBInterface):
             if isinstance(alert_data, AlertConfig)
             else AlertConfig.from_dict(alert_data)
         )
-        alert_instance.validate_required_fields()
 
         alert_data = alert_instance.to_dict()
         body = _as_json(alert_data)
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -4227,8 +4227,8 @@ class HTTPRunDB(RunDBInterface):
             if isinstance(alert_data, AlertConfig)
             else AlertConfig.from_dict(alert_data)
         )
-        # The validation must be here because the user can call this function directly
-        # via `mlrun.get_run_db().store_alert_config()`
+        # Validation is necessary here because users can directly invoke this function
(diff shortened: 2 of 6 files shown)


mlrun/alerts/alert.py
```diff
@@ -136,10 +136,6 @@ def __init__(
         if template:
             self._apply_template(template)
 
-    def validate_required_fields(self):
-        if not self.project or not self.name:
-            raise mlrun.errors.MLRunBadRequestError("Project and name must be provided")
-
     def _serialize_field(
```
yaelgen, line 143: Are we sure that no other place or any user uses this function? I see that from the server side we really enforce that `name` must be provided, but just to make sure that it wasn't in the httpdb for a reason 
@Yacouby do you have any additional insights or historical context on this?

tests/projects/test_project.py
```diff
@@ -2314,6 +2315,27 @@ def test_workflow_path_with_project_workdir():
...
 
 
+@pytest.mark.parametrize(
+    "alert_name_in_config, alert_name_as_func_param",
+    [
+        (None, None),
+        (None, ""),
+        ("", None),
+        ("", ""),
+    ],
```
yaelgen: Maybe add a case where alert_name_in_config or  alert_name_as_func_param  is provided and no err is raised?

yaelgen (changes requested):
LGTM, minor

mlrun/alerts/alert.py
moranbental: @yaelgen I didn't see any place where this function is used in the code.

mlrun/db/httpdb.py
```diff
@@ -4224,7 +4224,6 @@ def store_alert_config(
             if isinstance(alert_data, AlertConfig)
             else AlertConfig.from_dict(alert_data)
         )
-        alert_instance.validate_required_fields()
```
Yacouby, line 4227: its not a good idea to remove this from here, since the user can call this function directly ( mlrun.get_run_db().store_alert_config(..) ) and then in this case we do not do the validation.

Yacouby (changes requested):

mlrun/alerts/alert.py
Yacouby: i added a comment on this : https://github.com/mlrun/mlrun/pull/6183#discussion_r1721851851

tests/projects/test_project.py
moranbental: @yaelgen It has been moved to an integration test (as described in the PR description), and a test for that already exists.

tests/integration/sdk_api/alerts/test_alerts.py
```diff
@@ -283,6 +283,28 @@ def test_alert_templates(self):
...
+        project_name = "my-project"
+        project = mlrun.new_project(project_name)
+        alert_data = mlrun.alerts.alert.AlertConfig(
+            name=alert_name_in_config, project=None
+        )
+        with pytest.raises(
+            mlrun.errors.MLRunInvalidArgumentError, match="Alert name must be provided"
+        ):
+            project.store_alert_config(alert_data, alert_name=alert_name_as_func_param)
+
```
yaelgen: Integration tests tend to have a long runtime, so we typically avoid using parametrize to prevent further extending the execution time. However, I believe this test would be better suited as a unit test under `httpdb` tests, rather than remaining in the projects tests as integration.

yaelgen (changes requested):

mlrun/projects/project.py
```diff
@@ -4061,7 +4061,7 @@ def delete_api_gateway(
         mlrun.db.get_run_db().delete_api_gateway(name=name, project=self.name)
 
     def store_alert_config(
-        self, alert_data: AlertConfig, alert_name=None
+        self, alert_data: AlertConfig, alert_name: str = None
```
rokatyy: Optional[str]

tests/integration/sdk_api/alerts/test_alerts.py
moranbental: @yaelgen Which exact file are you referring to? Most unit tests for httpdb that I’ve seen only interact with sqldb.py or nopdb.py, not the httpdb file itself.

rokatyy (approved):

mlrun/db/httpdb.py
```diff
@@ -4224,6 +4227,8 @@ def store_alert_config(
             if isinstance(alert_data, AlertConfig)
             else AlertConfig.from_dict(alert_data)
         )
+        # The validation must be here because the user can call this function directly
+        # via `mlrun.get_run_db().store_alert_config()`
```
yaelgen:
```suggestion
    # Validation is necessary here because users can directly invoke this function
    # through `mlrun.get_run_db().store_alert_config()`.

```

yaelgen (approved):
LGTM 🔥  minor suggestion

Yacouby (approved):

liranbg (approved):

PR #6184, opened 2024-08-19
[Datastore] Prevent credentials from being exposed in the graph

alxtkr77:
https://iguazio.atlassian.net/browse/ML-6745
https://iguazio.atlassian.net/browse/ML-7622


mlrun/datastore/storytargets.py
```diff
@@ -0,0 +1,157 @@
...
+    storage_options = store.get_storage_options()
+    if storage_options and external_storage_options:
+        # merge external storage options with the store's storage options. storage_options takes precedence
+        storage_options = merge(external_storage_options, storage_options)
+    else:
+        storage_options = storage_options or external_storage_options
+    return url, storage_options
+
+
+class TDEngineStoreyTarget(storey.TDEngineTarget):
```
gtopper: Well, it you put it in its own module, you don't have to put storey in each class name.
```suggestion
class TDEngineTarget(storey.TDEngineTarget):
```

mlrun/datastore/storytargets.py
```diff
@@ -0,0 +1,157 @@
...
+from storey import V3ioDriver
+
+import mlrun
+
+from .utils import (
+    parse_kafka_url,
+)
+
+"""
+Storey targets expect storage_options, which may contain credentials.
```
gtopper: Storey is indeed spelled with an e. I'd rename this file from `storytargets` to just `storey`.

mlrun/datastore/storytargets.py
```diff
@@ -0,0 +1,157 @@
...
+To avoid passing it openly within the graph, we use those wrapper instead.
```
gtopper:
```suggestion
To avoid passing it openly within the graph, we use wrapper classes.
```

mlrun/datastore/__init__.py
```diff
@@ -22,6 +22,14 @@
...
     "StreamTarget",
     "KafkaTarget",
+    "ParquetStoreyTarget",
+    "CSVStoreyTarget",
+    "NoSqlStoreyTarget",
+    "StreamStoreyTarget",
+    "KafkaStoreyTarget",
+    "RedisNoSqlStoreyTarget",
+    "TSDBStoreyTarget",
+    "TDEngineStoreyTarget",
```
gtopper: Do we need to export this? It seems like it's meant for internal use to me.

mlrun/datastore/storytargets.py
```diff
@@ -0,0 +1,157 @@
...
+                "KafkaTarget requires a path (topic)"
+            )
+        kwargs["brokers"] = brokers
+        kwargs["topic"] = topic
+        super().__init__(*args, **kwargs, **attributes)
+
+
+class NoSqlStoreyTarget(storey.NoSqlTarget):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
gtopper: Seems like this one doesn't actually do anything. I don't think we need to force it where it's not necessary.

mlrun/datastore/storytargets.py
```diff
@@ -0,0 +1,157 @@
...
+    def __init__(self, *args, **kwargs):
+        path = kwargs.pop("path")
+        endpoint, uri = mlrun.datastore.targets.redis_get_server_endpoint(path)
+        kwargs["path"] = endpoint + "/" + uri
+        super().__init__(*args, **kwargs)
+
+
+class TSDBStoreyTarget(storey.TSDBTarget):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
gtopper: Seems like this one doesn't actually do anything. I don't think we need to force it where it's not necessary.

mlrun/datastore/targets.py
```diff
@@ -1486,48 +1476,51 @@ def prepare_spark_df(self, df, key_columns, timestamp_key=None, spark_options=No
         return df
 
 
+def redis_get_server_endpoint(path):
```
gtopper: Probably it still makes sense to put it in `RedisNoSqlTarget`, even if as a staticmethod.

mlrun/datastore/targets.py
```diff
@@ -1568,6 +1561,44 @@ def prepare_spark_df(self, df, key_columns, timestamp_key=None, spark_options=No
...
+        )
+        if not self.columns:
+            aggregate_features = (
+                [key for key, feature in features.items() if feature.aggregate]
+                if features
+                else []
+            )
+            column_list = [
+                col for col in column_list if col[0] not in aggregate_features
+            ]
```
gtopper: Looks like this duplicated code can be moved to the base class to avoid duplicating these lines.

mlrun/datastore/targets.py
```diff
@@ -1697,21 +1721,9 @@ def add_writer_step(
...
-        else:
-            attributes = copy(self.attributes)
-            brokers = attributes.pop(
-                "brokers", attributes.pop("bootstrap_servers", None)
-            )
-            topic, brokers = parse_kafka_url(self.get_target_path(), brokers)
+        path = self.get_target_path()
 
-        if not topic:
+        if not path:
```
gtopper, line 1694: Isn't `path` _not_ going to be defined if we rely on a datastore profile?

gtopper: Seems like the code moved to the wrapper and this bit should have been removed from here too.

mlrun/datastore/storytargets.py
alxtkr77: I prefer to have longer names rather then objects with the same name in different modules

mlrun/datastore/__init__.py
alxtkr77: I believe we do, since we use 'mlrun.datastore.StreamStoreyTarget' for example when we add step to graph

mlrun/datastore/storytargets.py
alxtkr77: I disagree. Better to have a clear rule to always use wrapper xxxStoreyTarget rather then sometimes storey.* and sometimes mlrun.datastore.*

mlrun/datastore/storytargets.py
alxtkr77: I disagree. Better to have a clear rule to always use wrapper xxxStoreyTarget rather then sometimes storey.* and sometimes mlrun.datastore.*

mlrun/datastore/targets.py
alxtkr77: In case of datastore the 'path' will be defined as 'ds://profilename/...'

mlrun/datastore/storytargets.py
alxtkr77: right, will rename

mlrun/datastore/targets.py
alxtkr77: will do

mlrun/datastore/targets.py
alxtkr77: ok

mlrun/datastore/storeytargets.py
```diff
@@ -0,0 +1,159 @@
...
+        # merge external storage options with the store's storage options. storage_options takes precedence
+        storage_options = merge(external_storage_options, storage_options)
+    else:
+        storage_options = storage_options or external_storage_options
+    return url, storage_options
+
+
+class TDEngineStoreyTarget(storey.TDEngineTarget):
+    def __init__(self, *args, **kwargs):
+        kwargs["url"] = mlrun.model_monitoring.helpers.get_tsdb_connection_string()
```
gtopper, line 45: Need to
```python
```
even if it somehow works without.

mlrun/datastore/__init__.py
gtopper: I'm pretty sure this is going to pollute the API that's visible to the user (most noticeably, the docs).

mlrun/datastore/storytargets.py
gtopper: Ok, but you can remove the redundant `__init__`.

mlrun/datastore/storytargets.py
gtopper: Ok, but you can remove the redundant `__init__.`

mlrun/datastore/__init__.py
```diff
@@ -63,6 +63,16 @@
...
 )
+from .storeytargets import (
+    CSVStoreyTarget,
+    KafkaStoreyTarget,
+    NoSqlStoreyTarget,
+    ParquetStoreyTarget,
+    RedisNoSqlStoreyTarget,
+    StreamStoreyTarget,
+    TDEngineStoreyTarget,
+    TSDBStoreyTarget,
```
gtopper: As discussed, it looks like these will be visible to the user, and even appear in API docs.

mlrun/datastore/storeytargets.py
```diff
@@ -0,0 +1,158 @@
...
+        if not topic:
+            raise mlrun.errors.MLRunInvalidArgumentError("KafkaTarget requires a topic")
+        kwargs["brokers"] = brokers
+        kwargs["topic"] = topic
+        super().__init__(*args, **kwargs, **attributes)
+
+
+class NoSqlStoreyTarget(storey.NoSqlTarget):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
gtopper: Even if we insist on the wrapper, we can still remove the boilerplate
```suggestion
    pass
```

mlrun/datastore/storeytargets.py
```diff
@@ -0,0 +1,158 @@
...
+        endpoint, uri = mlrun.datastore.targets.RedisNoSqlTarget.get_server_endpoint(
+            path
+        )
+        kwargs["path"] = endpoint + "/" + uri
+        super().__init__(*args, **kwargs)
+
+
+class TSDBStoreyTarget(storey.TSDBTarget):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
gtopper: Even if we insist on the wrapper, we can still remove the boilerplate
```suggestion
    pass
```

mlrun/datastore/storeytargets.py
```diff
@@ -0,0 +1,158 @@
...
+        url, storage_options = get_url_and_storage_options(
+            path, external_storage_options
+        )
+
+        if storage_options:
+            kwargs["storage_options"] = storage_options
+        if args:
+            args[0] = url
+        if "path" in kwargs:
+            kwargs["path"] = url
```
gtopper: This 14-line block is duplicated in `CSVStoreyTarget`, and can be deduplicated with a static method, or multiple inheritance (but better the former).

gtopper (commented):
Looks good otherwise.

mlrun/datastore/storeytargets.py
```diff
@@ -0,0 +1,158 @@
...
+    def __init__(self, *args, **kwargs):
+        args = list(args)
+
+        path = args[0] if args else kwargs.get("stream_path")
+        endpoint, storage_options = get_url_and_storage_options(path)
+
+        if not path:
+            raise mlrun.errors.MLRunInvalidArgumentError("StreamTarget requires a path")
+
+        access_key = storage_options.get("v3io_access_key")
```
gtopper, line 91: I just noticed this. `get_url_and_storage_options` can return `None`, but is assumed to always return dict. Looks like a bug.

mlrun/datastore/storeytargets.py
```diff
@@ -0,0 +1,146 @@
...
+    def __init__(self, *args, **kwargs):
+        args = list(args)
+
+        path = args[0] if args else kwargs.get("stream_path")
+        endpoint, storage_options = get_url_and_storage_options(path)
+
+        if not path:
+            raise mlrun.errors.MLRunInvalidArgumentError("StreamTarget requires a path")
+
+        access_key = storage_options.get("v3io_access_key") if storage_options else None
```
gtopper: I guess I would have made `get_url_and_storage_options` always return a dictionary. Otherwise, future callers will also have to deal with this easter egg.

Either way, do we have test coverage for it? Since apparently it was a real bug.

gtopper: Ok, apparently it wasn't a bug (only a potential bug), because storage options are always defined in this target, according to @alxtkr77.

gtopper (approved):

PR #6191, opened 2024-08-20
[Model Monitoring] Manage `last_analyzed` exclusively in the controller

jond01:
Use JSON files, one per endpoint.
Implements [ML-7457](https://iguazio.atlassian.net/browse/ML-7457).

Related CE config update PR:
https://github.com/mlrun/ce/pull/141

[ML-7457]: https://iguazio.atlassian.net/browse/ML-7457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/model_monitoring/controller.py
```diff
@@ -41,173 +39,174 @@
...
+            str,
+            mlrun.mlconf.get_model_monitoring_file_target_path(
+                project=project,
+                kind=mm_constants.FileTargetKind.MONITORING_SCHEDULES,
+            ),
+        ),
+        f"{endpoint}.json",
+    )
+
+
```
davesh0812: please move this function to `model_monitoring.helpers`.

mlrun/model_monitoring/controller.py
```diff
@@ -41,173 +39,174 @@
...
     end: datetime.datetime
 
 
+def _get_monitoring_schedules_file_path(*, project: str, endpoint: str) -> str:
+    return os.path.join(
+        cast(
+            str,
+            mlrun.mlconf.get_model_monitoring_file_target_path(
+                project=project,
+                kind=mm_constants.FileTargetKind.MONITORING_SCHEDULES,
```
davesh0812: You are using the online target logic (in spite it is an offline target), it means that you are always  get the  `mlrun.mlconf.model_endpoint_monitoring.store_prefixes.default` configuration.
In iguazio it will work fine but in CE I didn't saw that we overwrite this configuration (in default it using v3io).

mlrun/model_monitoring/controller.py
```diff
@@ -41,173 +39,174 @@
...
+        ) as err:
+            if isinstance(err, BotoClientError):
+                # accept only "NoSuchKey" errors codes - equivalent to `FileNotFoundError`
+                if err.response["Error"]["Code"] != "NoSuchKey":
+                    raise
+            last_analyzed = self._init_last_analyzed()
+            schedules = _Schedules({self._application: last_analyzed})
+            logger.info(
+                "The monitoring schedules file does not exist for this endpoint, creating it",
+                path=self._db.url,
```
davesh0812: Please add the creation of this json to be part of the model endpoint creation, and also we need to delete it when calling to `delete_model_endpoints_resources`.

mlrun/model_monitoring/controller.py
```diff
@@ -41,173 +39,174 @@
 )
 from mlrun.utils import datetime_now, logger
 
+_UTF_8 = "utf-8"
+_SECONDS_IN_DAY = int(datetime.timedelta(days=1).total_seconds())
+
+
+# `Schedules` is a dictionary of registered application name as key and Unix timestamp as value
+_Schedules = NewType("_Schedules", dict[str, int])
```
davesh0812: Can we create a `Schedules` class that get the json file and have few method like `get(), update(), update_db(), delete()` and some more maybe... I just think it can be more user friendly for us. And can be used also from  MEP creation/deletion API.

mlrun/model_monitoring/controller.py
```diff
@@ -41,173 +39,174 @@
...
+        for timestamp in range(self._start, self._stop - self._step + 1, self._step):
+            entered = True
+            start_time = datetime.datetime.fromtimestamp(
+                timestamp, tz=datetime.timezone.utc
+            )
+            end_time = datetime.datetime.fromtimestamp(
+                timestamp + self._step, tz=datetime.timezone.utc
+            )
+            yield _Interval(start_time, end_time)
+            self._update_last_analyzed(timestamp + self._step)
```
davesh0812: You don't think it is to much to update the store object on each app? Why no to update it one time per mep?
If we will have 10 apps and 100 mep we will reach to the db 1000 in each run of the controller.

davesh0812 (changes requested):
LookNice, I didn't finish all the  review.
Few important comments - the main one is that we have to delete/create the json when deleting/creating the model endpoint or delete it in case of project deletion.

mlrun/model_monitoring/controller.py
jond01: Right, I planned to update it in the CE config: https://github.com/mlrun/ce/pull/141

mlrun/model_monitoring/controller.py
jond01: The reason is to avoid data loss/duplication. These files are the persistent storage that ensures no loss upon unexpected controller failures.

Edit: Done - the mode endpoint schedules file is read once and written once per controller run.

mlrun/model_monitoring/controller.py
jond01: I think that I get what you mean, but this is basically the `DataItem` API we use here, isn't it?
I'll provide some more abstraction to reduce the duplication between MEP creation/deletion and the controller.

Edit: I united the APIs in the `ModelMonitoringSchedulesFile` class.

mlrun/model_monitoring/controller.py
```diff
@@ -45,164 +45,129 @@ class _Interval(NamedTuple):
...
                 application=self._application,
                 start=self._start,
                 stop=self._stop,
+                step=self._step,
             )
 
 
-class _BatchWindowGenerator:
-    def __init__(self, batch_dict: Union[dict, str]) -> None:
+class _BatchWindowGenerator(AbstractContextManager):
```
davesh0812, line 136: Why do we need to inherit from `AbstractContextManager`? 
You use composition relationship you with the `ModelMonitoringSchedulesFile` that already inherits `AbstractContextManager` no?

mlrun/model_monitoring/controller.py
```diff
@@ -491,18 +457,28 @@ def _push_to_applications(
...
+
+
+def handler(context: nuclio_sdk.Context, event: nuclio_sdk.Event) -> None:
     """
     Run model monitoring application processor
 
     :param context: the Nuclio context
     :param event:   trigger event
     """
+    context.platform.set_termination_callback(callback=_stop_running)
```
davesh0812: Can we run it only once when initializing the pod? Using `init_context(context)` facility of nuclio?

mlrun/model_monitoring/db/_schedules.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+        # `self._item` is the persistent version of the monitoring schedules.
+        self._item = mlrun.model_monitoring.helpers.get_monitoring_schedules_data(
+            project=project, endpoint_id=endpoint_id
+        )
+        self._path = self._item.url
+        self._fs = cast(fsspec.AbstractFileSystem, self._item.store.filesystem)
+        # `self._schedules` is an in-memory copy of the DB for all the applications for
+        # the same model endpoint.
+        self._schedules: dict[str, int] = {}
```
davesh0812: Needs to align with the actual inmemory data (self._item)

mlrun/model_monitoring/db/_schedules.py
```diff
@@ -0,0 +1,113 @@
...
+            )
+            self._item.delete()
+        else:
+            logger.debug(
+                "Model monitoring schedules file does not exist, nothing to delete",
+                path=self._item.url,
+            )
+
+    def _open(self) -> None:
+        self._schedules = json.loads(self._item.get().decode(encoding=self.ENCODING))
```
davesh0812: what happens if somehow there is no file created?

davesh0812 (commented):
Looks Good!
Few comments and questions :)
It seems that a few tests related to Windows machine updates were removed. Could we add those tests back under test_schedules.py?

mlrun/model_monitoring/db/_schedules.py
jond01: It's done in the context manager's `__enter__` method or `_open` method.
The constructor doesn't do IO intentionally, as it's unexpected and the reading is not always required - for example when creating a new file or deleting an existing file.

mlrun/model_monitoring/db/_schedules.py
jond01: Depending on the store, an error will be raised - `mlrun.errors.MLRunNotFoundError` or `BotoClientError`.

mlrun/model_monitoring/controller.py
jond01: Although the inheritance is not a must, it reminds us to use these classes as context managers when working with the underlying persistent data. It applies to both classes.
I get it might be irregular to see this. Do you have a suggestion?

mlrun/model_monitoring/db/_schedules.py
jond01: Added a flag and checks in 39e41ff 🙂

mlrun/model_monitoring/db/_schedules.py
jond01: Added try-except and logs in 935e89d 👍

davesh0812 (approved):
Very Nice!!

PR #6192, opened 2024-08-20
[Application Runtime] Pass body when invoke api gateway

rokatyy:
Jira -  https://iguazio.atlassian.net/browse/ML-7573

### DIFF
--- a/mlrun/runtimes/nuclio/api_gateway.py
+++ b/mlrun/runtimes/nuclio/api_gateway.py
@@ -386,7 +386,7 @@ class APIGateway(ModelObj):
         headers: dict = None,
         credentials: Optional[tuple[str, str]] = None,
         path: Optional[str] = None,
-        body: typing.Union[str, bytes, dict] = None,
+        body: Optional[Union[str, bytes, dict]] = None,
         **kwargs,
     ):
         """
--- a/mlrun/runtimes/nuclio/api_gateway.py
+++ b/mlrun/runtimes/nuclio/api_gateway.py
@@ -386,6 +386,7 @@ class APIGateway(ModelObj):
         headers: dict = None,
         credentials: Optional[tuple[str, str]] = None,
         path: Optional[str] = None,
+        body: typing.Union[str, bytes, dict] = None,
         **kwargs,
     ):
         """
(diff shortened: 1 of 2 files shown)


mlrun/runtimes/nuclio/application/application.py
```diff
@@ -509,6 +509,13 @@ def invoke(
 
         if not method:
             method = "POST" if body else "GET"
+
+        # Determine the correct keyword argument for the body
+        if isinstance(body, dict):
+            http_client_kwargs["json"] = body
+        elif isinstance(body, (str, bytes)):
+            http_client_kwargs["data"] = body
```
TomerShor: We should add the `body` argument to the api_gateway's `invoke` function, and this should be moved there.
Then body will be supported both in application and api gateway in general

TomerShor (changes requested):
Looks fine but have a suggestion

mlrun/runtimes/nuclio/application/application.py
rokatyy: @TomerShor api gw already supports data/json - whatever user sets in kwargs, it's just a matter of decision, but I think both options are fine, so will move to gws

mlrun/runtimes/nuclio/api_gateway.py
```diff
@@ -386,6 +386,7 @@ def invoke(
         headers: dict = None,
         credentials: Optional[tuple[str, str]] = None,
         path: Optional[str] = None,
+        body: typing.Union[str, bytes, dict] = None,
```
TomerShor:
```suggestion
        body: Optional[typing.Union[str, bytes, dict]] = None,
```

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -509,6 +509,7 @@ def invoke(
 
         if not method:
             method = "POST" if body else "GET"
+
         return self.status.api_gateway.invoke(
```
TomerShor, line 513: Add `body=body` to the invoke call

TomerShor (changes requested):

mlrun/runtimes/nuclio/application/application.py
rokatyy: right, nice catch

moranbental (approved):
LGTM

TomerShor (approved):

PR #6194, opened 2024-08-21
[Project delete] on failure to delete `kafka` stream or `tdengine` TSDB, do not fail the project deletion 

Eyal-Danieli:
This PR fix an issue in which the user has decided to delete his project but for some reason there is a communication issue with one of the external model monitoring resources that he defined (`kafka`/`tdengine`). In this case we will skip the error and delete other project resources. However, we do log the original error for the user and in addition suggesting him to delete there resources manually. 

Related JIRA: https://iguazio.atlassian.net/browse/ML-7545

### DIFF
--- a/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
+++ b/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
@@ -58,7 +58,12 @@ class TDEngineConnector(TSDBConnector):
         except taosws.QueryError:
             # Database already exists
             pass
-        conn.execute(f"USE {self.database}")
+        try:
+            conn.execute(f"USE {self.database}")
+        except taosws.QueryError as e:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"Failed to use TDEngine database {self.database}, {mlrun.errors.err_to_str(e)}"
+            )
         return conn
 
     def _init_super_tables(self):
--- a/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
+++ b/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
@@ -61,7 +61,7 @@ class TDEngineConnector(TSDBConnector):
         try:
             conn.execute(f"USE {self.database}")
         except taosws.QueryError as e:
-            raise mlrun.errors.MLRunInvalidArgumentError(
+            raise mlrun.errors.MLRunTSDBConnectionFailure(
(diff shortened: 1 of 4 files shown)


server/api/crud/model_monitoring/deployment.py
```diff
@@ -1000,6 +1008,11 @@ def _delete_model_monitoring_stream_resources(
                     topics=topics,
                     error=mlrun.errors.err_to_str(e),
                 )
+            except kafka.errors.NoBrokersAvailable as e:
+                # Raise an error that will be caught by the called and skip the deletion of the stream
```
gtopper: If it will always be caught by the caller, why even bother raising it and catching it again? Also, a typo:
```suggestion
                # Raise an error that will be caught by the caller and skip the deletion of the stream
```

server/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -544,7 +544,13 @@ def delete_model_endpoints_resources(
                         project=project_name
                     ),
                 )
-
+            except mlrun.errors.MLRunInvalidArgumentError as e:
+                logger.warning(
+                    "Failed to delete TSDB resources, you may need to delete them manually",
```
gtopper, line 549: A bit nitpicky, but all the other errors say `Can't delete` and this one says `Failed to delete`.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -58,7 +58,12 @@ def _create_connection(self):
         except taosws.QueryError:
             # Database already exists
             pass
-        conn.execute(f"USE {self.database}")
+        try:
+            conn.execute(f"USE {self.database}")
+        except taosws.QueryError as e:
+            raise mlrun.errors.MLRunInvalidArgumentError(
```
assaf758: Maybe use MLRunNotImplementedServerError?
Using MLRunInvalidArgumentError does not seem related here..

server/api/crud/model_monitoring/model_endpoints.py
gtopper: I'd say `Failed to delete` is clearer btw.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
Eyal-Danieli: amm `MLRunNotImplementedServerError` seems to be related to not supported features, what do you think about `MLRunInternalServerError`? because basically mlrun server is not able to connect these external services so maybe it does make sense.

assaf758 (approved):

PR #6197, opened 2024-08-21
[SDK] Add validation of node selectors at the project and function levels

yaelgen:
resolves: https://iguazio.atlassian.net/browse/ML-7538

This PR introduces validation for node selectors in the following areas:

Functions:
- Validation raise a warning to the user when using `with_node_selection` across all runtimes.

Projects:
- Validation is applied during project creation/update (BE side)

Runtime:
- Any invalid node selector that bypasses initial validation will fail at runtime (during run enrichment), with clear error messages identifying the invalid key and value.


mlrun/projects/project.py
```diff
@@ -3161,6 +3170,16 @@ def save_to_db(self, store=True):
...
+        if self.spec.default_function_node_selector:
+            try:
+                mlrun.utils.validate_node_selectors(
+                    node_selectors=self.spec.default_function_node_selector
+                )
+            except mlrun.errors.MLRunInvalidArgumentError as e:
+                new_message = (
+                    f"Project can't be saved, invalid node selectors defined: {e}"
+                )
+                raise mlrun.errors.MLRunInvalidArgumentError(new_message) from e
```
liranbg: why not letting the backend perform the validation?

mlrun/runtimes/utils.py
```diff
@@ -457,6 +457,17 @@ def resolve_node_selectors(
...
+            mlrun.utils.validate_node_selectors(
+                {
+                    **config_node_selector,
+                    **project_node_selector,
+                    **instance_node_selector,
+                }
+            )
+        except mlrun.errors.MLRunInvalidArgumentError as e:
+            message = f"Validation failed for node selectors defined at the function, project or service level: {e}"
+            raise mlrun.errors.MLRunInvalidArgumentError(message) from e
```
liranbg: the added functionality does not adhere to function's purpose. please remove and let the callee decide when to call it.

liranbg (changes requested):
the validation should occur on backend side and not client. if the backend progress and allow lesser validation (for some reason) old clients would be stuck with old code.
in addition to that, node selectors are used for k8s, if api now uses them on another arbitrary platform then it would be better to change the validation on server only and not both server and client

mlrun/utils/helpers.py
```diff
@@ -1718,6 +1718,44 @@ def validate_single_def_handler(function_kind: str, code: str):
             )
 
 
+def validate_node_selectors(node_selectors: dict) -> None:
```
moranbental:
```suggestion
def validate_node_selectors(node_selectors: typing.Dict[str, str]) -> None:
```

moranbental (changes requested):
Looks good!

mlrun/projects/project.py
```diff
@@ -1040,6 +1040,15 @@ def remove_artifact(self, key):
         if key in self._artifacts:
             del self._artifacts[key]
 
+    @property
+    def default_function_node_selector(self):
+        return self._default_function_node_selector
+
+    @default_function_node_selector.setter
+    def default_function_node_selector(self, node_selector):
```
```suggestion
    def default_function_node_selector(self, node_selector : typing.Dict[str, str]):
```

moranbental (changes requested):

mlrun/utils/helpers.py
yaelgen: linter raised a warning for `typing.Dict[str, str]`, I used `dict[str, str]` instead

mlrun/utils/helpers.py
```diff
@@ -1718,6 +1718,75 @@ def validate_single_def_handler(function_kind: str, code: str):
             )
 
 
+def validate_node_selectors(
```
moranbental: The function will never return False, so why is it returning a value at all?

moranbental (changes requested):

mlrun/utils/helpers.py
yaelgen: I saw other validation functions we are using returns a Boolean, so wanted to keep the same behavior

moranbental: But you never actually return False

yaelgen: I am, when `raise_on_error` is False, after the warning (line 1747)

moranbental: No, it is not. It will only work if you change the code to:
```
            if not pattern.match(item):
                if not handle_invalid(error_message):
                     return False
```
then the function `validate_node_selectors` will return False

moranbental (changes requested):

mlrun/utils/helpers.py
yaelgen: why? the return is happening after warning, it's not exiting the function.

moranbental (approved):
LGTM

mlrun/utils/helpers.py
yaelgen: Added a comment to explain :)

yaelgen:
> the validation should occur on backend side and not client. if the backend progress and allow lesser validation (for some reason) old clients would be stuck with old code. in addition to that, node selectors are used for k8s, if api now uses them on another arbitrary platform then it would be better to change the validation on server only and not both server and client

After discussion, the backend now enforces validation during project and function creation or updates. The validation for the `with_node_selection` function on client side is retained, but it will only issue a warning.

mlrun/utils/helpers.py
```diff
@@ -1718,6 +1718,77 @@ def validate_single_def_handler(function_kind: str, code: str):
...
+                f"Invalid Kubernetes value '{value}' for key '{key}' in the node selector field. "
+                "Value must start and end with an alphanumeric character (a–z, A–Z, 0–9), "
+                "be up to 63 characters long, and may contain '-', '_', and '.'.",
+            )
+            if value
+            else None,
+        ]
+
+        # Apply validation rules
+        for item, pattern, error_message in filter(None, rules):
```
TomerShor: While this code is correct, IMO it is a case where trying to optimize code causes it to be less clear to read.

1. Instead of tuples you can use a dictionary, where the keys are `"item, pattern, error_message"`, then you would see what is each field without the need to see the for loop below.
2. Instead of adding conditional `None`s and then filtering the out, to me the naive approach is a lot clearer:
```python
rules = [..]
if prefix:
    rules.append(...)
...
```

TomerShor (commented):
A thought about readability

mlrun/utils/helpers.py
```diff
@@ -1718,6 +1718,91 @@ def validate_single_def_handler(function_kind: str, code: str):
...
+                f"the node selector you've set does not comply with the validation rules. "
+                f"Check our documentation for the latest validation updates. "
+                f"If there's a newer validation standard, consider upgrading. "
+                f"Be aware that your function executions may fail if the node selector remains invalid."
+            )
+
+    node_selectors = node_selectors or {}
+    for key, value in node_selectors.items():
+        # Split key into prefix and name if applicable
+        prefix, name = key.split("/", 1) if "/" in key else ("", key)
```
liranbg: looks like a code-dup of `verify_label_key` / `verify_label_value` - perhaps we can merge behavior or so

mlrun/utils/helpers.py
```diff
@@ -1718,6 +1718,91 @@ def validate_single_def_handler(function_kind: str, code: str):
...
+        if raise_on_error:
+            raise mlrun.errors.MLRunInvalidArgumentError(message)
+        else:
+            warnings.warn(
+                f"{message} \nFor your current SDK version, "
+                f"the node selector you've set does not comply with the validation rules. "
+                f"Check our documentation for the latest validation updates. "
+                f"If there's a newer validation standard, consider upgrading. "
+                f"Be aware that your function executions may fail if the node selector remains invalid."
+            )
```
TomerShor: The validation failed not because the SDK version is old, it's because the node selector doesn't adhere to **k8s'** naming rules.
I wouldn't go into suggesting "try that, or maybe this, but be careful". It failed because it's not valid, and that's it.

mlrun/runtimes/sparkjob/spark3job.py
```diff
@@ -22,6 +22,7 @@
 from mlrun.config import config
 
 from ...execution import MLClientCtx
+from ...k8s_utils import validate_node_selectors
```
moranbental: use `import X` instead of `from X import Y`

server/api/db/sqldb/db.py
```diff
@@ -49,6 +49,7 @@
 )
 from mlrun.config import config
 from mlrun.errors import err_to_str
+from mlrun.k8s_utils import validate_node_selectors
```
moranbental: use `import X` instead of `from X import Y`

moranbental (changes requested):

mlrun/k8s_utils.py
```diff
@@ -185,3 +196,38 @@ def verify_label_value(value, label_key):
...
+
+    # Helper function for handling errors or warnings
+    def handle_invalid(message):
+        if raise_on_error:
+            raise
+        else:
+            warnings.warn(
+                f"{message}\n"
+                f"The node selector you’ve set does not meet the validation rules for the current Kubernetes version. "
+                f"Please note that invalid node selectors may cause function execution failures."
```
TomerShor:
```suggestion
                f"Please note that invalid node selectors may cause issues with function scheduling."
```

TomerShor (commented):
Minor log change, but otherwise LGTM!

TomerShor (approved):
👏

liranbg (approved):

PR #6205, opened 2024-08-26
[Notifications] Align Remote Workflow Start Notification Behavior

quaark:

We are removing the `send_start_notification` param and aligning both manual and scheduled workflows to not send a start notification by default. Instead, in both scenarios we send a start notification **ONLY** when a notification with `when=running` is defined in the notification list of the workflow.


mlrun/projects/pipelines.py
```diff
@@ -1078,6 +1060,13 @@ def load_and_run(
     if load_only:
         return
 
+    # extract "start" notification if exists
+    start_notifications = [
+        mlrun.model.Notification.from_dict(notification)
```
theSaarco, line 1064: But `notification` is of type `typing.Union[mlrun.model.Notification, dict]` - this won't work for a `Notification` object...

mlrun/projects/pipelines.py
```diff
@@ -1078,6 +1060,13 @@ def load_and_run(
...
+        for notification in notifications
+        if "running" in notification
```
theSaarco: Same here - it won't work for a `Notification`.
Actually, I don't think this will work for a `dict` as well - checking for `in` will only be true if the dict key is `running`, won't it? The key is `when` in our case.

server/api/crud/workflows.py
```diff
@@ -177,6 +180,7 @@ def _prepare_run_object_for_scheduling(
                     project.metadata.name,
                     meta_uid,
                 ),
+                notifications=notifications,
```
theSaarco, line 179: Wow, the fact that you need to pass both `notification_dicts` and `notifications` in different parts of the same function call is amazingly terrible. Not only passing it twice, but in different formats...
I can understand why this is needed, but still... 🤯

server/api/crud/workflows.py
```diff
@@ -331,12 +336,14 @@ def _prepare_run_object_for_single_run(
...
                     # save=True modifies the project.yaml (by enrichment) so the local git repo is becoming dirty
                     dirty=save,
+                    notifications=notification_dicts,
                     # remote pipeline pod stays alive for the whole lifetime of the pipeline.
                     # once the pipeline is done, the pod finishes (either successfully or not) and notifications
                     # can be sent.
                     wait_for_completion=True,
                 ),
                 handler="mlrun.projects.load_and_run",
+                notifications=notifications,
```
theSaarco, line 337: And again.

theSaarco (changes requested):
Pretty good, but some changes are required - and there's a topic there that just looks terrible but I really am not sure I have a quick solution for.

mlrun/projects/pipelines.py
quaark: Yup missed it thanks

mlrun/projects/pipelines.py
quaark: Yup this is virtually always a dict. But I added notification to the type hint as someone can use this method themselves. I'll skip this step if they already passed the model object

server/api/crud/workflows.py
quaark: I was also annoyed I can't pass the notifications as is to the run. 
I only have another option. That's that the load_and_run method will call the api to get its own run to extract the notifications from it. Also quite nasty but maybe less?

docs/concepts/notifications.md
```diff
@@ -133,13 +133,9 @@ notification = mlrun.model.Notification(
 project.run(..., notifications=[notification])
 ```
 
-For pipelines that have configured notifications, MLRun always sends a `pipeline started` notification. Assuming you have a few notifications 
-configured &mdash; the notification that includes `when=running` determines the parameters for the `pipeline started` notification, for 
+MLRun can also sends a `pipeline started` notification. In order to do so, configure a notification that includes
```
jillnogold: MLRun can also send a `pipeline started` notification. To do that, configure a notification that includes 
`when=running`. The `pipeline started` notification uses its own parameters, for 
example the webhook, credentials, etc., for the notification message.

mlrun/projects/pipelines.py
```diff
@@ -1078,6 +1059,13 @@ def load_and_run(
     if load_only:
         return
 
+    # extract "start" notification if exists
+    start_notifications = [
+        mlrun.model.Notification.from_dict(notification)
+        for notification in context.to_dict().get("spec", {}).get("notifications", [])
```
theSaarco: Would be much better to add an accessor to the notifications in the context object (it already has them in `self._notifications`). Probably was never done since nobody needed it so far...

theSaarco (approved):
Looks much better (no more eyes bleeding 😆). I have one more suggestion, but otherwise approved.

TomerShor (approved):

PR #6206, opened 2024-08-26
[Run] Remove artifacts cache

alonmr:
The cached artifacts are memory demanding. In the backend we already use artifact URIs. The artifacts will still be available to the run but they will be lazy loaded. We can't completely deprecate it without migration the DB which IMO we should want to avoid until we have some stale resources cleanup mechanism because we don't want to migrate so many objects in mlrun upgrade.

https://iguazio.atlassian.net/browse/ML-7347


mlrun/artifacts/model.py
```diff
@@ -506,10 +504,22 @@ def _get_extra(target, extra_data, is_dir=False):
...
+
+    # Remove future packaging links
+    if model_dict["spec"].get("extra_data"):
+        model_dict["spec"]["extra_data"] = {
+            key: item
+            for key, item in model_dict["spec"]["extra_data"].items()
+            if item is not ...
+        }
+
+    return yaml.safe_dump(model_dict)
```
liranbg: this function performs 2 things
1. it sanitizes some information
2. serializes the object

I suggest to break the function to 2 and have the former ut coveraged.

mlrun/execution.py
```diff
@@ -573,22 +581,24 @@ def log_iteration_results(self, best, summary: list, task: dict, commit=False):
         """Reserved for internal use"""
 
         if best:
+            best_context = MLClientCtx.from_dict(
+                task, store_run=False, include_status=True
+            )
```
liranbg, line 587: why do you need this specifically? to get the best iteration context logged artifacts?

mlrun/model.py
```diff
@@ -1285,15 +1285,59 @@ def __init__(
...
+    def artifacts(self):
+        self._artifacts = self._artifacts or []
+        existing_artifact_keys = [
+            artifact["metadata"]["key"] for artifact in self._artifacts
+        ]
+        for key, uri in self.artifact_uris.items():
+            if key not in existing_artifact_keys:
+                artifact = mlrun.datastore.get_store_resource(uri)
+                self._artifacts.append(artifact.to_dict())
+        return self._artifacts
```
liranbg: docstring why it is lazy loaded, etc

tests/system/runtimes/test_kubejob.py
```diff
@@ -327,6 +327,30 @@ def test_function_with_kwargs(self):
         run = function.run(params=params, handler="func")
         assert run.outputs["return"] == kwargs
 
+    # TODO: Un-skip
+    @pytest.mark.skip(
+        "Waiting for extra data parsing for default packager `pack` method"
+    )
```
liranbg, line 333: remove?

mlrun/artifacts/base.py
```diff
@@ -765,6 +765,10 @@ def upload_extra_data(
         return
     target_path = artifact.target_path
     for key, item in extra_data.items():
+        if item is ...:
+            # Skip future links
```
liranbg: why would item be `...` / Elipsis?

liranbg (changes requested):

mlrun/execution.py
alonmr: yes, added a comment

mlrun/artifacts/base.py
alonmr: clarified comment

mlrun/model.py
```diff
@@ -1285,15 +1285,65 @@ def __init__(
...
+        """
+        Artifacts are lazy loaded to reduce memory consumption.
+        We keep artifact_uris (key -> store URI dictionary) to be able to get the run artifacts easily.
+        If the artifact is not already in the cache, we get it from the store (DB).
+        :return: List of artifact dictionaries
+        """
+        self._artifacts = self._artifacts or []
+        existing_artifact_keys = [
+            artifact["metadata"]["key"] for artifact in self._artifacts
+        ]
```
moranbental: It is faster to check for an element in a set than in a list
```suggestion
        existing_artifact_keys = {artifact["metadata"]["key"] for artifact in self._artifacts}
```

mlrun/model.py
```diff
@@ -1285,15 +1285,65 @@ def __init__(
...
+        for key, uri in self.artifact_uris.items():
+            if key not in existing_artifact_keys:
+                artifact = mlrun.datastore.get_store_resource(uri)
+                self._artifacts.append(artifact.to_dict())
```
moranbental, line 1323: `artifact` can be None

mlrun/artifacts/manager.py
```diff
@@ -156,14 +156,12 @@ def ensure_artifact_source_file_exists(item, path, body):
 
     def artifact_list(self, full=False):
         artifacts = []
-        for artifact in self.artifacts.values():
-            if isinstance(artifact, dict):
-                artifacts.append(artifact)
+        for artifacts_uri in self.artifact_uris.values():
+            artifact: Artifact = mlrun.datastore.get_store_resource(artifacts_uri)
```
moranbental, line 160: `artifact` can be None

mlrun/artifacts/model.py
```diff
@@ -609,3 +569,58 @@ def update_model(
             project=model_spec.project,
         )
     return model_spec
+
+
+def _get_src_path(model_spec: ModelArtifact, filename):
```
moranbental:
```suggestion
def _get_src_path(model_spec: ModelArtifact, filename: str) -> str:
```

mlrun/artifacts/model.py
```diff
@@ -609,3 +569,58 @@ def update_model(
...
+    if model_spec.src_path:
+        return path.join(model_spec.src_path, filename)
+    return filename
```
moranbental:
```suggestion
    return path.join(model_spec.src_path, filename) if model_spec.src_path else filename
```

mlrun/artifacts/model.py
```diff
@@ -609,3 +569,58 @@ def update_model(
...
+
+
+def _load_model_spec(spec_path):
```
moranbental:
```suggestion
def _load_model_spec(spec_path) -> ModelArtifact:
```

mlrun/artifacts/model.py
```diff
@@ -609,3 +569,58 @@ def update_model(
...
+    data = mlrun.datastore.store_manager.object(url=spec_path).get()
+    spec = yaml.load(data, Loader=yaml.FullLoader)
+    return ModelArtifact.from_dict(spec)
+
+
+def _get_file_path(base_path: str, name: str, isdir=False):
```
moranbental:
```suggestion
def _get_file_path(base_path: str, name: str, isdir: bool = False) -> str:
```

mlrun/artifacts/model.py
```diff
@@ -609,3 +569,58 @@ def update_model(
...
+    if not is_relative_path(name):
+        return name
+    if not isdir:
+        base_path = path.dirname(base_path)
+    return path.join(base_path, name).replace("\\", "/")
+
+
+def _get_extra(target, extra_data, is_dir=False):
```
moranbental:
```suggestion
def _get_extra(target: str, extra_data: dict, is_dir: bool = False) -> dict:
```

mlrun/artifacts/model.py
```diff
@@ -609,3 +569,58 @@ def update_model(
...
+    extra_dataitems = {}
+    for k, v in extra_data.items():
+        extra_dataitems[k] = mlrun.datastore.store_manager.object(
+            url=_get_file_path(target, v, isdir=is_dir), key=k
+        )
+    return extra_dataitems
+
+
+def _sanitize_and_serialize_model_spec_yaml(model: ModelArtifact):
```
moranbental:
```suggestion
def _sanitize_and_serialize_model_spec_yaml(model: ModelArtifact) -> str:
```

mlrun/artifacts/model.py
```diff
@@ -609,3 +569,58 @@ def update_model(
...
+    if model_dict["spec"].get("extra_data"):
+        model_dict["spec"]["extra_data"] = {
+            key: item
+            for key, item in model_dict["spec"]["extra_data"].items()
+            if item is not ...
+        }
+    return model_dict
+
+
+def _serialize_model_spec_yaml(model_dict):
```
moranbental:
```suggestion
def _serialize_model_spec_yaml(model_dict: dict) -> str:
```

mlrun/execution.py
```diff
@@ -194,6 +194,11 @@ def artifacts(self):
         """Dictionary of artifacts (read-only)"""
         return deepcopy(self._artifacts_manager.artifact_list())
 
+    @property
+    def artifact_uris(self):
+        """Dictionary of artifact URIs (read-only)"""
+        return deepcopy(self._artifacts_manager.artifact_uris)
```
moranbental, line 200: No need for `deepcopy` for a dictionary when keys and values are strings

mlrun/execution.py
```diff
@@ -573,22 +581,25 @@ def log_iteration_results(self, best, summary: list, task: dict, commit=False):
...
-                self._results[k] = v
-            for artifact in get_in(task, ["status", RunKeys.artifacts], []):
-                self._artifacts_manager.artifacts[artifact["metadata"]["key"]] = (
-                    artifact
-                )
+            for key, result in best_context.results.items():
+                self._results[key] = result
+            for key, artifact_uri in best_context.artifact_uris.items():
+                self._artifacts_manager.artifact_uris[key] = artifact_uri
+                artifact = best_context.get_artifact(key)
```
moranbental, line 593: `artifact` can be None

mlrun/execution.py
```diff
@@ -852,10 +863,18 @@ def log_model(
...
-        return self._artifacts_manager.artifacts[key]
+        warnings.warn(
+            "get_cached_artifact is deprecated in 1.8.0 and will be removed in 1.10.0. Use get_artifact instead.",
+            FutureWarning,
+        )
+        return self.get_artifact(key)
 
-    def update_artifact(self, artifact_object):
-        """Update an artifact object in the cache and the DB"""
+    def get_artifact(self, key):
```
moranbental:
```suggestion
    def get_artifact(self, key: str) -> Artifact:
```

mlrun/model.py
```diff
@@ -1655,7 +1705,12 @@ def _artifact(self, key):
         ]
 
         if not matching_artifacts:
-            return None
+            if key not in self.status.artifact_uris:
+                return None
+
+            # Get artifact by store URI sanity (should have been enriched by now in status.artifacts property)
+            artifact_uri = self.status.artifact_uris[key]
+            matching_artifacts = [mlrun.datastore.get_store_resource(artifact_uri)]
```
moranbental: `mlrun.datastore.get_store_resource(artifact_uri)` can return None, resulting in `matching_artifacts` being a list that contains a single None element.
In this case, you can return the artifact immediately without needing to sort or find non latest artifacts, since it only results in one artifact

mlrun/package/packagers_manager.py
```diff
@@ -715,17 +717,24 @@ def _unpack_data_item(self, data_item: DataItem, type_hint: type):
...
 
-        :param key:       Key to look for.
-        :param artifacts: Artifacts to look in.
-        :param results:   Results to look in.
+        :param key:             Key to look for.
+        :param artifacts:       Artifacts to look in.
+        :param artifact_uris:   Artifacts URIs to look in.
+        :param results:         Results to look in.
 
         :return: The artifact or result with the same key or None if not found.
```
moranbental, line 732:
```suggestion
        :return: The artifact or artifact uri or result with the same key or None if not found.
```

mlrun/package/packagers_manager.py
```diff
@@ -715,17 +717,24 @@ def _unpack_data_item(self, data_item: DataItem, type_hint: type):
...
         """
+        artifact_uris = artifact_uris or {}
+        for _key, uri in artifact_uris.items():
+            if key == _key:
+                return uri
```
moranbental: Why not retrieve the artifact itself with `get_store_resource`? now extra_data holds the uri instead of the artifact, and I didn’t see you handle this case in `link_packages`

moranbental (changes requested):

mlrun/model.py
alonmr: `get_store_resource` will raise 404

mlrun/artifacts/manager.py
alonmr: same

mlrun/execution.py
alonmr: why not? the dictionary can be altered

mlrun/execution.py
alonmr: same it will raise 404

mlrun/package/packagers_manager.py
alonmr: Good catch 👍

mlrun/execution.py
liranbg: I think she means you can use regular copy without deep but I think the intention is to provide read-only structure which means it is safer to use deepcopy if in the future the structure of artifact uris would be different

mlrun/package/packagers_manager.py
```diff
@@ -715,17 +718,31 @@ def _unpack_data_item(self, data_item: DataItem, type_hint: type):
...
-        :param key:       Key to look for.
-        :param artifacts: Artifacts to look in.
-        :param results:   Results to look in.
+        :param key:             Key to look for.
+        :param artifacts:       Artifacts to look in.
+        :param artifact_uris:   Artifacts URIs to look in.
+        :param results:         Results to look in.
 
-        :return: The artifact or result with the same key or None if not found.
+        :return: The artifact or artifact uri or result with the same key or None if not found.
```
moranbental: Now this comment is incorrect, you are returning the artifact here, not the URI.

moranbental (approved):
Just one comment. Besides that, it looks very good!

liranbg (approved):

PR #6209, opened 2024-08-26
[Model Monitoring] Remove `current_stats` from tsdb

Eyal-Danieli:
Right now there are no consumers for the `current_stats` value within the TSDB. In addition, this values is not limited and therefore can cause storage issues within the TSDB when using a large model with many features.
In the near future, the user will be able to store the `current_stats` as part of the `result_extra_data` or generate a `current_stats` artifact for each run (if the value is too large). 

A fix for https://iguazio.atlassian.net/browse/ML-7519

### DIFF
--- a/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
+++ b/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
@@ -108,6 +108,7 @@ class TDEngineConnector(TSDBConnector):
             table_name = (
                 f"{table_name}_" f"{event[mm_schemas.ResultData.RESULT_NAME]}"
             ).replace("-", "_")
+            del event[mm_schemas.ResultData.CURRENT_STATS]
 
         else:
             # Write a new metric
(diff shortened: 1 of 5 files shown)


mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -129,6 +138,10 @@ def write_application_event(
         insert_statement.add_batch()
         insert_statement.execute()
 
+    @staticmethod
+    def _convert_to_datetime(key):
+        return datetime.fromisoformat(key) if isinstance(key, str) else key
```
gtopper: Not sure why it would be called `key`.
```suggestion
    def _convert_to_datetime(val):
        return datetime.fromisoformat(val) if isinstance(val, str) else val
```

jond01: And consider adding type hints :)

jond01 (approved):

gtopper (approved):

PR #6211, opened 2024-08-27
[Application] Fix setting of default-container annotation

rokatyy:
Jira - https://iguazio.atlassian.net/browse/NUC-236


alonmr (commented):
I couldn't find where this is converted to nuclio spec annotations. You can use `with_annotations` instead I assume that one works. Also please add some assertion for this in unit tests

rokatyy:
> I couldn't find where this is converted to nuclio spec annotations

@alonmr what do you mean? Annotations are kept in metadata in nuclio config. Reg `with_annotations` - Not sure if it's good idea as it cleans up the annotations, thus if user did set custom annotations, they will be removed..

TomerShor (approved):

alonmr (approved):

PR #6214, opened 2024-08-27
[Pipelines] Fix Jupyter pipeline run output not displaying stdout

quaark:
https://iguazio.atlassian.net/browse/ML-7613

Previously, the graph view of the pipeline would override the cell's entire display and therefore remove the stdout:
Screenshot 2024-08-21 at 19 03 47

Now, we make sure to create multiple IPython displays dedicated to their own content (default one for stdout, one for the http link to the pipeline web page and one for the graph visualization). Then we use `update_display` on the specific display instead of running `clear_output` which cleans up *all* displays of the cell:
Screenshot 2024-08-27 at 14 06 06


mlrun/utils/helpers.py
```diff
@@ -1410,6 +1411,26 @@ def is_running_in_jupyter_notebook() -> bool:
...
+
+        display_id = uuid.uuid4().hex
+        content = IPython.display.HTML(
+            f'<div id="{display_id}">Temporary Display Content</div>'
+        )
+        IPython.display.display(content, display_id=display_id)
+        return display_id
+
+    # returning None if IPython is not installed
+    logger.warn("IPython is not installed, cannot create IPython display")
```
alonmr: I think this log is not relevant if `is_ipython` is `False` then we are not using Jupyter and this logs an unrelated warning

alonmr (commented):
Looks great! Minor comment. Also please don't bump pipeline packages in a follow up - I would like to test the dependabot written for it but you will need to release of course

alonmr (approved):

quaark:
It seems our Backwards Compatibility Checks aren't compatible with changes in `pipeline-adapters` as only one version can be installed in the site-packages, and the Check assumes it can compare the 2 (old and new) version.
We'll merge here either way and we need to find a solution for this in a follow up.

PR #6215, opened 2024-08-27
[Alerts] Fix alert triggering for job events by adjusting the sliding window to account for monitoring interval

Yacouby:
This PR addresses a bug in the alerts flow where alerts configured to trigger based on a specific number of events within a defined period might not trigger correctly when monitoring failed jobs. 
The issue occurred because the monitoring runs flow executes every 30 seconds, which could result in missing the last event in a sequence if it occurs just before the monitoring interval.
so when the monitoring interval is longer than the time gap between the last event and the end of the defined period, the monitoring might miss triggering the alert.

resolves:
https://iguazio.atlassian.net/browse/ML-7549

### DIFF
--- a/server/api/utils/helpers.py
+++ b/server/api/utils/helpers.py
@@ -119,7 +119,9 @@ def is_request_from_leader(
     return False
 
 
-def string_to_timedelta(date_str, offset=0, raise_on_error=True):
+def string_to_timedelta(
+    date_str: str, offset: int = 0, raise_on_error: bool = True
+) -> Optional[datetime.timedelta]:
     date_str = date_str.strip().lower()
     try:
         seconds = parse_timespan(date_str) + offset
--- a/server/api/utils/helpers.py
+++ b/server/api/utils/helpers.py
@@ -119,10 +119,10 @@ def is_request_from_leader(
     return False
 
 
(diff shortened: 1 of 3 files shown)


server/api/utils/helpers.py
```diff
@@ -119,10 +119,10 @@ def is_request_from_leader(
     return False
 
 
-def string_to_timedelta(date_str, raise_on_error=True):
+def string_to_timedelta(date_str, interval=0, raise_on_error=True):
```
quaark:
```suggestion
def string_to_timedelta(date_str, offset=0, raise_on_error=True):
```
a more indicative name IMO

quaark (changes requested):
Looks good!
I'd only change the name from interval to offset

server/api/utils/helpers.py
```diff
@@ -119,10 +119,10 @@ def is_request_from_leader(
     return False
 
 
-def string_to_timedelta(date_str, raise_on_error=True):
+def string_to_timedelta(date_str, offset=0, raise_on_error=True):
```
TomerShor: Add type hints

quaark (approved):

PR #6218, opened 2024-08-27
[API] Add error message on monitored runs

alonmr:
Tries to fetch the error message from the k8s resource when updating run to an error state and update the run accordingly.

https://iguazio.atlassian.net/browse/ML-7232

### DIFF
--- a/server/api/runtime_handlers/base.py
+++ b/server/api/runtime_handlers/base.py
@@ -1698,8 +1698,8 @@ class BaseRuntimeHandler(ABC):
         run_updates = {
             "status.state": run_state,
             "status.last_update": now_date().isoformat(),
-            "status.reason": reason,
-            "status.status_text": message,
+            "status.reason": reason or "",
+            "status.status_text": message or "",
             "status.error": "",
         }
         run = db.update_run(db_session, run_updates, uid, project)
--- a/server/api/runtime_handlers/base.py
+++ b/server/api/runtime_handlers/base.py
@@ -771,6 +771,15 @@ class BaseRuntimeHandler(ABC):
 
         return in_terminal_state, last_container_completion_time, run_state
 
+    def _resolve_container_error_status(self, pod: dict) -> tuple[str, str]:
+        container_statuses = pod["status"].get("container_statuses", [])
+        for container_status in container_statuses:
+            terminated = container_status.get("state", {}).get("terminated")
+            if terminated:
+                return terminated.get("reason", ""), terminated.get("message", "")
+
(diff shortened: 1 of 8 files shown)


server/api/runtime_handlers/base.py
```diff
@@ -771,6 +771,15 @@ def _resolve_pod_status_info(
 
         return in_terminal_state, last_container_completion_time, run_state
 
+    def _resolve_container_error_status(self, pod: dict) -> tuple[str, str]:
+        container_statuses = pod["status"].get("container_statuses", [])
```
moranbental:
```suggestion
        container_statuses = pod.get("status", {}).get("container_statuses", [])
```

server/api/runtime_handlers/mpijob/v1.py
```diff
@@ -270,6 +270,15 @@ def _resolve_crd_object_status_info(
             )
         return in_terminal_state, completion_time, desired_run_state
 
+    def _resolve_container_error_status(self, crd_object: dict) -> tuple[str, str]:
```
moranbental, line 273: It seems like this function is duplicated from the one you added in `server/api/runtime_handlers/base.py`. It could be moved to a helper module or something similar?

moranbental (changes requested):
Looks good! Just a few comments below

server/api/runtime_handlers/mpijob/v1.py
alonmr: they may look similar but they are not the same. the runtime resource structure is different

rokatyy (approved):
LGTM

moranbental (approved):
LGTM

server/api/runtime_handlers/base.py
```diff
@@ -1671,12 +1690,16 @@ def _ensure_run_state(
...
+                reason, message = self._resolve_container_error_status(runtime_resource)
+
         logger.info("Updating run state", run_uid=uid, run_state=run_state)
         run_updates = {
             "status.state": run_state,
             "status.last_update": now_date().isoformat(),
-            # run is not in terminal state, so reset reason and error
-            "status.reason": "",
+            "status.reason": reason or "",
+            "status.status_text": message or "",
```
TomerShor, line 1702: you don't need the `or ""`, they're already initialized as `""` and the resolve method can also return `""` and not None

TomerShor (changes requested):
Minor

TomerShor (approved):

PR #6221, opened 2024-08-28
[Application] Validate authentication params

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-7539

### DIFF
--- a/mlrun/runtimes/nuclio/application/application.py
+++ b/mlrun/runtimes/nuclio/application/application.py
@@ -327,6 +327,14 @@ class ApplicationRuntime(RemoteRuntime):
         self._ensure_reverse_proxy_configurations(self)
         self._configure_application_sidecar()
 
+        if (
+            authentication_mode == schemas.APIGatewayAuthenticationMode.basic
+            and not authentication_creds
+        ):
+            raise mlrun.errors.MLRunBadRequestError(
+                "Authentication credentials not provided"
+            )
+
         # We only allow accessing the application via the API Gateway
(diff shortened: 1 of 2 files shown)


mlrun/runtimes/nuclio/application/application.py
```diff
@@ -327,6 +327,14 @@ def deploy(
         self._ensure_reverse_proxy_configurations(self)
         self._configure_application_sidecar()
 
+        if (
```
alonmr, line 330: Validations are generally preformed in the server side. Also this validation is misplaced. You are validating the API gateway in the application runtime so there is mixup in responsibility. The validation should be in the server create api gateway endpoint. I understand the need to fail fast but if deploy was successful and application gateway creation failed - users can create the api gateway without deploying the function again so it is not a big blow.

alonmr (changes requested):
One more note about validation on server side - the reason for that is making the server the vocal point for validations and enrichment (there are some that must be on the client). Think about client having validation X in mlrun 1.7 and in mlrun 1.8 we changed the validation to Y. Now the client must upgrade to pass validation. This does not apply here because the validation is unlikely to change but it s a general best practice for validations when we can.

mlrun/runtimes/nuclio/application/application.py
rokatyy: @alonmr in api gateway you can't not to pass creds:
image

empty-string credentials are still credentials from Nuclio point of view.

PR #6222, opened 2024-08-28
[Model Monitoring] Some improvements in tests debugging

assaf758:
1. In case env-vars are missing for a test, print one of them.
2. fix mm test infra to allow logging from _V3IORecordsChecker custom_setup()

### DIFF
--- a/tests/system/base.py
+++ b/tests/system/base.py
@@ -183,7 +183,7 @@ class TestMLRunSystem:
         return pytest.mark.skipif(
             not configured,
             reason=f"This is a system test, add the needed environment variables {*mandatory_env_vars,} "
-            "in tests/system/env.yml to run it",
+            f"in tests/system/env.yml to run it. Currently at least {env_var} is missing",
         )(test)
 
     @classmethod
--- a/tests/system/base.py
+++ b/tests/system/base.py
@@ -170,20 +170,21 @@ class TestMLRunSystem:
         )
         if cls._has_marker(test, cls.model_monitoring_marker_name):
             mandatory_env_vars += cls.model_monitoring_mandatory_env_vars
-        configured = True
+
+        missing_env_vars = []
         try:
             env = cls._get_env_from_file()
         except FileNotFoundError:
-            configured = False
+            missing_env_vars = mandatory_env_vars
         else:
             for env_var in mandatory_env_vars:
                 if env_var not in env or env[env_var] is None:
-                    configured = False
+                    missing_env_vars.append(env_var)
 
         return pytest.mark.skipif(
-            not configured,
+            len(missing_env_vars) > 0,
             reason=f"This is a system test, add the needed environment variables {*mandatory_env_vars,} "
-            f"in tests/system/env.yml to run it. Currently at least {env_var} is missing",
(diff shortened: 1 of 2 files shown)


tests/system/model_monitoring/test_app.py
```diff
@@ -431,7 +431,8 @@ def custom_setup_class(cls) -> None:
         cls.run_db = mlrun.get_run_db()
 
     def custom_setup(self) -> None:
-        _V3IORecordsChecker.custom_setup(project_name=self.project_name)
+        # skips TestMLRunSystem, calls custom_setup() of _V3IORecordsChecker
+        super(TestMLRunSystem, self).custom_setup(project_name=self.project_name)
```
gtopper, line 443: Or we can refactor this to _not_ use multiple inheritance.

tests/system/base.py
```diff
@@ -183,7 +183,7 @@ def skip_test_if_env_not_configured(cls, test):
         return pytest.mark.skipif(
             not configured,
             reason=f"This is a system test, add the needed environment variables {*mandatory_env_vars,} "
-            "in tests/system/env.yml to run it",
+            f"in tests/system/env.yml to run it. Currently at least {env_var} is missing",
```
gtopper: It only takes 2 extra lines to report all missing env vars. Something like:
```python
        missing_env_vars = []
        try:
            env = cls._get_env_from_file()
        except FileNotFoundError:
            configured = False
        else:
            for env_var in mandatory_env_vars:
                if env_var not in env or env[env_var] is None:
                    missing_env_vars.append(env_var)
                    configured = False

        return pytest.mark.skipif(
            not configured,
            reason=f"This is a system test, add the needed environment variables {*mandatory_env_vars,} "
            f"in tests/system/env.yml to run it. These environment variables are missing: "
            + ", ".join(missing_env_vars),
```

tests/system/model_monitoring/test_app.py
assaf758: I'm not sure either, but I guess there are pros and cons to this. Can be discussed in one of our meetings f2f.

gtopper (approved):

tests/system/model_monitoring/test_app.py
gtopper: I'm pretty sure the comment isn't correct btw. Why do you think it skips `TestMLRunSystem`? `TestMLRunSystem` is the one that defines the logger.

assaf758: Now entirely sure - but still few references:

1. See python [docs](https://docs.python.org/3/library/functions.html#super) for `super`:
>The object_or_type determines the [method resolution order](https://docs.python.org/3/glossary.html#term-method-resolution-order) to be searched. The search starts from the class right after the type.

2. From another view, looking at `TestMLRunSystem` custom_setup() you can see it has a different signature (without project_name) so it would have raised an exception if called from here.

3. As I understand it, the reason the change is working is that it calls custom_setup() on the TestMonitoringAppFlow class, which has it's _logger class attribute set from the TestMLRunSystem inherited class. On the hand, the original code called _V3IORecordsCheck custom_setup() - and on that class the _logger attribute in not initialized.

PR #6225, opened 2024-08-28
[Projects] Change the background task timeout for project deletion

moranbental:
When creating the background task for project deletion, the timeout should be configured using the default value `mlrun.mlconf.background_tasks.default_timeouts.operations.delete_project` from `config.py`, which is set to 15 minutes. However, the background task was hardcoded to wait only 6 minutes. 

The fix is to use the default value from the config to ensure consistency
https://iguazio.atlassian.net/browse/ML-7686
https://iguazio.atlassian.net/browse/ML-6595

### DIFF
--- a/server/api/utils/clients/iguazio.py
+++ b/server/api/utils/clients/iguazio.py
@@ -618,7 +618,7 @@ class Client(
 
         job_state, job_result = mlrun.utils.helpers.retry_until_successful(
             self._wait_for_job_completion_retry_interval,
-            timeout,
+            int(timeout),
             self._logger,
             False,
             _verify_job_in_terminal_state,
--- a/server/api/utils/clients/iguazio.py
+++ b/server/api/utils/clients/iguazio.py
@@ -293,7 +293,7 @@ class Client(
             self._wait_for_job_completion(
                 session,
                 job_id,
-                mlrun.mlconf.background_tasks.default_timeouts.operations.delete_project,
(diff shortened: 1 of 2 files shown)


quaark (approved):
👑

server/api/utils/clients/iguazio.py
```diff
@@ -610,7 +618,7 @@ def _verify_job_in_terminal_state():
 
         job_state, job_result = mlrun.utils.helpers.retry_until_successful(
             self._wait_for_job_completion_retry_interval,
-            360,
+            timeout,
```
alonmr: config timeout is a string
```suggestion
            int(timeout),
```

alonmr (commented):
Looks great! minor comment

alonmr (approved):

server/api/utils/clients/iguazio.py
```diff
@@ -290,7 +290,11 @@ def delete_project(
                 name=name,
                 job_id=job_id,
             )
-            self._wait_for_job_completion(session, job_id)
+            self._wait_for_job_completion(
+                session,
+                job_id,
+                mlrun.mlconf.background_tasks.default_timeouts.operations.delete_project,
```
alonmr: Actually
```suggestion
                timeout=mlrun.mlconf.background_tasks.default_timeouts.operations.delete_project,
```

alonmr (changes requested):
I think you passed the timeout as error message

alonmr (approved):

liranbg (approved):

PR #6227, opened 2024-08-28
[Application] Separate Application deploy and API gateway creation

alonmr:
Design fix for Application runtime - extracting the API gateway creation from the deploy process is needed for the following reasons:
- Clear separation between api gateway and application runtime entities for developers and for users
- Easier development process
- Shorter deploy process (less things can go wrong)
- Robustness to changes
- Option to redeploy without creating api gateway again

In mlrun they are separate entities but an application is not accessible until it has an API gateway so at 1st we wanted to make life easier for the user and give them everything they need with a single deploy call. However this makes life difficult for developers since it introduces cupeling and mixing responsibilities. More than that it will be useful for the users to know that they can change the API gateway without being dependent on the deploy process and vice versa.

Still to make life simple, we provide a default API gateway with opt-out option.

https://iguazio.atlassian.net/browse/ML-7698

### DIFF
--- a/mlrun/runtimes/nuclio/application/application.py
+++ b/mlrun/runtimes/nuclio/application/application.py
@@ -428,7 +428,7 @@ class ApplicationRuntime(RemoteRuntime):
         path: str = None,
         direct_port_access: bool = False,
         authentication_mode: schemas.APIGatewayAuthenticationMode = None,
-        authentication_creds: tuple[str] = None,
+        authentication_creds: tuple[str, str] = None,
         ssl_redirect: bool = None,
         set_as_default: bool = False,
     ):
@@ -499,6 +499,8 @@ class ApplicationRuntime(RemoteRuntime):
             api_gateway = APIGateway.from_scheme(api_gateway_scheme)
             api_gateway.wait_for_readiness()
             url = api_gateway.invoke_url
+            # Update application status
+            self._get_state(raise_on_exception=False)
 
         logger.info("Successfully created API gateway", url=url)
         return url
--- a/mlrun/runtimes/nuclio/application/application.py
+++ b/mlrun/runtimes/nuclio/application/application.py
@@ -281,10 +281,6 @@ class ApplicationRuntime(RemoteRuntime):
         is_kfp=False,
         mlrun_version_specifier=None,
         show_on_failure: bool = False,
-        direct_port_access: bool = False,
-        authentication_mode: schemas.APIGatewayAuthenticationMode = None,
-        authentication_creds: tuple[str] = None,
-        ssl_redirect: bool = None,
     ):
         """
         Deploy function, builds the application image if required (self.requires_build()) or force_build is True,
@@ -302,11 +298,6 @@ class ApplicationRuntime(RemoteRuntime):
         :param is_kfp:                  Deploy as part of a kfp pipeline
         :param mlrun_version_specifier: Which mlrun package version to include (if not current)
         :param show_on_failure:         Show logs only in case of build failure
-        :param direct_port_access:      Set True to allow direct port access to the application sidecar
-        :param authentication_mode:     API Gateway authentication mode
-        :param authentication_creds:    API Gateway authentication credentials as a tuple (username, password)
-        :param ssl_redirect:            Set True to force SSL redirect, False to disable. Defaults to
-                                        mlrun.mlconf.force_api_gateway_ssl_redirect()
 
         :return: True if the function is ready (deployed)
         """
@@ -328,10 +319,6 @@ class ApplicationRuntime(RemoteRuntime):
         self._configure_application_sidecar()
 
         # We only allow accessing the application via the API Gateway
-        name_tag = tag or self.metadata.tag
-        self.status.api_gateway_name = (
-            f"{self.metadata.name}-{name_tag}" if name_tag else self.metadata.name
-        )
         self.spec.add_templated_ingress_host_mode = (
             NuclioIngressAddTemplatedIngressModes.never
         )
@@ -344,9 +331,7 @@ class ApplicationRuntime(RemoteRuntime):
             builder_env=builder_env,
         )
         logger.info(
-            "Successfully deployed function, creating API gateway",
-            api_gateway_name=self.status.api_gateway_name,
-            authentication_mode=authentication_mode,
+            "Successfully deployed function",
         )
 
         # Restore the source in case it was removed to make nuclio not consider it when building
@@ -354,14 +339,10 @@ class ApplicationRuntime(RemoteRuntime):
             self.spec.build.source = self.status.application_source
         self.save(versioned=False)
 
-        ports = self.spec.internal_application_port if direct_port_access else []
-        return self.create_api_gateway(
-            name=self.status.api_gateway_name,
-            ports=ports,
-            authentication_mode=authentication_mode,
-            authentication_creds=authentication_creds,
-            ssl_redirect=ssl_redirect,
+        logger.info(
+            "Application online, proceed to API gateway creation to make it accessible."
         )
+        return True
 
     def with_source_archive(
         self,
@@ -429,14 +410,34 @@ class ApplicationRuntime(RemoteRuntime):
         self,
         name: str = None,
         path: str = None,
-        ports: list[int] = None,
+        direct_port_access: bool = False,
         authentication_mode: schemas.APIGatewayAuthenticationMode = None,
         authentication_creds: tuple[str] = None,
         ssl_redirect: bool = None,
     ):
+        """
+        Create the application API gateway. Once the application is deployed, the API gateway can be created.
+        An application without an API gateway is not accessible.
+        :param name:                    The name of the API gateway, defaults to <function-name>-<function-tag>
+        :param path:                    Optional path of the API gateway, default value is "/"
+        :param direct_port_access:      Set True to allow direct port access to the application sidecar
+        :param authentication_mode:     API Gateway authentication mode
+        :param authentication_creds:    API Gateway authentication credentials as a tuple (username, password)
+        :param ssl_redirect:            Set True to force SSL redirect, False to disable. Defaults to
+                                        mlrun.mlconf.force_api_gateway_ssl_redirect()
+
+        :return:    The API gateway URL
+        """
+        self.status.api_gateway_name = name or (
+            f"{self.metadata.name}-{self.metadata.tag}"
+            if self.metadata.tag
+            else self.metadata.name
+        )
+        ports = self.spec.internal_application_port if direct_port_access else []
+
         api_gateway = APIGateway(
             APIGatewayMetadata(
-                name=name,
+                name=self.status.api_gateway_name,
                 namespace=self.metadata.namespace,
                 labels=self.metadata.labels,
                 annotations=self.metadata.annotations,
@@ -452,10 +453,10 @@ class ApplicationRuntime(RemoteRuntime):
         if ssl_redirect is None:
             ssl_redirect = mlrun.mlconf.force_api_gateway_ssl_redirect()
         if ssl_redirect:
-            # force ssl redirect so that the application is only accessible via https
+            # Force ssl redirect so that the application is only accessible via https
             api_gateway.with_force_ssl_redirect()
 
-        # add authentication if required
+        # Add authentication if required
         authentication_mode = (
             authentication_mode
             or mlrun.mlconf.function.application.default_authentication_mode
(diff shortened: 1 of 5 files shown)


mlrun/runtimes/nuclio/application/application.py
```diff
@@ -344,24 +331,18 @@ def deploy(
...
 
-        ports = self.spec.internal_application_port if direct_port_access else []
-        return self.create_api_gateway(
-            name=self.status.api_gateway_name,
-            ports=ports,
-            authentication_mode=authentication_mode,
-            authentication_creds=authentication_creds,
-            ssl_redirect=ssl_redirect,
+        logger.info(
+            "Application online, proceed to API gateway creation to make it accessible."
```
TomerShor: this might need to be a warning to "pop out" to the user.
Something like
```suggestion
        logger.warn(
            "Application is online but is not accessible. Use the `create_api_gateway` method to make it accessible."
```

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -281,34 +281,29 @@ def deploy(
...
+        :param auth_info:                   Service AuthInfo (deprecated and ignored)
+        :param builder_env:                 Env vars dict for source archive config/credentials
+                                            e.g. builder_env={"GIT_TOKEN": token}
+        :param force_build:                 Set True for force building the application image
+        :param with_mlrun:                  Add the current mlrun package to the container build
+        :param skip_deployed:               Skip the build if we already have an image for the function
+        :param is_kfp:                      Deploy as part of a kfp pipeline
+        :param mlrun_version_specifier:     Which mlrun package version to include (if not current)
+        :param show_on_failure:             Show logs only in case of build failure
+        :param create_default_api_gateway:  When deploy finishes will create the default API gateway for the
```
TomerShor:
```suggestion
        :param create_default_api_gateway:  When deploy finishes the default API gateway will be created for the
```

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -344,24 +335,30 @@ def deploy(
             builder_env=builder_env,
         )
         logger.info(
-            "Successfully deployed function, creating API gateway",
-            api_gateway_name=self.status.api_gateway_name,
-            authentication_mode=authentication_mode,
+            "Successfully deployed function.",
```
TomerShor, line 338: 2 minor notes about log consistency that I noticed:
1. It's not very consistent across MLRun but we usually put the "successfully" at the end (e.g. `"Pipeline submitted successfully"`)
2. Either always end with a period or don't.

Just something to think about in general

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -344,24 +335,30 @@ def deploy(
...
-            ssl_redirect=ssl_redirect,
-        )
+        if create_default_api_gateway:
+            try:
+                api_gateway_name = self._resolve_default_api_gateway_name()
+                return self.create_api_gateway(api_gateway_name, set_as_default=True)
+            except Exception as exc:
+                logger.warning(
+                    "Failed to create default API gateway, application will not be accessible",
+                    mlrun.errors.err_to_str(exc),
```
TomerShor:
```suggestion
                    exc_info=mlrun.errors.err_to_str(exc),
```

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -344,24 +335,30 @@ def deploy(
...
-            authentication_creds=authentication_creds,
-            ssl_redirect=ssl_redirect,
-        )
+        if create_default_api_gateway:
+            try:
+                api_gateway_name = self._resolve_default_api_gateway_name()
+                return self.create_api_gateway(api_gateway_name, set_as_default=True)
+            except Exception as exc:
+                logger.warning(
+                    "Failed to create default API gateway, application will not be accessible",
```
TomerShor: Maybe worth adding the same about the need to create api gateway to make it accessible here too

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -429,11 +426,33 @@ def create_api_gateway(
...
+        set_as_default: bool = False,
     ):
+        """
+        Create the application API gateway. Once the application is deployed, the API gateway can be created.
+        An application without an API gateway is not accessible.
+        :param name:                    The name of the API gateway, defaults to <function-name>-<function-tag>
+        :param path:                    Optional path of the API gateway, default value is "/"
+        :param direct_port_access:      Set True to allow direct port access to the application sidecar
+        :param authentication_mode:     API Gateway authentication mode
+        :param authentication_creds:    API Gateway authentication credentials as a tuple (username, password)
```
TomerShor:
```suggestion
        :param authentication_creds:    API Gateway basic authentication credentials as a tuple (username, password)
```

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -429,11 +426,33 @@ def create_api_gateway(
         self,
         name: str = None,
```
TomerShor, line 428: `name` is not enriched, can it still be `None`?

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -429,11 +426,33 @@ def create_api_gateway(
...
-        authentication_creds: tuple[str] = None,
+        authentication_creds: tuple[str, str] = None,
         ssl_redirect: bool = None,
+        set_as_default: bool = False,
     ):
+        """
+        Create the application API gateway. Once the application is deployed, the API gateway can be created.
+        An application without an API gateway is not accessible.
+        :param name:                    The name of the API gateway, defaults to <function-name>-<function-tag>
+        :param path:                    Optional path of the API gateway, default value is "/"
```
TomerShor, line 440: where is this default set? in the BE / Nuclio?

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -469,14 +488,22 @@ def create_api_gateway(
...
+            self.status.api_gateway = APIGateway.from_scheme(api_gateway_scheme)
+            self.status.api_gateway.wait_for_readiness()
+            self.url = self.status.api_gateway.invoke_url
+            url = self.url
+        else:
+            api_gateway = APIGateway.from_scheme(api_gateway_scheme)
+            api_gateway.wait_for_readiness()
+            url = api_gateway.invoke_url
+            # Update application status
+            self._get_state(raise_on_exception=False)
```
TomerShor: why is this needed here?

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -493,7 +520,10 @@ def invoke(
         self._sync_api_gateway()
         # If the API Gateway is not ready or not set, try to invoke the function directly (without the API Gateway)
         if not self.status.api_gateway:
-            super().invoke(
+            logger.warning(
+                "Default API gateway is not configured, will use invocation URL."
```
TomerShor:
```suggestion
                "Default API gateway is not configured, invoking function invocation URL."
```

TomerShor (changes requested):
Awesome stuff, have some comments and suggestions

mlrun/runtimes/nuclio/application/application.py
alonmr: Period at the end is subjected to having log kwargs ot not because the a `:` is appended. This should probably move to the logger (to add "." when needed)

mlrun/runtimes/nuclio/application/application.py
alonmr: it will fail in `APIGatewayMetadata` constructor

mlrun/runtimes/nuclio/application/application.py
alonmr: Seems like it's in the API gateway schema

mlrun/runtimes/nuclio/application/application.py
alonmr: to get the external invocation. it is enriched only in the server but the client doesn't get it

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -469,14 +489,22 @@ def create_api_gateway(
...
-        self.url = self.status.api_gateway.invoke_url
 
-        logger.info("Successfully created API gateway", url=self.url)
-        return self.url
+        if set_as_default:
+            self.status.api_gateway_name = api_gateway_scheme.metadata.name
+            self.status.api_gateway = APIGateway.from_scheme(api_gateway_scheme)
+            self.status.api_gateway.wait_for_readiness()
+            self.url = self.status.api_gateway.invoke_url
+            url = self.url
```
quaark, line 503: Maybe also delete the old default if one exists here?

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -665,3 +696,12 @@ def _sync_api_gateway(self):
...
         self.status.api_gateway.wait_for_readiness()
         self.url = self.status.api_gateway.invoke_url
+
+    def _resolve_default_api_gateway_name(self):
+        api_gateway_name = (
+            f"{self.metadata.name}-{self.metadata.tag}"
+            if self.metadata.tag
+            else self.metadata.name
+        )
+        api_gateway_name += "-default"
```
quaark: are we sure we want this `-default` suffix on every default app runtime gw?
I feel the default should reflect the common use case, and the common use case would be just the name of the function with the tag.
Since we have the validation in the `create_api_gateway` method, there shouldn't be overlap issues, so I don't think this is needed.

quaark (changes requested):
Looks great!
Couple minor suggestions

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -344,24 +335,31 @@ def deploy(
...
+                return self.create_api_gateway(api_gateway_name, set_as_default=True)
+            except Exception as exc:
+                logger.warning(
+                    "Failed to create default API gateway, application will not be accessible. "
+                    "Use the `create_api_gateway` method to make it accessible",
+                    exc=mlrun.errors.err_to_str(exc),
+                )
+        else:
+            logger.warning(
+                "Application is online but is not accessible. "
```
theSaarco:
```suggestion
                "Application is online but may not be accessible since default gateway creation was not requested."
```
It's still a user's choice to get here - so make sure it's understood this is happening due to the user's choice, not due to an error.
I'm also thinking to use a "may not" here since it's possible this is a re-deploy and manual GWs already exist that point to that function.

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -429,11 +427,33 @@ def create_api_gateway(
...
+        :param authentication_creds:    API Gateway basic authentication credentials as a tuple (username, password)
+        :param ssl_redirect:            Set True to force SSL redirect, False to disable. Defaults to
+                                        mlrun.mlconf.force_api_gateway_ssl_redirect()
+        :param set_as_default:          Set the API gateway as the default for the application (`status.api_gateway`)
+
+        :return:    The API gateway URL
+        """
+        if not set_as_default and name == self._resolve_default_api_gateway_name():
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"API gateway {name=} conflicts with default while {set_as_default=}."
```
theSaarco:
```suggestion
                f"Non-default API gateway cannot use the default gateway name, {name=}."
```

mlrun/runtimes/nuclio/application/application.py
theSaarco: I believe in our discussion this was mentioned to happen automatically since you're using the same name for the GW, so nuclio will replace.

theSaarco (commented):
Looks good, I only have some logging suggestions.

mlrun/runtimes/nuclio/application/application.py
quaark: Not always anymore, as this code allows you to create a GW with a different name and set it as the default, overriding anything that was there before

mlrun/runtimes/nuclio/application/application.py
TomerShor: IMO better to have it as a required arg here instead of failing later

mlrun/runtimes/nuclio/application/application.py
TomerShor: we can also add a check if `self.status.api_gateway` is None or not

TomerShor (approved):

PR #6228, opened 2024-08-28
[Application] Don't allow disabling default http trigger

TomerShor:
Since application runtime is only accessible via API gateway, the function must have an http trigger to handle the API gateway's requests.

Thus, we wan't to forbid disabling http trigger for application runtime.

### DIFF
--- a/mlrun/runtimes/nuclio/function.py
+++ b/mlrun/runtimes/nuclio/function.py
@@ -698,7 +698,7 @@ class RemoteRuntime(KubeResource):
         """
         self.spec.disable_default_http_trigger = True
 
-    @min_nuclio_versions("1.12.8")
+    @min_nuclio_versions("1.13.1")
     def enable_default_http_trigger(
         self,
     ):
--- a/mlrun/runtimes/nuclio/application/application.py
+++ b/mlrun/runtimes/nuclio/application/application.py
@@ -552,6 +552,20 @@ class ApplicationRuntime(RemoteRuntime):
             reverse_proxy_func.metadata.name, reverse_proxy_func.metadata.project
         )
 
+    @min_nuclio_versions("1.13.1")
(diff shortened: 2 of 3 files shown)


mlrun/runtimes/nuclio/function.py
```diff
@@ -698,7 +698,7 @@ def disable_default_http_trigger(
         """
         self.spec.disable_default_http_trigger = True
 
-    @min_nuclio_versions("1.12.8")
+    @min_nuclio_versions("1.13.1")
```
alonmr, line 701: should be able to do `@min_nuclio_versions("1.12.8", "1.13.1")`

TomerShor: it will work but what's the benefit?

alonmr: if users have nuclio 1.12.9 shouldn't this be supported? if you remove `1.12.8` it will require users to upgrade to 1.13.1

TomerShor: First of all, the `disable` method is supported only from 1.13.1, so I aligned the `enable` method because it doesn't make sense to only support enable (it's already  enabled by default).

Second of all, the `min_nuclio_version` decorator iterates over the versions, and returns False if _either one_ of the versions is greater than the current version. So I don't really see why we should have multiple versions at all tbh.
BTW the same function is used for min iguazio version.

alonmr: second point is a bug.. look at the previous implementation https://github.com/mlrun/mlrun/commit/87244882c16090977448d7b1500842d3732921fe#diff-91002e128e55ae0b16d8d62abbd2c9f54e83d5d584a287b43f0df436df7b2f88L74

alonmr (approved):

PR #6230, opened 2024-08-28
[Model Monitoring] Avoid creating run-context in monitoring apps

jond01:
Implements [ML-7687](https://iguazio.atlassian.net/browse/ML-7687), fixes [ML-7663](https://iguazio.atlassian.net/browse/ML-7663) and [ML-7621](https://iguazio.atlassian.net/browse/ML-7621).

Refactor `MonitoringApplicationContext` - do not inherit from `MLClientCtx` anymore. Use the project object to log artifacts and an MLRun/Nuclio logger as the messages logger.

[ML-7687]: https://iguazio.atlassian.net/browse/ML-7687?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ML-7663]: https://iguazio.atlassian.net/browse/ML-7663?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ML-7621]: https://iguazio.atlassian.net/browse/ML-7621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/model_monitoring/applications/context.py
```diff
@@ -49,79 +56,89 @@ class MonitoringApplicationContext(MLClientCtx):
...
-            ctx = (
-                super().from_dict(
-                    attrs=attrs.get(mm_constants.ApplicationEvent.MLRUN_CONTEXT, {}),
-                    **kwargs,
-                ),
-            )
-        else:
-            ctx = context
-            cls._enrich_data(ctx)
+        self.project_name = cast(str, mlrun.mlconf.default_project)
```
theSaarco: Why `default_project`? That's not the project that this serving function belongs to. I think you can extract the current project from the `GraphContext`.
When you do it like this all artifacts will be logged under the default project, which is not what we want.

mlrun/model_monitoring/applications/context.py
```diff
@@ -49,79 +56,89 @@ class MonitoringApplicationContext(MLClientCtx):
...
+        self.project = mlrun.load_project(url=self.project_name)
```
theSaarco, line 82: Would prefer to use `get_or_create_project` here. It basically calls `load_project`, but it's better to use it.

tests/model_monitoring/test_applications/test_app_context.py
```diff
@@ -0,0 +1,28 @@
...
+from mlrun.model_monitoring.applications.context import MonitoringApplicationContext
+from mlrun.projects import MlrunProject
+
+
+@pytest.mark.parametrize("method", ["log_artifact", "log_dataset", "log_model"])
+def test_log_object_signature(method: str) -> None:
+    """Future-proof the `log_x` method of MM app context with respect to the project object"""
+    assert inspect.signature(
+        getattr(MonitoringApplicationContext, method)
+    ) == inspect.signature(getattr(MlrunProject, method))
```
theSaarco, line 28: Nice.
Maybe add a test that actually makes sure that when logging artifacts through the context, they're created with the expected labels, and belong to the project as expected?

tests/system/model_monitoring/test_app.py
```diff
@@ -649,8 +649,7 @@ def _test_artifacts(self) -> None:
             if app_data.artifacts:
                 app_name = app_data.class_.NAME
                 self._logger.debug("Checking app artifacts", app_name=app_name)
-                for name in app_data.artifacts:
-                    key = f"{app_name}-logger_{name}"
+                for key in app_data.artifacts:
```
theSaarco, line 652: Perhaps here you can verify that the artifacts indeed have the labels etc.

tests/system/model_monitoring/test_model_monitoring.py
```diff
@@ -920,16 +919,16 @@ def test_batch_drift(self):
...
+            == mm_constants.ResultStatusApp.detected
+        )
 
         # Validate that the artifacts were logged under the generated context
-        assert len(project.list_artifacts(name="~drift_table_plot")) == 1
-        assert len(project.list_artifacts(name="~features_drift_results")) == 1
-        # TODO: take the artifacts from the original context when ML-5792 is done
-        # artifacts = context.artifacts
+        assert len(project.list_artifacts(name="drift_table_plot")) == 1
+        assert len(project.list_artifacts(name="features_drift_results")) == 1
```
theSaarco: here as well.

theSaarco (changes requested):
Looks good, but I think you're not retrieving the project correctly.
Also, since the main user-facing change here is the logging of artifacts, I think it's important to validate that they're logged as expected, in the correct project scope and with the proper labels.

mlrun/model_monitoring/applications/context.py
jond01: The default project is overridden in a Nuclio runtime with the appropriate environment variable, e.g.:
image

The artifacts logging is checked at the project level by unit tests and system tests, and they all passed 🙂

Do you prefer taking it from the graph context? After checking its implementation, I'm not sure it's cleaner.

tests/model_monitoring/test_applications/test_app_context.py
jond01: This is covered in:
* `tests/model_monitoring/test_application_steps.py::TestEventPreparation::test_prepare_monitoring_event` - a new unit test, with the labels check, not exactly with the project. I'll add the project check to it.
* The existing system tests you commented on - with the project but without checking the labels. I will add the labels check to them.

mlrun/model_monitoring/applications/context.py
theSaarco: To be honest, I wasn't aware of this - now that I know it's there I can see where it's happening in the code, and I see why it works. However, this can be very confusing, assuming you're not aware of this little behind-the-scenes magic MLRun is doing, and reading the code can be misleading. 
Also, once we enable debug-mode for MM apps and allow running them in some job or locally (see [ML-6362](https://iguazio.atlassian.net/browse/ML-6362)) - this will stop working, since this enrichment is unique to nuclio functions.
 
So, while for now it works I think it may be better to not rely on this. And I tend to agree that the graph-context implementation also seems a bit shaky, but I think it works across the board.
Another option is to keep this, but add a proper comment there to explain why this works *for nuclio* and also a TODO to fix once debug mode is implemented.

mlrun/model_monitoring/applications/context.py
jond01: I addressed all of your comments except for this one, which I didn't understand.
I would not like to create a project here "by mistake", so I think calling `load_project` is preferable.
Do I miss something?

davesh0812 (approved):

theSaarco (approved):
👍 LGTM.

PR #6231, opened 2024-08-28
[Application] Fix invoking with basic auth credentials

TomerShor:
`application.invoke` used the `auth_info` object for inferring the basic auth credentials.
There are a couple of issues with this:
1. `auth_info` is an object that is only used in the server side.
2. An API gateway with basic authentication can be created with any username-password, and not only the calling user's credentials.

This PR adds `credentials` argument to the application's `invoke` function, but still keeps support of extracting the credentials from the auth_info, just in case.


mlrun/runtimes/nuclio/application/application.py
```diff
@@ -480,20 +481,33 @@ def create_api_gateway(
...
         self._sync_api_gateway()
+
+        if not credentials and auth_info:
+            credentials = (auth_info.username, auth_info.password)
+
         # If the API Gateway is not ready or not set, try to invoke the function directly (without the API Gateway)
         if not self.status.api_gateway:
-            super().invoke(
+            # create a requests auth object if credentials are provided
+            auth = (
```
alonmr: this blocks providing nuclio auth info in `http_client_kwargs` (for access key for example). can do something like:
```
auth = http_client_kwargs.pop("auth") or ...
```

liranbg (approved):

PR #6237, opened 2024-08-29
[Artifacts] Lock artifacts on mark as best iteration flow

alonmr:
It's not entirely clear why we are experiencing this deadlock. The test is running a workflow with many hyper runs and logs a dataset. The tree/producer_id for all the artifacts is the workflow id but the key is generated for each iteration. It seems to be some race condition with another flow where we lock the artifacts.

https://iguazio.atlassian.net/browse/ML-6869

### DIFF
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -1121,47 +1121,57 @@ class SQLDB(DBInterface):
         if link_key:
             key = link_key
 
-        # We perform two consecutive SELECT queries and modify the two artifact records that are returned.
-        # Without no_autoflush, SQLAlchemy offloads the transaction data to the DB during the second SELECT query,
-        # which can result in a deadlock due to assumed conflicts between the transactions.
-        # Using no_autoflush ensures that all modifications don't arrive to the DB until the transaction is committed.
-        with session.no_autoflush:
-            # get the best iteration artifact record
-            query = self._query(session, ArtifactV2).filter(
-                ArtifactV2.project == project,
-                ArtifactV2.key == key,
-                ArtifactV2.iteration == link_iteration,
+        # Lock the artifacts with the same project and key (and producer_id when available) to avoid unexpected
+        # deadlocks and conform to our lock-once-when-starting logic - ML-6869
+        lock_query = self._query(
+            session,
+            ArtifactV2,
+            project=project,
+            key=key,
+        ).with_entities(ArtifactV2.id)
+        if link_tree:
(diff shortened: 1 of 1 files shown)


server/api/db/sqldb/db.py
```diff
@@ -1121,47 +1121,57 @@ def _mark_best_iteration_artifact(
...
+        # Lock the artifacts with the same project and key (and producer_id when available) to avoid unexpected
+        # deadlocks and conform to our lock-once-when-starting logic - ML-6869
+        lock_query = self._query(
+            session,
+            ArtifactV2,
+            project=project,
+            key=key,
+        ).with_entities(ArtifactV2.id)
+        if link_tree:
+            lock_query = lock_query.filter(ArtifactV2.producer_id == link_tree)
```
theSaarco, line 1133: Before this change we had 2 queries here that always returned different records (if returning anything). Now we're adding a 3rd query that contains the results of the other 2. So, there's a question - is it better to retrieve just the IDs here and go to the DB twice more for the other queries, or maybe it's better to retrieve the full objects here and perform the processing client-side on the results without querying again?
On one hand, the current approach has less memory and network requirements since only getting IDs here, and retrieving full objects only for pin-point queries with at most 1 result. OTOH, we do 3 queries with the DB, and this first query is not expected to return many results since it only asks for a specific key within a specific tree (and multiple iterations, true). 
I'm raising this here, but I really am not sure what's better... I guess our memory/processing load is higher than query overhead on the DB, so we can keep it like this.

server/api/db/sqldb/db.py
```diff
@@ -1121,47 +1121,57 @@ def _mark_best_iteration_artifact(
...
+
+        lock_query.order_by(
+            ArtifactV2.id.asc()
+        ).populate_existing().with_for_update().all()
```
theSaarco, line 1137: Do we really need to order the query? Why?

theSaarco (approved):
Looks OK, but I have some existential dilemmas there... Don't think I'd change this, but still - for awareness.

server/api/db/sqldb/db.py
alonmr: tbh I think we don't but we do it in tag artifacts where we lock them there. I think it's a sanity in case the DB locks the artifacts one by one it makes the locking deadlock proof if its not atomic

server/api/db/sqldb/db.py
alonmr: since we have been burnt by these memory issues again I prefer keeping it like this as you said. but I agree this is a question

PR #6241, opened 2024-08-30
[Docs] Check for broken links

rokatyy:
Please take a moment to read this message, as it provides important details about the changes introduced with this and review it. These updates have not only addressed and found issues but will also help us reduce the time spent verifying that all links (both relative and external) are functional and accessible. 

Some new rules and recommendations this PR introduces:

:no_entry_sign: Avoid relative links to `.html` pages if your source file is in `.md` or `.ipynb` format. Instead, use the actual link format (e.g., `file.md#anchor` rather than `file.html#anchor`). During the documentation generation process, all links will be automatically converted to `.html` (except those which are in raw-html format already)

:link: Link checking: Broken link check will now be part of the PR creation CI process (see the "Check for broken links" job) and part of the documentation generation itself (myst will inform you if any links are broken). If the check fails, you can review the job logs to see the detailed report.

:computer: Local execution: To check for broken links locally, use `make linkcheck` or `make -b linkcheck --keep-going html`.

:warning: Handling false positives: If you are confident that a link is correct but it still triggers an error, add it to the `linkcheck_ignore` list in `docs/conf.py` and kindly tag me in the PR review.

:star2: Additional benefit: Another advantage of this change is that we can once again see IDE suggestions for broken links, improving our development workflow.

:link: Code links: Please use `{py:class}` for links to class objects and `{py:meth}` for methods. 
Examples: 
{py:meth}`~mlrun.feature_store.FeatureVector.get_online_feature_service`  , 
{py:class}`~mlrun.serving.remote.RemoteStep`.


docs/tutorials/demos.md
```diff
@@ -1,8 +1,7 @@
-(demos)=
 # Demos
 
 See:
-- [Fine tune an LLM and build a BOT](https://github.com/mlrun/demo-llm-tuning/blob/main>)
+- [Fine tune an LLM and build a BOT](https://github.com/mlrun/demo-llm-tuning/blob/main)
```
jillnogold: comment is obsolete - file is totally reworked: https://github.com/mlrun/mlrun/pull/6253/files#diff-ec25285ff5e65a70889f9a3f3dbbca0501641a685af14d7e2d9f31b7a04e35a3

docs/conf.py
```diff
@@ -192,6 +191,23 @@ def current_version() -> str:
 
 smartquotes = False
 
+linkcheck_ignore = [
+    # Ignore all HTML links to local files
```
alonmr: Not sure if you meant to ignore all HTTP links here but the comment seems irrelevant to the regex
```suggestion
    # Ignore all HTTP links to local files
```

.github/workflows/docs-language-linter.yaml
```diff
@@ -35,6 +35,24 @@ jobs:
       - uses: errata-ai/vale-action@reviewdog
         with:
           reporter: github-pr-check
+  docs_linkcheck:
```
alonmr, line 38: Can be part of the docs lint job instead of taking another github runner (we get limited runners)

alonmr (commented):
Great stuff! Small comments

docs/conf.py
rokatyy: @alonmr there are links to local .html files which are in raw html format (for instance [here](https://github.com/mlrun/mlrun/blob/f1763a902e4049cb5422278c2233031ea94c4bf8/docs/index.md#L25)), we want to ignore them as these file don't exist in documentation folder.

alonmr: ok but the regex in the line below is for http so I don't understand how to comment matches the regex

rokatyy: @alonmr no, it excludes all the http/https links,I've just removed `html` at all

.github/workflows/docs-language-linter.yaml
rokatyy: @alonmr moved

liranbg (approved):

jillnogold (approved):

alonmr (approved):

TomerShor (approved):

PR #6244, opened 2024-09-01
[Linting] Select all pep8-naming rules and fix/ignore errors

jond01:
In cases of new code in 1.7, fix the errors as there's no BC concern.

https://docs.astral.sh/ruff/rules/#pep8-naming-n

Remove the `tests/http_srv.py` file - dead code without any test.


server/api/crud/model_monitoring/deployment.py
```diff
@@ -897,7 +897,7 @@ async def delete_monitoring_function(
                     function_names=[function_name],
                     access_key=access_key,
                 )
-            except mlrun.errors.MLRunStreamConnectionFailure as e:
+            except mlrun.errors.MLRunStreamConnectionFailureError as e:
                 logger.warning(
                     "Failed to delete stream resources, you may need to delete them manually",
                     project_name=project,
```
rokatyy, line 913: Why adding error in the end (is it something in PEP saying that it should be done like this)? Cause it looks a bit redundant taking into account that it's already in `mlrun.errors` module..

mlrun/common/schemas/api_gateway.py
```diff
@@ -77,7 +77,7 @@ class APIGatewaySpec(_APIGatewayBaseModel):
     name: str
     description: Optional[str]
     path: Optional[str] = "/"
-    authenticationMode: Optional[APIGatewayAuthenticationMode] = (
+    authentication_mode: Optional[APIGatewayAuthenticationMode] = (
```
rokatyy: We need it like this because we convert it to json [here](https://github.com/mlrun/mlrun/blob/3c9f80ed300620305399101b59f0620fd3d6c363/server/api/utils/clients/async_nuclio.py#L112)  and then it's used in nuclio - https://github.com/nuclio/nuclio/blob/672b8e36f9edd6e42b4685ec1d27cabae3c5f045/pkg/platform/types.go#L476

rokatyy (changes requested):

mlrun/common/schemas/api_gateway.py
jond01: Thanks, fixed

server/api/crud/model_monitoring/deployment.py
jond01: https://docs.astral.sh/ruff/rules/error-suffix-on-exception-name/
https://peps.python.org/pep-0008/#exception-names

In addition, please note almost all of the errors in `mlrun.errors` follow this guideline

rokatyy (approved):
LGTM

mlrun/errors.py
```diff
@@ -205,15 +208,15 @@ class MLRunTimeoutError(MLRunHTTPStatusError, TimeoutError):
     error_status_code = HTTPStatus.GATEWAY_TIMEOUT.value
 
 
-class MLRunInvalidMMStoreType(MLRunHTTPStatusError, ValueError):
+class MLRunInvalidMMStoreTypeError(MLRunHTTPStatusError, ValueError):
```
Eyal-Danieli, line 211: That's great but you should also refactor the class name in functions description, e.g. https://github.com/mlrun/mlrun/blob/a34fbe609a9c1ac6359279db592b2ee7834270c5/mlrun/model_monitoring/db/stores/__init__.py#L104

mlrun/frameworks/sklearn/mlrun_interface.py
```diff
@@ -97,7 +97,7 @@ def mlrun_fit(cls):
 
         def wrapper(
             self: SKLearnTypes.ModelType,
-            X: SKLearnTypes.DatasetType,
+            X: SKLearnTypes.DatasetType,  # noqa: N803 - should be lowercase "x", kept for BC
```
Eyal-Danieli, line 100: tbh I understand that this is a pep rule but there is general convention in which `X` represents machine learning inputs (either train set or test set). For instance, take a look at https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html

Eyal-Danieli (commented):
nice! 2 comments

mlrun/errors.py
jond01: Done in 9144259

mlrun/frameworks/sklearn/mlrun_interface.py
jond01: I see your point, but it should be used primarily on examples/tutorials and not as a public API.
If there's a need for it - a "noqa" waiver can be added.

Eyal-Danieli (approved):
LGTM

PR #6245, opened 2024-09-01
[Backround-tasks] Authorize internal background task with project

alonmr:
https://iguazio.atlassian.net/browse/ML-7484

### DIFF
--- a/server/api/api/endpoints/background_tasks.py
+++ b/server/api/api/endpoints/background_tasks.py
@@ -199,7 +199,7 @@ async def _authorize_get_background_task_request(
     background_task: mlrun.common.schemas.BackgroundTask,
     auth_info: mlrun.common.schemas.AuthInfo,
 ):
-    # Iguazio manifest doesn't support the global background task resource - therefore if the background task has a
+    # Iguazio manifest doesn't support the global background task resource yet - therefore if the background task has a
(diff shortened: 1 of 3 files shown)


liranbg (approved):

server/api/api/endpoints/background_tasks.py
```diff
@@ -197,17 +196,25 @@ async def list_internal_background_tasks(
...
+    # project (e.g. delete project), we can authorize on the project
+    if background_task.metadata.project:
+        return await server.api.utils.auth.verifier.AuthVerifier().query_resource_permissions(
+            mlrun.common.schemas.AuthorizationResourceTypes.project,
+            background_task.metadata.project,
             mlrun.common.schemas.AuthorizationAction.read,
             auth_info,
         )
+
+    # If there is no project we have to just omit authorization
```
moranbental: Why are these comments here?

moranbental (changes requested):

server/api/api/endpoints/background_tasks.py
alonmr: This needs to be supported one day so I would keep the comment to have a reference for it

liranbg (approved):

moranbental (approved):

PR #6247, opened 2024-09-02
[Model Monitoring] Add some logs for deleting model monitoring resources

moranbental:
Adding some logs to understand the flow when project deletion fails on model monitoring
https://iguazio.atlassian.net/browse/ML-7719

### DIFF
--- a/server/api/crud/model_monitoring/model_endpoints.py
+++ b/server/api/crud/model_monitoring/model_endpoints.py
@@ -633,6 +633,10 @@ class ModelEndpoints:
                 function_names=model_monitoring_applications,
                 access_key=model_monitoring_access_key,
             )
+            logger.debug(
+                "Successfully deleted model monitoring stream resources",
+                project_name=project_name,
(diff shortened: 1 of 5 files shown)


davesh0812 (approved):

server/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -631,6 +641,11 @@ def _delete_model_monitoring_stream_resources(
                 error=mlrun.errors.err_to_str(e),
             )
 
+        logger.debug(
+            "Successfully deleted model monitoring stream resources",
+            project_name=project_name,
+        )
```
Eyal-Danieli: I suggest moving it inside the try clause at (L635) because if there is a failure we don't break the flow, just printing a warning.

Eyal-Danieli (commented):
nice, a single comment

Eyal-Danieli (approved):

PR #6249, opened 2024-09-02
[Docs] Add note on schedules validation

alonmr:
Adding a note to describe https://iguazio.atlassian.net/browse/ML-6839 until it is fixed.
We do not want to force this validation so we need to change the mechanism in future versions.

### DIFF
--- a/docs/concepts/scheduled-jobs.md
+++ b/docs/concepts/scheduled-jobs.md
@@ -3,7 +3,7 @@
 
 Oftentimes you may want to run a `job` on a regular schedule. For example, fetching from a datasource every morning, compiling an analytics report every month, or detecting model drift every hour.
 
-> Schedules has a minimum interval that will be allowed between two scheduled jobs - by default, a job wouldn't be allowed to be scheduled to run more than 2 times in 10 minutes. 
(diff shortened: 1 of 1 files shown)


docs/concepts/scheduled-jobs.md
```diff
@@ -3,6 +3,10 @@
 
 Oftentimes you may want to run a `job` on a regular schedule. For example, fetching from a datasource every morning, compiling an analytics report every month, or detecting model drift every hour.
 
+> Schedules have a minimum interval that will be allowed between two scheduled jobs - by default, a job wouldn't be allowed to be scheduled to run more than 2 times in 10 minutes.
```
jillnogold: Is this correct?
Schedules have a required minimum interval between two scheduled jobs. By default, a job is not allowed to be scheduled twice in a 10 minute period from 00-59 minutes of the hour. 
Currently, schedules like */13 * * * * (every 13th minute), in which the job would triggered at the 52nd minute and then again at the start of the next hour (minute 0) (with only 8 minutes between runs) are not allowed.

alonmr: I removed this part `from 00-59 minutes of the hour.` but other than that it makes sense

jillnogold (approved):

liranbg (approved):

PR #6254, opened 2024-09-02
[Utils] Read verify_ssl config and disable_warnings accordingly

roei3000b:
Before making request, we want to check if we should disable warnings of `urllib3`.

### DIFF
--- a/mlrun/utils/http.py
+++ b/mlrun/utils/http.py
@@ -114,6 +114,12 @@ class HTTPSessionWithRetry(requests.Session):
         )
         while True:
             try:
+                # import here to prevent import cycle
+                from mlrun.config import config as mlconf
+
+                verify_ssl = mlconf.httpdb.http.verify
+                if not verify_ssl:
+                    urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
                 response = super().request(method, url, **kwargs)
                 return response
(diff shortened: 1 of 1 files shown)


mlrun/utils/http.py
```diff
@@ -114,6 +114,12 @@ def request(self, method, url, **kwargs):
         )
         while True:
             try:
+                # import here to prevent import cycle
+                from mlrun.config import config as mlconf
+
+                verify_ssl = mlconf.httpdb.http.verify
+                if not verify_ssl:
+                    urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
TomerShor, line 122: This is duplicated from `datastore/base.py`. move it to a common location and call it from both places

TomerShor (changes requested):

alonmr (changes requested):
Couple of notes about the import
1. You have `from ..config import config` imported already
2. The convention is to use `from mlrun import mlconf` or `mlrun.mlconf` and not `from mlrun.config import config as mlconf`
3. When you encounter a cycle import - try to fix it 1st before introducing a new dynamic import

About the solution - IIRC the solution was to change the default verify value but then we need some way for developers to work with no-verify and disable warnings. We shouldn't always disable warnings because they may be important to the users

PR #6255, opened 2024-09-02
[Datastore] Add prepare put data

tomerm-iguazio:
Add support for bytearray type in put operations.
Add exceptions for unsupported data types, except in the in-memory datastore.
Rename variables and add tests.

Notes: 
- The Redis connector always returns decoded data.
- There will be a PR that merge all common tests to some parent class.

https://iguazio.atlassian.net/browse/ML-7078


mlrun/datastore/base.py
```diff
@@ -157,6 +157,19 @@ def query(self, key, query="", **kwargs):
     def put(self, key, data, append=False):
         pass
 
+    def _prepare_put_data(self, data, append=False):
+        mode = "a" if append else "w"
+        if isinstance(data, bytearray):
+            data = bytes(data)
+        if isinstance(data, bytes):
```
gtopper: Better separate if from if-elif-else block to avoid it being misread as part of the same block.
```suggestion
            data = bytes(data)

        if isinstance(data, bytes):
```

mlrun/datastore/base.py
```diff
@@ -157,6 +157,19 @@ def query(self, key, query="", **kwargs):
...
+            return data, f"{mode}b"
+        elif isinstance(data, str):
+            return data, mode
+        else:
+            raise TypeError(
+                f"Data type unknown. Unable to put in {self.__class__.__name__}"
```
```suggestion
                f"Unable to put a value of type {type(self).__name__}"
```

tests/datastore/test_filestore.py
```diff
@@ -45,3 +45,21 @@ def test_rm_file_not_found(self, prefix):
         not_exist_url = f"{prefix}/path/to/file/not_exist_file.txt"
         data_item = mlrun.run.get_dataitem(not_exist_url)
         data_item.delete()
+
+    @pytest.mark.parametrize("data", [b"test", bytearray(b"test")])
+    def test_put_types(self, data, prefix):
+        try:
+            with tempfile.NamedTemporaryFile(suffix=".txt", delete=False) as temp_file:
```
gtopper: Why not let it do the delete for you?
```suggestion
            with tempfile.NamedTemporaryFile(suffix=".txt") as temp_file:
```

mlrun/datastore/base.py
```diff
@@ -157,6 +157,19 @@ def query(self, key, query="", **kwargs):
     def put(self, key, data, append=False):
         pass
 
+    def _prepare_put_data(self, data, append=False):
+        mode = "a" if append else "w"
+        if isinstance(data, bytearray):
```
gtopper, line 162: What's the motivation to support this?

tests/integration/google_cloud_storage/test_google_cloud_storage.py
```diff
@@ -222,8 +222,19 @@ def test_perform_google_cloud_storage_tests(
...
 
+    @pytest.mark.parametrize(
+        "setup_by, use_secrets",
+        [
+            ("credentials_file", False),
+            ("credentials_file", True),
+            ("serialized_json", False),
+            ("serialized_json", True),
+        ],
+    )
```
gtopper: Small style thing – independent parameters would normally be specified as:
```suggestion
    @pytest.mark.parametrize(
        "setup_by", ["credentials_file", "serialized_json"],
    )
    @pytest.mark.parametrize(
        "use_secrets", [False, True],
    )
```
to avoid repetition and communicate better the fact that all permutations are tried.

tomerm-iguazio: nice idea!

gtopper (approved):
LGTM

PR #6257, opened 2024-09-03
[Model Monitoring] Add test to plot stream pod graph

gtopper:
With necessary refactoring (dependency injection, moved I/O out of constructor).

**V3IO only**
v3io-v3io

**V3IO + TDEngine**
v3io-tdengine

**MySQL + V3IO**
mysql-v3io

**MySQL + TDEngine**
mysql-tdengine


mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -47,9 +47,13 @@ def __init__(
...
         self._tdengine_connection_string = kwargs.get("connection_string")
         self.database = database
-        self._connection = self._create_connection()
+
+        self._connection = None
         self._init_super_tables()
 
+    def init(self):
+        self._connection = self._create_connection()
+
```
jond01: I think that change is an obvious code smell, please mock what you need instead.
This applies to all the places you made this change.

gtopper: I disagree. In my view, I/O doesn't belong in the constructor, and dependency injection is a common pattern to enable unit testing. The need for mocking is the smell.

gtopper: I could replace `init()` with lazy init (maybe in a property), though.

jond01: > I/O doesn't belong in the constructor

Then we will probably not agree here :)
We both can find plenty of examples for it in mlrun and many other places.

This pattern, to be used in production according to this PR:
```py
a = A()
a.init()
```
Is:
1. Error-prone
2. Not natural in Python and ugly IMO

Note that you add all of this for a test, which is nice for plots, but doesn't test much TBH.

gtopper: Replaced with a property.

I would like the code to be testable for the purpose of adding more tests going forward.

gtopper: Currently, the graph in `stream_processing.py` is only tested by the full system test `test_app.py` afaict, which takes a long time to run and requires building `mlrun/mlrun-api` and replacing `mlrun-api` when testing a change.

gtopper: The issue with I/O in the constructor in this case (putting aside the general principle), is that it's not reasonable to have to mock out a bunch of I/O to be able to create the graph (a purely functional operation).

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -61,9 +61,15 @@ def __init__(
             )
 
         self._sql_connection_string = kwargs.get("store_connection_string")
-        self._engine = get_engine(dsn=self._sql_connection_string)
+        self._engine = None
         self._init_tables()
 
+    @property
+    def engine(self):
```
jond01: Can you please add the returned type hint?

mlrun/model_monitoring/db/stores/v3io_kv/kv_store.py
```diff
@@ -100,13 +100,18 @@ def __init__(
...
-        # Initialize a V3IO client instance
-        self.client = mlrun.utils.v3io_clients.get_v3io_client(
-            endpoint=mlrun.mlconf.v3io_api,
-        )
+        self._client = None
         # Get the KV table path and container
         self.path, self.container = self._get_path_and_container()
 
+    @property
+    def client(self):
```
jond01: Same here

mlrun/model_monitoring/db/tsdb/__init__.py
```diff
@@ -102,4 +102,5 @@ def get_tsdb_connector(
     tsdb_connector_factory = ObjectTSDBFactory(tsdb_connector_type)
 
     # Convert into TSDB connector object
-    return tsdb_connector_factory.to_tsdb_connector(project=project, **kwargs)
+    tsdb_connector = tsdb_connector_factory.to_tsdb_connector(project=project, **kwargs)
+    return tsdb_connector
```
jond01: Just return - same as in the previous version

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -47,9 +47,16 @@ def __init__(
             )
         self._tdengine_connection_string = kwargs.get("connection_string")
         self.database = database
-        self._connection = self._create_connection()
+
+        self._connection = None
         self._init_super_tables()
 
+    @property
+    def connection(self):
```
jond01: Type hint

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -64,14 +64,17 @@ def __init__(
...
-        )
-
+        self._frames_client: Optional[v3io_frames.client.ClientBase] = None
         self._init_tables_path()
+        self._create_table = create_table
 
-        if create_table:
-            self.create_tables()
+    @property
+    def frames_client(self):
```
jond01: Type hint

server/api/crud/model_monitoring/deployment.py
```diff
@@ -384,14 +384,26 @@ def _initial_model_monitoring_stream_processing_function(
...
 
+        secret_provider = server.api.crud.secrets.get_project_secret_provider(
+            project=self.project
+        )
+
+        tsdb_connector = mlrun.model_monitoring.get_tsdb_connector(
+            project=self.project, secret_provider=secret_provider
+        )
+        store_object = mlrun.model_monitoring.get_store_object(
+            project=self.project, secret_provider=secret_provider
```
jond01, line 395: IIRC we have a helper function for these on the server side.

Edit: looks like it's implemented only for the `get_store_object` function.

tests/model_monitoring/test_stream_processing.py
```diff
@@ -0,0 +1,56 @@
...
+    processor.apply_monitoring_serving_graph(fn, tsdb_connector, store_object)
+
+    graph = fn.spec.graph.plot(rankdir="TB")
+    print()
+    print(
+        f"Graphviz graph definition with tsdb_connector={tsdb_connector}, endpoint_store={endpoint_store}"
+    )
+    print("Feed this to graphviz, or to https://dreampuf.github.io/GraphvizOnline")
+    print()
+    print(graph)
```
jond01, line 57: How about logs instead of `print`s in the test? It will be cleaner

tests/model_monitoring/test_stream_processing.py
```diff
@@ -0,0 +1,56 @@
...
+        project_name,
+        1000,
+        10,
+        "mytarget",
+    )
+
+    fn = mlrun.new_function(
+        kind="serving",
+        name="my-fn",
+    )
```
jond01: Do we have an updated version of this?
I understood that `mlrun.code_to_function` and `mlrun.new_function` are deprecated.
Just asking, because I also encountered it recently.

server/api/crud/model_monitoring/deployment.py
```diff
@@ -384,14 +384,26 @@ def _initial_model_monitoring_stream_processing_function(
...
             function,
-            secret_provider=server.api.crud.secrets.get_project_secret_provider(
-                project=self.project
-            ),
+            tsdb_connector,
+            store_object,
         )
 
+        tsdb_connector.create_tables()
+
```
jond01: Why did you add it here?
It's already a part of the `set_credentials` flow in this module.

jond01 (commented):
Looks much cleaner, thanks.
I still have some comments.

tests/model_monitoring/test_stream_processing.py
gtopper: I'm not aware of anything to that effect. They're not deprecated in the code though.

tests/model_monitoring/test_stream_processing.py
gtopper: The logger messes up the graph output. The rest could use a logger, but then it'll be a mix of logs and prints.

server/api/crud/model_monitoring/deployment.py
gtopper: So what's the suggestion here?

jond01 (approved):
Looks good!

tests/model_monitoring/test_stream_processing.py
jond01: See this: #6235

gtopper: Ok, I changed it :)

jond01 (approved):

PR #6264, opened 2024-09-03
[Functions] Add foo func spec

adamdelman:
Example implementation for MLRun coding exercise


mlrun/common/schemas/client_spec.py
```diff
@@ -68,3 +68,4 @@ class ClientSpec(pydantic.BaseModel):
     packagers: typing.Optional[dict]
     external_platform_tracking: typing.Optional[dict]
     alerts_mode: typing.Optional[str]
+    foo: typing.Optional[str]
```
alonmr: The parent config is already a part of the client-spec. See `function: typing.Optional[Function]`.
The foo spec should be added to the `Function` class which I see you already did so this should be removed

mlrun/config.py
```diff
@@ -219,6 +219,7 @@
     },
     "function": {
         "spec": {
+            "foo": "bar",
```
alonmr: in general we want to allow expanding the configuration in the future so it should look like:
```
"foo": {
    "default": "bar",
}
```

mlrun/projects/project.py
```diff
@@ -2437,9 +2437,21 @@ def _instantiate_function(
             )
 
         if isinstance(func, str):
-            # in hub or db functions name defaults to the function name
-            if not name and not (func.startswith("db://") or func.startswith("hub://")):
-                raise ValueError("Function name must be specified")
+            # In hub or db functions name defaults to the function name
+            # If name is specified, we try to create a new function with this name.
```
alonmr: looks like this is from some other PR - can be removed

mlrun/run.py
```diff
@@ -547,7 +547,7 @@ def new_function(
             runner.spec.function_handler = handler
         else:
             runner.spec.default_handler = handler
-
+    runner.spec.foo = runner.spec.foo or mlconf.function.spec.foo
```
alonmr: why here? is should be enriched in the constructor

mlrun/runtimes/base.py
```diff
@@ -130,6 +132,7 @@ def __init__(
         # The build.source is cloned/extracted to the specified clone_target_dir
         # if a relative path is specified, it will be enriched with a temp dir path
         self._clone_target_dir = clone_target_dir or None
+        self.foo = foo
```
alonmr: default should be enriched here

mlrun/runtimes/base.py
```diff
@@ -167,6 +170,16 @@ def enrich_function_preemption_spec(self):
     def validate_service_account(self, allowed_service_accounts):
         pass
 
+    def with_foo(self, val: str):
```
alonmr: you may find shorthands around mlrun but they are legacy. latest convention is to not use them so change `val` to something like `foo_value`

alonmr: also this shouldn't be in the `FunctionSpec`. `BaseRuntime` is the class that holds the functionality. Spec is more like a dataclass

server/api/db/sqldb/db.py
```diff
@@ -4607,6 +4621,9 @@ def _find_functions(
         if hash_key is not None:
             query = query.filter(Function.uid == hash_key)
 
+            if foo_spec:
```
alonmr: this looks over indented

tests/api/api/test_functions.py
```diff
@@ -1000,19 +1002,43 @@ def failing_func():
...
 
+def test_list_functions_by_foo_spec(
+    db: sqlalchemy.orm.Session, client: fastapi.testclient.TestClient, monkeypatch
+):
+    project = "test-adam"
+    foo_spec = "adam"
+    function_1 = _generate_function(
+        function_name="adam-func", project=project, foo_spec=foo_spec, kind="job"
+    )
+    SQLDB().store_function(
```
alonmr: use either the fastapi client or the `get_db()` method for conventions consistency

tests/api/api/test_functions.py
```diff
@@ -1000,19 +1002,43 @@ def failing_func():
         assert background_task.status.state == expected_status_result
 
 
+def test_list_functions_by_foo_spec(
```
alonmr, line 1005: add some more cases like creating a function that should be filtered out and make sure you don't get it in the response

alonmr (commented):
Awesome stuff! A few comments

mlrun/projects/project.py
```diff
@@ -2437,9 +2437,21 @@ def _instantiate_function(
...
+                    existing_func_defs = self.list_functions(name=func)
+                    if existing_func_defs:
+                        existing_func_names = [
+                            existing_func_def.metadata.name
+                            for existing_func_def in existing_func_defs
+                        ]
+                    if not existing_func_defs or func not in existing_func_names:
+                        raise mlrun.MLRunInvalidArgumentError(
+                            f'Function "{func}" does not exist in the project "{self.name}".'
+                        )
```
quaark: Since we are only using this `existing_func_names` to check the current func anyway, I'd probably have the if already in the comprehension. Also the if existing_func_defs is a bit redundant and causes the if to be a bit complicated. How about?
```suggestion
                    existing_func_names = [
                        existing_func_def.metadata.name
                        for existing_func_def in self.list_functions(name=func)
                        if func == existing_func_def.metadata.name
                    ]
                    # expect list with length 1
                    if not existing_func_names:
                        raise mlrun.MLRunInvalidArgumentError(
                            f'Function "{func}" does not exist in the project "{self.name}".'
                        )
```

server/api/db/sqldb/db.py
```diff
@@ -1753,10 +1754,14 @@ def store_function(
...
         fn = self._get_class_instance_by_uid(session, Function, name, project, uid)
         if not fn:
+            foo_spec = function.get("spec", {}).get("foo")
             fn = Function(
                 name=name,
                 project=project,
                 uid=uid,
+                foo_spec=FunctionSpecFoo(
+                    data=foo_spec,
+                ),
```
TomerShor: what if `foo_spec` is `None`? should we still add it here?

TomerShor: Also, if the function - what will happen to the corresponding `FunctionFooSpec` record?

server/api/api/endpoints/functions.py
```diff
@@ -751,3 +751,67 @@ async def _get_function_status(data, auth_info: mlrun.common.schemas.AuthInfo):
             HTTPStatus.BAD_REQUEST.value,
             reason=f"Runtime error: {err_to_str(err)}",
         )
+
+
+@router.get("/projects/{project}/functions/by_foo_spec/{foo_spec}")
+async def list_functions_by_foo_spec(
```
TomerShor, line 757: you created a new endpoint, how can clients invoke it easily?
An SDK maybe? 👀

server/api/api/endpoints/runs.py
```diff
@@ -502,3 +502,57 @@ async def abort_run(
     )
 
     return background_task
+
+
+@router.get("/projects/{project}/completed_runs")
+async def list_completed_runs(
```
TomerShor, line 508: Same here

server/api/db/sqldb/models.py
```diff
@@ -769,6 +774,20 @@ class TimeWindowTracker(Base, mlrun.utils.db.BaseModel):
         def get_identifier_string(self) -> str:
             return f"{self.key}"
 
+    class FunctionSpecFoo(Base, mlrun.utils.db.BaseModel):
+        __tablename__ = "function_spec_foo"
```
quaark:
```suggestion
        __tablename__ = "function_spec_foos"
```
We usually use plurals for table names

server/api/migrations/versions/a1973a4e89df_added_foo_to_functionspec.py
```diff
@@ -0,0 +1,49 @@
...
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "a1973a4e89df"
+down_revision = "fcf2ea01f99a"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
quaark: shouldn't this migration add the relationship to the functions table as well?

quaark (approved):
Looks great!
The different uses of the MLRun abstractions seem good.
Just a few minor comments

quaark (changes requested):
approved by mistake 😅

server/api/db/sqldb/models.py
```diff
@@ -769,6 +774,20 @@ class TimeWindowTracker(Base, mlrun.utils.db.BaseModel):
...
+
+        data = Column(
+            String(length=255, collation=SQLTypesUtil.collation()), primary_key=True
+        )
+        function = relationship(Function, back_populates="foo_spec")
```
TomerShor: can you explain what `back_populates` means?

TomerShor (changes requested):
Nice job!
left some comments and suggestions

server/api/db/sqldb/models.py
adamdelman: back_populates is like using backref, but when using backref the other side of the relationship is created dynamically and is not explicitly written in the code.

server/api/migrations/versions/a1973a4e89df_added_foo_to_functionspec.py
adamdelman: WDYM?
The function.id becomes a foreign key on the function_spec_foos table, the relationship is managed by SQLAlchemy when it is defined on the ORM class.

mlrun/projects/project.py
adamdelman: This was committed by accident and was removed 😅

mlrun/runtimes/base.py
```diff
@@ -111,6 +112,7 @@ def __init__(
         pythonpath=None,
         disable_auto_mount=False,
         clone_target_dir=None,
+        foo=None,
```
roei3000b, line 115: You need to add the foo arg also to all runtimes that extend this class.

server/api/db/sqldb/db.py
```diff
@@ -1753,10 +1754,18 @@ def store_function(
             function.setdefault("metadata", {})["name"] = name
         fn = self._get_class_instance_by_uid(session, Function, name, project, uid)
         if not fn:
+            foo_spec = function.get("spec", {}).get("foo")
+            func_foo_spec = None
+            if foo_spec:
+                func_foo_spec = FunctionSpecFoo(
+                    data=foo_spec,
+                )
```
```suggestion
            func_foo_spec = FunctionSpecFoo(data=foo_spec) if foo_spec else None
```

server/api/db/sqldb/models.py
```diff
@@ -769,6 +774,18 @@ class TimeWindowTracker(Base, mlrun.utils.db.BaseModel):
         def get_identifier_string(self) -> str:
             return f"{self.key}"
 
+    class FunctionSpecFoo(Base, mlrun.utils.db.BaseModel):
```
moranbental, line 777: You should also create a `get_identifier_string` function for this class. Otherwise, it will fail at [this point](https://github.com/mlrun/mlrun/blob/7105add883de2fba974221605f2ce9239bbfbe0e/server/api/db/sqldb/db.py#L4514).
This is an abstract method [defined here](https://github.com/mlrun/mlrun/blob/30aadb6f3d83b27db15c5704d6c0be0d7503032d/mlrun/utils/db.py#L42).

server/api/api/endpoints/functions.py
moranbental, line 757: This function is duplicated from `list_functions`. Create a common function that can be called from both places.

server/api/api/endpoints/runs.py
```diff
@@ -502,3 +502,57 @@ async def abort_run(
...
+    if project != "*":
+        await server.api.utils.auth.verifier.AuthVerifier().query_project_permissions(
+            project,
+            mlrun.common.schemas.AuthorizationAction.read,
+            auth_info,
+        )
+
+    paginator = server.api.utils.pagination.Paginator()
+
+    async def _filter_runs(_runs):
```
moranbental, line 529: This function is duplicated from `list_runs`. maybe refactoring it into a private method and using it in both places.

moranbental (changes requested):
Great job! Just a few comments below.

roei3000b (changes requested):
Looks good :)
minor.

BTW, You need also to make the client get the default value of foo(see for example: `default_function_node_selector` on `ClientSpec`).

mlrun/db/httpdb.py
```diff
@@ -910,6 +910,108 @@ def list_runs(
...
+            # If running in Jupyter, can use the .show() function to display the results
+            db.list_completed_runs(name="", project=project_name).show()
+
+
+        :param name: Name of the run to retrieve.
+        :param uid: Unique ID of the run, or a list of run UIDs.
+        :param project: Project that the runs belongs to.
+        :param sort: Whether to sort the result according to their start time. Otherwise, results will be
+            returned by their internal order in the DB (order will not be guaranteed).
+        :param last: Deprecated - currently not used (will be removed in 1.8.0).
```
alonmr: no reason to have this since its deprecated

mlrun/db/httpdb.py
```diff
@@ -1281,6 +1383,31 @@ def list_functions(
         responses = self.paginated_api_call("GET", path, error, params=params)
         return self.process_paginated_responses(responses, "funcs")
 
+    def list_functions_by_foo_spec(
+        self, foo_spec=None, project=None, tag=None, labels=None, since=None, until=None
```
alonmr: always add type hints for new functions

mlrun/db/nopdb.py
```diff
@@ -23,6 +23,7 @@
 import mlrun.errors
 
 from ..config import config
+from ..lists import RunList
```
alonmr: relative imports are discouraged as they cause circular import errors. most of them are due to legacy reasons but we decided as a convention to use full imports.

mlrun/projects/project.py
```diff
@@ -3821,6 +3821,27 @@ def list_functions(self, name=None, tag=None, labels=None):
...
+
+        example::
+
+            functions = project.list_functions_by_foo_spec(foo_spec="bar")
+
+
+        :param foo_spec: Return only functions with a specific foo_spec.
+        :param tag: Return function versions with specific tags. To return only tagged functions, set tag to ``"*"``.
+        :param labels: Return functions that have specific labels assigned to them.
+        :returns: List of function objects.
```
alonmr: General suggestion on docstring formatting
```suggestion
        Example::

            functions = project.list_functions_by_foo_spec(foo_spec="bar")


        :param foo_spec: Return only functions with a specific foo_spec.
        :param tag:      Return function versions with specific tags. To return only tagged functions, set tag to ``"*"``.
        :param labels:   Return functions that have specific labels assigned to them.

        :returns: List of function objects.
```

mlrun/projects/project.py
```diff
@@ -3929,6 +3950,58 @@ def list_runs(
...
+        """Retrieve a list of runs, filtered by various options.
+
+        The returned result is a `` (list of dict), use `.to_objects()` to convert it to a list of RunObjects,
+        `.show()` to view graphically in Jupyter, `.to_df()` to convert to a DataFrame, and `compare()` to
+        generate comparison table and PCP plot.
+
+        Example::
+
+            # return a list of runs matching the name and label and compare
+            runs = project.list_runs(name="download", labels="owner=admin")
```
alonmr: Docstring need some updates
```suggestion
            runs = project.list_completed_runs(name="download", labels="owner=admin")
```

mlrun/projects/project.py
```diff
@@ -3929,6 +3950,58 @@ def list_runs(
...
+        name: Optional[str] = None,
+        uid: Optional[Union[str, list[str]]] = None,
+        project: Optional[str] = None,
+        sort: bool = True,
+        iter: bool = False,
+        start_time_from: datetime = None,
+        start_time_to: datetime = None,
+        **kwargs,
+    ) -> mlrun.lists.RunList:
+        """Retrieve a list of runs, filtered by various options.
```
```suggestion
        """Retrieve a list of completed runs, filtered by various options.
```

server/api/crud/client_spec.py
```diff
@@ -121,6 +121,7 @@ def get_client_spec(
             ),
             packagers=self._get_config_value_if_not_default("packagers"),
             alerts_mode=self._get_config_value_if_not_default("alerts.mode"),
+            foo_spec=self._get_config_value_if_not_default("function.spec.foo"),
```
alonmr: this is not needed because it is passed under
```
function=self._get_config_value_if_not_default("function"),
```

server/api/db/sqldb/db.py
```diff
@@ -1766,6 +1767,27 @@ def store_function(
...
+                foo_spec_obj = FunctionSpecFoo(
+                    data=foo_spec,
+                    project=project,
+                    function_id=fn.id,
+                )
+                fn.foo_spec = existing_foo_spec
+        objects = [fn]
+        if foo_spec_obj is not None:
+            objects.append(foo_spec_obj)
+        self._upsert(session, objects)
```
alonmr: function is upserted twice (line 1769)

server/api/db/sqldb/db.py
```diff
@@ -1856,7 +1880,7 @@ def get_function(
 
     def delete_function(self, session: Session, project: str, name: str):
         logger.debug("Removing function from db", project=project, name=name)
-
+        self._delete(session, FunctionSpecFoo, project=project)
```
alonmr: I don't see a reason to delete them separately, they should be cascaded with the functions.
You could remove dangling once like we do with labels below.

server/api/db/sqldb/db.py
```diff
@@ -4545,7 +4574,7 @@ def _try_commit_obj():
                 config.httpdb.db.commit_retry_interval,
                 config.httpdb.db.commit_retry_timeout,
                 logger,
-                False,
+                True,
```
alonmr: what's this?

alonmr (commented):
Very nice! Some comments below

server/api/db/sqldb/db.py
adamdelman: I'm not a biq fan of one line expressions like this

PR #6266, opened 2024-09-03
[Project] Skip sync functions when scheduling workflows

alonmr:
https://iguazio.atlassian.net/browse/ML-6194

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -3040,7 +3040,7 @@ class MlrunProject(ModelObj):
                 "Remote repo is not defined, use .create_remote() + push()"
             )
 
-        if engine not in ["remote"]:
+        if schedule or engine not in ["remote"]:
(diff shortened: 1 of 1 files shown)


mlrun/projects/project.py
```diff
@@ -3040,7 +3040,7 @@ def run(
                 "Remote repo is not defined, use .create_remote() + push()"
             )
 
-        if engine not in ["remote"]:
+        if schedule or engine not in ["remote"]:
             # for remote runs we don't require the functions to be synced as they can be loaded dynamically during run
```
liranbg, line 3044: why is it true for schedule as well?

alonmr: schedules use remote engine. a few lines below there is the engine override if a schedule was given

TomerShor (approved):

PR #6274, opened 2024-09-04
[Model Monitoring] Fix skip stream deletion if failed

Eyal-Danieli:
Currently, the delete project API fails if the model monitoring stream resources (either `v3io` or `kafka`) cannot be deleted, except for specific errors that are already handled. In this PR, we catch all possible errors, log them as a warning in the API, and ensure the delete project API does not fail.

Related JIRA: https://iguazio.atlassian.net/browse/ML-7719

### DIFF
--- a/server/api/crud/model_monitoring/deployment.py
+++ b/server/api/crud/model_monitoring/deployment.py
@@ -980,11 +980,10 @@ class MonitoringDeployment:
                     logger.debug(
                         "Deleted v3io stream", project=project, stream_path=stream_path
                     )
-                except v3io.dataplane.response.HttpResponseError as e:
-                    logger.warning(
-                        "Failed to delete v3io stream",
-                        stream_path=stream_path,
-                        error=mlrun.errors.err_to_str(e),
+                except Exception as exc:
+                    # Raise an error that will be caught by the caller and skip the deletion of the stream
+                    raise mlrun.errors.MLRunStreamConnectionFailure(
+                        f"Failed to delete v3io stream {stream_path}, {mlrun.errors.err_to_str(exc)}"
                     )
         elif stream_paths[0].startswith("kafka://"):
             # Delete Kafka topics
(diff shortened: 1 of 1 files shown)


server/api/crud/model_monitoring/deployment.py
```diff
@@ -1007,22 +1006,10 @@ def _delete_model_monitoring_stream_resources(
...
-                    "Kafka model monitoring topics not found, probably not created",
-                    topics=topics,
-                    error=mlrun.errors.err_to_str(e),
-                )
-            except kafka.errors.NoBrokersAvailable as e:
+            except Exception as exc:
                 # Raise an error that will be caught by the caller and skip the deletion of the stream
                 raise mlrun.errors.MLRunStreamConnectionFailure(
-                    f"Failed to delete kafka topics {topics}, no brokers available, {mlrun.errors.err_to_str(e)}"
+                    f"Failed to delete kafka topics {topics}, {mlrun.errors.err_to_str(exc)}"
```
gtopper: Should we use `from exc` here, to include the original stack trace?

server/api/crud/model_monitoring/deployment.py
```diff
@@ -980,11 +980,10 @@ def _delete_model_monitoring_stream_resources(
...
                     )
-                except v3io.dataplane.response.HttpResponseError as e:
-                    logger.warning(
-                        "Failed to delete v3io stream",
-                        stream_path=stream_path,
-                        error=mlrun.errors.err_to_str(e),
+                except Exception as exc:
+                    # Raise an error that will be caught by the caller and skip the deletion of the stream
+                    raise mlrun.errors.MLRunStreamConnectionFailure(
+                        f"Failed to delete v3io stream {stream_path}, {mlrun.errors.err_to_str(exc)}"
```
gtopper: Should we use from exc here, to include the original stack trace?

gtopper: Why do we have to repackage the exception anyway? Why not just let it propagate up the stack?

Eyal-Danieli: > Why do we have to repackage the exception anyway? Why not just let it propagate up the stack?

We want to wrap this part in the **`MLRunStreamConnectionFailure`** error, which we know the caller can handle. If a different error occurs before it, we may want the process to fail.

gtopper (approved):

PR #6276, opened 2024-09-04
[GenAI] Added and Updated on steps

ZeevRispler:
Added sentiment step, edited retrival step, changed main file as guy suggested


mlrun/genai/chains/retrieval.py
```diff
@@ -56,36 +59,70 @@ class DocumentRetriever:
...
             verbose=verbose,
         )
-        handler = DocumentCallbackHandler()
-        handler.verbose = verbose
-        self.chain_type = chain_type
+        self.cb = DocumentCallbackHandler()
+        self.cb.verbose = verbose
         self.verbose = verbose
-        self.cb = handler
+        self.chain_type = chain_type  # TODO check what this is
```
yonishelach: According to the langchain docstrings:
```
chain_type: Type of document combining chain to use. Should be one of "stuff",
            "map_reduce", "refine" and "map_rerank".
```
So, it is look like the strategy is to load the documents into the prompt. But it seems that this option is being deprecated. need to check it out: https://api.python.langchain.com/en/latest/chains/langchain.chains.question_answering.chain.load_qa_chain.html

mlrun/genai/chains/sentiment_analysis.py
```diff
@@ -0,0 +1,68 @@
...
+from mlrun.genai.chains.base import ChainRunner
+
+
+class SentimentAnalysisStep(ChainRunner):
+    """
+    Processes sentiment analysis on a given text.
+    """
+
+    # Default model to use as model and tokenizer if not given
+    DEFAULT_MODEL = "cardiffnlp/twitter-roberta-base-sentiment"
```
yonishelach, line 24: I am not sure if this is the best place to configure the default model, maybe its better in the config

yonishelach (commented):
Looks good, added some information regarding chain_type, maybe its better to add some of what I wrote to the todo.

mlrun/genai/chains/sentiment_analysis.py
ZeevRispler: need to talk about where to put defaults @guy1992l

guy1992l:
can be closed.

PR #6281, opened 2024-09-06
[API Gateway] Fix joining URL when getting invoke_url on server side

rokatyy:
Fixes the issue where `/` wasn't added between host and path when generating `invoke_url` on server side. 
Now it will be always added and we also ensure that `/` isn't duplicated. 

Jira - https://iguazio.atlassian.net/browse/NUC-252

### DIFF
--- a/mlrun/common/schemas/api_gateway.py
+++ b/mlrun/common/schemas/api_gateway.py
@@ -103,11 +103,11 @@ class APIGateway(_APIGatewayBaseModel):
         ]
 
     def get_invoke_url(self):
-        return (
-            self.spec.host + self.spec.path
-            if self.spec.path and self.spec.host
-            else self.spec.host
-        ).rstrip("/")
+        if self.spec.host and self.spec.path:
+            if not self.spec.path.startswith("/"):
+                return f"{self.spec.host.rstrip('/')}/{self.spec.path}"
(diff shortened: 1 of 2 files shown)


mlrun/common/schemas/api_gateway.py
```diff
@@ -103,11 +103,11 @@ def get_function_names(self):
...
     def get_invoke_url(self):
-        return (
-            self.spec.host + self.spec.path
-            if self.spec.path and self.spec.host
-            else self.spec.host
-        ).rstrip("/")
+        if self.spec.host and self.spec.path:
+            if not self.spec.path.startswith("/"):
+                return f"{self.spec.host.rstrip('/')}/{self.spec.path}"
+            return f"{self.spec.host.rstrip('/')}{self.spec.path}"
```
TomerShor: You can `rstrip` the host and `lstrip` the path in both cases, without an if condition

TomerShor: Also, we previously removed the last slash to not duplicate urls with and without slash suffix.
Make sure this is still the case (should be done in the BE IIRC)

TomerShor (changes requested):

liranbg (approved):

PR #6285, opened 2024-09-08
[Pipelines] Add project name filter to list_pipelines when no filter provided

yaelgen:
resolves: https://iguazio.atlassian.net/browse/ML-7673

When no filter is provided and the project is not `*`, the project name will automatically be appended as a filter, ensuring that the query only returns runs containing the project name. This mirrors the current behavior in the UI and improves query performance from SDK. This change is backward compatible as the logic to append the project name to the run name has been in place since 2021.


liranbg (approved):
🚀

mlrun/utils/helpers.py
```diff
@@ -1007,6 +1007,10 @@ def get_workflow_url(project, id=None):
     return url
 
 
+def get_kfp_project_filter(project_name: str) -> str:
+    return f'{{"predicates": [{{"key": "name", "op": 9, "string_value": "{project_name}"}}]}}'
```
liranbg: Add explanation that we filter kfp runs using substring predicate as a heuristic to get a specific project runs.

TomerShor: Also, it's better to use a variable than simply a number so we know what's going on here in the future.
https://github.com/kubeflow/pipelines/blob/master/backend/api/v1beta1/filter.proto#L43-L45

```suggestion
def get_kfp_project_filter(project_name: str) -> str:
    is_substring_op = 9
    return f'{{"predicates": [{{"key": "name", "op": {is_substring_op}, "string_value": "{project_name}"}}]}}'
```

TomerShor (approved):

alonmr (commented):
I think in general we want to leave a backdoor to listing with no filter (i.e default filter should be `None` and if `filter=""` than we don't enrich with default filter).
It might be possible to provide a run name without a project prefix though it's possibly an extreme edge case. Either way I guess users could get a specific pipeline with run id so I'm ok with the fix as is but just a suggestion.

liranbg (approved):

quaark (approved):

PR #6287, opened 2024-09-08
[Projects] Adding project full_sync description

moranbental:


server/api/utils/projects/follower.py
```diff
@@ -79,6 +79,9 @@ def initialize(self):
                 # full_sync=True was a temporary measure to handle the move of mlrun from single instance to
                 # chief-worker model. Now it is possible to delete projects that are not in the leader therefore
                 # we don't necessarily need to archive projects that are not in the leader.
+                # A full sync occurs only during initialization (and not in the periodic task) to avoid cases where
+                # a bug in Iguazio returns an empty list of projects, which could potentially result in the
+                # deletion of all MLRun projects
```
alonmr: we don't delete the projects
```suggestion
                # archiving of all MLRun projects
```

moranbental: We will first archive the projects, and then delete them when receiving a delete request when the project is in the archived state
https://github.com/mlrun/mlrun/blob/fefd8ab253b8928e21693fc22653891b62d12a3c/server/api/api/endpoints/projects.py#L281

alonmr (approved):

PR #6289, opened 2024-09-08
[Project] Add owner when logging artifacts

TomerShor:
When artifact are logged via a context, the artifact producer has an `owner` field which is the run owner.
When logging via a project, the owner field was left empty.

In this PR we populate the owner field when logging artifacts via a project according to the following logic:
- If we have the current username at hand (in the `V3IO_USERNAME` env var) - use it as the artifact owner.
- If not - use the project owner as the artifact owner.

Resolves https://iguazio.atlassian.net/browse/ML-7730

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -4341,9 +4341,7 @@ class MlrunProject(ModelObj):
         return self._get_hexsha() or str(uuid.uuid4())
 
     def _resolve_artifact_owner(self):
-        if username := os.getenv("V3IO_USERNAME"):
-            return username
-        return self.spec.owner
+        return os.getenv("V3IO_USERNAME") or self.spec.owner
 
 
 def _set_as_current_default_project(project: MlrunProject):
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -18,6 +18,7 @@ import glob
 import http
 import importlib.util as imputil
 import json
+import os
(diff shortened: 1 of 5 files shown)


tests/projects/test_project.py
```diff
@@ -1109,6 +1114,32 @@ def test_replace_exported_artifact_producer(rundb_mock):
...
+    [
+        ("project-owner", None),
+        (None, "username"),
+        ("project-owner", "username"),
+    ],
+)
+def test_artifact_owner(rundb_mock, project_owner, username):
+    if username:
+        # set the V3IO_USERNAME env var
+        os.environ["V3IO_USERNAME"] = username
```
alonmr: use `monkeypatch.setenv` instead of try and finally

mlrun/projects/project.py
```diff
@@ -4337,6 +4340,11 @@ def _resolve_existing_artifact(
     def _get_project_tag(self):
         return self._get_hexsha() or str(uuid.uuid4())
 
+    def _resolve_artifact_owner(self):
+        if username := os.getenv("V3IO_USERNAME"):
+            return username
+        return self.spec.owner
```
```suggestion
        return  os.getenv("V3IO_USERNAME") or self.spec.owner
```

alonmr (commented):
Please add a ut in `test_execution.py` as well

liranbg (approved):

PR #6294, opened 2024-09-09
[Model Monitoring] Scale out stream function [1.7.x]

gtopper:
[ML-7731](https://iguazio.atlassian.net/browse/ML-7731)

Ran `TestMonitoringAppFlow::test_app_flow` and `TestModelMonitoringKafka::test_model_monitoring_with_kafka_stream`.

[ML-7731]: https://iguazio.atlassian.net/browse/ML-7731?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gtopper:
@davesh0812, please give [deployment.py](https://github.com/mlrun/mlrun/pull/6294/files#diff-1f2285f1be3191e8c3d4d98ec9e51e5dd0f914df73e8d575f465e7178ced5bdc) a look, because I had a major merge conflict with your changes in #6305.

davesh0812 (commented):
deployment.py looks ok from my side

mlrun/serving/v2_serving.py
```diff
@@ -225,13 +230,29 @@ def _pre_event_processing_actions(self, event, event_body, op):
         request = self.preprocess(event_body, op)
         return self.validate(request, op)
 
+    @property
+    def versioned_model_name(self):
+        version = self.version or "latest"
+        return f"{self.name}:{version}"
+
```
davesh0812: please take a look on `def _init_endpoint_record`  function and use the same logic for the model version and enrich it from the model artifact:
```
    # Generating version model value based on the model name and model version
    if model.model_path and model.model_path.startswith("store://"):
        # Enrich the model server with the model artifact metadata
        model.get_model()
        if not model.version:
            # Enrich the model version with the model artifact tag
            model.version = model.model_spec.tag
        model.labels = model.model_spec.labels
        versioned_model_name = f"{model.name}:{model.version}"
    else:
        versioned_model_name = f"{model.name}:latest"
```

mlrun/serving/v2_serving.py
```diff
@@ -225,13 +230,29 @@ def _pre_event_processing_actions(self, event, event_body, op):
...
         original_body = event.body
         event_body = _extract_input_data(self._input_path, event.body)
         event_id = event.id
         op = event.path.strip("/")
+
+        # Generating model endpoint ID based on function uri and model version
+        model_endpoint_uid = mlrun.common.model_monitoring.create_model_endpoint_uid(
+            function_uri=self.context.server.function_uri,
+            versioned_model=self.versioned_model_name,
+        ).uid
```
davesh0812: Why not to use `self.model_endpoint_uid`?

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1188,12 +1198,6 @@ def set_credentials(
             )
         )
 
-        self._create_stream_output(
-            stream_path=secrets_dict.get(
-                mlrun.common.schemas.model_monitoring.ProjectSecretKeys.STREAM_PATH
-            )
-        )
-
```
davesh0812: Why you want to delete this code?
We need some how to check that the cred of the stream are valid, you can do it differently.

mlrun/config.py
```diff
@@ -532,8 +533,54 @@
...
-        "application_stream_args": {"shard_count": 1, "retention_period_hours": 24},
+        "serving_stream": {
+            "v3io": {
+                "shard_count": 2,
+                "retention_period_hours": 24,
+                "num_workers": 1,
+                "min_replicas": 2,
+                "max_replicas": 2,
+            },
+            "kafka": {
```
davesh0812, line 544: Can we also add a retention  config for `kafka`?

davesh0812 (commented):
Nice :)
Few comments

server/api/crud/model_monitoring/deployment.py
gtopper: To avoid creating the stream twice, which was the cause of [ML-7922](https://iguazio.atlassian.net/browse/ML-7922).

I'll add a check for access that doesn't create the stream.

gtopper:
`TestMonitoringAppFlow::test_app_flow` passed.

mlrun/config.py
gtopper: Will leave it out for now because the old client we use doesn't expose a clear API for this, and this PR is already in the pipeline for too long, mostly due to 1.7.0 tasks preempting it.

gtopper:
`TestMonitoringAppFlow::test_app_flow` passed.

mlrun/serving/v2_serving.py
gtopper: It's not always initialized, e.g. when running a mock server. And anyway, we're planning to move `_init_endpoint_record` from here to the deployment phase.

davesh0812: If it's not initialized it means that there is no model endpoint.
Why do you need it when running a mock server?

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1198,11 +1209,11 @@ def set_credentials(
...
         )
 
-        self._create_stream_output(
-            stream_path=secrets_dict.get(
+        if not mlrun.mlconf.is_ce_mode():
+            stream_path = secrets_dict.get(
                 mlrun.common.schemas.model_monitoring.ProjectSecretKeys.STREAM_PATH
             )
-        )
+            self._verify_v3io_access(stream_path)
```
davesh0812, line 1216: What about Kafka?
Can you create `_verify_stream_access` and to verify it according to the stream path.

server/api/crud/model_monitoring/deployment.py
```diff
@@ -266,7 +266,10 @@ def deploy_model_monitoring_writer_application(
             )
 
     def apply_and_create_stream_trigger(
-        self, function: mlrun.runtimes.ServingRuntime, function_name: str
+        self,
+        function: mlrun.runtimes.ServingRuntime,
+        function_name: str,
+        stream_args,
```
davesh0812: please add type hints

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1212,6 +1223,23 @@ def set_credentials(
...
 
+    def _verify_v3io_access(self, stream_path: str):
+        import v3io.dataplane
+
+        stream_path = server.api.crud.model_monitoring.get_stream_path(
+            project=self.project, stream_uri=stream_path
+        )
+        v3io_client = v3io.dataplane.Client(endpoint=mlrun.mlconf.v3io_api)
+        container, path = split_path(stream_path)
+        print(f"111 stream_path={stream_path}, container={container}, path={path}")
```
davesh0812: necessary? If yes please use mlrun logger.

davesh0812:
I saw that you add a label of 1.7.0 to this pr... I changed it to 1.7.1 fill free to change it if I did something wrong

mlrun/serving/v2_serving.py
gtopper: If we rely on `post_init` here, then `self.model_endpoint_uid` will be `None` in mock server. I don't remember the specific failure that causes, but it did fail from what I remember.

Again, since our intention is to move `_init_endpoint_record` to the deployment stage ([ML-7930](https://iguazio.atlassian.net/browse/ML-7930)), I can't imagine why we would want to spend time on this, only to change the code back shortly after.

server/api/crud/model_monitoring/deployment.py
gtopper: Since it's not a regression in this PR, and not a part of the requirement, this will need a separate ticket/PR.

gtopper:
@davesh0812, 1.7.0 label was added on September 22, back when this feature was planned for 1.7.0. Your change to 1.7.1 is correct, thanks.

mlrun/config.py
gtopper: Added a TODO as discussed.

mlrun/serving/v2_serving.py
gtopper: I remembered that there was an issue with this, but it works fine when I try it now, so I pushed this change. Do note we will sort of have to change it back soon as part of [ML-7930](https://iguazio.atlassian.net/browse/ML-7930).

server/api/crud/model_monitoring/deployment.py
gtopper: Opened [ML-8123](https://iguazio.atlassian.net/browse/ML-8123) to track this.

davesh0812 (approved):

PR #6297, opened 2024-09-10
[API] Enhance and tweak logs

liranbg:
Enhanced debugging experience / logs readability for project deletion flow

### DIFF
--- a/server/api/crud/projects.py
+++ b/server/api/crud/projects.py
@@ -168,6 +168,10 @@ class Projects(
             skip_notification_secrets=True,
         )
 
+        logger.debug(
+            "Deleting project runtime resources",
+            project_name=name,
+        )
         # delete runtime resources
         server.api.crud.RuntimeResources().delete_runtime_resources(
             session,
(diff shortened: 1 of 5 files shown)


server/api/crud/projects.py
```diff
@@ -187,6 +191,10 @@ def delete_project_resources(
         ):
             server.api.crud.Logs().delete_project_logs_legacy(name)
 
+        logger.debug(
```
moranbental, line 194: Why not move all the logs inside the functions? For example, here in `delete_project_alert_events `

moranbental (changes requested):

server/api/utils/background_tasks/internal.py
```diff
@@ -64,15 +64,15 @@ def create_background_task(
 
         project_name = None
         if project := kwargs.get("project", None):
-            project_name = project.metadata.name
+            project_name = project.metadata.name if project else None
```
alonmr, line 67: project will never be None here because of the `:=` assignment

server/api/crud/projects.py
liranbg: because functions can get called from multiple places and lost context of the story im trying to tell using the delete project resources itself

moranbental: but most of the functions you added logs to are only called once, such as `_wait_for_nuclio_project_deletion`, `delete_project_alert_events` `delete_project_related_resources` and so on

moranbental (approved):

PR #6305, opened 2024-09-10
[Model Monitoring] Define 1.7.0 as the minimum Client Version which supports MM & delete strem trigger duplication of the stream pod

davesh0812:
Currently, we maintain two different stream triggers for the mm-stream function to accommodate older clients. In this PR, we are removing the old stream trigger, so only clients with mlrun >= 1.7.0 will be able to use the model monitoring facilities.

More of the same for now user have to be align between server and client (==1.7.0) in order to use MM APIS.

related to [ML-7762](https://iguazio.atlassian.net/browse/ML-7762)

[ML-7762]: https://iguazio.atlassian.net/browse/ML-7762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -532,7 +532,6 @@ default_config = {
         "store_prefixes": {
             "default": "v3io:///users/pipelines/{project}/model-endpoints/{kind}",
             "user_space": "v3io:///projects/{project}/model-endpoints/{kind}",
-            "stream": "",  # TODO: Delete in 1.9.0
             "monitoring_application": "v3io:///users/pipelines/{project}/monitoring-apps/",
         },
         # Offline storage path can be either relative or a full path. This path is used for general offline data
@@ -545,7 +544,6 @@ default_config = {
         "parquet_batching_max_events": 10_000,
         "parquet_batching_timeout_secs": timedelta(minutes=1).total_seconds(),
         # See mlrun.model_monitoring.db.stores.ObjectStoreFactory for available options
-        "store_type": "v3io-nosql",  # TODO: Delete in 1.9.0
         "endpoint_store_connection": "",
         # See mlrun.model_monitoring.db.tsdb.ObjectTSDBFactory for available options
         "tsdb_connection": "",
@@ -1126,7 +1124,7 @@ class Config:
         artifact_path: str = None,
         function_name: str = None,
         **kwargs,
-    ) -> typing.Union[str, list[str]]:
+    ) -> str:
         """Get the full path from the configuration based on the provided project and kind.
 
         :param project:         Project name.
@@ -1165,17 +1163,10 @@ class Config:
                     else f"{kind}-{function_name.lower()}",
                 )
             elif kind == "stream":  # return list for mlrun<1.6.3 BC
-                return [
-                    # TODO: remove the first stream in 1.9.0
-                    mlrun.mlconf.model_endpoint_monitoring.store_prefixes.default.format(
-                        project=project,
-                        kind=kind,
-                    ),  # old stream uri (pipelines) for BC ML-6043
-                    mlrun.mlconf.model_endpoint_monitoring.store_prefixes.user_space.format(
-                        project=project,
-                        kind=kind,
-                    ),  # new stream uri (projects)
-                ]
+                return mlrun.mlconf.model_endpoint_monitoring.store_prefixes.user_space.format(
+                    project=project,
+                    kind=kind,
+                )  # new stream uri (projects)
             else:
                 return mlrun.mlconf.model_endpoint_monitoring.store_prefixes.default.format(
                     project=project,
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -1121,8 +1121,8 @@ class Config:
         project: str = "",
         kind: str = "",
         target: str = "online",
-        artifact_path: str = None,
-        function_name: str = None,
+        artifact_path: typing.Optional[str] = None,
+        function_name: typing.Optional[str] = None,
         **kwargs,
     ) -> str:
         """Get the full path from the configuration based on the provided project and kind.
@@ -1140,8 +1140,7 @@ class Config:
                                 relative artifact path will be taken from the global MLRun artifact path.
         :param function_name:    Application name, None for model_monitoring_stream.
 
-        :return:                Full configured path for the provided kind. Can be either a single path
-                                or a list of paths in the case of the online model monitoring stream path.
+        :return:                Full configured path for the provided kind.
         """
 
         if target != "offline":
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -1162,11 +1162,11 @@ class Config:
                     if function_name is None
                     else f"{kind}-{function_name.lower()}",
                 )
-            elif kind == "stream":  # return list for mlrun<1.6.3 BC
+            elif kind == "stream":
                 return mlrun.mlconf.model_endpoint_monitoring.store_prefixes.user_space.format(
                     project=project,
                     kind=kind,
-                )  # new stream uri (projects)
+                )
             else:
                 return mlrun.mlconf.model_endpoint_monitoring.store_prefixes.default.format(
                     project=project,
--- a/mlrun/model_monitoring/helpers.py
+++ b/mlrun/model_monitoring/helpers.py
@@ -70,8 +70,6 @@ def get_stream_path(
             function_name=function_name,
         )
 
-    if isinstance(stream_uri, list):  # ML-6043 - user side gets only the new stream uri
-        stream_uri = stream_uri[1]  # get new stream path, under projects
     return mlrun.common.model_monitoring.helpers.parse_monitoring_stream_path(
         stream_uri=stream_uri, project=project, function_name=function_name
     )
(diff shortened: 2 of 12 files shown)


mlrun/config.py
```diff
@@ -545,7 +544,6 @@
         "parquet_batching_max_events": 10_000,
         "parquet_batching_timeout_secs": timedelta(minutes=1).total_seconds(),
         # See mlrun.model_monitoring.db.stores.ObjectStoreFactory for available options
-        "store_type": "v3io-nosql",  # TODO: Delete in 1.9.0
```
Eyal-Danieli, line 548: we need to align `CE` as well if we remove that `store_type`. Until we do that, I'm afraid that the CE installation will be broken (the whole CE not just mm). In other words we either need to update the CE before this PR is merged or maybe let's keep that `store_type` configuration just as a static configuration that doesn't affect anything and remove it after we update the CE installation.

mlrun/config.py
```diff
@@ -1126,7 +1124,7 @@ def get_model_monitoring_file_target_path(
         artifact_path: str = None,
         function_name: str = None,
         **kwargs,
-    ) -> typing.Union[str, list[str]]:
+    ) -> str:
```
Eyal-Danieli, line 1127:
```suggestion
    ) -> typing.Union[str, list[str]]:
```
I do think we need to keep that docstring as long as we support the mlrun<1.6.3 bc (L1165). We can modify the `return:` documentation, explaining that it can return a list only for older versions.

server/api/api/endpoints/model_monitoring.py
```diff
@@ -50,9 +53,19 @@ def __post_init__(self) -> None:
...
 
 async def _verify_authorization(
-    project: str, auth_info: mlrun.common.schemas.AuthInfo
+    project: str, auth_info: mlrun.common.schemas.AuthInfo, client_version: str
 ) -> None:
     """Verify project authorization"""
+    if (
+        semver.Version.parse(client_version)
+        < semver.Version.parse(MINIMUM_CLIENT_VERSION_FOR_MM)
+        or "unstable" in client_version
```
Eyal-Danieli, line 62: why to exclude `unstable`?

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1057,101 +982,14 @@ def check_if_credentials_are_set(
...
-                != mlrun.common.schemas.model_monitoring.ProjectSecretKeys.STREAM_PATH
-                else True
-                for key, val in credentials_dict.items()
-            ]
-        ):
+        if all([val is not None for key, val in credentials_dict.items()]):
             return
-        if with_upgrade_case_check:
-            with_upgrade_case_check = self._set_credentials_after_server_upgrade(
-                client_version=client_version
```
Eyal-Danieli, line 1073: remove both `client_version` and `with_upgrade_case_check` from func docs

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1057,101 +982,14 @@ def check_if_credentials_are_set(
...
-                replace_creds=True,
-            )
-        return set_cred
+        raise mlrun.errors.MLRunBadRequestError(
+            "Model monitoring credentials are not set. "
+            "Please set them using the set_model_monitoring_credentials API/SDK "
+            "or pass fetch_credentials_from_sys_config=True when using enable_model_monitoring API/SDK."
+        )
 
     def set_credentials(
```
Eyal-Danieli, line 994: remove `client_version` from func docs

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1271,31 +1103,23 @@ def set_credentials(
...
                 and mlrun.mlconf.is_ce_mode()
             ):
                 raise mlrun.errors.MLRunInvalidMMStoreTypeError(
                     "In CE mode, only kafka stream are supported for stream path"
                 )
-            elif stream_path == mm_constants.V3IO_MODEL_MONITORING_DB:
-                # TODO: Delete in 1.9.0 (for BC)
-                stream_path = ""
             elif stream_path:
                 if stream_path.startswith("kafka://") and "?topic" in stream_path:
```
```suggestion
                elif stream_path.startswith("kafka://") and "?topic" in stream_path:
```
due to the above changes we already know that there is a stream path in this part

server/api/crud/model_monitoring/deployment.py
```diff
@@ -1271,31 +1103,23 @@ def set_credentials(
...
                     raise mlrun.errors.MLRunInvalidMMStoreTypeError(
                         "Custom kafka topic is not allowed"
                     )
                 elif not stream_path.startswith("kafka://") and (
```
Eyal-Danieli: you can remove that `elif` and just raise the error because we know that there is a stream which is not kafka or v3io

server/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -593,30 +604,13 @@ def _delete_model_monitoring_stream_resources(
         :param db_session:                    A session that manages the current dialog with the database.
```
Eyal-Danieli:
```suggestion

```

Eyal-Danieli (commented):
nice, few comments

server/api/api/endpoints/model_monitoring.py
assaf758: I assume to make it easy to run tests from your ide/laptop.
@davesh0812, don't you need to do the same in `should_redeploy_monitoring_stream()`?

mlrun/config.py
davesh0812: We not support it any more, the function returning only a single string

server/api/api/endpoints/model_monitoring.py
davesh0812: For testing, like @assaf758 said:)
About - `should_redeploy_monitoring_stream()` - I removed this function so I don't sure what you mean.

server/api/crud/model_monitoring/deployment.py
davesh0812: not sure about it..

mlrun/config.py
davesh0812: https://github.com/mlrun/ce/pull/131 please review :)

Eyal-Danieli (approved):
LGTM

mlrun/config.py
assaf758, line 1127: please fix the docstring according to the funciton signature

mlrun/model_monitoring/helpers.py
```diff
@@ -63,7 +63,6 @@ def get_stream_path(
     )
 
     if not stream_uri or stream_uri == "v3io":
-        # TODO : remove the first part of this condition in 1.9.0
```
assaf758, line 66: you can also remove the `if isinstance(stream_uri, list)` if-clause few lines below

server/api/api/endpoints/nuclio.py
```diff
@@ -546,12 +546,20 @@ def _deploy_nuclio_runtime(
             )
 
         if serving_to_monitor:
-            if monitoring_deployment.should_redeploy_monitoring_stream(
-                fn_image=fn.spec.image, client_version=client_version
+            if (
+                fn.spec.image.startswith("mlrun/")
+                and client_version
```
assaf758: just to my understanding, do we really need to check for a case where client_version was not provided? I would think we should fail this request fast, and later in the code assume we have it.

server/api/__init__.py
```diff
@@ -12,3 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+
+# Minimum client version that supports model monitoring,
+# Will be fixed when MM will be defined as BC supported feature
+MINIMUM_CLIENT_VERSION_FOR_MM = "1.7.0"
```
assaf758, line 18: don't you prefer to define it in a more model-monitoring specific place like server/api/api/endpoints/model_monitoring.py?

server/api/crud/model_monitoring/deployment.py
```diff
@@ -288,54 +283,51 @@ def apply_and_create_stream_trigger(
...
-                        stream_path=stream_path,
-                        access_key=access_key,
-                        stream_args=stream_args,
-                    )
-                    # Generate V3IO stream trigger
-                    function.add_v3io_stream_trigger(
-                        stream_path=stream_path,
-                        name=f"monitoring_{function_name}_trigger{f'_{i}' if i != 0 else ''}",
-                        **kwargs,
+        if not mlrun.mlconf.is_ce_mode():
```
assaf758, line 330: I know it's not your relevant to your change, but still I would like to comment that it's better to fail here if it's ce_mode. Current code has 2 cons:
1. fails silently on an unsupported case
2. adds one more indentation which is less readable

server/api/__init__.py
davesh0812: We need this in few more places so when  I tried to put it in where I got some import issues.

assaf758 (approved):

PR #6306, opened 2024-09-10
[Alerts] Update alert name validation to allow only alphanumeric characters and hyphens

Yacouby:
resolves:

https://iguazio.atlassian.net/browse/ML-7753

### DIFF
--- a/server/api/crud/alerts.py
+++ b/server/api/crud/alerts.py
@@ -14,6 +14,7 @@
 #
 
 import datetime
+import re
 
 import sqlalchemy.orm
 
@@ -285,8 +286,8 @@ class Alerts(
 
         return False
 
-    @staticmethod
-    def _validate_alert(alert, name, project):
+    def _validate_alert(self, alert, name, project):
+        self.validate_alert_name(alert.name)
         if name != alert.name:
             raise mlrun.errors.MLRunBadRequestError(
                 f"Alert name mismatch for alert {name} for project {project}. Provided {alert.name}"
@@ -338,6 +339,13 @@ class Alerts(
         if alert.entities.project != project:
             raise mlrun.errors.MLRunBadRequestError(
                 f"Invalid alert entity project ({alert.entities.project}) for alert {name} for project {project}"
+            )
+
+    @staticmethod
+    def validate_alert_name(name: str) -> None:
+        if not re.fullmatch(r"^[a-zA-Z0-9-]+$", name):
+            raise mlrun.errors.MLRunBadRequestError(
+                f"Invalid alert name '{name}'. Alert names can only contain alphanumeric characters and hyphens."
             )
 
     @staticmethod
--- a/tests/api/crud/test_alerts.py
+++ b/tests/api/crud/test_alerts.py
@@ -13,6 +13,9 @@
 # limitations under the License.
 #
 
+from contextlib import AbstractContextManager
+from contextlib import nullcontext as does_not_raise
+
 import fastapi.concurrency
 import pytest
 import sqlalchemy.orm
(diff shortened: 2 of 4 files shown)


tests/api/crud/test_alerts.py
```diff
@@ -78,3 +78,64 @@ async def test_process_event_no_cache(
...
+@pytest.mark.parametrize(
+    "alert_name, should_pass",
+    [
+        ("invalid_name?", False),
+        ("name with spaces", False),
+        ("invalid/name", False),
+        ("invalid@name", False),
+        ("invalid_name", False),
+        ("valid-name", True),
+        ("valid-name-123", True),
```
TomerShor: Add a case that starts with an invalid character

tests/api/crud/test_alerts.py
```diff
@@ -78,3 +78,64 @@ async def test_process_event_no_cache(
...
+            server.api.crud.Alerts().store_alert(
+                db, project=project, name=alert_name, alert_data=alert_data
+            )
+        except mlrun.errors.MLRunBadRequestError:
+            pytest.fail("Expected the alert to pass validation but it failed.")
+    else:
+        with pytest.raises(mlrun.errors.MLRunBadRequestError):
+            server.api.crud.Alerts().store_alert(
+                db, project=project, name=alert_name, alert_data=alert_data
+            )
```
TomerShor: instead of `should_pass` + try-except, use `with expectation:`.
See how it done [here](https://github.com/mlrun/mlrun/blob/8c0a93206eefa7370159f09e5cd7579975e00506/tests/artifacts/test_artifacts.py#L392-L410) for example

tests/system/alerts/test_alerts.py
```diff
@@ -175,6 +175,11 @@ def _generate_events(
             system_performance_example,
         ]
 
+    @staticmethod
+    def _normalize_alert_name(alert_kind):
+        sanitized_name = alert_kind.value.replace("_", "-")
+        return sanitized_name
```
TomerShor: you can use `mlrun.utils.helpers.normalize_name` instead of creating a function here

tests/system/alerts/test_alerts.py
```diff
@@ -188,7 +193,8 @@ def _generate_alerts(
         ]
 
         for alert_kind in alerts_kind_to_test:
-            alert_name = f"drift_webhook_{alert_kind.value}"
+            normalized_kind = self._normalize_alert_name(alert_kind)
+            alert_name = f"drift-webhook-{normalized_kind}"
```
TomerShor: since the method is `normalize` name, it looks weird that you pass the kind.
You can do it all at once:
```suggestion
            alert_name = mlrun.utils.helpers.normalize_name(f"drift-webhook-{alert_kind}")
```

TomerShor (changes requested):
Logic LGTM, left some comments regarding the tests

server/api/crud/alerts.py
```diff
@@ -340,6 +341,14 @@ def _validate_alert(alert, name, project):
                 f"Invalid alert entity project ({alert.entities.project}) for alert {name} for project {project}"
             )
 
+    @staticmethod
+    def validate_alert_name(name):
```
moranbental:
```suggestion
    def validate_alert_name(name: str) -> None:
```

server/api/crud/alerts.py
```diff
@@ -340,6 +341,14 @@ def _validate_alert(alert, name, project):
...
+        pattern = re.compile(r"^[a-zA-Z0-9-]+$")
+        if not re.fullmatch(pattern, name):
```
moranbental: no need to compile the pattern separately if it’s only used once
```suggestion
        if not re.fullmatch(r"^[a-zA-Z0-9-]+$", name):
```

moranbental (changes requested):
Looks good!

TomerShor (approved):

moranbental (approved):

PR #6307, opened 2024-09-11
[Nuclio] Generate v3io stream trigger access key configuration

TomerShor:
To not expose the access key on the function spec, we will use the `$generate` keyword to let Nuclio populate the access key from the `"V3IO_ACCESS_KEY"` env var.

This is only supported from Nuclio 1.13.11 (see [this PR](https://github.com/nuclio/nuclio/pull/3350)).

Resolves https://iguazio.atlassian.net/browse/ML-7779


mlrun/runtimes/nuclio/function.py
```diff
@@ -1241,6 +1247,13 @@ def get_url(
 
         return self._resolve_invocation_url("", force_external_address)
 
+    @staticmethod
+    def _resolve_v3io_access_key():
+        # if nuclio version < 1.13.11 return None, otherwise return "$generate"
```
alonmr: I think this comment should explain why you do what you do instead of explaining the code which is pretty clear.

mlrun/runtimes/nuclio/function.py
```diff
@@ -495,6 +495,11 @@ def add_v3io_stream_trigger(
         extra_attributes = extra_attributes or {}
         if ack_window_size:
             extra_attributes["ackWindowSize"] = ack_window_size
+
+        access_key = kwargs.pop("access_key", None)
+        if not access_key:
+            access_key = self._resolve_v3io_access_key()
```
alonmr: IIUC if the user passes an access key it will still be exposed. can't we mask it?

TomerShor: Masking it in the trigger spec is a lot of work.
The access key should be passed in the env var that is already masked, and nuclio will read it from there.

I didn't want to deny the option to pass an access key, but it should be discouraged.

tests/api/runtimes/test_nuclio.py
```diff
@@ -234,38 +234,46 @@ def _assert_deploy_called_basic_config(
...
+            deepdiff.DeepDiff(
+                triggers_config["http"],
                 expected_struct,
                 ignore_order=True,
+                # TODO - (in Nuclio) There is a bug with canary configuration:
+                #        the nginx.ingress.kubernetes.io/canary-weight annotation gets assigned the host name
+                #        rather than the actual weight. Remove this once bug is fixed.
+                exclude_paths=[
+                    "root['annotations']['nginx.ingress.kubernetes.io/canary-weight']"
+                ],
```
rokatyy, line 252: Is there a jira ticket for this?

rokatyy (approved):
LGTM

tests/api/runtimes/test_nuclio.py
TomerShor: I'm actually not familiar with this bug at all, I just moved some code around that included this comment.
But i'll check it out before removing the comment

mlrun/runtimes/nuclio/function.py
alonmr: Since this is a security issue I actually think we should deny. Of course this breaks BC therefore it should be well documented.

TomerShor: If it breaks BC shouldn't we deprecate it first?

alonmr: There is a difference here because it's a serious bug. It shouldn't be considered as breaking BC

alonmr:
Reading more into the jira I see it is being masked in nuclio side so I believe the solution here should be different.
1. If access key isn't masked - do not include it in function yaml.
2. Once function is deployed, the function spec should be updated with the secret ref.

I don't think deprecating is the correct approach here because nuclio does mask it.
Still if we do deprecate, we need to mention versions end document it etc.

Edit:
This was discussed offline, currently we do not get the nuclio spec back after deploy

mlrun/runtimes/nuclio/function.py
```diff
@@ -495,6 +495,14 @@ def add_v3io_stream_trigger(
         extra_attributes = extra_attributes or {}
         if ack_window_size:
             extra_attributes["ackWindowSize"] = ack_window_size
+
+        access_key = kwargs.pop("access_key", None)
+        if access_key:
+            logger.warning(
+                "The access_key parameter is deprecated, use the V3IO_ACCESS_KEY environment variable instead"
+            )
```
alonmr: If access key is not given and nuclio version doesn't support generate then it will still appear in the dict.
What will happen if we remove it from the dict once nuclio created the trigger? will redeploy work?

TomerShor: > What will happen if we remove it from the dict once nuclio created the trigger? will redeploy work?

Nope, because it will take the new request body as the "desired" state.
The the access key will be missing and it fail to connect to the v3io stream.

mlrun/runtimes/nuclio/function.py
```diff
@@ -495,6 +495,14 @@ def add_v3io_stream_trigger(
         extra_attributes = extra_attributes or {}
         if ack_window_size:
             extra_attributes["ackWindowSize"] = ack_window_size
+
+        access_key = kwargs.pop("access_key", None)
+        if access_key:
+            logger.warning(
+                "The access_key parameter is deprecated, use the V3IO_ACCESS_KEY environment variable instead"
```
alonmr:
```suggestion
                "The access_key parameter is deprecated and will be ignored, use the V3IO_ACCESS_KEY environment variable instead"
```

alonmr (approved):

PR #6308, opened 2024-09-11
[Project] Fail sync functions silently on workflow runs

alonmr:
Sync functions reloads functions that are not necessarily related to the workflow, therefore it may fail on issues like file not found for functions that are not needed. This changes sync functions to fail silently when there is a failure to reload a function when trying to run a workflow. If a function fails to load and is later needed in the workflow, it shall be fetched from the DB or fail if it doesn't exist.

https://iguazio.atlassian.net/browse/ML-7706

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -2819,6 +2819,17 @@ class MlrunProject(ModelObj):
     ):
         """
         Reload function objects from specs and files.
+        The function objects are synced against the definitions spec in `self.spec._function_definitions`.
+        Referenced files/URLs in the function spec will be reloaded.
+        Function definitions are parsed by the following precedence:
+            1. Contains runtime spec.
+            2. Contains module in the project's context.
+            3. Contains path to function definition (yaml, DB, Hub).
+            4. Contains path to .ipynb or .py files.
+            5. Contains a Nuclio/Serving function image / an 'Application' kind definition.
+        If function definition is already an object, some project metadata updates will apply however,
+        it will not be reloaded.
+
         :param names:   Names of functions to reload, defaults to `self.spec._function_definitions.keys()`.
         :param always:  Force reloading the functions.
         :param save:    Whether to save the loaded functions or not.
@@ -2840,7 +2851,7 @@ class MlrunProject(ModelObj):
             if not function_definition:
                 if silent:
                     logger.warn(
-                        f"Function {name} definition was not found, skipping reload."
+                        "Function definition was not found, skipping reload", name=name
                     )
                     continue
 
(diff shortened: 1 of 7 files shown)


mlrun/projects/project.py
```diff
@@ -2810,47 +2810,81 @@ def push(
...
+                        message += f" Skipping {name} function reload."
+                        logger.warn(message)
+                        continue
+
+                    raise mlrun.errors.MLRunMissingDependencyError(message) from exc
+            else:
+                message = f"Function {name} must be an object or dict."
+                if silent:
+                    message += f" Skipping {name} function reload."
+                    logger.warn(message)
```
liranbg: why not abstract the log and move parameters to kwarg? is it possible? I know this warning is for clients but still we need to be consistent with log structure as much as possible

mlrun/projects/project.py
```diff
@@ -2810,47 +2810,81 @@ def push(
...
-        """reload function objects from specs and files"""
+    def sync_functions(
+        self,
+        names: list = None,
+        always: bool = True,
+        save: bool = False,
+        silent: bool = False,
+    ):
+        """
+        Reload function objects from specs and files.
```
liranbg, line 2821: worth adding _what_ does it sync and against what (and precedence, etc)

server/api/db/sqldb/db.py
```diff
@@ -4637,7 +4637,7 @@ def _delete(self, session, cls, **kw):
             session.delete(obj)
         session.commit()
 
-    def _find_lables(self, session, cls, label_cls, labels):
+    def _find_labels(self, session, cls, label_cls, labels):
```
liranbg, line 4640: 👏🏼

liranbg (changes requested):
minor

mlrun/projects/project.py
```diff
@@ -2810,47 +2810,92 @@ def push(
...
-                raise ValueError(f"function named {name} not found")
+            function_definition = self.spec._function_definitions.get(name)
+            if not function_definition:
+                if silent:
+                    logger.warn(
+                        "Function definition was not found, skipping reload", name=name
+                    )
+                    continue
+
+                raise ValueError(f"Function named {name} not found")
```
rokatyy, line 2858:
```suggestion
            if not function_definition:
                if not silent:
                    raise ValueError(f"Function named {name} not found")
                logger.warn(
                    "Function definition was not found, skipping reload", name=name
                )
```

mlrun/projects/project.py
```diff
@@ -2810,47 +2810,92 @@ def push(
...
-                    raise mlrun.errors.MLRunMissingDependencyError(
-                        f"File {exc.filename} not found while syncing project functions"
-                    ) from exc
+                    message = f"File {exc.filename} not found while syncing project functions."
+                    if silent:
+                        message += " Skipping function reload"
+                        logger.warn(message, name=name)
+                        continue
+
+                    raise mlrun.errors.MLRunMissingDependencyError(message) from exc
```
rokatyy, line 2884: same

mlrun/projects/project.py
```diff
@@ -2810,47 +2810,92 @@ def push(
...
+            else:
+                message = f"Function {name} must be an object or dict."
+                if silent:
+                    message += " Skipping function reload"
+                    logger.warn(message, name=name)
+                    continue
+                raise ValueError(message)
```
rokatyy, line 2891: same

rokatyy (commented):
Minor

mlrun/projects/project.py
alonmr: actually I still need to continue because I need to move on to the next function so I kept it as is

rokatyy (approved):

PR #6310, opened 2024-09-11
[Requirements] CEML-233: Bump boto packages to 1.35.16

alexpirogovski:


extras-requirements.txt
```diff
@@ -13,9 +13,9 @@
 s3fs>=2023.9.2, <2024.7
 gcsfs>=2023.9.2, <2024.7
 ossfs==2023.12.0
-boto3>=1.28.0,<1.29.0
+boto3>=1.35.16
```
liranbg, line 16:
```suggestion
boto3~=1.35.16
```
and do it across the its use. we do not give minimum only but either < or ~

extras-requirements.txt
```diff
@@ -13,9 +13,9 @@
...
 oss2==2.18.1
-aiobotocore>=2.5.0,<2.8
+aiobotocore>=2.15.0
```
liranbg, line 18:
```suggestion
aiobotocore~=2.15.0
```

liranbg (changes requested):
youre missing changing https://github.com/mlrun/mlrun/blob/6c420e47aa190fa931710df50891f0eff3cabd81/dependencies.py#L33 as well

assaf758:
@alexpirogovski what tests did you run?
@tomerm-iguazio can you please specify tests (upload etc) that excersize boto

automation/requirements.txt
```diff
@@ -2,7 +2,7 @@ click~=8.1
 paramiko~=3.4
 semver~=3.0
 requests~=2.31
-boto3>=1.28.0,<1.29.0
+boto3>=1.28.0,<=1.35.16
```
tomerm-iguazio, line 5: s3 functions that directly use boto3:
get, upload, put, stat and listdir.

Also s3fs depends on boto3, so I think It will be the best to run all s3 tests:

integration: `mlrun/tests/integration/aws_s3/test_aws_s3.py`
system tests: `mlrun/tests/system/datastore/test_aws_s3.py`

running spark s3 tests are not required for this change.

PR #6313, opened 2024-09-11
[Project] Improve `set_function` docstring

alonmr:
The reason I added `|` at new lines is to break them in sphinx docs

https://iguazio.atlassian.net/browse/ML-5686

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -2341,7 +2341,7 @@ class MlrunProject(ModelObj):
         | Creating a function from a single file is done by specifying ``func`` and disabling ``with_repo``.
         | Creating a function with project source:
         |   1. Specifying a relative ``func`` path and with_repo=True.
-        |   2. Specifying a module ``handler`` (e.g. ``handler=package.package.func``).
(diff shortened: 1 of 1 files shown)


mlrun/projects/project.py
```diff
@@ -2334,16 +2334,24 @@ def set_function(
...
-        """update or add a function object to the project
+        """
+        | Update or add a function object to the project.
+        | Function can be provided as an object (func) or a .py/.ipynb/.yaml URL.
 
-        function can be provided as an object (func) or a .py/.ipynb/.yaml url
-        support url prefixes::
+        | Creating a function from a single file is done by specifying ``func`` and disabling ``with_repo``.
+        | Creating a function with project source:
+        |   1. Specifying a relative ``func`` path and with_repo=True.
```
jillnogold: |   1. Specify a relative ``func`` path and with_repo=True.
        |   2. Specify a module ``handler`` (e.g. ``handler=package.package.func``) without ``func``.

theSaarco (commented):
Looks fine,  suggest to add examples in the examples section to the new options you documented, for the various options of working with project source, and the option of using `with_source_archive`.

liranbg (approved):

PR #6315, opened 2024-09-12
[Model Monitoring] Store `result_extra_data` in TSDB

Eyal-Danieli:
Implement the capability to store additional data (aka `result_extra_data`) from model monitoring application results within the TSDB, with a limit of approximately 1KB to prevent storage issues. If the extra data exceeds this limit, it will not be stored in the TSDB. Instead, the application will log a message notifying the user to either reduce the size of the extra data or save it as an artifact.

Additionally, this PR removes the `current_stats` from the result table in the KV/SQL store to prevent potential storage issues, as there are no current consumers for these values. 

Related JIRA: https://iguazio.atlassian.net/browse/ML-7460


mlrun/model_monitoring/applications/results.py
```diff
@@ -72,8 +81,21 @@ def to_dict(self):
             mm_constant.ResultData.RESULT_EXTRA_DATA: json.dumps(self.extra_data),
         }
 
+    @validator("extra_data")
+    @classmethod
+    def validate_extra_data_len(cls, result_extra_data: dict):
+        extra_data_len = len(json.dumps(result_extra_data))
+        if extra_data_len > _RESULT_EXTRA_DATA_MAX_SIZE:
+            logger.info(
```
davesh0812:
```suggestion
            logger.warnning(
```

mlrun/model_monitoring/applications/results.py
```diff
@@ -72,8 +81,21 @@ def to_dict(self):
             mm_constant.ResultData.RESULT_EXTRA_DATA: json.dumps(self.extra_data),
         }
 
+    @validator("extra_data")
+    @classmethod
+    def validate_extra_data_len(cls, result_extra_data: dict):
```
davesh0812, line 86: please add short explain

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -149,11 +150,12 @@ def read_metrics_data(
...
-        :param start:       The start time of the query.
-        :param end:         The end time of the query.
-        :param metrics:     The list of metrics to get the values for.
-        :param type:        "metrics" or "results" - the type of each item in metrics.
-        :return:            A list of result values or a list of metric values.
+        :param start:                  The start time of the query.
+        :param end:                    The end time of the query.
+        :param metrics:                The list of metrics to get the values for.
+        :param type:                   "metrics" or "results" - the type of each item in metrics.
+        :param with_result_extra_data: Whether to include the extra data in the results.
```
davesh0812:
```suggestion
        :param with_result_extra_data: Whether to include the extra data in the results, relevant only when type="result".
```

mlrun/model_monitoring/applications/histogram_data_drift.py
```diff
@@ -189,21 +189,24 @@ def _get_general_drift_result(
...
-                EventFieldType.CURRENT_STATS: json.dumps(
-                    monitoring_context.sample_df_stats
-                ),
-                EventFieldType.DRIFT_MEASURES: json.dumps(
-                    metrics_per_feature.T.to_dict()
-                    | {metric.name: metric.value for metric in metrics}
-                ),
-                EventFieldType.DRIFT_STATUS: status.value,
-            },
+            extra_data=model_endpoint_stats,
```
davesh0812: So we will save it to the tsdb by default? if it not to big...

mlrun/model_monitoring/applications/histogram_data_drift.py
```diff
@@ -189,21 +189,24 @@ def _get_general_drift_result(
...
+
+        model_endpoint_stats = {
+            EventFieldType.CURRENT_STATS: json.dumps(
+                monitoring_context.sample_df_stats
+            ),
+            EventFieldType.DRIFT_MEASURES: json.dumps(
+                metrics_per_feature.T.to_dict()
+                | {metric.name: metric.value for metric in metrics}
+            ),
+            EventFieldType.DRIFT_STATUS: status.value,
```
davesh0812: can be removed from my point of view

```suggestion
```

mlrun/model_monitoring/applications/histogram_data_drift.py
```diff
@@ -189,21 +189,24 @@ def _get_general_drift_result(
...
 
         status = self._value_classifier.value_to_status(value)
+
+        model_endpoint_stats = {
+            EventFieldType.CURRENT_STATS: json.dumps(
+                monitoring_context.sample_df_stats
+            ),
+            EventFieldType.DRIFT_MEASURES: json.dumps(
+                metrics_per_feature.T.to_dict()
+                | {metric.name: metric.value for metric in metrics}
```
davesh0812: If I understand correctly, this line represents the mean value for each metric. I think you can remove it here, as we’re already creating MM metrics for these values.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -422,7 +433,11 @@ def read_metrics_data(
             is_empty=df.empty,
         )
 
-        return df_handler(df=df, metrics=metrics, project=self.project)
+        df_handler_kwargs["df"] = df
+
```
davesh0812, line 436:
```suggestion
if not mm_schemas.ResultData.RESULT_EXTRA_DATA not in df.columns and type="result":
     df[mm_schemas.ResultData.RESULT_EXTRA_DATA] = ""
```

If you will use my suggestion, you can remove the new parameter from `df_to_results_values` methoods and revert the df_handler_kwargs changes, which I believe will help simplify the code.

of course it can be done in v3io connector as well

davesh0812 (commented):
Looks Good!
few comments :)

mlrun/model_monitoring/applications/histogram_data_drift.py
Eyal-Danieli: you right, I'll remove it from the hist data drift

Eyal-Danieli: and as a result we don't need both `DRIFT_STATUS` & `DRIFT_MEASURES` (your next 2 comments) at the moment. It will be calculated and stored according [to @royischoss ticket. ](https://iguazio.atlassian.net/browse/ML-7461)

davesh0812 (approved):

PR #6317, opened 2024-09-12
[Config] Configure ssl verification

moranbental:
Disable certificate warnings when `mlconf.httpdb.http.verify` is `False`

We can set this value using the environment variable:
`os.environ["MLRUN_HTTPDB__HTTP__VERIFY"] = "true"`

In the next PR, the default value for `mlconf.httpdb.http.verify` will be set to `True` to validate certificates by default.

### DIFF
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -326,7 +326,7 @@ default_config = {
         "http": {
             # when True, the client will verify the server's TLS
             # set to False for backwards compatibility.
-            "verify": True,
+            "verify": False,
         },
         "db": {
             "commit_retry_timeout": 30,
@@ -1448,6 +1448,9 @@ def read_env(env=None, prefix=env_prefix):
     # The default function pod resource values are of type str; however, when reading from environment variable numbers,
     # it converts them to type int if contains only number, so we want to convert them to str.
     _convert_resources_to_str(config)
+
+    if verify_ssl := config.get("httpdb", {}).get("http", {}).get("verify", None):
+        config["httpdb"]["http"]["verify"] = strtobool(verify_ssl)
     return config
 
 
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -326,7 +326,7 @@ default_config = {
         "http": {
             # when True, the client will verify the server's TLS
             # set to False for backwards compatibility.
-            "verify": False,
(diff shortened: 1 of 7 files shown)


mlrun/db/factory.py
```diff
@@ -59,9 +61,15 @@ def create_run_db(self, url="", secrets=None, force_reconnect=False):
             #  The SQLRunDB should always get its session from the FastAPI dependency injection.
             self._run_db = self._rundb_container.run_db(url)
 
+        self._verify_ssl()
```
TomerShor: why is it called here in the factory?

mlrun/db/factory.py
```diff
@@ -59,9 +61,15 @@ def create_run_db(self, url="", secrets=None, force_reconnect=False):
...
         self._run_db.connect(secrets=secrets)
         return self._run_db
 
+    @staticmethod
+    def _verify_ssl() -> None:
```
liranbg: you dont verify ssl here, you disable warnings upon user's preferences.

I think the functions logic can be moved to `read_env` (I know @quaark had some objections about it but this is the current place to put config -> logic functionality) and it will fix the scenario where user change the config during runtime

https://github.com/mlrun/mlrun/blob/d444546f8a3f02d47987c0188b607b86c03cabb6/mlrun/config.py#L1429-L1434

liranbg (changes requested):

mlrun/config.py
```diff
@@ -1351,6 +1354,14 @@ def _convert_str(value, typ):
     return typ(value)
 
 
+def _configure_ssl_verification(verify_ssl: bool) -> None:
+    """Configure SSL verification warnings based on the setting."""
+    if not verify_ssl:
+        urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+    else:
+        warnings.simplefilter("default", urllib3.exceptions.InsecureRequestWarning)
```
liranbg: needless

moranbental: If the default is False and it is changed to True at runtime, we need to enable this warning if using `set_env_from_file` after importing mlrun

alonmr: please add a code comment to explain it

PR #6318, opened 2024-09-12
[Docs] Add details about invalid node selector combinations

jillnogold:
[ML-7568](https://iguazio.atlassian.net/browse/ML-7568)

[ML-7568]: https://iguazio.atlassian.net/browse/ML-7568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/runtimes/configuring-job-resources.md
```diff
@@ -365,7 +365,12 @@ The zone label from the project level is completely removed, and the resulting c
 ```
 {"gpu": "true", "arch": "amd64"}
 ```
+### Preventing conflicts
 
+If you run a function and it is stuck with the status `pending` it's possible that the "specified" node selector does not exist. 
+Remember, the project and function level configurations override the service level and the function itself. Check the function.yaml and the 
+project.yaml for the node selector configurations. And check the UI (Projects > Jobs and workflows > Monitor jobs)
+for the node selector that was ultimately defined by all of the configurations and overrides.
```
yaelgen: Great! However, since this can be quite a confusing situation for users, I feel it might help to be a bit more detailed and specific. I've added a suggestion below with what I had in mind. Of course, feel free to modify it in any way you think is best:
```suggestion
If your function run is stuck with the status `pending`, it's possible that the "specified" node selector does not exist. There are three levels of node selectors in MLRun: function, project, and service. At runtime, the system combines these selectors and applies the resolved configuration to the pod.

How to Investigate:
1. Check the Configuration Files:
Look in the function.yaml and project.yaml files to see if there are any node selector settings.

2. Review Node Selectors in the UI:
Go to Projects > Jobs and Workflows > Monitor Jobs > Overview > Node Selector. This shows the node selector that was ultimately defined for the run after combining the function, project, and service settings.

3. Check Pod Errors in the UI:
Go to Projects > Jobs and Workflows > Monitor Job > Pods, you can see the pod details. If no nodes are found that match the specified node selector, the error will be displayed here.


Resolving Conflicts:
If the node selectors from the function, project, or service levels conflict or result in an impossible combination, you can resolve the issue by specifying the conflicting node selector key with an empty string value on your function. Be cautious with this approach and consult your project admin before making changes to ensure it won’t cause other issues.
```

yaelgen (changes requested):
Looks great, added a suggestion

yaelgen (approved):
🫶🏽

yaelgen (approved):

yaelgen (approved):

docs/runtimes/configuring-job-resources.md
```diff
@@ -397,9 +413,10 @@ This logic becomes part of the Spark CRD, ensuring that it is consistently appli
 ### Best Practice
 
 Node selection is often used for assigning jobs/pods to GPU nodes. But not all jobs/pods benefit from a GPU node.
-For example, a Databricks “helper” pod runs in a Spark service on Databricks and doesn’t follow the node-selector 
+For example:
+- A Databricks “helper” pod runs in a Spark service on Databricks and doesn’t follow the node-selector
```
alonmr: is this true @yaelgen / @tomerm-iguazio  ? databricks helper pod is a kubejob so maybe this is meant for the actual databricks job?

docs/runtimes/configuring-job-resources.md
```diff
@@ -365,7 +365,23 @@ The zone label from the project level is completely removed, and the resulting c
 ```
 {"gpu": "true", "arch": "amd64"}
 ```
+### Preventing conflicts
 
+If your function run is stuck with the status `pending`, it's possible that the "specified" node selector does not exist. There are three 
+levels of node selectors in MLRun: function, project, and service. At runtime, the system combines these selectors and applies the resolved
```
alonmr, line 371: we have 4 levels of node selector: run, function, project, service

alonmr (commented):
Looks good! Minor comments

docs/runtimes/configuring-job-resources.md
tomerm-iguazio: I think that the "helper pod" is the pod that run in our kubernetes and communicate with databricks (been used as a client only).
we do not run spark on the pod itself.

The job that run in databricks is running on a cluster that can be new or existing one according to user configuration.
There we can run spark:)

About the Node selector, what did you configure on databricks helper pod? @yaelgen

docs/runtimes/configuring-job-resources.md
yaelgen: I think this can be a little confusing for the users, since in most cases the run object is created at runtime from the function, and it's not something they are editing. The run level is only relevant for tasks and remote workflows and for the workflow runner, so If we are adding this here as part of the documentation we need to make sure it's clear enough

docs/runtimes/configuring-job-resources.md
yaelgen: looks like there might be a mix up between Spark and Databricks here, that I'm not sure I understand.
For Databricks, the node selector is only relevant for the "helper" pod running in our k8s cluster, as Tomer mentioned. This behaves similarly to how node selectors are applied in Kubejob. It does not affect the actual Databricks cluster, which is not based on Kubernetes and does not run in our cluster, so node selectors have no significance in that context.. maybe this is what this line meant by "doesn’t follow the node-selector"?

docs/runtimes/configuring-job-resources.md
jillnogold: @yaelgen how does the "run" node selector fit into the hierarchy of the other 3?

docs/runtimes/configuring-job-resources.md
jillnogold: @yaelgen @tomerm-iguazio @alonmr 
- Updated about Databricks.
- Does it need more explanation about Spark?
- Can you give more examples of where GPU node is "wasted"?

docs/runtimes/configuring-job-resources.md
alonmr: let's keep it as is for now

docs/runtimes/configuring-job-resources.md
```diff
@@ -392,14 +409,16 @@ a merge with precedence is performed with the project and MLRun config levels.
...
 ### Best Practice
 
 Node selection is often used for assigning jobs/pods to GPU nodes. But not all jobs/pods benefit from a GPU node.
-For example, a Databricks “helper” pod runs in a Spark service on Databricks and doesn’t follow the node-selector 
-(and doesn't benefit from being assigned to a GPU node). 
-A Spark Function includes an executor and a driver; the driver also does not benefit from a GPU node.
+For example:
+- With Databricks, the node selector is only relevant for the "helper" pod running in the MLRun Kubernetes cluster,  
+and it behaves similarly to how node selectors are applied in Kubejob. It does not affect the actual Databricks cluster, 
+which is not based on Kubernetes and does not run in our cluster: node selectors have no significance in that context.
```
alonmr: I don't think that's true. I would remove it
```suggestion
which does not run in our cluster: node selectors have no significance in that context.
```

yaelgen: It is true, but I'm not sure it's relevant

alonmr (approved):

PR #6324, opened 2024-09-12
[gRPC] Updating grpc tools for schema compilation

liranbg:
nothing specific to mention here, bumping deps and remedying vulnerabilities


liranbg:
blocked until we bump protobuf version on mlrun 1.8

PR #6326, opened 2024-09-15
[RuntimeResources] Enhance resources deletion upon project deletion

liranbg:
In a nutshell, upon project deletion - do not wait the entire runtime grace period but perform force removal of each runtime resources (specifically k8s one), that way, project deletion would occur faster

https://iguazio.atlassian.net/browse/ML-7803

In addition, removed repeated code of grace period initialization to a single place where needed

### DIFF
--- a/server/api/runtime_handlers/base.py
+++ b/server/api/runtime_handlers/base.py
@@ -125,7 +125,7 @@ class BaseRuntimeHandler(ABC):
         # 'force' is used to skip waiting X seconds *after* pod terminated.
         # 'grace_period' is used to set the time to wait *after* the pod terminated and before it's deleted.
         # if force is True and grace period is 0, simply delete the pod without waiting.
-        resource_deletion_grace_period = 0 if True and grace_period == 0 else None
+        resource_deletion_grace_period = 0 if force and grace_period == 0 else None
         # We currently don't support removing runtime resources in non k8s env
         if not server.api.utils.singletons.k8s.get_k8s_helper().is_running_inside_kubernetes_cluster():
             return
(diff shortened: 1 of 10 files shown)


server/api/runtime_handlers/base.py
```diff
@@ -115,16 +116,30 @@ def delete_resources(
...
+        grace_period: typing.Optional[int] = None,
     ):
         if grace_period is None:
             grace_period = config.runtime_resources_deletion_grace_period
+
+        # set resource deletion grace period to 0 when explicitly requested by force and grace period.
+        # 'force' is used to skip waiting X seconds *after* pod terminated.
+        # 'grace_period' is used to set the time to wait *after* the pod terminated and before it's deleted.
+        # if force is True and grace period is 0, simply delete the pod without waiting.
+        resource_deletion_grace_period = 0 if True and grace_period == 0 else None
```
alonmr:
```suggestion
        resource_deletion_grace_period = 0 if force and grace_period == 0 else None
```

server/api/runtime_handlers/base.py
```diff
@@ -1059,7 +1072,9 @@ def _delete_pod_resources(
                         )
 
                 server.api.utils.singletons.k8s.get_k8s_helper().delete_pod(
-                    pod.metadata.name, namespace
+                    pod.metadata.name,
+                    namespace,
+                    grace_period_seconds=resource_deletion_grace_period,
```
alonmr, line 1077: databricks jobs have a cancel task and if you force deletion here I believe it will not run. It needs to cancel the remote run so we might need to exclude databricks runtime here (see `termination_grace_period_seconds`)

alonmr (changes requested):
Looks very good! Couple of comments

alonmr (approved):

PR #6327, opened 2024-09-15
[Automation] Patch igz to support specific registry-url

liranbg:
Required when trying to login to specific register different than dockerhub
in addition, moved login to happen first, to avoid auth failures post build

### DIFF
--- a/automation/patch_igz/patch_remote.py
+++ b/automation/patch_igz/patch_remote.py
@@ -55,6 +55,7 @@ class MLRunPatcher:
         self._nodes = nodes
 
     def patch_mlrun_api(self):
+        self._docker_login_if_configured()
         vers = self._get_current_version()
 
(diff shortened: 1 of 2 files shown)


automation/patch_igz/patch_remote.py
```diff
@@ -119,17 +118,23 @@ def patch_mlrun_api(self):
...
-                [
-                    "docker",
-                    "login",
-                    "--username",
-                    registry_username,
-                    "--password",
-                    registry_password,
-                ],
-                live=True,
+        registry_url = self._config.get("REGISTRY_URL")
```
TomerShor, line 121: Add `REGISTRY_URL` to `patch_env_template.yml` with a proper comment.

TomerShor (changes requested):

PR #6331, opened 2024-09-15
[Artifacts] Fix the docstring for uploading an artifact when logging a directory artifact

moranbental:
When logging a directory artifact, the directory is not uploaded by default unless the `upload` flag is set to true
https://iguazio.atlassian.net/browse/ML-6064

### DIFF
--- a/mlrun/execution.py
+++ b/mlrun/execution.py
@@ -634,8 +634,9 @@ class MLClientCtx:
         :param viewer:        Kubeflow viewer type
         :param target_path:   Absolute target path (instead of using artifact_path + local_path)
         :param src_path:      Deprecated, use local_path
-        :param upload:        Upload to datastore (default is True). If logging a directory, it will not be
-                              uploaded unless this flag is set to `True`.
+        :param upload:        Whether to upload the artifact to the datastore. If not provided, and the `local_path`
+                              is not a directory, upload occurs by default. Directories are uploaded only when this
+                              flag is explicitly set to `True`.
         :param labels:        A set of key/value labels to tag the artifact with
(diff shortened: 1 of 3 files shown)


mlrun/execution.py
```diff
@@ -634,7 +634,8 @@ def log_artifact(
         :param viewer:        Kubeflow viewer type
         :param target_path:   Absolute target path (instead of using artifact_path + local_path)
         :param src_path:      Deprecated, use local_path
-        :param upload:        Upload to datastore (default is True)
+        :param upload:        Upload to datastore (default is True). If logging a directory, it will not be
+                              uploaded unless this flag is set to `True`.
```
yaelgen: I see the default is set to `None` not `True`, is this intentional: `(default is True)`?

yaelgen (changes requested):

mlrun/execution.py
moranbental: > I see the default is set to `None` not `True`, is this intentional: `(default is True)`?

Yes, because in the implementation, we upload by default unless it is a directory artifact
https://github.com/mlrun/mlrun/blob/a192b22dcaa51d64ba1248f30b8bb1f8b3df8384/mlrun/artifacts/manager.py#L302C9-L302C10

yaelgen: okay, I think this could be a bit confusing since the default value is `None`, but the behavior acts as if it's `True` when not specified. 
How about: 
`:param upload:        Upload to datastore. If not provided, and the target is not a directory, upload occurs by default. Directories are uploaded only when this flag is explicitly set to `True`.`
or something like that?

yaelgen (approved):
🔥

liranbg (approved):

PR #6332, opened 2024-09-15
[Tests] Do not verify SSL by default for system tests

moranbental:
related pr - https://github.com/mlrun/mlrun/pull/6323


alonmr (commented):
Why not add it to env.yml instead? IIRC our CI has an SSL certificate

moranbental:
> Why not add it to env.yml instead? IIRC our CI has an SSL certificate

I tried that, but it appears we are reading environment variables and updating the config at this line: https://github.com/mlrun/mlrun/blob/f07297b758eca327ecdc4a92a592c831d2ac229a/mlrun/config.py#L1290
Then, when we update dbpath, the first connection is made here: https://github.com/mlrun/mlrun/blob/f07297b758eca327ecdc4a92a592c831d2ac229a/mlrun/config.py#L1104
 As a result, the connection used the default value because the verify setting had not been updated yet

tests/system/env-template.yml
```diff
@@ -81,4 +81,9 @@ SNOWFLAKE_WAREHOUSE:
...
 MLRUN_MODEL_ENDPOINT_MONITORING__ENDPOINT_STORE_CONNECTION:
 MLRUN_MODEL_ENDPOINT_MONITORING__TSDB_CONNECTION:
-MLRUN_MODEL_ENDPOINT_MONITORING__STREAM_CONNECTION:
\ No newline at end of file
+MLRUN_MODEL_ENDPOINT_MONITORING__STREAM_CONNECTION:
+
+# Controls whether the client verifies the server's TLS certificate.
+# Set to `True` to enable TLS verification.
+# set to `False` when running in a lab environment where TLS verification may be disabled.
+MLRUN_HTTPDB__HTTP__VERIFY:
```
liranbg:
```suggestion
MLRUN_HTTPDB__HTTP__VERIFY:

```

tests/system/base.py
```diff
@@ -219,20 +218,36 @@ def _get_env_from_file(cls) -> dict:
...
+    def _process_env_var(cls, key, value):
+        if key in os.environ:
+            # Save old env vars for returning them on teardown
+            cls._old_env[key] = os.environ[key]
+
+        # Set the environment variable
+        if isinstance(value, bool):
+            os.environ[key] = "true" if value else "false"
+        elif value is not None:
+            os.environ[key] = value
```
liranbg: move below callee

mlrun/config.py
```diff
@@ -779,8 +779,8 @@ def __getattr__(self, attr):
         return val
 
     def __setattr__(self, attr, value):
-        # in order for the dbpath setter to work
-        if attr == "dbpath":
+        # in order for the dbpath and verify setters to work
+        if attr in {"dbpath", "verify"}:
```
liranbg: what calls mlconfig.verify now? uncertain why needed then

mlrun/config.py
```diff
@@ -1103,6 +1103,15 @@ def dbpath(self, value):
...
             mlrun.db.get_run_db(value, force_reconnect=True)
 
+    @property
+    def verify(self) -> bool:
+        return self._http_verify_ssl
+
+    @verify.setter
+    def verify(self, value: bool):
+        self._http_verify_ssl = value
+        _configure_ssl_verification(value)
```
liranbg: can u remind why needed?

liranbg (changes requested):

mlrun/config.py
moranbental: When updating the config at this point:
https://github.com/mlrun/mlrun/blob/0e9a5c3e60c96421b43d47dcf560784802c259f0/mlrun/config.py#L1299
the changes will trigger a call to:
https://github.com/mlrun/mlrun/blob/0e9a5c3e60c96421b43d47dcf560784802c259f0/mlrun/config.py#L802

mlrun/config.py
moranbental: Added a comment

PR #6333, opened 2024-09-15
[Workflows] Add support for workflow runner node selector in remote workflow execution

yaelgen:
resolves: [ML-7796](https://iguazio.atlassian.net/browse/ML-7796)

This PR introduces support for the `workflow_runner_node_selector` parameter, allowing users to define a specific node selector for the workflow runner pod when executing workflows with the remote engine.

[ML-7796]: https://iguazio.atlassian.net/browse/ML-7796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


liranbg (approved):
neat

mlrun/common/schemas/workflow.py
```diff
@@ -32,6 +32,7 @@ class WorkflowSpec(pydantic.BaseModel):
     schedule: typing.Union[str, ScheduleCronTrigger] = None
     run_local: typing.Optional[bool] = None
     image: typing.Optional[str] = None
+    workflow_runner_node_selector: typing.Optional[dict] = None
```
theSaarco:
```suggestion
    workflow_runner_node_selector: typing.Optional[dict[str, str]] = None
```
Let's be consistent across the board.

mlrun/projects/project.py
```diff
@@ -3060,6 +3060,7 @@ def run(
         source: str = None,
         cleanup_ttl: int = None,
         notifications: list[mlrun.model.Notification] = None,
+        workflow_runner_node_selector: dict[str, str] = None,
```
```suggestion
        workflow_runner_node_selector: typing.Optional[dict[str, str]] = None
```

mlrun/projects/project.py
```diff
@@ -3162,6 +3167,16 @@ def run(
             )
             inner_engine = get_workflow_engine(engine_kind, local).engine
         workflow_spec.engine = inner_engine or workflow_engine.engine
+        if workflow_engine.engine == "remote" and workflow_runner_node_selector:
```
theSaarco: Logically would be cleaner to do:
```
if workflow_runner_node_selector:
  if work_flow_engine.engine == "remote":
    ...
  else:
    ...
```

server/api/crud/workflows.py
```diff
@@ -363,6 +367,10 @@ def _prepare_run_object_for_single_run(
                     subpath=project.spec.subpath,
                 )
             )
+            if workflow_spec.workflow_runner_node_selector:
+                run_object.spec.node_selector.update(
```
theSaarco: Just curious - why are you using `update` here, and setting (using `deepcopy`) in all other places? Is there a specific reason for that, and if there is then why not do the same in all places?

theSaarco (changes requested):
Looks good overall, request some minor changes.

mlrun/projects/project.py
```diff
@@ -3060,6 +3060,7 @@ def run(
         source: str = None,
         cleanup_ttl: int = None,
         notifications: list[mlrun.model.Notification] = None,
+        workflow_runner_node_selector: typing.Optional[dict[str, str]] = None,
```
alonmr, line 3063: This should be part of `project.set_workflow` and not here. We need to be consistent with the way we work, see how we implemented setting the workflow runner image.

mlrun/projects/project.py
```diff
@@ -3096,7 +3097,11 @@ def run(
                           workflow and all its resources are deleted)
         :param notifications:
                           List of notifications to send for workflow completion
-
+        :param workflow_runner_node_selector:
+                          Defines the node selector for the workflow runner pod when using a remote engine.
+                          This allows you to control and specify where the workflow runner pod will be scheduled.
+                          This setting is only relevant when the engine is set to 'remote',
```
```suggestion
                          This setting is only relevant when the engine is set to 'remote' or for scheduled workflows.
```

server/api/crud/workflows.py
```diff
@@ -97,6 +98,7 @@ def schedule(
             workflow_request=workflow_request,
             labels=labels,
         )
+        runner.spec.node_selector = deepcopy(run_spec.spec.node_selector)
```
alonmr: I think we should either apply it on the runner or on the run. I don't see a reason to set it on both. probably having it on the run is better to make the function easy to export?

server/api/crud/workflows.py
```diff
@@ -230,6 +233,7 @@ def run(
         )
 
         artifact_path = workflow_request.artifact_path if workflow_request else ""
+        runner.spec.node_selector = deepcopy(run_spec.spec.node_selector)
```
alonmr: same here

tests/system/projects/test_project.py
```diff
@@ -595,6 +595,17 @@ def test_kfp_pipeline_with_resource_param_passed(self):
         function = project.get_function("func-1", ignore_cache=True)
         assert function.spec.resources["requests"]["memory"] == arguments["memory"]
 
+    def _load_remote_pipeline_project(self, name):
```
alonmr: please move private methods to the bottom

tests/system/projects/test_project.py
```diff
@@ -641,6 +645,53 @@ def _test_remote_pipeline_from_github(
...
+            workflow_runner_node_selector=runner_node_selector,
+        )
+        runner_run_result = project.list_runs(name=workflow_runner_name)[0]
+        assert runner_run_result["spec"]["node_selector"] == {
+            **project_default_function_node_selector,
+            **runner_node_selector,
+        }
+
+        # Test scheduled workflow
+        schedule = "*/30 * * * *"
```
alonmr: I think you need some schedule crontab that will never run so we won't have the schedule invoked twice by accident

alonmr (changes requested):
Looks very good! Some comments

mlrun/projects/project.py
alonmr: discussed offline

rokatyy (approved):
LGTM

theSaarco (approved):
LGTM

yaelgen:
Documentation Note:
We’ve decided to retain the workflow runner’s node selector on the runner/function object for two main reasons:

- User Visibility and Editability: When a user creates a schedule, the function object for the remote workflow runner is created and made available prior to the scheduled execution. Storing the node selector solely in the run object could lead to confusion regarding visibility, and it would prevent the user from modifying or adjusting it after the function has been created.

- Enrichment Process Considerations: Currently, the launcher enrichment process only takes the runtime object's node selector into account, ignoring the run spec’s node selector. While this may also be relevant for tasks, we’ve opted to maintain the node selector on the function object for now. We plan to revisit this in a separate discussions, but for the time being, this approach provides the clearest and most consistent handling.

mlrun/projects/project.py
```diff
@@ -3162,6 +3167,18 @@ def run(
...
+                workflow_spec.workflow_runner_node_selector = (
+                    workflow_runner_node_selector
+                )
+            else:
+                warnings.warn(
+                    "The 'workflow_runner_node_selector' is only applicable when using a remote engine. "
+                    "It defines the node selector for the workflow runner pod, which is not created when "
+                    "the engine is not set to 'remote'. "
+                    "The run will proceed, but this setting will have no effect."
+                )
```
alonmr: I think this is a bit much. You can replace all of these lines with a single line in
```
workflow_spec.workflow_runner_node_selector = workflow_runner_node_selector
```
and if it's not remote it will just be ignored.
If you still want to elaborate about runner node selector not being used, you should use the `logger` instead of `warnings` package and I also think log level should be `debug`.
last thing is try to make the log shorter. but again I would just remove it.

tests/system/projects/test_project.py
```diff
@@ -595,51 +595,52 @@ def test_kfp_pipeline_with_resource_param_passed(self):
...
+        project.save()
+
+        project.run(
             workflow_name,
-            watch=watch,
-            local=local,
-            engine=engine,
-            notifications=notifications,
+            watch=False,
+            local=False,
```
alonmr: these 2 can be removed

tests/system/projects/test_project.py
```diff
@@ -595,51 +595,52 @@ def test_kfp_pipeline_with_resource_param_passed(self):
...
-            run.state == mlrun_pipelines.common.models.RunStatuses.succeeded
-        ), "pipeline failed"
-        # run.run_id can be empty in case of a local engine:
-        assert run.run_id is not None, "workflow's run id failed to fetch"
+        # Test scheduled workflow
+        schedule = "0 0 30 2 *"
+        project.run(
+            workflow_name,
+            watch=False,
+            local=False,
```
alonmr: same

alonmr (changes requested):
Looks great! some small comments

alonmr (approved):

PR #6334, opened 2024-09-15
[Alerts] Fix alert name in test_app_flow system test

Yacouby:
adjust the alert name in MM system test, after the changes made in this PR:
https://github.com/mlrun/mlrun/pull/6306


davesh0812 (commented):
We are using those constants in other places also.. why not to normalize them in first place?
Are you sure the test is working? I got some pydentic error from the server...

Yacouby:
> We are using those constants in other places also.. why not to normalize them in first place?
That will demand a change in the schemas, which is not recommended to do now

> Are you sure the test is working? I got some pydentic error from the server...
Yes, I ran the test and it passed on my vmdev

PR #6338, opened 2024-09-16
[DevEnv] Improve patch_remote script

adamdelman:
Improve script for faster building and support for pushing mlrun image

### DIFF
--- a/automation/patch_igz/patch_remote.py
+++ b/automation/patch_igz/patch_remote.py
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
 import io
 import json
 import logging
@@ -34,12 +33,21 @@ logger = logging.getLogger("mlrun-patch")
 coloredlogs.install(level=log_level, logger=logger, fmt=fmt)
 
 
-class MLRunPatcher:
-    class Consts:
-        mandatory_fields = {"DATA_NODES", "SSH_USER", "SSH_PASSWORD", "DOCKER_REGISTRY"}
-        api_container = "mlrun-api"
-        log_collector_container = "mlrun-log-collector"
+class Constants:
+    mandatory_fields = {"DATA_NODES", "SSH_USER", "SSH_PASSWORD", "DOCKER_REGISTRY"}
+    api_container = "mlrun-api"
+    log_collector_container = "mlrun-log-collector"
+    api = "api"
+    mlrun = "mlrun"
+    log_collector = "log-collector"
+    targets_to_image_name = {
+        api: api_container,
+        mlrun: mlrun,
+        log_collector: log_collector,
+    }
 
+
+class MLRunPatcher:
     def __init__(
         self,
         conf_file: str,
@@ -48,7 +56,7 @@ class MLRunPatcher:
         image_tag: str,
         patch_log_collector_image: bool,
         patch_mlrun_image: bool,
-        patch_api: bool,
+        skip_patch_api: bool,
     ):
         self._config = yaml.safe_load(conf_file)
         patch_yaml_data = yaml.safe_load(patch_file)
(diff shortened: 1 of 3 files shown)


automation/patch_igz/patch_remote.py
```diff
@@ -40,101 +40,120 @@ class Consts:
...
+    def patch(self):
+        version = self._get_current_version()
+        image_tag = self._get_image_tag(version)
+        targets = []
+        if self._patch_api:
+            targets.append("api")
+        if self._patch_mlrun_image:
+            targets.append("mlrun")
+        if self._patch_log_collector_image:
+            targets.append("log-collector")
```
TomerShor: You can move the string to the `Consts` class as you use them in multiple places

automation/patch_igz/patch_remote.py
```diff
@@ -40,101 +40,120 @@ class Consts:
...
+        if not targets:
+            raise ValueError("No targets to patch")
+
+        target_to_built_images = self._make_targets(
+            targets=targets,
+            image_tag=image_tag,
+        )
+        # Docker login and push images
         self._docker_login_if_configured()
```
TomerShor: move this before making the targets to fail fast if needed

automation/patch_igz/patch_remote.py
```diff
@@ -40,101 +40,120 @@ class Consts:
...
-            )
-            built_images.append(built_log_collector_image)
-
-        built_images = self._tag_images_for_multi_node_registries(built_images)
+        built_images = self._tag_images_for_multi_node_registries(
+            target_to_built_images.values()
+        )
         self._push_docker_images(built_images)
 
+        # Connect to the first node and start deployment process
```
TomerShor:
```suggestion
        # Connect to the first node and start deployment patching process
```

automation/patch_igz/patch_remote.py
```diff
@@ -40,101 +40,120 @@ class Consts:
...
         node = self._nodes[0]
         self._connect_to_node(node)
-        try:
-            self._replace_deploy_policy()
-            self._replace_deployment_images(self.Consts.api_container, built_api_image)
-            if self._patch_log_collector:
-                # sanity
-                if not built_log_collector_image:
-                    raise RuntimeError("Log collector image not built")
+        if self._patch_api:
```
TomerShor: what if we only want to patch the log collector?
or the mlrun image? (not api)

automation/patch_igz/patch_remote.py
```diff
@@ -40,101 +40,120 @@ class Consts:
...
+                    "docker",
+                    "login",
+                    docker_registry or "",
+                    "--username",
+                    registry_username,
+                    "--password",
+                    registry_password,
+                ],
+                live=True,
             )
```
TomerShor: Why not failing on error?

automation/patch_igz/patch_remote.py
```diff
@@ -224,12 +254,13 @@ def _push_docker_images(self, built_images):
...
+            ]
             self._exec_local(
-                [
-                    "docker",
-                    "push",
-                    image,
-                ],
+                cmd,
                 live=True,
             )
```
TomerShor: Not that it's wrong, but is there a reasoning for this change?

docs/runtimes/dask-mlrun.ipynb
TomerShor, line 1: How is this related to this PR?

automation/patch_igz/patch_remote.py
```diff
@@ -590,13 +621,42 @@ def _resolve_overwrite_registry(self):
...
+    "--mlrun",
+    is_flag=True,
+    help="Deploy the mlrun image",
+)
+@click.option(
+    "-ap",
+    "--api",
+    is_flag=True,
+    help="Deploy the mlrun API image",
 )
```
TomerShor: Thinking out loud: Maybe we should make this "True" by default, and change it to "skip-api", as most of the time we will use this script to patch the api

TomerShor (changes requested):
Looking good!
Added a couple of comment below

automation/patch_igz/patch_remote.py
```diff
@@ -40,101 +40,120 @@ class Consts:
...
+                [
+                    "docker",
+                    "login",
+                    docker_registry or "",
+                    "--username",
+                    registry_username,
+                    "--password",
+                    registry_password,
+                ],
+                live=True,
```
liranbg: suggest to undo to my changes so allow reading password from stdin and not as plain text

automation/patch_igz/patch_remote.py
adamdelman: fixed

automation/patch_igz/patch_remote.py
```diff
@@ -34,97 +33,123 @@
...
 
-            if self._reset_db:
-                self._reset_mlrun_db()
-            else:
-                self._rollout_deployment()
+                if self._patch_log_collector_image:
+                    self._replace_deployment_images(
+                        Consts.log_collector_container,
+                        target_to_built_images[Consts.log_collector],
+                    )
```
alonmr: this should be outside of `if not self._skip_patch_api:` to support patching only the log collector

automation/patch_igz/patch_remote.py
```diff
@@ -34,97 +33,123 @@
 coloredlogs.install(level=log_level, logger=logger, fmt=fmt)
 
 
-class MLRunPatcher:
-    class Consts:
-        mandatory_fields = {"DATA_NODES", "SSH_USER", "SSH_PASSWORD", "DOCKER_REGISTRY"}
-        api_container = "mlrun-api"
-        log_collector_container = "mlrun-log-collector"
+class Consts:
```
alonmr:
```suggestion
class Constants:
```

automation/patch_igz/patch_remote.py
```diff
@@ -34,97 +33,123 @@
...
-            self._replace_deploy_policy()
-            self._replace_deployment_images(self.Consts.api_container, built_api_image)
-            if self._patch_log_collector:
-                # sanity
-                if not built_log_collector_image:
-                    raise RuntimeError("Log collector image not built")
+        if not self._skip_patch_api:
+            try:
+                # Replace deployment policies and images
+                self._replace_deploy_policy()
```
alonmr: please change the name it's not very indicative 
```suggestion
                self._patch_deployment_from_file()
```

alonmr (changes requested):
Very much needed! Some small comments

automation/patch_igz/patch_remote.py
```diff
@@ -34,97 +33,123 @@
...
-                "log-collector", image_tag, "log-collector"
-            )
-            built_images.append(built_log_collector_image)
-
-        built_images = self._tag_images_for_multi_node_registries(built_images)
+        target_to_built_images = self._make_targets(
+            targets=targets,
+            image_tag=image_tag,
+        )
+        # Docker login and push images
```
TomerShor:
```suggestion
        # build and push images
```

automation/patch_igz/patch_remote.py
TomerShor: This comment still stands..

automation/patch_igz/patch_remote.py
```diff
@@ -34,97 +33,122 @@
...
         logger.info(
-            "Deployed branch successfully! Yay! (Note this may not survive system restarts)"
+            "Deployed branch successfully! (Note: This may not survive system restarts)"
         )
 
     def _docker_login_if_configured(self):
         registry_username = self._config.get("REGISTRY_USERNAME")
         registry_password = self._config.get("REGISTRY_PASSWORD")
-        registry_url = self._config.get("REGISTRY_URL")
+        docker_registry = self._config.get("DOCKER_REGISTRY")
```
alonmr, line 145: need to remove `REGISTRY_URL` from `automation/patch_igz/patch_env_template.yml`

alonmr (approved):

PR #6339, opened 2024-09-16
[Alerts] Refactor EventKind values to use hyphens instead of underscores

Yacouby:
https://iguazio.atlassian.net/browse/ML-7880


davesh0812 (approved):

TomerShor:
Backwards compatibility test break because the schema was changed, but it is a new schema in 1.7.0 so we are good 👍

PR #6345, opened 2024-09-17
[Alerts] Disable alerts by default

TomerShor:
Alerts are Tech Preview in v1.7.0, so they are now disabled by default.

https://iguazio.atlassian.net/browse/ML-7869


Yacouby:
integration tests will fail after this change, maybe we should enable the alerts there ?

server/api/api/endpoints/alerts.py
```diff
@@ -20,6 +20,7 @@
 from sqlalchemy.orm import Session
 
 import mlrun.common.schemas
+import mlrun.common.schemas.alert
```
theSaarco: This seems redundant - is this a leftover? This is the only change in this file, so not sure why it's needed.

mlrun/db/httpdb.py
```diff
@@ -4209,6 +4209,11 @@ def store_alert_config(
         if not alert_data:
             raise mlrun.errors.MLRunInvalidArgumentError("Alert data must be provided")
 
+        if mlrun.mlconf.alerts.mode == mlrun.common.schemas.alert.AlertsModes.disabled:
+            logger.warning(
+                "Alerts are disabled, alert will still be stored but will not be triggered"
+            )
+
```
theSaarco, line 4219: Should add a similar check & warning to the `generate_event` method - since it will be disabled, but will not fail, on the client-side there will be no indication to the situation, and the debug log in the server side will not be visible.

theSaarco (changes requested):
Looks good - there's one minor addition I'd like to add here.

tests/integration/sdk_api/alerts/test_alerts.py
```diff
@@ -26,6 +26,14 @@
 
 
 class TestAlerts(tests.integration.sdk_api.base.TestMLRunIntegration):
+    @classmethod
+    def setup_class(cls):
+        cls.extra_env = cls.extra_env or {}
+        cls.extra_env["MLRUN_ALERTS__MODE"] = (
+            mlrun.common.schemas.alert.AlertsModes.enabled
+        )
+        super().setup_class()
```
Yacouby: This call is likely overwriting or ignoring cls.extra_env if it's being set after the parent class setup logic..
It should be called at the beginning of this flow and not at the end of it

Yacouby (changes requested):
Looks good.
One comment on what I think why the integration tests config are not taking the env-var that you set.

tests/integration/sdk_api/alerts/test_alerts.py
TomerShor: Yes I changed it to only set the `extra_env` class property and not override `setup_class`.

theSaarco (approved):
LGTM

PR #6347, opened 2024-09-17
[Pagination] Page Token Improvements

quaark:
On the last page, don't return the page token, but don't revoke it, so you can use it to go to previous pages (tokens are cleaned up after a configured amount of time of non-usage and when the table reaches a max-size).

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -332,19 +332,18 @@ class HTTPRunDB(RunDBInterface):
         first_page_params["page"] = 1
         first_page_params["page-size"] = config.httpdb.pagination.default_page_size
         response = _api_call(first_page_params)
-        page_token = response.json().get("pagination", {}).get("page-token")
-        if not page_token:
-            yield response
-            return
+
+        yield response
+        page_token = response.json().get("pagination", {}).get("page-token", None)
 
         while page_token:
-            yield response
             try:
                 response = _api_call({"page-token": page_token})
             except mlrun.errors.MLRunNotFoundError:
                 # pagination token expired
                 break
 
+            yield response
             page_token = response.json().get("pagination", {}).get("page-token", None)
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


mlrun/db/httpdb.py
```diff
@@ -332,21 +332,18 @@ def _api_call(_params):
...
-        params_with_page_token = deepcopy(params) or {}
-        params_with_page_token["page-token"] = page_token
+        yield response
+        page_token = response.json().get("pagination", {}).get("page-token", None)
+
         while page_token:
-            yield response
             try:
-                response = _api_call(params_with_page_token)
+                response = _api_call({"page-token": page_token})
```
liranbg, line 344: I suggest to add a comment to why you ignore the params and take only the page token

server/api/utils/pagination.py
```diff
@@ -134,6 +134,15 @@ async def paginate_permission_filtered_request(
             current_page = last_pagination_info.page + 1
             page_size = last_pagination_info.page_size
 
+        if page_size and len(result) < page_size:
+            # on the last page, we don't return the token, but we keep it live in the cache
+            # so the client can access previous pages.
+            # the token will be revoked after some time of none-usage.
+            self._logger.debug(
+                "Last page, not returning token", token=token, method=method.__name__
+            )
```
liranbg: too verbose I think

server/api/utils/pagination.py
```diff
@@ -193,12 +202,11 @@ async def paginate_request(
             )
         except (RuntimeError, StopIteration) as exc:
             if isinstance(exc, StopIteration) or "StopIteration" in str(exc):
+                # don't revoke the token here as we might still want to go to previous pages.
+                # the token will be revoked after some time of none-usage.
                 self._logger.debug(
                     "End of pagination", token=token, method=method.__name__
                 )
```
liranbg: too verbose

liranbg (changes requested):

PR #6349, opened 2024-09-17
[Artifacts] Limit the artifact body size when logging artifacts

moranbental:
To avoid handling large amounts of data in memory, we limit the body size to 16MB. 
If the body exceeds this limit, an error will be raised, directing users to log larger artifacts through files using the `src_path` parameter.
https://iguazio.atlassian.net/browse/ML-7240

### DIFF
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -280,9 +280,7 @@ def validate_artifact_body_size(body: typing.Union[str, bytes, None]) -> None:
             "The body of the artifact exceeds the maximum allowed size. "
             "Avoid embedding the artifact body. "
             "This increases the size of the project yaml file and could affect the project during loading and saving. "
-            "More information is available at "
-            "https://docs.mlrun.org/en/latest/projects/automate-project-git-source.html#setting-and-registering-the-project-artifacts."
-            " For larger objects, please use the `src_path` parameter to log artifacts through files instead."
+            "For larger artifacts, consider logging them through files instead."
         )
 
 
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -282,9 +282,9 @@ def validate_artifact_body_size(body: typing.Union[str, bytes, None]) -> None:
             "The body of the artifact exceeds the maximum allowed size. "
             "Avoid embedding the artifact body. "
             "This increases the size of the project yaml file and could affect the project during loading and saving. "
-            "More information is available at"
-            "https://docs.mlrun.org/en/latest/projects/automate-project-git-source.html#setting-and-registering-the-project-artifacts"
-            "Please use `src_path` instead."
+            "More information is available at "
+            "https://docs.mlrun.org/en/latest/projects/automate-project-git-source.html#setting-and-registering-the-project-artifacts. "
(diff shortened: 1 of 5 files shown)


mlrun/utils/helpers.py
```diff
@@ -267,14 +267,24 @@ def validate_artifact_key_name(
...
+def validate_artifact_body_size(body: typing.Union[str, bytes, None]) -> None:
+    """
+    Validates the size of the artifact body.
+
+    Args:
+        body: The artifact body, which can be a string, bytes, or None.
+
+    Raises:
+        mlrun.errors.MLRunBadRequestError: If the body exceeds the maximum allowed size.
+    """
```
rokatyy: Please follow the same format as here https://github.com/mlrun/mlrun/blob/aee322fc108f2e0a899a7be04a0f0d173c548115/mlrun/utils/helpers.py#L1484-L1488

moranbental: thanks

mlrun/utils/helpers.py
```diff
@@ -267,14 +267,22 @@ def validate_artifact_key_name(
...
+    :raises mlrun.errors.MLRunBadRequestError: If the body exceeds the maximum allowed size.
+    """
     if body and len(body) > MYSQL_MEDIUMBLOB_SIZE_BYTES:
         raise mlrun.errors.MLRunBadRequestError(
             "The body of the artifact exceeds the maximum allowed size. "
             "Avoid embedding the artifact body. "
             "This increases the size of the project yaml file and could affect the project during loading and saving. "
-            "More information is available at"
-            "https://docs.mlrun.org/en/latest/projects/automate-project-git-source.html#setting-and-registering-the-project-artifacts"
+            "More information is available at "
```
alonmr: A few comments about the error message.
1. I don't see any extra information in the link. IMO it can be removed.
2. No need to say "please" and don't specify parameters name of other methods you don't know where you cam from (could be packagers).

alonmr (commented):
I think in the future we should automate creating a file from the artifact body and uploading that tmp file without having the users do it themselves. Maybe worth adding a TODO.

rokatyy (approved):

mlrun/utils/helpers.py
```diff
@@ -274,14 +274,20 @@ def validate_artifact_key_name(
...
+    :raises mlrun.errors.MLRunBadRequestError: If the body exceeds the maximum allowed size.
+    """
     if body and len(body) > MYSQL_MEDIUMBLOB_SIZE_BYTES:
         raise mlrun.errors.MLRunBadRequestError(
             "The body of the artifact exceeds the maximum allowed size. "
             "Avoid embedding the artifact body. "
             "This increases the size of the project yaml file and could affect the project during loading and saving. "
-            "More information is available at"
-            "https://docs.mlrun.org/en/latest/projects/automate-project-git-source.html#setting-and-registering-the-project-artifacts"
+            "For larger artifacts, consider logging them through files instead."
```
liranbg: The error is not correct.
The embedding (upper part) is relevant for when using `inline=True` and the last part is relevant when it is False.

liranbg (changes requested):

liranbg (approved):

PR #6355, opened 2024-09-18
[Config] Raising more coherent message when config reload is failing

liranbg:
when providing wrong envvar for mlrun to load, the error is not user-friendly in terms of the root cause.

https://iguazio.atlassian.net/browse/CEML-32

### DIFF
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -801,10 +801,10 @@ class Config:
                     # input error that can lead to unexpected behavior.
                     # raise the exception to ensure configuration is loaded correctly and do not
                     # ignore any errors.
+                    config_value = getattr(self, key)
                     try:
-                        getattr(self, key).update(value)
+                        config_value.update(value)
                     except AttributeError as exc:
-                        config_value = getattr(self, key)
                         if not isinstance(config_value, (dict, Config)):
                             raise ValueError(
                                 f"Can not update `{key}` config. "
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -796,7 +796,14 @@ class Config:
         for key, value in cfg.items():
             if hasattr(self, key):
                 if isinstance(value, dict):
-                    getattr(self, key).update(value)
+                    try:
+                        getattr(self, key).update(value)
+                    except AttributeError as exc:
+                        config_value = getattr(self, key)
+                        if not isinstance(config_value, (dict, Config)):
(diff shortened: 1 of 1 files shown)


mlrun/config.py
```diff
@@ -796,7 +796,15 @@ def update(self, cfg, skip_errors=False):
...
                 if isinstance(value, dict):
-                    getattr(self, key).update(value)
+                    try:
+                        getattr(self, key).update(value)
+                    except AttributeError as exc:
+                        config_value = getattr(self, key)
+                        if not isinstance(config_value, (dict, Config)):
+                            raise ValueError(
+                                f"Can not update `{key}` config. "
+                                f"Expected a configuration but received {type(config_value)}"
```
moranbental, line 806:
```suggestion
                                f"Expected a configuration but received {type(config_value).__name__}"
```

mlrun/config.py
```diff
@@ -796,7 +796,15 @@ def update(self, cfg, skip_errors=False):
         for key, value in cfg.items():
             if hasattr(self, key):
                 if isinstance(value, dict):
-                    getattr(self, key).update(value)
+                    try:
+                        getattr(self, key).update(value)
+                    except AttributeError as exc:
+                        config_value = getattr(self, key)
```
moranbental:
```suggestion
                    config_value = getattr(self, key)
                    try:
                        config_value.update(value)
                    except AttributeError as exc:
```

moranbental (changes requested):

mlrun/config.py
```diff
@@ -796,7 +796,15 @@ def update(self, cfg, skip_errors=False):
...
+                        if not isinstance(config_value, (dict, Config)):
```
alonmr, line 808: I think for completeness you need an else clause that raises the original exception

mlrun/config.py
alonmr: IIUC what you received is `value` and not `config_value`
```suggestion
                                f"Expected a configuration but received {type(value)}"
```

alonmr (commented):
Couple of comments and also there is a `skip_errors` flag might want to use it or add a comment why we are not using it

mlrun/config.py
liranbg: `value` is what you expect it to be, `config_value` is what you have received

PR #6361, opened 2024-09-18
[Labels] Enforce labels validation

moranbental:
Add label validation for the following functions:
1. list_artifacts
2. list_projects
3. list_functions
4. list_model_endpoints
5. del_artifacts
6. list_runs
7. del_runs
8. list_features
9. list_features_v2
10. list_entities
11. list_entities_v2
12. list_feature_sets
13. list_feature_vectors

to check if the format is either `str` or `list[str]` or `dict[str, str]`.
https://iguazio.atlassian.net/browse/ML-7384


mlrun/common/schemas/common.py
```diff
@@ -41,3 +41,38 @@ class ImageBuilder(pydantic.BaseModel):
 
     class Config:
         extra = pydantic.Extra.allow
+
+
+class LabelsModel(pydantic.BaseModel):
```
liranbg, line 48:
```suggestion
class Labels(pydantic.BaseModel):
```

mlrun/common/schemas/common.py
```diff
@@ -41,3 +41,38 @@ class ImageBuilder(pydantic.BaseModel):
...
+              be a simple label name (e.g., 'label1') or a key-value pair in the format
+              'label=value'.
+            - If no labels are specified, the default is an empty list.
+    """
+
+    labels: typing.Optional[typing.Union[dict[str, str], list[str]]]
+
+    @pydantic.validator("labels", pre=True)
+    @classmethod
+    def validate_labels(cls, labels) -> list[str]:
```
liranbg:
```suggestion
    def validate(cls, labels) -> list[str]:
```

mlrun/common/schemas/common.py
```diff
@@ -41,3 +41,38 @@ class ImageBuilder(pydantic.BaseModel):
...
+        if labels is None:
+            return []
```
liranbg, line 72:
```suggestion
        if not labels:
            return []
```

liranbg (changes requested):
some suggestions and fixes

in addition, missing some ut coverage

mlrun/common/schemas/common.py
moranbental: Why? This is to handle cases where labels are passed as a boolean. In such cases, I would like to raise an exception

mlrun/common/schemas/common.py
liranbg: perhaps, the name is ok - having the Model part of the name

mlrun/db/httpdb.py
```diff
@@ -4332,6 +4406,15 @@ def list_alert_templates(self) -> list[mlrun.common.schemas.AlertTemplate]:
             results.append(mlrun.common.schemas.AlertTemplate(**item))
         return results
 
+    @staticmethod
+    def _validate_labels(labels):
```
alonmr: better to have validation on server side to support it for old clients and not restrict 1.8 clients if the validation changes

mlrun/common/schemas/common.py
```diff
@@ -41,3 +43,44 @@ class ImageBuilder(pydantic.BaseModel):
 
     class Config:
         extra = pydantic.Extra.allow
+
+
+class LabelsModel(pydantic.BaseModel):
```
alonmr, line 48: I believe you can do something like
```
Labels = Annotated[AfterValidator(validate_labels)]
```
And then
```
labels: Labels = Query([], alias="label"),
```

alonmr (commented):
Very nice! Couple of comments

mlrun/common/schemas/common.py
alonmr: https://docs.pydantic.dev/latest/concepts/validators/

moranbental: @alonmr `AfterValidator` is not supported in Pydantic v1

moranbental: @alonmr We cannot get the labels as `labels: Labels = Query([], alias='label')` because it is required to be sent in the body and not as a parameter
`AssertionError: Param: labels can only be a request body, using Body()`

mlrun/common/schemas/common.py
```diff
@@ -41,3 +43,44 @@ class ImageBuilder(pydantic.BaseModel):
...
+            if not all(isinstance(item, str) for item in labels):
+                raise mlrun.errors.MLRunValueError(
+                    "All items in the list must be strings."
+                )
+            return labels
+
+        if isinstance(labels, dict):
+            if not labels:
+                return []
+            return [f"{key}={value}" for key, value in labels.items()]
```
alonmr: We also support giving mix of labels with values and labels without. So a dict like:
```
{"a": "b", "a": None}
```
Should be valid and should output to
```
["a=b", "a"]
```

mlrun/db/httpdb.py
```diff
@@ -915,16 +926,20 @@ def del_runs(self, name=None, project=None, labels=None, state=None, days_ago=0)
...
+        :param labels: Filter runs that are labeled using these specific label values. This can be provided as:
+            - A dictionary in the format {"label": "value"} to match specific label key-value pairs.
+            - A list of strings formatted as "label=value" to match specific label key-value pairs.
+            - A list of strings formatted as "label" to match entities with the specified label key only.
         :param state: Filter only runs which are in this state.
         :param days_ago: Filter runs whose start time is newer than this parameter.
         """
 
         project = project or config.default_project
+        labels = self._validate_labels(labels)
```
alonmr: my main issue with this is that now if I add a new endpoint with labels param I need to remember to use this validator but I don't see a way around it that I like ATM.
Let's rename it to `_parse_labels` and comment that we do this to support providing dict from the SDK while it's not supported in the endpoints.
I think this clarifies that although we do validation here, the actual goal is to support dict.

alonmr (commented):
Latests thoughts

mlrun/projects/project.py
```diff
@@ -3816,8 +3816,10 @@ def list_artifacts(
             case-sensitive. This means that querying for ``~name`` may return artifacts named
             ``my_Name_1`` or ``surname``.
         :param tag: Return artifacts assigned this tag.
-        :param labels: Return artifacts that have these labels. Labels can either be a dictionary {"label": "value"} or
-            a list of "label=value" (match label key and value) or "label" (match just label key) strings.
+        :param labels: Return artifacts that have these labels. This can be provided as:
+            - A dictionary in the format {"label": "value"} to match specific label key-value pairs.
+            - A list of strings formatted as "label=value" to match specific label key-value pairs.
+            - A list of strings formatted as "label" to match entities with the specified label key only.
```
```suggestion
        :param labels: Filter artifacts by labels key-value pairs or key existence. This can be provided as:
            - A dictionary in the format {"label": "value"} to match specific label key-value pairs or {"label": None} for key existence.
            - A list of strings formatted as "label=value" to match specific label key-value pairs or "label" for key existence.
            - A comma separated string formatted as "label1=value1,label2" to match entities with the specified key-value pair/key existence.
```

alonmr: We also support list of comma separated labels. Need to allow this in validation.

moranbental: > We also support list of comma separated labels. Need to allow this in validation.

It is already supported in validation because if the input is a list, we check that all elements are strings. The actual validation is performed on the server https://github.com/mlrun/mlrun/blob/61f943eeeb23ea7e3c0b4379bdf927c694359eb3/server/api/db/sqldb/db.py#L4717

moranbental: For the last bullet, it should be a list of comma-separated strings, not just a single comma-separated string.

mlrun/common/schemas/common.py
```diff
@@ -41,3 +43,51 @@ class ImageBuilder(pydantic.BaseModel):
...
+
+
+class LabelsModel(pydantic.BaseModel):
+    """
+    This class accepts either a dictionary, a list, or a string for filtering by labels.
+
+    :param labels:
+        - If a dictionary is provided, it should be in the format
+          {'label_name': 'value'}. This will be converted to a list of strings
+          in the format 'label_name=value'.
```
alonmr: value can be `None` to filter only by key

mlrun/db/base.py
```diff
@@ -101,7 +101,14 @@ def del_run(self, uid, project="", iter=0):
...
 
     @abstractmethod
-    def del_runs(self, name="", project="", labels=None, state="", days_ago=0):
+    def del_runs(
+        self,
+        name: str = "",
+        project: str = "",
+        labels: Optional[Union[str, dict[str, Optional[str]], list[str]]] = None,
+        state: Optional[mlrun.common.runtimes.constants.RunStates] = None,
+        days_ago=0,
```
alonmr:
```suggestion
        days_ago: int = 0,
```

mlrun/db/httpdb.py
```diff
@@ -1144,23 +1167,38 @@ def list_artifacts(
...
-        :param labels: Choose artifacts which are labeled.
+        :param labels: Filter artifacts by label key-value pairs or key existence. This can be provided as:
+            - A dictionary in the format `{"label": "value"}` to match specific label key-value pairs,
+            or `{"label": None}` to check for key existence.
+            - A list of strings formatted as `"label=value"` to match specific label key-value pairs,
+            or just `"label"` for key existence.
+            - A comma-separated string formatted as `"label1=value1,label2"` to match entities with
+            the specified key-value pairs or key existence.
         :param days_ago: This parameter is deprecated and not used.
+        :param tree: Delete artifacts that are contained within this tree.
```
alonmr:
```suggestion
        :param tree: Delete artifacts filtered by tree.
```

tests/test_common.py
alonmr, line 1: please create a folder `tests/common` and move this to `tests/common/schemas.py`

alonmr (commented):
Very well! some minor comments. Also can add some more type hints if you already started 😃

alonmr (approved):
Great!

liranbg (approved):

PR #6367, opened 2024-09-19
[Tests] Fix accessing to the workflow_request when load_only=True

yaelgen:
Fix broken system tests by: https://github.com/mlrun/mlrun/pull/6333
When loading a project with `load_only=True`, the workflow_request is `None`. In such cases, we should avoid attempting to update the workflow_runner_node_selector.


liranbg (approved):

alonmr (commented):
This already happened to @quaark in https://github.com/mlrun/mlrun/pull/6226
Perhaps we need a different fix so that it does't happen again. Something like creating an empty spec.

quaark:
> This already happened to @quaark in #6226 Perhaps we need a different fix so that it does't happen again. Something like creating an empty spec.

@alonmr I think the fact we are using the workflow runner to load the project is both problematic and will also break with @yaelgen 's `load_and_run` changes now I think of it - @yaelgen we need to make your change switchable for the case where `workflow_request is None` to maintain the load and save.
The real solution is either to load the project from the api and not a run or have a general JobRunner that the WorkflowRunner will inherit from and keep the workflow_request only for the specific runner implementation that needs it.

PR #6368, opened 2024-09-19
[Artifacts] Update the read artifact query to include the tag in the response [1.7.x]

moranbental:
1. If the request includes a tag and a UID, and the tag is not "latest," the query will include the tag, and it will be present in the response.
2. If the request includes a tag but not a UID, the query will include the tag, and it will be present in the response.
3. If the request includes a tag and a UID, and the tag is "latest," the system will make a best-effort attempt to find the "latest" tag. It will be present in the response if the latest tag exists, otherwise, it will **not** be included.
4. If the request does not include a tag at all, the query will not include the tag, and it will **not** be present in the response.

Also, adding a deprecation warning for the UID parameter
 
https://iguazio.atlassian.net/browse/ML-7777


server/api/db/sqldb/db.py
```diff
@@ -777,7 +777,7 @@ def read_artifact(
         computed_tag = tag or "latest"
         enrich_tag = False
 
-        if tag and not uid:
+        if tag and (not uid or (uid and tag != "latest")):
```
alonmr: add a comment to explain this pls

alonmr (commented):
I believe you are missing part of the fix

> if a UID is given and the tag is latest - ignore the tag when querying, but do an outer join with the tags table (as if the user didn’t ask for a tag).
> Now that you have the list of artifacts + tags - 
> if the results have a result with latest - enrich the artifact with the tag.
> If no latest exists, return the artifact without a tag.

server/api/db/sqldb/db.py
```diff
@@ -777,16 +777,23 @@ def read_artifact(
...
+        if tag == "latest" and uid:
+            latest_query = query.outerjoin(ArtifactV2.Tag).filter(
+                ArtifactV2.Tag.name == "latest"
+            )
+            if latest_query.one_or_none():
+                enrich_tag = True
+        elif tag and (not uid or tag != "latest"):
+            enrich_tag = True
+            query = query.join(ArtifactV2.Tag).filter(
+                ArtifactV2.Tag.name == computed_tag
```
TomerShor: Is the computed_tag still needed?

server/api/db/sqldb/db.py
```diff
@@ -777,16 +777,23 @@ def read_artifact(
...
         if producer_id:
             query = query.filter(ArtifactV2.producer_id == producer_id)
 
+        if tag == "latest" and uid:
+            latest_query = query.outerjoin(ArtifactV2.Tag).filter(
+                ArtifactV2.Tag.name == "latest"
+            )
+            if latest_query.one_or_none():
+                enrich_tag = True
+        elif tag and (not uid or tag != "latest"):
```
TomerShor: add a comment on what is this condition for

tests/api/api/test_artifacts.py
```diff
@@ -983,3 +1058,15 @@ def _generate_artifact_body(
...
 
     return data
+
+
+def _get_artifact_url(uid: str = None, tag: str = None) -> str:
+    url = GET_API_ARTIFACT_v2_PATH.format(project=PROJECT, key=KEY)
+    params = []
+
+    if uid:
+        params.append(f"uid={uid}")
```
TomerShor: IIRC It's supposed to be 'object-uid'

TomerShor (changes requested):
Looking good, added some comments

tests/api/api/test_artifacts.py
moranbental: We support both options, but I will change that anyway https://github.com/mlrun/mlrun/blob/7e6e4666c23f7423bd8506c1caefa0d946457ae3/server/api/api/endpoints/artifacts_v2.py#L229C8-L229C38
I will add a deprecation for this field, similar to what we did for deleting artifacts https://github.com/mlrun/mlrun/blob/7e6e4666c23f7423bd8506c1caefa0d946457ae3/server/api/api/endpoints/artifacts_v2.py#L246

server/api/api/endpoints/artifacts_v2.py
```diff
@@ -205,7 +205,13 @@ async def get_artifact(
     tag: str = None,
     iter: int = None,
     object_uid: str = Query(None, alias="object-uid"),
-    uid: str = Query(None),
+    # TODO: remove deprecated uid parameter in 1.9.0
```
alonmr, line 208: nice, please make sure it is documented in https://iguazio.atlassian.net/browse/ML-5948 and not available or deprecated in httpdb

server/api/db/sqldb/db.py
```diff
@@ -773,20 +773,28 @@ def read_artifact(
...
             query = query.filter(ArtifactV2.producer_id == producer_id)
 
+        if tag == "latest" and uid:
+            # Make a best-effort attempt to find the "latest" tag. It will be present in the response if the
+            # latest tag exists, otherwise, it will not be included.
+            latest_query = query.outerjoin(ArtifactV2.Tag).filter(
+                ArtifactV2.Tag.name == "latest"
+            )
+            if latest_query.one_or_none():
+                enrich_tag = True
```
alonmr, line 791: outer join - `Combines both tables, including all As and Bs, even if they don’t have a match in the other table.`
are you sure this logic holds?

server/api/db/sqldb/db.py
```diff
@@ -773,20 +773,28 @@ def read_artifact(
...
-            query = query.join(ArtifactV2.Tag, ArtifactV2.Tag.obj_id == ArtifactV2.id)
-            query = query.filter(ArtifactV2.Tag.name == computed_tag)
         if uid:
             query = query.filter(ArtifactV2.uid == uid)
         if producer_id:
             query = query.filter(ArtifactV2.producer_id == producer_id)
 
+        if tag == "latest" and uid:
+            # Make a best-effort attempt to find the "latest" tag. It will be present in the response if the
+            # latest tag exists, otherwise, it will not be included.
```
alonmr, line 785: add - this is due to 'latest' being a special case and is enriched in the client side.

server/api/db/sqldb/db.py
```diff
@@ -773,20 +773,28 @@ def read_artifact(
...
+            latest_query = query.outerjoin(ArtifactV2.Tag).filter(
+                ArtifactV2.Tag.name == "latest"
+            )
+            if latest_query.one_or_none():
+                enrich_tag = True
+        elif tag and (not uid or tag != "latest"):
```
alonmr: `not uid or tag != "latest"` is the negated condition of the if clause so it is not needed (it will always be True)

server/api/db/sqldb/db.py
```diff
@@ -773,20 +773,28 @@ def read_artifact(
...
+            # If a specific tag is provided, handle all cases where UID may or may not be included.
+            # The case for UID with the "latest" tag is already covered above.
+            # Here, we join with the tags table to check for a match with the specified tag.
+            enrich_tag = True
+            query = query.join(ArtifactV2.Tag).filter(ArtifactV2.Tag.name == tag)
```
alonmr:
```suggestion
            query = query.join(ArtifactV2.Tag, ArtifactV2.Tag.obj_id == ArtifactV2.id).filter(ArtifactV2.Tag.name == tag)
```

tests/api/api/test_artifacts.py
```diff
@@ -45,6 +45,7 @@
 DELETE_API_ARTIFACTS_V2_PATH = V2_PREFIX + DELETE_API_ARTIFACTS_PATH
 STORE_API_ARTIFACTS_V2_PATH = V2_PREFIX + API_ARTIFACTS_PATH
 LIST_API_ARTIFACTS_V2_PATH = V2_PREFIX + API_ARTIFACTS_PATH
+GET_API_ARTIFACT_v2_PATH = V2_PREFIX + API_ARTIFACTS_PATH + "/{key}"
```
alonmr:
```suggestion
GET_API_ARTIFACT_V2_PATH = V2_PREFIX + API_ARTIFACTS_PATH + "/{key}"
```

alonmr (commented):
Great job! Some comments / questions

server/api/db/sqldb/db.py
moranbental: Yes, `latest_query.one_or_none()` will return None if there is no artifact with the 'latest' tag for the specified UID. There is unit test coverage for this case.

server/api/db/sqldb/db.py
```diff
@@ -773,20 +773,31 @@ def read_artifact(
...
         if uid:
             query = query.filter(ArtifactV2.uid == uid)
         if producer_id:
             query = query.filter(ArtifactV2.producer_id == producer_id)
 
+        if tag == "latest" and uid:
+            # Make a best-effort attempt to find the "latest" tag. It will be present in the response if the
+            # latest tag exists, otherwise, it will not be included.
+            # This is due to 'latest' being a special case and is enriched in the client side
+            latest_query = query.outerjoin(
```
alonmr: I can to a conclusion it makes more sense to use inner join here. It seems to me this only works because of the UID filter and it makes this confusing.

alonmr (approved):

TomerShor (approved):

PR #6370, opened 2024-09-19
[KFP] Obscuring KFP env credentials

adamdelman:
Modify the compiled YAML on the server side to include references to kubernetes secrets instead of plain text values

https://iguazio.atlassian.net/browse/ML-7735


tests/api/api/test_pipelines.py
```diff
@@ -280,7 +280,7 @@ def test_list_pipelines_specific_project(
     importlib.reload(server.api.crud)
 
 
-def test_create_pipeline(
+def wtest_create_pipeline(
```
liranbg:
```suggestion
def test_create_pipeline(
```

server/api/crud/pipelines.py
```diff
@@ -262,6 +270,12 @@ def create_pipeline(
 
         logger.debug("Writing pipeline to temp file", content_type=content_type)
 
+        env_vars = server.api.api.utils.create_secret_env_vars_for_pipeline(auth_info)
```
liranbg: make sure you delete the secret when pipeline is deleted

server/api/api/utils.py
```diff
@@ -773,6 +774,61 @@ def ensure_function_has_auth_set(
...
+                secret_key_ref=V1SecretKeySelector(
+                    name=mlrun_auth_secret_name,
+                    key=mlrun.common.schemas.AuthSecretData.get_field_secret_key(
+                        "access_key"
+                    ),
+                )
+            ),
+        ),
+    ]
+    return env_vars
```
liranbg: so we do need both v3io_access_key *and* mlrun_auth_session?

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/patcher.py
```diff
@@ -68,22 +68,29 @@ def _create_enriched_mlrun_workflow(
...
-                        function_obj,
-                    )
+            if "dag" in kfp_step_template:
+                continue
+            env_vars = kfp_step_template["container"].get("env", [])
+            kfp_step_template["container"]["env"] = [
+                env_var
+                for env_var in env_vars
+                if env_var["name"] not in ["MLRUN_AUTH_SESSION", "V3IO_ACCESS_KEY"]
+            ]
```
liranbg: some comments would be nice for other developers trying to understand why you are filtering out those envvars

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/patcher.py
```diff
@@ -68,22 +68,29 @@ def _create_enriched_mlrun_workflow(
...
-                        kfp_step_template,
-                        function_obj,
-                        "PriorityClassName",
-                        "priority_class_name",
-                    )
-                    _enrich_kfp_pod_security_context(
-                        kfp_step_template,
-                        function_obj,
-                    )
+            if "dag" in kfp_step_template:
```
liranbg: explain why skipping dag

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +684,130 @@ def _enrich_node_selector(function):
...
+    byte_buffer: bytes,
+    content_type: str,
+    env_vars: list[V1EnvVar],
+) -> bytes:
+    if content_type.endswith("zip"):
+        modified_zip_bytes = _enrich_kfp_workflow_zip_credentials(
+            yaml_bytes=byte_buffer,
+            env_vars=env_vars,
+        )
+        return modified_zip_bytes
```
liranbg, line 699: this is a cpu bound operation. could that happen as a subprocess?

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/models.py
```diff
@@ -102,3 +110,9 @@ def transient_statuses():
             for status in RunStatuses.all()
             if status not in RunStatuses.stable_statuses()
         ]
+
+
+class EngineType(StrEnum):
+    LOCAL = "local"
+    REMOTE = "remote"
+    KFP = "kfp"
```
liranbg: 👏🏼

liranbg (changes requested):
some suggestions and fixes, overall - looking very well!

mlrun/projects/pipelines.py
```diff
@@ -354,11 +358,18 @@ def enrich_function_object(
         f.enrich_runtime_spec(
             project.spec.default_function_node_selector,
         )
-
+    skip_automount_for_kfp = False
+    if project.spec.workflows:
+        skip_automount_for_kfp = (
+            project.spec.workflows[0].get("engine")
+            == mlrun_pipelines.common.models.EngineType.KFP
+        )
```
theSaarco: Not sure I understand what the logic here - this method (`enrich_function_object`) is called in several flows, not all of them are workflow-related.
First, generally speaking - why is this check needed? Why do we need a specific logic for `kfp` functions in this case?
Also, it seems that in this check, if the project has a workflow of type `kfp` you'll always skip auto-mounting functions, even if I'm running a function outside of the scope of the workflow? Is this the expected behavior?

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +684,130 @@ def _enrich_node_selector(function):
...
+
+
+def _enrich_kfp_workflow_yaml_credentials(
+    yaml_bytes: bytes,
+    env_vars: list[V1EnvVar],
+) -> bytes:
+    """
+    Modifies the given workflow dictionary to add environment variables from a Kubernetes secret
+    to the container specifications.
+    The function checks if the workflow is compatible with KFP v1 or KFP v2 and adds the
```
theSaarco: Just to verify I have this right - this is placed in the common pipelines code because even when running with kfp v2, you can still work with the old (i.e. argo) format, right? So it's not possible to place this in the 1.8/2.0 adapter code?

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +684,130 @@ def _enrich_node_selector(function):
...
+                                    "name": env_var.value_from.secret_key_ref.name,
+                                    "key": env_var.value_from.secret_key_ref.key,
+                                }
+                            },
+                        }
+                    else:
+                        env_var = {
+                            "name": env_var.name,
+                            "value": env_var.value,
+                        }
```
theSaarco: This code block is repeating twice, put it in an internal function. 
Also, since it's actually not specifically related to pipeline code (it's basically converting formats of env vars), it can actually sit in k8s_utils or some common utilities location.

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +684,130 @@ def _enrich_node_selector(function):
...
+            if "container" in template:
+                container = template["container"]
+                for env_var in env_vars:
+                    if env_var.value_from is not None:
+                        env_var = {
+                            "name": env_var.name,
+                            "valueFrom": {
+                                "secretKeyRef": {
+                                    "name": env_var.value_from.secret_key_ref.name,
+                                    "key": env_var.value_from.secret_key_ref.key,
```
theSaarco: Be careful here - the `value_from` property can have additional options (see [here](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1EnvVarSource.md)) - so it doesn't necessarily have the `secret_key_ref` field. Should validate it.

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/patcher.py
theSaarco: Agreed, and also keep in mind that this is an implementation decision made for now - to drop whatever arrived from the FE and instead plant the REST call key. This needs to be explained, since it's not necessarily what we'd want to do in the future - we may want to replace the clear-text access key with an auth secret that has its value in it. So, this needs to be documented to explain the considerations here.

server/api/api/utils.py
```diff
@@ -773,6 +774,61 @@ def ensure_function_has_auth_set(
...
 
+def create_secret_env_vars_for_pipeline(
+    auth_info: mlrun.common.schemas.AuthInfo,
+) -> list[V1EnvVar]:
+    """
+    :param auth_info:   The auth info of the request.
+    """
+
+    if not auth_info.access_key:
+        auth_info.access_key = (
```
theSaarco: I'm not a fan of this replacing the access-key in the auth-info. You don't necessarily know who's going to use that auth-info after you, and replacing the access-key in-place seems dangerous.

server/api/api/utils.py
```diff
@@ -773,6 +774,61 @@ def ensure_function_has_auth_set(
...
+            server.api.utils.auth.verifier.AuthVerifier().get_or_create_access_key(
+                session=auth_info.get_key(),
```
theSaarco: I think the `get_key()` function is a bit overloaded. In this use-case you already know there's no access-key, and you're just looking for the session. So instead of `get_key` I would create a `get_session` function that will just replace whatever session exists.
Then (and continuing my previous comment about not modifying the auth-info in-place), I would have a local `access_key` parameter that is either the auth-info `access_key` or the result of this call. Then the rest of the logic would use that local `access_key` parameter.

server/api/api/utils.py
```diff
@@ -773,6 +774,61 @@ def ensure_function_has_auth_set(
...
+                planes=[
+                    server.api.utils.clients.iguazio.SessionPlanes.data,
```
theSaarco: Why `data` plane? Is it enough for the workflow's purposes? Doesn't it also need it for control-plane?

server/api/api/utils.py
theSaarco: Auth session is control-plane. This also relates to my previous question - if we need both, then just a `data` plane access key will not suffice.

server/api/crud/pipelines.py
theSaarco: @liranbg - you cannot do that. The auth secret relates to a given username/access-key combination. It's very likely that the same auth-secret will be used by multiple runtimes/workflows, and it should only be deleted once the access-key is deleted (though we don't really do that today...)

theSaarco (changes requested):
Looks good, but there are some changes needed, and I'm also not so sure about some of the changes done and why they're needed.
Also, the original bug shows there are 2 locations that need cleaning - the env vars, and the spec of the function when a `deploy_function` is done. In this PR I can only see you handling the 1st case, not the 2nd. Are you planning to do that in a separate PR?

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/models.py
```diff
@@ -16,7 +16,15 @@
 from enum import Enum
 
 
-class RunStatuses(str, Enum):
+class StrEnum(str, Enum):
```
moranbental: This type already exists, and you can use it https://github.com/mlrun/mlrun/blob/4a33636f3deced38af89bf2c5461ee93582737e5/mlrun/common/types.py#L19

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +684,130 @@ def _enrich_node_selector(function):
...
+    in_memory_zip = io.BytesIO(byte_buffer)
+    with zipfile.ZipFile(in_memory_zip, "r") as zip_read:
+        file_list = zip_read.namelist()
+        files_data = {}
+        for file_name in file_list:
+            with zip_read.open(file_name) as f:
+                files_data[file_name] = f.read()
+
+    for file_name, file_data in files_data.items():
+        if file_name.endswith(".yaml") or file_name.endswith(".yml"):
```
moranbental:
```suggestion
        if file_name.endswith((".yaml", ".yml"))
```

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +684,130 @@ def _enrich_node_selector(function):
...
+    content_type: str,
+    env_vars: list[V1EnvVar],
+) -> bytes:
+    if content_type.endswith("zip"):
+        modified_zip_bytes = _enrich_kfp_workflow_zip_credentials(
+            yaml_bytes=byte_buffer,
+            env_vars=env_vars,
+        )
+        return modified_zip_bytes
+    elif content_type.endswith("yaml") or content_type.endswith("plain"):
```
moranbental:
```suggestion
    elif content_type.endswith(("yaml", "plain")):
```

moranbental (changes requested):
Looks good! some suggestions

server/api/crud/pipelines.py
adamdelman: @theSaarco @liran Couldn't we use an K8S owner annotation to GC the secret once the workflow runner pod is deleted?

mlrun/projects/pipelines.py
adamdelman: AFAIK the engine type of kfp is only used for workflows, so the change in automount behavior only affect this flow.

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
adamdelman: yes, thats correct.

server/api/api/utils.py
adamdelman: From what I can tell this seems to be enough, I'm not sure why we would need the control plane as well for the workflow.

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/patcher.py
adamdelman: Since this part is being done on the client side, I'm not sure how we would create the k8s secret here, as we have no k8s client initialized.

mlrun/projects/pipelines.py
alonmr: Saar means that you are checking if the project HAS a workflow but this may disable auto mount for regular jobs as well since they may pass here so the check is not accurate.
Project object having a workflow spec doesn't mean that this workflow is what's currently running.

server/api/crud/pipelines.py
liranbg: Lets take it offline for now

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +684,121 @@ def _enrich_node_selector(function):
...
+    byte_buffer: bytes,
+    content_type: str,
+    env_vars: list[V1EnvVar],
+) -> bytes:
+    if content_type.endswith("zip"):
+        modified_zip_bytes = _enrich_kfp_workflow_zip_credentials(
+            yaml_bytes=byte_buffer,
+            env_vars=env_vars,
+        )
+        return modified_zip_bytes
```
liranbg: this looks like some cpu-bound operation. could that be running aside from the api-serving process?

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/patcher.py
```diff
@@ -68,22 +68,32 @@ def _create_enriched_mlrun_workflow(
...
```
alonmr: why not keeping the `kfp_step_template.get("container")` condition to make sure we have container settings instead of checking for dag?
This can be changed to
```
if not kfp_step_template.get("container"):
    continue
```
This verifies what we need without relying on kfp logic (that dag is the only case that there is no container)

tests/api/api/test_pipelines.py
```diff
@@ -296,11 +306,22 @@ def test_create_pipeline(
...
+    app.dependency_overrides[deps.authenticate_request] = mock_authenticate_request
+
+    with patch(
+        target="server.api.utils.auth.verifier.AuthVerifier.get_or_create_access_key",
+        return_value="mock-access-key",
+    ):
+        with patch(
+            target="server.api.api.utils.create_secret_env_vars_for_pipeline",
+            return_value=[V1EnvVar(name="test", value="test")],
+        ):
```
alonmr: You can chain these
```suggestion
    ), patch(
            target="server.api.api.utils.create_secret_env_vars_for_pipeline",
            return_value=[V1EnvVar(name="test", value="test")],
        ):
```

alonmr: actually don't mock this - you can use `k8s_secrets_mock` and then you can also assert that the secret was created as expected

mlrun/projects/pipelines.py
alonmr: discussed offline, use `pipeline_context` here

alonmr (commented):
Very nice! small comments

tests/api/conftest.py
```diff
@@ -123,6 +123,10 @@ def set_base_url_for_test_client(
     client.base_url = client.base_url.join(prefix)
 
 
+async def mock_authenticate_request():
+    return mlrun.common.schemas.AuthInfo(username="test_user", token="mock_token")
```
alonmr: is this needed? looks like no one is using it

mlrun/common/schemas/workflow.py
```diff
@@ -16,6 +16,7 @@
 
 import pydantic
 
+from ..types import StrEnum
```
alonmr: One of the new conventions in mlrun is not using relative imports

mlrun/projects/pipelines.py
```diff
@@ -354,11 +360,17 @@ def enrich_function_object(
         f.enrich_runtime_spec(
             project.spec.default_function_node_selector,
         )
-
+    skip_automount_for_kfp = False
+    if project.spec.workflows:
+        skip_automount_for_kfp = (
+            pipeline_context.workflow.get("engine") == mlrun.EngineType.KFP
```
theSaarco: Are you guaranteed that `pipeline_context.workflow` is set when you get here? If all you do is run a function (even though the project has workflows defined), will you have a valid workflow in the context?
I think the `if` should change to `if pipeline_context.workflow:` which is a better indication that we're running in workflow context.

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +685,141 @@ def _enrich_node_selector(function):
...
+    )
+
+    if "argoproj.io" in api_version:  # KFP Argo Workflow
+        for template in workflow_dict["spec"]["templates"]:
+            if "container" in template:
+                container = template["container"]
+                for env_var in env_vars:
+                    env_var = create_env_for_container(env_var)
+                    if "env" in container and container["env"] is not None:
+                        container["env"].append(env_var)
```
theSaarco: I think it would actually be better to check if the env already contains a variable of that name, and if it does - replace it. Appending blindly may cause a situation where you have multiple copies of the same env var, which can cause unpredictable results.

mlrun/projects/pipelines.py
```diff
@@ -354,11 +360,18 @@ def enrich_function_object(
         f.enrich_runtime_spec(
             project.spec.default_function_node_selector,
         )
-
+    skip_automount_for_kfp = False
+    if project.spec.workflows:
```
theSaarco: Shouldn't this check if there's a `pipeline_context.workflow`? Looking here may still return a positive result even if just running a function and there's no workflow in the pipeline context.

theSaarco (changes requested):
Looks very good. Have some relatively minor comments.
Also - still can't see any reference to replacing secrets from function spec in the deploy-function case. Will this be treated in a different PR?

mlrun/projects/pipelines.py
```diff
@@ -354,11 +360,18 @@ def enrich_function_object(
...
+        skip_automount_for_kfp = (
+            pipeline_context.workflow.engine
+            == mlrun.common.schemas.workflow.EngineType.KFP
+        )
```
liranbg: add doc - what is happening here

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +685,141 @@ def _enrich_node_selector(function):
         project_node_selector, function_node_selector
     )
     return mlrun.utils.helpers.to_non_empty_values_dict(function_node_selector)
+
+
+def enrich_wrapper(queue, byte_buffer, content_type, env_vars):
```
liranbg:
```suggestion
def _enrich_wrapper(queue, byte_buffer, content_type, env_vars):
```

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +685,141 @@ def _enrich_node_selector(function):
...
+            yaml_bytes=byte_buffer,
+            env_vars=env_vars,
+        )
+        return modified_zip_bytes
+    elif content_type.endswith(("yaml", "plain")):
+        modified_yaml_bytes = _enrich_kfp_workflow_yaml_credentials(
+            yaml_bytes=byte_buffer,
+            env_vars=env_vars,
+        )
+        return modified_yaml_bytes
```
liranbg: make sure this part runs inline and not subprocess as mlrun api spend most of its time here

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
liranbg: reference for existing solution https://github.com/mlrun/mlrun/blob/ffb8e69f75a7cfee8e8f6fe005439615b126c952/server/api/runtime_handlers/daskjob.py#L331

mlrun/projects/pipelines.py
adamdelman: fixed

liranbg (changes requested):
up

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +692,296 @@ def _enrich_node_selector(function):
         project_node_selector, function_node_selector
     )
     return mlrun.utils.helpers.to_non_empty_values_dict(function_node_selector)
+
```
alonmr, line 695: don't forget to bump mlrun pipelines common package in `pipeline-adapters/mlrun-pipelines-kfp-common/setup.py`

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +692,296 @@ def _enrich_node_selector(function):
...
+    The function checks if the workflow uses Argo Workflows or Tekton Pipelines and injects the
+    environment variables accordingly.
+    """
+    workflow_dict = yaml.safe_load(yaml_bytes)
+    # Determine the KFP version by checking the 'apiVersion' field
+    api_version = (
+        workflow_dict.get("api_version") or workflow_dict.get("apiVersion", "").lower()
+    )
+
+    if "argoproj.io" in api_version:  # KFP Argo Workflow
```
alonmr, line 800: IIUC this if is to check if kfp version is 1.8 or 2 and I think it misses the point of the adapters abstraction. I would try to move this to the matching packages but it doesn't look trivial so not needed for now but I think it should be moved

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +692,296 @@ def _enrich_node_selector(function):
...
+):
+    result = _enrich_kfp_workflow_zip_credentials(
+        byte_buffer=byte_buffer,
+        env_var_names=env_var_names,
+        secrets_store=secrets_store,
+    )
+    queue.put(result)
+
+
+def _enrich_kfp_workflow_credentials_in_subprocess(
```
alonmr: please maintain reading from top to bottom meaning functions called by other functions should be below them.
In this case `_enrich_kfp_workflow_credentials_in_subprocess ` should be above `_enrich_kfp_workflow_zip_credentials`

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +692,296 @@ def _enrich_node_selector(function):
...
+    cmd_parts: list[str],
+    env_var_names: list[str],
+    secret_name_to_secret_ref: dict[str, V1EnvVar],
+    secrets_store: "SecretsStore",
+):
+    """
+    Replaces specified environment variables in the function spec within cmd_parts.
+    """
+    for cmd_part_index, cmd_part in enumerate(cmd_parts):
+        if cmd_part == "-r" and cmd_part_index + 1 < len(cmd_parts):
```
alonmr, line 886: please add a comment to explain what is `-r` it looks like magic

server/api/crud/pipelines.py
```diff
@@ -259,9 +268,17 @@ def create_pipeline(
                 http.HTTPStatus.BAD_REQUEST.value,
                 reason=f"unsupported pipeline type {content_type}",
             )
-
         logger.debug("Writing pipeline to temp file", content_type=content_type)
-
+        logger.error("Raw pipeline yaml", workflow_dict=yaml.safe_load(data))
```
alonmr: was this log meant to be deleted?

server/api/crud/pipelines.py
```diff
@@ -259,9 +268,17 @@ def create_pipeline(
...
+        data = mlrun_pipelines.common.ops.replace_kfp_plaintext_secret_env_vars_with_secret_refs(
+            byte_buffer=data,
+            content_type=content_type,
+            env_var_names=["MLRUN_AUTH_SESSION", "V3IO_ACCESS_KEY"],
+            secrets_store=server.api.crud.Secrets(),
+        )
+        logger.error(
+            "Generated enriched pipeline yaml", workflow_dict=yaml.safe_load(data)
+        )
```
alonmr: same

tests/api/api/test_pipelines.py
```diff
@@ -296,13 +305,19 @@ def test_create_pipeline(
     with open(str(pipeline_file_path)) as file:
         contents = file.read()
     _mock_pipelines_creation(kfp_client_mock)
+    app.dependency_overrides[deps.authenticate_request] = mock_authenticate_request
```
alonmr: I'm not familiar with this mocking functionality. will it be reset after the test?

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +692,296 @@ def _enrich_node_selector(function):
...
     return mlrun.utils.helpers.to_non_empty_values_dict(function_node_selector)
+
+
+def replace_kfp_plaintext_secret_env_vars_with_secret_refs(
+    byte_buffer: bytes,
+    content_type: str,
+    env_var_names: list[str],
+    secrets_store: "SecretsStore",
+) -> bytes:
+    if content_type.endswith("zip"):
```
alonmr: please add a comment to explain when it will be a zip and when yaml / plain

alonmr (commented):
Looks very well! Some comments/questions

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +692,296 @@ def _enrich_node_selector(function):
...
+            container = template.get("container")
+            if container is not None:
+                updated_container = _replace_secret_envs_in_argocd_template(
+                    env_var_names=env_var_names,
+                    container=container,
+                    secrets_store=secrets_store,
+                )
+                template["container"] = updated_container
+
+        workflow_dict["spec"] = spec
```
theSaarco: Don't think this line is needed (the dict was updated in-place).

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
adamdelman: It isn't necessarily linked to the KFP version, you can have ArgoCD workflows in kfp 2.0 as well.

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +692,296 @@ def _enrich_node_selector(function):
...
+    step_envs = step_template.get("env", [])
+    for step_env_index, step_env in enumerate(step_envs):
+        env_var_name = step_env.get("name")
+        if env_var_name in env_var_names:
+            if env_var_name in secret_name_to_secret_ref:
+                secret_var = secret_name_to_secret_ref[env_var_name]
+            else:
+                secret_var = _create_secret_env_var_for_pipeline(
+                    name=step_env["name"],
+                    value=step_env["value"],
```
theSaarco: What if the incoming pipeline already has the env var pointing at a secret? It's not the common case, but might happen, and this code would fail since there's no `value` parameter. This entire transformation should be done only if the env variable is provided in cleartext format (i.e. has the `value` parameter) - should verify this before doing any change.

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +692,296 @@ def _enrich_node_selector(function):
...
+                    secrets_store=secrets_store,
+                )
+                secret_name_to_secret_ref[env_var_name] = secret_var
+
+            step_envs[step_env_index] = secret_var
+    step_template["env"] = step_envs
+    task["stepTemplate"] = step_template  # Ensure the modified step_template is updated
```
theSaarco: Don't think this is really needed since modifying in-place. Did you see this is mandatory to have the changes applied?

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +692,298 @@ def _enrich_node_selector(function):
...
+
+            func_spec = func_data.get("spec", {})
+            func_envs = func_spec.get("env", [])
+            func_spec["env"] = _replace_env_vars_with_secrets(
+                env_vars=func_envs,
+                env_var_names=env_var_names,
+                secret_name_to_secret_ref=secret_name_to_secret_ref,
+                secrets_store=secrets_store,
+            )
+            func_data["spec"] = func_spec
```
theSaarco: Here again - don't think this is needed, since `func_spec` itself hasn't changed, just its internal `env` member.

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py
```diff
@@ -680,3 +692,298 @@ def _enrich_node_selector(function):
...
+                secret_var = secret_name_to_secret_ref[env_var_name]
+            else:
+                secret_var = _create_secret_env_var_for_pipeline(
+                    name=step_env["name"],
+                    value=step_env["value"],
+                    secrets_store=secrets_store,
+                )
+                secret_name_to_secret_ref[env_var_name] = secret_var
+
+            step_envs[step_env_index] = secret_var
```
theSaarco: Can't you use `_replace_env_vars_with_secrets` on `step_envs`?

tests/api/api/test_pipelines.py
```diff
@@ -296,13 +303,18 @@ def test_create_pipeline(
...
         f"projects/{project}/pipelines",
         data=contents,
         headers={"content-type": "application/yaml"},
+        auth=BasicAuth(username="admin", password="mock_token"),
     )
     response_body = response.json()
     assert response_body["id"] == "some-run-id"
+    assert k8s_secrets_mock.auth_secrets_map[
+        "secret-ref-V3IO_ACCESS_KEY-some-session"
+    ] == {"accessKey": "some-session", "username": "V3IO_ACCESS_KEY"}
```
theSaarco, line 317: A thought that could improve this test - since the `kfp_client_mock` mocks all calls to the KFP `Client`, you could check the call actually being generated and verify that the contents of the pipeline really has the secrets obfuscated. This is a more direct and complete check than verifying the secrets-mock has this auth secret being created.
It's not a must, but if it's simple enough to do, would try to add that here.

theSaarco (dismissed):
Looks good - I'm approving as functionality seems complete, but have some suggestions for improvements that would make this cleaner.

liranbg (dismissed):

PR #6375, opened 2024-09-19
[ProjectSummaries] Fix wrong schedule jobs count

roei3000b:
Fixing bug on project summaries schedule jobs count.
To determine if schedule object is jobs/workflow we check its labels:

1. Schedule workflow will have label with name `workflow` and value of the name of the workflow, but also will have label with name `kind` and value `job`.
2. Schedule job will have  label with name `kind` and value `job`.

On the first implementation we count twice schedule workflow, one as workflow (because the name of the label is workflow), and another time as a schedule job - which is wrong.

Fixing it by first creating list of scheduled pipelines per project and only then iterate over schedules and add the schedule only if it not workflow.

for more details see my [comment](https://iguazio.atlassian.net/browse/ML-7875?focusedCommentId=158377).

https://iguazio.atlassian.net/browse/ML-7875

### DIFF
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -27,7 +27,7 @@ from typing import Any
 import fastapi.concurrency
 import mergedeep
 import pytz
-from sqlalchemy import MetaData, and_, delete, distinct, func, or_, select, text
+from sqlalchemy import MetaData, and_, case, delete, distinct, func, or_, select, text
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.inspection import inspect
 from sqlalchemy.orm import Session, aliased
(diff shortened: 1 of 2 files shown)


alonmr (commented):
I think the fix should be in the DB query. It would make this much simpler.
Why not grouping the DB response by schedule name?

server/api/db/sqldb/db.py
```diff
@@ -2858,20 +2893,19 @@ def _calculate_schedules_counters(
...
-                result[2].to_dict()["name"]
-                == mlrun_constants.MLRunInternalLabels.workflow
-            ):
-                project_to_schedule_pending_workflows_count[result[0]] += 1
-            elif result[2].to_dict()["value"] == "job":
-                project_to_schedule_pending_jobs_count[result[0]] += 1
+            project_name, schedule_name, kind = result
+            if kind == mlrun_constants.MLRunInternalLabels.workflow:
+                project_to_schedule_pending_workflows_count[project_name] += 1
+            elif kind == mlrun.common.schemas.ScheduleKinds.job.value:
```
alonmr: Since `ScheduleKinds` is `StrEnum`
```suggestion
            elif kind == mlrun.common.schemas.ScheduleKinds.job:
```

tests/api/db/test_schedule.py
```diff
@@ -68,6 +72,43 @@ def test_delete_schedules(db: DBInterface, db_session: Session):
...
+
+def test_calculate_schedules_counters(db: DBInterface, db_session: Session):
+    next_minute = datetime.now(timezone.utc) + timedelta(hours=1)
+
+    # Store schedule job
+    db.store_schedule(
+        db_session,
+        project="project1",
+        name="job1",
+        labels={mlrun_constants.MLRunInternalLabels.kind: "job"},
```
alonmr:
```suggestion
        labels={mlrun_constants.MLRunInternalLabels.kind: mlrun.runtimes.RuntimeKinds.job},
```

tests/api/db/test_schedule.py
```diff
@@ -68,6 +72,43 @@ def test_delete_schedules(db: DBInterface, db_session: Session):
...
+
+    pipelines_name = ["some_name", "some_name2", "some_name3"]
+    for name in pipelines_name:
+        # Store schedule pipeline
+        db.store_schedule(
+            db_session,
+            project="project2",
+            name=name,
+            labels={
+                mlrun_constants.MLRunInternalLabels.kind: "job",
```
alonmr: same

server/api/db/sqldb/db.py
```diff
@@ -2846,7 +2846,42 @@ def _calculate_schedules_counters(
...
+                Schedule.project,
+                Schedule.name,
+                # The logic here is the following:
+                # If the schedule has a label with the name "workflow" then we take the value of that label
+                # name. Otherwise, we take the value of the label with the name "kind".
+                # The reason for that is that for schedule workflow we have both workflow label and kind label
+                # with job, and on schedule job we have only kind label with job and because of that we first
+                # want to check for workflow label and if it doesn't exist then we take the kind label.
+                func.coalesce(
+                    func.max(
```
alonmr, line 2859: IIUC how everything works then the `func.max` is not needed right?

roei3000b: No, we still need the max because on workflow we can get both kind and workflow as label name(because of the group by) and we want to get the workflow and only if it doesn't exists get the kind

alonmr (approved):

PR #6378, opened 2024-09-22
[Feature Store] Fix remote ingest `start_time` and `end_time` parameters of `ParquetSource`

tomerm-iguazio:
1) Edited to_dict to send string instead of datetime in fields : start_time and end_time.
2) Added remote test.

[ML-7918](https://iguazio.atlassian.net/browse/ML-7918)

[ML-7918]: https://iguazio.atlassian.net/browse/ML-7918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/system/feature_store/test_feature_store.py
```diff
@@ -900,41 +900,65 @@ def test_featureset_column_types(self):
...
+            ],
             return_df=True,
+            run_config=run_config,
+        )
+        result_offline_target = get_offline_target(
+            measurements, name="second_parquet_target"
         )
-        assert len(resp) == 0
+        with pytest.raises(FileNotFoundError):
+            result_offline_target.as_df()
```
gtopper, line 970: Maybe it would be better to write a subset of the data and assert on that? Otherwise it's not possible to distinguish filtering from something just not working properly.

mlrun/model.py
```diff
@@ -2072,6 +2074,17 @@ def __init__(
...
+    def _serialize_field(
+        self, struct: dict, field_name: str = None, strip: bool = False
+    ) -> typing.Any:
+        # We pull the field from self and not from struct because it was excluded from the struct when looping over
+        # the fields to save.
+        if field_name in ["start_time", "end_time"]:
+            value = getattr(self, field_name, None)
+            if value and isinstance(value, datetime):
+                return value.isoformat()
+        return super()._serialize_field(struct, field_name, strip)
```
gtopper: Can be simplified a bit.
```suggestion
        value = super()._serialize_field(struct, field_name, strip)
        # We pull the field from self and not from struct because it was excluded from the struct when looping over
        # the fields to save.
        if field_name in ("start_time", "end_time") and isinstance(value, datetime):
            return value.isoformat()
        return value
```

gtopper (commented):
LGTM otherwise.

mlrun/model.py
tomerm-iguazio: nice, tnx!

tests/system/feature_store/test_feature_store.py
tomerm-iguazio: np, so it is relevant for the previous assert ...

tomerm-iguazio: in this assert, the ingest does not write to this target because it's the df become empty after filtering.

gtopper (approved):

PR #6387, opened 2024-09-23
[Project] Terminate pipelines before deleting pods on delete job [1.7.x]

alonmr:
We see pods being created during the delete job which interferes with the project delete job.
This will terminate the pipelines before starting to delete pods.

https://iguazio.atlassian.net/browse/ML-7786


moranbental (approved):
LGTM

PR #6389, opened 2024-09-23
[Schedule] Synchronize labels on upgrade

rokatyy:
After implementing [ML-7349](https://iguazio.atlassian.net/browse/ML-7349), which aligns the labels in schedules and schedule objects, we need to account for objects created prior to this change. To address this, a new data migration (version 8) has been added. This migration retrieves all schedules from the database and aligns their labels accordingly.

Jira - https://iguazio.atlassian.net/browse/ML-7914

[ML-7349]: https://iguazio.atlassian.net/browse/ML-7349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/api/initial_data.py
+++ b/server/api/initial_data.py
@@ -35,6 +35,7 @@ import server.api.db.sqldb.models
 import server.api.utils.db.alembic
 import server.api.utils.db.backup
 import server.api.utils.db.mysql
+import server.api.utils.scheduler
 from mlrun.artifacts.base import fill_artifact_object_hash
 from mlrun.config import config
 from mlrun.errors import MLRunPreconditionFailedError, err_to_str
@@ -245,6 +246,8 @@ def _perform_data_migrations(db_session: sqlalchemy.orm.Session):
                 _perform_version_6_data_migrations(db, db_session)
             if current_data_version < 7:
                 _perform_version_7_data_migrations(db, db_session)
+            if current_data_version < 8:
+                _perform_version_8_data_migrations(db, db_session)
 
             db.create_data_version(db_session, str(latest_data_version))
 
@@ -877,6 +880,12 @@ def _perform_version_7_data_migrations(
     _create_project_summaries(db, db_session)
 
 
+def _perform_version_8_data_migrations(
+    db: server.api.db.sqldb.db.SQLDB, db_session: sqlalchemy.orm.Session
+):
+    _align_schedule_labels(db, db_session)
+
+
 def _create_project_summaries(db, db_session):
     # Create a project summary record for all projects.
     # We need to create them manually because a summary record is created only when a new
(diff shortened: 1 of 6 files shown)


server/api/initial_data.py
```diff
@@ -894,6 +903,34 @@ def _create_project_summaries(db, db_session):
...
+def _align_schedule_labels(db, db_session):
+    db_schedules: list[mlrun.common.schemas.ScheduleRecord] = db.list_schedules(
+        session=db_session
+    )
+    schedules_update = []
+    for db_schedule in db_schedules:
+        # convert list[LabelRecord] to dict
+        db_schedule_labels = {label.name: label.value for label in db_schedule.labels}
+        # merging labels
+        merged_labels = server.api.utils.scheduler.Scheduler()._merge_schedule_and_schedule_object_labels(
```
alonmr: Perhaps use `ensure_scheduler` and `get_scheduler` just for consistency

server/api/initial_data.py
```diff
@@ -894,6 +903,34 @@ def _create_project_summaries(db, db_session):
...
+            db_session, db_schedule.project, db_schedule.name
+        )
+        db._update_schedule_body(
+            schedule=schedule,
+            scheduled_object=db_schedule.scheduled_object,
+            labels=merged_labels,
+        )
+        schedules_update.append(schedule)
+
+    db._upsert(db_session, schedules_update)
```
quaark: 1. This method is very expensive. It is going to preform `O(n)` queries for each schedule in the DB. I suggest adding a `as_records: bool = False,` param to `list_schedules` (similar to `list_artifacts`). You can still extract the schedule struct with `_transform_schedule_record_to_scheme` but then you don't need to re-query for each record after getting them all.
2. This method is using loads of the `db`'s private methods, I don't think we should make them public, but maybe this method should be part of the `sqldb` interface?

quaark (changes requested):
Looks good! Got a couple of comments on the alignment implementation

server/api/initial_data.py
```diff
@@ -894,6 +903,34 @@ def _create_project_summaries(db, db_session):
     db._upsert(db_session, project_summaries, ignore=True)
 
 
+def _align_schedule_labels(db, db_session):
+    db_schedules: list[mlrun.common.schemas.ScheduleRecord] = db.list_schedules(
+        session=db_session
+    )
+    schedules_update = []
+    for db_schedule in db_schedules:
```
alonmr: I'm thinking that we need to optimize this for scale systems. I would take the following steps:
1. Add `as_records` flag similar to `list_artifacts`.
2. Use sqlalchemy lazy load - return the query and then for loop on the query instead of using query.all() and loading all schedules in memory.
3. Transform the DB schedule to schema schedule then you have both and you don't need to get the DB one (saves us O(n) queries to the DB/cache).
```suggestion
    schedules_update = []
    for db_schedule in db.list_schedules(session=db_session, as_records=True):
        schedule_record = db._transform_schedule_record_to_scheme(db_schedule)
```

tests/api/test_initial_data.py
```diff
@@ -228,6 +237,39 @@ def test_create_project_summaries():
     assert migrated_project_summary.name == project.metadata.name
 
 
+def test_align_schedule_labels():
```
alonmr: please add some more schedules with edge cases. we need to make sure this will not break in field

alonmr (commented):
Very well done! A few improvement suggestions

rokatyy:
@quaark @alonmr Great minds think alike :) 

Thanks for suggestions!  Have fixed them + also moved code from scheduler to helpers to enable access from both scheduler and sql avoiding circular imports.

server/api/utils/helpers.py
```diff
@@ -165,3 +165,96 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     return decorator
+
+
+def set_scheduled_object_labels(
```
alonmr, line 170: I think this methods can be `staticmethods` of the Scheduler class it would make more sense to me

alonmr (approved):
Very well done! Minor suggestion

server/api/utils/helpers.py
rokatyy: @alonmr then we will have circular imports ad this method is used in `helpers` below.

liranbg (approved):

PR #6390, opened 2024-09-23
[MySQL] Update the DSN regex to also support an optional password

moranbental:
To run MLRun using an external MySQL database, we need to include the password in the `MLRUN_HTTPDB__DSN` configuration. The format should be:
`MLRUN_HTTPDB__DSN=mysql+pymysql://root:<password>@<ip>:<port>/mlrun`
However, when using the internal MySQL, the password is optional, and the value can simply be:
`MLRUN_HTTPDB__DSN=mysql+pymysql://root@<ip>:<port>/mlrun`

https://iguazio.atlassian.net/browse/ML-6808


tests/api/utils/db/test_mysql.py
```diff
@@ -0,0 +1,54 @@
...
+                "database": "mlrun",
+            },
+        ),
+        ("mysql+pymysql://@localhost:3307/mlrun", None),
+        ("mysql+pymysql://root:pass@localhost:3307", None),
+        ("sqlite:///db/mlrun.db?check_same_thread=false", None),
+    ],
+)
+def test_get_mysql_dsn_data(http_dsn: str, expected_output: dict):
+    os.environ["MLRUN_HTTPDB__DSN"] = http_dsn
```
liranbg: use pytest func to set envvar (so it wont override python's process env)

liranbg (changes requested):
overall lgtm, minor test fix is required

liranbg (approved):

PR #6428, opened 2024-09-25
[Pipelines] Fix log [1.7.x]

moranbental:
https://iguazio.atlassian.net/browse/ML-7958

### DIFF
--- a/mlrun/projects/pipelines.py
+++ b/mlrun/projects/pipelines.py
@@ -628,11 +628,13 @@ class _KFPRunner(_PipelineRunner):
 
     @staticmethod
     def wait_for_completion(run, project=None, timeout=None, expected_statuses=None):
+        project_name = project.metadata.name if project else ""
         logger.info(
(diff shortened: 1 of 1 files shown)


mlrun/projects/pipelines.py
```diff
@@ -628,11 +628,13 @@ def run(
 
     @staticmethod
     def wait_for_completion(run, project=None, timeout=None, expected_statuses=None):
```
liranbg:
```suggestion
    def wait_for_completion(run, project=None, timeout=None, expected_statuses=None):
```
I think it work updating the type hint as well

```
    def wait_for_completion(
        run: "_PipelineRunStatus",
        project: typing.Optional["mlrun.projects.MlrunProject"] = None,
        timeout: typing.Optional[int] = None,
        expected_statuses: list[str] = None,
    ):
```
wdyt

liranbg (approved):

PR #6439, opened 2024-09-29
[Workflows] Fix ability to opt-out node selector for workflow runner pod

yaelgen:
Optional way to resolve: https://iguazio.atlassian.net/browse/ML-7926

In the submit_job endpoint, `_fill_workflow_missing_fields_from_project` calls `_update_dict`, which removes keys with empty values, as unfilled fields are set to None during `WorkflowSpec` initialization.
For node selectors, None values are acceptable because they will be included in the resulting node selector of the job and take precedence, and `to_non_empty_values_dict` will remove them as expected.
This PR overrides the result of `_update_dict` in the NS field with the original value of `workflow_runner_node_selector` to maintain the intended behavior for node selectors.


liranbg (changes requested):
add ut coverage and we can safely merge it

alonmr (approved):
LGTM

liranbg (approved):

PR #6440, opened 2024-09-29
[Background Task] Reverting the check for resource permissions on background tasks [1.7.x]

moranbental:
When a project is deleted, its associated policies are also removed, leading to "access denied" errors for users trying to access the background task. 
https://iguazio.atlassian.net/browse/ML-7951

related issue 
https://iguazio.atlassian.net/browse/ML-7484

### DIFF
--- a/server/api/api/endpoints/background_tasks.py
+++ b/server/api/api/endpoints/background_tasks.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import asyncio
 import datetime
 import typing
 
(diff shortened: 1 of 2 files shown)


server/api/api/endpoints/background_tasks.py
```diff
@@ -199,14 +200,22 @@ async def _authorize_get_background_task_request(
     background_task: mlrun.common.schemas.BackgroundTask,
     auth_info: mlrun.common.schemas.AuthInfo,
 ):
+    return await asyncio.sleep(1)
```
liranbg:
```suggestion
    return await asyncio.sleep(0)
```

but actually perhaps you can simply return without even sleep

liranbg (approved):

PR #6451, opened 2024-09-29
[Notifications] Resolve notification params from the db [1.7.x]

roei3000b:
On this [pr](https://github.com/mlrun/mlrun/pull/4334/files) we exclude notifications params.
Because of that, the `workflow-runner` pod which should send the `start running` notification didn't send the starting message (the notifications on `MLRUN_EXEC_CONFIG` didn't include the params as I mentioned above).
Here we introduce solution to this problem by getting the notifications from the api (with new `RunFormat.notifications`) which should contains also the params that is missing.

image

### DIFF
--- a/mlrun/common/formatters/run.py
+++ b/mlrun/common/formatters/run.py
@@ -22,5 +22,8 @@ class RunFormat(ObjectFormat, mlrun.common.types.StrEnum):
     # No enrichment, data is pulled as-is from the database.
     standard = "standard"
 
+    # Performs run enrichment, including the run's artifacts..
+    notifications = "notifications"
+
     # Performs run enrichment, including the run's artifacts. Only available for the `get` run API.
     full = "full"
--- a/mlrun/common/formatters/run.py
+++ b/mlrun/common/formatters/run.py
@@ -22,7 +22,7 @@ class RunFormat(ObjectFormat, mlrun.common.types.StrEnum):
     # No enrichment, data is pulled as-is from the database.
     standard = "standard"
 
-    # Performs run enrichment, including the run's artifacts.
+    # Enrich run with full notifications since the notification params are subtracted from the run body.
     notifications = "notifications"
 
     # Performs run enrichment, including the run's artifacts. Only available for the `get` run API.
--- a/mlrun/common/formatters/run.py
+++ b/mlrun/common/formatters/run.py
@@ -22,7 +22,7 @@ class RunFormat(ObjectFormat, mlrun.common.types.StrEnum):
     # No enrichment, data is pulled as-is from the database.
     standard = "standard"
 
-    # Performs run enrichment, including the run's artifacts..
+    # Performs run enrichment, including the run's artifacts.
     notifications = "notifications"
 
     # Performs run enrichment, including the run's artifacts. Only available for the `get` run API.
--- a/mlrun/execution.py
+++ b/mlrun/execution.py
@@ -24,6 +24,7 @@ from dateutil import parser
 
 import mlrun
 import mlrun.common.constants as mlrun_constants
+import mlrun.common.formatters
 from mlrun.artifacts import ModelArtifact
 from mlrun.datastore.store_resources import get_store_resource
 from mlrun.errors import MLRunInvalidArgumentError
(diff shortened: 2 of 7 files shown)


server/api/db/sqldb/db.py
```diff
@@ -335,14 +336,18 @@ def update_runs_requested_logs(
         )
         session.commit()
 
-    def read_run(self, session, uid, project=None, iter=0):
+    def read_run(self, session, uid, project=None, iter=0, with_notifications=False):
         project = project or config.default_project
         run = self._get_run(session, uid, project, iter)
```
quaark: probably a bug :) 
```suggestion
        run = self._get_run(session, uid, project, iter, with_notifications=with_notifications)
```

quaark (changes requested):
Looks good!
Think there's a bug in the code, probably why the test is failing as well

mlrun/common/formatters/run.py
```diff
@@ -22,5 +22,8 @@ class RunFormat(ObjectFormat, mlrun.common.types.StrEnum):
     # No enrichment, data is pulled as-is from the database.
     standard = "standard"
 
+    # Performs run enrichment, including the run's artifacts..
```
alonmr: Fix comment
> including the run's artifacts..

mlrun/projects/pipelines.py
```diff
@@ -1074,10 +1075,18 @@ def load_and_run(
         return
 
     # extract "start" notification if exists
+    run_db = mlrun.get_run_db()
+
+    # Here we want to get the run object with the notifications because we want to also get the notifications params
+    # (the context.get_notifications() doesn't include notifications params,
+    # see https://github.com/mlrun/mlrun/pull/4334/files)
```
alonmr: Not a fan of long comments. PR can be added in the description instead.
```suggestion
    # Get the full notifications from the DB since the run context does not contain the params due to bloating
```

mlrun/common/formatters/run.py
```diff
@@ -22,5 +22,8 @@ class RunFormat(ObjectFormat, mlrun.common.types.StrEnum):
     # No enrichment, data is pulled as-is from the database.
     standard = "standard"
 
+    # Performs run enrichment, including the run's artifacts.
```
alonmr:
```suggestion
    # Enrich run with full notifications since the notification params are subtracted from the run body.
```

mlrun/projects/pipelines.py
```diff
@@ -1074,10 +1075,18 @@ def load_and_run(
...
+    run = run_db.read_run(
+        context.uid, format_=mlrun.common.formatters.RunFormat.notifications
+    )
```
alonmr: why not move this into `context.get_notifications()` ?

server/api/crud/runs.py
```diff
@@ -125,7 +125,15 @@ def get_run(
...
         run = server.api.utils.singletons.db.get_db().read_run(
-            db_session, uid, project, iter
+            db_session,
+            uid,
+            project,
+            iter,
+            with_notifications=format_
+            in [
+                mlrun.common.formatters.RunFormat.notifications,
+                mlrun.common.formatters.RunFormat.full,
```
alonmr: for 1.7 I don't see a reason to make full also return notifications I think it adds risk because full is used in many places. we should probably change these places to use a new "enrich just artifacts" format

alonmr (changes requested):
Looks nice! A few comments

mlrun/execution.py
```diff
@@ -928,9 +929,16 @@ def set_hostname(self, host: str):
 
     def get_notifications(self):
         """Get the list of notifications"""
+
+        run_db = mlrun.get_run_db()
+
+        # Get the full notifications from the DB since the run context does not contain the params due to bloating
+        run = run_db.read_run(
```
alonmr:
```suggestion
        run = self._rundb.read_run(
```

mlrun/projects/pipelines.py
```diff
@@ -25,6 +25,7 @@
 import mlrun_pipelines.utils
 
 import mlrun
+import mlrun.common.formatters
```
alonmr: needed?

server/api/crud/runs.py
```diff
@@ -124,8 +124,15 @@ def get_run(
         format_: mlrun.common.formatters.RunFormat = mlrun.common.formatters.RunFormat.full,
     ) -> dict:
         project = project or mlrun.mlconf.default_project
+
+        # TODO: 1.8 - add notification also when format is full.
```
alonmr:
```suggestion
        # TODO: 1.8 - add notifications for full format as well.
```

alonmr (changes requested):
Minor

alonmr (approved):

liranbg (approved):

PR #6488, opened 2024-10-07
[Notifications] Mask notification secret params on submit_workflow [1.7.x]

roei3000b:
We need to mask  notification secret params on submit_workflow so we can store the notification secret params as k8s secret.

https://iguazio.atlassian.net/browse/ML-8029

### DIFF
--- a/server/api/api/utils.py
+++ b/server/api/api/utils.py
@@ -223,22 +223,58 @@ def mask_notification_params_on_task(
     action: server.api.constants.MaskOperations,
 ):
     """
+    Mask notification config params from the task dictionary
+    :param task:    The task object to mask
+    :param action:  The masking operation to perform on the notification config params (conceal/redact)
+    """
+    run_uid = get_in(task, "metadata.uid")
+    project = get_in(task, "metadata.project")
+    notifications = task.get("spec", {}).get("notifications", [])
+    if notifications:
+        task.setdefault("spec", {})["notifications"] = _mask_notifications_params(
+            run_uid,
+            project,
+            [
+                mlrun.model.Notification.from_dict(notification)
+                for notification in notifications
+            ],
+            action,
+        )
+
+
+def mask_notification_params_on_task_object(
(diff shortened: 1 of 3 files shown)


server/api/api/utils.py
```diff
@@ -223,22 +223,58 @@ def mask_notification_params_on_task(
...
     notifications = task.get("spec", {}).get("notifications", [])
-    masked_notifications = []
     if notifications:
-        for notification in notifications:
-            notification_object = mlrun.model.Notification.from_dict(notification)
-            masked_notifications.append(
-                mask_op(project, run_uid, notification_object).to_dict()
-            )
-        task.setdefault("spec", {})["notifications"] = masked_notifications
+        task.setdefault("spec", {})["notifications"] = _mask_notifications_params(
```
theSaarco: Looks like in this flow you're missing a `to_dict()` on the masked notifications - before your change the code would do `mask_op(...).to_dict()`, which is something you don't do now.

theSaarco (changes requested):
Looks fine, but there's a missing conversion in one flow.

server/api/api/utils.py
roei3000b: You right, thanks 👍

liranbg (approved):

theSaarco (approved):
LGTM.

server/api/crud/workflows.py
```diff
@@ -233,6 +240,11 @@ def run(
             run_name=runner.metadata.name,
             load_only=load_only,
         )
+        # We want to store the secret params as k8s secret, so later we can access them with the project internal secret
+        # key that was created.
+        server.api.api.utils.mask_notification_params_on_task_object(
+            run_spec, server.api.constants.MaskOperations.CONCEAL
+        )
```
alonmr, line 247: there is some common code here together with node selector enrichment with the schedule flow can be merged to method

PR #6493, opened 2024-10-08
[Logs] Return log collection legacy fallback and fix periodic collection

alonmr:
1. Revert the legacy log collection fallback in case periodic job malfunctions.
2. Change get runs filter for periodic collection to get last 4 hours so that it doesn't collect logs for removed pods

https://iguazio.atlassian.net/browse/ML-8027

### DIFF
--- a/server/api/main.py
+++ b/server/api/main.py
@@ -417,16 +417,18 @@ async def _initiate_logs_collection(start_logs_limit: asyncio.Semaphore):
         ) - datetime.timedelta(
             seconds=int(config.runtime_resources_deletion_grace_period)
         )
+
+        # aborted means the pods were deleted and logs were already fetched
+        run_states = mlrun.common.runtimes.constants.RunStates.terminal_states()
+        run_states.remove(mlrun.common.runtimes.constants.RunStates.aborted)
         runs_uids.extend(
             await fastapi.concurrency.run_in_threadpool(
                 get_db().list_distinct_runs_uids,
                 db_session,
                 requested_logs_modes=[False],
-                # get only uids as there might be many runs which reached terminal state while the API was down, the
-                # run objects will be fetched in the next step
                 only_uids=True,
                 last_update_time_from=last_update_time,
-                states=mlrun.common.runtimes.constants.RunStates.terminal_states(),
+                states=run_states,
             )
         )
         if runs_uids:
--- a/server/api/main.py
+++ b/server/api/main.py
@@ -403,7 +403,7 @@ async def _initiate_logs_collection(start_logs_limit: asyncio.Semaphore):
     """
     db_session = await fastapi.concurrency.run_in_threadpool(create_session)
     try:
-        # list all the runs in the system which we didn't request logs collection for yet
+        # list all the runs currently still running in the system which we didn't request logs collection for yet
         runs_uids = await fastapi.concurrency.run_in_threadpool(
             get_db().list_distinct_runs_uids,
             db_session,
@@ -418,7 +418,8 @@ async def _initiate_logs_collection(start_logs_limit: asyncio.Semaphore):
             seconds=int(config.runtime_resources_deletion_grace_period)
         )
 
-        # aborted means the pods were deleted and logs were already fetched
+        # Add all the completed/failed runs in the system which we didn't request logs collection for yet.
+        # Aborted means the pods were deleted and logs were already fetched.
         run_states = mlrun.common.runtimes.constants.RunStates.terminal_states()
         run_states.remove(mlrun.common.runtimes.constants.RunStates.aborted)
         runs_uids.extend(
--- a/server/api/main.py
+++ b/server/api/main.py
@@ -403,17 +403,31 @@ async def _initiate_logs_collection(start_logs_limit: asyncio.Semaphore):
     """
     db_session = await fastapi.concurrency.run_in_threadpool(create_session)
     try:
-        # we don't want initiate logs collection for aborted runs
-        run_states = mlrun.common.runtimes.constants.RunStates.all()
-        run_states.remove(mlrun.common.runtimes.constants.RunStates.aborted)
-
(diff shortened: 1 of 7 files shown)


server/api/main.py
```diff
@@ -403,17 +403,31 @@ async def _initiate_logs_collection(start_logs_limit: asyncio.Semaphore):
...
+        runs_uids.extend(
+            await fastapi.concurrency.run_in_threadpool(
+                get_db().list_distinct_runs_uids,
+                db_session,
+                requested_logs_modes=[False],
+                # get only uids as there might be many runs which reached terminal state while the API was down, the
+                # run objects will be fetched in the next step
+                only_uids=True,
+                last_update_time_from=last_update_time,
+                states=mlrun.common.runtimes.constants.RunStates.terminal_states(),
```
roei3000b: Remove the aborted status as it was before

server/api/runtime_handlers/base.py
```diff
@@ -1631,6 +1637,25 @@ def _get_run_label_selector(project: str, run_uid: str):
...
+            logs_from_k8s = crud.Logs()._get_logs_legacy_method(
+                db_session, project, uid, source=LogSources.K8S, run=run
+            )
+            if logs_from_k8s:
+                logger.info("Storing run logs", project=project, uid=uid)
+                server.api.crud.Logs().store_log(
+                    logs_from_k8s, project, uid, append=False
+                )
+                # Tell the periodic log collection to not request logs (this assumes the run is in terminal state)
+                db.update_runs_requested_logs(db_session, [uid], requested_logs=True)
```
roei3000b: 1. Why do we assume that the run is in terminal state and not check it?
2. Why do we update the requested_logs only if we have logs from k8s? don't we need to update it also if we doesn't have logs from k8s?

roei3000b (commented):
Loogs great!
Some questions.

server/api/runtime_handlers/base.py
alonmr: 1. We already check it outside of the function but I guess we can check it here as well to make sure - added
2. We let the periodic log collection job handle it because it will mark it as requested logs after `max_consecutive_start_log_requests` threshold passed

roei3000b (approved):
LGTM

server/api/main.py
```diff
@@ -403,17 +403,33 @@ async def _initiate_logs_collection(start_logs_limit: asyncio.Semaphore):
...
+            states=mlrun.common.runtimes.constants.RunStates.non_terminal_states(),
+        )
+
+        last_update_time = datetime.datetime.now(
+            datetime.timezone.utc
+        ) - datetime.timedelta(
+            seconds=int(config.runtime_resources_deletion_grace_period)
+        )
+
+        # aborted means the pods were deleted and logs were already fetched
```
quaark: I'd add another comment here to make it clear why we have this extra query:
```suggestion
        # add all the completed/failed runs in the system which we didn't request logs collection for yet.
        # aborted means the pods were deleted and logs were already fetched.
```

And I'd add to the previous query a comment saying it's querying the still running runs (added a separate comment there)

server/api/main.py
```diff
@@ -403,17 +403,33 @@ async def _initiate_logs_collection(start_logs_limit: asyncio.Semaphore):
     """
     db_session = await fastapi.concurrency.run_in_threadpool(create_session)
     try:
-        # we don't want initiate logs collection for aborted runs
-        run_states = mlrun.common.runtimes.constants.RunStates.all()
-        run_states.remove(mlrun.common.runtimes.constants.RunStates.aborted)
-
         # list all the runs in the system which we didn't request logs collection for yet
```
quaark:
```suggestion
        # list all the runs currently still running in the system which we didn't request logs collection for yet
```

server/api/runtime_handlers/base.py
```diff
@@ -1631,6 +1637,28 @@ def _get_run_label_selector(project: str, run_uid: str):
...
+                logger.info("Storing run logs", project=project, uid=uid)
+                server.api.crud.Logs().store_log(
+                    logs_from_k8s, project, uid, append=False
+                )
+                if run.get("status", {}).get("state") in RunStates.terminal_states():
+                    # Tell the periodic log collection to not request logs
+                    db.update_runs_requested_logs(
+                        db_session, [uid], requested_logs=True
+                    )
+
```
quaark: This is mostly a styling thing, but IMO, I usually prefer methods like these with many `if` indentations to do `if not`s and return, to make the method slightly easier to read.
Also, since we are re-introducing this mainly for the fallback I'd rewrite the BC comment to emphasize the fallback first and then BC.
```suggestion
    @staticmethod
    def _ensure_run_logs_collected(
        db: DBInterface, db_session: Session, project: str, uid: str, run: dict = None
    ):
        # We use this method as a fallback in case the periodic collect job malfunctions,
        # and also for backwards compatibility in case we would not use the log collector but rather
        # the legacy method to pull logs.
        log_file_exists, _ = crud.Logs().log_file_exists_for_run_uid(project, uid)
        if log_file_exists:
            # nothing to ensure
            return
        
       
        logs_from_k8s = crud.Logs()._get_logs_legacy_method(
            db_session, project, uid, source=LogSources.K8S, run=run
        )
        if not logs_from_k8s:
            # no log to store
            return
        
        logger.info("Storing run logs", project=project, uid=uid)
        server.api.crud.Logs().store_log(
            logs_from_k8s, project, uid, append=False
        )
        if run.get("status", {}).get("state") in RunStates.terminal_states():
            # Tell the periodic log collection to not request logs
            db.update_runs_requested_logs(
                db_session, [uid], requested_logs=True
            )
```

quaark (changes requested):
Looks good!
Left a couple of comments mainly relating to readability - to ensure we understand why we have this behaviour still.

quaark (approved):

PR #6497, opened 2024-10-08
[KFP] If we got empty secret value skip the replacing of the env vars [1.7.x]

roei3000b:
In case we got empty secret value, we should skip the `_create_secret_env_var_for_pipeline`


adamdelman (approved):
lgtm

alonmr (approved):

PR #6511, opened 2024-10-09
[K8S] Fix K8S client error handling when storing secrets [1.7.x]

adamdelman:
1. Fix the exception handling for the K8S client was not working correctly.
This is due to trying to catch exceptions inheriting from DynamicApiError, which require a call to
kubernetes.dynamic.exceptions.api_exception to convert to the correct exception class.
2. Improve the cache mechanism used in the secret creation, preventing redundant API calls.
3. Add unit tests for K8sHelper

https://iguazio.atlassian.net/browse/ML-8041

### DIFF
--- a/server/api/utils/singletons/k8s.py
+++ b/server/api/utils/singletons/k8s.py
@@ -599,6 +599,7 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
         type_: str = SecretTypes.opaque,
         labels: typing.Optional[dict] = None,
     ):
+        logger.info("Creating secret", secret_name=secret_name)
         k8s_secret = client.V1Secret(
             type=type_,
             metadata=client.V1ObjectMeta(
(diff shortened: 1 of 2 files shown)


server/api/utils/singletons/k8s.py
```diff
@@ -613,14 +614,15 @@ def _create_secret(
...
             # There was a conflict while we tried to create the secret.
-            logger.warning(
-                "Secret was created while we tried to create it,",
-                secret_name=k8s_secret.metadata.name,
-                exc=mlrun.errors.err_to_str(exc),
-            )
-            raise
+            if isinstance(exc, k8s_dynamic_exceptions.ConflictError):
+                logger.warning(
+                    "Secret was created while we tried to create it,",
```
alonmr:
```suggestion
                    "Failed to create secret, Secret might have been created while we tried to create it",
```

server/api/utils/singletons/k8s.py
```diff
@@ -599,6 +599,7 @@ def _create_secret(
         type_: str = SecretTypes.opaque,
         labels: typing.Optional[dict] = None,
     ):
+        logger.info("Creating secret", secret_name=secret_name)
```
alonmr: These logs should be debug level IMO

alonmr (changes requested):
A few requests:
1. can you explain what was the original issue and how you solved it? Code is a but hard to follow I don't understand why you needed to change the exception type.
2. Please add a ut in `tests/api/utils/singletons/test_k8s_utils.py`
3. We usually do `exc.status == 40x` to check the api exception kind and I'm OK if you want to convert it to the K8s SDK way just make sure this is properly tested since it is for 1.7
4. Add [1.7.x] in the title :)

PR #6513, opened 2024-10-09
[Notifications] Fill notification params from project secrets [1.7.x]

roei3000b:
Fixing bug that we didn't send `running` notification when webhook defined on the `secret_params` of the notification.
The `workflow-runner` should send the `running` notification when start running the workflow.
We got the notifications for the specific run from the db (see [here](https://github.com/mlrun/mlrun/pull/6451))
When we define the webhook inside the `secret_params`, the entity in the db doesn't contains the secret param itself but reference to the **internal project secret** that was created. Currently, the `workflow-runner` pod doesn't mount this internal project secret.

For getting the webhook(and the other secret params), for this version, we introduce the following workaround:
1. The user should create **project secret** the contains all the `secret_params` of that notification.
2. Define the notification with `secret_params={"secret": "<PROJECT_SECRET_NAME>"`}`

And then he should get the `running` notification.

In this solution, we read the notification from the db and filling the notification params from the project secret key that was provided as the secret_params.

https://iguazio.atlassian.net/browse/ML-8029

### DIFF
--- a/mlrun/execution.py
+++ b/mlrun/execution.py
@@ -934,10 +934,20 @@ class MLClientCtx:
         run = self._rundb.read_run(
             self.uid, format_=mlrun.common.formatters.RunFormat.notifications
         )
-        return [
-            mlrun.model.Notification.from_dict(notification)
-            for notification in run["spec"]["notifications"]
-        ]
+
+        notifications = []
+        for notification in run["spec"]["notifications"]:
+            notification: mlrun.model.Notification = mlrun.model.Notification.from_dict(
+                notification
+            )
+            # We fill the params from the secret params to allow the notification to get the project secret
+            # TODO: This is a workaround to allow the notification to get the secret params from project instead of
+            #       getting them from from the k8s secret. We should mount the internal project secret that was created
+            #       to the workflow-runner and get the secret from there.
+            notification.fill_params_from_secret_params()
+            notifications.append(notification)
+
+        return notifications
 
     def to_dict(self):
         """Convert the run context to a dictionary"""
--- a/mlrun/execution.py
+++ b/mlrun/execution.py
@@ -940,8 +940,8 @@ class MLClientCtx:
             notification: mlrun.model.Notification = mlrun.model.Notification.from_dict(
                 notification
             )
-            # We fill the secret params from the project secret params to allow the  get what it need for sending
-            # the notification(the webhook for example).
+            # Fill the secret params from the project secret. We cannot use the server side internal secret mechanism
+            # here as it is the client side.
             # TODO: This is a workaround to allow the notification to get the secret params from project secret
             #       instead of getting them from the internal project secret that should be mounted.
             #       We should mount the internal project secret that was created to the workflow-runner
--- a/mlrun/execution.py
+++ b/mlrun/execution.py
@@ -943,13 +943,13 @@ class MLClientCtx:
             # Fill the secret params from the project secret. We cannot use the server side internal secret mechanism
             # here as it is the client side.
             # TODO: This is a workaround to allow the notification to get the secret params from project secret
-            #       instead of getting them from the internal project secret that should be mounted.
+            #       instead of getting them from the intmernal project secret that should be mounted.
             #       We should mount the internal project secret that was created to the workflow-runner
             #       and get the secret from there.
             try:
                 notification.fill_secret_params_from_project_secret()
                 notifications.append(notification)
-            except Exception:
+            except mlrun.errors.MLRunValueError:
                 logger.warning(
                     "Failed to fill secret params from project secret for notification."
                     "Skip this notification.",
--- a/mlrun/execution.py
+++ b/mlrun/execution.py
@@ -940,12 +940,13 @@ class MLClientCtx:
             notification: mlrun.model.Notification = mlrun.model.Notification.from_dict(
                 notification
             )
-            # We fill the params from the secret params to allow the notification to get the project secret
-            # TODO: This is a workaround to allow the notification to get the secret params from project instead of
-            #       getting them from the internal project secret that should be mounted.
+            # We fill the secret params from the project secret params to allow the  get what it need for sending
+            # the notification(the webhook for example).
+            # TODO: This is a workaround to allow the notification to get the secret params from project secret
+            #       instead of getting them from the internal project secret that should be mounted.
             #       We should mount the internal project secret that was created to the workflow-runner
             #       and get the secret from there.
-            notification.fill_params_from_secret_params()
+            notification.fill_secret_params_from_project_secret()
             notifications.append(notification)
 
         return notifications
--- a/mlrun/execution.py
+++ b/mlrun/execution.py
@@ -946,8 +946,15 @@ class MLClientCtx:
             #       instead of getting them from the internal project secret that should be mounted.
             #       We should mount the internal project secret that was created to the workflow-runner
             #       and get the secret from there.
-            notification.fill_secret_params_from_project_secret()
-            notifications.append(notification)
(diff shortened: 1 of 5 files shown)


mlrun/execution.py
```diff
@@ -934,10 +934,20 @@ def get_notifications(self):
...
+        notifications = []
+        for notification in run["spec"]["notifications"]:
+            notification: mlrun.model.Notification = mlrun.model.Notification.from_dict(
+                notification
+            )
+            # We fill the params from the secret params to allow the notification to get the project secret
+            # TODO: This is a workaround to allow the notification to get the secret params from project instead of
+            #       getting them from from the k8s secret. We should mount the internal project secret that was created
+            #       to the workflow-runner and get the secret from there.
+            notification.fill_params_from_secret_params()
```
theSaarco: The comment is not entirely correct - when mounting the internal project secret we will still do the same here and call the same function. The difference is that it will also work for the general case since the internal secret will be mounted as env variable. We won't get it from the k8s secret directly.

theSaarco (approved):
Looks good, might want to clarify the comment a bit.

mlrun/execution.py
roei3000b: You right, thanks for the clarification!

mlrun/model.py
```diff
@@ -774,6 +774,22 @@ def validate_notification_params(self):
 
         notification_class.validate_params(secret_params | params)
 
+    def fill_params_from_secret_params(self):
+        """
+        Fill the notification params from the secret_params.
+        """
+        if "secret" in self.secret_params:
+            secret = self.secret_params["secret"]
```
```suggestion
        secret = self.secret_params.get("secret")
        if secret:
```

mlrun/model.py
```diff
@@ -774,6 +774,22 @@ def validate_notification_params(self):
...
+            secret_value = mlrun.get_secret_or_env(secret)
+            if secret_value:
+                try:
+                    self.params.update(json.loads(secret_value))
```
alonmr: this worries me. you are taking a secret and saving it in plaintext params. what if we then save the notification? the secret will not be masked. I'm looking at the `SlackNotification` class and I see it gets the secret as you did here but it doesn't save it to params. Perhaps the change should be there?

alonmr (changes requested):
Have an issue with saving the secret to params. see below 👇 perhaps @quaark or @Yacouby can shed light on how this should be done

mlrun/model.py
roei3000b: [Here](https://github.com/mlrun/mlrun/blob/74cc3e1bd4a2fb008945de5c8aa6ee51d34bf54e/mlrun/utils/notifications/notification_pusher.py#L221) we also update params from secret_params, so should we also change this behavior?

theSaarco: In general, I wouldn't do this change in the `SlackNotification` class since it's just one type of notification and we need to support secrets for all of the types.
The classes that inherit from `NotificationBase` do not have `secret_params` at all - they only have `params` and they are not meant to be serialized, but rather just used for sending the notifications. I believe they are only used in the `NotificationPusher`. 
So, maybe this part should be done when they (the `NotificationBase` classes) are created - I'm not sure exactly where that happens, I guess @quaark or @Yacouby can indeed help there.

mlrun/execution.py
```diff
@@ -940,12 +940,13 @@ def get_notifications(self):
             notification: mlrun.model.Notification = mlrun.model.Notification.from_dict(
                 notification
             )
-            # We fill the params from the secret params to allow the notification to get the project secret
-            # TODO: This is a workaround to allow the notification to get the secret params from project instead of
-            #       getting them from the internal project secret that should be mounted.
+            # We fill the secret params from the project secret params to allow the  get what it need for sending
+            # the notification(the webhook for example).
```
quaark:
```suggestion
            # Fill the secret params from the project secret. We cannot use the server side internal secret mechanism
            # here as it is the client side.
```

mlrun/model.py
```diff
@@ -774,20 +774,22 @@ def validate_notification_params(self):
...
                 try:
-                    self.params.update(json.loads(secret_value))
+                    self.secret_params = json.loads(secret_value)
                 except ValueError:
                     logger.warning(
-                        "Failed to load secret value to params",
-                        secret_value=secret_value,
+                        "Failed to load secret",
+                        secret=secret,
                     )
```
quaark: I think it's best to raise an exception here as this will make the notification fail to send.
It shouldn't cancel the workflow from running though so wrap the call to `get_notifications()` with a `try/except` and in that except log a warning that you failed to load the notification and don't send it.

quaark (changes requested):
Couple suggestions

mlrun/model.py
roei3000b: I don't think so, because sometime the webhook is not on the secret params but on the params.

quaark: Assume you need 100% of the configured params. If you are loading params from a secret and they exist but they are malformed, you cannot assume you have some other params. This is a bad state and you shouldn't try sending a notification with only half of its params.

mlrun/model.py
```diff
@@ -774,20 +774,22 @@ def validate_notification_params(self):
 
         notification_class.validate_params(secret_params | params)
 
-    def fill_params_from_secret_params(self):
+    def fill_secret_params_from_project_secret(self):
```
alonmr: rename to `unmask_notification_params_secret` it should match the server side function as it does a similar operation

quaark (approved):
🚀

mlrun/execution.py
```diff
@@ -940,12 +940,13 @@ def get_notifications(self):
...
+            # TODO: This is a workaround to allow the notification to get the secret params from project secret
+            #       instead of getting them from the internal project secret that should be mounted.
             #       We should mount the internal project secret that was created to the workflow-runner
             #       and get the secret from there.
-            notification.fill_params_from_secret_params()
+            notification.fill_secret_params_from_project_secret()
```
alonmr: I don't think this should be here tbh
We should unmask the secret just before we push it
@quaark wdyt?

alonmr (commented):
Looks better, couple of comments

mlrun/execution.py
```diff
@@ -940,14 +940,21 @@ def get_notifications(self):
             notification: mlrun.model.Notification = mlrun.model.Notification.from_dict(
                 notification
             )
-            # We fill the secret params from the project secret params to allow the  get what it need for sending
-            # the notification(the webhook for example).
+            # Fill the secret params from the project secret. We cannot use the server side internal secret mechanism
+            # here as it is the client side.
             # TODO: This is a workaround to allow the notification to get the secret params from project secret
-            #       instead of getting them from the internal project secret that should be mounted.
+            #       instead of getting them from the intmernal project secret that should be mounted.
```
alonmr:
```suggestion
            #       instead of getting them from the internal project secret that should be mounted.
```

mlrun/model.py
```diff
@@ -787,10 +787,7 @@ def fill_secret_params_from_project_secret(self):
                 try:
                     self.secret_params = json.loads(secret_value)
                 except ValueError:
-                    logger.warning(
-                        "Failed to load secret",
-                        secret=secret,
-                    )
+                    raise mlrun.errors.MLRunValueError("Failed to parse secret value")
```
```suggestion
                    raise mlrun.errors.MLRunValueError("Failed to parse secret value") from exc
```

mlrun/execution.py
quaark: agree

alonmr (approved):

quaark (approved):

PR #6514, opened 2024-10-09
[Artifact] Enhance artifact key validation regex

Yacouby:
This PR enhances the validation for artifact key and db_key in the backend by updating the regex pattern.
The artifact key needs to start and end with alphanumeric character, and can contain -_. in the middle. This is basically the same validation the UI currently has.

Existing artifacts will remain valid; this change applies only to newly created artifacts.

resolves:
https://iguazio.atlassian.net/browse/ML-5809

### DIFF
--- a/mlrun/artifacts/manager.py
+++ b/mlrun/artifacts/manager.py
@@ -246,6 +246,8 @@ class ArtifactManager:
                 # otherwise, we do not want to override it.
                 # this is mainly relevant for imported artifacts that have an explicit db_key value already set
                 db_key = item.db_key or key
+        if db_key != key:
+            validate_artifact_key_name(db_key, "artifact.db_key")
         item.db_key = db_key or ""
         item.viewer = viewer or item.viewer
         item.tree = producer.tag
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -268,16 +268,13 @@ def validate_tag_name(
 def validate_artifact_key_name(
     artifact_key: str, field_name: str, raise_on_failure: bool = True
 ) -> bool:
-    if field_name == "artifact.key":
-        type = "key"
-    else:
-        type = "db_key"
+    field_type = "key" if field_name == "artifact.key" else "db_key"
     return mlrun.utils.helpers.verify_field_regex(
         field_name,
         artifact_key,
         mlrun.utils.regex.artifact_key,
         raise_on_failure=raise_on_failure,
-        log_message=f"The artifact {type} must start and end with an alphanumeric character and may only contain "
+        log_message=f"Artifact {field_type} must start and end with an alphanumeric character, and may only contain "
         "letters, numbers, hyphens, underscores, and dots.",
     )
 
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -268,13 +268,17 @@ def validate_tag_name(
 def validate_artifact_key_name(
     artifact_key: str, field_name: str, raise_on_failure: bool = True
 ) -> bool:
+    if field_name == "artifact.key":
+        type = "key"
+    else:
+        type = "db_key"
     return mlrun.utils.helpers.verify_field_regex(
         field_name,
         artifact_key,
         mlrun.utils.regex.artifact_key,
         raise_on_failure=raise_on_failure,
-        log_message="The artifact key must start and end with an alphanumeric character and may only contain letters, "
-        "numbers, hyphens, underscores, and dots.",
+        log_message=f"The artifact {type} must start and end with an alphanumeric character and may only contain "
+        "letters, numbers, hyphens, underscores, and dots.",
     )
 
 
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -273,7 +273,8 @@ def validate_artifact_key_name(
         artifact_key,
         mlrun.utils.regex.artifact_key,
         raise_on_failure=raise_on_failure,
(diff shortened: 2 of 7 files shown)


moranbental (changes requested):
You are validating the key, but the ticket also mentions validating the db_key. 
The `validate_artifact_key_name` function should also be called for the `db_key`

mlrun/utils/helpers.py
```diff
@@ -268,12 +268,17 @@ def validate_tag_name(
 def validate_artifact_key_name(
     artifact_key: str, field_name: str, raise_on_failure: bool = True
 ) -> bool:
+    if field_name == "artifact.key":
+        type = "key"
+    else:
+        type = "db_key"
```
moranbental: type is built in function

```suggestion
   field_type = "key" if field_name == "artifact.key" else "db_key"
```

mlrun/artifacts/manager.py
```diff
@@ -246,6 +246,8 @@ def log_artifact(
                 # otherwise, we do not want to override it.
                 # this is mainly relevant for imported artifacts that have an explicit db_key value already set
                 db_key = item.db_key or key
+        if db_key != key:
```
moranbental, line 249: When `producer.kind == run` it will change the value of `db_key`. 
Please add a test to cover this issue

moranbental (changes requested):

tests/utils/test_helpers.py
```diff
@@ -321,11 +321,19 @@ def test_validate_tag_name(tag_name, expected):
...
         ),
         ("", pytest.raises(mlrun.errors.MLRunInvalidArgumentError)),
+        (
+            "artifact-name#",
+            pytest.raises(mlrun.errors.MLRunInvalidArgumentError),
+        ),
+        ("artifact@name", pytest.raises(mlrun.errors.MLRunInvalidArgumentError)),
+        ("artifact#name", pytest.raises(mlrun.errors.MLRunInvalidArgumentError)),
+        ("artifact-name#", pytest.raises(mlrun.errors.MLRunInvalidArgumentError)),
+        ("artifact:name", pytest.raises(mlrun.errors.MLRunInvalidArgumentError)),
```
moranbental, line 331: one more test case
`("artifact_name$", pytest.raises(MLRunInvalidArgumentError))`

tests/artifacts/test_artifacts.py
```diff
@@ -357,6 +357,45 @@ def test_log_artifact_with_target_path_and_upload_options(target_path, upload_op
...
+    "artifact_key,expected",
+    [
+        ("artifact@key", pytest.raises(mlrun.errors.MLRunInvalidArgumentError)),
+        ("artifact#key", pytest.raises(mlrun.errors.MLRunInvalidArgumentError)),
+        ("artifact!key", pytest.raises(mlrun.errors.MLRunInvalidArgumentError)),
+        ("artifact_key!", pytest.raises(mlrun.errors.MLRunInvalidArgumentError)),
+        ("artifact/key", pytest.raises(mlrun.errors.MLRunInvalidArgumentError)),
+        ("artifact_key123", does_not_raise()),
+        ("artifact-key", does_not_raise()),
+        ("artifact.key", does_not_raise()),
```
moranbental, line 370: Why include all these tests if they are already validated in t`est_validate_artifact_name`? 
I think you could use just one invalid artifact key here and modify `test_validate_artifact_name` to also test `db_key`, not just `artifact.key`

moranbental (changes requested):

tests/artifacts/test_artifacts.py
Yacouby: this test is intended to test the flow of log_artifact, so i think it is okay to test it with some invalid keys.
and also it tests log_artifact when db_key != key

moranbental (approved):
LGTM

tests/api/db/test_artifacts.py
```diff
@@ -556,6 +556,46 @@ def test_store_artifact_with_different_key(
...
+            )
+
+        # store the artifact with invalid db_key
+        artifact_invalid_db_key = "artifact#!key"
+        artifact_body["spec"]["db_key"] = artifact_invalid_db_key
+        with pytest.raises(mlrun.errors.MLRunInvalidArgumentError):
+            db.store_artifact(
+                db_session,
+                artifact_valid_key,
+                artifact_body,
```
TomerShor, line 583: doesn't the body still contain the invalid key?

mlrun/utils/helpers.py
```diff
@@ -268,12 +268,14 @@ def validate_tag_name(
...
     artifact_key: str, field_name: str, raise_on_failure: bool = True
 ) -> bool:
+    field_type = "key" if field_name == "artifact.key" else "db_key"
     return mlrun.utils.helpers.verify_field_regex(
         field_name,
         artifact_key,
         mlrun.utils.regex.artifact_key,
         raise_on_failure=raise_on_failure,
-        log_message="Slashes are not permitted in the artifact key (both \\ and /)",
+        log_message=f"The artifact {field_type} must start and end with an alphanumeric character and may only contain "
```
TomerShor:
```suggestion
        log_message=f"Artifact {field_type} must start and end with an alphanumeric character, and may only contain "
```

TomerShor (commented):
Minor

tests/api/db/test_artifacts.py
Yacouby: fixed

TomerShor (approved):

PR #6517, opened 2024-10-10
[K8S] Fix k8s api exception not converted to their status code dependent specific exception class [1.7.x]

adamdelman:
When creating a k8s secret and getting a conflict, the exception class was not correctly converted to the specific exception class for a conflict.
Also fix and relocate k8s to the correct file

### DIFF
--- a/server/api/utils/singletons/k8s.py
+++ b/server/api/utils/singletons/k8s.py
@@ -615,6 +615,7 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
                 body=k8s_secret,
             )
         except k8s_client_rest.ApiException as exc:
+            exc = k8s_dynamic_exceptions.api_exception(exc)
             # There was a conflict while we tried to create the secret.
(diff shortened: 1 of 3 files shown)


server/api/utils/singletons/k8s.py
```diff
@@ -615,14 +615,15 @@ def _create_secret(
...
+            exc = k8s_dynamic_exceptions.api_exception(exc)
             # There was a conflict while we tried to create the secret.
             if isinstance(exc, k8s_dynamic_exceptions.ConflictError):
                 logger.warning(
                     "Failed to create secret, Secret might have been created while we tried to create it",
                     secret_name=k8s_secret.metadata.name,
                     exc=mlrun.errors.err_to_str(exc),
                 )
-            raise k8s_dynamic_exceptions.api_exception(exc)
+            raise
```
alonmr: I'm not sure how this syntactic sugar works since you are reassigning `exc` I would be explicit here and change it to `raise exc` in case python keeps the original exception in some other variable

alonmr (approved):

PR #6520, opened 2024-10-13
[Tests] Add support for K8S client in system tests

adamdelman:
- Pass `kubeconfig-content` from GitHub Actions secrets to system tests.
- Modify `SystemTestPreparer` to handle `kubeconfig_content`.
- Update pytest session setup to handle missing or invalid Kubeconfig gracefully.


tests/system/conftest.py
```diff
@@ -27,20 +27,34 @@
 from pytest import CallInfo, ExitCode, Function, TestReport
 
 
+def missing_kubeclient(self):
+    raise AttributeError("Kubeclient was not setup and is unavailable")
```
alonmr: move below method that uses it (i.e. `pytest_sessionstart`)

tests/system/conftest.py
```diff
@@ -27,20 +27,34 @@
...
-    if not base64_kubeconfig_content:
-        raise ValueError("The Kubeconfig for the system test session was not provided.")
-    kubeconfig_content = base64.b64decode(base64_kubeconfig_content)
-    with NamedTemporaryFile() as tempfile:
-        tempfile.write(kubeconfig_content)
-        config.load_kube_config(
-            config_file=tempfile.name,
-        )
-        session.kube_client = k8s_client.CoreV1Api()
+    try:
```
alonmr, line 50: I would extract this to a separate method since it can be classified as "one of the things" that is done in `pytest_sessionstart` and not what `pytest_sessionstart` does.

tests/system/conftest.py
```diff
@@ -27,20 +27,34 @@
...
+        base64_kubeconfig_content = os.environ["SYSTEM_TEST_KUBECONFIG"]
+        kubeconfig_content = base64.b64decode(base64_kubeconfig_content)
+    except (ValueError, KeyError):
+        print("Kubeconfig was empty or invalid.")
```
alonmr: 1. Print the exception error message?
2. Any chance to use a logger here?

alonmr (approved):
LGTM. Some small flavor comments

tests/system/conftest.py
adamdelman: fixed

liranbg (approved):

alonmr (approved):

PR #6526, opened 2024-10-13
[Monitoring] Fix get or create run conflict attempt 2 [1.7.x]

alonmr:
During monitoring runs we try to create a run for an existing pod when we didn't find it by listing runs previously. This is prune to session isolation level errors as it may have been created after the transaction started and therefore we can't read it.
Trying to catch sql integrity errors and then getting the run instead of presumably committing the session to get a new snapshot / new transaction.

https://iguazio.atlassian.net/browse/ML-8028

### DIFF
--- a/server/api/db/base.py
+++ b/server/api/db/base.py
@@ -61,6 +61,16 @@ class DBInterface(ABC):
     ):
         pass
 
+    def create_or_get_run(
+        self,
+        session,
+        run_data: dict,
+        uid: str,
+        project: str = "",
+        iter: int = 0,
+    ):
+        pass
+
     @abstractmethod
     def update_run(self, session, updates: dict, uid, project="", iter=0):
         pass
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -227,20 +227,47 @@ class SQLDB(DBInterface):
                 start_time=run_start_time(run_data) or now,
                 requested_logs=False,
             )
-        self._ensure_run_name_on_update(run, run_data)
-        labels = run_labels(run_data)
-        self._update_run_state(run, run_data)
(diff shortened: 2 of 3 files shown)


server/api/db/sqldb/db.py
```diff
@@ -227,21 +227,48 @@ def store_run(
...
+            iteration=iter,
+            state=run_state(run_data),
+            start_time=run_start_time(run_data) or now,
+            requested_logs=False,
+        )
+        self._enrich_run_model(now, run, run_data)
+        try:
+            self._upsert(session, [run], silent=True)
+        except mlrun.errors.MLRunConflictError:
+            # Session was rollbacked and we now get a new snapshot
```
adamdelman, line 268: I'm not sure sqlalchemy really does a rollback on an exception, I think you might need to add         `session.rollback()`

adamdelman (changes requested):

server/api/db/base.py
```diff
@@ -61,6 +61,16 @@ def store_run(
     ):
         pass
 
+    def create_or_get_run(
```
moranbental, line 64: why not `get_or_create_run` like `get_or_create_project`

moranbental (changes requested):
Looks good! minor suggestion

server/api/db/sqldb/db.py
alonmr: we do the rollback in `_try_commit_obj`

server/api/db/base.py
alonmr: thanks. I wanted to emphasis that we 1st do a create and then we get the run so I reversed the name

moranbental (approved):

adamdelman (approved):

PR #6528, opened 2024-10-13
[Model Monitoring] Fix for checking infer data on the controller

Eyal-Danieli:
Resolved two issues introduced i https://iguazio.atlassian.net/browse/ML-8046 which were caused by the use of batch inference. These issues led to incorrect behavior: either the applications were triggered with no data, or they weren’t triggered at all when the MEP was deployed from a model server and invoked only by the batch inference process. For a detailed RCA, please refer to the ticket.

**Solution**
The controller will now check the infer parquet files directly, instead of querying the TSDB. This ensures that applications are triggered only when the model has actually been invoked.

### DIFF
--- a/mlrun/model_monitoring/controller.py
+++ b/mlrun/model_monitoring/controller.py
@@ -287,9 +287,11 @@ class MonitoringApplicationController:
         )
 
         self.model_monitoring_access_key = self._get_model_monitoring_access_key()
-        self.tsdb_connector = mlrun.model_monitoring.get_tsdb_connector(
-            project=self.project
-        )
+        self.storage_options = None
+        if not mlrun.mlconf.is_ce_mode():
+            self._initialize_v3io_configurations()
+        elif os.environ.get("MLRUN_ARTIFACT_PATH").startswith("s3://"):
+            self.storage_options = mlrun.mlconf.get_s3_storage_options()
 
     @staticmethod
     def _get_model_monitoring_access_key() -> Optional[str]:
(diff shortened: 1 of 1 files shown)


mlrun/model_monitoring/controller.py
```diff
@@ -415,12 +415,12 @@ def model_endpoint_process(
                 )
 
                 for start_infer_time, end_infer_time in batch_window.get_intervals():
-                    prediction_metric = tsdb_connector.read_predictions(
-                        endpoint_id=endpoint_id,
-                        start=start_infer_time,
-                        end=end_infer_time,
+                    df = m_fs.to_dataframe(
```
jond01, line 421: I assume you tested it in Iguazio and it works, but will it work without the `storage_options` in CE?

mlrun/model_monitoring/controller.py
```diff
@@ -286,9 +287,11 @@ def __init__(self) -> None:
         )
 
         self.model_monitoring_access_key = self._get_model_monitoring_access_key()
-        self.tsdb_connector = mlrun.model_monitoring.get_tsdb_connector(
-            project=self.project
-        )
+        self.storage_options = None
+        if not mlrun.mlconf.is_ce_mode():
+            self._initialize_v3io_configurations()
+        elif os.environ.get("MLRUN_ARTIFACT_PATH").startswith("s3://"):
```
jond01: I think this is more elegant:
```suggestion
        elif mlconf.artifact_path.startswith("s3://"):
```

Need to add of course
```py
from mlrun.config import config as mlconf
```

mlrun/model_monitoring/controller.py
```diff
@@ -298,6 +301,12 @@ def _get_model_monitoring_access_key() -> Optional[str]:
             access_key = mlrun.mlconf.get_v3io_access_key()
         return access_key
 
+    def _initialize_v3io_configurations(self) -> None:
+        self.storage_options = dict(
+            v3io_access_key=self.model_monitoring_access_key,
+            v3io_api=mlrun.mlconf.v3io_api,
+        )
```
jond01: Since V3IO worked even without it, is it even necessary?

jond01 (approved):

jond01 (approved):

PR #6532, opened 2024-10-14
[Docs] Running notification send only the default notification [1.7.x]

roei3000b:
When we send running notification it sent from the old notification mechanism where we cannot set the message and the severity.
Clarify it on the docs.

https://iguazio.atlassian.net/browse/ML-8063

### DIFF
--- a/docs/concepts/notifications.md
+++ b/docs/concepts/notifications.md
@@ -161,6 +161,8 @@ slack_notification = mlrun.model.Notification(
     secret_params={"secret": "SLACK_SECRET1"},
 )
 ```
+> ℹ️ **Info:** On running notification you can set only the webhook, and the message will be the default message.
+
 
 ### Remote pipeline notifications
 In remote pipelines, the pipeline end notifications are sent from the MLRun API. This means you don't need to watch the pipeline in order for its notifications to be sent.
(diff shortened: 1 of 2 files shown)


mlrun/projects/pipelines.py
```diff
@@ -593,7 +593,7 @@ def run(
             logger.warning(
                 "Setting notifications on kfp pipeline runner uses old notification behavior. "
                 "Notifications will only be sent if you wait for pipeline completion. "
-                "To use the new notification behavior, use the remote pipeline runner."
+                "Some of the features (like setting message or severity level)."
```
quaark: Some of the features what? 😄 
Guess you mean "Some of the features are not supported" ?

quaark (changes requested):

docs/concepts/notifications.md
```diff
@@ -161,6 +161,8 @@ slack_notification = mlrun.model.Notification(
     secret_params={"secret": "SLACK_SECRET1"},
 )
 ```
+> ℹ️ **Info:** On running notification you can set only the webhook, and the message will be the default message.
```
jillnogold: ### "running" notifications 
"running" notifications have additional considerations:
- You can set only the webhook; the message is the default message.
- If the webhook is stored in the `secret_params`, you should first set the project secret and then use this project secret in the notification. For example:
...

jillnogold (approved):

quaark (approved):

PR #6535, opened 2024-10-14
[Artifacts] Limit the artifact body size when logging artifacts

moranbental:
To avoid handling large amounts of data in memory, we limit the body size to 16 MB. 
If the body exceeds this limit, an error will be raised, directing users to log larger artifacts through files using the `src_path` parameter.

This change has already been approved in PR [#6349](https://github.com/mlrun/mlrun/pull/6349). 
This PR specifically addresses the [error](https://github.com/mlrun/mlrun/pull/6364): `The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all()` encountered when logging artifacts where the body is a DataFrame. In this case, we now check the body size differently

https://iguazio.atlassian.net/browse/ML-7240


mlrun/utils/helpers.py
```diff
@@ -277,15 +278,36 @@ def validate_artifact_key_name(
...
+    :param is_inline: A flag indicating whether the artifact body is inline.
+
+    :raises mlrun.errors.MLRunBadRequestError: If the body exceeds the maximum allowed size.
+    """
+    body_size = 0
+
+    if isinstance(body, pd.DataFrame):
+        body_size = body.memory_usage(deep=True).sum()
+    elif isinstance(body, (str, bytes)):
+        body_size = len(body)
```
rokatyy, line 297: actual size of string in bytes really depends on encoding so just make sure it's utf-8
```suggestion
       body_size = len(body.encode('utf-8')) if isinstance(body, str) else len(body)
```

PR #6541, opened 2024-10-15
[Requirements] Bump storey to 1.8.0, add test

gtopper:
[ML-7818](https://iguazio.atlassian.net/browse/ML-7818)

[ML-7818]: https://iguazio.atlassian.net/browse/ML-7818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/feature-store/test_steps.py
+++ b/tests/feature-store/test_steps.py
@@ -19,6 +19,7 @@ import unittest.mock
 import numpy as np
 import pandas as pd
 import pytest
+import storey
 
 import mlrun
 import mlrun.feature_store as fstore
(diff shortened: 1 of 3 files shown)


tests/feature-store/test_steps.py
```diff
@@ -733,6 +734,59 @@ def test_imputer_default_value(rundb_mock, engine):
     assert not imputed_df.isnull().values.any()
 
 
+class MyChoice(storey.Choice):
+    def select_outlets(self, event):
```
Eyal-Danieli, line 738: just a suggestion
```suggestion 
    @staticmethod
    def select_outlets(event):
```

Eyal-Danieli (approved):
LGRM
just a single suggestion

tests/feature-store/test_steps.py
gtopper: `select_outlets()` is an instance method in `storey.Choice`. It's not good practice to override an instance method with a static method. It also causes static analysis warnings due to the signature not matching.

guy1992l (commented):
Looks good to me 👍

PR #6554, opened 2024-10-16
[Data migration] Bump latest data migration version

rokatyy:
We have 8 version of migration, but latest_data_version wasn't bumped

PR #6555, opened 2024-10-16
[Schedule] Synchronize labels on upgrade [1.7.x]

rokatyy:
backport #6389 
https://iguazio.atlassian.net/browse/ML-7914

### DIFF
--- a/server/api/initial_data.py
+++ b/server/api/initial_data.py
@@ -35,6 +35,7 @@ import server.api.db.sqldb.models
 import server.api.utils.db.alembic
 import server.api.utils.db.backup
 import server.api.utils.db.mysql
+import server.api.utils.scheduler
 from mlrun.artifacts.base import fill_artifact_object_hash
 from mlrun.config import config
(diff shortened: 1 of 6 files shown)


TomerShor (approved):

server/api/initial_data.py
```diff
@@ -245,6 +246,8 @@ def _perform_data_migrations(db_session: sqlalchemy.orm.Session):
                 _perform_version_6_data_migrations(db, db_session)
             if current_data_version < 7:
                 _perform_version_7_data_migrations(db, db_session)
+            if current_data_version < 8:
+                _perform_version_8_data_migrations(db, db_session)
```
liranbg, line 250: aren't you supposed to bump the latest data version then?

liranbg (changes requested):

server/api/initial_data.py
rokatyy: @liranbg already did it and it's merged #6554

TomerShor: And in #6556 for 1.7.x, so maybe not aligned with upstream

PR #6568, opened 2024-10-21
[Automation] Add script for managing old pipeline runs

yaelgen:
This PR introduces a new script, clean_pipelines.py, designed to delete old pipeline runs associated with a specific MLRun project, with date filtering.
This can be used as an MLRun function/schedule  to automate pipeline cleanup tasks.

Additionally, it includes a README file that provides usage instructions, function descriptions, and parameter details for integrating the script into MLRun projects.

ticket: https://iguazio.atlassian.net/browse/ML-8077


automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,392 @@
...
+        dt_object = parser.parse(date_input)
+        if dt_object.tzinfo is not None:
+            # Convert to UTC if it's in a different timezone
+            dt_object = dt_object.astimezone(pytz.utc)
+        else:
+            # If no timezone info is present, assume it's in local time
+            local_tz = pytz.timezone("UTC")
+            dt_object = local_tz.localize(dt_object)
+
+        formatted_date = dt_object.isoformat().replace("+00:00", "Z")
```
roei3000b, line 99: Maybe add before that section explanation that you do it for enforcing RFC 3339 format.

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,392 @@
...
+            page_token=page_token,
+            page_size=mlrun.common.schemas.PipelinesPagination.max_page_size,
+            sort_by=sort_by if page_token == "" else "",
+            filter=query_filter if page_token == "" else "",
+        )
+        runs.extend([PipelineRun(run) for run in response.runs or []])
+        page_token = response.next_page_token
+        if counter % 50 == 0:
+            mlrun.utils.logger.info(
+                "Collecting pipelines runs:", runs_count=(10000 * counter)
```
roei3000b: From where the `10000` comes?
Shouldn't it be `mlrun.common.schemas.PipelinesPagination.max_page_size * counter` ?

roei3000b (changes requested):
Looks good, minor comments.
In general, if the purpose if this script is only cleanup pipelines, I would suggest that instead of store all runs in memory(`list_pipelines_runs`) I think should use the pagination mechanism for deleting and not store all the runs on memory.

I would also suggest that the dry run will pass down to the delete function (because it is where the actual deletion happened and make this function decide what it should be if dry run is set to True).

automation/scripts/clean_pipelines/README.md
```diff
@@ -0,0 +1,90 @@
+# Clean Pipelines Script 
+
+This script is designed to delete old pipeline runs associated with a specific MLRun project,
+with flexible date filtering.
+You can use this as an MLRun function to automate pipeline cleanup tasks.
```
alonmr, line 5: Add some warning that this is destructive and should be used with extra care

automation/scripts/clean_pipelines/README.md
```diff
@@ -0,0 +1,90 @@
...
+    handler="delete_project_old_pipelines",
+)
+```
+
+The `delete_project_old_pipelines` function deletes old pipeline runs associated with a specified project.
+
+#### Function parameters:
+- **context**: The context object to log results (automatically passed by MLRun).
+- **project_name** (str): The name of the project for which to delete old pipelines.
+- **end_date** (str): The cutoff date for deleting pipeline runs. All runs created on or before this date will be considered for deletion (format: YYYY-MM-DDTHH:MM:SSZ).
```
alonmr: Can `(format: YYYY-MM-DDTHH:MM:SSZ)` be translated to something readable? is it isoformat?

automation/scripts/clean_pipelines/README.md
```diff
@@ -0,0 +1,90 @@
...
+- **start_date** (str, optional): If provided, only runs created on or after this date will be considered for deletion (default is empty, no filtering).
+- **target_path** (str, optional): The path where logs or output artifacts from the function run will be saved,
```
alonmr: I don't think this should be a function param. It is already a parameter in `func.run()`

automation/scripts/clean_pipelines/clean_pipelines.py
alonmr, line 1: See (10) in https://iguazio.atlassian.net/wiki/spaces/TLE/pages/2457650/MLRun+Python+-+Coding+Conventions
> When using a function in the same file, it should be declared below the calling function as it makes it easier to understand the functionality.

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,392 @@
...
+            dt_object = local_tz.localize(dt_object)
+
+        formatted_date = dt_object.isoformat().replace("+00:00", "Z")
+        if not formatted_date.endswith("Z"):
+            formatted_date += "Z"
+
+        return formatted_date
+    except (ValueError, OverflowError) as e:
+        raise ValueError(
+            f"Invalid date format: {date_input}. Please provide a valid date."
```
alonmr:
```suggestion
            f"Invalid date format: {date_input}. Date format must adhere ...."
```

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,392 @@
...
+            # Convert to UTC if it's in a different timezone
+            dt_object = dt_object.astimezone(pytz.utc)
+        else:
+            # If no timezone info is present, assume it's in local time
+            local_tz = pytz.timezone("UTC")
+            dt_object = local_tz.localize(dt_object)
+
+        formatted_date = dt_object.isoformat().replace("+00:00", "Z")
+        if not formatted_date.endswith("Z"):
+            formatted_date += "Z"
```
alonmr:
```suggestion
        formatted_date = formatted_date.rstrip("Z") + "Z"
```

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,392 @@
...
+        raise ValueError(
+            f"Invalid date format: {date_input}. Please provide a valid date."
+        ) from e
+
+
+def get_kfp_client(
+    kfp_url=mlrun.mlconf.kfp_url, namespace: str = "default-tenant"
+) -> Client:
+    kfp_client = mlrun_pipelines.utils.get_client(kfp_url, namespace)
+    return kfp_client
```
alonmr, line 114: you can possibly use `mlrun.mlconf.resolve_namespace()` and don't use `default-tenant` because it's not a thing in CE

automation/scripts/clean_pipelines/clean_pipelines.py
alonmr: Agree we need some explanation of these magic numbers (% 50 ?)

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,392 @@
...
+
+
+def get_experiment_name(kfp_client: Client, experiment_id: str) -> str:
+    experiment = kfp_client.get_experiment(experiment_id=experiment_id)
+    return experiment.name if experiment else ""
+
+
+def filter_out_non_related_runs(
```
alonmr:
```suggestion
def filter_project_runs(
```

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,392 @@
...
+
+    :param kfp_client: KFP client for interacting with the pipeline API.
+    :param project_name: Name of the project for filtering the runs.
+    :param query_filter: Filter for querying the runs.
+    """
+    runs = list_pipelines_runs(kfp_client, query_filter)
+
+    # Filter out non project-related runs if project was provided
+    project_runs = (
+        filter_out_non_related_runs(project_name, runs) if project_name != "*" else runs
```
alonmr: I would also move the `if project != *` into `filter_out_non_related_runs`

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,392 @@
...
+    if not dry_run:
+        # Delete runs
+        delete_runs(context, kfp_client, runs, target_path)
+
+        # Find and delete empty experiments
+        delete_empty_experiments(context, kfp_client, experiments_ids, target_path)
+
+    else:
+        mlrun.utils.logger.info(f"Dry run: {len(runs)} runs would be deleted")
+        context.log_result(key="runs_to_be_deleted", value=runs)
```
alonmr: Will be nice to add the latest run end time

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,392 @@
...
+    for item_id, name in items:
+        try:
+            delete_func(item_id)
+            deleted_count += 1
+            deleted_items.append((name, item_id))
+            if deleted_count % 100 == 0:
+                mlrun.utils.logger.info(
+                    f"Deleted {deleted_count}/{total_items_amount} {item_type}s successfully"
+                )
+                sleep(10)
```
alonmr: why sleep?

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,392 @@
...
+    """
+    delete_items(
+        context,
+        runs,
+        lambda run_id: kfp_client._run_api.delete_run(run_id),
+        target_path=target_path,
+    )
+
+
+def delete_empty_experiments(
```
alonmr: In mlrun we don't care if the experiment is empty or not so why here?
This seems very heavy (doing a list for every experiment id) can you explain why we need it?
IIUC experiment and runs are 1 to 1 ? If I'm wrong then maybe kfp will block us from deleting a non-empty experiment and then we don't need to check?

alonmr (changes requested):
Very nice an thorough solution. A few comments below and please add jira ticket to description

automation/scripts/clean_pipelines/clean_pipelines.py
yaelgen: Actually, experiments can contain multiple runs, each time you execute the same pipeline under the same project, kfp creates a new run under the same experiment. This means an experiment can accumulate runs over time, and a project can have multiple experiments.

When we delete resources in MLRun for project deletion, we remove all assets associated with a project, including experiments created based on the project and workflow names. However, if we're deleting runs prior to a specific date, some runs within an experiment may remain. Deleting the entire experiment would remove runs that don't meet the deletion criteria. Therefore, we only need to delete an experiment if it becomes empty after all its associated runs have been deleted.

automation/scripts/clean_pipelines/clean_pipelines.py
yaelgen: to avoid continuously loading kfp and sending excessive deletion requests when removing multiple runs over several hours. I do think 10 seconds is too long, we used it in Vistra cause it was the first time we ran it and wasn't sure how loading it's going to be, but I believe we can remove it or lower it, wdyt?

automation/scripts/clean_pipelines/README.md
```diff
@@ -0,0 +1,92 @@
...
+### Define the `end_date` and any other necessary parameters, for example:
+
+```python
+end_date = "2024-10-20T10:10:06Z"
+target_path = "/path/to/logs"
+
+params = {
+    "project_name": "remote-workflow-example1",
+    "end_date": end_date,
+    "artifact_path": artifact_path,
```
moranbental: It looks like you meant to refer to `target_path` not to `artifact_path`
```suggestion
    "target_path": target_path,
```

automation/scripts/clean_pipelines/README.md
```diff
@@ -0,0 +1,92 @@
...
+    func=code_path,
+    kind="job",
+    image="mlrun/mlrun",
+    handler="delete_project_old_pipelines",
+)
+```
+
+The `delete_project_old_pipelines` function deletes old pipeline runs associated with a specified project.
+
+#### Function parameters:
```
moranbental: Move all the parameter descriptions and the param object before step 4, as they are used there, which can be confusing

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,397 @@
...
+import pandas as pd
+import pytz
+from dateutil import parser
+from kfp import Client
+from mlrun_pipelines.models import PipelineRun
+
+import mlrun
+
+
+def delete_project_old_pipelines(
```
moranbental, line 32: All functions in this file can be private except for `delete_project_old_pipelines`

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,397 @@
...
+    filters = {
+        "predicates": [
+            {
+                "key": "created_at",
+                "op": 7,  # Operation 7 corresponds to '<=' (less than or equal)
+                "timestamp_value": end_date,
+            },
+        ]
+    }
+    if not project_name == "*":
```
moranbental:
```suggestion
    if project_name != "*":
```

moranbental (changes requested):
Looks very good! Please add some unit tests to cover cases with none/empty fields

yaelgen:
> Looks good, minor comments. In general, if the purpose if this script is only cleanup pipelines, I would suggest that instead of store all runs in memory(`list_pipelines_runs`) I think should use the pagination mechanism for deleting and not store all the runs on memory.
> 
> I would also suggest that the dry run will pass down to the delete function (because it is where the actual deletion happened and make this function decide what it should be if dry run is set to True).

Will be updated to use pagination once the paginated run query will be merged: 
https://iguazio.atlassian.net/browse/ML-8242

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,431 @@
...
+    deleted_count = 0
+    failed_items = []
+    deleted_items = []
+
+    for item_id, name in items:
+        try:
+            delete_func(item_id)
+            deleted_count += 1
+            deleted_items.append((name, item_id))
+            if deleted_count % 100 == 0:
```
roei3000b: Maybe also make `100` as variable

roei3000b (commented):
Looks very good, last minor suggestion.

automation/scripts/clean_pipelines/README.md
```diff
@@ -0,0 +1,97 @@
...
+Use MLRun version ***1.7.0*** or greater to build the function as follows
+(building is only necessary if your MLRun version is lower than 1.7.0):
+
+```python
+function_build = project.build_function(
+    func,
+    base_image="python:3.9",
+    with_mlrun=False,
+    force_build=True,
+    requirements=["mlrun==1.7.0-rc55"],
```
moranbental: maybe use here 1.7.0 and not rc

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,431 @@
...
+    experiment_ids = set(run.experiment_id for run in project_runs if run.experiment_id)
+
+    return runs, experiment_ids
+
+
+def _list_pipelines_runs(
+    kfp_client: Client, query_filter: str, page_token: str = "", sort_by: str = ""
+) -> list[PipelineRun]:
+    runs = []
+    batch_size = 1000  # Adjust batch size for progress updates
```
moranbental: Change `batch_size` to be a function parameter with a default value of 1000

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,431 @@
...
+    while page_token is not None:
```
moranbental: Can `page_token` be an empty string? You should stop when this happens

moranbental (changes requested):

automation/scripts/clean_pipelines/README.md
yaelgen: yep the PR opened before 1.7 was released 👍

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,437 @@
...
+
+    return runs, experiment_ids
+
+
+def _list_pipelines_runs(
+    kfp_client: Client,
+    query_filter: str,
+    page_token: str = "",
+    sort_by: str = "",
+    batch_size=1000,
```
moranbental:
```suggestion
    batch_size: int = 1000,
```

moranbental (approved):
LGTM

roei3000b (approved):
Great! :)

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,422 @@
...
+# Interval for logging deletion progress
+DELETION_LOG_INTERVAL = 100
+
+
+def delete_project_old_pipelines(
+    context: mlrun.MLClientCtx,
+    project_name: str,
+    end_date: str,
+    start_date: str = "",
+    dry_run: bool = True,
```
alonmr: readme says the default is false
```suggestion
    dry_run: bool = False,
```

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,422 @@
...
+        try:
+            delete_func(item_id)
+            deleted_count += 1
+            deleted_items.append((name, item_id))
+            if deleted_count % DELETION_LOG_INTERVAL == 0:
+                mlrun.utils.logger.info(
+                    f"Deleted {deleted_count}/{total_items_amount} {item_type}s successfully"
+                )
+                # A 5-second sleep is used to balance KFP load and limit the rate of deletion requests.
+                sleep(5)
```
alonmr, line 381: I'm assuming it's possible that one of the empty experiments will get a new run while deleting all of the experiments right? Is there something we can do about it? verify no matching schedules?

automation/scripts/clean_pipelines/clean_pipelines.py
```diff
@@ -0,0 +1,422 @@
...
+
+
+def _list_pipelines_runs(
+    kfp_client: Client,
+    query_filter: str,
+    page_token: str = "",
+    sort_by: str = "",
+    batch_size: int = 1000,
+) -> list[PipelineRun]:
+    runs = []
```
alonmr, line 190: I fear this can explode in memory. Maybe we should add some default limit to number of runs (10k?). If someone passes a long time range and we need to pull 500k runs this can kill the pod

alonmr (commented):
Nicely done! A couple of small concerns

automation/scripts/clean_pipelines/clean_pipelines.py
alonmr: 5 seconds sounds good to me

automation/scripts/clean_pipelines/clean_pipelines.py
yaelgen: If an experiment is deleted and a new run is created afterward, the experiment will automatically be recreated. However, this script is made to be used to delete older runs, meaning that if an experiment becomes empty as a result, there have likely been no new runs in the past few weeks (based on the filtering criteria). Therefore, the likelihood of a new run being added immediately after deletion is quite low, making this scenario a real edge case.

That said, I can look into adding a verification step to check for matching schedules to further minimize this possibility, though I believe the chances of it happening is quite low.

automation/scripts/clean_pipelines/clean_pipelines.py
yaelgen: This will be paginated in a follow-up PR, where the deletion will happen after each page, so it shouldn't be a concern.

automation/scripts/clean_pipelines/clean_pipelines.py
alonmr: That's ok, worst case it will fail and continue

alonmr (approved):

liranbg (approved):

PR #6569, opened 2024-10-22
[Model Monitoring] Edit TDEngine tables & Connector

royischoss:
https://iguazio.atlassian.net/browse/ML-7456
Added errors table to TDEngine super tables, added functionality to TDEngineConnector, including error extractor step and metadata queries.
System test test_app_flow from tests/system/model_monitoring/test_app.py using both TDEngine as tsdb and v3io


mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -446,3 +446,26 @@ def df_to_results_values(
             )
 
         return metrics_values
+
+    @staticmethod
+    def _get_start_end(
+        start: Union[str, datetime],
+        end: Union[str, datetime],
```
Eyal-Danieli: just to keep consistency with the syntax of other functions in this class. I’ve noticed that this issue also appears in other parts of the class, so please update those as well. Afterward, you can remove the `from typing import Union` statement.
```suggestion
        start: typing.Union[str, datetime],
        end: typing.Union[str, datetime],
```

mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
```diff
@@ -277,3 +310,18 @@ def __init__(self, database: Optional[str] = None):
...
         super().__init__(super_table, columns, tags, database)
+
+
+@dataclass
+class Errors(TDEngineSchema):
+    def __init__(self, database: Optional[str] = None):
+        super_table = mm_schemas.TDEngineSuperTables.ERRORS
+        columns = {
+            mm_schemas.EventFieldType.TIME: _TDEngineColumn.TIMESTAMP,
+            mm_schemas.EventFieldType.MODEL_ERROR: _TDEngineColumn.BINARY_10000,
```
Eyal-Danieli: What happens if the error description exceeds 10k? In that case, the data won't be stored, and you'll get an error from tdengine. Additionally, if there are a lot of errors, storing 10k for each one could lead to storage issues down the line.

To address this, I suggest the following:

- Reduce the limit to 1k characters.
- Ensure that the error description does not exceed 1k using the new `ErrorExtractor` step you added. If it does exceed this limit, we can store the last 1k and make sure to explain the user that the error exceeds the limit, so only the last 1k chars will be stored.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -245,6 +277,10 @@ def _get_records(
         limit: typing.Optional[int] = None,
         sliding_window_step: typing.Optional[str] = None,
         timestamp_column: str = mm_schemas.EventFieldType.TIME,
+        group_by: typing.Optional[Union[list[str], str]] = None,
+        preform_agg_columns: typing.Optional[list] = None,
+        order_by: typing.Optional[str] = None,
+        desc: typing.Optional[bool] = None,
```
Eyal-Danieli, line 283: update the func documentation with the new params

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -438,50 +478,216 @@ def read_predictions(
...
-        start: Union[datetime, str] = "now-24h",
-        end: Union[datetime, str] = "now",
+        start: datetime = None,
+        end: datetime = None,
     ) -> pd.DataFrame:
-        pass
+        endpoint_ids = (
+            endpoint_ids if isinstance(endpoint_ids, list) else [endpoint_ids]
+        )
+        start = start or "now"
```
Eyal-Danieli: why not doing this calculation at this point? instead of generating a dummy string of "now" which will be converted anyway in the `self._get_start_end()` func to mlrun.utils.datetime_now()
```suggestion
        start = start or mlrun.utils.datetime_now()
```

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -734,12 +734,14 @@ def get_last_request(
...
         endpoint_ids: Union[str, list[str]],
-        start: Union[datetime, str] = "now-24h",
-        end: Union[datetime, str] = "now",
+        start: datetime = None,
+        end: datetime = None,
     ) -> pd.DataFrame:
         endpoint_ids = (
             endpoint_ids if isinstance(endpoint_ids, list) else [endpoint_ids]
         )
+        start = start or "now"
```
Eyal-Danieli: same comment as above

tests/system/model_monitoring/test_app.py
```diff
@@ -171,18 +171,8 @@ def _test_metrics_kv_record(cls, ep_id: str) -> None:
...
-        else:
-            # TDEngine
-            df: pd.DataFrame = cls._tsdb_storage._get_records(
-                table=mm_constants.TDEngineSuperTables.APP_RESULTS,
-                start=datetime.now().astimezone()
-                - timedelta(minutes=10 * cls.app_interval),
-                end=datetime.now().astimezone(),
-                timestamp_column=mm_constants.WriterEvent.END_INFER_TIME,
-            )
+        # V3IO TSDB
```
Eyal-Danieli:
```suggestion

```

Eyal-Danieli (commented):
Nice! 
I have a few comments, but overall, it looks great!

tests/model_monitoring/test_tdengine.py
```diff
@@ -186,35 +241,116 @@ def test_get_records_query(
         start: str,
         end: str,
```
Eyal-Danieli:
```suggestion
        start: datetime.datetime,,
        end: datetime.datetime,,
```
similar change in L397 and L398

Eyal-Danieli (approved):
NICE!

royischoss:
System test test_app_flow from tests/system/model_monitoring/test_app.py using both TDEngine as tsdb and v3io

assaf758 (approved):

PR #6586, opened 2024-10-27
[Alerts] Refine alert configuration to skip reset on non-functional changes

Yacouby:
Resolves:
https://iguazio.atlassian.net/browse/ML-7749

This PR addresses the behavior where existing alert configurations were being reset on every modification, regardless of the type of change. 
To improve functionality, the reset behavior now only applies to modifications that impact the alert's decision logic.

Key changes:
1. Non-functional changes – Adjustments to fields such as Description, Summary, Severity, Notifications, and Reset-policy no longer trigger an alert reset.

2. Functional changes – Changes to fields tied to alert conditions: Entity, Trigger, Criteria will reset the alert.
3. Force Reset Option – Added a force_reset parameter (default: False) to allow optional reset even when modifying non-functional fields. This parameter is accessible via the SDK and REST API.

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -4327,6 +4327,7 @@ class HTTPRunDB(RunDBInterface):
         alert_name: str,
         alert_data: Union[dict, AlertConfig],
         project="",
+        force_reset: bool = False,
     ) -> AlertConfig:
         """
         Create/modify an alert.
@@ -4334,6 +4335,7 @@ class HTTPRunDB(RunDBInterface):
         :param alert_name: The name of the alert.
         :param alert_data: The data of the alert.
         :param project:    The project that the alert belongs to.
+        :param force_reset: If true and the alert already exist then we should force reset the alert.
         :returns:          The created/modified alert.
         """
         if not alert_data:
@@ -4358,7 +4360,10 @@ class HTTPRunDB(RunDBInterface):
 
         alert_data = alert_instance.to_dict()
         body = _as_json(alert_data)
-        response = self.api_call("PUT", endpoint_path, error_message, body=body)
+        params = {"force_reset": bool2str(force_reset)}
+        response = self.api_call(
+            "PUT", endpoint_path, error_message, params=params, body=body
+        )
         return AlertConfig.from_dict(response.json())
 
     def get_alert_config(self, alert_name: str, project="") -> AlertConfig:
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -4335,7 +4335,7 @@ class HTTPRunDB(RunDBInterface):
         :param alert_name: The name of the alert.
         :param alert_data: The data of the alert.
         :param project:    The project that the alert belongs to.
-        :param force_reset: If true and the alert already exist then we should force reset the alert.
+        :param force_reset: If True and the alert already exists, the alert would be reset.
         :returns:          The created/modified alert.
         """
         if not alert_data:
@@ -4360,7 +4360,7 @@ class HTTPRunDB(RunDBInterface):
 
         alert_data = alert_instance.to_dict()
         body = _as_json(alert_data)
-        params = {"force_reset": bool2str(force_reset)}
+        params = {"force_reset": bool2str(force_reset)} if force_reset else {}
         response = self.api_call(
             "PUT", endpoint_path, error_message, params=params, body=body
         )
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -4213,7 +4213,7 @@ class MlrunProject(ModelObj):
 
         :param alert_data: The data of the alert.
         :param alert_name: The name of the alert.
-        :param force_reset: If true and the alert already exist then we should force reset the alert.
+        :param force_reset: If True and the alert already exists, the alert would be reset.
         :return: the created/modified alert.
         """
         if not alert_data:
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -4203,13 +4203,17 @@ class MlrunProject(ModelObj):
         mlrun.db.get_run_db().delete_api_gateway(name=name, project=self.name)
 
     def store_alert_config(
-        self, alert_data: AlertConfig, alert_name: typing.Optional[str] = None
+        self,
+        alert_data: AlertConfig,
+        alert_name: typing.Optional[str] = None,
+        force_reset: bool = False,
     ) -> AlertConfig:
         """
         Create/modify an alert.
 
         :param alert_data: The data of the alert.
         :param alert_name: The name of the alert.
+        :param force_reset: If true and the alert already exist then we should force reset the alert.
         :return: the created/modified alert.
         """
         if not alert_data:
@@ -4223,7 +4227,9 @@ class MlrunProject(ModelObj):
                 project=alert_data.project,
             )
         alert_data.project = self.metadata.name
-        return db.store_alert_config(alert_name, alert_data, project=self.metadata.name)
+        return db.store_alert_config(
+            alert_name, alert_data, project=self.metadata.name, force_reset=force_reset
+        )
 
     def get_alert_config(self, alert_name: str) -> AlertConfig:
         """
--- a/server/api/api/endpoints/alerts.py
+++ b/server/api/api/endpoints/alerts.py
@@ -36,6 +36,7 @@ async def store_alert(
     project: str,
     name: str,
     alert_data: mlrun.common.schemas.AlertConfig,
+    force_reset: bool = False,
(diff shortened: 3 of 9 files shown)


mlrun/db/httpdb.py
```diff
@@ -4327,13 +4327,15 @@ def store_alert_config(
...
         project="",
+        force_reset: bool = False,
     ) -> AlertConfig:
         """
         Create/modify an alert.
 
         :param alert_name: The name of the alert.
         :param alert_data: The data of the alert.
         :param project:    The project that the alert belongs to.
+        :param force_reset: If true and the alert already exist then we should force reset the alert.
```
```suggestion
        :param force_reset: If True and the alert already exists, the alert would be reset.
```

mlrun/db/httpdb.py
```diff
@@ -4358,7 +4360,10 @@ def store_alert_config(
 
         alert_data = alert_instance.to_dict()
         body = _as_json(alert_data)
-        response = self.api_call("PUT", endpoint_path, error_message, body=body)
+        params = {"force_reset": bool2str(force_reset)}
```
TomerShor: If false it can be ignored
```suggestion
        params = {"force_reset": bool2str(force_reset)} if force_reset else {}
```

mlrun/projects/project.py
```diff
@@ -4203,13 +4203,17 @@ def delete_api_gateway(
...
+        alert_data: AlertConfig,
+        alert_name: typing.Optional[str] = None,
+        force_reset: bool = False,
     ) -> AlertConfig:
         """
         Create/modify an alert.
 
         :param alert_data: The data of the alert.
         :param alert_name: The name of the alert.
+        :param force_reset: If true and the alert already exist then we should force reset the alert.
```
```suggestion
        :param force_reset: If True and the alert already exists, the alert would be reset.
```

server/api/crud/alerts.py
```diff
@@ -57,6 +58,13 @@ def store_alert(
         if alert is not None:
             self._delete_notifications(alert)
             self._get_alert_by_id_cached().cache_remove(session, alert.id)
+
+            for kind in alert.trigger.events:
+                server.api.crud.Events().remove_event_configuration(
+                    project, kind, alert.id
+                )
+            alert_data.created = alert.created
+            alert_data.id = alert.id
```
TomerShor: Why is this needed?
Add a comment explaining the Why

server/api/crud/alerts.py
```diff
@@ -57,6 +58,13 @@ def store_alert(
         if alert is not None:
```
TomerShor: I would rename `alert` to `existing_alert` to make it easier to differentiate between the new alert and the "old" alert.

server/api/crud/alerts.py
```diff
@@ -87,7 +87,17 @@ def store_alert(
                 project, kind, new_alert.id
             )
 
-        self.reset_alert(session, project, new_alert.name)
+        # if the alert already exist we should check if it should be reset or not
```
TomerShor:
```suggestion
        # if the alert already exists we should check if it should be reset or not
```

server/api/crud/alerts.py
```diff
@@ -375,6 +385,20 @@ def reset_alert(self, session: sqlalchemy.orm.Session, project: str, name: str):
         self._get_alert_state_cached().cache_remove(session, alert.id)
         self._clear_alert_states(alert)
 
+    @staticmethod
+    def _should_reset_alert(old_alert_data, alert_data, force_reset):
+        if force_reset:
+            return True
+
+        # if one of these fields was modified then we should reset the alert
```
TomerShor: You are commenting what you are doing and not why.
```suggestion
        # the alert should be reset if a functional parameter was changed, which could affect the decision on when to trigger the alert
```

server/api/crud/alerts.py
```diff
@@ -375,6 +385,20 @@ def reset_alert(self, session: sqlalchemy.orm.Session, project: str, name: str):
...
+        if (
+            old_alert_data.entities != alert_data.entities
+            or old_alert_data.trigger != alert_data.trigger
+            or old_alert_data.criteria != alert_data.criteria
+        ):
+            return True
+        return False
+
```
TomerShor: With some python magic 🪄 :
```suggestion
        return any(getattr(old_obj, attr) != getattr(new_obj, attr) for attr in [
            "entities",
            "trigger",
            "criteria",
        ])

```

TomerShor: Also - can the `EventEntities` instances be compared?
Don't they need the `__eq__` method?

tests/api/crud/test_alerts.py
```diff
@@ -104,33 +94,198 @@ async def test_validate_alert_name(
...
+                )
+            ],
+            False,
+        ),
+        ("reset_policy", alert_objects.ResetPolicy.AUTO, False),
+        # Functional fields:
+        (
+            "entities",
+            alert_objects.EventEntities(
+                kind=alert_objects.EventEntityKind.JOB,
```
TomerShor, line 156: add a case where the kind changes

tests/api/crud/test_alerts.py
```diff
@@ -104,33 +94,198 @@ async def test_validate_alert_name(
...
+    # validate the state based on whether it should have reset
+    expected_state = (
+        alert_objects.AlertActiveState.INACTIVE
+        if should_reset or force_reset
+        else alert_objects.AlertActiveState.ACTIVE
+    )
+    assert alert.state == expected_state
+
+    # reset the alert so the next test starts clean
+    server.api.crud.Alerts().reset_alert(session=db, project=project, name=alert_name)
```
TomerShor: If one of the `assert`s above fails we would never get here.
You can maybe use a fixture to make sure it is always called.

tests/api/crud/test_alerts.py
```diff
@@ -104,33 +94,198 @@ async def test_validate_alert_name(
...
+            ),
+            True,
+        ),
+    ],
+)
+@pytest.mark.parametrize(
+    "force_reset",
+    [False, True],
+)
+async def test_alert_reset_with_fields_updates(
```
TomerShor, line 234: I would also test that changing multiple fields simultaneously works as expected

TomerShor (changes requested):
Looking great!
Added some suggestions for improvement

server/api/crud/alerts.py
Yacouby: Since EventEntities is a pydantic.BaseModel, it automatically includes an __eq__ method that compares all fields by default.
This means two EventEntities instances will be considered equal if all their attributes match

Yacouby: > With some python magic 🪄 :
Nice!

server/api/api/endpoints/alerts.py
```diff
@@ -71,6 +72,7 @@ async def store_alert(
         project,
         name,
         alert_data,
+        force_reset,
```
alonmr, line 75: I don't understand something. In httpdb you used `bool2str` which converts this to "yes" or "no" so how is this a bool here? Do you need to use `strtobool` like `with_mlrun = strtobool(data.get("with_mlrun", "on"))`
I see the same method is used in other places like `list_runs` I wonder if this is a bug

tests/api/crud/test_alerts.py
```diff
@@ -21,66 +21,62 @@
 import sqlalchemy.orm
 
 import mlrun.common.schemas.alert
+import mlrun.common.schemas.alert as alert_objects
 import server.api.crud
 import tests.api.conftest
 
 
+@pytest.fixture
+def reset_alert(db, project="project-name", alert_name="failed-alert"):
```
alonmr: do you need this fixture? I believe the DB should start from scratch on each test

alonmr (commented):
Generally looks good to me. A couple of comments/questions

server/api/api/endpoints/alerts.py
Yacouby: I checked that out and seems like the FastApi converts the "yes" or "no", to True or False

tests/api/crud/test_alerts.py
Yacouby: it does not start from scratch.. without it only the first test pass and the others fail since the state of the alert is modified in the first test

TomerShor (approved):
Nice job!

tests/api/crud/test_alerts.py
alonmr: I think this fixture should also create the alert. You have duplicated the project and alert_name params and this fixture doesn't have any meaning if the alert was not created.
I'm still not sure why the DB doesn't start from scratch between tests since the fixture uses a tmp file as the DB file so that makes me think there might be some other issue here

alonmr (approved):
Nicely done!

PR #6587, opened 2024-10-27
[Logger] Allow users to use custom logger format

GiladShapira94:
https://iguazio.atlassian.net/browse/ML-7770
This PR allow users to use their own custom logger format, for using custom logger the user should create two env's see below - 
    1. `"MLRUN_LOG_FORMATTER" = "custom"` - change the default log formatter.
    2. `"MLRUN_CUSTOM_FORMAT" = "> {timestamp} [{level}] Running module: {module} {message} {more}"` - the format for the logger


mlrun/config.py
```diff
@@ -101,6 +101,8 @@
     "log_level": "INFO",
     # log formatter (options: human | human_extended | json)
     "log_formatter": "human",
+    # custom logger format, workes only with log_formatter: custom
```
alonmr, line 104: add which keys need to be in the format

mlrun/utils/logger.py
```diff
@@ -93,6 +95,68 @@ def _record_with(self, record):
         return record_with
 
 
+class CUSTOMFormatter(HumanReadableFormatter):
```
alonmr:
```suggestion
class CustomFormatter(HumanReadableFormatter):
```

mlrun/utils/logger.py
```diff
@@ -93,6 +95,68 @@ def _record_with(self, record):
...
+    """
+    For using cusom logger you need to add the below env's :
```
alonmr:
```suggestion
    To enable custom logger formatter, configure MLRun with the following env variables:
```

mlrun/utils/logger.py
```diff
@@ -93,6 +95,68 @@ def _record_with(self, record):
...
+    1. "MLRUN_LOG_FORMATTER" = "custom" - change the default log formatter.
+    2. "MLRUN_CUSTOM_FORMAT" = "> {timestamp} [{level}] Running module: {module} {message} {more}" - logger format
+        * Please note that your custom format must include those 4 fields - timestamp, level, message and more
```
alonmr:
```suggestion
        * Note that your custom format must include those 4 fields - timestamp, level, message and more
```

mlrun/utils/logger.py
```diff
@@ -93,6 +95,68 @@ def _record_with(self, record):
...
+    If the custom format does not configure properly , MLRun will use the default logger (human format).
```
alonmr:
```suggestion
    If the custom format is not configured properly , MLRun will use the default logger (human format).
```

mlrun/utils/logger.py
```diff
@@ -93,6 +95,68 @@ def _record_with(self, record):
...
+                custom_format_keys = [
+                    key
+                    for _, key, _, _ in formatter.parse(custom_format)
+                    if key is not None
+                ]
+                fail_on_missing_default_flags = [
+                    default_key
+                    for default_key in default_keys
+                    if default_key not in custom_format_keys
+                ]
```
alonmr:
```suggestion
                missing_default_flags = list(set(default_keys) - set(custom_format_keys))
```

mlrun/utils/logger.py
```diff
@@ -93,6 +95,68 @@ def _record_with(self, record):
...
+                _custom_format = custom_format.format(
+                    timestamp=self.formatTime(record, self.datefmt),
+                    level=record.levelname.lower(),
+                    message=record.getMessage().rstrip(),
+                    more=more or "",
+                    **record_dict,
+                )
+        except KeyError as e:
+            logger.warning(
+                f"Failed to create custom logger due to missing format key in the log record {e}"
```
alonmr:
```suggestion
                f"Failed to create custom logger due to missing format key in the log record", error=mlrun.errors.err_to_str(e)
```

mlrun/utils/logger.py
```diff
@@ -93,6 +95,68 @@ def _record_with(self, record):
...
+            )
+            _format = (
+                f"> {self.formatTime(record, self.datefmt)} "
+                f"[{record.levelname.lower()}] "
+                f"{record.getMessage().rstrip()}"
+                f"{more}"
+            )
+        except Exception as e:
+            logger.warning(e)
```
alonmr: add a proper warning message and explain what you are doing next

tests/utils/logger/test_logger.py
```diff
@@ -177,3 +178,26 @@ def test_child_logger():
...
+    formatted_time = current_time.strftime("%Y-%m-%d %H:%M:%S,%f")[:-3]
+    expected_logger = f"> {formatted_time} [debug] Running module: logger test custom"
+    mlrun.mlconf.custom_format = format
+    logger = create_logger(
+        "debug",
+        name="test-logger",
+        stream=stream,
+        formatter_kind=FormatterKinds.CUSTOM.name,
+    )
+    logger.debug("test custom")
```
alonmr: Add some kwargs
```suggestion
    logger.debug("test custom", a=1, b="2")
```

mlrun/utils/logger.py
```diff
@@ -272,17 +336,24 @@ class FormatterKinds(Enum):
...
+        HumanReadableExtendedFormatter,
+        JSONFormatter,
+        CUSTOMFormatter,
+    ]
 ]:
     return {
         FormatterKinds.HUMAN: HumanReadableFormatter,
         FormatterKinds.HUMAN_EXTENDED: HumanReadableExtendedFormatter,
         FormatterKinds.JSON: JSONFormatter,
+        FormatterKinds.CUSTOM: CUSTOMFormatter,
```
alonmr:
```suggestion
        FormatterKinds.CUSTOM: CustomFormatter,
```

alonmr (commented):
Nice job! Some small comments

mlrun/config.py
```diff
@@ -101,6 +101,9 @@
...
+    # Note that your custom format must include those 4 fields - timestamp, level, message and more
+    "custom_format": None,
```
liranbg:
```suggestion
    "log_format_override": None,
```

mlrun/utils/logger.py
```diff
@@ -93,6 +96,71 @@ def _record_with(self, record):
...
+    """
+    To enable custom logger formatter, configure MLRun with the following env variables:
+    1. "MLRUN_LOG_FORMATTER" = "custom" - change the default log formatter.
+    2. "MLRUN_CUSTOM_FORMAT" = "> {timestamp} [{level}] Running module: {module} {message} {more}" - logger format
+        * Note that your custom format must include those 4 fields - timestamp, level, message and more
+    If the custom format is not configured properly , MLRun will use the default logger (human format).
+    """
+
+    def format(self, record) -> str:
+        logger = create_logger("INFO", "human", "logger", sys.stdout)
```
liranbg: why basing the logger on human and not on what given by config (human_extendned / json / human)? also you're setting the logger to be info and ignoring the config as well

liranbg (changes requested):

mlrun/utils/logger.py
```diff
@@ -93,6 +96,73 @@ def _record_with(self, record):
...
+    1. "MLRUN_LOG_FORMATTER" = "custom" - change the default log formatter.
+    2. "MLRUN_LOG_FORMAT_OVERRIDE" = "> {timestamp} [{level}] Running module: {module} {message} {more}" - logger format
+        * Note that your custom format must include those 4 fields - timestamp, level, message and more
+    If the custom format is not configured properly , MLRun will use the default logger (human format).
+    """
+
+    def format(self, record) -> str:
+        logger = create_logger(
+            config.log_level, config.log_formatter, "logger", sys.stdout
+        )
```
liranbg: looking at it again. I think the implementation here is wrong, creating a logger _every_ time you log a message is anti pattern.
I am not sure even why do you create it - is it for the warnings only? if so you can use `print` to stderr. either way, you should not create the logger instance at all here.

In addition to that, since the config  would not change in between the messaging, you can initialize the formatter structure once and reuse it (format it) every time you log

liranbg (changes requested):

liranbg (approved):

PR #6590, opened 2024-10-27
[Schedule] Enrich kind label on upgrade [1.7.x]

roei3000b:
On [this commit](https://github.com/mlrun/mlrun/commit/bc86a7db897614b86193e38952683aaf70acafc4) we fixed the problem with enriching kind label of schedule during submit job. Meaning it fixed only from 1.7.
On this pr we created new data migration (from 1.6) that enrich the kind of the schedule.

https://iguazio.atlassian.net/browse/ML-8110

### DIFF
--- a/server/api/db/sqldb/db.py
+++ b/server/api/db/sqldb/db.py
@@ -2416,13 +2416,20 @@ class SQLDB(DBInterface):
             main_table_identifier_values=names,
         )
 
-    def align_schedule_labels(self, session: Session):
+    def align_schedule_labels(self, session: Session, enrich_kind: bool = False):
         schedules_update = []
         for db_schedule in self.list_schedules(session=session, as_records=True):
             schedule_record = self._transform_schedule_record_to_scheme(db_schedule)
             db_schedule_labels = {
                 label.name: label.value for label in db_schedule.labels
             }
+            if (
+                enrich_kind
+                and mlrun_constants.MLRunInternalLabels.kind not in db_schedule_labels
+            ):
+                # We will assume that if the kind of schedule is empty, it should be job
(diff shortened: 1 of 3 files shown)


server/api/db/sqldb/db.py
```diff
@@ -2416,13 +2416,20 @@ def delete_schedules(
...
             db_schedule_labels = {
                 label.name: label.value for label in db_schedule.labels
             }
+            if (
+                enrich_kind
+                and mlrun_constants.MLRunInternalLabels.kind not in db_schedule_labels
+            ):
+                # We will assume that if the kind of schedule is empty, it should be job
+                # because currently we only support job schedules
+                db_schedule_labels[mlrun_constants.MLRunInternalLabels.kind] = "job"
```
alonmr, line 2432: use enum for "job"

server/api/db/sqldb/db.py
```diff
@@ -2416,13 +2416,20 @@ def delete_schedules(
...
             schedule_record = self._transform_schedule_record_to_scheme(db_schedule)
             db_schedule_labels = {
                 label.name: label.value for label in db_schedule.labels
             }
+            if (
+                enrich_kind
+                and mlrun_constants.MLRunInternalLabels.kind not in db_schedule_labels
+            ):
+                # We will assume that if the kind of schedule is empty, it should be job
+                # because currently we only support job schedules
```
alonmr, line 2431: I don't think the comment is accurate. It should say something like - the current schedule kind is miss-used as it is always job. We need to enrich it for the counters but it needs to be fixed. Please rephrase a bit and mark as TODO

PR #6597, opened 2024-10-29
[ProjectSummaries] Fixing schedule count logic for jobs [1.7.x]

roei3000b:
About schedule count for jobs, instead of increasing the counter only if the kind is job, we increase it anyway if it is not workflow.
We did this because [here](https://github.com/mlrun/mlrun/pull/6389) we sync between the schedule object labels with the schedule labels. (Therefore, if for example the internal task kind is mpijob, the schedule label kind will be also mpijob and before this fix we didn't count it).

https://iguazio.atlassian.net/browse/ML-8110


server/api/db/sqldb/db.py
```diff
@@ -3004,7 +3004,7 @@ def _calculate_schedules_counters(
             project_name, schedule_name, kind = result
             if kind == mlrun_constants.MLRunInternalLabels.workflow:
                 project_to_schedule_pending_workflows_count[project_name] += 1
-            elif kind == mlrun.common.schemas.ScheduleKinds.job:
+            else:
```
alonmr: Add a comment that we check the workflow label because the schedule kind is not used properly (not setting pipelines kind for workflow schedules)

alonmr (approved):

PR #6598, opened 2024-10-29
[DB] Implement cascade deletion constraints

alonmr:
DB level cascade constraints on foreign keys:
* Tag
* Label
* Feature
* Entity

Deleting project resources is inefficient because we don't have cascade constraint on tags and labels so we have to get the resources, delete the tags and labels and then delete the resources. Then, we create a huge query with all of the parent resource ids and check if the tags/labels match them.
Instead, we can add a cascade constraint and just delete the resources without needing to the fetch them beforehand. This moves the responsibility to the DB layer which is much more efficient.

Cascade explanation with foreign key and ORM relationships:
https://docs.sqlalchemy.org/en/14/orm/cascades.html#using-foreign-key-on-delete-cascade-with-orm-relationships

https://iguazio.atlassian.net/browse/ML-8080


server/api/db/sqldb/db.py
```diff
@@ -2439,13 +2458,21 @@ def align_schedule_labels(self, session: Session):
...
     @staticmethod
     def _delete_multi_objects(
         session: Session,
         main_table: mlrun.utils.db.BaseModel,
-        related_tables: list[mlrun.utils.db.BaseModel],
         project: str,
-        main_table_identifier: str,
+        related_tables: typing.Optional[list[mlrun.utils.db.BaseModel]] = None,
+        main_table_identifier: typing.Optional[Column] = None,
         main_table_identifier_values: typing.Union[str, list[str]] = None,
```
```suggestion
        main_table_identifier_values: typing.Optional[typing.Union[str, list[str]]] = None,
```

server/api/db/sqldb/db.py
```diff
@@ -2439,13 +2458,21 @@ def align_schedule_labels(self, session: Session):
...
+        )
+
     @staticmethod
     def _delete_multi_objects(
         session: Session,
         main_table: mlrun.utils.db.BaseModel,
-        related_tables: list[mlrun.utils.db.BaseModel],
         project: str,
-        main_table_identifier: str,
+        related_tables: typing.Optional[list[mlrun.utils.db.BaseModel]] = None,
```
moranbental, line 2471: Update the parameters order in the docstring

moranbental (changes requested):

server/api/db/sqldb/db.py
```diff
@@ -2460,13 +2487,24 @@ def _delete_multi_objects(
...
-                tables=[main_table] + related_tables,
-            )
-            return 0
+        related_tables = related_tables or []
+        if project != "*":
+            where_clause = main_table.project == project
+            if main_table_identifier:
+                where_clause = and_(
+                    where_clause,
+                    main_table_identifier.in_(main_table_identifier_values),
```
roei3000b, line 2508: main_table_identifier_values can be None, in that case don't we want to skip the deletion?

server/api/db/sqldb/models.py
```diff
@@ -57,23 +57,27 @@ def post_table_definitions(base_cls):
         cls for cls in base_cls.__subclasses__() if hasattr(cls, "Notification")
     ]
     _classes = [cls for cls in base_cls.__subclasses__()]
+    for _labeled_cls in _labeled:
+        _labeled_cls.Label.parent = relationship(_labeled_cls, back_populates="labels")
+    for _tagged_cls in _tagged:
+        _tagged_cls.Tag.parent = relationship(_tagged_cls, back_populates="tags")
```
TomerShor: Add a comment on what you are doing here

server/api/db/sqldb/db.py
```diff
@@ -2478,33 +2516,7 @@ def _delete_multi_objects(
...
-                            main_table_identifier == value
-                            for value in main_table_identifier_values
-                        )
-                    )
-                    .subquery()
-                )
+            subquery = select(cls.id).join(main_table).where(where_clause).subquery()
             stmt = (
                 delete(cls)
                 .where(cls.id.in_(aliased(subquery)))
```
TomerShor, line 2529: I don't fully understand why do you still need to delete the related tables manually, if we add teh CASCADE option.
I see from the docs that all of the related tables [need to be loaded](https://docs.sqlalchemy.org/en/14/orm/cascades.html#using-foreign-key-on-delete-cascade-with-orm-relationships:~:text=If%20the%20my_parent.children%20collection%20is%20unloaded%2C%20then%20no%20DELETE%20statements%20are%20emitted.%20If%20the%20relationship.passive_deletes%20flag%20were%20not%20set%20on%20this%20relationship()%2C%20then%20a%20SELECT%20statement%20for%20unloaded%20Child%20objects%20would%20have%20been%20emitted.), so it might makes sense to run the select as above, but when the parent table record is deleted, the children should be deleted automatically...

Perhaps there's something I'm missing?

roei3000b (commented):
Looks geat, small question.

server/api/db/sqldb/db.py
alonmr: I left this to allow using this method with other relationships that are not cascading although I don't see a need for it at this time

TomerShor: So won't it make the optimization redundent?
As you are still deleting the related tables first, then the main table, so you don't really optimize

server/api/db/sqldb/db.py
```diff
@@ -4465,6 +4472,7 @@ def _delete_tagged_object(
         commit=True,
         **kwargs,
     ):
+        # TODO: Tag is now cascaded in the DB level so this should not be needed anymore
```
moranbental, line 4475: If it is not needed, why not delete it?

moranbental (changes requested):

server/api/db/sqldb/db.py
alonmr: It requires some changes that I prefer not to do right now. You need to rewrite how to delete objects by tag

liranbg (approved):

server/api/db/sqldb/db.py
alonmr: no because I don't pass the related tables in the params so they are skipped

TomerShor (approved):

PR #6599, opened 2024-10-29
[Workflows] Prevent project fields from being overwritten by remote/scheduled workflows

yaelgen:
This PR refactors the `WorkflowRunners` class to enhance code clarity and maintainability by:

Splitting WorkflowRunner into three classes:
- `BaseRunner`: Provides shared functionality.
- `LoadRunner`: Dedicated to project loading.
- `WorkflowRunners`: Focused on workflow execution.

Separating the `load_and_run` function into two handlers:
`project_loader`: Loads the project for `LoadRunner`.
`load_and_run_workflow`: Loads the project and runs the workflow/schedule - for `WorkflowRunners`.

Key Changes:
- Removed the `load_only` parameter to simplify logic and avoid conditional statements.
- Changed behavior of `load_and_run_workflow` only for `WorkflowRunners`:
 The function now loads the project without saving it, solely to fetch remote resource files needed by the runner.
It then retrieves the project using `get_or_create_project`, which gets it from the database if it exists, or creates it from the previously loaded local directory if it doesn't.
This ensures that the runner has access to remote files without overwriting existing project data in the database from the remote project.yaml, and resolves [ML-7801](https://iguazio.atlassian.net/browse/ML-7801).
In the future, we plan to change the PUT behavior of `load_project` into a PATCH on the client side, and this change will no longer be necessary ([ML-7810]( https://iguazio.atlassian.net/browse/ML-7810))

[ML-7801]: https://iguazio.atlassian.net/browse/ML-7801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/projects/pipelines.py
```diff
@@ -1109,18 +1094,156 @@ def load_and_run(
...
+        url=url,
+        name=project_name,
+        init_git=init_git,
+        subpath=subpath,
+        clone=clone,
+        save=save,
+        sync_functions=True,
+    )
+
+    context.logger.info(f"Loaded project {project.name} successfully")
```
liranbg, line 1249:
```suggestion
    context.logger.info("Successfully loaded project", project=project.name)
```

tests/api/crud/test_workflows.py
```diff
@@ -154,54 +154,58 @@ def test_run_workflow_with_local_source(
             .secret_params.get("secret", "")
             .startswith("mlrun.notifications.")
         )
-        assert run.spec.handler == "mlrun.projects.load_and_run"
```
liranbg, line 157: could it break older exported runs/workflows that specifying the handler that way?
perhaps we can aliasing `load_and_run` to its new representation (load_and_run_workflow)

liranbg (commented):
first round, minor comments

liranbg:
closing in favour of https://github.com/mlrun/mlrun/pull/6620

PR #6600, opened 2024-10-30
[CI] Introducing per-PR OS system test

liranbg:
This PR introducing the capability to run open source system test against a given PR, on demand. The motivation is to be able to raise the confidence level of PRs being merged by running the PR code on a patched version of MLRun CE.

MLRun Community Edition is MLOps stack which gives us the easiness of spinning MLRun in production-like environments, and this proposed PR, would build the derived MLRun version from users PR and patch it onto MLRun CE environments.

The build phase currently builds mlrun **api**, **log collector** and **mlrun** images.

> NOTE: The api/log collector are patched to the MLRun CE helm chart while `mlrun` image is inferred during runtime, thus, no need to patch it onto the helm chart.

In addition to that, Ive added the ability to select the `pytest` markers to run, so that once MLRun system tests are marked correctly, it would be easy to run specific test suites, e.g: "nuclio or feature_set" / "secrets" so only specific tests would run.

A horror story from trying to run MLRun CE atop minikube using docker-driver and github-runners:
MySQL failed to start because it could not create a unix-socket as OS reported with permission denied so we fallback to use sqllite as a workaround. Trying to fix it led me to understand that this is due to github runners running azure-flavor docker runtime which some hardening or modifications that limits the way mysql pod containers shared a volume. since mlrun's db is mysql that uses init container to prepare its data and then the main container simply takes the data and start the container, it failed because the init container data was reported (wrongly) to belong to another user (they had same user id).
Changing minikube atop docker-driver to **k3s** (containerd) solved it perfectly.


.github/workflows/system-tests-opensource.yml
```diff
@@ -55,165 +56,215 @@ on:
...
+      - name: Set computed versions params
+        id: computed_params
+        run: |
+          export mlrun_hash=$(echo ${{ steps.git_upstream_info.outputs.mlrun_hash }})
+          export ui_hash=$(echo ${{ steps.git_upstream_info.outputs.ui_hash }})
+          echo "mlrun_hash=$mlrun_hash" >> $GITHUB_OUTPUT
+          echo "ui_hash=$ui_hash" >> $GITHUB_OUTPUT
+
+          echo "mlrun_version=$(echo ${{ steps.git_upstream_info.outputs.unstable_version_prefix }}+$mlrun_hash)" >> $GITHUB_OUTPUT
+          echo "mlrun_ui_version=${{ steps.git_upstream_info.outputs.unstable_version_prefix }}-$ui_hash" >> $GITHUB_OUTPUT
```
alonmr, line 114: this might be needed
```suggestion
          echo "mlrun_ui_version=$(echo ${{ steps.git_upstream_info.outputs.unstable_version_prefix }}-$ui_hash)" >> $GITHUB_OUTPUT
```

Makefile
```diff
@@ -48,6 +48,7 @@ MLRUN_RAISE_ON_ERROR ?= true
 MLRUN_SKIP_CLONE ?= false
 MLRUN_RELEASE_NOTES_OUTPUT_FILE ?=
 MLRUN_SYSTEM_TESTS_CLEAN_RESOURCES ?= true
+MLRUN_SYSTEM_TEST_MARKERS ?=
```
alonmr, line 51:
```suggestion
MLRUN_SYSTEM_TEST_MARKERS ?= "not enterprise"
```
and then you don't need the if in
```
-m $(if $(MLRUN_SYSTEM_TEST_MARKERS),"$(MLRUN_SYSTEM_TEST_MARKERS)","not enterprise") \
```
it can be
```
-m $(MLRUN_SYSTEM_TEST_MARKERS) \
```
although it's not exactly the same because then users can set MLRUN_SYSTEM_TEST_MARKERS="" if we don't want to allow it

automation/system_test/k3s_docker_registry.yaml
```diff
@@ -0,0 +1,119 @@
...
+          mountPath: /var/lib/registry
+        env:
+        - name: REGISTRY_HTTP_ADDR
+          value: :5000
+        - name: REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY
+          value: /var/lib/registry
+      volumes:
+      - name: storage
+        persistentVolumeClaim:
+          claimName: docker-registry
```
alonmr: newline

.github/workflows/system-tests-opensource.yml
```diff
@@ -55,165 +56,215 @@ on:
...
+
     # since github-actions gives us 14G only, and fills it up with some garbage
     - name: Freeing up disk space
       run: |
         "${GITHUB_WORKSPACE}/automation/scripts/github_workflow_free_space.sh"
-    - name: Set up python
-      uses: actions/setup-python@v5
+
+    - name: Install uv
+      uses: astral-sh/setup-uv@v3
```
alonmr, line 164: I think if you setup python and then setup uv then you don't need to `source venv/bin/activate` on each step
https://github.com/charliermarsh/autobot/blob/e42c66659bf97b90ca9ff305a19cc99952d0d43f/.github/workflows/ci.yaml
Not sure about it but worth checking

alonmr (commented):
Amazing! here is my 2cents

.github/workflows/build-internal.yaml
```diff
@@ -91,12 +107,14 @@ jobs:
       run: |
         export version_suffix=$(echo "$GITHUB_REF_NAME" | grep -E "^[0-9]+\.[0-9]+\.x$" | tr -d '.');
         export unstable_tag=$(if [ -z "$version_suffix" ]; then echo "unstable-cache"; else echo "unstable-cache-$version_suffix";fi);
+        export unstable_tag=$(if [ -z "$INPUT_CACHE_TAG_SUFFIX" ]; then echo "$unstable_tag" ; else echo "$unstable_tag-$INPUT_CACHE_TAG_SUFFIX";fi);
```
TomerShor:
```suggestion
        unstable_tag="unstable-cache${version_suffix:+-$version_suffix}${INPUT_CACHE_TAG_SUFFIX:+-$INPUT_CACHE_TAG_SUFFIX}"
        export unstable_tag
```

.github/workflows/system-tests-opensource.yml
```diff
@@ -55,165 +56,215 @@ on:
...
+        detached: true
+
+    - name: Setup Local Docker Registry
+      run: |
+        # fill /etc/rancher/k3s/registries.yaml
+        registries_file="/etc/rancher/k3s/registries.yaml"
+        echo "mirrors:"                              | sudo tee $registries_file
+        echo "  registry.localhost:"                 | sudo tee --append $registries_file
+        echo "    endpoint:"                         | sudo tee --append $registries_file
+        echo "    - http://registry.localhost:80"  | sudo tee --append $registries_file
```
TomerShor:
```suggestion
        sudo tee "$registries_file" << EOF
        mirrors:
          registry.localhost:
            endpoint:
            - http://registry.localhost:80
        EOF
```

.github/workflows/system-tests-opensource.yml
```diff
@@ -17,34 +17,35 @@ name: System Tests Open Source
...
-        required: true
-        default: 'mlrun'
+      pr_number:
+        description: 'PR number to run the system tests against (default: run on the latest commit in the development branch)'
+        required: false
+        default: ''
+      pytest_markers:
+        description: 'Pytest markers to run'
+        required: false
+        default: 'not enterprise and smoke'
```
TomerShor, line 38: general question - are the used/supported markers written anywhere?

TomerShor (changes requested):
Looks good!
Some minor comments

.github/workflows/system-tests-opensource.yml
liranbg: it is actually not needed. the variables are templates without the need to echo-ed them

Makefile
liranbg: well, I agree we need to ensure that some of the markers are not given to the developers and are run regardless. Id take it on a followup pr
regarding the variable, I thought it could be served for the enterprise system test afterwards, and thus I did not set a global default but rather a test-specific one

.github/workflows/system-tests-opensource.yml
liranbg: note that they run uv run python so it is basically like sourcing the python venv

the reason I had to install python and venv is because I can't use system python but rather the venv one

.github/workflows/system-tests-opensource.yml
liranbg: u manually create markers  by using `pytest.markers.<X>` where X is any arbitrary marker

alonmr (approved):
LGTM

PR #6603, opened 2024-10-30
[Alert Activations] Create alert activations table

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-8132

This PR introduces the `alert_history` table with range-based partitioning by `activation_time` to optimize time-based data queries. The partitioning is configured by` YEARWEEK(activation_time)` by default. 

**Key points**
* Partitioning Configuration:
  MySQL requires at least one partition to be defined when setting up a range-based partition. As such, an initial partition (p0) is created by default to ensure the table is operable upon creation. This default partition can be adjusted or expanded as necessary by setting `PARTITION_INTERVAL` to `DAY`, `MONTH` and etc ([everything that mysql supports](https://dev.mysql.com/doc/refman/8.4/en/date-and-time-functions.html)). Partitioning is to be configured when we already created a table in order to make a partitioning configurable and allow this only on migration. 
* Primary Key Selection:
The primary key now includes both `activation_time` and `id` to satisfy MySQL's requirement that a primary key contains any column used in the partitioning expression.

* `Data` field:
The idea of data field is to keep both criteria and notification state of alert history there.


server/api/db/sqldb/models.py
```diff
@@ -769,6 +769,36 @@ class TimeWindowTracker(Base, mlrun.utils.db.BaseModel):
         def get_identifier_string(self) -> str:
             return f"{self.key}"
 
+    class AlertHistory(Base, mlrun.utils.db.BaseModel):
+        __tablename__ = "alert_history"
+        __table_args__ = (
+            PrimaryKeyConstraint("activation_time", "id", name="_alert_history_uc"),
+        )
```
liranbg: I suggest to add indexes according to the expected lists we will want to support (from ui perspective) 
like "all project alerts" -> add index on projects / "all kind X / entity" and so on. I suggest to follow the UI narrative to begin with

server/api/migrations/versions/38b2f0b84eef_add_alert_history.py
```diff
@@ -0,0 +1,85 @@
...
+
+    # Add initial partitioning
+    partition_interval = os.getenv("PARTITION_INTERVAL", "YEARWEEK")
+    default_partition_value = os.getenv("DEFAULT_PARTITION_VALUE", "2023-01-01")
+    partition_sql = f"""
+        ALTER TABLE alert_history
+        PARTITION BY RANGE ({partition_interval}(activation_time)) (
+            PARTITION p0 VALUES LESS THAN ({partition_interval}('{default_partition_value}'))
+        );
+    """
```
liranbg: was it not possible to achieve the partitioning using the orm definitions?

liranbg (changes requested):
minor

server/api/migrations/versions/38b2f0b84eef_add_alert_history.py
rokatyy: @liranbg no, it's not supported in orm

server/api/db/sqldb/models.py
```diff
@@ -769,6 +769,36 @@ class TimeWindowTracker(Base, mlrun.utils.db.BaseModel):
...
+
+        id = Column(Integer)
+        activation_time = Column(SQLTypesUtil.datetime(), nullable=False)
+        name = Column(String(255, collation=SQLTypesUtil.collation()), nullable=False)
+        project = Column(
+            String(255, collation=SQLTypesUtil.collation()), nullable=False
+        )
+        data = Column(JSON)
```
TomerShor, line 886: You can use `_full_object` for this, and we can maybe store the entire object here and not only specific data.
This aligns with what we are doing with all other models, and provides the dumping capabilities using properties.

server/api/migrations/versions/67b0169ba6b4_add_alert_history.py
```diff
@@ -0,0 +1,107 @@
...
+        "ix_alert_history_project", "alert_history", ["project"], unique=False
+    )
+    op.create_index(
+        "ix_alert_history_severity", "alert_history", ["severity"], unique=False
+    )
+    # ### end Alembic commands ###
+
+    # Add initial partitioning
+    partition_interval = os.getenv("PARTITION_INTERVAL", "YEARWEEK")
+    default_partition_value = os.getenv("DEFAULT_PARTITION_VALUE", "2023-01-01")
```
TomerShor: This migration will run once, and then the revision will be set to the revision defined in this file.
How will we be able to modify these partitioning values later?

TomerShor (changes requested):

server/api/db/sqldb/models.py
Yacouby: I agree with @TomerShor , you can look at the full_object that we used in the AlertConfig & AlertState classes.

server/api/migrations/versions/67b0169ba6b4_add_alert_history.py
rokatyy: @TomerShor we don't have such a requirement. Technically it is possible, but according to the requirement we should not support this.

server/api/db/sqldb/models.py
rokatyy: @TomerShor @Yacouby I chose JSON for the following reasons:

1. We won’t use it to store fields that require validation.
2. Performance considerations:
* Direct JSON columns eliminate the need for manual json.loads or json.dumps conversions in Python, saving CPU cycles and improving response time.
* JSON data stored directly in the database benefits from database-level optimizations, resulting in faster data access compared to using properties or custom serializers in SQLAlchemy.


Are there specific benefits to using _full_object over JSON in this case, aside from aligning with other tables? Just as a note, we use a standard JSON column for project summaries..

server/api/migrations/versions/67b0169ba6b4_add_alert_history.py
TomerShor: So what if QA will want to have partitioning < 1 week?
Will it not be possible?

rokatyy: @TomerShor it is possible by setting `PARTITION_INTERVAL` envvar on initial setup.

rokatyy: From requirement:

```
partitioning granularity is configurable. This should be considered under the following restrictions:

* This configuration will not be dynamic nor will it be able to change during upgrades. Meaning that a system installed with partitioning granularity X will have to keep that granularity across future upgrades. There will not be any migration process.

* Given this limitation, this configuration (if available) will only be used by QA to facilitate testing. We should not document nor use it in customer deployments.
```

server/api/db/sqldb/models.py
```diff
@@ -769,6 +769,42 @@ class TimeWindowTracker(Base, mlrun.utils.db.BaseModel):
...
+    class AlertHistory(Base, mlrun.utils.db.BaseModel):
+        __tablename__ = "alert_history"
+        __table_args__ = (
+            PrimaryKeyConstraint("activation_time", "id", name="_alert_history_uc"),
+            Index("ix_alert_history_project", "project"),
+            Index("ix_alert_history_entity_kind", "entity_kind"),
+            Index("ix_alert_history_event_kind", "event_kind"),
+            Index("ix_alert_history_severity", "severity"),
+            Index("ix_alert_history_entity_id", "entity_id"),
+            Index("ix_alert_history_name", "name"),
```
theSaarco: Some of these indexes will not have any benefit and will most likely never be used, for example consider the `severity` field - it is a field with very few distinct values and the DB will probably never use it as part of a query plan. Same for `entity_kind`.
Reason is that the DB maintains statistics for each index, and one of the important stats is the cardinality of the index. If the DB knows that the cardinality of an index is 3, and the table contains 100,000 records then there really is little benefit for using this index since it's expected to return 1/3 if the records, and the DB is likely to use any other index instead (or resort to scanning the table). 
Usually you'd want to create indexes on fields/combination of fields that will generate a good query optimization under specific queries that are performed. For example, in our case an index on project + name would be a good combination. Also an index on the activation time will be meaningful here since we expect queries on "bring me activations on last 24h" etc.

server/api/migrations/versions/67b0169ba6b4_add_alert_history.py
```diff
@@ -0,0 +1,107 @@
...
+    partition_sql = f"""
+         ALTER TABLE alert_history
+         PARTITION BY RANGE ({partition_interval}(activation_time)) (
+             PARTITION p0 VALUES LESS THAN ({partition_interval}('{default_partition_value}'))
+         );
```
theSaarco: This depends on values retrieved from outside, so at a minimum you should verify that both `partition_interval` and `default_partition_value` don't contain special chars and only contain valid chars for SQL.
The other reason (even though less of a concern here due to being a specific installation flow) is that this is subject to SQL-injection exploits. Many static-analysis tools will catch this code and warn against it. Ideally other ways should be used to inject parameters into SQL statements executed.

server/api/migrations/versions/67b0169ba6b4_add_alert_history.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index("ix_alert_history_severity", table_name="alert_history")
+    op.drop_index("ix_alert_history_project", table_name="alert_history")
+    op.drop_index("ix_alert_history_name", table_name="alert_history")
+    op.drop_index("ix_alert_history_event_kind", table_name="alert_history")
+    op.drop_index("ix_alert_history_entity_kind", table_name="alert_history")
+    op.drop_index("ix_alert_history_entity_id", table_name="alert_history")
```
theSaarco: Do you need to drop the indexes before dropping the table? I think this is redundant.

theSaarco (changes requested):
Looks good, I do have some comments that need addressing.

server/api/migrations/versions/67b0169ba6b4_add_alert_history.py
rokatyy: @theSaarco agreed 💯 , I've rewritten the logic.

server/api/db/sqldb/models.py
TomerShor: I do see some other "full_object" fields that use a direct JSON Column.
Not sure there is another specific benefit for it, but I guess calling "data" is not reflective of what it actually stores.
Storing the full object enables us to add more non-columned fields to the object in the future without changing the code db code.

server/api/migrations/versions/650f0ce2da6f_add_alert_history.py
```diff
@@ -0,0 +1,130 @@
...
+    else:
+        partition_value = now_utc.strftime("%Y-%m-%d")
+        partition_expression = f"{partition_interval}(activation_time)"
+
+    # Construct SQL for partitioning
+    partition_sql = f"""
+        ALTER TABLE alert_history
+        PARTITION BY RANGE ({partition_expression}) (
+            PARTITION p{partition_name} VALUES LESS THAN ({partition_value})
+        );
```
TomerShor: I would move it to a separate function, e.g. `create_partition_query`, to keep it more contained and concise

TomerShor (changes requested):

server/api/migrations/versions/650f0ce2da6f_add_alert_history.py
rokatyy: @TomerShor moved partly to alert history crud. As SQL part should still be there as it's specific to initial migration..

server/api/db/sqldb/models.py
rokatyy: During slack discussion it was decided to go with JSON field, so resolving the comment

server/api/db/sqldb/models.py
```diff
@@ -769,6 +769,38 @@ class TimeWindowTracker(Base, mlrun.utils.db.BaseModel):
...
+            Index("ix_alert_activation_activation_time", "activation_time"),
+        )
+
+        id = Column(Integer)
+        activation_time = Column(SQLTypesUtil.datetime(), nullable=False)
+        name = Column(String(255, collation=SQLTypesUtil.collation()), nullable=False)
+        project = Column(
+            String(255, collation=SQLTypesUtil.collation()), nullable=False
+        )
+        data = Column(JSON)
```
Yacouby, line 886: in this field you mean the notifications list data ? if yes then maybe we should change the name to make its purpose clearer ( maybe notification_details ? )

Yacouby (commented):
LGTM, minor comment

server/api/db/sqldb/models.py
rokatyy: The idea of data field is to keep both criteria and notification state of alert history there.

server/api/crud/alert_activation.py
```diff
@@ -0,0 +1,44 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+import mlrun.utils.singleton
+
+valid_partition_intervals = ["DAY", "YEARWEEK", "MONTH"]
+partition_name_format_mapping = {
+    "YEARWEEK": "%Y%V",  # ISO week number (for weeks starting Monday)
```
theSaarco:
```suggestion
    "YEARWEEK": "%Y%W",  # ISO week number (for weeks starting Monday)
```

server/api/crud/alert_activation.py
```diff
@@ -0,0 +1,44 @@
...
+    "MONTH": "%Y%m",  # Format as 'YYYYMM' (year and month)
+    "DAY": "%Y%m%d",  # Format as 'YYYYMMDD' (year, month, day)
+}
+
+
+class AlertActivation(
+    metaclass=mlrun.utils.singleton.Singleton,
+):
+    @staticmethod
+    def get_partition_info_for_datetime(partition_interval, partition_datetime):
```
theSaarco: It would help to add a docstring explaining what this function returns, as it returns a 3-way tuple which is a bit hard to follow.

server/api/crud/alert_activation.py
```diff
@@ -0,0 +1,44 @@
...
+        partition_name = partition_datetime.strftime(
+            partition_name_format_mapping[partition_interval]
+        )
+
+        if partition_interval == "YEARWEEK":
+            partition_value = int(
+                partition_datetime.strftime("%Y%V")
```
theSaarco:
```suggestion
                partition_datetime.strftime("%Y%W")
```

server/api/migrations/versions/650f0ce2da6f_add_alert_activation.py
```diff
@@ -0,0 +1,119 @@
...
+    valid_partition_intervals = ["DAY", "YEARWEEK", "MONTH"]
+
+    # Validate the partition interval
+    if partition_interval not in valid_partition_intervals:
+        raise ValueError(
+            f"Partition interval can only be one of the following: {valid_partition_intervals}"
+        )
+
+    # Current UTC date
+    now_utc = datetime.utcnow()
```
theSaarco, line 96: This value is going to be used as the `LESS THAN` value in the partition created. Of course, we're not directly using it but rather its translation to day/month/week, but it's still possible that the only partition created at this step will be already expired immediately after - this will lead into SQL errors when trying to insert to the table until a new partition is created.
Since we're using `LESS THAN`, it makes sense to create the 1st partition to be less than now + (partition interval), so the partition will just start at `now`, not end at it.

Another option is to create a "catchall" partition using `LESS THAN MAXVALUE` to handle situations where a partition is not created for some reason. See https://dev.mysql.com/doc/refman/8.0/en/partitioning-range.html for details. I think we should not go for this approach, but it's something to keep in mind.

theSaarco (changes requested):
Some small fixes, and comments on the partition management (initial partition should be forward-looking, not ending at `now`).

server/api/crud/alert_activation.py
rokatyy: @theSaarco In MySQL, using `YEARWEEK(date, 1)` aligns with the ISO 8601 standard for week numbering, which considers Monday as the first day of the week and defines the first week of the year as the one containing at least four days in the new year. To correctly match this ISO I used `V` formatter

server/api/migrations/versions/650f0ce2da6f_add_alert_activation.py
rokatyy: @theSaarco the idea of this PR is to create a migration which creates a table. That's mysql limitation that partitioned table can't exist without at least 1 partition. As a part of next ticket, the logic of creating partitions on start of chief for the next weeks will be implemented, so we will be safe there.

server/api/crud/alert_activation.py
rokatyy: @theSaarco Added 👍

server/api/crud/alert_activation.py
```diff
@@ -0,0 +1,64 @@
...
+    "MONTH": "%Y%m",  # Format as 'YYYYMM' (year and month)
+    "DAY": "%Y%m%d",  # Format as 'YYYYMMDD' (year, month, day)
+}
+
+
+class AlertActivation(
+    metaclass=mlrun.utils.singleton.Singleton,
+):
+    @staticmethod
+    def get_partition_info_for_datetime(partition_interval, partition_datetime):
```
liranbg:
```suggestion
    def get_partition_info(partition_interval, partition_datetime):
```

the _for_x_ is inferred by the `partition_datetime` thus, redundant

server/api/db/sqldb/models.py
```diff
@@ -869,6 +869,38 @@ class TimeWindowTracker(Base, mlrun.utils.db.BaseModel):
         def get_identifier_string(self) -> str:
             return f"{self.key}"
 
+    class AlertActivation(Base, mlrun.utils.db.BaseModel):
+        __tablename__ = "alert_activation"
+        __table_args__ = (
+            PrimaryKeyConstraint("activation_time", "id", name="_alert_activation_uc"),
+            Index("ix_alert_activation_project_name", "project", "name"),
+            Index("ix_alert_activation_activation_time", "activation_time"),
```
liranbg, line 877: 💯

liranbg (commented):
minor suggestions

server/api/crud/alert_activation.py
```diff
@@ -0,0 +1,64 @@
...
+    "MONTH": "%Y%m",  # Format as 'YYYYMM' (year and month)
+    "DAY": "%Y%m%d",  # Format as 'YYYYMMDD' (year, month, day)
+}
+
+
+class AlertActivation(
+    metaclass=mlrun.utils.singleton.Singleton,
+):
+    @staticmethod
+    def get_partition_info(partition_interval, partition_datetime):
```
TomerShor: add typehints (for return value too)

server/api/db/sqldb/models.py
```diff
@@ -869,6 +869,38 @@ class TimeWindowTracker(Base, mlrun.utils.db.BaseModel):
...
+        event_kind = Column(
+            String(255, collation=SQLTypesUtil.collation()), nullable=False
+        )
+        severity = Column(
+            String(255, collation=SQLTypesUtil.collation()), nullable=False
+        )
+        number_of_events = Column(Integer, nullable=False)
+
+        def get_identifier_string(self) -> str:
+            return f"{self.id}"
```
TomerShor: since you have already the project name column, add the project to the identifier string. It will make it easier for debugging (it is printed in the logs in case of failures).
```suggestion
        def get_identifier_string(self) -> str:
            return f"{self.project}/{self.id}"
```

TomerShor (changes requested):
minor

server/api/crud/alert_activation.py
theSaarco: :+1: I guess my concern is that `%V` is an extended format, not necessarily supported on all platforms and in some cases not documented (as mentioned [here](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes)). Given that this is strictly server-side code, I guess that's not really an issue. 

Out of curiosity, is there a specific reason you went with the ISO-8601 implementation for `YEARWEEK` (mode 1)? It supports many other modes (and that is not the default mode).

tests/api/crud/test_alert_activations.py
```diff
@@ -0,0 +1,67 @@
...
+            "202410",
+            "2024-11-01",
+            "MONTH(activation_time)",
+        ),
+        (
+            "YEARWEEK",
+            datetime(2024, 10, 30),
+            "202444",
+            "202445",
+            "YEARWEEK(activation_time, 1)",
```
theSaarco, line 45: Given our discussion on ISO format and other formats, it may be useful to test an edge case where mode 1 of `YEARWEEK` is different than other modes (week 53 last year vs. week 1 this year).

theSaarco (approved):
👍 
Nicely done. Minor suggestion, but otherwise excellent.

server/api/crud/alert_activation.py
rokatyy: @theSaarco
> Out of curiosity, is there a specific reason you went with the ISO-8601 implementation for YEARWEEK (mode 1)? It supports many other modes (and that is not the default mode).
 
 just because in python it is much easier to work with weeks which start from Monday rather than from Sunday

tests/api/crud/test_alert_activations.py
rokatyy: @theSaarco Actually, I noticed that it works incorrectly at times with the 31st of December, resulting in mismatches with MySQL outputs. Therefore, I decided to eliminate the use of %V completely and switched to the logic with isocalendar(). Additionally, I added more test cases and populated each test with values from the actual database output.

server/api/db/sqldb/models.py
rokatyy: @TomerShor but we can have many alerts per project, how it will help?

TomerShor: This method is used in [here](https://github.com/mlrun/mlrun/blob/4a6e0aec96d781836d6497e130ce0647580be62d/server/api/db/sqldb/db.py#L4669-L4672) for logging only.
IMO It makes more sense that the log will show the project, and not only a random id, so we'll have more information (the project at the very least) on the record that failed to commit, instead of having to look in the DB what is this record.

rokatyy: @TomerShor I think we can use {project}/{alert_name}/{id} then. What do you think?

TomerShor: @rokatyy even better 👍

server/api/crud/alert_activation.py
```diff
@@ -0,0 +1,78 @@
...
+        elif partition_interval == "MONTH":
+            partition_boundary_date = (
+                partition_datetime.replace(day=1) + timedelta(days=32)
+            ).replace(day=1)
+            partition_value = partition_boundary_date.strftime(
+                partition_name_format_mapping["MONTH"]
+            )
+            partition_expression = "MONTH(activation_time)"
+        else:
+            raise ValueError(f"Unsupported partition interval: {partition_interval}")
```
TomerShor:
```suggestion
        if partition_interval == "DAY":
            partition_boundary_date = partition_datetime + timedelta(days=1)
        elif partition_interval == "MONTH":
            partition_boundary_date = (
                partition_datetime.replace(day=1) + timedelta(days=32)
            ).replace(day=1)
        else:
            raise ValueError(f"Unsupported partition interval: {partition_interval}")
            
        partition_value = partition_boundary_date.strftime(
            partition_name_format_mapping[partition_interval]
        )
        partition_expression = f"{partition_interval}(activation_time)"
```

server/api/crud/alert_activation.py
```diff
@@ -0,0 +1,78 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from datetime import datetime, timedelta
+
+import mlrun.utils.singleton
+
+valid_partition_intervals = ["DAY", "YEARWEEK", "MONTH"]
```
TomerShor: since you're using these string multiple times, it can be an `StrEnum` class: `class PartitionIntervals(StrEnum)`, which provides an `isValid` function

server/api/migrations/versions/650f0ce2da6f_add_alert_activation.py
```diff
@@ -0,0 +1,119 @@
...
+    op.create_index(
+        "ix_alert_activation_project_name",
+        "alert_activation",
+        ["project", "name"],
+        unique=False,
+    )
+    # ### end Alembic commands ###
+
+    partition_interval = os.getenv("PARTITION_INTERVAL", "YEARWEEK").upper()
+    valid_partition_intervals = ["DAY", "YEARWEEK", "MONTH"]
```
TomerShor: duplicate of the same list in the `crud`

TomerShor (changes requested):
Last comments

server/api/migrations/versions/650f0ce2da6f_add_alert_activation.py
rokatyy: @TomerShor good catch, it should be removed from crud and be left only in migration script

TomerShor (approved):
Great stuff

liranbg (approved):

PR #6605, opened 2024-10-30
[Pagination] Artifacts Pagination Endpoint Support 

moranbental:
This PR modifies the `list_artifacts` endpoint to support pagination by adding the `page`, `page-size`, and `page-token` parameters. Essentially, it implements the pagination mechanism previously established for the initial stage of adding pagination to the `list_runs` and `list_functions` endpoints.

Additionally, it utilizes this mechanism in the SDK by sending a default paginated query, waiting until all responses are received, and returning the complete response at once.

Also, fixing enumeration member conflict for `ArtifactFormat`, `FeatureSetFormat` and `FunctionFormat`: 
```
>   ???
E   pydantic.error_wrappers.ValidationError: 1 validation error for list_artifacts_schema
E   format_
E     value is not a valid enumeration member; permitted: 'minimal' (type=type_error.enum; enum_values=[minimal])
```
This issue arises because `ArtifactFormat` inherits from both `ObjectFormat` and `StrEnum`, which can create confusion with the enumeration members. In Python, `enum.Enum` typically expects all members to be defined at the same level in the class. Mixing members from different parent classes can lead to validation issues. The solution is to define all enum members in one place.

https://iguazio.atlassian.net/browse/ML-8162


tests/api/api/test_artifacts.py
```diff
@@ -1008,6 +1009,75 @@ def test_list_artifacts_with_time_filters(db: Session, unversioned_client: TestC
...
+    artifact_path = LIST_API_ARTIFACTS_V2_PATH.format(project=PROJECT)
+    response = unversioned_client.get(
+        artifact_path,
+        params={
+            "page": 1,
+            "page-size": page_size,
+        },
+    )
+
+    tests.api.api.utils.assert_pagination_info(
```
TomerShor, line 1046: call with kw args to make it clearer:
```
expected_results_count=..., 
expected_page=..., 
```

server/api/crud/artifacts.py
```diff
@@ -144,8 +145,8 @@ def list_artifacts(
         self,
         db_session: sqlalchemy.orm.Session,
         project: str = None,
-        name: str = "",
-        tag: str = "",
+        name: typing.Optional[str] = "",
+        tag: typing.Optional[str] = "",
```
rokatyy:
```suggestion
        name: typing.Optional[str] = None,
        tag: typing.Optional[str] = None,
```

server/api/db/sqldb/db.py
```diff
@@ -1398,6 +1402,8 @@ def _find_artifacts(
         attach_tags: bool = False,
         limit: int = None,
         with_entities: list[Any] = None,
+        page: typing.Optional[int] = None,
+        page_size: typing.Optional[int] = None,
```
liranbg, line 1435: I think that we need to block requests requesting both limit and page size.
the limit is mainly used by ui to reduce the amount of artifacts they requested. so now that the endpoint is paginated, they could use that instead.
what I suggest is that
if limit is given - use as usual
if limit is given with page size, raise conflict and message should say that page_size and limit are conflicting and only one can be given

liranbg (changes requested):
minor

liranbg (approved):

TomerShor (approved):
LGTM

PR #6607, opened 2024-10-31
[Linting] Fix implicit `Optional` type hints

jond01:
Use Ruff's RUF013 rule:
https://docs.astral.sh/ruff/rules/implicit-optional/

Note: this rule is limited. It does not catch aliases or custom classes, e.g.:

* https://github.com/mlrun/mlrun/blob/8e973600d3bfb8ff6af8cfd2ce21f6a72a0c77a0/mlrun/db/base.py#L88
* https://github.com/mlrun/mlrun/blob/8e973600d3bfb8ff6af8cfd2ce21f6a72a0c77a0/mlrun/db/httpdb.py#L1518

Pyright's [reportArgumentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportArgumentType) check catches such errors. We can add it to better cover this check.
See also https://github.com/astral-sh/ruff/issues/14018.


liranbg:
@jond01 since this was an automated change, please hold until https://github.com/mlrun/mlrun/pull/6615 is merged

automation/deployment/deployer.py
```diff
@@ -48,23 +48,23 @@ class ExcecutionParams:
...
         self,
         registry_url: str,
-        registry_secret_name: str = None,
-        chart_version: str = None,
-        mlrun_version: str = None,
-        override_mlrun_api_image: str = None,
-        override_mlrun_log_collector_image: str = None,
-        override_mlrun_ui_image: str = None,
-        override_jupyter_image: str = None,
+        registry_secret_name: typing.Optional[str] = None,
```
assaf758, line 51: For brevity, please import `Optional` in modules where it's missing.

liranbg:
you have the green light to continue

automation/deployment/deployer.py
jond01: @assaf758 this change requires a lot of manual work or a clever way to detect imports across many modules.
I think it's ok for now, new code can choose to use `from typing import Optional` instead of `import typing`.

assaf758: Fair enough.
Please just fix the file conflict.

assaf758 (approved):

PR #6610, opened 2024-10-31
[Pagination] Fix Pagination Cache Serializer

quaark:
https://iguazio.atlassian.net/browse/ML-8036

The mechanism used to serialize the paginated requests into a storable object in the database was using a pydantic schema based on the paginated method's arguments. However, instead of using pydantic's `json` method and converting the arguments to a string, we used `dict` which conserves object types (like `datetime`) which sqlalchmey does not automatically support serializing in the json column. Pydantic's `json` method however produces a 'pure' JSON so any object gets serialized to a int/float/string/list/map in a way that pydantic can then decode back to the original object. (This is the whole point of the mechanism that was missed with using the `dict` method instead of the `json` method)

* Added datetime objects to the unit tests so this higher-level serialization and de-serialization are properly tested as well.

### DIFF
--- a/server/api/utils/pagination.py
+++ b/server/api/utils/pagination.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 #
 import inspect
+import json
 import typing
 
 import pydantic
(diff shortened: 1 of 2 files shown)


server/api/utils/pagination.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 #
 import inspect
+import json
```
liranbg: u may use orjson for fastness serialization

liranbg (changes requested):
small request

PR #6612, opened 2024-10-31
[Artifacts] Order artifacts by updated date DESC

moranbental:
When listing artifacts, the order was incorrect: (artifact2, artifact0, artifact4, artifact3, artifact1). 

Added `ORDER BY updated DESC` to the artifact query to sort by the updated field in descending order. 
This PR also addresses the case when a limit is applied

https://iguazio.atlassian.net/browse/ML-8179


server/api/db/sqldb/db.py
```diff
@@ -1480,6 +1480,8 @@ def _find_artifacts(
             )
 
         if limit:
+            # Order the query before applying the limit
+            query = query.order_by(ArtifactV2.updated.desc())
```
liranbg: comment _why_ not just _what_

tests/api/api/test_artifacts.py
```diff
@@ -1008,6 +1008,47 @@ def test_list_artifacts_with_time_filters(db: Session, unversioned_client: TestC
...
+    for counter in range(number_of_artifacts):
+        artifact_name = f"artifact-{counter}"
+        json_data = _generate_artifact_body(key=artifact_name)
+        resp = unversioned_client.put(
+            STORE_API_ARTIFACTS_V2_PATH.format(project=PROJECT) + f"/{artifact_name}",
+            json=json_data,
+        )
+        assert resp.status_code == HTTPStatus.OK.value
+
+    # Make the request
```
TomerShor: It's 'Do the request', but anyway you can delete the comment because it's redundant

server/api/db/sqldb/db.py
```diff
@@ -1492,6 +1494,8 @@ def _find_artifacts(
 
         outer_query = outer_query.join(subquery, ArtifactV2.id == subquery.c.id)
 
+        outer_query = outer_query.order_by(ArtifactV2.updated.desc())
```
TomerShor: If we have limit we already ordered the query, why is this needed again?
Maybe we can order first regardless of the limit?

TomerShor (changes requested):

TomerShor (approved):

liranbg (approved):

server/api/db/sqldb/db.py
moranbental: Added a comment

server/api/db/sqldb/db.py
moranbental: we need to ensure that the results are ordered after all filtering and joins.
I added a comment

tests/api/api/test_artifacts.py
TomerShor, line 1: The test is good, but I think it is in the wrong place.
This test file is under the `api` directory, so it should test the functionality of the Artifacts API.
Your changes are in the db layer, so the test fits better in `tests/api/db/test_artifacts.py`, where we test DB-related functionality.

tests/api/db/test_artifacts.py
```diff
@@ -1179,6 +1179,38 @@ def test_list_artifact_for_tagging_fallback(
...
+        # Create artifacts
+        number_of_artifacts = 10
+        for counter in range(number_of_artifacts):
+            artifact_key = f"artifact-{counter}"
+            artifact_body = self._generate_artifact(artifact_key, project=project)
+            db.store_artifact(db_session, artifact_key, artifact_body, project=project)
+
+        artifacts = db.list_artifacts(db_session, project=project, limit=limit)
+
+        expected_count = limit if limit else number_of_artifacts
```
TomerShor:
```suggestion
        expected_count = limit or number_of_artifacts
```

TomerShor (changes requested):
last thing

PR #6620, opened 2024-11-03
[Workflows] Prevent project fields from being overwritten by remote/scheduled workflows [1.7.x]

yaelgen:
This PR refactors the `WorkflowRunners` class to enhance code clarity and maintainability by:

Splitting WorkflowRunner into three classes:
- `BaseRunner`: Provides shared functionality.
- `LoadRunner`: Dedicated to project loading.
- `WorkflowRunners`: Focused on workflow execution.

Separating the `load_and_run` function into two handlers:
`project_loader`: Loads the project for `LoadRunner`.
`load_and_run_workflow`: Loads the project and runs the workflow/schedule - for `WorkflowRunners`.

Key Changes:
- Removed the `load_only` parameter to simplify logic and avoid conditional statements.
- Changed behavior of `load_and_run_workflow` only for `WorkflowRunners`:
 The function now loads the project without saving it, solely to fetch remote resource files needed by the runner.
It then retrieves the project using `get_or_create_project`, which gets it from the database if it exists, or creates it from the previously loaded local directory if it doesn't.
This ensures that the runner has access to remote files without overwriting existing project data in the database from the remote project.yaml, and resolves [ML-7801](https://iguazio.atlassian.net/browse/ML-7801).
In the future, we plan to change the PUT behavior of `load_project` into a PATCH on the client side, and this change will no longer be necessary ([ML-7810]( https://iguazio.atlassian.net/browse/ML-7810))

[ML-7801]: https://iguazio.atlassian.net/browse/ML-7801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/projects/pipelines.py
```diff
@@ -984,7 +984,7 @@ def github_webhook(request):
     return {"msg": "pushed"}
 
 
-def load_and_run(
```
liranbg, line 987: add alias function to load_and_run that points to load_and_run_workflow with relevant flags

mlrun/projects/pipelines.py
```diff
@@ -984,14 +984,47 @@ def github_webhook(request):
...
+):
+    """
+    This function serves as an alias to `load_and_run_workflow`,
+    allowing to continue using `load_and_run` without modifying existing workflows or exported runs.
+    This approach ensures backward compatibility,
+    while directing all new calls to the updated `load_and_run_workflow` function.
+    """
+    args = locals().copy()
+    args.pop("load_only")
+    load_and_run_workflow(**args)
```
liranbg: using the `locals().copy()` is inefficient as you copy all we have here, you can change the function signature to
`def load_and_run(*args, **kwargs):`
and run `kwargs.pop("load_only", None)`

liranbg (approved):

mlrun/projects/pipelines.py
```diff
@@ -1109,18 +1098,156 @@ def load_and_run(
...
     if wait_for_completion:
+        handle_workflow_completion(
+            run=run,
+            project=project,
+            context=context,
+            workflow_log_message=workflow_log_message,
+        )
+
+
+def load_remote_project(
```
quaark: I suggest changing the name of this method (and maybe `project_loader` as well) to be more specific to their goals. 
It's quite confusing trying to track which combination of the words `load` `remote` and `project` result in which kind of functionality, especially since they all eventually call `mlrun.load_project` with different parameters.

I'm thinking maybe:
1. `project_loader` -> `import_remote_project` or `load_remote_project_to_db`
2. `load_remote_project` -> `pull_remote_project_files`

quaark (changes requested):
Looks great IMO, I have only a slight naming change I suggest

quaark (approved):
LGTM 🚀

PR #6621, opened 2024-11-03
[API] Block worker requests when instance not online [1.7.x]

liranbg:
This was raised during an upgrade from mlrun version X to Y where the migration completed, chief moved to state "migration completed" and worker started to handle requests and failed because the functionality of "move api to online" was yet to complete. this results in failing some of the jobs.

This solution makes the worker handle requests only when `online` state is set to ensure black listing all states other than online.

https://iguazio.atlassian.net/browse/ML-8184

### DIFF
--- a/server/api/api/deps.py
+++ b/server/api/api/deps.py
@@ -74,6 +74,23 @@ def verify_api_state(request: Request):
             )
             raise mlrun.errors.MLRunPreconditionFailedError(message)
 
+    # defensive approach, if we are not in online state we should not allow
+    # any requests. this is a safety net in case we missed adding a check on a specific
+    # state that the chief is waiting for.
+    if mlrun.mlconf.httpdb.state not in [mlrun.common.schemas.APIStates.online]:
+        enabled_endpoints = [
+            "healthz",
+            "background-tasks",
+            "client-spec",
+            "migrations",
+            "clusterization-spec",
+            "memory-reports",
+        ]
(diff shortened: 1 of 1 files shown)


server/api/api/deps.py
```diff
@@ -74,6 +74,23 @@ def verify_api_state(request: Request):
             )
             raise mlrun.errors.MLRunPreconditionFailedError(message)
 
+    # defensive approach, if we are not in online state we should not allow
+    # any requests. this is a safety net in case we missed adding a check on a specific
+    # state that the chief is waiting for.
+    if mlrun.mlconf.httpdb.state not in [mlrun.common.schemas.APIStates.online]:
+        enabled_endpoints = [
```
alonmr, line 81: this is duplicated, suggest to write it once somehow (class or getter?)

alonmr (approved):
Generally looks goos, though not sure what was the issue. The only thing the workers do when moving to online is init project member. was that the issue?

moranbental:
> Generally looks goos, though not sure what was the issue. The only thing the workers do when moving to online is init project member. was that the issue?

Yes, the issue was that init project member (https://github.com/mlrun/mlrun/blob/061da8476f9653c1507beaa86289ffcc4e6a9636/server/api/utils/singletons/project_member.py#L27) does not occur, which causes the error when trying to access `get_project_member().ensure_project`, as `get_project_member` is None

PR #6628, opened 2024-11-04
[Notifications] Add default notification params

roei3000b:
We want to add functionality for enriching default params for notifications by kind.
The basic logic of the `enrich_default_params` is only to enrich missing params from the default params to the notification params (and not overriding the existing params).

For example:
If we have `mail` notification and the end user didn’t give us `server_host` param, we should enrich that param from the default params.

https://iguazio.atlassian.net/browse/ML-8190


server/api/crud/workflows.py
```diff
@@ -341,6 +343,18 @@ def _prepare_run_object_for_single_run(
                 mlrun.model.Notification.from_dict(notification.dict())
                 for notification in workflow_request.notifications or []
             ]
+            default_notification_params = (
+                server.api.utils.notification_pusher.get_notifications_default_params()
+            )
+            for notification in notifications:
+                notification_type = NotificationTypes(
+                    notification.kind or NotificationTypes.console
```
liranbg: why are you defaulting to console? I am not sure why would the notification kind would be None?

server/api/crud/workflows.py
```diff
@@ -341,6 +343,18 @@ def _prepare_run_object_for_single_run(
...
+                ).get_notification()
+
+                default_params = default_notification_params.get(notification.kind, {})
+                notification.params = notification_type.enrich_default_params(
+                    notification.params, default_params
```
liranbg: why do we do the enrichment on the notification here ? the enrichment should happen on server side

mlrun/utils/notifications/notification_pusher.py
```diff
@@ -100,8 +100,13 @@ class NotificationPusher(_NotificationPusherBase):
...
     }
 
-    def __init__(self, runs: typing.Union[mlrun.lists.RunList, list]):
+    def __init__(
+        self,
+        runs: typing.Union[mlrun.lists.RunList, list],
+        default_notification_params: dict = None,
+    ):
         self._runs = runs
+        self._default_notification_params = default_notification_params or {}
```
liranbg:
```suggestion
        self._default_params = default_params or {}
```
I think it is redundant to name it x_notification_y as the class it notification anyway

mlrun/utils/notifications/notification/base.py
```diff
@@ -59,6 +69,13 @@ def load_notification(
     ) -> None:
         self.params = params or {}
 
+    @classmethod
+    def enrich_default_params(cls, params: dict, default_params: dict = None) -> dict:
+        default_params = default_params or {}
+        returned_params = default_params.copy()
+        returned_params.update(params)
+        return returned_params
+
```
liranbg: worth changing it to deepcopy (as the default might include nested values)

mlrun/utils/notifications/notification/base.py
```diff
@@ -24,9 +24,19 @@ def __init__(
         self,
         name: str = None,
         params: dict[str, str] = None,
+        default_params: dict[str, str] = None,
     ):
+        """
+        Initialize the notification object.
```
liranbg:
```suggestion
        NotificationBase is the base class for all notification types.
```

mlrun/utils/notifications/notification/base.py
```diff
@@ -24,9 +24,19 @@ def __init__(
...
+
+        :param name: The name of the notification object.
```
liranbg:
```suggestion
        :param name: The name of the notification.
```

mlrun/utils/notifications/notification/base.py
```diff
@@ -24,9 +24,19 @@ def __init__(
...
+        :param params: The parameters of the notification object.
```
liranbg:
```suggestion
        :param params: The parameters of the notification.
```

mlrun/utils/notifications/notification/base.py
```diff
@@ -24,9 +24,19 @@ def __init__(
...
+        :param default_params: The default parameters of the notification object.
+                                If one of the params is not provided, the default from this dict will be used.
```
liranbg:
```suggestion
        :param default_params: The default parameters of the notification. Used for server-side enrichment purposes.
```

liranbg (changes requested):
few suggestions

server/api/crud/workflows.py
roei3000b: Good question, I saw it also [here](https://github.com/mlrun/mlrun/blob/08344c885833faa0464afc4da34f1a9656716d96/mlrun/utils/notifications/notification_pusher.py#L222-L224) saw I assumed it is the way for getting the notification type.
Anyway, on this point I don't think that the kind can be None,  so maybe I will remove it.

server/api/crud/workflows.py
roei3000b: 1.  It is server side code
2. [Here](https://github.com/roei3000b/mlrun/blob/08344c885833faa0464afc4da34f1a9656716d96/server/api/crud/workflows.py#L342-L345) we create notification with `from_dict`, meaning that the `__init__` of the object isn't called, and because of that I need to manually call to the enrichment here.

mlrun/utils/notifications/notification/base.py
```diff
@@ -59,6 +69,13 @@ def load_notification(
     ) -> None:
         self.params = params or {}
 
+    @classmethod
+    def enrich_default_params(cls, params: dict, default_params: dict = None) -> dict:
+        default_params = default_params or {}
+        returned_params = deepcopy(default_params)
+        returned_params.update(params)
+        return returned_params
+
```
quaark, line 78: Since each notification class calls the `enrich_default_params`, I'm not sure `get_notifications_default_params` is needed.
Each class can have its own `default_params` method which implements the default params however it needs, and the base `enrich_default_params` can call the inherited `default_params` of each notification object.

quaark (changes requested):
Looks good, though I think the double step for getting the default params and then setting them can be combined into one

mlrun/utils/notifications/notification/base.py
roei3000b: You definitely right, but because for the current version the notifications are under the client code(`mlrun/mlrun/utils/notifications/notification`) and for mail notification for example we need some server code for the default params I choose to do it on the server side code.
After implementing [Make running notification send from the server side](https://iguazio.atlassian.net/browse/ML-8069) we can move the all notifications classes to the server side code and then I will also make the `default_params` inside the notification itself by the type of the notification.
For now, maybe I will add TODO about it. WDYT?

mlrun/utils/notifications/notification/base.py
```diff
@@ -59,6 +69,13 @@ def load_notification(
     ) -> None:
         self.params = params or {}
 
+    @classmethod
```
liranbg, line 72: why is it a classmethod and not a self?

mlrun/utils/notifications/notification/base.py
```diff
@@ -59,6 +69,13 @@ def load_notification(
...
+    def enrich_default_params(cls, params: dict, default_params: dict = None) -> dict:
```
liranbg, line 73:
```suggestion
    def _enrich(cls, params: dict, default_params: dict = None) -> dict:
```
do you call it externally at all?

liranbg (commented):
minor

mlrun/utils/notifications/notification/base.py
roei3000b: The answer for the both question is that I call it on [validate_notification_params](https://github.com/roei3000b/mlrun/blob/9ad20791019778b0322c929fc41c21fd933fc4a9/mlrun/model.py#L760)

So:

1. It should be class method because on [this line](https://github.com/roei3000b/mlrun/blob/9e0ca0db3d8db1cf50e6af16c988454d51c5d69d/mlrun/model.py#L756) we use the class object for enrich_default_params.
2. It should be public because we call it here.

mlrun/utils/notifications/notification/base.py
roei3000b: See above

mlrun/utils/notifications/notification/base.py
quaark: yes add a TODO that would be best

quaark (approved):
LGTM - just add the TODO please

PR #6633, opened 2024-11-04
[Alert Activation] Add partition creation and retention for alert_activation table

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-8133
Jira - https://iguazio.atlassian.net/browse/ML-8134

This PR introduces partition management for the `alert_activation` table, including automatic partition creation and retention. Partition handling now runs as a background job (`_start_periodic_partition_management`) that executes at the specified retention interval. It removes partitions older than the retention threshold and pre-creates partitions for the next `2 * retention_time`.

Retention time can be configured by setting `config.crud.alert_activations.retention_days` (in days). Minimum value is 1 week (7 days), maximum is year (`7 * 53` days).

Example of partition creation from the system with patched mlrun:
image


server/py/services/api/db/sqldb/db.py
```diff
@@ -5384,6 +5384,87 @@ def _enrich_notification(_notification):
...
+        # Query partitions older than cutoff_partition_name
+        partitions_to_drop = session.execute(
+            text("""
+            SELECT GROUP_CONCAT(PARTITION_NAME) AS partitions
+            FROM information_schema.PARTITIONS
+            WHERE TABLE_NAME = :table_name
+              AND PARTITION_NAME < :cutoff_partition_name
+        """),
+            {"table_name": table_name, "cutoff_partition_name": cutoff_partition_name},
+        ).scalar()
```
alonmr: can't this be done with sqlalchemy SDK instead of plaintext query?

server/py/services/api/db/sqldb/db.py
```diff
@@ -5384,6 +5384,87 @@ def _enrich_notification(_notification):
...
+
+        # Only proceed if there are partitions to drop
+        if partitions_to_drop:
+            drop_sql = f"ALTER TABLE {table_name} DROP PARTITION {partitions_to_drop}"
```
alonmr: pretty sure this is prune to sql injection. you should use parameters like you did in the select above

server/py/services/api/db/sqldb/db.py
rokatyy: @alonmr No, it's not supported

server/py/services/api/db/sqldb/db.py
rokatyy: @alonmr it's designed to be used internally with internal arguments

mlrun/common/schemas/alert.py
TomerShor, line 1: Now that the `PartitionInterval` has grown, it seems unrelated to alerts specifically. Maybe worth moving it to its own file. (We might also want to use this logic for other tables in the future)

mlrun/config.py
```diff
@@ -134,6 +134,9 @@
         "resources": {
             "delete_crd_resources_timeout": "5 minutes",
         },
+        "alert_activations": {
+            "retention_days": 14 * 7,  # days
```
TomerShor: redundant, you can say "14 weeks" if you want to explain, but `days` is already written in the key 
```suggestion
            "retention_days": 14 * 7, 
```

mlrun/config.py
```diff
@@ -1038,6 +1041,19 @@ def verify_security_context_enrichment_mode_is_allowed(self):
                 f"is not allowed for iguazio version: {igz_version} < 3.5.1"
             )
 
+    @staticmethod
+    def validate_alert_activation_retention():
+        if config.crud.alert_activations.retention_days < 7 and not os.getenv(
+            "PARTITION_INTERVAL"
+        ):
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "Partition interval must be greater than a week"
```
TomerShor:
```suggestion
                "Alert activation partition interval must be greater than a week"
```

mlrun/config.py
```diff
@@ -1038,6 +1041,19 @@ def verify_security_context_enrichment_mode_is_allowed(self):
...
+            )
+        elif config.crud.alert_activations.retention_days > 53 * 7:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "Partition interval must be less than a year"
```
TomerShor:
```suggestion
                "Alert activation partition interval must be less than a year"
```

server/py/services/api/crud/alert_activation.py
```diff
@@ -15,54 +15,117 @@
...
+
+    def create_partitions(
+        self,
+        session: Session,
+        partition_number: int,
+    ):
+        # Retrieve the partition function from the database
+        partition_expression, partition_interval = (
+            self.get_partition_expression_and_interval(session)
+        )
```
TomerShor: You already have this information from the caller function above (create_and_drop_partitions), you can reuse it and skip another db query

server/py/services/api/crud/alert_activation.py
```diff
@@ -15,54 +15,117 @@
...
-                partition_datetime.replace(day=1) + timedelta(days=32)
-            ).replace(day=1)
-            # Format as 'YYYYMM' (year and month)
-            partition_value = partition_boundary_date.strftime("%Y%m")
-        else:
-            raise ValueError(f"Unsupported partition interval: {partition_interval}")
-        partition_expression = f"{partition_interval}(activation_time)"
-
-        return partition_name, partition_value, partition_expression
+        _, partition_interval = self.get_partition_expression_and_interval(session)
```
TomerShor: Same here

server/py/services/api/crud/alert_activation.py
```diff
@@ -15,54 +15,117 @@
...
+        :param retention_days: The number of days to retain partitions.
         """
-        Generates partition details for a specified interval and datetime.
 
-        :param partition_interval: The partitioning interval type, e.g., "DAY", "MONTH", or "YEARWEEK".
-        :param partition_datetime: The datetime used for generating partition details.
+        # Retrieve the partition expression and interval from the database.
+        partition_expression, partition_interval = (
+            self.get_partition_expression_and_interval(session)
+        )
```
TomerShor: Actually I now see that `partition_expression` is not used in this function at all.
But anyway you can get both values here and pass them to the `create` and `drop` functions

server/py/services/api/db/sqldb/db.py
```diff
@@ -5384,6 +5384,90 @@ def _enrich_notification(_notification):
...
+        :param session: SQLAlchemy session for database connection.
+        :param table_name: Name of the table where partitions will be created.
+        :param partitioning_information_list: List of tuples, each containing:
+            - partition_name: The name for the partition.
+            - partition_value: The "LESS THAN" boundary value for the partition.
+            - partition_expression: The SQL partition expression.
+        """
+        alter_table_template = f"""
+            ALTER TABLE {table_name}
+            PARTITION BY RANGE ({partitioning_information_list[0][2]}) (
```
TomerShor: Why are you always using the same value here?

server/py/services/api/crud/alert_activation.py
TomerShor, line 1: I think all of the partition logic can also move to a separate file as it is not really related to the alert activation crud.
Maybe a file in the API utils would be more fitting, as again - we would maybe want to reuse this logic for something else in the future.

TomerShor (changes requested):
Looking nice!
Left some suggestions below

server/py/services/api/db/sqldb/db.py
```diff
@@ -5384,6 +5384,90 @@ def _enrich_notification(_notification):
...
+            FROM information_schema.PARTITIONS
+            WHERE TABLE_NAME = :table_name
+              AND PARTITION_NAME < :cutoff_partition_name
+        """),
+            {"table_name": table_name, "cutoff_partition_name": cutoff_partition_name},
+        ).scalar()
+
+        # Only proceed if there are partitions to drop
+        if partitions_to_drop:
+            drop_sql = f"ALTER TABLE {table_name} DROP PARTITION {partitions_to_drop}"
```
Yacouby: this query should be parametrized to prevent sql injection

server/py/services/api/db/sqldb/db.py
```diff
@@ -5384,6 +5384,90 @@ def _enrich_notification(_notification):
...
+        :param cutoff_partition_name: The cutoff partition name for dropping old partitions.
+        """
+        # Query partitions older than cutoff_partition_name
+        partitions_to_drop = session.execute(
+            text("""
+            SELECT GROUP_CONCAT(PARTITION_NAME) AS partitions
+            FROM information_schema.PARTITIONS
+            WHERE TABLE_NAME = :table_name
+              AND PARTITION_NAME < :cutoff_partition_name
+        """),
```
Yacouby: if the table has many partitions this query can be slow, you can add an index on the PARTITION_NAME column to speed up the query lookup.

server/py/services/api/tests/unit/crud/test_alert_activations.py
```diff
@@ -97,3 +102,153 @@ def test_get_partition_info_for_datetime(
...
+    "partition_interval, retention_days, test_date, expected_cutoff_name",
+    [
+        ("DAY", 4 * 7, datetime(2024, 1, 1), "p20231204"),
+        ("DAY", 1, datetime(2024, 1, 1), "p20231231"),
+        ("MONTH", 6 * 7, datetime(2024, 7, 15), "p202406"),
+        ("YEARWEEK", 12 * 7, datetime(2024, 6, 1), "p202410"),
+        ("YEARWEEK", 14 * 7, datetime(2024, 6, 1), "p202408"),
+    ],
+)
+def test_drop_old_partitions(
```
Yacouby: what about adding a test case when there are no partitions to drop ?

mlrun/common/schemas/alert.py
```diff
@@ -214,3 +214,77 @@ def is_valid(cls, value: str) -> bool:
...
+        """
+
+        partition_name = self.get_partition_name(partition_datetime)
+        partition_boundary_date = self.get_next_partition_time(partition_datetime)
+        partition_value = self.get_partition_name(partition_boundary_date)
+        partition_expression = self.get_partition_expression()
+
+        return partition_name, partition_value, partition_expression
+
+    def get_next_partition_time(self, current_datetime: datetime) -> datetime:
```
Yacouby: add a comment to explain this method's purpose and how the next partition time is calculated as it is confusing by only reading the implementation.

server/py/services/api/crud/alert_activation.py
```diff
@@ -15,54 +15,117 @@
...
+
+        for _ in range(partition_number):
+            partitioning_information_list.append(
+                partition_interval.get_partition_info(current_datetime)
+            )
+
+            # Move to the next interval based on the partition_interval
+            current_datetime = mlrun.common.schemas.alert.PartitionInterval(
+                partition_interval
+            ).get_next_partition_time(current_datetime=current_datetime)
```
Yacouby: can you optimize the partition creation process by precomputing the partition boundaries and avoiding multiple calls to get_next_partition_time within the loop ? this can reduce the overhead of repeated calls to get_next_partition_time and make the loop more efficient.

Yacouby (commented):
some suggestions for improvements

server/py/services/api/db/sqldb/db.py
rokatyy: @Yacouby since `information_schema.PARTITIONS` is a MySQL system table, you cannot modify its schema or add indexes to it directly. The query performance on information_schema tables can be slower when there are many partitions, but indexing or altering information_schema tables is not possible in MySQL.

server/py/services/api/db/sqldb/db.py
rokatyy: @Yacouby since `partitions_to_drop` is a list there is no a good way to do so, here all the values are internal, I think we can leave it as is

server/py/services/api/db/sqldb/db.py
rokatyy: @TomerShor because the structure of partitioning_information_list is always the same and second value is always a partitioning expression

server/py/services/api/tests/unit/crud/test_alert_activations.py
rokatyy: @Yacouby this test doesn't really test actual partition deletion - it tests cutoff date calculation as we use sqlite in tests not mysql

server/py/services/api/db/sqldb/db.py
rokatyy: I've reworked this logic a bit to avoid keeping the same value N times

server/py/services/api/main.py
```diff
@@ -642,6 +645,23 @@ def _start_periodic_project_summaries_calculation():
         )
 
 
+def _start_periodic_partition_management():
+    retention_days = config.crud.alert_activations.retention_days
+    logger.info(
+        "Starting periodic partition management for alert_activations",
```
TomerShor:
```suggestion
        "Starting periodic partition management for alert activations",
```

server/py/services/api/crud/alert_activation.py
```diff
@@ -13,56 +13,37 @@
...
-    @staticmethod
-    def get_partition_info(
-        partition_interval: str,
-        partition_datetime: datetime,
-    ) -> tuple[str, str, str]:
+    def create_and_drop_partitions(
+        self,
+        session: Session,
+        retention_days: int,
+    ) -> None:
```
TomerShor: I now think that this wrapper function is not really useful, and we can just call the Partitioner from `main` with the `alert_activations` table name.

TomerShor (changes requested):

server/py/services/api/crud/alert_activation.py
```diff
@@ -13,56 +13,13 @@
 # limitations under the License.
 #
 
-from datetime import datetime, timedelta
 
 import mlrun.common.schemas.alert
+import mlrun.common.schemas.partition
```
TomerShor: Not needed

TomerShor (commented):
Last small thing, otherwise approved 👌🏼
(+ CI should pass)

TomerShor (approved):
LGTM

mlrun/config.py
```diff
@@ -134,6 +134,9 @@
         "resources": {
             "delete_crd_resources_timeout": "5 minutes",
         },
+        "alert_activations": {
+            "retention_days": 14 * 7,
+        },
```
theSaarco: While we're dealing only with alert activations at this point, I'm seeing the partitioning mechanism as something we'd like to generalize later to additional tables. So, I'm thinking it's better to reverse the order here, i.e. something like:
```
"object_retentions": {
    "alert_activations": 14*7,
},
```
Then in the `validate` function, you can go over all values in the `object_retentions` section and perform the same validation for all of them.

mlrun/config.py
```diff
@@ -1038,6 +1041,19 @@ def verify_security_context_enrichment_mode_is_allowed(self):
                 f"is not allowed for iguazio version: {igz_version} < 3.5.1"
             )
 
+    @staticmethod
+    def validate_alert_activation_retention():
```
theSaarco: As mentioned in previous comment - I think this needs to be `validate_object_retentions`, and internally iterate over all members of that config (currently only alert activations).

server/py/services/api/db/sqldb/db.py
```diff
@@ -5656,6 +5656,106 @@ def _enrich_notification(_notification):
...
+        Execute the drop operation for partitions older than the cutoff partition name.
+
+        :param session: SQLAlchemy session.
+        :param table_name: The name of the table with partitions.
+        :param cutoff_partition_name: The cutoff partition name for dropping old partitions.
+        """
+        # Query partitions older than cutoff_partition_name
+        partitions_to_drop = session.execute(
+            text("""
+            SELECT GROUP_CONCAT(PARTITION_NAME) AS partitions
```
theSaarco: Using `GROUP_CONCAT` is problematic:
1. It's a MySQL thing. Postgres doesn't support it (it has other alternatives).
2. It's limited to 1024 chars by default (see [here](https://dev.mysql.com/doc/refman/8.4/en/aggregate-functions.html#function_group-concat)). Not a big concern for us since we can fit >100 partitions in 1024 chars, but still.

I suggest we don't use this here, but rather can just query the partition names and aggregate later in the Python code.

server/py/services/api/db/sqldb/db.py
```diff
@@ -5656,6 +5656,106 @@ def _enrich_notification(_notification):
...
+            FROM information_schema.PARTITIONS
+            WHERE TABLE_NAME = :table_name
+              AND PARTITION_NAME < :cutoff_partition_name
```
theSaarco, line 5728: `GROUP_CONCAT` is an aggregated function over groups, and it's best practice to add the relevant `GROUP_BY` clause here (would be `TABLE_NAME` in this case). MySQL knows how to deduce it if needed, but there actually is a setting in MySQL controlling whether it's required to always put the exact `GROUP_BY` clause or let the DB deduce it. 
We don't enforce it, but it may be used in some deployment scenarios, especially if working with a user-deployed DB.

server/py/services/api/db/sqldb/db.py
```diff
@@ -5656,6 +5656,106 @@ def _enrich_notification(_notification):
...
+
+        alter_table_template = f"""
+            ALTER TABLE {table_name}
+            ADD PARTITION (
+                {", ".join(new_partitions)}
+            );
+        """
+
+        session.execute(text(alter_table_template))
+        session.commit()
```
theSaarco: `commit` is redundant after DDL commands (https://dev.mysql.com/doc/refman/8.4/en/implicit-commit.html). While putting it here is good for readability, it has some unneeded overhead. You can replace it with a comment saying you don't need a commit 😄

server/py/services/api/db/sqldb/db.py
```diff
@@ -5656,6 +5656,106 @@ def _enrich_notification(_notification):
...
+              AND PARTITION_NAME < :cutoff_partition_name
+        """),
+            {"table_name": table_name, "cutoff_partition_name": cutoff_partition_name},
+        ).scalar()
+
+        # Only proceed if there are partitions to drop
+        if partitions_to_drop:
+            drop_sql = f"ALTER TABLE {table_name} DROP PARTITION {partitions_to_drop}"
+            session.execute(text(drop_sql))
+            session.commit()
```
theSaarco: Same commit comment.

server/py/services/api/main.py
```diff
@@ -642,6 +646,31 @@ def _start_periodic_project_summaries_calculation():
         )
 
 
+def _start_periodic_partition_management():
+    retention_days = config.crud.alert_activations.retention_days
```
theSaarco: Here again, I'd generalize this (the name of this function is already generalized to "partition management" which is great) - iterate over the config values available and call the `_manage_partitions` function for each of them (you'll probably need to pass the pass the table name to the function).
This means that the config keys will have to match the DB table names, but that should be fine.

theSaarco (changes requested):
Looks very good. I do have some comments, mostly these:
1. The mechanism should be generalized, as we'd like to apply it to more tables later. Either do it now or in subsequent PRs, but I don't think it's much more complicated to make it generic at this point.
2. Have some comments on the SQL statements used.

Yacouby (approved):

theSaarco (approved):
🔥 Nice work - LGTM.

liranbg (approved):

PR #6646, opened 2024-11-05
[Model Monitoring] Write res & metrics only to the tsdb & Update UI endpoints - /metrics to use tsdb connector

davesh0812:
https://iguazio.atlassian.net/browse/ML-8148

also resolving https://iguazio.atlassian.net/browse/ML-8187


mlrun/common/schemas/model_monitoring/model_endpoints.py
```diff
@@ -284,7 +284,14 @@ class ModelEndpointMonitoringMetric(BaseModel):
     app: str
     type: ModelEndpointMonitoringMetricType
     name: str
-    full_name: str
+    full_name: Optional[str] = None
+    kind: Optional[ResultKindApp] = None
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
Eyal-Danieli: Are you sure we need that *args? because the parent class `BaseModel` has no *args

mlrun/model_monitoring/db/stores/sqldb/sql_store.py
```diff
@@ -338,70 +318,12 @@ def list_model_endpoints(
...
-        return "_".join(
-            [
-                event[mm_schemas.WriterEvent.ENDPOINT_ID],
-                event[mm_schemas.WriterEvent.APPLICATION_NAME],
-                name,
-            ]
-        )
-
     @staticmethod
     def _get_filter_criteria(
```
Eyal-Danieli: I think you can remove this func as well

Eyal-Danieli (approved):

PR #6653, opened 2024-11-06
[Notifications] Add mail notification kind

roei3000b:
The mail notification params are:
**Server**
- server_host - network address of the SMTP server (for example email-smtp.us-east-1.amazonaws.com).
- server_port - network port to use. If not specified, the default SMTP port of 587 will be assumed.
- sender_address - email address of sender.
- use_tls - the TLS handshake occurs when the connection is established, and all traffic is encrypted. This type of connection should generally be used where available.
- start_tls - an initial unencrypted connection is made, EHLO/HELO greetings are exchanged, and the connection is upgraded in place once the client requests it by sending the STARTTLS command. Most servers require an upgrade before allowing AUTH commands.
- validate_certs - should validate certificate of the smtp server?
- username - The user to login to the SMTP server with.
- password - password for the username provided.

**Email-related parameters**
- email_address - the email address to send the email to - we should support multiple mail addresses.
- subject - an optional subject.
- message - an optional message to add to the description of the mail, like is done in other types of notifications.

 
MLRun should read the default params of smtp configuration from the secret (the secret name is in the mlconf under `notifications.smtp.config_secret_name`

We are using the `enrich_default_params` (https://github.com/mlrun/mlrun/pull/6628) function for enriching the server parameters.

https://iguazio.atlassian.net/browse/ML-8191


mlrun/common/schemas/notification.py
```diff
@@ -45,6 +45,13 @@ class NotificationKind(mlrun.common.types.StrEnum):
     slack: str = "slack"
     """**webhook** - The slack webhook to which to send the notification."""
 
+    mail: str = "mail"
+    """
+    **email_address** - The target mails\n
+    **subject** - The subject of the mail\n
+    **body** - The body of the mail\n
+    """
```
rokatyy:
```suggestion
    **email_addresses** - The target mails\n
    """
```

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,106 @@
...
+        email_address,
+        sender_address,
+        server_host,
+        server_port,
+        username,
+        password,
+        use_tls,
+        subject="",
+        body="",
+        **kwargs,
```
rokatyy: type annotations?

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,106 @@
...
+    ):
+        # Create the email message
+        message = EmailMessage()
+        message["From"] = sender_address
+        message["To"] = email_address
+        message["Subject"] = subject
+        message.set_content(body)
```
rokatyy: Does it mean that it supports only a single email address for now or list can be passed here?

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,106 @@
...
+    @staticmethod
+    async def _send_email(
+        email_address,
+        sender_address,
+        server_host,
+        server_port,
+        username,
+        password,
+        use_tls,
+        subject="",
```
rokatyy: An empty subject might be confusing if the user forgets to set it.. It may be worth adding an 'MLRun Notification' prefix as a default or simply not allowing it to be empty.

rokatyy (changes requested):
Looking good!

Having couple of questions about a support of multiple email addresses

mlrun/utils/notifications/notification/mail.py
roei3000b: The `To` should be comma separated string that contains the email addresses, for example:

1. Single address: `a@b.com`
2. Multiple addresses: `a@b.com, b@b.com, c@b.com`

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,117 @@
...
+        for required_param in cls.required_params:
+            if required_param not in params:
+                raise ValueError(
+                    f"Parameter '{required_param}' is required for MailNotification"
+                )
+
+        if type(params["email_addresses"]) not in [str, list]:
+            raise ValueError(
+                "Parameter 'email_addresses' must be a string or a list of strings"
+            )
```
rokatyy: By this code below it seems that `email_addresses` is always a string.

```
      if type(email_addresses) is list:
            email_addresses = ",".join(email_addresses)
        params["email_addresses"] = email_addresses
```

rokatyy (approved):
Great! 

Just one minor question regarding type check in `validate_params`. Otherwise approved 👍

mlrun/utils/notifications/notification/mail.py
roei3000b: The `validate_params` function is called before the `enrich_default_params`.
So first we want to validate that the `email_addresses` is either string or list, and ,later we want to enrich it to string.

server/py/services/api/utils/notification_pusher.py
```diff
@@ -198,5 +201,22 @@ def resolve_notifications_default_params():
...
         NotificationTypes.slack: {},
+        NotificationTypes.mail: _get_mail_notification_default_params(),
         NotificationTypes.webhook: {},
     }
+
+
+def _get_mail_notification_default_params():
+    global mail_notification_default_params
+    if mail_notification_default_params is not None:
+        return mail_notification_default_params
```
quaark: In the future this will be part of the `Notification` class and the `mail_notification_default_params` can live as a static member there.

Thinking if its worth making this function(and `resolve_notifications_default_params`) as part of the server side `NotificationPusher` class above, so this can still be a static member here instead of a global, which can often be confusing.

tests/utils/test_notifications.py
```diff
@@ -1012,3 +1013,99 @@ def _generate_run_result(
...
+    )
+    def test_validate_mail_params(self, params, should_raise):
+        try:
+            mlrun.utils.notifications.notification.MailNotification.validate_params(
+                params
+            )
+        except ValueError:
+            assert should_raise
+        else:
+            assert not should_raise
```
quaark: In a lot of mlrun tests we have the concept putting in the parametrize an `"expectation"`, that can have an empty context manager or a `pytest.raises(ValueError)`.

And then you can run:
```
with expectation:
    mlrun.utils.notifications.notification.MailNotification.validate_params(
        params
    )
```

quaark (changes requested):
Looks good!
A couple of suggestions

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,119 @@
...
+        message.set_content(body)
+
+        # Send the email
+        await aiosmtplib.send(
+            message,
+            hostname=server_host,
+            port=server_port,
+            username=username,
+            password=password,
+            use_tls=use_tls,
```
adamdelman, line 146: From the aiosmtplib docs:

TARTTLS connections
By default, if the server advertises STARTTLS support, aiosmtplib will upgrade the connection automatically. Setting use_tls=True for STARTTLS servers will typically result in a connection error.

So we probably need a way to let the user set start_tls as well.

mlrun/utils/notifications/notification/__init__.py
```diff
@@ -20,6 +20,7 @@
 from .console import ConsoleNotification
 from .git import GitNotification
 from .ipython import IPythonNotification
+from .mail import MailNotification
```
moranbental: use `import x` instead `from x import y`
https://github.com/mlrun/mlrun/blob/development/CONTRIBUTING.md#:~:text=When%20importing%20from%20local%20code%2C%20NEVER%20use%3A%20from%20X%20import%20Y%20instead%20use%3A%20import%20X%20or%20import%20X%20as%20Y.%20For%20external%20packages%2C%20it%27s%20acceptable%20to%20use%20from%20X%20import%20Y%20since%20they%20won%27t%20try%20to%20import%20back%20to%20our%20code.

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,119 @@
...
+import typing
+from email.message import EmailMessage
+
+import aiosmtplib
+
+import mlrun.common.schemas
+import mlrun.lists
+import mlrun.utils.helpers
+
+from .base import NotificationBase
```
moranbental: same here

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @classmethod
+    def validate_params(cls, params):
+        for required_param in cls.required_params:
+            if required_param not in params:
+                raise ValueError(
+                    f"Parameter '{required_param}' is required for MailNotification"
+                )
+
+        if type(params["email_addresses"]) not in [str, list]:
```
moranbental:
```suggestion
        if not isinstance(email_addresses, (str, list)):
```
You also need to validate that if it is a list, all the elements are strings

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,119 @@
...
+    async def push(
+        self,
+        message: str,
+        severity: typing.Union[
+            mlrun.common.schemas.NotificationSeverity, str
+        ] = mlrun.common.schemas.NotificationSeverity.INFO,
+        runs: typing.Union[mlrun.lists.RunList, list] = None,
+        custom_html: typing.Optional[str] = None,
+        alert: mlrun.common.schemas.AlertConfig = None,
+        event_data: mlrun.common.schemas.Event = None,
```
moranbental:
```suggestion
        runs: typing.Optional[typing.Union[mlrun.lists.RunList, list]] = None,
        custom_html: typing.Optional[str] = None,
        alert: typing.Optional[mlrun.common.schemas.AlertConfig] = None,
        event_data: typing.Optional[mlrun.common.schemas.Event] = None,
```

requirements.txt
```diff
@@ -42,3 +42,5 @@ orjson>=3.9.15, <4
 # mlrun pipeline adapters
 mlrun-pipelines-kfp-common~=0.2.1
 mlrun-pipelines-kfp-v1-8~=0.2.0
+# TODO: Remove it after moving notificaions to the server side(ML-8069)
+aiosmtplib~=3.0
```
moranbental: A line is missing at the end of the file

moranbental (changes requested):
Looks good, just few comments

mlrun/utils/notifications/notification/mail.py
roei3000b: I think we should set start_tls to `False` anyway, because I don't think that the user needs to know all these implementation details.
If we pass `use_tls` as `True` it should be encrupted. 
Another question, what do you think about setting the `validate_certs` to the value of `use_tls` that I got from the user? is it correct or maybe I should add another params for `validate_certs` that by default will be `True`.
@adamdelman, @quaark , @liranbg , @theSaarco

roei3000b: As we talked offline, we will give option to configure the `use_tls`, `start_tls` and `validate_certs`.
By default:
```
use_tls = True
start_tls = False
validate_certs = True
```

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,141 @@
...
+        email_addresses = params.get("email_addresses", default_mail_address)
+        if type(email_addresses) is list:
+            email_addresses = ",".join(email_addresses)
+        params["email_addresses"] = email_addresses
+
+        return params
+
+    @staticmethod
+    def _set_param_default_value(params, key, default_value):
+        if type(params.get(key, "")) is str:
```
adamdelman: I think it's slightly mot pythonic to do 
`isinstance(params.get(key, ""),str)`

adamdelman (changes requested):
Looks good to me, just one minor style fix

adamdelman (approved):

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,141 @@
...
+                raise ValueError(
+                    f"Parameter '{required_param}' is required for MailNotification"
+                )
+
+        if not isinstance(params["email_addresses"], (str, list)):
+            raise ValueError(
+                "Parameter 'email_addresses' must be a string or a list of strings"
+            )
+
+        if isinstance(params["email_addresses"], list):
```
moranbental: If `email_addresses` is a str, you should also validate that it follows a valid email regex format

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,141 @@
...
+
+        cls._set_param_default_value(params, "use_tls", "true")
+        cls._set_param_default_value(params, "start_tls", "false")
+        cls._set_param_default_value(params, "validate_certs", "true")
+
+        params.setdefault("server_port", DEFAULT_SMTP_PORT)
+
+        default_mail_address = params.pop("default_email_addresses", "")
+        email_addresses = params.get("email_addresses", default_mail_address)
+        if type(email_addresses) is list:
```
moranbental:
```suggestion
        if isinstance(email_addresses, list):
```

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,141 @@
...
+            email_addresses = ",".join(email_addresses)
+        params["email_addresses"] = email_addresses
+
+        return params
+
+    @staticmethod
+    def _set_param_default_value(params, key, default_value):
+        if isinstance(params.get(key, ""), str):
+            params[key] = json.loads(params.get(key, default_value))
```
moranbental: What happens if `json.loads()` fails?

mlrun/utils/notifications/notification_pusher.py
```diff
@@ -29,13 +29,14 @@
...
 import mlrun.utils.helpers
+import mlrun.utils.notifications.notification.base as base
 import mlrun_pipelines.common.ops
 import mlrun_pipelines.models
 import mlrun_pipelines.utils
 from mlrun.utils import logger
 from mlrun.utils.condition_evaluator import evaluate_condition_in_separate_process
 
-from .notification import NotificationBase, NotificationTypes
+from .notification import NotificationTypes
```
moranbental: use import x

server/py/framework/utils/notifications/notification_pusher.py
```diff
@@ -20,17 +20,55 @@
 import mlrun.errors
 import mlrun.model
 import mlrun.utils.helpers
+import mlrun.utils.notifications.notification.base as base
 from mlrun.utils import logger
-from mlrun.utils.notifications.notification import NotificationBase, NotificationTypes
+from mlrun.utils.notifications.notification import NotificationTypes
```
moranbental: use import x

moranbental (changes requested):
minor

server/py/framework/utils/singletons/k8s.py
```diff
@@ -629,12 +629,13 @@ def _update_secret(
         except k8s_client_rest.ApiException as exc:
             raise k8s_dynamic_exceptions.api_exception(exc)
 
-    def _read_secret(
+    def read_secret(
```
liranbg: move up

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,141 @@
...
+        severity: typing.Optional[
+            typing.Union[mlrun.common.schemas.NotificationSeverity, str]
+        ] = mlrun.common.schemas.NotificationSeverity.INFO,
+        runs: typing.Optional[typing.Union[mlrun.lists.RunList, list]] = None,
+        custom_html: typing.Optional[typing.Optional[str]] = None,
+        alert: typing.Optional[mlrun.common.schemas.AlertConfig] = None,
+        event_data: typing.Optional[mlrun.common.schemas.Event] = None,
+    ):
+        self.params.setdefault("subject", f"[{severity}] {message}")
+        self.params.setdefault("body", message)
```
liranbg, line 73: why does the subject includes the message?

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,141 @@
...
+        default_mail_address = params.pop("default_email_addresses", "")
+        email_addresses = params.get("email_addresses", default_mail_address)
+        if type(email_addresses) is list:
+            email_addresses = ",".join(email_addresses)
+        params["email_addresses"] = email_addresses
+
+        return params
+
+    @staticmethod
+    def _set_param_default_value(params, key, default_value):
```
quaark: should be part of notification base

quaark (changes requested):
looks really good, very minor comment

mlrun/utils/notifications/notification/mail.py
roei3000b: As the requirement requires:
```
The emails sent with this notification will have the following on the subject line:
[<severity] <message>
For example:
[info] Run completed: Function Completed
```

I will open another pr for the right body (https://iguazio.atlassian.net/browse/ML-8193)

mlrun/utils/notifications/notification/base.py
```diff
@@ -145,3 +146,13 @@ def _get_overview_type_and_url(
             overview_type = "Model endpoint"
 
         return overview_type, url
+
+    @staticmethod
+    def _load_json_from_str_param_with_default_json_value(
+        params, key, default_json_value
+    ):
+        if isinstance(params.get(key, ""), str):
```
moranbental: Can `params.get(key)` be a list or any other type besides a string?

moranbental (changes requested):
just one question

mlrun/utils/notifications/notification/base.py
roei3000b: Yes, it can be, for example see validate_certs here:
`mail_notification = mlrun.model.Notification(kind="mail", when=["completed", "error"], name="ENXuIbyqEx", message="JMTpbLAldTZTAeWatuIpWzVTwSKdSaDq", condition="", severity="verbose", params={"start_tls": "true", "use_tls": "false", "validate_certs": False, "email_addresses": ["roei3000b@gmail.com","hegeg56369@inikale.com"]})`

moranbental: @roei3000b So you need to handle that case because right now you're only handling str

moranbental (changes requested):

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,153 @@
...
+        cls, params: dict, default_params: typing.Optional[dict] = None
+    ) -> dict:
+        params = super().enrich_default_params(params, default_params)
+        params.setdefault("use_tls", True)
+        params.setdefault("start_tls", False)
+        params.setdefault("validate_certs", True)
+        params.setdefault("server_port", DEFAULT_SMTP_PORT)
+
+        default_mail_address = params.pop("default_email_addresses", "")
+        email_addresses = params.get("email_addresses", default_mail_address)
```
quaark, line 87: Not sure I understand, what is the `default_email_addresses` param, when would I use it and not `email_addresses`?

mlrun/utils/notifications/notification/mail.py
```diff
@@ -0,0 +1,153 @@
...
+            )
+
+        if not isinstance(params["email_addresses"], (str, list)):
+            raise ValueError(
+                "Parameter 'email_addresses' must be a string or a list of strings"
+            )
+
+        email_addresses = params["email_addresses"]
+        if isinstance(params["email_addresses"], str):
+            email_addresses = params["email_addresses"].split(",")
```
quaark: maybe create a small method that does this to eliminate the code duplication in the file

mlrun/utils/notifications/notification/mail.py
roei3000b: When it comes from the secret.

moranbental (approved):
LGTM

liranbg (approved):

PR #6654, opened 2024-11-06
[Pagination] Add paginated sdks for artifacts and models

moranbental:
Add new methods to the SDK (which involves adding SDK methods to `MlrunProject` and `HTTPRunDB`):
1. paginated_list_artifacts()
2. paginated_list_models()

Fix the BC test to run `test_save_openapi_json` inside the base code, as it was previously running inside the head code, which caused issues. The issue occurs because the CWD was set in such a way that pytest imported the HEAD code before the BASE tests, leading to incorrect imports and potentially causing test failures

https://iguazio.atlassian.net/browse/ML-8208


mlrun/db/httpdb.py
```diff
@@ -312,9 +312,26 @@ def paginated_api_call(
...
         version=None,
+        return_all=False,
     ) -> typing.Generator[requests.Response, None, None]:
         """
-        Calls the api with pagination, yielding each page of the response
+        Calls the API with pagination and yields each page of the response.
+
+        Depending on the `return_all` parameter:
+        - If `return_all` is `True`, fetches and yields all pages of results.
+        - If `return_all` is `False`, fetches and yields only the first page of results.
```
TomerShor: it's more "a single page" rather than "the first page", because you can get a specific page != 1

mlrun/db/httpdb.py
```diff
@@ -330,38 +347,50 @@ def _api_call(_params):
...
+
+        if page_params.get("page-token") is None and page_params.get("page") is None:
+            page_params["page"] = 1
+
+        if page_params.get("page-size") is None:
+            page_params["page-size"] = config.httpdb.pagination.default_page_size
 
+        response = _api_call(page_params)
+
+        # Yield the first page of results
```
TomerShor: same here

mlrun/db/httpdb.py
```diff
@@ -1144,6 +1175,162 @@ def list_artifacts(
...
+
+        This method retrieves a paginated list of artifacts based on the specified filter parameters.
+        Pagination is controlled using the `page`, `page_size`, and `page_token` parameters. The method
+        will return a list of artifacts that match the filtering criteria provided.
+
+        Examples::
+
+            # Show latest version of all artifacts in project
+            latest_artifacts, token = db.paginated_list_artifacts(
+                "", tag="latest", project="iris"
```
TomerShor: why this empty string?

mlrun/db/httpdb.py
```diff
@@ -1144,6 +1175,162 @@ def list_artifacts(
...
+            producer_uri=producer_uri,
+            format_=format_,
+            limit=limit,
+            page=page,
+            page_size=page_size,
+            page_token=page_token,
+            return_all=False,
+        )
+
+    def _list_or_paginated_artifacts(
```
TomerShor: I'd just call it `_list_artifacts`

mlrun/db/httpdb.py
```diff
@@ -1144,6 +1175,162 @@ def list_artifacts(
...
+        format_: Optional[
+            mlrun.common.formatters.ArtifactFormat
+        ] = mlrun.common.formatters.ArtifactFormat.full,
+        limit: int = None,
+        page: Optional[int] = None,
+        page_size: Optional[int] = None,
+        page_token: Optional[str] = None,
+        return_all: bool = False,
+    ) -> tuple[ArtifactList, Optional[str]]:
+        """Handles both list and paginated artifact fetching."""
```
```suggestion
        """Handles list artifacts, both paginated and not."""
```

mlrun/projects/project.py
```diff
@@ -3919,6 +4016,85 @@ def list_models(
...
+        return db.paginated_list_artifacts(
+            name,
+            self.metadata.name,
+            tag,
+            labels=labels,
+            since=since,
+            until=until,
+            iter=iter,
+            best_iteration=best_iteration,
+            kind="model",
```
```suggestion
            kind= mlrun.artifacts.model.ModelArtifact.kind,
```

TomerShor (changes requested):
Looking good, have some minor suggestions

mlrun/db/httpdb.py
```diff
@@ -1144,6 +1176,162 @@ def list_artifacts(
...
+        producer_uri: str = None,
+        format_: Optional[
+            mlrun.common.formatters.ArtifactFormat
+        ] = mlrun.common.formatters.ArtifactFormat.full,
+        limit: int = None,
+        page: Optional[int] = None,
+        page_size: Optional[int] = None,
+        page_token: Optional[str] = None,
+    ) -> tuple[ArtifactList, Optional[str]]:
+        """List artifacts with support for pagination and various filtering options.
```
liranbg, line 1206: instead of doc-dup list_artifacts paramters, you can just set a reference to list_Artifacts doc_string. the examples are ok to be left.

also, instead of accepting named arguments, Id perhaps have the function being declared as
```
def paginated_list_artifacts(self, *args, page=None, ...., **kwargs):
  pass
```

please add example of iterating the pages automatically without the need to explicitly state the page

liranbg (changes requested):

mlrun/db/httpdb.py
```diff
@@ -330,38 +347,50 @@ def _api_call(_params):
...
 
+            while page_token:
+                try:
+                    # Use the page token to get the next page.
+                    # No need to supply any other parameters as the token informs the pagination cache
+                    # which parameters to use.
+                    response = _api_call({"page-token": page_token})
+                except mlrun.errors.MLRunNotFoundError:
+                    # pagination token expired
+                    break
```
rokatyy: Do we want to inform a user in this case?

rokatyy (approved):
Looking really nice!

Left a minor question.

mlrun/db/httpdb.py
moranbental: @rokatyy No, it simply means that we've reached the last page (I updated the comment).

mlrun/db/httpdb.py
```diff
@@ -1144,6 +1176,111 @@ def list_artifacts(
...
+            # Automatically iterate over all pages without explicitly specifying the page number
+            artifacts = []
+            token = None
+            while True:
+                page_artifacts, token = db.paginated_list_artifacts(
+                    project="my-project", page_token=token, page_size=5
+                )
+                artifacts.extend(page_artifacts)
+                if not token:  # If no token is returned, we've reached the last page
+                    break
```
liranbg: no token is the end or no artifacts returned?

mlrun/projects/project.py
```diff
@@ -3858,6 +3859,61 @@ def list_artifacts(
...
+        :param page_size: The number of items per page to retrieve. Up to `page_size` responses are expected.
+        :param page_token: A pagination token used to retrieve the next page of results. Should not be provided
+            for the first request.
+
+        :returns: A tuple containing the list of artifacts and an optional `page_token` for pagination.
+        """
+        db = mlrun.db.get_run_db(secrets=self._secrets)
+        return db.paginated_list_artifacts(
+            *args, page=page, page_size=page_size, page_token=page_token, **kwargs
+        )
```
liranbg: are you suppose to pass the project name?

liranbg (commented):
left two questions to address

mlrun/projects/project.py
moranbental: yes thanks!

mlrun/db/httpdb.py
moranbental: @liranbg actually both, I updated the comment

mlrun/db/httpdb.py
```diff
@@ -1144,6 +1176,114 @@ def list_artifacts(
...
+        return self._list_artifacts(
+            *args,
+            page=page,
+            page_size=page_size,
+            page_token=page_token,
+            return_all=False,
+            **kwargs,
+        )
+
+    def _list_artifacts(
```
liranbg: move to bottom (private)

liranbg (commented):
minor

liranbg (approved):

PR #6655, opened 2024-11-07
[Microservices] Create service base class

alonmr:
Add a service and daemon base classes
The `Service` defines the interface and common code between microservices
The `Daemon` class initializes the service and shall configure it in the future

This is part 1 of https://iguazio.atlassian.net/browse/ML-8211


server/py/framework/service/__init__.py
```diff
@@ -0,0 +1,192 @@
...
+            openapi_url=f"{self.SERVICE_PREFIX}/openapi.json",
+            docs_url=f"{self.SERVICE_PREFIX}/docs",
+            redoc_url=f"{self.SERVICE_PREFIX}/redoc",
+            default_response_class=fastapi.responses.ORJSONResponse,
+            lifespan=self.lifespan,
+        )
+
+    # https://fastapi.tiangolo.com/advanced/events/
+    @contextlib.asynccontextmanager
+    async def lifespan(self, app_: fastapi.FastAPI):
```
adamdelman, line 76: Maybe add a return type annotation?

alonmr: wdym? adding `-> None` ?

server/py/services/api/daemon.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, service_cls: framework.service.Service.__class__):
+        self._service = service_cls()
+
+    def initialize(self):
+        self._service.initialize()
+
+
+daemon = Daemon(service_cls=services.api.main.Service)
+daemon.initialize()
+app = daemon.app
```
liranbg, line 32:
```suggestion
if __name__ == "__main__":
  daemon = Daemon(service_cls=services.api.main.Service)
  daemon.initialize()
  app = daemon.app
```

alonmr: This doesn't work with our current uvircorn command. May be able to get it to work with the new fastapi CLI but it requires some investigation - can be done later IMO

liranbg (approved):

server/py/services/api/daemon.py
```diff
@@ -0,0 +1,32 @@
...
+class Daemon(framework.service.Daemon):
+    def __init__(self, service_cls: framework.service.Service.__class__):
+        self._service = service_cls()
+
+    def initialize(self):
+        self._service.initialize()
+
+    @property
+    def service(self) -> services.api.main.Service:
+        return self._service
```
TomerShor, line 27: @alonmr Post-merge question: Why duplicate the same methods from the `framework.service.Daemon` class? I know it is an `ABC` but it seems redundant...

PR #6659, opened 2024-11-07
[Tests] Explicitly declare common fixtures in server tests

alonmr:
This makes `config_test_base` fixture run even when running tests from pycharm


server/py/services/api/tests/unit/conftest.py
```diff
@@ -60,6 +61,14 @@
 tests_root_directory = pathlib.Path(__file__).absolute().parent
 assets_path = tests_root_directory.joinpath("assets")
 
+if str(tests_root_directory) in os.getcwd():
+    # If this is the top level conftest - we need to explicitly declare the base common fixtures to
+    # make pytest use them. If this is not the top level conftest (e.g. when running the tests from the project root)
+    # then providing pytest_plugins is not allowed.
```
TomerShor, line 67: Is this really the top-level?
I'd expect it to be vice-versa - top-level is from the root folder

alonmr: top-level is relative to cwd

PR #6663, opened 2024-11-07
[Typing] Fix type hints with `datetime` module instead of the class

jond01:
Cherry-picked from #6607:
https://github.com/mlrun/mlrun/pull/6607/commits/582256d5d249a402e6e66869cb68538d7b3aec74


jond01:
Entered as a part of #6607.

PR #6674, opened 2024-11-10
[Alert Activation] Save alert activation to the DB for auto reset policy

Yacouby:
This PR implements the functionality to save alert activations to the alert_activation table for alerts with an auto-reset policy.
For alerts with manual reset policy the functionality will be added in an upcoming PR.

resolves :
https://iguazio.atlassian.net/browse/ML-8135


mlrun/common/schemas/alert.py
```diff
@@ -202,6 +202,24 @@ def templates_differ(self, other):
         )
 
 
+class NotificationState(pydantic.BaseModel):
+    kind: str
+    status: Optional[str]
```
TomerShor: This should be in `common/schemas/notification.py`

server/py/services/api/crud/alert_activation.py
```diff
@@ -15,13 +15,71 @@
...
+        services.api.utils.singletons.db.get_db().store_alert_activation(
+            session, alert_data, event_data, notifications_states
+        )
+
+    def list_alerts_activations(
+        self,
+        session: sqlalchemy.orm.Session,
+        project: str = "",
+    ) -> list[mlrun.common.schemas.AlertActivation]:
+        # add filters later
```
TomerShor:
```suggestion
        # TODO: add filters
```

server/py/services/api/crud/alert_activation.py
```diff
@@ -15,13 +15,71 @@
...
+            if notification_kind not in notification_states:
+                notification_states[notification_kind] = (
+                    "error"
+                    if notification_status
+                    == mlrun.common.schemas.NotificationStatus.ERROR
+                    else ""
+                )
+            else:
+                if notification_status == mlrun.common.schemas.NotificationStatus.ERROR:
+                    notification_states[notification_kind] = "error"
```
TomerShor: The same thing but a bit more efficient:
```suggestion
            notification_states.setdefault(
                notification_kind,
                notification_status if notification_status == mlrun.common.schemas.NotificationStatus.ERROR else ""
            )
```

server/py/services/api/db/sqldb/db.py
```diff
@@ -5772,6 +5773,83 @@ def _delete_alert_notifications(
...
+        # For JOB entities, construct entity_id as "name.uid" format
+        if alert_data.entities.kind == mlrun.common.schemas.alert.EventEntityKind.JOB:
+            run_name = alert_data.entities.ids[0]
+            run_uid = event_data.value_dict.get("uid")
+            entity_id = f"{run_name}.{run_uid}"
+        else:
+            entity_id = alert_data.entities.ids[0]
+
+        alert_activation_record = AlertActivation(
+            id=alert_data.id,
```
TomerShor: The `id` column is autogenerated by the DB when you insert a record, and the value is automatically increased by 1 on each insertion - it should not be manually set.
Furthermore, setting the activation id to the alert id is wrong, because they are 2 different things. A single alert can have multiple activations, and each activation will be a different record with a unique id.
```suggestion
```

server/py/services/api/db/sqldb/db.py
```diff
@@ -5772,6 +5773,83 @@ def _delete_alert_notifications(
...
+            number_of_events=alert_data.criteria.count,
+            data=extra_data,
+        )
+
+        self._upsert(session, [alert_activation_record])
+
+    def list_alerts_activations(
+        self, session: Session, project: str = None
+    ) -> list[mlrun.common.schemas.AlertActivation]:
+        # add filters later
```
TomerShor:
```suggestion
        # TODO: add filters
```

server/py/services/api/db/sqldb/db.py
```diff
@@ -5772,6 +5773,83 @@ def _delete_alert_notifications(
...
+
+    @staticmethod
+    def _transform_alert_activation_record_to_scheme(
+        alert_activation_record: AlertActivation,
+    ) -> mlrun.common.schemas.AlertActivation:
+        if alert_activation_record is None:
+            return None
+
+        alert_activation = mlrun.common.schemas.AlertActivation(
+            id=alert_activation_record.id,
```
TomerShor: In addition to my comment above, we do not usually return the `id` when listing/reading resources, as this `id` is only relevant for the DB, and does not have any relevance to mlrun (as opposed to UIDs for example)
```suggestion
```

server/py/services/api/tests/unit/crud/test_alert_activations.py
```diff
@@ -97,3 +102,150 @@ def test_get_partition_info_for_datetime(
...
+        services.api.crud.Events().process_event,
+        session=db,
+        event_data=event_data,
+        event_name=event_data.kind,
+        project=project,
+    )
+
+    # since the reset policy of the alert if auto, then now it should be reset and added to the alert
+    # activation table
+    alerts = services.api.crud.AlertActivation().list_alerts_activations(
```
TomerShor:
```suggestion
    activations = services.api.crud.AlertActivation().list_alerts_activations(
```

server/py/services/api/tests/unit/crud/test_alert_activations.py
```diff
@@ -97,3 +102,150 @@ def test_get_partition_info_for_datetime(
...
+    assert alerts[0].project == project
+    assert (
+        alerts[0].entity_id
+        == f"{alert_data.entities.ids[0]}.{event_data.value_dict.get('uid')}"
+    )
+    assert alerts[0].entity_kind == alert_entity.kind
+    # since criteria is not defined the default count number is 1
+    assert alerts[0].number_of_events == 1
+    assert alerts[0].activation_time == event_data.timestamp
+    assert alerts[0].notifications == expected_notifications_states
```
TomerShor: move to an helper function `_assert_alert_activation`

TomerShor (changes requested):
I know it's a draft but got a head start with some comments and suggestions.

Also a question - do we need any validations on the data before storing the activation record? Not sure because its only stored from within the API...

server/py/services/api/db/sqldb/db.py
Yacouby: you are right, this was some leftover that was going to be deleted 👍

mlrun/common/schemas/alert.py
rokatyy: In the requirement we approved another scheme:
```
NotificationState:
    kind: str
    err: Optional[str] #empty error means that notification was sent successfully
```

It is already used by FE to mock data and please also notice that the idea was to show the actual error, not just boolean value.

mlrun/common/schemas/alert.py
```diff
@@ -202,6 +202,24 @@ def templates_differ(self, other):
...
+
+
+class AlertActivation(pydantic.BaseModel):
+    name: str
+    project: str
+    severity: AlertSeverity
+    activation_time: datetime
+    entity_id: str
+    entity_kind: EventEntityKind
+    criteria: Optional[AlertCriteria]
```
rokatyy: Why is it optional? What if value is nil?

server/py/services/api/crud/alert_activation.py
```diff
@@ -15,13 +15,72 @@
...
+            notification_kind = alert_notification.notification.kind
+            notification_status = alert_notification.notification.status
+
+            if notification_kind not in notification_states:
+                notification_states[notification_kind] = (
+                    "error"
+                    if notification_status
+                    == mlrun.common.schemas.NotificationStatus.ERROR
+                    else ""
+                )
```
rokatyy: The same comment here, we should support an actual error, not just a state

server/py/services/api/crud/alerts.py
```diff
@@ -215,6 +215,9 @@ def process_event(
                 AlertNotificationPusher().push(alert, event_data)
 
                 if alert.reset_policy == "auto":
+                    services.api.crud.AlertActivation().store_alert_activation(
+                        session, alert, event_data
+                    )
                     self.reset_alert(session, alert.project, alert.name)
```
rokatyy: The idea was to write a db record on alert deactivation, so I think even here it's important to follow this order 
```suggestion
                    self.reset_alert(session, alert.project, alert.name)
                    services.api.crud.AlertActivation().store_alert_activation(
                        session, alert, event_data
                    )
```

server/py/services/api/db/sqldb/db.py
```diff
@@ -5756,6 +5757,83 @@ def _delete_alert_notifications(
...
+        alert_activation_record = AlertActivation(
+            id=alert_data.id,
+            name=alert_data.name,
+            project=alert_data.project,
+            activation_time=event_data.timestamp,
+            entity_id=entity_id,
+            entity_kind=alert_data.entities.kind,
+            event_kind=event_data.kind,
+            severity=alert_data.severity,
+            number_of_events=alert_data.criteria.count,
```
rokatyy: What if criteria is nil in this case (if it is possible as scheme says that it is)?

server/py/services/api/db/sqldb/db.py
```diff
@@ -5756,6 +5757,83 @@ def _delete_alert_notifications(
...
+            return None
+
+        alert_activation = mlrun.common.schemas.AlertActivation(
+            id=alert_activation_record.id,
+            name=alert_activation_record.name,
+            project=alert_activation_record.project,
+            severity=alert_activation_record.severity,
+            activation_time=alert_activation_record.activation_time.replace(
+                tzinfo=timezone.utc
+            ),
```
rokatyy: Please notice that the `activation_time` is of `datetime` format in db, which means that it should be already in UTC (DATETIME is always stored and returned as UTC)

rokatyy (changes requested):
Some initial comments :)

server/py/services/api/db/sqldb/db.py
Yacouby: in our case the id column is not autogenerated by the DB, since we have two primary keys in the table ( id & activation_time) and mysql does not support autoincrement for composite primary keys

mlrun/common/schemas/alert.py
Yacouby: in a second though this field can never be None, since if you try to store an alert without a criteria then we assign it a default one, so the Optional can be removed.

server/py/services/api/db/sqldb/db.py
Yacouby: the datetime format has no associated timezone.
so when i run the basic test of the alert activations it fails on : 
```
Expected :datetime.datetime(2024, 11, 11, 16, 39, 12, 309820, tzinfo=datetime.timezone.utc)
Actual   :datetime.datetime(2024, 11, 11, 16, 39, 12, 309820)
```

you can see here that the actual time has no time zone info

rokatyy: @Yacouby tests should not require code changes. Is there another place in the actual code where this change is needed?

Yacouby: the Actual datetime lacks timezone information, so it is treated as naive, meaning that it could represent any time, and there is no guarantee it is in UTC

rokatyy: @Yacouby then add comment that it's already in UTC and we just set it explicitly to avoid confusion

server/py/framework/utils/notifications/notification_pusher.py
```diff
@@ -172,11 +173,21 @@ def _update_notification_status(
...
         notification.sent_time = sent_time or notification.sent_time
 
+        # fill reason only if failed
+        if notification.status == mlrun.common.schemas.NotificationStatus.ERROR:
+            notification.reason = reason or notification.reason
+
+            # limit reason to a max of 255 characters (for db reasons) but also for human readability reasons.
+            notification.reason = notification.reason[:255]
+        else:
+            notification.reason = None
```
rokatyy, line 239: Can reason be not None if status is not error?

server/py/framework/db/base.py
```diff
@@ -956,6 +956,21 @@ def store_alert_notifications(
...
+    def store_alert_activation(
+        self,
+        session,
+        alert_data: mlrun.common.schemas.AlertConfig,
+        event_data: mlrun.common.schemas.Event,
+        notifications_states: list[mlrun.common.schemas.NotificationState],
+    ):
+        pass
+
+    def list_alerts_activations(
```
rokatyy:
```suggestion
    def list_alert_activations(
```

server/py/framework/db/sqldb/db.py
```diff
@@ -5875,6 +5876,85 @@ def _delete_alert_notifications(
...
+            entity_kind=alert_data.entities.kind,
+            event_kind=event_data.kind,
+            severity=alert_data.severity,
+            number_of_events=alert_data.criteria.count,
+            data=extra_data,
+        )
+
+        self._upsert(session, [alert_activation_record])
+
+    def list_alerts_activations(
```
```suggestion
    def list_alert_activations(
```

mlrun/common/schemas/alert.py
```diff
@@ -200,3 +200,30 @@ def templates_differ(self, other):
...
+    MONTH = "MONTH"
+    YEARWEEK = "YEARWEEK"
+
+    @classmethod
+    def is_valid(cls, value: str) -> bool:
+        return value in cls._value2member_map_
+
+    @classmethod
+    def valid_intervals(cls) -> list:
+        return list(cls._value2member_map_.keys())
```
rokatyy: Duplicate of https://github.com/mlrun/mlrun/blob/20a496cee30ef173dde1518987b8339b15d39ed2/mlrun/common/schemas/partition.py#L20

server/py/services/api/crud/alert_activation.py
```diff
@@ -13,10 +13,62 @@
...
+        # be stored. if no failure occurred, the status will be an empty string indicating success.
+        notification_states = {}
+
+        for alert_notification in alert.notifications:
+            notification_kind = alert_notification.notification.kind
+            notification_reason = alert_notification.notification.reason
+
+            notification_states.setdefault(
+                notification_kind,
+                notification_reason if notification_reason is not None else "",
```
rokatyy: `err` can be None, so no need to set it explicitly to `""`

server/py/services/api/crud/alert_activation.py
```diff
@@ -13,10 +13,62 @@
...
+            )
+
+        notification_states = [
+            mlrun.common.schemas.NotificationState(kind=kind, err=err)
+            for kind, err in notification_states.items()
+        ]
+
+        return notification_states
```
rokatyy:
```suggestion
    notification_states =   [
        mlrun.common.schemas.NotificationState(
            kind=alert_notification.notification.kind,
            err=alert_notification.notification.reason,
        )
        for alert_notification in alert.notifications
    ]
    return notification_states
```

server/py/services/api/crud/alert_activation.py
```diff
@@ -13,10 +13,62 @@
...
+    def list_alerts_activations(
+        self,
+        session: sqlalchemy.orm.Session,
+        project: str = "",
+    ) -> list[mlrun.common.schemas.AlertActivation]:
+        # TODO: add filters
+        project = project or mlrun.mlconf.default_project
+        return framework.utils.singletons.db.get_db().list_alerts_activations(
+            session, project
+        )
```
rokatyy: We don't really need this now (as unit tests aren't applicable), so I think we can leave it for next PR

server/py/services/api/tests/unit/crud/test_alerts.py
```diff
@@ -315,38 +313,3 @@ async def test_alert_reset_with_fields_updates(
...
-        name=name,
-        description=description,
-        summary=summary,
-        severity=severity,
-        entities=entity,
-        trigger=trigger,
-        criteria=criteria,
-        notifications=notifications,
-        reset_policy=reset_policy,
-    )
```
rokatyy: Do we still need to move this?

rokatyy (changes requested):
Looking good!

Added some questions and suggestions

server/py/services/api/tests/unit/crud/test_alerts.py
Yacouby: yes, we can use it in our upcoming work when adding alert activations tests to crud ( that mocks the db )

server/py/services/api/crud/alert_activation.py
```diff
@@ -13,10 +13,42 @@
...
+        # if at least one notification of that type failed, the error message from the most recent failure will
+        # be stored. if no failure occurred, the status will be an empty string indicating success.
+        notification_states = [
+            mlrun.common.schemas.NotificationState(
+                kind=alert_notification.notification.kind,
+                err=alert_notification.notification.reason,
+            )
+            for alert_notification in alert.notifications
+        ]
+        return notification_states
```
rokatyy: The problem is that we won't have updated notification from db in the alert object at this moment. Furthermore, in the case of auto-reset policy, it would be good to be able to avoid any additional db calls.

server/py/services/api/crud/alert_activation.py
Yacouby: this way for each notification we will have a different entry in the list, which is not aligned with the requirements that says we should have only one entry for each notification type..

rokatyy (changes requested):
It seems the current implementation of propagation notification reason isn't working. To make it work we will need to request notification again from db which is unnecessary for auto-reset policy. 

What we can do is to improve this logic https://github.com/mlrun/mlrun/blob/6e22aa8c6b831b3d231ccf0bb5f2ca2658eca8d6/server/py/framework/utils/notifications/notification_pusher.py#L97-L104 to make it gather the information after we have all the errors/exceptions in hands, not a single one. This way, we can determine if all notification attempts failed or just some. Then, we could log only the first error (logging all errors could be excessive). For this, we can modify `_push_notification_async` to return a `NotificationState`, which would allow us to aggregate the results into a new `NotificationState` with unique types.

Also, what really bothers me is the fact that we don't pass any session object here:
https://github.com/mlrun/mlrun/blob/6e22aa8c6b831b3d231ccf0bb5f2ca2658eca8d6/server/py/framework/utils/notifications/notification_pusher.py#L192-L193

Also I'm not sure that this - https://github.com/mlrun/mlrun/blob/6e22aa8c6b831b3d231ccf0bb5f2ca2658eca8d6/server/py/framework/utils/notifications/notification_pusher.py#L130-L137 should be a part of try-except flow for sending a notification as if we fail on updating notification in db, we will go to except part and say `"Failed to send notification"` which is not true.  So we can move the part from `try` below (after all except logic).

server/py/services/api/crud/alert_activation.py
Yacouby: why we dont have the updated notifications in the alert object ?
when updating the notification status in the notification_pusher, we iterate on the notifications of the alert.. the status and the reason are saved there, and when storing the alert activation we have these info.
so no other modifications should be done in the notification pusher side..
in my opinion it should not be that excessive to iterate on the notifications and check whether some of them failed to sent or not.

rokatyy: Agree, I didn't notice that it's a part of alert, all good then 👍

server/py/services/api/crud/alert_activation.py
TomerShor: What about this ^, why did you eventually change it to the full list instead of unique item per kind?

Yacouby: this code is not relevant anymore

server/py/framework/db/sqldb/db.py
```diff
@@ -5887,6 +5888,83 @@ def _delete_alert_notifications(
...
+
+        self._upsert(session, [alert_activation_record])
+
+    def list_alert_activations(
+        self, session: Session, project: typing.Optional[str] = None
+    ) -> list[mlrun.common.schemas.AlertActivation]:
+        # TODO: add filters
+        query = self._query(session, AlertActivation)
+
+        if project and project != "*":
```
moranbental, line 5933: This approach (of using project == '*') was changed in this PR: https://github.com/mlrun/mlrun/pull/6683. Please apply the same change here

mlrun/common/schemas/alert.py
```diff
@@ -17,7 +17,7 @@
 
 import pydantic
 
-from mlrun.common.schemas.notification import Notification
+from mlrun.common.schemas.notification import Notification, NotificationState
```
moranbental: use import x instead from x import y
https://github.com/mlrun/mlrun/blob/development/CONTRIBUTING.md#:~:text=When%20importing%20from%20local%20code%2C%20NEVER%20use%3A%20from%20X%20import%20Y%20instead%20use%3A%20import%20X%20or%20import%20X%20as%20Y.%20For%20external%20packages%2C%20it%27s%20acceptable%20to%20use%20from%20X%20import%20Y%20since%20they%20won%27t%20try%20to%20import%20back%20to%20our%20code.

server/py/framework/db/sqldb/db.py
```diff
@@ -5887,6 +5888,83 @@ def _delete_alert_notifications(
...
+                notification.dict() for notification in notifications_states
+            ],
+            "criteria": alert_data.criteria.dict() if alert_data.criteria else None,
+        }
+
+        # For JOB entities, construct entity_id as "name.uid" format
+        if alert_data.entities.kind == mlrun.common.schemas.alert.EventEntityKind.JOB:
+            run_name = alert_data.entities.ids[0]
+            run_uid = event_data.value_dict.get("uid")
+            entity_id = f"{run_name}.{run_uid}"
```
TomerShor: Not sure if it is even possible to not have a `uid` in the event for a JOB failure, but better be defensive and not end up with random periods.
```suggestion
            run_uid = event_data.value_dict.get("uid")
            entity_id = f"{run_name}.{run_uid}" if run_uid else run_name
```

server/py/framework/db/sqldb/db.py
```diff
@@ -5887,6 +5888,83 @@ def _delete_alert_notifications(
...
+        )
+
+        return alert_activations
+
+    @staticmethod
+    def _transform_alert_activation_record_to_scheme(
+        alert_activation_record: AlertActivation,
+    ) -> mlrun.common.schemas.AlertActivation:
+        if alert_activation_record is None:
+            return None
```
TomerShor: if `alert_activation_record` is allowed to be None, it should be reflected in the type hints with `Optional`.

server/py/services/api/crud/alert_activation.py
TomerShor, line 1: Now that we have an alert service directory - why is this file in the still in the api service directory?..

server/py/framework/db/sqldb/db.py
```diff
@@ -5887,6 +5888,83 @@ def _delete_alert_notifications(
...
+        alert_activations = list(
+            map(self._transform_alert_activation_record_to_scheme, query.all())
+        )
+
+        return alert_activations
+
+    @staticmethod
+    def _transform_alert_activation_record_to_scheme(
+        alert_activation_record: AlertActivation,
+    ) -> mlrun.common.schemas.AlertActivation:
```
```suggestion
        alert_activation_record: Optional[AlertActivation]
    ) -> Optional[mlrun.common.schemas.AlertActivation]:
```

server/py/framework/db/sqldb/db.py
```diff
@@ -5887,6 +5888,83 @@ def _delete_alert_notifications(
...
+            ),
+            entity_id=alert_activation_record.entity_id,
+            entity_kind=alert_activation_record.entity_kind,
+            event_kind=alert_activation_record.event_kind,
+            number_of_events=alert_activation_record.number_of_events,
+            notifications=alert_activation_record.data.get("notifications", []),
+            criteria=alert_activation_record.data.get("criteria"),
+        )
+
+        return alert_activation
```
```suggestion
        return mlrun.common.schemas.AlertActivation(
            name=alert_activation_record.name,
            project=alert_activation_record.project,
            severity=alert_activation_record.severity,
            # the activation_time is already stored in UTC in the database as a naive datetime.
            # we explicitly set the timezone to UTC here to make it timezone-aware, avoiding any ambiguity.
            activation_time=alert_activation_record.activation_time.replace(
                tzinfo=timezone.utc
            ),
            entity_id=alert_activation_record.entity_id,
            entity_kind=alert_activation_record.entity_kind,
            event_kind=alert_activation_record.event_kind,
            number_of_events=alert_activation_record.number_of_events,
            notifications=alert_activation_record.data.get("notifications", []),
            criteria=alert_activation_record.data.get("criteria"),
        )
```

server/py/services/api/crud/alert_activation.py
```diff
@@ -13,10 +13,85 @@
...
+        for alert_notification in notifications:
+            kind = alert_notification.notification.kind
+            reason = alert_notification.notification.reason
+
+            # initialize an entry for each kind if it doesn't exist
+            if kind not in notification_errors:
+                notification_errors[kind] = {
+                    "errors": set(),
+                    "success_count": 0,
+                }
```
TomerShor: you can use a `defaultdict`here:
```suggestion
        notification_errors = collections.defaultdict(
            lambda: {
                "errors": set(),
                "success_count": 0,
            },
        )

        # process each notification and gather errors by type
        for alert_notification in notifications:
            kind = alert_notification.notification.kind
            reason = alert_notification.notification.reason
```

server/py/services/api/crud/alert_activation.py
```diff
@@ -13,10 +13,85 @@
...
+                if success_count == 0:
+                    error_message = (
+                        f"All {kind} notifications failed. Errors: {', '.join(errors)}"
+                    )
+                else:
+                    error_message = (
+                        f"Some {kind} notifications failed. Errors: {', '.join(errors)}"
+                    )
+            else:
+                error_message = ""  # indicates success if there are no errors
```
TomerShor: Styling convention not to use inline comments
```suggestion
            else:
                # indicates success if there are no errors
                error_message = ""
```

server/py/framework/db/sqldb/db.py
```diff
@@ -5887,6 +5888,83 @@ def _delete_alert_notifications(
...
+        query = self._query(session, AlertActivation)
+
+        if project and project != "*":
+            query = query.filter(AlertActivation.project == project)
+
+        alert_activations = list(
+            map(self._transform_alert_activation_record_to_scheme, query.all())
+        )
+
+        return alert_activations
```
moranbental: This is more readable and pythonic than using map
```suggestion
        return [
            self._transform_alert_activation_record_to_scheme(record)
            for record in query.all()
        ]
```

TomerShor (changes requested):
Looking nice!
Let some suggestions and questions.
Also - I am missing a test that the activation record was indeed saved when an alert was reset. You can also add this check to an existing alert test.

moranbental (changes requested):
Looks good! left some comments

server/py/services/api/crud/alert_activation.py
Yacouby: since also @rokatyy is working on that, it was decided with @alonmr  that we will make a separate PR for moving the alert activations after this PR and the one @rokatyy working on ( https://iguazio.atlassian.net/browse/ML-8139 ) are merged

server/py/framework/db/sqldb/db.py
Yacouby: this PR is still not merged, we can modify it in the upcoming PR for this req: https://iguazio.atlassian.net/browse/ML-8139

moranbental: @quaark Fyi

moranbental (approved):
LGTM

rokatyy (approved):
Great job!

TomerShor (approved):
Good job!

PR #6683, opened 2024-11-11
[API] Query DB for only allowed projects when listing with `project=*`

quaark:
https://iguazio.atlassian.net/browse/ML-8260

When querying a `list` endpoint with the parameter `project=*`, first query all the project names from the DB, use OPA's `filter_allowed` to filter out of the list any project the user isn't allowed to view. Then query the records from the DB with a `project IN (X, Y, Z)` query for each project in the list. This is to ensure this filtering happens before paginating the results so the page size remains consistent.

This affects the following resources:
- alerts
- runs
- functions
- runtime_resources (already works this way - no change)
- schedules
- pipelines (not a DB implementation, required new handling for filtering for multiple projects)


liranbg (approved):

server/py/services/api/crud/projects.py
```diff
@@ -286,6 +287,36 @@ def list_projects(
...
+        if project != "*":
+            await (
+                framework.utils.auth.verifier.AuthVerifier().query_project_permissions(
+                    project,
+                    mlrun.common.schemas.AuthorizationAction.read,
+                    auth_info,
+                )
+            )
+            return [project]
+        else:
```
roei3000b: Redundant else

server/py/services/api/tests/unit/api/test_runs.py
```diff
@@ -990,3 +993,18 @@ def assert_time_range_request(client: TestClient, expected_run_uids: list, **fil
...
+def _generate_project(name="project-name"):
+    return mlrun.common.schemas.Project(
+        metadata=mlrun.common.schemas.ProjectMetadata(
+            name=name,
+            created=datetime.utcnow() - timedelta(seconds=1),
+            labels={
+                "some-label": "some-label-value",
+            },
+        ),
+        spec=mlrun.common.schemas.ProjectSpec(
```
roei3000b, line 1007: Is there reason why not using [_generate_project](https://github.com/mlrun/mlrun/blob/40f3ea3ad5fb53e2bd3f3071d1232c1147a8df67/server/py/services/api/tests/unit/db/test_projects.py#L371-L383) from here?

roei3000b (changes requested):
Looks great!
minor.

server/py/framework/db/sqldb/db.py
```diff
@@ -4798,7 +4807,9 @@ def _find_functions(
         """
         query = session.query(Function, Function.Tag.name)
 
-        if project != "*":
+        if isinstance(project, list):
+            query = query.filter(Function.project.in_(project))
+        elif project and project != "*":
```
liranbg: DRY, I think you can move to a base common func
`query = self._filter_project_resource(query, <resource>, project)`, whereas resource would be `X.project`
which will do exactly that

server/py/services/api/tests/unit/api/test_runs.py
quaark: You usually don't import helper functions from other test suites.
To make this a common helper utility for all tests, I'd make it more robust and accept more parameters to the point where it's easier to just build the schema object and not use the function.
Usually different suites have different cases, and in these 2 cases, the project is empty, but the 2 suites might pull these methods in different directions later. There are indeed other `_generate_project` methods in the tests other than these 2 that are more specific.

server/py/services/api/crud/pipelines.py
```diff
@@ -127,7 +131,17 @@ def list_pipelines(
...
                 mlrun_pipelines.models.PipelineRun(run) for run in response.runs or []
             ]
+
+            if isinstance(project, list) and project:
+                project_runs = []
+                for run in runs:
+                    run_project = self.resolve_project_from_pipeline(run)
+                    if run_project in project:
+                        project_runs.append(run)
+                runs = project_runs
```
moranbental:
```suggestion
                runs = [run for run in runs if self.resolve_project_from_pipeline(run) in project]
```

moranbental (approved):
Nice work! just one suggestion

moranbental (changes requested):
I think you should also handle this case: when listing artifacts with project == '*'
https://github.com/mlrun/mlrun/blob/40f3ea3ad5fb53e2bd3f3071d1232c1147a8df67/server/py/framework/db/sqldb/db.py#L3196

quaark:
> I think you should also handle this case: when listing artifacts with project == '*'
> 
> https://github.com/mlrun/mlrun/blob/40f3ea3ad5fb53e2bd3f3071d1232c1147a8df67/server/py/framework/db/sqldb/db.py#L3196

Maybe in a separate PR, it's an added feature that we're not sure is wanted / needed

server/py/services/api/crud/pipelines.py
quaark: This code is being reverted anyway, pipelines sadly needs to be handled differently

roei3000b (approved):

moranbental (approved):

server/py/services/api/crud/pipelines.py
```diff
@@ -315,6 +273,75 @@ def create_pipeline(
...
+                sort_by,
+                filter_,
+            )
+            runs = [mlrun_pipelines.models.PipelineRun(run) for run in runs or []]
+            # In-memory filtering turns Kubeflow's counting inaccurate if there are multiple pages of data
+            # so don't pass it to the client in such case
+            if next_page_token:
+                total_size = -1
+            else:
+                total_size = len(runs)
```
moranbental:
```suggestion
            total_size = len(runs) if next_page_token is None else -1

```

server/py/services/api/crud/pipelines.py
```diff
@@ -315,6 +273,75 @@ def create_pipeline(
...
+        else:
+            runs = []
+            while next_page_token is not None:
```
moranbental: Can `next_page_token` be an empty string?

moranbental (changes requested):
minor

moranbental (approved):
LGTM

server/py/services/api/crud/pipelines.py
```diff
@@ -315,6 +273,72 @@ def create_pipeline(
...
+                page_runs, next_page_token = self._list_runs_from_kfp(
+                    kfp_client,
+                    page_token,
+                    page_size or mlrun.common.schemas.PipelinesPagination.max_page_size,
+                    sort_by,
+                    filter_,
+                )
+                runs.extend(
+                    [mlrun_pipelines.models.PipelineRun(run) for run in page_runs or []]
+                )
```
yaelgen: Maybe move this section into a separate function that calls `_list_runs_from_kfp` and directly returns a list of `PipelineRun` instances, and it would help to add comments before each if and else block to clarify the purpose.

yaelgen (changes requested):
Looks great! added a minor suggestion

yaelgen (approved):
🚀

PR #6687, opened 2024-11-12
[CI] Add support for running the linter with Python 3.11

moranbental:
In MLRun 1.8, the official runtime remains Python 3.9. However, in preparation for the transition to Python 3.11, the following changes have been implemented:

**CI Pipeline Update**: Python 3.11 has been added to the CI matrix alongside Python 3.9. This change will ensure that the MLRun codebase is tested for compatibility with both Python versions.

**Testing on Python 3.11**: Running unit and integration tests on Python 3.11 has caused some issues with protobuf, which will be addressed in a separate [task](https://iguazio.atlassian.net/browse/ML-8396).

Also, bumping the `pyyaml` package to version `>=6.0.2` to support both Python 3.9 and Python 3.11

https://iguazio.atlassian.net/browse/ML-7282


Makefile
```diff
@@ -633,12 +633,20 @@ fmt: ## Format the code using Ruff and blacken-docs
 	python -m ruff check --fix-only
 	python -m ruff format
 	@echo "Formatting the code blocks with blacken-docs..."
+	# TODO: Update format to use py311 for Python 3.11
```
adamdelman: If we have the python version available in MLRUN_PYTHON_VERSION, why not use that to determine which value gets passed to the `-t` argument?

adamdelman (changes requested):

Makefile
```diff
@@ -633,12 +633,26 @@ fmt: ## Format the code using Ruff and blacken-docs
...
 	@echo "Checking the code blocks with blacken-docs"
-	git ls-files -z -- '*.md' | xargs -0 blacken-docs -t=py39 --check
+	@if [ "$(MLRUN_PYTHON_VERSION)" = "3.9" ]; then \
+	    git ls-files -z -- '*.md' | xargs -0 blacken-docs -t=py39 --check; \
+	elif [ "$(MLRUN_PYTHON_VERSION)" = "3.11" ]; then \
+	    git ls-files -z -- '*.md' | xargs -0 blacken-docs -t=py311 --check; \
+	else \
+	    echo "Unsupported Python version: $(MLRUN_PYTHON_VERSION)"; \
+	    exit 1; \
+	fi
```
alonmr: you can do this in 1-2 liner
see this example
```
# expected to be in the form of '-py<major><minor>' e.g. '-py39'
MLRUN_ANACONDA_PYTHON_DISTRIBUTION := $(shell echo "$(MLRUN_PYTHON_VERSION)" | awk -F. '{print "-py"$$1$$2}')
```
Calculate the env var you need and use it in the command
e.g.
```
MLRUN_LINT_PYTHON_VERSION=... # will be py39 or py311
git ls-files -z -- '*.md' | xargs -0 blacken-docs -t="${MLRUN_LINT_PYTHON_VERSION}" --check;
```

.github/workflows/ci.yaml
```diff
@@ -83,44 +83,82 @@ jobs:
...
         export version_suffix=$(echo "${{ steps.git_info.outputs.branch }}" | grep -E "^[0-9]+\.[0-9]+\.x$" | tr -d '.')
         export unstable_tag=$(if [ -z "$version_suffix" ]; then echo "unstable-cache"; else echo "unstable-cache-$version_suffix";fi)
         echo "tag=$(echo $unstable_tag)" >> $GITHUB_OUTPUT
+
+    - name: Install automation scripts dependencies
+      run: pip install -r automation/requirements.txt
+
     - name: Run Dockerized tests
-      run: MLRUN_DOCKER_REGISTRY=ghcr.io/ MLRUN_DOCKER_CACHE_FROM_TAG=${{ steps.docker_cache.outputs.tag }} make test-integration-dockerized
+      run: MLRUN_DOCKER_REGISTRY=ghcr.io/ MLRUN_DOCKER_CACHE_FROM_TAG=${{ steps.docker_cache.outputs.tag }} MLRUN_PYTHON_VERSION=${{ matrix.python-version }} make test-integration-dockerized
```
TomerShor: Easier to read and extend
```suggestion
      run: |
          MLRUN_DOCKER_REGISTRY=ghcr.io/ \
          MLRUN_DOCKER_CACHE_FROM_TAG=${{ steps.docker_cache.outputs.tag }} \
          MLRUN_PYTHON_VERSION=${{ matrix.python-version }} \
          make test-integration-dockerized
```

.github/workflows/ci.yaml
```diff
@@ -83,44 +83,82 @@ jobs:
...
+    - name: Install automation scripts dependencies
+      run: pip install -r automation/requirements.txt
+
     - name: Run Dockerized tests
-      run: MLRUN_DOCKER_REGISTRY=ghcr.io/ MLRUN_DOCKER_CACHE_FROM_TAG=${{ steps.docker_cache.outputs.tag }} make test-dockerized
+      run: |
+        MLRUN_DOCKER_REGISTRY=ghcr.io/
+        MLRUN_DOCKER_CACHE_FROM_TAG=${{ steps.docker_cache.outputs.tag }}
+        MLRUN_PYTHON_VERSION=${{ matrix.python-version }}
+        make test-dockerized
```
TomerShor: The env vars are defined for the same make command, and no as separate commands
```suggestion
      run: |
        MLRUN_DOCKER_REGISTRY=ghcr.io/ \
        MLRUN_DOCKER_CACHE_FROM_TAG=${{ steps.docker_cache.outputs.tag }} \
        MLRUN_PYTHON_VERSION=${{ matrix.python-version }} \
        make test-dockerized
```

.github/workflows/docs-language-linter.yaml
```diff
@@ -42,6 +42,8 @@ jobs:
       - uses: actions/checkout@v4
       - name: Set up Python
         uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'  # Use Python 3.9 for link checking
```
TomerShor: why not 3.11? is there some package that doesn't support it?

.github/workflows/docs-language-linter.yaml
```diff
@@ -72,5 +74,8 @@ jobs:
           python -m pip install --upgrade pip~=24.0
           pip install -r dev-requirements.txt
       - name: Lint
-        run: SKIP_VALE_CHECK=true make lint-docs
+        run: |
+          # Set the MLRUN_PYTHON_VERSION environment variable based on matrix
+          echo "MLRUN_PYTHON_VERSION=${{ matrix.python-version }}" >> $GITHUB_ENV
+          make lint-docs
```
TomerShor: What about `SKIP_VALE_CHECK=true`?

TomerShor: Also, shouldn't you use the new `MLRUN_LINT_PYTHON_VERSION` env var here?

TomerShor (changes requested):

.github/workflows/docs-language-linter.yaml
```diff
@@ -38,12 +38,16 @@ jobs:
   docs_linkcheck:
     name: Check for broken links
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ 3.9, 3.11 ]
```
liranbg: I think that python 3.9 is enough at this stage

.github/workflows/docs-language-linter.yaml
```diff
@@ -38,10 +38,18 @@ jobs:
   docs_linkcheck:
     name: Check for broken links
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # TODO: Ensure compatibility with Python 3.11 as well
+        # Currently, MLRun requires KFP with protobuf, which fails to install on Python 3.11.
+        python-version: [ 3.9]
```
TomerShor:
```suggestion
        python-version: [ 3.9 ]
```

.github/workflows/ci.yaml
```diff
@@ -83,44 +83,87 @@ jobs:
...
 
   integration-tests:
     name: Run Dockerized Integration Tests
     runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        # TODO: Ensure compatibility with Python 3.11 as well
+        # Currently, MLRun requires KFP with protobuf, which fails to install on Python 3.11.
+        python-version: [ 3.9 ]
```
TomerShor: actually, I see that we sometimes use quotation and sometimes not (`"3.9"` vs `3.9`). Let's be aligned with that

TomerShor (approved):

.github/workflows/docs-language-linter.yaml
```diff
@@ -52,16 +60,16 @@ jobs:
...
         env:
           FORCE_COLOR: 1  # Force color output
-        run: make linkcheck
+        run: MLRUN_PYTHON_VERSION=${{ matrix.python-version }} make linkcheck
   lint:
     name: Lint docs
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ "3.9", "3.11" ]
```
liranbg: I think the value of having linting docs on both Python versions is currently very small, meaning, docs eventually is going to be either 3.9 or 3.11 but not both. I will leave that for the end, after we move to python 3.11 and because eventually is code is compiled with 3.9 and each strategy takes a runner from github workflow - we can safely remove the 3.11 from the matrix - leaving it to run on python 3.9 only.

liranbg (changes requested):

PR #6689, opened 2024-11-12
[Microservices] Initial alerts service separation

alonmr:
This PR moves the alerts service components to a different package while it still runs in a single instance with the API.
The alerts service currently assumes a single instance as it assumes "chief" configuration e.g.
```
mlrun.mlconf.httpdb.clusterization.role == mlrun.common.schemas.ClusterizationRole.chief
```
where needed.
The service container injects the alerts service to the alert/event endpoints with
```
# Overriding ``ServiceContainer`` with ``AlertsServiceContainer``:
@containers.override(framework.service.ServiceContainer)
class AlertsServiceContainer(containers.DeclarativeContainer):
    service = providers.Object(daemon.service)
```
⚠️ For the injection to work, the container must be wired after the container override. 
⚠️ Meaning we need to define the container with the override decorator, and then wire the container which is done in the daemon initialization method. 

Other endpoints are running as usual with no changes in the API service package.
The container needs to be expanded to support injection of the correct service to the activated endpoint by prefix and it needs to support holding multiple services.

The application starts the API service with mounts the alerts fastapi application to support all endpoints in a single instance. This is temporary as we plan to:
1. Separate the services
2. Create a hydra service that runs multiple services with a path prefix for each
3. Add tunneling for requesting a specific service where multiple services may answer (healthz)

https://iguazio.atlassian.net/browse/ML-8211


server/py/framework/service/__init__.py
```diff
@@ -104,6 +123,7 @@ async def _custom_setup_service(self):
 
     async def _teardown_service(self):
         await self._custom_teardown_service()
+        # TODO: Don't cancel periodic of mounted services
```
liranbg: Elaborate?

liranbg (approved):

server/py/framework/service/__init__.py
alonmr: The cancel is called twice:
1. On API teardown
2. On alerts teardown

But it cancels all periodic functions so the 2nd time is void

PR #6692, opened 2024-11-13
[Alert Activation] Set id column to auto-increment in alert_activation table

Yacouby:
This PR modifies the id column in the alert_activation table to auto increment, ensuring that each entry has a unique, sequential ID.

The current table definition uses a composite primary key on both `id` and `activation_time` to satisfy partitioning requirements, which prevented the id column from being a unique identifier on its own. As a result, the database did not automatically increment the id.


server/py/services/api/migrations/versions/21b8e805f7ab_add_autoincrement_to_alert_activation.py
TomerShor, line 1: The file is called `add auto increment` but it's only adding a unique constraint.
1. Make sure the auto increment does indeed apply.
2. If it does and it's not needed in the migration - rename the migration file.

TomerShor (changes requested):

server/py/services/api/migrations/versions/21b8e805f7ab_add_autoincrement_to_alert_activation.py
Yacouby: seems like Mysql does not like the unique constraint:
"A UNIQUE INDEX must include all columns in the table's partitioning function"

rokatyy: https://dev.mysql.com/doc/refman/8.0/en/partitioning-limitations-partitioning-keys-unique-keys.html

`every unique key on the table must use every column in the table's partitioning expression`

rokatyy (commented):
This will fix everything:https://github.com/mlrun/mlrun/pull/6693

PR #6693, opened 2024-11-13
[Alert Activation] Make id column auto-incrementable

rokatyy:
What this PR does is quite a bad magic in order to:

1. Support the existing unit tests for SQLite
2. Enable `autoincrement` for `id` by changing the order in which PK is created (`id` should be 1st). Also, you can notice that we don't have `autoincrementable=True` in models.py for `id`. It's done for purpose as we manually make it autoincrementable in the migration and for SQLite it remains non-autoincrementable.
image


Yacouby (approved):
nice !

server/py/services/api/migrations/versions/12d8000e609b_change_alert_activation_pk.py
```diff
@@ -0,0 +1,63 @@
...
+    op.drop_constraint("_alert_activation_uc", "alert_activation", type_="primary")
+    op.create_primary_key(
+        "_alert_activation_uc", "alert_activation", ["id", "activation_time"]
+    )
+    # Only add AUTO_INCREMENT if not using SQLite
+    conn = op.get_bind()
+    if conn.dialect.name != "sqlite":
+        op.execute("""
+            ALTER TABLE alert_activation
+            MODIFY COLUMN id INT NOT NULL AUTO_INCREMENT
```
TomerShor, line 44: What about UNIQUE?

rokatyy: @TomerShor we don't need it, as id is autoincremental and a part of PK. And even more, it is not possible to add it because of this limitation - https://dev.mysql.com/doc/refman/8.0/en/partitioning-limitations-partitioning-keys-unique-keys.html.

liranbg (approved):

TomerShor (approved):

PR #6695, opened 2024-11-14
[API] Allow listing kind-specific functions

liranbg:
It is required to let clients filter functions by their kind (remote / local / handler / ... all other runtime kinds)
this way, clients such as sdk / ui can show better granular tables filtering by the function kinds

https://iguazio.atlassian.net/browse/ML-7075


mlrun/db/httpdb.py
```diff
@@ -1389,7 +1389,9 @@ def list_functions(
         name: Optional[str] = None,
         project: Optional[str] = None,
         tag: Optional[str] = None,
+        kind: Optional[str] = None,
```
alonmr: I believe we need to be adding parameters at the end of functions to keep BC of clients using positional arguments

server/py/framework/db/sqldb/db.py
```diff
@@ -1898,6 +1898,7 @@ def store_function(
         fn.updated = updated
         labels = get_in(function, "metadata.labels", {})
         update_labels(fn, labels)
+        fn.kind = function.pop("kind", None)
```
alonmr, line 1903: IIUC you pop to avoid data duplication. I think it's worth adding a comment

alonmr (commented):
Looks good! Minor comment

alonmr (approved):

mlrun/db/base.py
```diff
@@ -213,11 +213,13 @@ def delete_function(self, name: str, project: str = ""):
     def list_functions(
         self,
         name: Optional[str] = None,
-        project: Optional[str] = "",
-        tag: Optional[str] = "",
+        project: Optional[str] = None,
+        tag: Optional[str] = None,
+        kind: Optional[str] = None,
```
moranbental, line 218: Move the `kind` and `format_` fields to the bottom

mlrun/db/nopdb.py
```diff
@@ -230,12 +230,14 @@ def delete_function(self, name: str, project: str = ""):
...
+        name: Optional[str] = None,
+        project: Optional[str] = None,
+        tag: Optional[str] = None,
+        kind: Optional[str] = None,
         labels: Optional[Union[str, dict[str, Optional[str]], list[str]]] = None,
-        since=None,
-        until=None,
+        format_: mlrun.common.formatters.FunctionFormat = mlrun.common.formatters.FunctionFormat.full,
+        since: Optional[datetime.datetime] = None,
+        until: Optional[datetime.datetime] = None,
```
moranbental, line 240: Move the `kind` and `format_` fields to the bottom

server/py/framework/db/sqldb/db.py
```diff
@@ -4807,6 +4816,9 @@ def _find_functions(
         if hash_key is not None:
             query = query.filter(Function.uid == hash_key)
 
+        if kind is not None:
```
moranbental, line 4819: if kind is an empty string?

moranbental (changes requested):
Looks good! just minor suggestion

server/py/framework/db/sqldb/db.py
liranbg: there are function with kind empty (used to be local or handler, cant really recall) - and that should take action

PR #6700, opened 2024-11-15
[Microservices] Move alert templates to alerts service

rokatyy:
Move alert templates to alerts service


server/py/services/alerts/main.py
TomerShor, line 1: @rokatyy @alonmr 
If I understand correctly, if a service is in charge of multiple resources, all of the resources' endpoint methods are implemented in the same `Service` class.
In this case, the alerts service is in charge of alert_configs, events, alert_templates and alert_activations, and the `Service` class implements every `store`, `list` etc for each one.

IMO this is not ideal and makes quite a mess and harder to maintain, as opposed to how we had it previously where we had a separate file for each resource.

alonmr: We can explore separating the endpoints implementation in different ways but this is not part of the MVP so not part of the current considerations. The implementation needs to be service specific as we want separation of responsibilities and concerns. We might want to do something similar as in zebo with resources but again not part of the current effort.

server/py/framework/routers/alert_template.py
```diff
@@ -0,0 +1,110 @@
...
+        name,
+        auth_info,
+        db_session,
+    )
+
+
+@router.get("", response_model=list[mlrun.common.schemas.AlertTemplate])
+@inject
+async def list_alert_templates(
+    auth_info: mlrun.common.schemas.AuthInfo = Depends(deps.authenticate_request),
```
alonmr, line 80: add `request: Request,` it will be needed when the API service will forward requests to alerts service

server/py/framework/routers/alert_template.py
```diff
@@ -0,0 +1,110 @@
...
+    db_session: Session = Depends(deps.get_db_session),
+    service: framework.service.Service = Depends(
+        Provide[framework.service.ServiceContainer.service]
+    ),
+) -> list[mlrun.common.schemas.AlertTemplate]:
+    return await service.handle_request(
+        "list_alert_templates",
+        auth_info,
+        db_session,
+    )
```
alonmr, line 91:
```suggestion
        request,
        auth_info,
        db_session,
    )
```

server/py/services/alerts/main.py
```diff
@@ -288,6 +289,99 @@ async def post_event(
...
+            mlrun.mlconf.httpdb.clusterization.role
+            != mlrun.common.schemas.ClusterizationRole.chief
+        ):
+            chief_client = framework.utils.clients.chief.Client()
+            data = await request.json()
+            return await chief_client.store_alert_template(
+                name=name, request=request, json=data
+            )
+
+        self.logger.debug("Storing alert template", name=name)
```
alonmr:
```suggestion
        self._logger.debug("Storing alert template", name=name)
```

server/py/services/alerts/main.py
```diff
@@ -288,6 +289,99 @@ async def post_event(
...
+        )
+        # TODO: Once alerts runs in its own pod - remove chief check
+        if (
+            mlrun.mlconf.httpdb.clusterization.role
+            != mlrun.common.schemas.ClusterizationRole.chief
+        ):
+            chief_client = framework.utils.clients.chief.Client()
+            return await chief_client.delete_alert_template(name=name, request=request)
+
+        self.logger.debug("Deleting alert template", name=name)
```
alonmr: same

server/py/services/alerts/main.py
```diff
@@ -288,6 +289,99 @@ async def post_event(
...
+            auth_info,
+        )
+
+        return await run_in_threadpool(
+            services.alerts.crud.AlertTemplates().get_alert_template, db_session, name
+        )
+
+    async def list_alert_templates(
+        self,
+        auth_info: mlrun.common.schemas.AuthInfo,
```
alonmr, line 346:
```suggestion
        request: fastapi.Request,
        auth_info: mlrun.common.schemas.AuthInfo,
```

alonmr (changes requested):
Well done! Minor comments

alonmr (approved):

PR #6705, opened 2024-11-17
[Artifacts] Added producer_uri column to Artifact Table

moranbental:
Artifact `producer_uri` filtered After db query when listing artifacts. As with other filters applied after the SQL query, this approach causes issues with the number of items and the order of pages returned in the API.

To resolve this, several changes were made:
1. Added a producer_uri column to the artifact table.
2. Modified store_artifact to populate the new producer_uri column.
3. Updated list_artifacts to use this column in the SQL query for filtering by producer_uri.
4. Implemented a data migration to populate the producer_uri column for all existing artifacts during an upgrade.

https://iguazio.atlassian.net/browse/ML-8259


alonmr (approved):
Nicely done!

server/py/services/api/initial_data.py
```diff
@@ -877,40 +877,84 @@ def _perform_version_8_data_migrations(
...
+        "kind",
+        _handle_function_kind,
+    )
+    _migrate_data(
+        db,
+        db_session,
+        framework.db.sqldb.models.ArtifactV2,
+        "producer_uri",
+        _handle_artifact_producer_uri,
+    )
```
TomerShor: I think it is now too general and you can't understand what is the migration doing.
There are 2 options:
1. have another wrapper function for each (e.g. `_ensure_function_kind` and `_add_producer_uri_to_artifact` that will call `_migrate_data`
2. Add comments on what each call does.

Although the first option seems like an overkill I think I prefer it for readability and functionality, but your choice.

server/py/services/api/initial_data.py
```diff
@@ -877,40 +877,84 @@ def _perform_version_8_data_migrations(
...
-            db._commit(db_session, functions)
-        if len(functions) < chunk_size:
-            logger.info("No more functions to migrate")
+        logger.info(f"Fetching {model.__name__.lower()}s to migrate", offset=offset)
+
+        # Build query to fetch the data in chunks
+        query = (
+            db._query(db_session, model)
+            .order_by(getattr(model, "id"))
+            .limit(chunk_size)
```
theSaarco: Don't we want to filter by `model.field_name != None`? Since this is chunked, it's possible that migration would go half-way and then the service is restarted. In this case you want to only work on records that weren't already migrated.
Even without restart, doing it this way means you won't need `.offset()` anymore since any new query starts from offset 0 (anything that was already updated is not returned by the DB), and this actually make the query more efficient overall.

server/py/services/api/initial_data.py
```diff
@@ -877,40 +877,84 @@ def _perform_version_8_data_migrations(
...
+    chunk_size: int = 500,
 ):
     offset = 0
-    number_of_functions_to_migrate = (
-        db._query(db_session, framework.db.sqldb.models.Function)
-        .filter(framework.db.sqldb.models.Function.kind.is_(None))
+    # Get count of records to migrate
+    count = (
+        db._query(db_session, model)
+        .filter(getattr(model, field_name).is_(None))
```
TomerShor: If you generalizing the `_migrate_data` function I would accept the filter function as an argument too, and not limit to only checking if a field is `None`

server/py/services/api/initial_data.py
```diff
@@ -877,40 +877,84 @@ def _perform_version_8_data_migrations(
...
+        db_session,
+        framework.db.sqldb.models.ArtifactV2,
+        "producer_uri",
+        _handle_artifact_producer_uri,
+    )
+
 
+def _handle_function_kind(records: list, record):
+    function_dict = record.struct
+    record.kind = function_dict.pop("kind", None)
```
theSaarco: I know this isn't part of this PR, but @liranbg - why are you `pop`-ing the field from the dictionary? You'll need to re-introduce it into the dictionary when retrieving objects. Any reason for that?

TomerShor (changes requested):
Great stuff, have a couple of comments wrt generalizing the migration

server/py/services/api/initial_data.py
theSaarco: Here you're filtering on `is none` to count the records correctly. But it's not consistent with the actual query done later...

theSaarco (changes requested):
Nice job, but I think the migration query is sub-optimal, see my comments.

server/py/services/api/initial_data.py
liranbg: to avoid data duplication. I re-add it back when you retrieve the function

alonmr: it is to avoid data duplication, we discussed it on his PR and he added unit tests to ensure it is reintroduced on pulls

theSaarco: I'm not sure why data duplication is an issue here. Most of the fields in the DB are also part of the dictionary to begin with (project, name etc.). Making this a one-time behavior doesn't make the code any clearer, as it's extra logic that is unique to this case.

liranbg: you're stating a problem, not a solution. since I do not intend to preserve the correctness of both body/column data, ive set the value to be taken from the column and enrich the body at GET time to ensure smooth BC.

server/py/services/api/initial_data.py
```diff
@@ -877,41 +877,104 @@ def _perform_version_8_data_migrations(
...
+    chunk_size: int = 500,
 ):
-    offset = 0
-    number_of_functions_to_migrate = (
-        db._query(db_session, framework.db.sqldb.models.Function)
-        .filter(framework.db.sqldb.models.Function.kind.is_(None))
-        .count()
+    def handle_function_kind(record):
+        function_dict = record.struct
+        record.kind = function_dict.pop("kind", None)
```
liranbg: since you are now iterating over function kinds and filter by `None` - it is a must to default to `""` instead of `None`. because if there is (by any chance) a function with kind `None`, then the data migration would iterate over it forever.

liranbg (changes requested):

server/py/services/api/initial_data.py
```diff
@@ -877,41 +877,104 @@ def _perform_version_8_data_migrations(
...
+    chunk_size: int = 500,
+):
+    # Query for records that need migration
+    records = db._query(db_session, model).filter(filter_func).limit(chunk_size).all()
+
+    if not records:
+        logger.info(f"No records to migrate for {model.__name__.lower()}")
+        return
+
+    logger.info(f"Starting migration for {len(records)} {model.__name__.lower()}s")
```
TomerShor: since `artifactv2s` looks weird, I'd go with:
```suggestion
    logger.info(f"Starting migration for {len(records)} {model.__name__.lower()} records")
```

server/py/services/api/initial_data.py
```diff
@@ -877,41 +877,104 @@ def _perform_version_8_data_migrations(
...
+        to_commit = [
+            handled_record
+            for record in records
+            if (handled_record := handle_field_record_func(record)) is not None
+        ]
+
+        # Commit if there are records to migrate
+        if to_commit:
+            logger.info(
+                f"Committing migrated {model.__name__.lower()}s", count=len(to_commit)
```
TomerShor:
```suggestion
                f"Committing migrated records", count=len(to_commit), model=model.__name__
```

server/py/services/api/initial_data.py
```diff
@@ -877,41 +877,104 @@ def _perform_version_8_data_migrations(
...
+            db._commit(db_session, to_commit)
+
+        # Fetch next batch of records to migrate (if any)
+        records = (
+            db._query(db_session, model).filter(filter_func).limit(chunk_size).all()
+        )
+
+        # If no records left to migrate, stop
+        if not records:
+            logger.info(f"No more {model.__name__.lower()}s to migrate")
```
TomerShor:
```suggestion
            logger.info(f"No more records to migrate", model=model.__name__)
```

server/py/services/api/initial_data.py
```diff
@@ -877,41 +877,104 @@ def _perform_version_8_data_migrations(
 def _perform_version_9_data_migrations(
     db: framework.db.sqldb.db.SQLDB, db_session: sqlalchemy.orm.Session
 ):
-    _migrate_function_kind(db, db_session, chunk_size=500)
+    _migrate_function_kind(db, db_session)
+    _migrate_artifact_producer_uri(db, db_session)
 
 
 def _migrate_function_kind(
```
TomerShor: Although this is a part of a "migration", I still prefer `_ensure_function_kind` because we're not really migrating the kind from anywhere. Better be explicit on what we are actually doing here

server/py/services/api/initial_data.py
```diff
@@ -877,41 +877,104 @@ def _perform_version_8_data_migrations(
...
-            functions.append(function)
-        if functions:
-            logger.info("Committing migrated functions", count=len(functions))
-            db_session.add_all(functions)
-            db._commit(db_session, functions)
-        if len(functions) < chunk_size:
-            logger.info("No more functions to migrate")
+
+
+def _migrate_artifact_producer_uri(
```
TomerShor: Same here with `_add_producer_uri_to_artifact`

TomerShor (changes requested):
Minor comments

server/py/services/api/initial_data.py
moranbental: Fixed it and also updated the unit tests to cover that

server/py/services/api/initial_data.py
theSaarco: All I'm saying is that you're not solving anything like this - you're still copying from and to the DB column, with the only difference being that you're removing the value from the dictionary while doing that. It is still exposed to bugs, only different ones - instead of having a potential difference between values, you might run into a situation where on a new flow someone forgot to update the dict and it's still different. 
Given that you just changed the problem, not solved it, I think it's preferred to have uniform code/logic across the board.

theSaarco (approved):
LGTM

liranbg (approved):

quaark (approved):
🚀

TomerShor (approved):

PR #6706, opened 2024-11-17
[Model Monitoring] Avoid project deletion failure when TDEngine drop tables fails

Eyal-Danieli:
If an issue occurs while deleting TDEngine resources, we will log the error as a warn and ensure it does not disrupt the project deletion API. In this case, we will suggest the user to delete the resources manually. 

https://iguazio.atlassian.net/browse/ML-8246

### DIFF
--- a/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
+++ b/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
@@ -242,9 +242,9 @@ class TDEngineConnector(TSDBConnector):
                 )
             except Exception as e:
                 logger.warning(
-                    "Failed to delete TDEngine resources, you may need to delete them manually."
-                    "Please note that currently, the available project resources can be found "
-                    "under the following supertables: 'app_results,' 'metrics,' and 'predictions.'",
+                    "Failed to drop TDEngine tables. You may need to drop them manually. "
+                    "These can be found under the following supertables: app_results, "
+                    "metrics, and predictions.",
                     project=self.project,
                     error=mlrun.errors.err_to_str(e),
                 )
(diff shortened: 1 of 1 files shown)


mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -234,9 +234,18 @@ def delete_tsdb_resources(self):
...
-            )
+            try:
+                self.connection.run(
+                    statements=drop_statements,
+                    timeout=self._timeout,
+                    retries=self._retries,
+                )
+            except Exception as e:
+                logger.warning(
+                    "Failed to delete TDEngine resources, you may need to delete them manually",
```
gtopper: How will the user know what to delete though?

Eyal-Danieli: It quite depends the error,  but I’ve improved the error message to guide the user to the specific resources - https://github.com/mlrun/mlrun/pull/6706/commits/51d6391be1dcf42e5aa902dad1fc1837c7a90d7d

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -234,9 +234,20 @@ def delete_tsdb_resources(self):
...
+                self.connection.run(
+                    statements=drop_statements,
+                    timeout=self._timeout,
+                    retries=self._retries,
+                )
+            except Exception as e:
+                logger.warning(
+                    "Failed to delete TDEngine resources, you may need to delete them manually."
+                    "Please note that currently, the available project resources can be found "
+                    "under the following supertables: 'app_results,' 'metrics,' and 'predictions.'",
```
gtopper: Can we be more succinct about it? Something like:
```suggestion
                    "Failed to drop TDEngine tables. You may need to drop them manually. "
                    "These an be found under the following supertables: app_results, "
                    "metrics, and predictions.",
```

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
gtopper: You mean you improved the warning :)

gtopper (approved):

PR #6715, opened 2024-11-18
[Tutorials] Fix some errors in MM notebook

jond01:
Port #6712 to dev.


assaf758 (approved):

docs/tutorials/05-model-monitoring.ipynb
```diff
@@ -82,7 +82,7 @@
     "### Log the model with training data\n",
     "\n",
     "See the parameter descriptions in {py:meth}`~mlrun.projects.MlrunProject.log_model`. \n",
-    "{Download the pickle file}`pickle file <./src/model2.pkl>` used in this example."
+    "Download the {download}`pickle file <./src/model.pkl>` used in this example."
```
jillnogold, line 85: There's another pkl file for a different demo - I compared the contents and they were not the same. Hence the name model2.
Are you saying that they should be the same?

docs/tutorials/05-model-monitoring.ipynb
```diff
@@ -119,9 +119,10 @@
     "model_name = \"RandomForestClassifier\"\n",
     "project.log_model(\n",
     "    model_name,\n",
-    "    model_file=\"model2.pkl\",\n",
+    "    model_file=\"src/model.pkl\",\n",
```
jillnogold, line 122: same here

docs/tutorials/05-model-monitoring.ipynb
jond01: Which different demo do you refer to? I looked up references to this file (`src/model.pkl`) and found only this notebook.

jillnogold: Looking back, there was another in a transient version of a tutorial. So name change is good.

PR #6716, opened 2024-11-18
[Alert activations] Change periodic partition management interval

rokatyy:
* Change _start_periodic_partition_management interval to partition_interval 
 * reduce the number of partitions that we keep from 2*retention_time to retention_time+3*partition_time

### DIFF
--- a/server/py/services/api/main.py
+++ b/server/py/services/api/main.py
@@ -558,7 +558,9 @@ class Service(framework.service.Service):
                 services.api.utils.db.partitioner.MySQLPartitioner().get_partition_interval,
                 table_name=table_name,
             )
-            interval_in_seconds = partition_interval.as_duration().total_seconds()
+            interval_in_seconds = int(
+                partition_interval.as_duration().total_seconds() / 2
+            )
             run_function_periodically(
                 interval_in_seconds,
                 f"{self._manage_partitions.__name__}_{table_name}",
--- a/server/py/services/api/main.py
+++ b/server/py/services/api/main.py
@@ -554,7 +554,11 @@ class Service(framework.service.Service):
                 f"Starting periodic partition management for table {table_name}",
                 retention_days=retention_days,
             )
-            interval_in_seconds = retention_days * 24 * 60 * 60
+            partition_interval: mlrun.common.schemas.partition.PartitionInterval = framework.db.session.run_function_with_new_db_session(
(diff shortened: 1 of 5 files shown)


server/py/services/api/utils/db/partitioner.py
```diff
@@ -46,9 +47,12 @@ def create_and_drop_partitions(
             table_name,
         )
 
-        # Ensure partitions for double the retention time.
+        # Ensure partitions for the retention time plus
+        # mlrun.config.partitions_buffer_multiplier * partitioned_interval
```
TomerShor: no need to write the math twice
```suggestion
        # Ensure partitions for the retention time plus a buffer
```

TomerShor (changes requested):
Great stuff, really minor comment

TomerShor (approved):

server/py/services/api/main.py
```diff
@@ -554,7 +554,13 @@ def _start_periodic_partition_management(self):
                 f"Starting periodic partition management for table {table_name}",
                 retention_days=retention_days,
             )
-            interval_in_seconds = retention_days * 24 * 60 * 60
+            partition_interval = framework.db.session.run_function_with_new_db_session(
+                services.api.utils.db.partitioner.MySQLPartitioner().get_partition_interval,
+                table_name=table_name,
+            )
+            interval_in_seconds = int(
+                partition_interval.as_duration().total_seconds() / 2
```
moranbental, line 562: In the case where partition_interval is not DAY, MONTH, or YEARWEEK, it will return None, and attempting to call total_seconds() on None will result in an error. It may be better to raise a ValueError with the message 'Unsupported partition interval' in the as_duration function in that case.

moranbental (changes requested):
Looks good! just one suggestion

server/py/services/api/main.py
TomerShor: It's only half true - 
If we have a valid partition interval then it is an `StrEnum` so it must be one of the valid options.
However - `get_partition_interval` can possibly return `None`, and then we will get an error:
```
AttributeError: 'NoneType' object has no attribute 'as_duration'
```
So I agree there should be some defense here.

rokatyy: Changed the code of `from_function` to make it always either return `PartitionInterval` or raise an exception

moranbental (approved):
LGTM

TomerShor (approved):

PR #6717, opened 2024-11-18
[Microservices] Use uvicorn factory application

alonmr:
This allows us to wire the containers only once which will enable hydra initialization


server/py/framework/service/__init__.py
```diff
@@ -56,6 +56,7 @@ def initialize(self, mounts: dict):
 
     def _mount_services(self, mounts: dict):
         for path, service in mounts.items():
+            service.initialize()
```
liranbg, line 63:
```suggestion
            service.initialize(mounts)
```

liranbg (approved):

PR #6719, opened 2024-11-18
[Dependencies] Make Pydantic Schemas Pydantic V2 Compatible

quaark:
https://iguazio.atlassian.net/browse/ML-4122

From `pydantic==1.10` all the v1 pydantic features are namespaced in `pydantic.v1` which is still supported in pydantic v2. This is exactly for our use-case where we want to slowly migrate from v1 to v2 while the src supports both versions, during the transition - https://docs.pydantic.dev/latest/migration/#continue-using-pydantic-v1-features .


moranbental (approved):
LGTM

requirements.txt
```diff
@@ -20,11 +20,9 @@ requests~=2.32
 # >=0.8.6 from kfp 1.6.0 (and still up until 1.8.10)
 tabulate~=0.8.6
 v3io~=0.6.9
-# pydantic 1.10.8 fixes a bug with literal and typing-extension 4.6.0
-# https://docs.pydantic.dev/latest/changelog/#v1108-2023-05-23
-# TODO: loosen upperbound and remove the below comment once fixed upstream
-# pydantic 1.10.15 breaks backwards compatibility due to https://github.com/pydantic/pydantic/pull/9042
-pydantic>=1.10.8, <1.10.15
+# pydantic 2 requires python kfp 2
```
liranbg: no it doesnt :)
```suggestion
# kfp 1.8 blocks pydantic 2
```

liranbg (changes requested):

PR #6720, opened 2024-11-18
[Microservices] Implement inter-service communication

alonmr:
The final straw to completely enable running alerts service independently (helm chart is WIP)
---
### Discovery client
* Holds in memory services mapping. 
* Currently we have 2-3 services so this is a simple interpretation just to enable the microservices framework. 
* Resolves from configuration if runs as hydra or as microservices.
### Messaging client
* Enables forwarding requests
* Used by the API to forward requests to alerts service
* Uses discovery to find the services that matches the endpoint

TO DO:
* Sync messaging client for promptly triggering endpoints (e.g. when deleting project, the API will send a request to alerts to delete events)
* Adding endpoints to alerts must be added to discovery client as well. Find a way to force this or at least document. Ideally we would have service registration where the service could tell the API which requests it listens to (with open api docs?)
* Do not create db session if forwarding request
* More...

Closes:
* https://iguazio.atlassian.net/browse/ML-5139
* https://iguazio.atlassian.net/browse/ML-8212
* https://iguazio.atlassian.net/browse/ML-8332


server/py/services/api/tests/unit/test_daemon.py
```diff
@@ -0,0 +1,27 @@
...
+import framework.service
+
+
+def test_service_injection():
+    mlconf.services.hydra.services = "*"
+    service_container = framework.service.ServiceContainer()
+    service = service_container.service()
+    assert "services.alerts.main.Service" in str(service)
+
+    mlconf.services.hydra.services = ""
```
liranbg, line 25: Use a fixture to reset the value post change

server/py/framework/utils/clients/chief.py
```diff
@@ -13,24 +13,22 @@
...
-from mlrun.utils import logger
+
+import framework.utils.clients.messaging
 
 
 # we were thinking to simply use httpdb, but decided to have a separated class for simplicity for now until
 # this class evolves, but this should be reconsidered when adding more functionality to the class
 class Client(
-    metaclass=mlrun.utils.singleton.AbstractSingleton,
+    framework.utils.clients.messaging.BaseClient,
```
liranbg: I think that rather chief implement messaging client, it should hold that as a dependency, and the reason is that chief client is not a messaging client but rather holds it + state upon need.

server/py/framework/utils/clients/discovery.py
```diff
@@ -0,0 +1,150 @@
...
+        self.initialize()
+
+    def initialize(self):
+        # We use an ordered dict for control over service matching order.
+        # This is important for services that may have overlapping routes.
+        self.services: dict[str, ServiceInstance] = collections.OrderedDict()
+        if mlconf.services.hydra.services != "*":
+            self.register_service(service_name="alerts")
+        self.register_service(service_name="api")
+        # Chief is last to allow other services to override its routes
```
liranbg:
```suggestion
        # Must be last. Allowing other services to override its routes
```

server/py/framework/utils/clients/discovery.py
```diff
@@ -0,0 +1,150 @@
...
+    def _resolve_service_url(service_name: str) -> str:
+        """Resolve service URL by service name."""
+        return f"http://mlrun-{service_name}.{mlconf.namespace}.svc.cluster.local:8080"
+
+    @staticmethod
+    def _service_routes(service_name: str) -> list:
+        """Get all routes for a service."""
+        return {
+            "api-chief": [
+                # Alerts routes that require chief capabilities
```
liranbg: can you elaborate what do you meant by here? almost why doesnt `get` alter_templates is here?

server/py/framework/utils/clients/messaging.py
```diff
@@ -0,0 +1,268 @@
...
+        pass
+
+    def put(self):
+        pass
+
+    def delete(self):
+        pass
+
+    def get(self):
+        pass
```
liranbg: why not async?

liranbg (changes requested):
few comments and a question
how do you handle the project deletion sends request to alerts service deleting its events?

server/py/framework/utils/clients/discovery.py
```diff
@@ -0,0 +1,150 @@
...
+import mlrun.utils.singleton
+from mlrun import mlconf
+
+
+@dataclass
+class ServiceInstance:
+    name: str
+    url: str
+    method_routes: dict[str, list[re.Pattern]] = None
+    status: str = "UP"  # UP, DOWN, UNKNOWN
```
roei3000b: Any reason why don't use enum?

server/py/framework/utils/clients/discovery.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def initialize(self):
+        # We use an ordered dict for control over service matching order.
+        # This is important for services that may have overlapping routes.
+        self.services: dict[str, ServiceInstance] = collections.OrderedDict()
+        if mlconf.services.hydra.services != "*":
+            self.register_service(service_name="alerts")
+        self.register_service(service_name="api")
+        # Chief is last to allow other services to override its routes
+        self.register_service(service_name="api-chief")
```
roei3000b: I think it is a bit strange that the `discovery` client "know" each micro service.
Why the  `register_service` isn't call from the micro service itself?
If it is because overlapping routes and for making some order, anyway I don't think it is good idea that we will have overlapping routers, maybe to solve it we should add the service name prefix before the route?

server/py/framework/utils/clients/discovery.py
```diff
@@ -0,0 +1,150 @@
...
+            name=service_name,
+            url=url,
+            method_routes=method_routes,
+        )
+        self._logger.info("Registered service", service_name=service_name, url=url)
+
+    def deregister_service(self, service_name: str):
+        """Deregister a service instance."""
+        if service_name in self.services:
+            del self.services[service_name]
```
roei3000b:
```suggestion
        self.services.pop(service_name, None)
```

server/py/framework/utils/clients/discovery.py
```diff
@@ -0,0 +1,150 @@
...
+        :param path: URL path to match against service patterns
+        :return: ServiceInstance matching the request or None if no match
+        """
+        method = method.lower()
+        if service_name := self._find_service(method, path):
+            return self.get_service(service_name)
+        return None
+
+    def get_service(self, service_name: str) -> typing.Optional[ServiceInstance]:
+        """Get all healthy instances of a service."""
```
roei3000b: Isn't the services dict values are only one service per key?
If so, what is the meaning of all healthy instances of a service?

server/py/framework/utils/clients/discovery.py
```diff
@@ -0,0 +1,150 @@
...
+
+        :param method: HTTP method of the request
+        :param path: URL path to match against service patterns
+        :return: Name of matching service or None if no match
+        """
+        for service_name, service_instance in self.services.items():
+            routes_patterns = service_instance.method_routes[method]
+            for route_pattern in routes_patterns:
+                if route_pattern.match(path):
+                    return service_name
```
roei3000b, line 123: We will call this function very often, and the values on it will not change during the lifecycle of the discovery service, consider adding caching to this function. (Invalidate the cache when registering or derigstering service)

server/py/framework/utils/clients/discovery.py
```diff
@@ -0,0 +1,150 @@
...
+                (["*"], "projects/.+/alerts.*"),
+                (["post"], "projects/.+/events.*"),
+            ],
+            "api": [],
+            "alerts": [
+                (["put", "get", "delete"], "alert_templates.*"),
+                (["*"], "projects/.+/alerts.*"),
+                (["post"], "projects/.+/events.*"),
+            ],
+        }[service_name]
```
roei3000b: Again I think it is a bit strange that the discovery service "know" the microservices under it.

roei3000b (changes requested):
Looks very good!! :)
Left some questions and suggestions.

server/py/framework/utils/clients/discovery.py
alonmr: This is a dummy discovery client. It's not possible to call the discovery from the services since they are not initialized in all instances. The real way to do this is when the service starts it sends a request to register to a remote KV DB like etcd. But this will not happen in 1.8

server/py/framework/utils/clients/discovery.py
alonmr: This is for some hypothetical situation where you have some workers as hydra and some not. This would make them forward the request straight to chief instead of going into loop of workers. It's not really needed but I added it as a thought that it might be nice to use the messaging client as a router to chief as well.

server/py/services/api/tests/unit/test_daemon.py
alonmr: This is not needed because `config_test_base` resets then entire config to default values and it is auto used

server/py/framework/utils/clients/messaging.py
alonmr: I was contemplating if there is any point to keep these in this PR. Decided to remove them

server/py/framework/utils/clients/discovery.py
alonmr: Same about dummy discovery. I agree this is not the best place for defining the routed. I don't have a better idea atm. This should probably be held by the API service somewhere

server/py/framework/utils/clients/discovery.py
alonmr: removed as it's not really used

liranbg (approved):

server/py/framework/utils/clients/discovery.py
```diff
@@ -0,0 +1,142 @@
...
+        :param path: URL path to match against service patterns
+        :return: ServiceInstance matching the request or None if no match
+        """
+        method = method.lower()
+        if service_name := self._find_service(method, path):
+            return self.get_service(service_name)
+        return None
+
+    def get_service(self, service_name: str) -> typing.Optional[ServiceInstance]:
+        """Get the registered instances of a service."""
```
roei3000b:
```suggestion
        """Get the registered instance of a service."""
```

roei3000b (changes requested):
Minor

server/py/framework/utils/clients/discovery.py
```diff
@@ -0,0 +1,142 @@
...
+        self.services[service_name] = ServiceInstance(
+            name=service_name,
+            url=url,
+            method_routes=method_routes,
+        )
+        self._logger.info("Registered service", service_name=service_name, url=url)
+
+    def deregister_service(self, service_name: str):
+        """Deregister a service instance."""
+        self.services.pop(service_name, None)
```
moranbental, line 64: If `service_name` does not exist in `self.services`, you will log an incorrect message

server/py/framework/utils/clients/discovery.py
```diff
@@ -0,0 +1,142 @@
...
+):
+    def __init__(self):
+        self._logger = mlrun.utils.logger.get_child(__name__)
+        self.services = None
+        self.initialize()
+
+    def initialize(self):
+        # We use an ordered dict for control over service matching order.
+        # This is important for services that may have overlapping routes.
+        self.services: dict[str, ServiceInstance] = collections.OrderedDict()
```
moranbental, line 41: Why not initialize `self.services` directly in the constructor?

PR #6721, opened 2024-11-19
[Alerts] Add condition for resetting alert on reset policy change

Yacouby:
Added logic to handle reset policy changes from "manual" to "auto" when the alert is in an active state. 
This ensures the alert is reset appropriately in such scenarios, improving the alignment of alert behavior with reset policy transitions.

Jira:
https://iguazio.atlassian.net/browse/ML-7749

### DIFF
--- a/docs/concepts/alerts.md
+++ b/docs/concepts/alerts.md
@@ -157,7 +157,8 @@ The `ResetPolicy` options are:
 - manual &mdash; for manual reset of the alert
 - auto &mdash; if the criteria contains a time period such that the alert is reset once there are no more invocations in the relevant time window.
 
-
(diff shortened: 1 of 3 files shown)


rokatyy (approved):
Looks good to me! I think that's good idea not to leave it in active state until manual reset as it was before. 

Just a note - we will probably need to update documentation to explain this behaviour.

Yacouby:
> Just a note - we will probably need to update documentation to explain this behaviour.

I don’t think additional documentation is necessary here since this is the standard and correct behavior for handling reset policy transitions

rokatyy:
@Yacouby I think adding additional explanations is always a good idea.

TomerShor (approved):

TomerShor (approved):

PR #6723, opened 2024-11-19
[Alert activations] Implement api for project's alert activations

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-8139

Key changes:

* Added the API endpoint `/projects/{project}/alert-activations` (with pagination from get-go)
* Implemented httpdb methods to handle both paginated and non-paginated requests for this API.
* There was a discussion regarding the cleanup of alert activations during project deletion. It was decided that activations would not be cleaned up. Instead, the system will always filter alert activations based on the project name and project creation time, ensuring that activations are only considered if they occurred after the project's creation.


mlrun/db/httpdb.py
```diff
@@ -4666,6 +4666,43 @@ def list_alert_templates(self) -> list[mlrun.common.schemas.AlertTemplate]:
             results.append(mlrun.common.schemas.AlertTemplate(**item))
         return results
 
+    def list_alert_activations(
```
moranbental, line 4669: missing docstring and return value

mlrun/db/httpdb.py
```diff
@@ -4666,6 +4666,43 @@ def list_alert_templates(self) -> list[mlrun.common.schemas.AlertTemplate]:
...
+            name=name,
+            since=since,
+            until=until,
+            entity=entity,
+            severity=severity,
+            return_all=True,
+        )
+        return alert_activations
+
+    def paginated_list_alert_activations(
```
moranbental, line 4712: missing docstring and return value

mlrun/db/nopdb.py
```diff
@@ -865,3 +865,24 @@ def get_alert_template(self, template_name: str):
...
+    def list_alert_activations(
+        self,
+        project: Optional[str] = None,
+        name: Optional[str] = None,
+        since: Optional[datetime.datetime] = None,
+        until: Optional[datetime.datetime] = None,
+        entity: Optional[str] = None,
+        severity: Optional[list[str]] = None,
+    ):
+        pass
```
moranbental, line 884: In nopdb we usually return something. For example, in list_artifact, we return mlrun.lists.ArtifactList()

mlrun/db/nopdb.py
```diff
@@ -865,3 +865,24 @@ def get_alert_template(self, template_name: str):
...
+
+    def paginated_list_alert_activations(
+        self,
+        *args,
+        page: Optional[int] = None,
+        page_size: Optional[int] = None,
+        page_token: Optional[str] = None,
+        **kwargs,
+    ):
+        pass
```
moranbental, line 894: same here

server/py/framework/rundb/sqldb.py
```diff
@@ -1271,6 +1271,27 @@ def get_alert_template(self, template_name: str):
...
+
+    def paginated_list_alert_activations(
+        self,
+        *args,
+        page: Optional[int] = None,
+        page_size: Optional[int] = None,
+        page_token: Optional[str] = None,
+        **kwargs,
+    ):
+        pass
```
moranbental:
```suggestion
        raise NotImplementedError()
```

server/py/framework/rundb/sqldb.py
```diff
@@ -1271,6 +1271,27 @@ def get_alert_template(self, template_name: str):
...
+    def list_alert_activations(
+        self,
+        project: Optional[str] = None,
+        name: Optional[str] = None,
+        since: Optional[datetime.datetime] = None,
+        until: Optional[datetime.datetime] = None,
+        entity: Optional[str] = None,
+        severity: Optional[list[str]] = None,
+    ):
+        pass
```
moranbental: I think you need to implement this function

server/py/framework/db/sqldb/db.py
```diff
@@ -5929,13 +5930,55 @@ def store_alert_activation(
...
+        # excluding activations linked to any previous instances of the project.
+        # TODO: reconsider this approach when we move alerts out of main MLRun db
+        conditions = []
+        for project, created in project_with_creation_time:
+            conditions.append(
+                and_(
+                    AlertActivation.project == project,
+                    AlertActivation.activation_time > created,
+                )
+            )
```
```suggestion
        conditions = [
            and_(
                AlertActivation.project == project,
                AlertActivation.activation_time > created,
            )
            for project, created in project_with_creation_time
        ]
```

server/py/framework/db/sqldb/helpers.py
```diff
@@ -87,6 +89,19 @@ def generate_query_predicate_for_name(column, query_string):
         return column.__eq__(query_string)
 
 
+def generate_query_for_name_with_wildcard(column, query_string):
+    if query_string.startswith("~"):
+        return column.ilike(translate_wildcard_to_sql(query_string[1:]))
+    else:
+        column.__eq__(query_string)
```
moranbental: it doesn't return anything when the query_string does not start with ~

server/py/services/api/api/api.py
```diff
@@ -189,6 +190,11 @@
     tags=["nuclio"],
     dependencies=[Depends(deps.authenticate_request)],
 )
+api_router.include_router(
+    alert_activations.router,
+    tags=["alert-activations"],
+    dependencies=[Depends(deps.authenticate_request)],
```
moranbental, line 196: @alonmr more for you - why do we need to pass `dependencies=[Depends(deps.authenticate_request)]` for every `include_router` if it's the same for all?

moranbental (changes requested):
Nice! However, there are some missing tests regarding pagination in both the SDK and the server:

For the server side, see [test_list_artifacts_with_pagination](https://github.com/mlrun/mlrun/blob/40f3ea3ad5fb53e2bd3f3071d1232c1147a8df67/server/py/services/api/tests/unit/api/test_artifacts.py#L1014).
For the SDK, see [test_list_artifacts_with_pagination](https://github.com/mlrun/mlrun/blob/f2e5312ff81bf956d90d45964ee2cbbe02c4323f/tests/rundb/test_httpdb.py#L863).

server/py/framework/db/sqldb/helpers.py
rokatyy: @moranbental nice catch!

moranbental (changes requested):
Another thing: what about creating an SDK method for the project, something like `project.paginated_list_alert_activations()`?

server/py/framework/rundb/sqldb.py
rokatyy: after offline discussion it was decided it's not relevant

mlrun/db/nopdb.py
rokatyy: after offline discussion it was decided it's not relevant

rokatyy:
> Another thing: what about creating an SDK method for the project, something like `project.paginated_list_alert_activations()`?

as discussed in slack, it's part of another ticket and out of scope for this one

server/py/framework/db/sqldb/db.py
```diff
@@ -5929,13 +5930,55 @@ def store_alert_activation(
         self._upsert(session, [alert_activation_record])
 
     def list_alert_activations(
-        self, session: Session, project: typing.Optional[str] = None
+        self,
+        session: Session,
+        project_with_creation_time: list[tuple[str, datetime]],
```
```suggestion
        projects_with_creation_time: list[tuple[str, datetime]],
```

server/py/framework/db/sqldb/helpers.py
```diff
@@ -87,6 +89,19 @@ def generate_query_predicate_for_name(column, query_string):
         return column.__eq__(query_string)
 
 
+def generate_query_for_name_with_wildcard(column, query_string):
```
TomerShor, line 117: add a docstring with a simple example

server/py/services/api/api/endpoints/alert_activations.py
TomerShor, line 1: Why not implement this in the alerts service?

server/py/services/api/api/endpoints/alert_activations.py
```diff
@@ -0,0 +1,86 @@
...
+    since: Optional[str] = None,
+    until: Optional[str] = None,
+    entity: Optional[str] = None,
+    severity: Optional[list[str]] = None,
+    page: int = Query(None, gt=0),
+    page_size: int = Query(None, alias="page-size", gt=0),
+    page_token: str = Query(None, alias="page-token"),
+    auth_info: mlrun.common.schemas.AuthInfo = Depends(deps.authenticate_request),
+    db_session: Session = Depends(deps.get_db_session),
+):
```
TomerShor: Don't we support `entity_kind` and `event_kind` query params as [written in the HLD](https://iguazio.atlassian.net/wiki/spaces/ARC/pages/165085272/Alerts+history#3.1.2.-Logical-Components)?

mlrun/db/httpdb.py
TomerShor, line 1: Add methods to the `MlrunProject` that call these httpdb methods in https://github.com/mlrun/mlrun/blob/development/mlrun/projects/project.py

TomerShor (changes requested):
Looking good!
Have some minor comments

mlrun/db/httpdb.py
TomerShor: Just saw [your last comment](https://github.com/mlrun/mlrun/pull/6723#issuecomment-2485905282), that this will be in a followup PR.

server/py/services/api/api/endpoints/alert_activations.py
rokatyy: @TomerShor it will be moved in a separate PR, because this work was started before we started moving everything to alerts service

server/py/services/api/api/endpoints/alert_activations.py
rokatyy: @TomerShor will add them, missed

rokatyy: Added

mlrun/common/schemas/auth.py
```diff
@@ -90,6 +91,7 @@ def to_resource_string(
             AuthorizationResourceTypes.run: "/projects/{project_name}/runs/{resource_name}",
             AuthorizationResourceTypes.event: "/projects/{project_name}/events/{resource_name}",
             AuthorizationResourceTypes.alert: "/projects/{project_name}/alerts/{resource_name}",
+            AuthorizationResourceTypes.alert_activations: "/projects/{project_name}/alerts/{resource_name}/activations",
```
moranbental, line 94: It shouldn't be `/projects/{project_name}/alert-activations`?

mlrun/db/httpdb.py
```diff
@@ -4666,6 +4666,98 @@ def list_alert_templates(self) -> list[mlrun.common.schemas.AlertTemplate]:
...
+        name: Optional[str] = None,
+        since: Optional[datetime] = None,
+        until: Optional[datetime] = None,
+        entity: Optional[str] = None,
+        severity: Optional[list[str]] = None,
+        entity_kind: Optional[str] = None,
+        event_kind: Optional[str] = None,
+    ) -> list:
+        """
+        Retrieve list of all alert activations.
```
moranbental: missing docstring for each parameter

mlrun/db/httpdb.py
```diff
@@ -4915,6 +5007,46 @@ def _list_runs(
...
+        until: Optional[datetime] = None,
+        entity: Optional[str] = None,
+        severity: Optional[list[str]] = None,
+        entity_kind: Optional[str] = None,
+        event_kind: Optional[str] = None,
+        page: Optional[int] = None,
+        page_size: Optional[int] = None,
+        page_token: Optional[str] = None,
+        return_all: bool = False,
+    ):
```
moranbental:
```suggestion
    ) -> tuple[list, Optional[str]]:
```

server/py/services/api/api/endpoints/alert_activations.py
```diff
@@ -0,0 +1,90 @@
...
+@router.get("/projects/{project}/alert-activations")
+async def list_alert_activations(
+    project: str,
+    name: Optional[str] = None,
+    since: Optional[str] = None,
+    until: Optional[str] = None,
+    entity: Optional[str] = None,
+    severity: Optional[list[str]] = None,
+    entity_kind: Optional[str] = None,
+    event_kind: Optional[str] = None,
```
moranbental: Also, verify the entity_kind in the system test to ensure it is covered
```suggestion
    entity_kind: Optional[str] = Query(None, alias="entity-kind"),
    event_kind: Optional[str] = Query(None, alias="event-kind"),
```

tests/system/alerts/test_alerts.py
```diff
@@ -84,6 +85,57 @@ def test_job_failure_alert(self):
...
+        entities.remove(activations[0]["entity_id"])
+
+        activations, token = self._run_db.paginated_list_alert_activations(
+            project=self.project_name,
+            name=alert_name,
+            severity=[
+                alert_objects.AlertSeverity.LOW,
+                alert_objects.AlertSeverity.HIGH,
+            ],
+            entity=f"~{run_id}*",
```
moranbental: There's no need to resend all these filters, just the token

tests/system/alerts/test_alerts.py
```diff
@@ -84,6 +85,57 @@ def test_job_failure_alert(self):
...
+            page_size=1,
+        )
+        assert len(activations) == 1
```
moranbental, line 127: Also, verify that the token **is not** None and that you have the first alert activation object, not the second

tests/system/alerts/test_alerts.py
```diff
@@ -84,6 +85,57 @@ def test_job_failure_alert(self):
...
+            project=self.project_name,
+            name=alert_name,
+            severity=[
+                alert_objects.AlertSeverity.LOW,
+                alert_objects.AlertSeverity.HIGH,
+            ],
+            entity=f"~{run_id}*",
+            page_token=token,
+        )
+        assert len(activations) == 1
```
moranbental, line 134: Also, verify that the token **is** None and that you have the second alert activation object, not the first

server/py/framework/db/sqldb/db.py
```diff
@@ -5929,13 +5930,62 @@ def store_alert_activation(
...
+        if severity:
+            query = query.filter(AlertActivation.severity.in_(severity))
+
+        if event_kind:
+            query = query.filter(AlertActivation.event_kind == event_kind)
+
+        if entity_kind:
+            query = query.filter(AlertActivation.entity_kind == entity_kind)
+
+        query = self._paginate_query(query, page, page_size)
```
moranbental, line 5991: Before running the pagination, shouldn't it be ordered by `activation_time`?

moranbental (changes requested):

mlrun/common/schemas/auth.py
rokatyy: @moranbental the idea behind here is that user should have access to the alert config on which this activation is. So everything is correct here

tests/system/alerts/test_alerts.py
rokatyy: @moranbental it is already verified by deletion logic from entities dict

tests/system/alerts/test_alerts.py
rokatyy: @moranbental it is already verified by deletion logic from entities dict

server/py/framework/db/sqldb/db.py
rokatyy: @moranbental they by default sorted by ascending order, but we indeed need it in  descending order, so will add

mlrun/db/httpdb.py
```diff
@@ -4666,6 +4666,108 @@ def list_alert_templates(self) -> list[mlrun.common.schemas.AlertTemplate]:
...
+    def list_alert_activations(
+        self,
+        project: Optional[str] = None,
+        name: Optional[str] = None,
+        since: Optional[datetime] = None,
+        until: Optional[datetime] = None,
+        entity: Optional[str] = None,
+        severity: Optional[list[str]] = None,
+        entity_kind: Optional[str] = None,
+        event_kind: Optional[str] = None,
```
moranbental: It seems like there should be enums here. Are there enums for these parameters (severity, entity_kind, event_kind)?

tests/system/alerts/test_alerts.py
moranbental: @rokatyy How do you verify that you got this page by descending order of activation_time?

moranbental (changes requested):
minor

mlrun/db/httpdb.py
Yacouby: sure, all the enums for these params are present in: `mlrun/common/schemas/alert.py`

mlrun/db/httpdb.py
```diff
@@ -4666,6 +4666,108 @@ def list_alert_templates(self) -> list[mlrun.common.schemas.AlertTemplate]:
...
+    ) -> list:
```
Yacouby:
```suggestion
    ) -> list[mlrun.common.schemas.AlertActivation]:
```

mlrun/db/httpdb.py
```diff
@@ -4666,6 +4666,108 @@ def list_alert_templates(self) -> list[mlrun.common.schemas.AlertTemplate]:
             results.append(mlrun.common.schemas.AlertTemplate(**item))
         return results
 
+    def list_alert_activations(
+        self,
+        project: Optional[str] = None,
+        name: Optional[str] = None,
+        since: Optional[datetime] = None,
+        until: Optional[datetime] = None,
+        entity: Optional[str] = None,
```
Yacouby, line 4675: since we have entity_kind here and can be misleading, lets change this to entity_id

mlrun/db/httpdb.py
```diff
@@ -4666,6 +4666,108 @@ def list_alert_templates(self) -> list[mlrun.common.schemas.AlertTemplate]:
...
+        return alert_activations
+
+    def paginated_list_alert_activations(
+        self,
+        *args,
+        page: Optional[int] = None,
+        page_size: Optional[int] = None,
+        page_token: Optional[str] = None,
+        **kwargs,
+    ) -> tuple[list, Optional[str]]:
```
Yacouby, line 4719:
```suggestion
    ) -> tuple[list[mlrun.common.schemas.AlertActivation], Optional[str]]:
```

mlrun/db/httpdb.py
```diff
@@ -4915,6 +5017,46 @@ def _list_runs(
...
+        until: Optional[datetime] = None,
+        entity: Optional[str] = None,
+        severity: Optional[list[str]] = None,
+        entity_kind: Optional[str] = None,
+        event_kind: Optional[str] = None,
+        page: Optional[int] = None,
+        page_size: Optional[int] = None,
+        page_token: Optional[str] = None,
+        return_all: bool = False,
+    ) -> tuple[list, Optional[str]]:
```
Yacouby:
```suggestion
    ) -> tuple[list[mlrun.common.schemas.AlertActivation], Optional[str]]:
```

mlrun/db/nopdb.py
```diff
@@ -865,3 +865,26 @@ def get_alert_template(self, template_name: str):
...
+    def list_alert_activations(
+        self,
+        project: Optional[str] = None,
+        name: Optional[str] = None,
+        since: Optional[datetime.datetime] = None,
+        until: Optional[datetime.datetime] = None,
+        entity: Optional[str] = None,
+        severity: Optional[list[str]] = None,
+        entity_kind: Optional[str] = None,
+        event_kind: Optional[str] = None,
```
Yacouby: use enums for severity, entity_kind, event_kind

server/py/framework/db/base.py
```diff
@@ -967,8 +967,20 @@ def store_alert_activation(
...
+        projects_with_creation_time: list[tuple[str, datetime.datetime]],
+        name: Optional[str] = None,
+        since: Optional[str] = None,
+        until: Optional[str] = None,
+        entity: Optional[str] = None,
+        severity: Optional[list[str]] = None,
+        entity_kind: Optional[str] = None,
+        event_kind: Optional[str] = None,
+        page: typing.Optional[int] = None,
+        page_size: typing.Optional[int] = None,
```
Yacouby, line 987: use enums for severity, entity_kind, event_kind

server/py/framework/db/sqldb/db.py
```diff
@@ -5929,14 +5930,63 @@ def store_alert_activation(
...
+        projects_with_creation_time: list[tuple[str, datetime]],
+        name: typing.Optional[str] = None,
+        since: typing.Optional[str] = None,
+        until: typing.Optional[str] = None,
+        entity: typing.Optional[str] = None,
+        severity: typing.Optional[list[str]] = None,
+        entity_kind: typing.Optional[str] = None,
+        event_kind: typing.Optional[str] = None,
+        page: typing.Optional[int] = None,
+        page_size: typing.Optional[int] = None,
```
Yacouby, line 5948: use enums for severity, entity_kind, event_kind

tests/system/alerts/test_alerts.py
rokatyy: @moranbental will add a check to the test

server/py/framework/db/sqldb/helpers.py
```diff
@@ -87,6 +89,49 @@ def generate_query_predicate_for_name(column, query_string):
...
+
+        # For a direct equality query
+        query_string = "test"
+        condition = generate_query_for_name_with_wildcard(
+            my_table.column_name, query_string
+        )
+        # condition evaluates to: column_name == "test"
+        ```
+    """
+
```
Yacouby, line 147: can the query_string be empty ? or starts with ~ but empty ?

moranbental (approved):
Great!

server/py/framework/rundb/sqldb.py
```diff
@@ -1271,6 +1271,29 @@ def get_alert_template(self, template_name: str):
...
+        self,
+        project: Optional[str] = None,
+        name: Optional[str] = None,
+        since: Optional[datetime.datetime] = None,
+        until: Optional[datetime.datetime] = None,
+        entity: Optional[str] = None,
+        severity: Optional[list[str]] = None,
+        entity_kind: Optional[str] = None,
+        event_kind: Optional[str] = None,
+    ):
```
Yacouby, line 1284: use enums for severity, entity_kind, event_kind

server/py/services/api/api/endpoints/alert_activations.py
```diff
@@ -0,0 +1,90 @@
...
+async def list_alert_activations(
+    project: str,
+    name: Optional[str] = None,
+    since: Optional[str] = None,
+    until: Optional[str] = None,
+    entity: Optional[str] = None,
+    severity: Optional[list[str]] = None,
+    entity_kind: Optional[str] = Query(None, alias="entity-kind"),
+    event_kind: Optional[str] = Query(None, alias="event-kind"),
+    page: int = Query(None, gt=0),
```
Yacouby, line 49: use enums for severity, entity_kind, event_kind

server/py/services/api/crud/alert_activation.py
```diff
@@ -94,3 +96,31 @@ def _prepare_notifications_states(
...
+        self,
+        session: sqlalchemy.orm.Session,
+        projects_with_creation_time: list[tuple[str, datetime.datetime]],
+        name: Optional[str] = None,
+        since: Optional[datetime.datetime] = None,
+        until: Optional[datetime.datetime] = None,
+        entity: Optional[str] = None,
+        severity: Optional[list[str]] = None,
+        entity_kind: Optional[str] = None,
+        event_kind: Optional[str] = None,
```
Yacouby: enums

tests/system/alerts/test_alerts.py
```diff
@@ -84,6 +85,58 @@ def test_job_failure_alert(self):
...
+            name=alert_name,
+            severity=[
+                alert_objects.AlertSeverity.LOW,
+                alert_objects.AlertSeverity.HIGH,
+            ],
+            entity=f"~{run_id}*",
+            entity_kind=alert_objects.EventEntityKind.JOB,
+            event_kind=alert_objects.EventKind.FAILED,
+        )
+        assert len(activations) == 2
```
Yacouby, line 100: why do we have two activations here ? 
according to the test the alert should be triggered once

tests/system/alerts/test_alerts.py
```diff
@@ -84,6 +85,58 @@ def test_job_failure_alert(self):
         self._validate_notifications_on_nuclio(
             nuclio_function_url, expected_notifications
         )
+        # get all activations
```
Yacouby, line 88: i would move all of this to a separate function ( _validate_alert_activations ? )
so the test will not be so long and if we want to add more checks to it in the future it would be more readable

Yacouby (changes requested):
Looks good, added some comments.
more improvements to consider:
- I think it would be better to add an integration test for this flow.. Since system tests don’t run in CI, adding an integration test will ensure that this functionality is continuously tested and covered.
- what about checking the id column in the alert activations table ? to see it is auto incrementing after adding each entry

tests/system/alerts/test_alerts.py
rokatyy: @Yacouby because I have changed the test and now I run it twice

tests/system/alerts/test_alerts.py
rokatyy: @Yacouby what exactly do you want to move here? Every check here is specific to this test

rokatyy:
@Yacouby 

> I think it would be better to add an integration test for this flow.. Since system tests don’t run in CI, adding an integration test will ensure that this functionality is continuously tested and covered.

test is there already, for all other tests, I'd do a big test in a separate PR (when we have all SDKs) as we discussed before

> what about checking the id column in the alert activations table ? to see it is auto incrementing after adding each entry
We don't return id anywhere and I've checked this already on an actual cluster.

server/py/framework/db/sqldb/helpers.py
rokatyy: @Yacouby if it's empty, we won't get there at all. If it's ~ and empty, we will just request all the alerts

mlrun/db/httpdb.py
rokatyy: @Yacouby in API we have already confirmed `entity` as a name and also I think that it's more user friendly, so I would not change that.

mlrun/db/base.py
```diff
@@ -845,6 +845,35 @@ def get_alert_template(self, template_name: str):
...
+        since: Optional[datetime.datetime] = None,
+        until: Optional[datetime.datetime] = None,
+        entity: Optional[str] = None,
+        severity: Optional[
+            list[Union[mlrun.common.schemas.alert.AlertSeverity, str]]
+        ] = None,
+        entity_kind: Optional[
+            Union[mlrun.common.schemas.alert.EventEntityKind, str]
+        ] = None,
+        event_kind: Optional[Union[mlrun.common.schemas.alert.EventKind, str]] = None,
```
TomerShor, line 862: Why also allowing general `str`?

server/py/framework/db/sqldb/db.py
```diff
@@ -5929,14 +5930,67 @@ def store_alert_activation(
...
     ) -> list[mlrun.common.schemas.AlertActivation]:
-        # TODO: add filters
         query = self._query(session, AlertActivation)
 
-        if project and project != "*":
-            query = query.filter(AlertActivation.project == project)
+        # Filter alert activations for the project created after the project creation date,
+        # excluding activations linked to any previous instances of the project.
+        # TODO: reconsider this approach when we move alerts out of main MLRun db
+        conditions = [
```
TomerShor: Just for explicitness
```suggestion
        project_filter_conditions = [
```

server/py/framework/db/sqldb/db.py
```diff
@@ -5929,14 +5930,67 @@ def store_alert_activation(
...
+
+        if since or until:
+            since = since or datetime.min
+            until = until or datetime.max
+            query = query.filter(
+                and_(
+                    AlertActivation.activation_time >= since,
+                    AlertActivation.activation_time <= until,
+                )
+            )
```
TomerShor: this is duplicated in 2 other places (Functions and Artifacts), can be made a common function

TomerShor (changes requested):

mlrun/db/base.py
rokatyy: @TomerShor I noticed that it's a general rule in httpdb to support either dict/str or pydantic entity/enum, so followed the same pattern

server/py/framework/db/sqldb/db.py
rokatyy: @TomerShor not sure how it can be generalised when model's and column's names are different

TomerShor: Something like this (need to test):
```python
    def _attach_time_range_query(self, query, cls, field_name, since, until):
        since = since or datetime.min
        until = until or datetime.max

        field = getattr(cls, field_name)
        return query.filter(and_(field >= since, field <= until))
```

rokatyy: Good idea!

TomerShor (approved):
Nicely done

Yacouby (approved):
Nice job !

PR #6725, opened 2024-11-19
[Alerts] Fix alert state count reset when storing existing alerts

Yacouby:
- Fixed an issue where storing an existing alert did not reset its state count correctly.
- Renamed the method `get_enriched_alert` to `get_alert` to improve clarity and reduce confusion.
- Removed unused `get_alert` method from crud.

Jira:
https://iguazio.atlassian.net/browse/ML-8315

### DIFF
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -5568,6 +5568,7 @@ class SQLDB(DBInterface):
             return self._create_alert(session, alert)
         alert_record.full_object = alert.dict()
         alert_state = self.get_alert_state(session, alert_record.id)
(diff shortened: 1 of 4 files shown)


server/py/framework/db/sqldb/db.py
```diff
@@ -5568,6 +5568,7 @@ def store_alert(
             return self._create_alert(session, alert)
         alert_record.full_object = alert.dict()
         alert_state = self.get_alert_state(session, alert_record.id)
+        alert_state.count = 0
```
rokatyy, line 5571: Does it mean that count will be zero-ed if I update alert config's description?

PR #6727, opened 2024-11-19
[KFP] Convert modifier functions to modify the runtime spec directly

adamdelman:
**Summary**:
This PR removes dependencies on KFP by refactoring modifier functions to directly modify the runtime specification.

*Key Changes:*

- Refactored `mlrun.runtimes.pod` to modify runtime specs directly.
- Updated `mlrun.runtimes.mounts` and `mlrun.runtimes.mixins` to align with the new approach.
- Removed KFP-related `ContainerOp` logic and dependencies.

https://iguazio.atlassian.net/browse/ML-6118
https://iguazio.atlassian.net/browse/ML-8275


pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/mounts.py
```diff
@@ -39,29 +43,25 @@ def v3io_cred(api="", user="", access_key=""):
...
-            )
-            .add_env_variable(k8s_client.V1EnvVar(name="V3IO_USERNAME", value=_user))
-            .add_env_variable(
-                k8s_client.V1EnvVar(name="V3IO_ACCESS_KEY", value=_access_key)
-            )
-            .add_env_variable(
-                k8s_client.V1EnvVar(name="V3IO_FRAMESD", value=v3io_framesd)
-            )
+        runtime.spec.env.extend(
+            [
```
liranbg: does add_env_variable handling duplicates?

liranbg (commented):
overall looking good. left a question for the implementation

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/mounts.py
adamdelman: no, it simply appends to the env list

liranbg (approved):

mlrun/runtimes/pod.py
```diff
@@ -412,6 +413,26 @@ def _verify_and_set_limits(
                         resource_value,
                     )
 
+    def add_volumes(self, volumes):
```
alonmr: A few comments here since this is now user facing:
1. Add type hints
2. Move above `_set_volume_mount` to maintain reading order
3. Suggest to rename - we usually use `set_...` or `with_...` I think set is more fitting and that we should keep some conventions here to make it easier to find.
4. Expand the docstring - IIUC users shouldn't really be using this and this is mainly for `auto_mount` usage

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/mounts.py
```diff
@@ -19,7 +19,9 @@
 VolumeMount = namedtuple("Mount", ["path", "sub_path"])
 
 
-def _enrich_and_validate_v3io_mounts(remote="", volume_mounts=None, user=""):
+def _enrich_and_validate_v3io_mounts(
```
alonmr: please add type hints

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/mounts.py
```diff
@@ -39,29 +43,25 @@ def v3io_cred(api="", user="", access_key=""):
...
+                {"name": "V3IO_API", "value": web_api},
+                {"name": "V3IO_USERNAME", "value": _user},
+                {"name": "V3IO_ACCESS_KEY", "value": _access_key},
```
alonmr: This potentially exposes the access key on the runtime spec. I don't think it's a good idea to add it as plain text. I'm not sure why not keep this enrichment on the kfp object?

alonmr (commented):
Partial review 😬

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/mounts.py
adamdelman: I think this will only be exposed if running locally, when the workflow is submitted to the MLRun API, the secret is replaced with a reference to a K8S secret.

alonmr: Right but my point is that after you compile the workflow, the runtime now has the access key as plaintext in its spec on the client side. We should be OK because of the implementation of `mlrun.runtimes.pod.KubeResourceSpec._enrich_field` which is used on `to_dict` - but IMO it's not good practice to allow this

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/mounts.py
```diff
@@ -144,52 +151,65 @@ def add_vol(name, mount_path, host_path):
         igz_version = mlconf.get_parsed_igz_version()
         if igz_version and igz_version >= semver.VersionInfo.parse("3.2.3-b1"):
```
alonmr: 3.2.3 is dead this can be removed
```suggestion
        if igz_version:
```

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/mounts.py
```diff
@@ -234,65 +254,63 @@ def mount_s3(
...
         logger.warning(
-            "it is recommended to use k8s secret (specify secret_name), "
-            "specifying the aws_access_key/aws_secret_key directly is unsafe"
+            "It is recommended to use k8s secret (specify secret_name), "
+            "specifying aws_access_key/aws_secret_key directly is unsafe."
         )
 
-    def _use_s3_cred(container_op):
+    def _use_s3_cred(runtime: "KubeResource"):
         from os import environ
```
alonmr: not sure why this import is here, can be moved to the top

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/mounts.py
```diff
@@ -465,13 +528,10 @@ def set_env_variables(env_vars_dict: Optional[dict[str, str]] = None, **kwargs):
     for key, value in kwargs.items():
         env_data[key] = value
 
-    def _set_env_variables(container_op: kfp.dsl.ContainerOp):
-        from kubernetes import client as k8s_client
+    def _set_env_variables(runtime: "KubeResource"):
+        for key, value in env_data.items():
+            runtime.spec.env.append({"name": key, "value": value})
```
alonmr: why not use `runtime.set_envs(env_data)` ?

alonmr (commented):
Minor comments

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/mounts.py
```diff
@@ -259,24 +250,16 @@ def mount_s3(
...
-        _endpoint_url = endpoint_url or environ.get(prefix + "S3_ENDPOINT_URL")
+        _access_key = aws_access_key or os.environ.get(prefix + "AWS_ACCESS_KEY_ID")
+        _secret_key = aws_secret_key or os.environ.get(prefix + "AWS_SECRET_ACCESS_KEY")
+        _endpoint_url = endpoint_url or os.environ.get(prefix + "S3_ENDPOINT_URL")
 
         if _endpoint_url:
-            runtime.spec.env.append(
-                {"name": prefix + "S3_ENDPOINT_URL", "value": _endpoint_url}
-            )
+            runtime.set_envs({prefix + "S3_ENDPOINT_URL": _endpoint_url})
```
alonmr: can use `set_env` when adding a single var

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/mounts.py
adamdelman: This behavior has not changed, previously (in `mlrun_pipelines.utils.apply_kfp`), we we would create a ContainerOp, fill it with all the relevant env vars, volumes and volumemounts, and then copy a serialised version (through k8s client) back into the runtime.

server/py/services/api/tests/unit/runtimes/test_dask.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+# Standard library imports
```
alonmr:
```suggestion
```

mlrun/runtimes/mounts.py
```diff
@@ -0,0 +1,574 @@
...
+from mlrun.utils import logger
+
+if typing.TYPE_CHECKING:
+    from mlrun.runtimes import KubeResource
+
+
+VolumeMount = namedtuple("Mount", ["path", "sub_path"])
+
+
+def _enrich_and_validate_v3io_mounts(
```
alonmr: please move private methods below public ones

mlrun/runtimes/mounts.py
```diff
@@ -0,0 +1,574 @@
...
+                    "name": name,
+                }
+            )
+
+        # this is a legacy path for the daemon shared memory
+        host_path = "/dev/shm/"
+
+        # path to shared memory for daemon was changed in Iguazio 3.2.3-b1
+        igz_version = mlconf.get_parsed_igz_version()
+        if igz_version:
```
alonmr: actually all of this is not needed. we can assume iguazio and iguazio version >= 3.5 here

mlrun/runtimes/mounts.py
```diff
@@ -0,0 +1,574 @@
...
+        return mount_pvc(
+            pvc_name=pvc_name,
+            volume_mount_path=volume_mount_path,
+            volume_name=volume_name or "shared-persistency",
+        )
+    if "MLRUN_PVC_MOUNT" in os.environ:
+        return mount_pvc(
+            volume_name=volume_name or "shared-persistency",
+        )
+    # In the case of MLRun-kit when working remotely, no env variables will be defined but auto-mount
```
alonmr:
```suggestion
    # In the case of CE when working remotely, no env variables will be defined but auto-mount
```

mlrun/runtimes/mounts.py
```diff
@@ -0,0 +1,574 @@
...
+    # parameters may still be declared - use them in that case.
+    if config.storage.auto_mount_type == "pvc":
+        return mount_pvc(**config.get_storage_auto_mount_params())
+    if "V3IO_ACCESS_KEY" in os.environ:
+        return mount_v3io(name=volume_name or "v3io")
+
+    raise ValueError("failed to auto mount, need to set env vars")
```
alonmr:
```suggestion
    raise ValueError("Failed to auto mount, need to set env vars")
```

mlrun/runtimes/mounts.py
```diff
@@ -0,0 +1,574 @@
...
+    mount_path: str,
+    volume_name: str = "secret",
+    items: typing.Optional[list[dict]] = None,
+) -> typing.Callable[["KubeResource"], "KubeResource"]:
+    """Modifier function to mount a Kubernetes secret as file(s).
+
+    :param secret_name: Kubernetes secret name
+    :param mount_path: Path inside the container to mount
+    :param volume_name: Unique volume name
+    :param items:        If unspecified, each key-value pair in the Data field
```
alonmr, line 388: the indentation here is weird

mlrun/runtimes/mounts.py
```diff
@@ -0,0 +1,574 @@
...
+    mount_path: str,
+    volume_name: str = "configmap",
+    items: typing.Optional[list[dict]] = None,
+) -> typing.Callable[["KubeResource"], "KubeResource"]:
+    """Modifier function to mount a Kubernetes ConfigMap as file(s).
+
+    :param configmap_name: Kubernetes ConfigMap name
+    :param mount_path: Path inside the container to mount
+    :param volume_name: Unique volume name
+    :param items:           If unspecified, each key-value pair in the Data field
```
alonmr, line 436: same

alonmr (commented):
Looks great! some small nitpicking

alonmr (approved):

PR #6729, opened 2024-11-19
[Microservices] Enable hydra opt-out

alonmr:
Added minimal configuration to determine wether the API should mount alerts service as well or if it is standalone.
In order to enable running the initial services in a single instance as before, we are creating a dummy hydra which is consolidated in the api. The end goal is to create a hydra daemon and service controller which will be able to manage multiple services in a single instance. Once that is supported, the API will not mount any applications by itself.

### DIFF
--- a/server/py/services/alerts/main.py
+++ b/server/py/services/alerts/main.py
@@ -18,6 +18,7 @@ import fastapi
 import semver
 import sqlalchemy.orm
 from fastapi.concurrency import run_in_threadpool
+from starlette.responses import Response
 
 import mlrun.common.runtimes.constants
 import mlrun.common.schemas
@@ -44,10 +45,6 @@ from framework.routers import alert_template, alerts, auth, healthz
 
 
 class Service(framework.service.Service):
-    # TODO: Change service name to alerts once they are fully separated - this allows to mount the application on the
-    #  api Router without implementing tunneling
-    service_name = "api"
-
     async def store_alert(
         self,
         request: fastapi.Request,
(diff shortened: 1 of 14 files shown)


server/py/services/alerts/main.py
```diff
@@ -506,3 +475,11 @@ def _generate_event_on_failed_runs(
                 project=project,
                 validate_event=True,
             )
+
+    @staticmethod
+    def _is_chief_or_standalone():
+        return (
```
liranbg, line 490: add a comment explaining the parts on the "or"

server/py/services/alerts/tests/unit/conftest.py
```diff
@@ -34,9 +36,12 @@
 
 @pytest.fixture()
 def app() -> fastapi.FastAPI:
+    old_service_name = mlconf.services.service_name
+    mlconf.services.service_name = "alert"
     yield services.alerts.daemon.app()
+    mlconf.services.service_name = old_service_name
```
liranbg:
```suggestion
    try:
      mlconf.services.service_name = "alert"
      yield services.alerts.daemon.app()
    finally:
      mlconf.services.service_name = old_service_name
```

server/py/services/api/daemon.py
```diff
@@ -21,13 +23,12 @@
...
-        self._service: framework.service.Service = service_cls()
-
     @property
     def mounts(self) -> dict[str, framework.service.Service]:
-        # Mount the alerts application until we have service routing/tunneling
-        return {"/": alerts_daemon.service}
+        if mlconf.services.hydra.services == "*":
+            # Mount the alerts application until we have proper hydra
+            return {"/": alerts_daemon.service}
+        return {}
```
liranbg: return list of services instead of prefix per service
in hydra, later on, once adding a service you would need to return more than a single service instance

moranbental (approved):

liranbg (approved):

PR #6735, opened 2024-11-20
[Tests] Moving test_httpdb.py to integration tests

moranbental:


tests/rundb/test_unit_httpdb.py
alonmr, line 1: can rename to `test_httpdb.py`

alonmr (approved):
Generally looks good as long as CI passes but I don't really see a reason to keep these integration tests fixtures.
The docker fixture is an old version of `TestMLRunIntegration` framework and the server fixture is a nice addition but not sure if it is needed. I would rather have all tests using a common framework.

PR #6738, opened 2024-11-20
[Alert activation] Add project index in alert_activation

rokatyy:
Adding this index improves listing performance


server/py/framework/db/sqldb/models.py
```diff
@@ -855,6 +855,7 @@ class AlertActivation(Base, mlrun.utils.db.BaseModel):
             PrimaryKeyConstraint("id", "activation_time", name="_alert_activation_uc"),
             Index("ix_alert_activation_project_name", "project", "name"),
```
TomerShor, line 856: Doesn't this ondez already cover the needed optimization?

rokatyy: @TomerShor A dedicated index on project is needed because the current composite index (ix_alert_activation_project_name) includes both project and name, which means it’s less efficient when we query using project alone. Now, after thinking about the PR #6723 , I realised that probably the best idea here is to remove single activation_time index and change it to composite index with project+activation time, because every time we do request, we request by both project and activation time

TomerShor: What about when we query for all projects?
I think the default query will be for the last day / week so indexing activation time might still make sense.

rokatyy: @TomerShor we never actually skip filtering by project, because even if `*` is passed, we request all the allowed project and filter by them. Thus, we never request activation_time alone without project.

TomerShor: Right, missed that.
Sounds good then 👍

TomerShor (approved):

PR #6742, opened 2024-11-20
[CI] Fix mlrun-api dockerfile cache layers

liranbg:
install requirement before adding rest of code (code . .) to avoid pip deps cache invalidation


adamdelman (approved):
lgtm

adamdelman (approved):

PR #6739, opened 2024-11-20
[Model Monitoring] Skip feature set update for router endpoints [1.7.x]

gtopper:
[ML-8316](https://iguazio.atlassian.net/browse/ML-8316)

[ML-8316]: https://iguazio.atlassian.net/browse/ML-8316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6747, opened 2024-11-21
[Model Monitoring] Skip feature set update for router endpoints

gtopper:
[ML-8316](https://iguazio.atlassian.net/browse/ML-8316)

Development version of #6739.

[ML-8316]: https://iguazio.atlassian.net/browse/ML-8316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/model_monitoring/stream_processing.py
+++ b/mlrun/model_monitoring/stream_processing.py
@@ -30,6 +30,7 @@ import mlrun.model_monitoring.db
 import mlrun.serving.states
 import mlrun.utils
 from mlrun.common.schemas.model_monitoring.constants import (
+    EndpointType,
     EventFieldType,
     EventKeyMetrics,
     EventLiveStats,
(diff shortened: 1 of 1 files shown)


mlrun/model_monitoring/stream_processing.py
```diff
@@ -783,6 +784,10 @@ def do(self, event: dict):
             if not feature_names and self._infer_columns_from_data:
                 feature_names = self._infer_feature_names_from_data(event)
 
+            is_router_endpoint = endpoint_record.get(
```
Eyal-Danieli: I suggest moving it inside the if clause (`if not feature_names: ...`) since it is only relevant to that specific case.
Additionally, on L852 we call `endpoint_record.get(EventFieldType.ENDPOINT_TYPE)` again. It would be nice to merge these calls to avoid redundancy, though this is more of a nice-to-have improvement.

gtopper: But it's also needed inside the `if not label_columns:` block. Anyway, it doesn't matter following refactor based on your second point. Please check again.

gtopper: Sadly though, now I have to verify it again 😬

gtopper: Notebook completed successfully.

Eyal-Danieli (approved):

PR #6751, opened 2024-11-22
[Alerts activations] Fix path to mlrun config for partitions_buffer_multiplier

rokatyy:
Path was broken


server/py/services/api/utils/db/partitioner.py
```diff
@@ -50,7 +50,7 @@ def create_and_drop_partitions(
         # Ensure partitions for the retention time plus a buffer
         partition_number = partition_interval.get_number_of_partitions(
             days=retention_days
-            + mlrun.config.partitions_buffer_multiplier
+            + mlrun.config.config.partitions_buffer_multiplier
```
TomerShor:
```suggestion
            + mlrun.mlconf.partitions_buffer_multiplier
```

TomerShor (changes requested):

liranbg (approved):

PR #6752, opened 2024-11-22
[Alert activation] Save alert activation for manual policy

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-8138

Within this task:

1. process_event function was refactored to improve readability
2. logic to save activation for manual policy was added
3. Improved compatibility of tests with SQLite with getting back the `autoincrement=True` 
4. logic to save `number_of_event` when resetting event in manual policy was added


server/py/framework/db/init_db.py
```diff
@@ -15,8 +15,17 @@
...
+    engine = get_engine()
+    if engine.name == "sqlite":
+        tables_to_create = [
+            table
+            for table in Base.metadata.tables.values()
+            if table.name != AlertActivation.__tablename__
+        ]
+        Base.metadata.create_all(bind=engine, tables=tables_to_create)
+    else:
+        Base.metadata.create_all(bind=engine)
```
TomerShor: 1. How does this allow running unit tests if we don't create the tables at all?
2. a more extendable version will be something like a function `get_sqlite_tables` that resolves the list. It will keep the `init_db` function leaner.

server/py/framework/db/sqldb/db.py
```diff
@@ -4691,6 +4691,16 @@ def _upsert(self, session, objects, ignore=False, silent=False):
             session.add(object_)
         self._commit(session, objects, ignore, silent)
 
+    def _upsert_object_and_flush_to_get_field(self, session, object_, field):
+        session.add(object_)  # Add the object to the session
+        session.flush()  # Flush the session to generate the database values
```
```suggestion
        # Add the object to the session
        session.add(object_)
        # Flush the session to generate the database values
        session.flush()
```

server/py/framework/db/sqldb/db.py
```diff
@@ -5927,7 +5937,30 @@ def store_alert_activation(
...
+        # if reset_policy is MANUAL, we need to keep id to be able to update number_of_events
+        # when policy is reset
+        if alert_data.reset_policy == mlrun.common.schemas.alert.ResetPolicy.MANUAL:
+            return self._upsert_object_and_flush_to_get_field(
+                session, alert_activation_record, "id"
+            )
+        else:
+            return self._upsert(session, [alert_activation_record])
+
+    def update_alert_activation(
```
```suggestion
    def update_alert_activation_number_of_events(
```

This is the only thing we allow updating

server/py/framework/db/sqldb/db.py
```diff
@@ -5927,7 +5937,30 @@ def store_alert_activation(
...
+        self,
+        session,
+        activation_id: int,
+        activation_time: datetime,
```
TomerShor, line 5955: Why is the activation time required if we have the record id?

server/py/framework/db/sqldb/models.py
```diff
@@ -861,7 +861,7 @@ class AlertActivation(Base, mlrun.utils.db.BaseModel):
             ),
         )
 
-        id = Column(Integer)
+        id = Column(Integer, autoincrement=True)
```
TomerShor, line 864: This was already handled in a migration. Won't re-adding this cause conflicts?

server/py/services/alerts/crud/alerts.py
```diff
@@ -175,76 +175,34 @@ def process_event(
...
-                    session,
-                    alert.project,
-                    alert.name,
-                    count=state["count"],
-                    last_updated=event_data.timestamp,
-                    obj=state_obj,
-                    active=active,
-                )
+        # Append event timestamp for active state or criteria
+        if state["active"] or alert.criteria is not None:
```
TomerShor: comment **Why** you check "or criteria"

server/py/services/alerts/crud/alerts.py
```diff
@@ -175,76 +175,34 @@ def process_event(
...
+            state_obj["events"].append(event_data.timestamp)
 
-            if update_state:
-                # we don't want to update the state if reset_alert() was called, as we will override the reset
-                self._states[alert.id] = state_obj
+        # Exit early if state is active (no further processing needed)
+        if state["active"]:
+            self._states[alert.id] = state_obj
+            return
```
TomerShor, line 190: So to clarify - the number of events on the activation is not incremented every event, but will stay the same until the alert is reset? even if more events are still coming in?
Won't it make more sense to update it on each event?

server/py/services/alerts/crud/alerts.py
```diff
@@ -260,6 +218,61 @@ def populate_event_cache(self, session: sqlalchemy.orm.Session):
         services.alerts.crud.Events().cache_initialized = True
         logger.debug("Finished populating event cache for alerts")
 
+    def should_send_notification(self, alert, state_obj):
```
TomerShor: can be a private method

server/py/services/alerts/crud/alerts.py
```diff
@@ -260,6 +218,61 @@ def populate_event_cache(self, session: sqlalchemy.orm.Session):
...
+                self._normalize_events(
+                    state_obj,
+                    framework.utils.helpers.string_to_timedelta(
+                        alert.criteria.period, offset, raise_on_error=False
+                    ),
+                )
+            return len(state_obj["events"]) >= alert.criteria.count
+        return True
+
+    def get_number_of_events(self, alert_id: int) -> int:
```
TomerShor: can be a private method

server/py/services/alerts/crud/alerts.py
```diff
@@ -260,6 +218,61 @@ def populate_event_cache(self, session: sqlalchemy.orm.Session):
...
+        state_obj = self._states.get(alert_id, {"events": []})
+        return len(state_obj["events"])
+
+    @staticmethod
+    def _get_event_offset(alert):
+        if alert.entities.kind == mlrun.common.schemas.alert.EventEntityKind.JOB:
+            return int(mlconfig.monitoring.runs.interval)
+        return 0
+
+    def _handle_notification(self, session, alert, state, state_obj, event_data):
```
TomerShor: add type hints

server/py/services/alerts/crud/alerts.py
```diff
@@ -260,6 +218,61 @@ def populate_event_cache(self, session: sqlalchemy.orm.Session):
...
+        framework.utils.singletons.db.get_db().store_alert_state(
+            session,
+            alert.project,
+            alert.name,
+            count=state["count"],
+            last_updated=event_data.timestamp,
+            obj=state_obj,
+            active=active,
+        )
+        return update_state
```
TomerShor: it looks wrong that `update_state` can be false but store the alert state anyway.
I understand that you use it outside for wether to update the state cache, but it's not clear what and why when reading this function.

server/py/services/alerts/crud/alerts.py
```diff
@@ -391,6 +404,27 @@ def reset_alert(self, session: sqlalchemy.orm.Session, project: str, name: str):
...
+                activation_id = alert_state.get("full_object", {}).get(
+                    "last_activation_id"
+                )
+                if activation_time and activation_id:
+                    framework.utils.singletons.db.get_db().update_alert_activation(
+                        session,
+                        activation_id=activation_id,
+                        activation_time=activation_time,
+                        number_of_events=number_of_events,
+                    )
```
TomerShor: Maybe we can do it in a transaction with the `store_alert_state` call in the line afterwards?
Trying to minimize some db calls..

TomerShor (changes requested):

server/py/framework/db/sqldb/models.py
rokatyy: @TomerShor the problem is that without this string, flush works weird and overall this string just shows to ORM that the row is autoincrementable. Also, we saw the warning in the logs because of this. So, this approach is more correct in terms of working with ORM. 

> Won't re-adding this cause conflicts?

No, it's completely alright

server/py/framework/db/sqldb/db.py
rokatyy: @TomerShor that's a good question. And it's indeed done on purpose. The thing is that in our case we have composite PK (id, activation_time) and if we don't pass activation time, MySQL will always go through all the partitions (even if id is passed). So we need it to improve performance.

server/py/services/alerts/crud/alerts.py
rokatyy: @TomerShor update state indicates whether internal `_states` should be updated... But yeah, looks confusing, maybe rename to keep_cache?

server/py/services/alerts/crud/alerts.py
rokatyy: @TomerShor it is done only in the cases where number_or_alerts is higher that count in criteria, so not really sure if it is worth it

server/py/framework/db/init_db.py
rokatyy: @TomerShor  the problem is that:

1. `SQLIte` doesn't support autoincrement on composite PK
2. our code considers that `id` is autoincrementable thus we don't generate id

as a result, it doens't make sense even to create alert_activations table within tests as we can do nothing with table. Before this PR, i tried to make a trick with no putting `autoincrememt`. But this approach is quite hacky for the real code itself. So I decided that it's better to affect the tests rather than the actual code. All the tests for partitioned table should be integration/system until we decide to move to mysql for unit tests.

server/py/services/alerts/crud/alerts.py
rokatyy: @TomerShor it will create more load on database. Not that it's not feasible, but for now we decided to do like this (please check @theSaarco comment in slack)

server/py/framework/db/sqldb/db.py
```diff
@@ -4691,6 +4691,18 @@ def _upsert(self, session, objects, ignore=False, silent=False):
             session.add(object_)
         self._commit(session, objects, ignore, silent)
 
+    def _upsert_object_and_flush_to_get_field(self, session, object_, field):
+        # Add the object to the session
+        session.add(object_)
+        # Flush the session to generate the database values
+        session.flush()
```
theSaarco, line 4698: Why do you need to explicitly use `flush()` here? Since you're committing to DB anyway later in this function, wouldn't doing the `getattr` after the `commit` behave exactly the same?

server/py/services/alerts/crud/alerts.py
theSaarco: Correct. We certainly want to avoid going to the DB on every event (after all, that's why we keep cache).

server/py/services/alerts/crud/alerts.py
```diff
@@ -391,6 +415,27 @@ def reset_alert(self, session: sqlalchemy.orm.Session, project: str, name: str):
...
+                    framework.utils.singletons.db.get_db().get_alert_state_dict(
+                        session, alert.id
+                    )
+                )
+                activation_time = alert_state.get("last_updated")
+                activation_id = alert_state.get("full_object", {}).get(
+                    "last_activation_id"
+                )
+                if activation_time and activation_id:
+                    framework.utils.singletons.db.get_db().update_alert_activation_number_of_events(
```
theSaarco, line 425: Don't we store in the DB also the reset time? We know when it was activated, but we also want to know when it was reset, don't we?

server/py/framework/db/init_db.py
theSaarco: I tend to agree with @TomerShor in that I'd maintain in the `models.py` file a method called `get_partitioned_table_names()` or something - just so it's co-located with the tables themselves. This code will check that the table name is not in the returned list. This way, a developer that makes another partitioned table only needs to update that function, and doesn't need to be aware that this code here also needs to change.

theSaarco (changes requested):
Looks good - some comments and suggestions.

server/py/framework/db/init_db.py
```diff
@@ -15,8 +15,17 @@
 
 from mlrun.common.db.sql_session import get_engine
 
-from framework.db.sqldb.models import Base
+from framework.db.sqldb.models import AlertActivation, Base
```
Yacouby: we tend to not use imports this way:
https://github.com/mlrun/mlrun/blob/development/CONTRIBUTING.md#:~:text=When%20importing%20from%20local%20code%2C%20NEVER%20use%3A%20from%20X%20import%20Y%20instead%20use%3A%20import%20X%20or%20import%20X%20as%20Y.%20For%20external%20packages%2C%20it%27s%20acceptable%20to%20use%20from%20X%20import%20Y%20since%20they%20won%27t%20try%20to%20import%20back%20to%20our%20code

server/py/framework/db/sqldb/db.py
```diff
@@ -5927,7 +5939,30 @@ def store_alert_activation(
...
 
-        self._upsert(session, [alert_activation_record])
+        # if reset_policy is MANUAL, we need to keep id to be able to update number_of_events
+        # when policy is reset
+        if alert_data.reset_policy == mlrun.common.schemas.alert.ResetPolicy.MANUAL:
+            return self._upsert_object_and_flush_to_get_field(
+                session, alert_activation_record, "id"
+            )
+        else:
+            return self._upsert(session, [alert_activation_record])
```
Yacouby: in case of "auto" reset policy, is this return needed ?

server/py/services/alerts/crud/alerts.py
```diff
@@ -175,76 +175,36 @@ def process_event(
...
-                    count=state["count"],
-                    last_updated=event_data.timestamp,
-                    obj=state_obj,
-                    active=active,
-                )
+        # Append the event timestamp if the alert is active or has criteria.
+        # If the alert is active, it follows a manual reset policy, so we need to continue counting events.
+        # If the criteria is None and the alert is inactive, we trigger the alert immediately and do not cache events.
+        if state["active"] or alert.criteria is not None:
+            state_obj["events"].append(event_data.timestamp)
```
Yacouby:
```suggestion
            state_obj["events_timestamps"].append(event_data.timestamp)
```
this will make the state_obj more understandable

server/py/services/alerts/crud/alerts.py
```diff
@@ -175,76 +175,36 @@ def process_event(
...
 
-            if update_state:
-                # we don't want to update the state if reset_alert() was called, as we will override the reset
-                self._states[alert.id] = state_obj
+        # Exit early if state is active (no further processing needed)
+        if state["active"]:
+            self._states[alert.id] = state_obj
+            return
+
+        # Check notification criteria and send notification if needed
```
Yacouby: there is no such thing "notification criteria"

server/py/services/alerts/crud/alerts.py
```diff
@@ -391,6 +415,27 @@ def reset_alert(self, session: sqlalchemy.orm.Session, project: str, name: str):
                 f"Alert {name} for project {project} does not exist"
             )
 
+        if alert.reset_policy == mlrun.common.schemas.alert.ResetPolicy.MANUAL:
+            # before resetting an alert, update number_of_events if required
+            number_of_events = self._get_number_of_events(alert.id)
+            if number_of_events > alert.criteria.count:
```
Yacouby, line 414: what if number_of_events == alert.criteria.count ?

tests/integration/sdk_api/alerts/test_alerts.py
```diff
@@ -674,6 +732,45 @@ def _validate_alert(
...
+        project_name=None,
+        alert_name=None,
+        alert_event_name=None,
+        alert_severity=None,
+        alert_criteria=None,
+        alert_entity_id=None,
+        alert_entity_kind=None,
+        alert_event_kind=None,
+        number_of_events=None,
+        alert_notifications=None,
```
Yacouby, line 747: in your validations you only pass to this method: alert_activation, project_name, alert_name & number of events, 
so the rest of the params here are redundant.
also, this seems like a code duplication of the _validate_alert method here in the same test, so why not combine them and have one validator ? ( either pass it the alert_data or the alert_activation )

Yacouby (commented):
Looks good, minor comments

server/py/services/alerts/crud/alerts.py
rokatyy: @Yacouby then this number is already there

server/py/services/alerts/crud/alerts.py
rokatyy: @theSaarco no, we don't as it wasn't it the scope originally. We can discuss it and implement as a separate ticket

tests/integration/sdk_api/alerts/test_alerts.py
rokatyy: @Yacouby I think it's good to have them separated as even though some fields do match, some of them don't, so it will decrease readability, I'd prefer it as is.

server/py/framework/db/sqldb/db.py
rokatyy: will remove

server/py/framework/db/sqldb/db.py
rokatyy: @theSaarco yeah, that's what I expected, but after checking it - it didn't work for some reason(`id` was empty after committing )

server/py/services/alerts/crud/alerts.py
rokatyy: agree!

server/py/framework/db/init_db.py
rokatyy: @theSaarco Added the function and updated CONTRIBUTION.md with instructions.

server/py/framework/db/sqldb/db.py
```diff
@@ -5927,7 +5939,30 @@ def store_alert_activation(
             data=extra_data,
         )
 
-        self._upsert(session, [alert_activation_record])
+        # if reset_policy is MANUAL, we need to keep id to be able to update number_of_events
+        # when policy is reset
```
```suggestion
        # when the alert is reset
```

server/py/framework/db/sqldb/db.py
```diff
@@ -5927,7 +5939,30 @@ def store_alert_activation(
...
+        activation_id: int,
+        activation_time: datetime,
+        number_of_events: int,
+    ):
+        query = self._query(
+            session, AlertActivation, id=activation_id, activation_time=activation_time
+        )
+        activation = query.one_or_none()
+        if not activation:
+            raise ValueError(f"Activation with id {activation_id} not found")
```
TomerShor: Easier to cmd+F 😉 
```suggestion
            raise mlrun.errors.MLRunNotFoundError(f"Alert activation not found for id: {activation_id}")
```

server/py/services/alerts/crud/alerts.py
```diff
@@ -175,76 +175,35 @@ def process_event(
...
-            if update_state:
-                # we don't want to update the state if reset_alert() was called, as we will override the reset
-                self._states[alert.id] = state_obj
+        # Exit early if state is active (no further processing needed)
+        if state["active"]:
+            self._states[alert.id] = state_obj
+            return
+
+        send_notification = self._should_send_notification(alert, state_obj)
+        update_state = True
```
```suggestion
        update_state_cache = True
```

TomerShor (changes requested):
Much better! Last few comments

server/py/services/alerts/crud/alerts.py
```diff
@@ -175,76 +175,35 @@ def process_event(
...
-                    alert.name,
-                    count=state["count"],
-                    last_updated=event_data.timestamp,
-                    obj=state_obj,
-                    active=active,
-                )
+        # Append the event timestamp if the alert is active or has criteria.
+        # If the alert is active, it follows a manual reset policy, so we need to continue counting events.
+        # If the criteria is None and the alert is inactive, we trigger the alert immediately and do not cache events.
+        if state["active"] or alert.criteria is not None:
```
Yacouby: I'm looking at the alert.criteria field, and looks like it never can be None, 
since when storing the alert, if the criteria is not defined, we give it a default value ( with count=1 & period=None ), so looks like the check here is redundant.

server/py/services/alerts/crud/alerts.py
```diff
@@ -260,6 +219,70 @@ def populate_event_cache(self, session: sqlalchemy.orm.Session):
         services.alerts.crud.Events().cache_initialized = True
         logger.debug("Finished populating event cache for alerts")
 
+    def _should_send_notification(
+        self, alert: mlrun.common.schemas.AlertConfig, state_obj: dict
+    ) -> bool:
+        if alert.criteria:
```
Yacouby: same comment as above

Yacouby (changes requested):
another comment

server/py/services/alerts/crud/alerts.py
rokatyy: @Yacouby good point, I just copied this logic from the before, so will change to alert.critiria >1

server/py/services/alerts/crud/alerts.py
```diff
@@ -175,76 +175,35 @@ def process_event(
...
-                    alert.name,
-                    count=state["count"],
-                    last_updated=event_data.timestamp,
-                    obj=state_obj,
-                    active=active,
-                )
+        # Append the event timestamp if the alert is active or has criteria.
+        # If the alert is active, it follows a manual reset policy, so we need to continue counting events.
+        # If the criteria is 1 and the alert is inactive, we trigger the alert immediately and do not cache events.
+        if state["active"] or alert.criteria.count > 1:
```
Yacouby: what if alert.criteria.count =, period=None ( the default when not passing criteria ) 
notification will not be sent in the current implementation

Yacouby (approved):

server/py/services/alerts/crud/alerts.py
TomerShor: Add a comment that the activation is already saved with the `criteria.count` to avoid future questions

rokatyy: @TomerShor added

TomerShor (approved):

TomerShor (approved):

PR #6754, opened 2024-11-23
[AutoMount] Add mlrun runtime mounts back to platforms for backwards compatibility

liranbg:
Side effect of https://github.com/mlrun/mlrun/pull/6727

https://iguazio.atlassian.net/browse/ML-8405


mlrun/platforms/__init__.py
```diff
@@ -24,6 +25,48 @@
...
+        return getattr(self._lazy_load(), attr)
+
+    def _lazy_load(self, *args, **kwargs):
+        import mlrun.runtimes.mounts as mlrun_mounts
+
+        return getattr(mlrun_mounts, self._new_target)
+
+    def _warn(self):
+        warnings.warn(
+            f"mlrun.platforms.{self._new_target} is deprecated since version {self._version}, "
```
alonmr, line 50: We also specify when it will be removed (and add a `# TODO: remove in 1.10`)
```suggestion
            f"mlrun.platforms.{self._new_target} is deprecated since version {self._version}, and will be removed in 1.10"
```

alonmr (approved):

PR #6760, opened 2024-11-24
[Alert activations] Fix list activations by severity

rokatyy:
Request params isn't passed correctly without `Query([], alias="severity")`

jira - https://iguazio.atlassian.net/browse/ML-8430

BC tests fail exactly because of this change. Filter by severity has never worked and api is new, so we are safe.

### DIFF
--- a/server/py/services/api/api/endpoints/alert_activations.py
+++ b/server/py/services/api/api/endpoints/alert_activations.py
@@ -40,7 +40,7 @@ async def list_alert_activations(
     entity: Optional[str] = None,
     severity: Optional[
         list[Union[mlrun.common.schemas.alert.AlertSeverity, str]]
-    ] = None,
+    ] = Query([], alias="severity"),
     entity_kind: Optional[
         Union[mlrun.common.schemas.alert.EventEntityKind, str]
     ] = Query(None, alias="entity-kind"),
--- a/tests/integration/sdk_api/alerts/test_alerts.py
+++ b/tests/integration/sdk_api/alerts/test_alerts.py
@@ -269,6 +269,63 @@ class TestAlerts(tests.integration.sdk_api.base.TestMLRunIntegration):
             number_of_events=1,
         )
 
+    def test_list_alert_activations_by_severity(self):
+        project_name = "my-new-project"
+        event_name = alert_objects.EventKind.DATA_DRIFT_DETECTED
+        alert_name = "drift"
+        alert_summary = "Model {{project}}/{{entity}} is drifting."
+        alert_entity_kind = alert_objects.EventEntityKind.MODEL_ENDPOINT_RESULT
+        alert_entity_project = project_name
+
+        project = mlrun.new_project(alert_entity_project)
+        self._create_alert(
+            alert_entity_project,
+            alert_name,
+            alert_entity_kind,
+            alert_entity_project,
+            alert_summary,
+            event_name,
+            reset_policy=mlrun.common.schemas.alert.ResetPolicy.AUTO,
+        )
+        self._post_event(alert_entity_project, event_name, alert_entity_kind)
+        self._create_alert(
+            alert_entity_project,
+            alert_name,
+            alert_entity_kind,
+            alert_entity_project,
+            alert_summary,
+            event_name,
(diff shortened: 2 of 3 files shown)


tests/integration/sdk_api/alerts/test_alerts.py
```diff
@@ -633,30 +680,32 @@ def _create_alert(
...
+            alert_entity_kind or alert.entities.kind,
+            alert_entity_project or alert.project,
+            alert_summary or alert.summary,
+            event_name or alert.trigger.events[0].name,
+            severity or alert.severity,
+            criteria or alert.criteria,
+            # notification needs to be sent every time due to not having secret params in the client side
             notifications,
-            reset_policy,
+            reset_policy or alert.reset_policy,
```
TomerShor: Use kwargs to be on the safe side

tests/integration/sdk_api/alerts/test_alerts.py
```diff
@@ -269,6 +269,53 @@ def test_alert_after_project_deletion(self):
...
+        )
+        self._post_event(alert_entity_project, event_name, alert_entity_kind)
+
+        activations = project.list_alert_activations(name=alert_name)
+        assert len(activations) == 2
+
+        for severity in [
+            mlrun.common.schemas.alert.AlertSeverity.LOW,
+            mlrun.common.schemas.alert.AlertSeverity.HIGH,
+        ]:
```
TomerShor: The endpoint also supports severity as a list, add a test case for that too

server/py/services/api/api/endpoints/alert_activations.py
```diff
@@ -40,7 +40,7 @@ async def list_alert_activations(
     entity: Optional[str] = None,
     severity: Optional[
         list[Union[mlrun.common.schemas.alert.AlertSeverity, str]]
```
TomerShor, line 42: Actually, why do we support a random string at all?
The idea of using the schemas is that we will have validations on the input and of won't be able to send something random.

TomerShor (changes requested):

server/py/services/api/api/endpoints/alert_activations.py
rokatyy: @TomerShor firstly, This is just typing.. from real-world side it can be anything. Also, it is for filtering, so I thought that allowing filtering by string is better for any BC issues, but if you think we need it, I can explicitly translate it to AlertSeverity object. However I haven't found any place in the code where we would do it like this.

TomerShor: In FastAPI, pydantic is enforcing the endpoint params types, so typing does matter.
There is not BC here as this is a new endpoint, and the severity is also enforced on the [AlertConfig object](https://github.com/mlrun/mlrun/blob/1b288a97709161603f2586ada547ac1db346ede4/mlrun/common/schemas/alert.py#L151), so you can't set it to a random string.
Lastly, I think it is better to fail-fast if user tries to filter on something that we don't support

rokatyy: @TomerShor we don't translate the given object to AlertConfig/AlertSeverity when listing alert activations. In terms of BC I was talking about another case. Imagine client has newer version of MLRun which supports new type of AlertSeverity. If user passes this new type in the filter along with other severities, we will fail fast. Do we want to do this instead of just ignoring it?

TomerShor: If the client is newer then the server, this new severity won't be supported in the backend anyway.
IMO if we choose using the schemas as the query param types, then we should enforce them. If we don't want to enforce specific values then why use schemas at all?..

rokatyy: @TomerShor I agree that we should have validation when storing data. But when filtering.. Especially if we allow the list here, I'd go with best-effort approach

rokatyy: Also, it is out of scope for this PR. Whether we decide to hardcode schemas here or not, it can be done in separate PR.

moranbental (approved):
LGTM

TomerShor (approved):

PR #6762, opened 2024-11-25
[SystemTest] Fix needless timezone specifier

liranbg:


liranbg:
Fails on unrealted test, would be fix on a follow-up pr

PR #6768, opened 2024-11-25
[CI] Enforce fully qualified imports

adamdelman:
[CI] Add import validation flow for modified Python files

    - Script checks for fully qualified imports in modified files.
    - Supports file exclusion via CLI arguments.


.github/workflows/enforce_imports.yml
```diff
@@ -0,0 +1,43 @@
...
+      with:
+        python-version: '3.9'
+
+    - name: Get Modified Files
+      run: |
+        BASE_BRANCH=${{ github.event.pull_request.base.ref }}
+        git diff --name-only $(git merge-base HEAD $BASE_BRANCH) > modified_files.txt
+
+
+
```
TomerShor: Newline much?
```suggestion

```

automation/scripts/check_fully_qualified_imports.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    Scan all provided files and report disallowed imports.
+    """
+    all_violations = []
+    for file_path in files:
+        if file_path.endswith(".py"):
+            all_violations.extend(check_disallowed_imports(file_path))
+
+    if all_violations:
+        print("The following disallowed imports were found:")
```
TomerShor: "Disallowed" is mostly a verb, and not an adjective.
I'd use "forbidden", "invalid" or "non-conforming" - you can choose :)
(Relevant to all occurrences of this word)

automation/scripts/check_fully_qualified_imports.py
```diff
@@ -0,0 +1,59 @@
...
+        for violation in all_violations:
+            print(f"❌ {violation}")
+        sys.exit(1)
+    else:
+        print("✅ All imports are valid.")
+        sys.exit(0)
+
+
+if __name__ == "__main__":
+    main(sys.argv[1:])
```
TomerShor: If you're not using argparse or similar, at least add a comment of what the script expects (=> a list of file names)

automation/scripts/check_fully_qualified_imports.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def check_disallowed_imports(file_path: pathlib.Path) -> list[str]:
+    """
+    Check for disallowed imports in a file and return a list of violations.
+    """
+    violations = []
+    with open(file_path) as file:
+        for line_num, line in enumerate(file, start=1):
+            if DISALLOWED_PATTERN.match(line.strip()):
```
TomerShor: I think you can skip lines that don't start with or have the word "from", because you don't really need to match them.

TomerShor (changes requested):
Nice one, added some comments below

automation/scripts/check_fully_qualified_imports.py
adamdelman: I'm not using arg parse because I want this to be runnable in the future as part of a pre-commit git hook

automation/scripts/check_fully_qualified_imports.py
adamdelman: I think this is probably premature optimization, regex is fast enough to not process the entire line if the beginning doesnt match

.github/workflows/enforce_imports.yml
```diff
@@ -0,0 +1,41 @@
...
+name: Enforce Fully Qualified Imports
+
+on:
+  pull_request:
+    paths:
+      - '**/*.py'
+
+jobs:
+  check_imports:
+    runs-on: ubuntu-latest
```
moranbental, line 24:
```suggestion
    runs-on: ubuntu-latest
    strategy:
      matrix:
        python-version: [ "3.9"]
```

.github/workflows/enforce_imports.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.9'
```
moranbental:
```suggestion
    - name: Set up python ${{ matrix.python-version }}
      uses: actions/setup-python@v5
      with:
        python-version: ${{ matrix.python-version }}
        cache: pip
```

.github/workflows/enforce_imports.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    - name: Get Modified Files
+      run: |
+        BASE_BRANCH=${{ github.event.pull_request.base.ref }}
+        git diff --name-only $(git merge-base HEAD $BASE_BRANCH) > modified_files.txt
+
+    - name: Run Import Check
+      run: |
+        python /home/runner/work/mlrun/mlrun/automation/scripts/check_fully_qualified_imports.py $(cat modified_files.txt)
```
moranbental: Why do you need to specify the entire path? I think you can run it here with `automation/scripts/check_fully_qualified_imports.py`, as we are doing in `ci.yaml`

moranbental (changes requested):
Looks good!
We need to run it on Python 3.11 as well (in my task), so I made some changes to support it in the future. For now, it will run on Python 3.9

moranbental (approved):
LGTM

TomerShor (approved):
Nice work

PR #6770, opened 2024-11-25
[Datastore] Add vectordb support

alxtkr77:
https://iguazio.atlassian.net/browse/ML-6379

### DIFF
--- a/mlrun/artifacts/document.py
+++ b/mlrun/artifacts/document.py
@@ -1,3 +1,17 @@
+# Copyright 2024 Iguazio
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 import ast
 import re
 import tempfile
@@ -78,9 +92,9 @@ class DocumentLoader:
 
     def __new__(
         cls,
-        artifact_key: str,
         source_path: str,
         loader_spec: "DocumentLoaderSpec",
+        artifact_key="doc%%",
         producer: Optional[Union["MlrunProject", str, "MLClientCtx"]] = None,  # noqa: F821
         upload: bool = False,
     ):
@@ -92,9 +106,9 @@ class DocumentLoader:
                 self,
                 source_path,
                 loader_spec,
-                artifact_key="%%",
-                producer=None,
-                upload=False,
+                artifact_key,
+                producer,
+                upload,
             ):
                 self.producer = producer
                 self.artifact_key = (
--- a/mlrun/artifacts/document.py
+++ b/mlrun/artifacts/document.py
@@ -1,0 +1,299 @@
+import ast
+import re
+import tempfile
+from collections.abc import Iterator
+from copy import deepcopy
+from importlib import import_module
+from typing import Optional, Union
+
+import mlrun
+from mlrun.artifacts import Artifact, ArtifactSpec
+from mlrun.model import ModelObj
+
+from ..utils import generate_artifact_uri
+
+
+class DocumentLoaderSpec(ModelObj):
+    """
+    A class to load a document from a file path using a specified loader class.
+
+    This class is responsible for loading documents from a given source path using a specified loader class.
+    The loader class is dynamically imported and instantiated with the provided arguments. The loaded documents
+    can be optionally uploaded as artifacts.
+
+    Attributes:
+        loader_class_name (str): The name of the loader class to use for loading documents.
+        src_name (str): The name of the source attribute to pass to the loader class.
+        kwargs (Optional[dict]): Additional keyword arguments to pass to the loader class.
+
+    Methods:
+        make_loader(src_path): Creates an instance of the loader class with the specified source path.
+    """
+
+    _dict_fields = ["loader_class_name", "src_name", "kwargs"]
+
+    def __init__(
+        self,
+        loader_class_name: str = "langchain_community.document_loaders.TextLoader",
+        src_name: str = "file_path",
+        kwargs: Optional[dict] = None,
+    ):
+        """
+        Initialize the document loader.
+
+        Args:
+            loader_class_name (str): The name of the loader class to use.
+            src_name (str): The source name for the document.
+            kwargs (Optional[dict]): Additional keyword arguments to pass to the loader class.
+        """
+        self.loader_class_name = loader_class_name
+        self.src_name = src_name
+        self.kwargs = kwargs
+
+    def make_loader(self, src_path):
+        module_name, class_name = self.loader_class_name.rsplit(".", 1)
+        module = import_module(module_name)
+        loader_class = getattr(module, class_name)
+        kwargs = deepcopy(self.kwargs or {})
+        kwargs[self.src_name] = src_path
+        loader = loader_class(**kwargs)
+        return loader
+
+
+class DocumentLoader:
+    """
+    A factory class for creating instances of a dynamically defined document loader.
+
+    Args:
+        artifact_key (str): The key for the artifact to be logged.It can include '%%' which will be replaced
+        by a hex-encoded version of the source path.
+        source_path (str): The source path of the document to be loaded.
+        loader_spec (DocumentLoaderSpec): Specification for the document loader.
+        producer (Optional[Union[MlrunProject, str, MLClientCtx]], optional): The producer of the document
+        upload (bool, optional): Flag indicating whether to upload the document.
+
+    Returns:
+        DynamicDocumentLoader: An instance of a dynamically defined subclass of BaseLoader.
+    """
+
+    def __new__(
+        cls,
+        artifact_key: str,
+        source_path: str,
+        loader_spec: "DocumentLoaderSpec",
+        producer: Optional[Union["MlrunProject", str, "MLClientCtx"]] = None,  # noqa: F821
+        upload: bool = False,
+    ):
+        # Dynamically import BaseLoader
+        from langchain_community.document_loaders.base import BaseLoader
+
+        class DynamicDocumentLoader(BaseLoader):
+            def __init__(
+                self,
+                source_path,
+                loader_spec,
+                artifact_key="%%",
+                producer=None,
+                upload=False,
+            ):
+                self.producer = producer
+                self.artifact_key = (
+                    DocumentLoader.artifact_key_instance(artifact_key, source_path)
+                    if "%%" in artifact_key
+                    else artifact_key
+                )
+                self.loader_spec = loader_spec
+                self.source_path = source_path
+                self.upload = upload
+
+                # Resolve the producer
+                if not self.producer:
+                    self.producer = mlrun.mlconf.default_project
+                if isinstance(self.producer, str):
+                    self.producer = mlrun.get_or_create_project(self.producer)
+
+            def lazy_load(self) -> Iterator["Document"]:  # noqa: F821
+                artifact = self.producer.log_document(
+                    key=self.artifact_key,
+                    document_loader=self.loader_spec,
+                    src_path=self.source_path,
+                    upload=self.upload,
+                )
+                yield artifact.to_langchain_documents()
+
+        # Return an instance of the dynamically defined subclass
+        instance = DynamicDocumentLoader(
+            artifact_key=artifact_key,
+            source_path=source_path,
+            loader_spec=loader_spec,
+            producer=producer,
+            upload=upload,
+        )
+        return instance
+
+    @staticmethod
+    def artifact_key_instance(artifact_key: str, src_path: str) -> str:
+        if "%%" in artifact_key:
+            pattern = mlrun.utils.regex.artifact_key[0]
+            # Convert anchored pattern (^...$) to non-anchored version for finditer
+            search_pattern = pattern.strip("^$")
+            result = []
+            current_pos = 0
+
+            # Find all valid sequences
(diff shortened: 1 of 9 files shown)


tests/system/datastore/test_vectorstore.py
```diff
@@ -0,0 +1,193 @@
...
+            assert langchain_documents[0].metadata["mlrun_chunk"] == "0"
+
+            # Test logging a document localy
+            artifact = self.project.log_document(
+                "test_document_artifact", src_path=temp_file.name, upload=True
+            )
+
+            stored_artifcat = self.project.get_artifact("test_document_artifact")
+            stored_langchain_documents = stored_artifcat.to_langchain_documents()
+            stored_langchain_documents[0].metadata["mlrun_target_path"]
```
assaf758, line 79: what is this line?

tests/system/datastore/test_vectorstore.py
```diff
@@ -0,0 +1,193 @@
...
+                langchain_documents[0].metadata["source"]
+                == f"{self.project.name}/test_document_artifact"
+            )
+            assert langchain_documents[0].metadata["original_source"] == temp_file.name
+            assert langchain_documents[0].metadata["mlrun_object_uri"] == artifact.uri
+            assert langchain_documents[0].metadata["mlrun_chunk"] == "0"
+
+            # Test logging a document localy
+            artifact = self.project.log_document(
+                "test_document_artifact", src_path=temp_file.name, upload=True
```
assaf758, line 74: make `test_document_artifact` a constant

tests/system/datastore/test_vectorstore.py
```diff
@@ -0,0 +1,193 @@
...
+            # Test logging a document localy
+            artifact = self.project.log_document(
+                "test_document_artifact", src_path=temp_file.name, upload=False
+            )
+            langchain_documents = artifact.to_langchain_documents()
+
+            assert len(langchain_documents) == 1
+            assert (
+                langchain_documents[0].page_content
+                == "This is a test document for vector store."
```
assaf758, line 62: make constant

tests/system/datastore/test_vectorstore.py
```diff
@@ -0,0 +1,193 @@
...
+            # Test logging a document localy
+            loader = DocumentLoader(
+                source_path=temp_file.name,
+                loader_spec=DocumentLoaderSpec(),
+                artifact_key="test_document_artifact",
+                producer=self.project,
+                upload=False,
+            )
+            lc_documents = loader.load()
+            assert len(lc_documents) == 1
```
assaf758, line 120: check page_content

mlrun/datastore/datastore_profile.py
```diff
@@ -81,6 +81,24 @@ class DatastoreProfileBasic(DatastoreProfile):
     private: typing.Optional[str] = None
 
 
+class VectorStoreProfile(DatastoreProfile):
+    type: str = pydantic.Field("vector")
+    _private_attributes = ("kwargs_private",)
+    vector_store_class: str
+    kwargs_public: typing.Optional[dict] = None
```
assaf758, line 88: how the user know what fields to provide?

assaf758 (changes requested):

assaf758 (approved):
Very cool.
Some minor comments, please address in follow-up pr.

mlrun/execution.py
```diff
@@ -861,6 +867,46 @@ def log_model(
...
+        tag: str = "",
+        upload: Optional[bool] = False,
+        labels: Optional[dict[str, str]] = None,
+        **kwargs,
+    ) -> DocumentArtifact:
+        """
+        Log a document as an artifact.
+
+        :param key: Artifact key
+        :param target_path: Path to the local file
```
theSaarco, line 884: This is missing from the function signature. In general, this should not be `target_path` but rather `source_path` (or `local_path`) and it should point at the local file to create the artifact from.

mlrun/projects/project.py
```diff
@@ -1839,6 +1856,72 @@ def log_model(
...
+        tag: str = "",
+        upload: Optional[bool] = False,
+        labels: Optional[dict[str, str]] = None,
+        **kwargs,
+    ) -> DocumentArtifact:
+        """
+        Log a document as an artifact.
+
+        :param key: Artifact key
+        :param target_path: Path to the local file
```
theSaarco, line 1902: Here too, this parameters is missing.

mlrun/execution.py
```diff
@@ -861,6 +867,46 @@ def log_model(
...
+        :param kwargs: Additional keyword arguments
+        :return: DocumentArtifact object
+        """
+        doc_artifact = DocumentArtifact(
+            key=key,
+            document_loader=document_loader,
+            **kwargs,
+        )
+
+        item = self._artifacts_manager.log_artifact(
```
theSaarco, line 899: This method should get the `local_path` as well, currently I'm not sure where it gets the contents from.

mlrun/projects/project.py
```diff
@@ -1520,6 +1533,10 @@ def register_artifacts(self):
                     is_retained_producer=is_retained_producer,
                 )
 
+    def update_artifact(self, artifact_object: Artifact):
+        artifacts_manager = self._get_artifact_manager()
+        artifacts_manager.update_artifact(artifact_object, artifact_object)
```
theSaarco, line 1538: 1st parameter of `update_artifact` is the producer - I believe it should be `self` in this case.

mlrun/artifacts/document.py
```diff
@@ -0,0 +1,313 @@
...
+    """
+
+    def __new__(
+        cls,
+        source_path: str,
+        loader_spec: "DocumentLoaderSpec",
+        artifact_key="doc%%",
+        producer: Optional[Union["MlrunProject", str, "MLClientCtx"]] = None,  # noqa: F821
+        upload: bool = False,
+    ):
```
theSaarco, line 100: Why do you need the `__new__` method for this, and this external class? Can't the `DynamicDocumentLoader` simply be the `MLRunDocumentLoader` and its `__init__` (as already implemented) will be used for initialization?

mlrun/artifacts/document.py
```diff
@@ -0,0 +1,313 @@
...
+
+        # Return an instance of the dynamically defined subclass
+        instance = DynamicDocumentLoader(
+            artifact_key=artifact_key,
+            source_path=source_path,
+            loader_spec=loader_spec,
+            producer=producer,
+            upload=upload,
+        )
+        return instance
```
theSaarco, line 146: I find it extremely confusing the you create an object of type `DocumentLoader` and you get back an internal `DynamicDocumentLoader` class instance. I'm not sure I understand why this is really needed.

mlrun/artifacts/document.py
```diff
@@ -0,0 +1,313 @@
...
+                loader_spec,
+                artifact_key,
+                producer,
+                upload,
+            ):
+                self.producer = producer
+                self.artifact_key = (
+                    DocumentLoader.artifact_key_instance(artifact_key, source_path)
+                    if "%%" in artifact_key
+                    else artifact_key
```
theSaarco, line 117: You do this check again in `artifact_key_instance` - this is redundant.

mlrun/artifacts/document.py
```diff
@@ -0,0 +1,313 @@
...
+class DocumentLoader:
+    """
+    A factory class for creating instances of a dynamically defined document loader.
+
+    Args:
+        artifact_key (str): The key for the artifact to be logged.It can include '%%' which will be replaced
+        by a hex-encoded version of the source path.
+        source_path (str): The source path of the document to be loaded.
+        loader_spec (DocumentLoaderSpec): Specification for the document loader.
+        producer (Optional[Union[MlrunProject, str, MLClientCtx]], optional): The producer of the document
```
theSaarco, line 86: Explain a bit more, that a `str` is the project name, and if not provided, the current project will be used.

mlrun/artifacts/document.py
```diff
@@ -0,0 +1,313 @@
...
+            # Convert anchored pattern (^...$) to non-anchored version for finditer
+            search_pattern = pattern.strip("^$")
+            result = []
+            current_pos = 0
+
+            # Find all valid sequences
+            for match in re.finditer(search_pattern, src_path):
+                # Add hex values for characters between matches
+                for char in src_path[current_pos : match.start()]:
+                    result.append(hex(ord(char))[2:].zfill(2))
```
theSaarco, line 161: Hex values are a bit distracting here, resulting in things like `22tmp22myfile44txt` where it's hard to separate the file-name from the filled chars. Why not simply replace with `-`? Will result in something like `-tmp-myfile-txt` in this case.

mlrun/artifacts/document.py
```diff
@@ -0,0 +1,313 @@
...
+    A factory class for creating instances of a dynamically defined document loader.
+
+    Args:
+        artifact_key (str): The key for the artifact to be logged.It can include '%%' which will be replaced
+        by a hex-encoded version of the source path.
+        source_path (str): The source path of the document to be loaded.
+        loader_spec (DocumentLoaderSpec): Specification for the document loader.
+        producer (Optional[Union[MlrunProject, str, MLClientCtx]], optional): The producer of the document
+        upload (bool, optional): Flag indicating whether to upload the document.
+
```
theSaarco, line 88: An example of using this method should also be provided, helping the user understand the `DocumentLoaderSpec` usage, for example.

mlrun/artifacts/document.py
```diff
@@ -0,0 +1,313 @@
...
+    Returns:
+        DynamicDocumentLoader: An instance of a dynamically defined subclass of BaseLoader.
+    """
+
+    def __new__(
+        cls,
+        source_path: str,
+        loader_spec: "DocumentLoaderSpec",
+        artifact_key="doc%%",
```
theSaarco, line 97: I'm thinking that perhaps the `key` should be a mandatory parameter, even maybe the 1st parameter...

mlrun/artifacts/document.py
```diff
@@ -0,0 +1,313 @@
...
+    """
+    A specific artifact class inheriting from generic artifact, used to maintain Document meta-data.
+
+    Methods:
+        to_langchain_documents(splitter): Create LC documents from the artifact.
+        collection_add(collection_id): Add a collection ID to the artifact.
+        collection_remove(collection_id): Remove a collection ID from the artifact.
+    """
+
+    class DocumentArtifactSpec(ArtifactSpec):
```
theSaarco, line 188: Why inside the `DocumentArtifact` class?

mlrun/artifacts/document.py
```diff
@@ -0,0 +1,313 @@
...
+        _dict_fields = ArtifactSpec._dict_fields + [
+            "document_loader",
+            "collections",
+            "original_source",
+        ]
+
+        def __init__(
+            self,
+            *args,
```
theSaarco, line 197: You should add the class-specific fields as specific parameters here.

mlrun/artifacts/document.py
```diff
@@ -0,0 +1,313 @@
...
+            **kwargs,
+        ):
+            super().__init__(*args, **kwargs)
+            self.document_loader = None
+            self.collections = set()
```
theSaarco, line 202: If the class is created and `collections` are passed to it, this will simply ignore this value and throw it away. You should verify that `to_dict` and `from_dict` for this class works properly when these values exist.

mlrun/artifacts/document.py
```diff
@@ -0,0 +1,313 @@
...
+            self._spec.collections = ast.literal_eval(self._spec.collections)
+
+    @property
+    def inputs(self):
+        # To keep the interface consistent with the project.update_artifact() when we update the artifact
+        return None
+
+    @property
+    def source(self):
+        return generate_artifact_uri(self.metadata.project, self.spec.db_key)
```
theSaarco, line 248: I'm not a fan of hiding logic behind a property. Would much prefer changing this to `get_source()` which hints to users that something is being done behind the scenes.

theSaarco (commented):
Very good start. I have a bunch of comments and some decisions that I'm not sure about their reasoning. Didn't go over the tests yet.

mlrun/artifacts/document.py
alxtkr77: will fix

mlrun/artifacts/document.py
alxtkr77: The MLRunDocumentLoader has to be derived from BaseLoader - it is tested by the langchain down the road. That is the reason for __new__

mlrun/artifacts/document.py
alxtkr77: Langchain assumes the first parameter of a loader is the path (In DirectoryLoader for example)

mlrun/artifacts/document.py
alxtkr77: Because its internal to the DocumentArtifact ? Isn't it a good practice?

mlrun/artifacts/document.py
alxtkr77: My impression was that this class is set only by the DocumentArtifact.spec setter and never using passing the parameters. Am I wrong?

alxtkr77: But you are right - will add parameters anyway to keep it generic

mlrun/artifacts/document.py
theSaarco: Yes, sure, but why do you need the `DynamicDocumentLoader` class wrapper? Can't you just define the `MLRunDocumentLoader` and directly call its `__init__` for creation? What's the added benefit of having this external class?

mlrun/artifacts/document.py
theSaarco: So 2nd 😄

mlrun/artifacts/document.py
theSaarco: Well, sort of. Given that Python doesn't really have private members, nested classes are mostly for namespacing and you can say that putting the classes in the same py file (package) serves roughly the same purpose through import. The advantages of encapsulation you'll get in most languages don't really hold for Python.
In the MLRun code the specs of objects are usually top-level classes, so I was thinking to keep with how things are usually done in MLRun.

PR #6776, opened 2024-11-25
[Alerts] Fix reset alert reason logging

Yacouby:
Updated the _should_reset_alert method to handle reset policy changes more explicitly. 
Now, the log correctly reports changes in reset-policy when the reset is triggered by a change in the alert's reset policy. 

Resolves : 
https://iguazio.atlassian.net/browse/ML-8434

### DIFF
--- a/server/py/services/alerts/crud/alerts.py
+++ b/server/py/services/alerts/crud/alerts.py
@@ -407,12 +407,11 @@ class Alerts(
         # reset the alert if the policy was modified from manual to auto while the state is active
         old_reset_policy = getattr(old_alert_data, "reset_policy")
         new_reset_policy = getattr(alert_data, "reset_policy")
-        reset_due_to_policy_change = (
+        if (
             old_alert_data.state == mlrun.common.schemas.AlertActiveState.ACTIVE
             and old_reset_policy == mlrun.common.schemas.alert.ResetPolicy.MANUAL
             and new_reset_policy == mlrun.common.schemas.alert.ResetPolicy.AUTO
-        )
-        if reset_due_to_policy_change:
+        ):
             return "reset-policy changed from manual to auto"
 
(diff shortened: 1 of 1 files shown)


server/py/services/alerts/crud/alerts.py
```diff
@@ -101,16 +101,13 @@ def store_alert(
...
-                "force_reset being True"
-                if force_reset
-                else "changes in entities, criteria, or trigger of the alert",
+        if existing_alert is not None:
+            should_reset, reset_reason = self._should_reset_alert(
+                existing_alert, alert_data, force_reset
             )
-            self.reset_alert(session, project, new_alert.name)
+            if should_reset:
+                logger.debug("Resetting alert due to %s", reset_reason)
```
TomerShor: What alert are we resetting?
Also, it's better to use arguments over formatted strings, as the logs stay consistent.
```suggestion
                logger.debug("Resetting alert before storing", project=project, alert_name=name, reason=reset_reason)
```

TomerShor (changes requested):
A general comment - do we really need to log the reason?
It's in the API so it's not visible to the client...

Yacouby:
> A general comment - do we really need to log the reason? It's in the API so it's not visible to the client...

it may be good for debugging

server/py/services/alerts/crud/alerts.py
```diff
@@ -400,29 +402,26 @@ def reset_alert(self, session: sqlalchemy.orm.Session, project: str, name: str):
     @staticmethod
     def _should_reset_alert(old_alert_data, alert_data, force_reset):
```
TomerShor, line 403: Because the function is **should**_reset_alert, I prefer it to still to return (bool, str) with wether it should be reset, and the reason, and not only the reason

TomerShor (approved):

PR #6782, opened 2024-11-26
[KFP] Add Retry Functionality

adamdelman:
This PR introduces a retry mechanism for Kubernetes Pipelines (KFP) by adding a `retry_pipeline` function to MLRun. Key updates include:

New retry_pipeline method: A method was added to the HTTPRunDB class to enable retrying pipeline runs. If a pipeline run fails or encounters an error, this method allows the pipeline to be retried.

`ExtendedKfpClient` class: This new class extends the KFP client to handle pipeline retries. The `retry_run` method checks the status of the pipeline run. If the run is in a retryable state (i.e., failed or errored), it re-executes the run via the KFP retry endpoint.

Fallback to New Run Creation: If the pipeline run is not in a valid state for retry (i.e., not in a failed or errored state), a new pipeline run is created with the same configuration. This ensures that users can reattempt execution even when a pipeline is not eligible for direct retry. The process works as follows:
Pipeline Retry Endpoint: A new FastAPI endpoint (POST `/projects/{project}/pipelines/{run_id}/retry`) has been added to trigger the retry or new run creation based on the pipeline’s status.

https://iguazio.atlassian.net/browse/ML-8438
https://iguazio.atlassian.net/browse/ML-8276


mlrun/db/httpdb.py
```diff
@@ -2239,6 +2239,75 @@ def get_pipeline(
...
+                error=str(exc),
+            )
+            if isinstance(exc, mlrun.errors.MLRunHTTPError):
+                raise exc  # Re-raise known HTTP errors
+            raise mlrun.errors.MLRunRuntimeError(
+                f"Unexpected error while retrying pipeline run '{run_id}'."
+            ) from exc
+
+        logger.info(
+            "Successfully retried pipeline run.",
```
liranbg:
```suggestion
            "Successfully retried pipeline run",
```

server/py/services/api/api/endpoints/pipelines.py
```diff
@@ -122,18 +129,53 @@ async def create_pipeline(
...
+    db_session: sqlalchemy.orm.Session = fastapi.Depends(
+        framework.api.deps.get_db_session
+    ),
+):
+    await (
+        framework.utils.auth.verifier.AuthVerifier().query_project_resource_permissions(
+            mlrun.common.schemas.AuthorizationResourceTypes.pipeline,
+            project,
+            run_id,
+            mlrun.common.schemas.AuthorizationAction.read,
```
liranbg: I think we need to ensure ".create" and not just "read" as we rerun it

server/py/services/api/crud/pipelines.py
```diff
@@ -199,16 +200,68 @@ def get_pipeline(
...
 
         return run
 
+    def retry_pipeline(
+        self,
+        db_session: sqlalchemy.orm.Session,
+        run_id: str,
+        project: typing.Optional[str] = None,
+        namespace: typing.Optional[str] = None,
+    ):
```
liranbg: add docstring for the function flow

server/py/services/api/crud/pipelines.py
```diff
@@ -199,16 +200,68 @@ def get_pipeline(
...
+                    )
+        mlrun.utils.logger.debug(
+            "Retrying KFP run",
+            run_id=run_id,
+            run_name=run.get("name"),
+            project=project,
+        )
+        kfp_client.retry_run(
+            run_id=run_id,
+        )
```
liranbg: wouldnt it be better to return here and make retry_run returns the updated run?

liranbg (changes requested):
few notes here
1. better off doing the code updates on a separated PR to avoid noise on reading the changes
2. missing UT/System test coverage
3. missing function docstrings

liranbg (approved):

server/py/services/api/api/endpoints/pipelines.py
```diff
@@ -122,18 +129,53 @@ async def create_pipeline(
     return response
 
 
+@router.post("/{run_id}/retry")
+async def retry_pipeline(
+    run_id: str,
```
moranbental, line 134: Should the `request: fastapi.Request parameter` also be here? I remember @alonmr mentioned that

server/py/services/api/crud/pipelines.py
```diff
@@ -199,16 +200,87 @@ def get_pipeline(
...
+        if (
+            run.status
+            not in mlrun_pipelines.common.models.RunStatuses.stable_statuses()
+        ):
+            raise mlrun.errors.MLRunBadRequestError(
+                f"Pipeline run {run_id} is not in a completed state. Current status: {run.status}"
+            )
+
+        if run:
+            if project and project != "*":
```
moranbental: We are no longer using `project="*"`. This was changed in PR [#6683](https://github.com/mlrun/mlrun/pull/6683). Use `list_allowed_project_names` instead

server/py/services/api/crud/pipelines.py
```diff
@@ -199,16 +200,87 @@ def get_pipeline(
...
+        :param namespace: (Optional) The Kubernetes namespace in which the pipeline is running.
+                          Defaults to the configured namespace if not specified.
+        :type namespace: Optional[str]
+        :raises MLRunBadRequestError: If the pipeline run is not in a retryable state.
+        :raises MLRunNotFoundError: If the pipeline run does not belong to the specified project
+                                    or if the run ID is not found.
+        :raises MLRunRuntimeError: If there is an error retrieving the pipeline run details.
+        :raises MLRunHTTPStatusError: If there is an HTTP error interacting with KFP.
+        :return: The unique identifier of the retried pipeline run.
+        :rtype: str
```
moranbental: This is not our docstring format, @rokatyy, am I right?

server/py/services/api/crud/pipelines.py
```diff
@@ -199,16 +200,87 @@ def get_pipeline(
...
+            api_run_detail = kfp_client.get_run(run_id)
+        except kfp_server_api.ApiException as exc:
+            raise mlrun.errors.err_for_status_code(
+                exc.status, mlrun.errors.err_to_str(exc)
+            ) from exc
+        except mlrun.errors.MLRunHTTPStatusError:
+            raise
+        except Exception as exc:
+            raise mlrun.errors.MLRunRuntimeError(
+                f"Failed getting kfp run: {mlrun.errors.err_to_str(exc)}"
```
moranbental:
```suggestion
                f"Failed getting KFP run: {mlrun.errors.err_to_str(exc)}"
```

server/py/services/api/crud/pipelines.py
```diff
@@ -199,16 +200,87 @@ def get_pipeline(
...
+        # Check if the pipeline is in a completed state
+        if (
+            run.status
+            not in mlrun_pipelines.common.models.RunStatuses.stable_statuses()
+        ):
+            raise mlrun.errors.MLRunBadRequestError(
+                f"Pipeline run {run_id} is not in a completed state. Current status: {run.status}"
+            )
+
+        if run:
```
moranbental: Why is it necessary to check `if run`? Can it be None? If so, you've already checked `run.status` before, so that could fail

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/utils.py
```diff
@@ -12,11 +12,91 @@
...
+        :type run_id: str
+        :raises ApiException: If the API request fails during the retry or new run creation process.
+        :raises ValueError: If the experiment ID cannot be found for the given run ID, or if
+                            the original run does not contain a valid pipeline specification.
+        :raises FileNotFoundError: If a temporary file for the workflow manifest cannot be created or accessed.
+        :return: The ID of the new or retried run.
+        :rtype: str
+        """
+        # Fetch run details
+        run_details = self.get_run(run_id).run
```
moranbental, line 46: Can `get_run` return None?

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/utils.py
```diff
@@ -12,11 +12,91 @@
...
+            mlrun_pipelines.common.models.RunStatuses.failed,
+            mlrun_pipelines.common.models.RunStatuses.error,
+        }
+        if run_status in valid_states_for_retry:
+            self._experiment_api.api_client.call_api(
+                f"/apis/v1beta1/runs/{run_id}/retry",
+                "POST",
+                response_type="ApiRun",
+                auth_settings=["BearerToken"],
+            )
```
moranbental: What if this API call fails? Maybe add a try-except here

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/utils.py
```diff
@@ -12,11 +12,91 @@
...
+            ),
+            None,
+        )
+        if not experiment_id:
+            raise ValueError(f"Experiment ID not found for run ID: {run_id}")
+
+        valid_states_for_retry = {
+            mlrun_pipelines.common.models.RunStatuses.failed,
+            mlrun_pipelines.common.models.RunStatuses.error,
+        }
```
moranbental, line 64: In your HLD, you also mention the `Succeeded` and `Skipped` statuses
https://iguazio.atlassian.net/wiki/spaces/MLRUN/pages/183534076/Retry+Workflow+Feature+HLD#:~:text=Valid%20States%20for%20Retry%3A

moranbental (changes requested):
Nice! I added some notes

quaark (commented):
Looks great!
I'm wondering if it's worth adding a retry method to the `_PipelineRunStatus` class.
Since the way we run pipelines in the SDK, we call `pipeline = project.run(...)` and receive a `_PipelineRunStatus` object (with the run id embedded), which can theoretically then retried with `pipeline.retry()`

server/py/services/api/crud/pipelines.py
adamdelman: I was looking at `mlrun.db.httpdb.HTTPRunDB.list_runs` for example, is there any official docstring format for mlrun?

server/py/services/api/api/endpoints/pipelines.py
adamdelman: It doesnt look like it's used here, I think it was only required when passing the request to a different microservice than the one that received it

server/py/services/api/crud/pipelines.py
adamdelman: fixed

server/py/services/api/crud/pipelines.py
adamdelman: fixed

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/utils.py
adamdelman: it should raise an exception if the run is not found

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/utils.py
adamdelman: Suceeded and Skipped can be retried, but in those cases the a new workflow will be created as a clone of the previous one.

server/py/services/api/crud/pipelines.py
adamdelman: as discussed, the changes in the references PR have mostly to do with pagination and filtering projects efficiently, which is not really relevant here as the project can be retrieved from the run annotations

server/py/services/api/crud/pipelines.py
moranbental: We're not really using `:type`

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/utils.py
```diff
@@ -12,11 +12,113 @@
...
+import mlrun_pipelines.helpers
+import mlrun_pipelines.imports
+
+logger = mlrun.utils.create_logger(
+    level=mlrun.config.config.log_level,
+    formatter_kind=mlrun.config.config.log_formatter,
+    name="mlrun",
+    stream=sys.stdout,
+)
+
```
liranbg: use get logger from mlrun utils

server/py/services/api/crud/pipelines.py
```diff
@@ -199,16 +200,86 @@ def get_pipeline(
...
+            raise mlrun.errors.MLRunBadRequestError(
+                f"Pipeline run {run_id} is not in a completed state. Current status: {run.status}"
+            )
+
+        if project and project != "*":
+            run_project = self.resolve_project_from_pipeline(run)
+            if run_project != project:
+                raise mlrun.errors.MLRunNotFoundError(
+                    f"Pipeline run with id {run_id} is not of project {project}"
+                )
```
liranbg: move before line 259

liranbg (commented):
minor

server/py/services/api/crud/pipelines.py
adamdelman: fixed

liranbg (approved):

PR #6783, opened 2024-11-26
[API] Missing function kind on list functions

liranbg:
take the `kind` and shove it to struct to ensure format will return the `kind` as top-level field
https://iguazio.atlassian.net/browse/ML-8461


adamdelman:
I think the commit message could be a bit more specific 😄

PR #6785, opened 2024-11-26
[Alert activation] Add reset_time to alert activation

rokatyy:
Adds new reset_time to alert activation table, so we can determine:

1. whether alert activation is now active or not
2. when alert activation was reset

Jira - https://iguazio.atlassian.net/browse/ML-8432


server/py/framework/db/base.py
```diff
@@ -968,12 +968,12 @@ def store_alert_activation(
         pass
 
     @abstractmethod
-    def update_alert_activation_number_of_events(
+    def update_alert_activation_on_reset(
```
TomerShor: Despite my comment in the previous PR, let's just call it `update_alert_activation` if we'll maybe use it in other places and update other things.

server/py/framework/db/sqldb/db.py
```diff
@@ -5946,14 +5946,17 @@ def store_alert_activation(
                 session, alert_activation_record, "id"
             )
         else:
+            # for auto reset policy reset_time is the same as the activation time
+            # for manual reset policy, we keep it empty until policy reset
```
```suggestion
            # for manual reset policy, we keep it empty until the alert is reset
```

server/py/services/alerts/crud/alerts.py
```diff
@@ -410,24 +410,23 @@ def reset_alert(self, session: sqlalchemy.orm.Session, project: str, name: str):
...
-                )
-                activation_time = alert_state.get("last_updated")
-                activation_id = alert_state.get("full_object", {}).get(
-                    "last_activation_id"
+            alert_state = framework.utils.singletons.db.get_db().get_alert_state_dict(
+                session, alert.id
+            )
+            activation_time = alert_state.get("last_updated")
+            activation_id = alert_state.get("full_object", {}).get("last_activation_id")
+            if activation_time and activation_id:
```
TomerShor, line 418: If one or both are None - is it a valid case?
If not, we should at least log a warning.

TomerShor (changes requested):
Minor comments

server/py/framework/db/sqldb/db.py
rokatyy: @TomerShor actually, that's alert being reset, reset policy just indicates when to reset it

server/py/services/alerts/crud/alerts.py
rokatyy: @TomerShor nope - not a valid case, will add warning

server/py/framework/db/sqldb/db.py
TomerShor: exactly, the alert is reset.
"until policy reset" means the the `policy` is reset, which is not the case.

server/py/services/alerts/crud/alerts.py
```diff
@@ -410,24 +410,30 @@ def reset_alert(self, session: sqlalchemy.orm.Session, project: str, name: str):
...
+                    number_of_events=number_of_events
+                    if number_of_events > alert.criteria.count
+                    else None,
                 )
-                activation_time = alert_state.get("last_updated")
-                activation_id = alert_state.get("full_object", {}).get(
-                    "last_activation_id"
+            else:
+                logger.warning(
+                    "No activation id or last activation time found for alert",
```
```suggestion
                    "No activation id or last activation time found for alert, skipping activation update on reset",
```

server/py/framework/db/sqldb/db.py
rokatyy: @TomerShor right, sorry :facepalm:

Yacouby (approved):
Nice 👍

TomerShor (approved):

PR #6792, opened 2024-11-27
[SQLRunDB] Create new session in session getter

alonmr:
When we send the notifications we create a dictionary from the run in multiple places with run.to_dict() there it tries to fetch the artifacts to fulfill the dict

It fails because it uses a global db session from the run db factory - in the server we assume that the db session is injected from the fastapi endpoint but it’s not the case here.

We decided to keep the artifacts in the run dict but we'll create a new db session before accessing the db

https://iguazio.atlassian.net/browse/ML-8781
https://iguazio.atlassian.net/browse/ML-8463

### DIFF
--- a/server/py/framework/rundb/sqldb.py
+++ b/server/py/framework/rundb/sqldb.py
@@ -48,15 +48,20 @@ class SQLRunDB(RunDBInterface):
         dsn,
         session=None,
     ):
-        self.session = session
+        self._session = session
         self.dsn = dsn
         self.db = None
 
     def connect(self, secrets=None):
-        if not self.session:
-            self.session = create_session()
         self.db = SQLDB(self.dsn)
         return self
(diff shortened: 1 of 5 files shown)


server/py/framework/rundb/sqldb.py
```diff
@@ -48,16 +48,21 @@ def __init__(
         dsn,
         session=None,
     ):
-        self.session = session
+        self._session = session
         self.dsn = dsn
         self.db = None
 
     def connect(self, secrets=None):
```
roei3000b, line 55: It strange that the `connect` doesn't do the actual `connect` (`create_session`). I think we should add here comment that explain it.

roei3000b (changes requested):
Very nice solution :)
Minor.

PR #6794, opened 2024-11-27
[Notifications] Make notification pusher resilient to errors

roei3000b:
We want to make the notification pusher more resistant to exceptions so even if one of the notification cannot be send the others will send.

https://iguazio.atlassian.net/browse/ML-8482

### DIFF
--- a/mlrun/utils/notifications/notification_pusher.py
+++ b/mlrun/utils/notifications/notification_pusher.py
@@ -118,39 +118,43 @@ class NotificationPusher(_NotificationPusherBase):
         ] = []
 
         for run in self._runs:
+            self._process_run(run)
+
+    def _process_run(self, run):
+        try:
+            if isinstance(run, dict):
+                run = mlrun.model.RunObject.from_dict(run)
+
+            for notification in run.spec.notifications:
+                self._process_notification(notification, run)
+        except Exception as exc:
+            logger.warning(
+                "Failed to load notification from run",
+                run_uid=run.metadata.uid,
+                error=mlrun.errors.err_to_str(exc),
+            )
+
+    def _process_notification(self, notification, run):
+        try:
             try:
-                if isinstance(run, dict):
-                    run = mlrun.model.RunObject.from_dict(run)
-
-                for notification in run.spec.notifications:
-                    try:
-                        try:
-                            notification.status = run.status.notifications.get(
-                                notification.name
-                            ).get(
-                                "status",
-                                mlrun.common.schemas.NotificationStatus.PENDING,
-                            )
-                        except (AttributeError, KeyError):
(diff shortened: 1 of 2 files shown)


mlrun/utils/notifications/notification_pusher.py
```diff
@@ -118,21 +118,39 @@ def __init__(
...
+                            run_uid=run.metadata.uid,
+                            notification=notification,
+                            error=mlrun.errors.err_to_str(exc),
+                        )
+            except Exception as exc:
+                logger.warning(
+                    "Failed to load notification from run",
+                    run_uid=run.metadata.uid,
+                    error=mlrun.errors.err_to_str(exc),
+                )
```
moranbental: I would suggest splitting it into two functions when
```
for run in self._runs:
    self._process_run(run)
    
def _process_run(run):
    if isinstance(run, dict):
        run = mlrun.model.RunObject.from_dict(run)

    for notification in run.spec.notifications:
        self._process_notification(run, notification)
```

moranbental (changes requested):
Looks good! just one suggestion

mlrun/utils/notifications/notification_pusher.py
```diff
@@ -118,21 +118,39 @@ def __init__(
...
-
-                if self._should_notify(run, notification):
-                    self._load_notification(run, notification)
+            try:
+                if isinstance(run, dict):
+                    run = mlrun.model.RunObject.from_dict(run)
+
+                for notification in run.spec.notifications:
+                    try:
+                        try:
```
liranbg: It could be look better I think, find a way to use less try/except and less indentation
overall the idea of not failing on sending a single notificaiton is legit, just need to ensure that code-looking and robustness is not affected by it

liranbg (changes requested):

mlrun/utils/notifications/notification_pusher.py
```diff
@@ -118,21 +118,43 @@ def __init__(
...
+            logger.warning(
+                "Failed to load notification from run",
+                run_uid=run.metadata.uid,
+                error=mlrun.errors.err_to_str(exc),
+            )
 
-                if self._should_notify(run, notification):
-                    self._load_notification(run, notification)
+    def _process_notification(self, notification, run):
+        try:
```
liranbg: remove the outer scope of `try` and move it to whoever call `_process_notification` and on the exception message, replace the `Failed to load notification` with `Failed to process notification`
this way you gain a correct warning log message and less indented code

liranbg (changes requested):

moranbental (approved):
LGTM

liranbg (approved):

PR #6799, opened 2024-11-28
[Tests] Fix KFP imports for tests

adamdelman:
Fix broken KFP imports in tests


quaark (approved):

liranbg (changes requested):
I think that we cannot mutate such tests as they simulate user stories.
what we can do is marking those tests and skip them upon need

PR #6810, opened 2024-11-28
[Datastore] Vectordb bugs and design-implementation mismatches

alxtkr77:
https://iguazio.atlassian.net/browse/ML-8498
https://iguazio.atlassian.net/browse/ML-8488
https://iguazio.atlassian.net/browse/ML-8465
https://iguazio.atlassian.net/browse/ML-8464
https://iguazio.atlassian.net/browse/ML-8466

### DIFF
--- a/mlrun/artifacts/document.py
+++ b/mlrun/artifacts/document.py
@@ -303,13 +303,32 @@ class DocumentArtifact(Artifact):
         return results
 
     def collection_add(self, collection_id: str) -> None:
+        """
+        Add a collection ID to the artifact's collection list.
+
+        Adds the specified collection ID to the artifact's collection mapping if it
+        doesn't already exist.
+        This method only modifies the client-side artifact object and does not persist
+        the changes to the MLRun DB. To save the changes permanently, you must call
+        project.update_artifact() after this method.
+
+        Args:
+            collection_id (str): The ID of the collection to add
+        """
         if collection_id not in self.spec.collections:
             self.spec.collections[collection_id] = "1"
-            project = mlrun.get_or_create_project(mlrun.mlconf.default_project)
-            project.update_artifact(self)
 
     def collection_remove(self, collection_id: str) -> None:
+        """
+        Remove a collection ID from the artifact's collection list.
+
+        Removes the specified collection ID from the artifact's local collection mapping.
+        This method only modifies the client-side artifact object and does not persist
+        the changes to the MLRun DB. To save the changes permanently, you must call
+        project.update_artifact() after this method.
+
+        Args:
+            collection_id (str): The ID of the collection to remove
+        """
         if collection_id in self.spec.collections:
             self.spec.collections.pop(collection_id)
-            project = mlrun.get_or_create_project(mlrun.mlconf.default_project)
-            project.update_artifact(self)
--- a/mlrun/artifacts/document.py
+++ b/mlrun/artifacts/document.py
@@ -325,7 +325,7 @@ class DocumentArtifact(Artifact):
         Removes the specified collection ID from the artifact's local collection mapping.
         This method only modifies the client-side artifact object and does not persist
         the changes to the MLRun DB. To save the changes permanently, you must call
-        project.update_artifact() after this method.
+        project.update_artifact() or context.update_artifact() after this method.
 
         Args:
             collection_id (str): The ID of the collection to remove
--- a/mlrun/artifacts/document.py
+++ b/mlrun/artifacts/document.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import ast
 import re
 import tempfile
 from collections.abc import Iterator
@@ -129,7 +128,7 @@ class MLRunLoader:
             def lazy_load(self) -> Iterator["Document"]:  # noqa: F821
                 artifact = self.producer.log_document(
                     key=self.artifact_key,
-                    document_loader=self.loader_spec,
+                    document_loader_spec=self.loader_spec,
                     local_path=self.local_path,
                     upload=self.upload,
                 )
@@ -196,9 +195,9 @@ class DocumentArtifact(Artifact):
         def __init__(
             self,
             *args,
-            document_loader=None,
-            collections=None,
-            original_source=None,
+            document_loader: Optional[DocumentLoaderSpec] = None,
+            collections: Optional[dict] = None,
+            original_source: Optional[str] = None,
             **kwargs,
         ):
             super().__init__(*args, **kwargs)
@@ -221,12 +220,14 @@ class DocumentArtifact(Artifact):
     def __init__(
         self,
         original_source: Optional[str] = None,
-        document_loader: Optional[DocumentLoaderSpec] = None,
+        document_loader_spec: Optional[DocumentLoaderSpec] = None,
         **kwargs,
     ):
         super().__init__(**kwargs)
         self.spec.document_loader = (
-            document_loader.to_str() if document_loader else self.spec.document_loader
+            document_loader_spec.to_dict()
+            if document_loader_spec
+            else self.spec.document_loader
         )
         self.spec.original_source = original_source or self.spec.original_source
 
@@ -258,9 +259,8 @@ class DocumentArtifact(Artifact):
         Returns:
             list[Document]: A list of LangChain Document objects.
         """
-        dictionary = ast.literal_eval(self.spec.document_loader)
-        loader_spec = DocumentLoaderSpec.from_dict(dictionary)
 
+        loader_spec = DocumentLoaderSpec.from_dict(self.spec.document_loader)
         if self.get_target_path():
             with tempfile.NamedTemporaryFile() as tmp_file:
                 mlrun.datastore.store_manager.object(
--- a/mlrun/artifacts/document.py
+++ b/mlrun/artifacts/document.py
@@ -81,7 +81,7 @@ class MLRunLoader:
     Args:
         artifact_key (str): The key for the artifact to be logged.It can include '%%' which will be replaced
         by a hex-encoded version of the source path.
-        source_path (str): The source path of the document to be loaded.
+        local_path (str): The source path of the document to be loaded.
         loader_spec (DocumentLoaderSpec): Specification for the document loader.
         producer (Optional[Union[MlrunProject, str, MLClientCtx]], optional): The producer of the document
         upload (bool, optional): Flag indicating whether to upload the document.
--- a/mlrun/artifacts/document.py
+++ b/mlrun/artifacts/document.py
@@ -74,7 +74,7 @@ class DocumentLoaderSpec(ModelObj):
         return loader
 
 
-class DocumentLoader:
+class MLRunLoader:
     """
     A factory class for creating instances of a dynamically defined document loader.
 
@@ -104,7 +104,7 @@ class DocumentLoader:
         class DynamicDocumentLoader(BaseLoader):
             def __init__(
                 self,
-                source_path,
+                local_path,
                 loader_spec,
                 artifact_key,
                 producer,
@@ -112,12 +112,12 @@ class DocumentLoader:
             ):
                 self.producer = producer
                 self.artifact_key = (
-                    DocumentLoader.artifact_key_instance(artifact_key, source_path)
+                    MLRunLoader.artifact_key_instance(artifact_key, local_path)
                     if "%%" in artifact_key
                     else artifact_key
                 )
                 self.loader_spec = loader_spec
-                self.source_path = source_path
+                self.local_path = local_path
                 self.upload = upload
 
                 # Resolve the producer
@@ -130,15 +130,16 @@ class DocumentLoader:
                 artifact = self.producer.log_document(
                     key=self.artifact_key,
                     document_loader=self.loader_spec,
-                    src_path=self.source_path,
+                    local_path=self.local_path,
                     upload=self.upload,
                 )
-                yield artifact.to_langchain_documents()
+                res = artifact.to_langchain_documents()
+                yield res[0]
 
         # Return an instance of the dynamically defined subclass
         instance = DynamicDocumentLoader(
             artifact_key=artifact_key,
-            source_path=source_path,
+            local_path=source_path,
             loader_spec=loader_spec,
             producer=producer,
             upload=upload,
@@ -195,12 +196,15 @@ class DocumentArtifact(Artifact):
         def __init__(
             self,
             *args,
+            document_loader=None,
+            collections=None,
+            original_source=None,
             **kwargs,
         ):
             super().__init__(*args, **kwargs)
-            self.document_loader = None
-            self.collections = set()
-            self.original_source = None
+            self.document_loader = document_loader
+            self.collections = collections if collections is not None else {}
+            self.original_source = original_source
 
     """
     A specific artifact class inheriting from generic artifact, used to maintain Document meta-data.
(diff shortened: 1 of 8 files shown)


mlrun/artifacts/document.py
```diff
@@ -104,20 +104,20 @@ def __new__(
         class DynamicDocumentLoader(BaseLoader):
             def __init__(
                 self,
-                source_path,
+                local_path,
```
assaf758, line 106: please change source_path to loacal_path in the docstring

mlrun/projects/project.py
```diff
@@ -1888,38 +1888,44 @@ def get_or_create_vector_store_collection(
...
-        :param target_path: Path to the local file
+        :param tag: Version tag
+        :param local_path:    path to the local file we upload, will also be use
+                              as the destination subpath (under "artifact_path")
         :param artifact_path: Target path for artifact storage
         :param document_loader: Spec to use to load the artifact as langchain document
-        :param tag: Version tag
         :param upload: Whether to upload the artifact
         :param labels: Key-value labels
+        :param target_path: Path to the local file
```
assaf758: param description seems wrong

assaf758 (commented):
looks good, two minor comments.
Also if Chroma special logic would be replaced, consider removing it from this PR for claritiy.

mlrun/artifacts/document.py
```diff
@@ -104,20 +104,20 @@ def __new__(
         class DynamicDocumentLoader(BaseLoader):
```
theSaarco, line 103: I have to say I still don't get the reasoning for using an internal-class here and why this code (the `DynamicDocumentLoader::__new__` method) cannot simply be implemented in the `MLRunLoader::__init__` method. Of course, this means that the `MLRunLoader` class will inherit from `BaseLoader` directly.

mlrun/artifacts/document.py
```diff
@@ -130,15 +130,16 @@ def lazy_load(self) -> Iterator["Document"]:  # noqa: F821
...
+                yield res[0]
 
         # Return an instance of the dynamically defined subclass
         instance = DynamicDocumentLoader(
             artifact_key=artifact_key,
-            source_path=source_path,
+            local_path=source_path,
             loader_spec=loader_spec,
             producer=producer,
             upload=upload,
```
theSaarco, line 144: Repeating my comment from previous PR (regarding the `artifact_key_instance` method), let's discuss - 
Hex values are a bit distracting here, resulting in things like `22tmp22myfile44txt` where it's hard to separate the file-name from the filled chars. Why not simply replace with -? Will result in something like -tmp-myfile-txt in this case.

mlrun/artifacts/document.py
```diff
@@ -195,12 +196,15 @@ class DocumentArtifactSpec(ArtifactSpec):
         def __init__(
             self,
             *args,
+            document_loader=None,
+            collections=None,
+            original_source=None,
```
theSaarco: Please add type hints where applicable.

mlrun/artifacts/document.py
```diff
@@ -216,14 +220,15 @@ def __init__(
...
+        document_loader: Optional[DocumentLoaderSpec] = None,
         **kwargs,
     ):
-        super().__init__(key, **kwargs)
-        self.spec.document_loader = document_loader.to_str()
-        if "src_path" in kwargs:
-            self.spec.original_source = kwargs["src_path"]
+        super().__init__(**kwargs)
+        self.spec.document_loader = (
+            document_loader.to_str() if document_loader else self.spec.document_loader
```
theSaarco: Why `to_str()`? You're using `from_dict` to de-serialize, but in general - why not maintain the object as object? You only need to serialize it when serializing the entire object.

mlrun/artifacts/document.py
```diff
@@ -216,14 +220,15 @@ def __init__(
 
     def __init__(
         self,
-        key=None,
-        document_loader: DocumentLoaderSpec = DocumentLoaderSpec(),
+        original_source: Optional[str] = None,
+        document_loader: Optional[DocumentLoaderSpec] = None,
```
theSaarco:
```suggestion
        document_loader_spec: Optional[DocumentLoaderSpec] = None,
```

mlrun/artifacts/document.py
```diff
@@ -301,13 +297,19 @@ def to_langchain_documents(
...
                 )
                 results.append(doc)
         return results
 
     def collection_add(self, collection_id: str) -> None:
-        self.spec.collections.add(collection_id)
+        if collection_id not in self.spec.collections:
+            self.spec.collections[collection_id] = "1"
+            project = mlrun.get_or_create_project(mlrun.mlconf.default_project)
+            project.update_artifact(self)
```
theSaarco: The artifact should not take care of updating itself in the DB. It's up to the user to update if needed through the relevant producer.
Also, this code will work in some cases, but will always assume you're registering the document through a project, while the user can log the document through the MLRun execution context (i.e. the run is the producer).

mlrun/artifacts/document.py
```diff
@@ -301,13 +297,19 @@ def to_langchain_documents(
...
 
     def collection_remove(self, collection_id: str) -> None:
-        return self.spec.collections.discard(collection_id)
+        if collection_id in self.spec.collections:
+            self.spec.collections.pop(collection_id)
+            project = mlrun.get_or_create_project(mlrun.mlconf.default_project)
+            project.update_artifact(self)
```
theSaarco: Same comment here.

mlrun/artifacts/manager.py
```diff
@@ -325,7 +325,7 @@ def log_artifact(
 
     def update_artifact(self, producer, item: Artifact):
         self.artifact_uris[item.key] = item.uri
-        self._log_to_db(item.db_key, producer.project, producer.inputs, item)
+        self._log_to_db(item.db_key, producer.name, None, item)
```
theSaarco: This doesn't seem right. It will only work properly if the producer is a project, otherwise it will use the run-name as the project-name which is wrong.

theSaarco (changes requested):
I have some comments on the implementation, let's discuss some of the items (for example the class structure in `MLRunLoader`.

assaf758 (approved):

mlrun/artifacts/document.py
alxtkr77: We cannot inherit from BaseLoader directly, because when langchain is not installed it will generate errors

mlrun/artifacts/document.py
alxtkr77: Then we loose uniqueness - /tmp/file and /tmp~file will generate same key

mlrun/artifacts/document.py
alxtkr77: will fix

mlrun/artifacts/document.py
alxtkr77: will fix

mlrun/artifacts/document.py
alxtkr77: will fix

mlrun/artifacts/document.py
alxtkr77: will fix

mlrun/artifacts/document.py
alxtkr77: will fix

mlrun/artifacts/manager.py
alxtkr77: will fix

mlrun/artifacts/document.py
theSaarco: Indeed, but that's a pretty much of a corner case, don't you think? I think it's better to have better readability for the common case than sacrifice it for some cases that may never happen.

mlrun/artifacts/document.py
```diff
@@ -301,13 +297,38 @@ def to_langchain_documents(
...
 
     def collection_remove(self, collection_id: str) -> None:
-        return self.spec.collections.discard(collection_id)
+        """
+        Remove a collection ID from the artifact's collection list.
+
+        Removes the specified collection ID from the artifact's local collection mapping.
+        This method only modifies the client-side artifact object and does not persist
+        the changes to the MLRun DB. To save the changes permanently, you must call
+        project.update_artifact() after this method.
```
theSaarco:
```suggestion
        project.update_artifact() or context.update_artifact() after this method.
```

mlrun/artifacts/document.py
```diff
@@ -301,13 +297,38 @@ def to_langchain_documents(
...
+        doesn't already exist.
+        This method only modifies the client-side artifact object and does not persist
+        the changes to the MLRun DB. To save the changes permanently, you must call
+        project.update_artifact() after this method.
+
+        Args:
+            collection_id (str): The ID of the collection to add
+        """
+        if collection_id not in self.spec.collections:
+            self.spec.collections[collection_id] = "1"
```
theSaarco, line 319: I don't remember if we discussed this before - currently there's no point in having `collections` as a dict, if the value is just `"1"`. I don't think we planned to have additional data added per collection, so why do we maintain it as a dict here and not a list/set?

mlrun/datastore/datastore_profile.py
```diff
@@ -81,22 +81,31 @@ class DatastoreProfileBasic(DatastoreProfile):
...
-            attributes = merge(attributes, self.kwargs_private)
-        if kwargs:
-            attributes = merge(attributes, kwargs)
-        return attributes
+class ConfigProfile(DatastoreProfile):
+    """
+    A profile class for managing configuration data with nested public and private attributes.
+    This class extends DatastoreProfile to handle configuration settings, separating them into
+    public and private dictionaries. Both dictionaries support nested structures, and the class
+    provides functionality to merge these attributes when needed.
```
theSaarco, line 89: It would be great to have a small code sample here explaining how to use this class and set credentials in it.

mlrun/datastore/vectorstore.py
```diff
@@ -13,65 +13,39 @@
...
-
-        remove_itself_from_artifact(artifact: DocumentArtifact):
-            Removes the current object from the given artifact's collection and updates the artifact.
-
-        delete_artifacts(artifacts: list[DocumentArtifact]):
-            Deletes a list of DocumentArtifact objects from the collection and updates the MLRun context.
-            Raises NotImplementedError if the delete operation is not supported for the collection implementation.
+        collection_name (str): Name of the vector store collection
+        _collection_impl (VectorStore): The underlying vector store implementation
+        _mlrun_context (Union[MlrunProject, MLClientCtx]): MLRun context for artifact tracking
```
theSaarco: Usually we don't really document the inner workings of the class - this should be the user-facing documentation of how to use this class, not how it's built.

mlrun/datastore/vectorstore.py
```diff
@@ -112,40 +86,66 @@ def add_documents(
...
 
     def add_artifacts(self, artifacts: list[DocumentArtifact], splitter=None, **kwargs):
         """
-        Add a list of DocumentArtifact objects to the collection.
+        Add a list of DocumentArtifact objects to the vector store collection.
+
+        Converts artifacts to LangChain documents, adds them to the vector store, and
+        updates the MLRun context. If documents are split, the IDs are handled appropriately.
 
         Args:
```
theSaarco, line 97: I see you're using the Google-style docstrings here - did you verify this renders well in Sphinx as well, for our documentation site? Most of MLRun uses different docstring format, so I'm not sure if this works or not.

mlrun/datastore/vectorstore.py
```diff
@@ -112,40 +86,66 @@ def add_documents(
...
+            list: List of IDs for all added documents. When no custom IDs are provided:
+                - Without splitting: Vector store generates IDs automatically
+                - With splitting: Vector store generates separate IDs for each chunk
+                When custom IDs are provided:
+                - Without splitting: Uses provided IDs directly
+                - With splitting: Generates sequential IDs as "{original_id}_{i}" for each chunk
         """
         all_ids = []
-        for artifact in artifacts:
+        user_ids = kwargs.pop("ids", None)
```
theSaarco, line 116: Should verify that the `user_ids` collection (if exists) has the same length as the number of artifacts, otherwise the code will explode later, when looking for `user_ids[index]`. It's preferred to also verify that it is of type `list` or at least that it is iterable.

mlrun/projects/project.py
```diff
@@ -1862,70 +1863,60 @@ def log_model(
...
         :param upload: Whether to upload the artifact
         :param labels: Key-value labels
+        :param target_path: Target file path
         :param kwargs: Additional keyword arguments
         :return: DocumentArtifact object
         """
         doc_artifact = DocumentArtifact(
             key=key,
-            document_loader=document_loader,
+            original_source=target_path or local_path,
```
```suggestion
            original_source=local_path or target_path,
```
If both are provided, you definitely want to prefer the local path.

mlrun/projects/project.py
```diff
@@ -4502,6 +4493,25 @@ def register_datastore_profile(self, profile: DatastoreProfile):
             profile, self.name
         )
 
+    def get_config_profile(self, name: str) -> dict:
```
theSaarco: This is not really getting the profile, but rather is a utility to retrieve all the parameters (if you just want the profile you can call `get_datastore_profile()`). 
So, looking at it now, I actually think it's better to rename this to something like `get_config_profile_attributes` - will make it more accurate.

mlrun/execution.py
```diff
@@ -877,28 +877,35 @@ def log_model(
...
+        :param document_loader: Spec to use to load the artifact as langchain document
         :param upload: Whether to upload the artifact
         :param labels: Key-value labels
+        :param target_path: Path to the local file
         :param kwargs: Additional keyword arguments
         :return: DocumentArtifact object
         """
         doc_artifact = DocumentArtifact(
             key=key,
+            original_source=target_path or local_path,
```
theSaarco:
```suggestion
            original_source=local_path or target_path,
```

tests/system/datastore/test_vectorstore.py
```diff
@@ -103,66 +109,112 @@ def test_vectorstore_document_artifact(self):
...
+        )
+        loader = DirectoryLoader(
+            temp_dir,
+            glob="**/*.*",
+            loader_cls=MLRunLoader,
+            loader_kwargs={
+                "loader_spec": artifact_loader_spec,
+                "artifact_key": artifact_key,
+                "producer": self.project,
+                "upload": False,
```
theSaarco, line 172: 🔥 This is awesome - it can also serve as a good usage example both for QA and for the docs.

theSaarco (changes requested):
Very nice. Have some comments and some changes needed, but overall looks very good, and I really like the tests added - they can be used as reference for tests/docs.

mlrun/artifacts/document.py
alxtkr77: Dictionary is a basic class we use to serialize/deserialize specs. And I just want to keep it forward compatible

mlrun/artifacts/document.py
alxtkr77: Hmm, I disagree - when such a thing happens (and it will), you won't be able to find the problem. As a result, you will have an artifact with wrong content, and worse yet, you won't even know about it.

theSaarco: I have to say that I feel that users managing their files this way are bound to get confused one way or the other. BTW, this will cause issues only if they choose to upload the files to the MLRun store (otherwise the content is not handled by MLRun).

mlrun/artifacts/document.py
theSaarco: So, and this is somewhat nitpicking, but I would put `None` as value, not `"1"`.

theSaarco (approved):
Looks good. Have some minor comments, and for now we'll agree to disagree on the way the key is generated - but let's try to solicit feedback on that from people using the code.

PR #6814, opened 2024-11-29
[Alert activations] Add SDK for alert config

Yacouby:
This PR introduces the SDK implementation for listing alert activations in the alert config.

Also a new field was added to the alert config, that stores the timestamp of the last modification to the alert. 
This field is used for tracking changes and enables filtering alert activations based on the alert's last update.

Jira:
https://iguazio.atlassian.net/browse/ML-8441


mlrun/alerts/alert.py
```diff
@@ -253,3 +257,103 @@ def _apply_template(self, template):
...
+        :param page_size: The number of items per page to retrieve. Up to `page_size` responses are expected.
+        :param page_token: A pagination token used to retrieve the next page of results. Should not be provided
+            for the first request.
+        :param from_last_update: If set to True, retrieves alert activations since the alert's last update time.
+
+        :returns: A tuple containing the list of alert activations and an optional `page_token` for pagination.
+        """
+        if from_last_update and self.updated is not None:
+            if isinstance(self.updated, str):
+                self.updated = datetime.fromisoformat(self.updated)
```
rokatyy: This code is duplicated in two places. Can we implement a property method to perform all the translation to datetime there?

mlrun/alerts/alert.py
```diff
@@ -253,3 +257,103 @@ def _apply_template(self, template):
...
+        since: Optional[datetime] = None,
+        until: Optional[datetime] = None,
+        from_last_update: bool = False,
+    ) -> list[mlrun.common.schemas.alert.AlertActivation]:
+        """
+        Retrieve a list of all alert activations.
+
+        :param since: Filters for alert activations occurring after this timestamp.
+        :param until: Filters for alert activations occurring before this timestamp.
+        :param from_last_update: If set to True, retrieves alert activations since the alert's last update time.
```
rokatyy, line 298: Add info that if since and from_last_update=True is passed, from_last_update takes precedence

server/py/services/alerts/crud/alerts.py
```diff
@@ -70,6 +70,9 @@ def store_alert(
             alert_data.created = existing_alert.created
             alert_data.id = existing_alert.id
 
+            # set the updated field to reflect the latest modification time of the alert
+            alert_data.updated = datetime.datetime.now(tz=datetime.timezone.utc)
```
rokatyy: `
```suggestion
            alert_data.updated = mlrun.utils.now_date()
```

server/py/services/alerts/tests/unit/crud/test_alerts.py
```diff
@@ -351,3 +351,68 @@ async def test_alert_reset_with_fields_updates(
...
+            alert_data=alert_data,
+        )
+
+        alert = services.alerts.crud.Alerts().get_enriched_alert(
+            session=db,
+            project=project,
+            name=alert_name,
+        )
+        assert alert.updated is not None
+        assert alert.updated < datetime.now(tz=timezone.utc)
```
rokatyy:
```suggestion
        assert alert.updated < alert.created
```

server/py/services/alerts/tests/unit/crud/test_alerts.py
```diff
@@ -351,3 +351,68 @@ async def test_alert_reset_with_fields_updates(
...
+            name=alert_name,
+            alert_data=alert_data,
+        )
+
+        alert = services.alerts.crud.Alerts().get_enriched_alert(
+            session=db,
+            project=project,
+            name=alert_name,
+        )
+        assert alert.updated is None
```
rokatyy: I don't think we should allow it to be empty. If it was just created, then `created=updated`

server/py/services/alerts/tests/unit/db/test_alerts.py
```diff
@@ -15,42 +15,32 @@
 
 from datetime import datetime, timezone
 
-import mlrun.common.schemas
-import mlrun.model
+import mlrun.common.schemas.alert as alert_objects
```
rokatyy, line 18:
```suggestion
import mlrun.common.schemas.alert
```
And then use objects with full link following the MLRun development best practices

tests/integration/sdk_api/alerts/test_alerts.py
```diff
@@ -727,6 +773,12 @@ def _get_alert_activations(project_name, alert_name=None):
             project=project_name, name=alert_name
         )
 
+    @staticmethod
+    def _get_alert_config_activations(alert, from_last_update=False, paginated=False):
+        if paginated:
+            return alert.paginated_list_activations(from_last_update=from_last_update)
+        return alert.list_activations(from_last_update=from_last_update)
+
```
rokatyy: We have `self._get_alert_activations` already which allows passing alert name, please either use it or pure methods

rokatyy (changes requested):
Current logic won't work correctly for the case where alert was created once and never updated. 

I think that on creation we should always fill `updated` as well. And only for those clients which had alerts before the update to 1.8.0, if `from_last_update` is passed and `updated` is None (because we didn't have this field in 1.7.0), we should use `created` instead and leave the comment why.

server/py/services/alerts/tests/unit/db/test_alerts.py
TomerShor: Using `as` is not against the best practices (`from` is frowned upon):

> When importing from local code, NEVER use: `from X import Y` instead use: `import X` or `import X as Y`

server/py/services/alerts/tests/unit/crud/test_alerts.py
Yacouby: @rokatyy you mean > alert.created

tests/integration/sdk_api/alerts/test_alerts.py
Yacouby: but i dont want to pass the alert name, since this should be called from the alert config object itself, and the existing method go through the project,  so it may be better to separate between them.

rokatyy: @Yacouby then won't it be better to use alert.list_activations() and alert.paginated_list_activations?

TomerShor: I agree here. You don't really save anything with this helper method and you can just call each list method where needed.

mlrun/alerts/alert.py
```diff
@@ -133,10 +136,20 @@ def __init__(
         self.state = state
         self.created = created
         self.count = count
+        self.updated = updated
```
liranbg:
```suggestion
        self._updated = updated
```

mlrun/alerts/alert.py
```diff
@@ -133,10 +136,20 @@ def __init__(
...
 
         if template:
             self._apply_template(template)
 
+    @property
+    def updated_datetime(self) -> datetime:
```
liranbg:
```suggestion
    def updated(self) -> datetime:
```

mlrun/alerts/alert.py
```diff
@@ -133,10 +136,20 @@ def __init__(
...
+        """
+        Get the `updated` field as a datetime object.
+        """
+        if isinstance(self.updated, str):
+            return datetime.fromisoformat(self.updated)
+        return self.updated
```
liranbg:
```suggestion
        if isinstance(self._updated, str):
            return datetime.fromisoformat(self._updated)
        return self._updated
```

liranbg (changes requested):
minor
no need to indicate the return type over the function signature
you either use typing hinting or put the units on the variable name (e.g _updated_seconds)

mlrun/alerts/alert.py
```diff
@@ -55,6 +56,7 @@ def __init__(
         state: alert_objects.AlertActiveState = None,
         created: Optional[str] = None,
         count: Optional[int] = None,
+        updated: Optional[datetime] = None,
```
rokatyy: isn't it a string?

mlrun/alerts/alert.py
```diff
@@ -350,8 +367,8 @@ def paginated_list_activations(
 
         :returns: A tuple containing the list of alert activations and an optional `page_token` for pagination.
         """
-        if from_last_update and self.updated is not None:
-            kwargs["since"] = self.updated_datetime
+        if from_last_update and self._updated is not None:
```
rokatyy: `if from_last_update and self._updated` is better for the cases if `_updated` is an empty string to be on safe side.

rokatyy (approved):

TomerShor (approved):

PR #6816, opened 2024-12-01
[Notifications] Mail - Merge between default addresses and the provided addresses

roei3000b:
Currently, when there are `default_mail_addresses` and in the notification declaration the user specify `email_addresses`, the `email_addresses` will override the  `default_mail_addresses`.
Instead of overriding we want to **merge** between them.

https://iguazio.atlassian.net/browse/ML-8519

### DIFF
--- a/mlrun/utils/notifications/notification/mail.py
+++ b/mlrun/utils/notifications/notification/mail.py
@@ -95,7 +95,7 @@ class MailNotification(base.NotificationBase):
         cls,
         default_mail_address: typing.Union[str, list],
         email_addresses: typing.Union[str, list],
-    ):
+    ) -> str:
         if isinstance(default_mail_address, str):
             default_mail_address = (
                 default_mail_address.split(",") if default_mail_address else []
--- a/mlrun/utils/notifications/notification/mail.py
+++ b/mlrun/utils/notifications/notification/mail.py
@@ -84,12 +84,27 @@ class MailNotification(base.NotificationBase):
         params.setdefault("server_port", DEFAULT_SMTP_PORT)
 
         default_mail_address = params.pop("default_email_addresses", "")
-        email_addresses = params.get("email_addresses", default_mail_address)
-        if isinstance(email_addresses, list):
-            email_addresses = ",".join(email_addresses)
-        params["email_addresses"] = email_addresses
+        params["email_addresses"] = cls._merge_mail_addresses(
+            default_mail_address, params.get("email_addresses", "")
+        )
 
         return params
+
+    @classmethod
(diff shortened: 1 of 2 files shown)


mlrun/utils/notifications/notification/mail.py
```diff
@@ -84,13 +84,28 @@ def enrich_default_params(
...
+        )
 
         return params
 
+    @classmethod
+    def _merge_mail_addresses(
+        cls,
+        default_mail_address: typing.Union[str, list],
+        email_addresses: typing.Union[str, list],
+    ):
```
moranbental:
```suggestion
    ) -> str:
```

mlrun/utils/notifications/notification/mail.py
```diff
@@ -84,13 +84,28 @@ def enrich_default_params(
...
+        if isinstance(default_mail_address, str):
+            default_mail_address = (
+                default_mail_address.split(",") if default_mail_address else []
+            )
+        if isinstance(email_addresses, str):
+            email_addresses = email_addresses.split(",") if email_addresses else []
+        email_addresses.extend(default_mail_address)
```
moranbental, line 105: I think you should remove any duplicate emails if they exist

moranbental (changes requested):
minor

mlrun/utils/notifications/notification/mail.py
roei3000b: We don't want to remove duplicates because this is some of the ways to demonstrate long lists of recipients.

moranbental: But why do you need the same recipient email in the list?

roei3000b: Because we want it to be real email

moranbental (approved):

mlrun/utils/notifications/notification/mail.py
TomerShor: I agree with Moran here, it makes no sense to have the same recipient more than once.
You can use a `set()` to avoid duplications

roei3000b: This logic shouldn't be on the "client" side, the smtp server should do it.
See for example this:
image

Here I specified your mail 3 times.
Only after sending we can see your mail shown only once:
image

But as client I mentioned your mail 3 times, the remove duplication should be on the smtp server (because for example if there are groups, this logic is not simple as use `set` for example).

Because of that I don't think we should make the "client" (meaning the side who sending the mail) remove the duplications.

PR #6817, opened 2024-12-01
[Serving] Add `ModelRunner` step

gtopper:
[ML-7689](https://iguazio.atlassian.net/browse/ML-7689)

[ML-7689]: https://iguazio.atlassian.net/browse/ML-7689?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/serving/test_flow.py
```diff
@@ -17,8 +17,14 @@
 import pytest
 
 import mlrun
-from mlrun.serving import GraphContext, V2ModelServer
-from mlrun.serving.states import TaskStep
+from mlrun.serving import (
```
assaf758: seems this file is used mostly for "sync" engine testing and we have test_async_flow for testing async.

mlrun/serving/states.py
```diff
@@ -805,6 +806,113 @@ def plot(self, filename=None, format=None, source=None, **kw):
...
+        """Override to implement prediction logic if the logic requires asyncio."""
+        return body
+
+    def run(self, body: Any, path: str) -> Any:
+        return self.predict(body)
+
+    async def run_async(self, body: Any, path: str) -> Any:
+        return self.predict(body)
+
+    def do_event(self, event):
```
assaf758: As I understand, storey uses `run` and `run_async` - what flow exercises do_event?

gtopper: mlrun [picks do_event up](https://github.com/mlrun/mlrun/blob/v1.8.0-rc9/mlrun/serving/states.py#L487) and [wraps it in a storey.Map](https://github.com/mlrun/mlrun/blob/v1.8.0-rc9/mlrun/serving/states.py#L1697).

This is defined for using the `Model` class as a step, without a `ModelRunnerStep`. But as per our last DF/DS call, I'll delete this functionality and users will need to use `ModelRunnerStep` even for a single model.

tests/serving/test_flow.py
gtopper: I'll move it.

assaf758 (approved):

PR #6818, opened 2024-12-01
[Notifications] Mail - set body from message param

roei3000b:
When message param provided, we should override the message to be the message param that was provided.

https://iguazio.atlassian.net/browse/ML-8517


moranbental (changes requested):
Looks good! missing unit test to covert that

mlrun/utils/notifications/notification/mail.py
```diff
@@ -70,7 +70,7 @@ async def push(
         event_data: typing.Optional[mlrun.common.schemas.Event] = None,
     ):
         self.params.setdefault("subject", f"[{severity}] {message}")
```
liranbg, line 73: why not taking the message from param as well?

liranbg (approved):

moranbental (approved):

PR #6822, opened 2024-12-02
[Pagination] Smarter Last Page Logic

quaark:
https://iguazio.atlassian.net/browse/ML-8505
https://iguazio.atlassian.net/browse/ML-8521

In order to properly determine the current page is the last page, where `N=page_size`, instead of querying `OFFSET=(page-1)*N` and `LIMIT=N`, we can query `LIMIT=N+1`:
- If we receive `<=N` items, this is the last page and we can omit the token. 
- If we receive `N+1` items, we omit the last item and return the response with the token.

To implement this we:
- Refactor the server side DB client, to accept offset and limit parameters on all the list_ methods. 
- `Paginator` now calculates the offset and limit instead of passing the page and page_size to the db methods. And then performs the above logic to determine whether to omit the page token or not.


server/py/framework/db/base.py
```diff
@@ -236,6 +236,7 @@ def list_artifacts(
         producer_id: Optional[str] = None,
         producer_uri: Optional[str] = None,
         format_: mlrun.common.formatters.ArtifactFormat = mlrun.common.formatters.ArtifactFormat.full,
+        offset: Optional[int] = None,
         limit: Optional[int] = None,
```
theSaarco, line 240: hmm... The existing `limit` field is actually already used by the UI to limit the total number of objects returned from the query. You cannot just change its meaning like this, since there are already existing APIs that use it.
I think the cleanest approach would be to rename these params - maybe just call it `page_offset` and `page_size`? This implies we're getting a single page starting at the offset and of the given size. WDYT?

theSaarco (changes requested):
There's a conflict with the existing list-artifacts limit...

server/py/framework/db/base.py
quaark: Spoke offline with @theSaarco and agreed on the necessary counter measures and we'll want to deprecate the `limit` in the API, but we'll support it for now.

server/py/framework/db/sqldb/db.py
```diff
@@ -6943,12 +6937,14 @@ def _is_table_exists(session: Session, table_name: str) -> bool:
...
     def _paginate_query(
-        query, page: typing.Optional[int] = None, page_size: typing.Optional[int] = None
+        query, offset: typing.Optional[int] = None, limit: typing.Optional[int] = None
     ):
-        if page is not None:
-            page_size = page_size or config.httpdb.pagination.default_page_size
-            if query.count() < page_size * (page - 1):
+        if offset:
+            if query.count() < offset:
                 raise StopIteration
```
theSaarco: As discussed, while not related to this change - this generates another query to the DB which is resource intensive and should be avoided. It seems this doesn't really generate value, so should be removed.

theSaarco (changes requested):
Change looks great. Another fix is needed, as discussed.

theSaarco (approved):
sababa

server/py/framework/db/sqldb/db.py
```diff
@@ -6943,12 +6937,10 @@ def _is_table_exists(session: Session, table_name: str) -> bool:
...
-        if page is not None:
-            page_size = page_size or config.httpdb.pagination.default_page_size
-            if query.count() < page_size * (page - 1):
-                raise StopIteration
-            query = query.limit(page_size).offset((page - 1) * page_size)
-
+        if offset:
+            query = query.offset(offset)
+        if limit:
+            query = query.limit(limit)
```
liranbg: nitpick
if limit is 0, then you essentially return all results

I suggest to check None type

liranbg (commented):
Minor

server/py/services/api/utils/pagination.py
```diff
@@ -261,3 +265,16 @@ def _create_or_update_pagination_cache_record(
...
         )
         return token, page, page_size, method, kwargs_schema.dict(exclude_none=True)
+
+    @staticmethod
+    def _calculate_offset_and_limit(
+        page: typing.Optional[int] = None,
+        page_size: typing.Optional[int] = None,
+    ) -> tuple[typing.Optional[int], typing.Optional[int]]:
+        if page is not None:
+            page_size = page_size or mlconf.httpdb.pagination.default_page_size
```
moranbental, line 275: Check here that page and page_size are valid values (they are not negative)

moranbental (changes requested):
Looks good, just one comment

moranbental (approved):
Nice

PR #6795, opened 2024-11-27
[Model Monitoring] Support running applications locally with `app.evaluate()`

jond01:
Add the `evaluate` method to `ModelMonitoringApplicationBase` class, currently without parameters/inputs.
Implements [ML-8357](https://iguazio.atlassian.net/browse/ML-8357).

[ML-8357]: https://iguazio.atlassian.net/browse/ML-8357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6825, opened 2024-12-02
[Model Monitoring] Fix getting the project in appication context

jond01:
Fixes an error from #6795 - the context did not work in serving graphs.
The context is now supported in serving graphs and local/workflow runs.
I ran the system test and added a unit test.


davesh0812 (approved):
LGTM.
One comment is that the current code(`get_current_project()`) not supports remote job unless the user will run `get_or_create_project` on the pod.

PR #6831, opened 2024-12-02
[Alert activations] Add notification summary to NotificationState

rokatyy:
Add notification summary to NotificationState as it is needed for UI design.

### DIFF
--- a/mlrun/common/schemas/notification.py
+++ b/mlrun/common/schemas/notification.py
@@ -132,8 +132,14 @@ class SetNotificationRequest(pydantic.v1.BaseModel):
     notifications: list[Notification] = None
 
 
+class NotificationSummary(pydantic.v1.BaseModel):
+    failed: int = 0
+    succeeded: int = 0
+
+
 class NotificationState(pydantic.v1.BaseModel):
     kind: str
     err: Optional[
         str
     ]  # empty error means that the notifications were sent successfully
+    summary: NotificationSummary
--- a/server/py/services/api/crud/alert_activation.py
+++ b/server/py/services/api/crud/alert_activation.py
@@ -79,8 +79,8 @@ class AlertActivation(
         notification_states = []
         for kind, status_info in notification_errors.items():
             errors = list(status_info["errors"])
-            success_count = status_info["success_count"]
-            failed_count = status_info["failed_count"]
+            success_count = status_info.get("success_count", 0)
+            failed_count = status_info.get("failed_count", 0)
 
             if errors:
                 if success_count == 0:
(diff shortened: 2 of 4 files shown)


server/py/services/api/crud/alert_activation.py
```diff
@@ -69,6 +71,7 @@ def _prepare_notifications_states(
             # count successes and collect unique errors for failures
             if reason:
                 notification_errors[kind]["errors"].add(reason)
+                notification_errors[kind]["failed_count"] += 1
```
roei3000b, line 74: If it is the only place where you increments the `failed_count`, why should we have it? it should be the length of the `errors`.

roei3000b (commented):
Looks great.
Minor.

server/py/services/api/crud/alert_activation.py
rokatyy: @roei3000b no, it is not because errors is set, which means it has **unique** errors only

server/py/services/api/crud/alert_activation.py
```diff
@@ -77,6 +80,7 @@ def _prepare_notifications_states(
         for kind, status_info in notification_errors.items():
             errors = list(status_info["errors"])
             success_count = status_info["success_count"]
+            failed_count = status_info["failed_count"]
```
TomerShor: I'd do a `get` to be a bit more defensive
```suggestion
            success_count = status_info.get("success_count", 0)
            failed_count = status_info.get("failed_count", 0)
```

TomerShor:
@rokatyy please add more info in the description / link to a jira ticket to give context on why this is needed (a summary of the slack thread)

mlrun/common/schemas/notification.py
```diff
@@ -132,8 +132,14 @@ class SetNotificationRequest(pydantic.v1.BaseModel):
...
+    failed: int = 0
+    succeeded: int = 0
+
+
 class NotificationState(pydantic.v1.BaseModel):
     kind: str
     err: Optional[
         str
     ]  # empty error means that the notifications were sent successfully
+    summary: NotificationSummary
```
moranbental, line 145: Should it be an optional value?

moranbental (changes requested):

server/py/services/api/crud/alert_activation.py
```diff
@@ -69,6 +71,7 @@ def _prepare_notifications_states(
             # count successes and collect unique errors for failures
```
Yacouby:
```suggestion
            # count successes and fails and collect unique errors for failures
```

mlrun/common/schemas/notification.py
rokatyy: @moranbental no, since this is new feature, problems might appear only when migrating between rc, but in actual released version in should not be optional

Yacouby (approved):
minor comment

moranbental (approved):
LGTM

TomerShor (approved):

PR #6833, opened 2024-12-02
[Invoke] Fix joining urls

rokatyy:
PR title says it all
Jira - https://iguazio.atlassian.net/browse/ML-8500

### DIFF
--- a/mlrun/runtimes/nuclio/function.py
+++ b/mlrun/runtimes/nuclio/function.py
@@ -1192,9 +1192,6 @@ class RemoteRuntime(KubeResource):
         return results
 
     def _resolve_invocation_url(self, path, force_external_address):
-        if not path.startswith("/") and path != "":
-            path = f"/{path}"
-
         # internal / external invocation urls is a nuclio >= 1.6.x feature
         # try to infer the invocation url from the internal and if not exists, use external.
         # $$$$ we do not want to use the external invocation url (e.g.: ingress, nodePort, etc.)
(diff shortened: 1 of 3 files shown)


mlrun/runtimes/nuclio/function.py
```diff
@@ -1192,9 +1192,6 @@ async def _invoke_async(self, tasks, url, headers, secrets, generator):
         return results
 
     def _resolve_invocation_url(self, path, force_external_address):
```
liranbg, line 1194: worth covering with UT

liranbg (changes requested):

mlrun/utils/helpers.py
```diff
@@ -1886,3 +1886,15 @@ def run_with_retry(
...
+def join_urls(base_url: str, path: str) -> str:
+    """
+    Joins a base URL with a path, ensuring proper handling of slashes.
+
+    :param base_url: The base URL (e.g., "http://example.com").
+    :param path: The path to append to the base URL (e.g., "/path/to/resource").
+
+    :return: A unified URL with exactly one slash between base_url and path.
+    """
+    return f"{base_url.rstrip('/')}/{path.lstrip('/')}" if path else base_url
```
TomerShor: why not use `urllib.parse.urljoin` instead?

TomerShor (changes requested):

moranbental (commented):
Worth adding a short PR description

mlrun/utils/helpers.py
rokatyy: @TomerShor because it doesn't work correctly if there is no http(s):// in the base_url:
image

moranbental: @rokatyy You need to validate that the `base_url` is not empty

moranbental (changes requested):

mlrun/utils/helpers.py
rokatyy: @moranbental what do you mean?

moranbental: @rokatyy 
When `base_url = None`: The code will throw an error.
When `base_url = ""`: The code will return `"/path"` (Is that the expected behavior?)

rokatyy: @moranbental type annotation require both arguments to be string, so not sure if it worth checking, but okey, will add 

reg the case when base_url is empty string, I think it's expected.

moranbental (approved):
LGTM

TomerShor (approved):

liranbg (approved):

PR #6838, opened 2024-12-03
[K8S] Quietly read secret

roei3000b:


server/py/framework/utils/singletons/k8s.py
```diff
@@ -583,7 +583,8 @@ def read_secret(
         self, secret_name: str, namespace: str = "", silent=False
     ) -> typing.Optional[client.V1Secret]:
         namespace = self.resolve_namespace(namespace)
-        logger.debug("Reading secret", secret_name=secret_name, namespace=namespace)
+        if not silent:
```
moranbental, line 586: You are reading the secret in both cases, so I think you should add debug logging when it fails to read the secret and silent=True

moranbental (changes requested):

server/py/framework/utils/singletons/k8s.py
roei3000b: See [this](https://github.com/mlrun/mlrun/pull/6793):
> We have periodic task for [refreshing the smtp configuration from the secret](https://github.com/mlrun/mlrun/blob/9bc379edb0d8756198f3d2d56b6299ce3b60d1bc/server/py/services/api/main.py#L568).
> If the secret doesn’t exists on the system the logs are flooded with Failed to retrieve k8s secret.

moranbental (approved):

liranbg (approved):

PR #6839, opened 2024-12-03
[Alert activations] Fix cross project alert activations and add test

rokatyy:
There was a bug when only names where returned with no creation time (no ticket)

This PR fixes the bug and adds a test for cross project request of activations

### DIFF
--- a/server/py/services/api/crud/projects.py
+++ b/server/py/services/api/crud/projects.py
@@ -347,11 +347,16 @@ class Projects(
             format_=mlrun.common.formatters.ProjectFormat.name_and_creation_time,
             **project_filters,
         )
-        return await framework.utils.auth.verifier.AuthVerifier().filter_projects_by_permissions(
+        allowed_project_names = await framework.utils.auth.verifier.AuthVerifier().filter_projects_by_permissions(
             [project[0] for project in projects_output.projects],
             auth_info,
             action=action,
         )
+        return [
+            project
+            for project in projects_output.projects
+            if project[0] in allowed_project_names
+        ]
 
     async def list_project_summaries(
         self,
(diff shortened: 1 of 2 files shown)


server/py/services/api/crud/projects.py
```diff
@@ -347,11 +347,16 @@ async def list_allowed_project_names_with_creation_time(
...
+        allowed_project_names = await framework.utils.auth.verifier.AuthVerifier().filter_projects_by_permissions(
             [project[0] for project in projects_output.projects],
             auth_info,
             action=action,
         )
+        return [
+            project
+            for project in projects_output.projects
+            if project[0] in allowed_project_names
+        ]
```
liranbg, line 361: worth adding comment why it is needed to add extra iteration
also, you may utilize the list once and use twice (see line 351 for dup)

liranbg (changes requested):

server/py/services/api/crud/projects.py
```diff
@@ -347,11 +347,18 @@ async def list_allowed_project_names_with_creation_time(
...
             **project_filters,
         )
-        return await framework.utils.auth.verifier.AuthVerifier().filter_projects_by_permissions(
+        allowed_project_names = await framework.utils.auth.verifier.AuthVerifier().filter_projects_by_permissions(
             [project[0] for project in projects_output.projects],
             auth_info,
             action=action,
         )
+        # Filter the original list based on allowed names
+        # we need to return full project object (not only project names)
```
TomerShor: You don't need the full project object (i.e. `Project`), you need the tuple that includes the name + creation time.
(As the function name suggests, along with its return type`list[tuple[str, datetime.datetime]]`)

TomerShor (changes requested):

server/py/services/api/crud/projects.py
rokatyy: @TomerShor full project object here means what we have in `projects_output.projects`. Will remove "full" part as it can be indeed confused with one of format types.

TomerShor (approved):

PR #6844, opened 2024-12-03
[Alerts] Fix listing alerts for all projects

TomerShor:
The `list_alerts` endpoint always ran `ensure_project` at the entrypoint, blocking the option to list alerts with `project="*"`, although the rest of the code does support this option.
For cases where `project != "*"` we do want to keep using `ensure_project` so we can fail fast with an informative message ("Project <project-name> does not exist"), but for all projects we can just skip it.


server/py/services/alerts/tests/unit/api/test_alerts.py
```diff
@@ -30,6 +30,69 @@ def test_store_alerts(self, db: Session, client: TestClient):
...
+        )
+        assert resp.status_code == HTTPStatus.OK.value
+        alerts = resp.json()
+        assert len(alerts) == 1
+        assert alerts[0]["name"] == alert_name
+
+    def test_list_alerts_for_all_projects(self, db: Session, client: TestClient):
+        for i in range(2):
+            project = f"test-alerts-{i}"
+            alert_name = f"alert-name-{i}"
```
rokatyy: Why not calling alert the same name for both project? It seems to be good edgecase to check

rokatyy (approved):
Great catch!

Really minor thing, otherwise nicely done

server/py/services/alerts/tests/unit/api/test_alerts.py
TomerShor: Done 👍

server/py/services/alerts/tests/unit/api/test_alerts.py
```diff
@@ -30,6 +30,69 @@ def test_store_alerts(self, db: Session, client: TestClient):
         project = "test-alerts"
         alert_name = "alert-name"
         self._create_project(db, project)
+        alert_config = self._create_alert_config(alert_name, project)
```
Yacouby: why not use here the generate_alert_data that we have in `alerts/tests/unit/crud/utils.py` ?
I think that it is good that we start reduce these duplications as part of the ongoing refactoring of the alerts code.

Yacouby (commented):
LGTM, minor suggestion.

moranbental (approved):

server/py/services/alerts/tests/unit/api/test_alerts.py
TomerShor: No reason, updated the test to use it.

liranbg (approved):

Yacouby (approved):

PR #6847, opened 2024-12-04
[Serving] Add `headers_expression` flag to the `RemoteStep` 

GiladShapira94:
https://iguazio.atlassian.net/browse/ML-6373

This PR adds the headers_expression flag to the RemoteStep class, that allow the user to set dynamic headers to the RemoteStep request base on the event body.


mlrun/serving/remote.py
```diff
@@ -227,7 +238,6 @@ def _generate_request(self, event, body):
                 body = self._body_function_handler(body)
             body = json.dumps(body)
             headers["Content-Type"] = "application/json"
-
```
gtopper: Please don't change the spacing.

mlrun/serving/remote.py
```diff
@@ -205,8 +215,9 @@ def do_event(self, event):
 
     def _generate_request(self, event, body):
         method = self.method or event.method or "POST"
+        if self._headers_function_handler(body):
+            self.headers = self._headers_function_handler(body)
         headers = self.headers or {}
-
```
gtopper, line 209: Please don't change the spacing.

mlrun/serving/remote.py
```diff
@@ -86,6 +87,7 @@ def __init__(
         :param retries:     number of retries (in exponential backoff)
         :param backoff_factor: A backoff factor in seconds to apply between attempts after the second try
         :param timeout:     How long to wait for the server to send data before giving up, float in seconds
+        :param headers_expression: an expression for getting the request headers from the event,e.g. "event['headers']"
```
gtopper:
```suggestion
        :param headers_expression: an expression for getting the request headers from the event, e.g. "event['headers']"
```

mlrun/serving/remote.py
```diff
@@ -205,8 +215,9 @@ def do_event(self, event):
 
     def _generate_request(self, event, body):
         method = self.method or event.method or "POST"
+        if self._headers_function_handler(body):
+            self.headers = self._headers_function_handler(body)
         headers = self.headers or {}
```
gtopper: `self.headers` shouldn't be mutated here.
```suggestion
            headers = self._headers_function_handler(body)
        else:
            headers = self.headers or {}
```

gtopper (changes requested):
Please also add or extend an existing test [here](https://github.com/mlrun/mlrun/blob/v1.8.0-rc5/tests/serving/test_remote.py#L235).

Not really a good thing that we use `eval` in this class, but that's a pre-existing issue.

mlrun/serving/remote.py
```diff
@@ -205,8 +215,10 @@ def do_event(self, event):
 
     def _generate_request(self, event, body):
         method = self.method or event.method or "POST"
-        headers = self.headers or {}
-
+        if self._headers_function_handler:
+            headers = self._headers_function_handler(body)
+        else:
+            headers = self.headers or {}
```
gtopper: Not a regression in this PR, but I'm pretty sure this should be
```suggestion
            headers = copy(self.headers) or {}
```
to avoid mutating `self.headers`.

tests/serving/test_remote.py
```diff
@@ -234,19 +234,31 @@ def test_remote_advance(httpserver, engine):
...
+                    "data": {"x": 5},
+                    "body_headers": {"Test-header": "true"},
+                }
+            }
+        )
     except Exception as e:
         raise e
     finally:
         server.wait_for_completion()
+    # resp_body = resp['body']
```
gtopper:
```suggestion
```

tests/serving/test_remote.py
```diff
@@ -234,19 +234,31 @@ def test_remote_advance(httpserver, engine):
...
+    resp__headers = resp["req"].pop("body_headers")
```
gtopper:
```suggestion
    resp_headers = resp["req"]["body_headers"]
```

tests/serving/test_remote.py
```diff
@@ -234,19 +234,31 @@ def test_remote_advance(httpserver, engine):
...
     assert resp == {"req": {"url": "/dog", "data": {"x": 5}}, "resp": {"post": "ok"}}
+    assert resp__headers["Test-header"] == "true"
```
gtopper:
```suggestion
    assert resp_headers["Test-header"] == "true"
```

tests/serving/test_remote.py
GiladShapira94: i wanted to remove the `body_headers` from the response so if i do not using `pop` the headers will be keep in the response

gtopper: Why do you want to remove it? It seems inconsequential.

GiladShapira94: the reason i wanted to remove the headers is because if i do not do it this test fail - resp == {"req": {"url": "/dog", "data": {"x": 5}}, "resp": {"post": "ok"}} , becuase the headers are in the {"req": and their values are not constant.
i thought to remove the headers and then keep the old test and check the only add key for the headers

gtopper (approved):

PR #6848, opened 2024-12-04
[API] Allow filtering schedules by its next run time

liranbg:
Implementing https://iguazio.atlassian.net/browse/ML-8362 - allowing sdk / api to filter schedules by range of next schedule runtime


roei3000b (approved):
LGTM

PR #6850, opened 2024-12-05
[ProjectSummaries] Add alert activations counters

Yacouby:
Added the following counters to the project summaries:
- endpoint_alerts_count
- job_alerts_count
- other_alerts_count

Also, added a new method to httpdb: 
`get_project_summary`: This method retrieves the summary of a specific project, including the newly added alert counters, by making an API call to fetch the data. 

This was added to test the counters of the alert activations which cant be done in unit tests since unit tests are in SQLite.
This change was tested in a system test since the update of the project resources is done in the periodic flow of the system.

Jira:
https://iguazio.atlassian.net/browse/ML-8136

Screenshot 2024-12-05 at 9 44 09

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -4836,7 +4836,9 @@ class HTTPRunDB(RunDBInterface):
             **kwargs,
         )
 
-    def get_project_summary(self, project="") -> mlrun.common.schemas.ProjectSummary:
+    def get_project_summary(
+        self, project: Optional[str] = None
+    ) -> mlrun.common.schemas.ProjectSummary:
         """
         Retrieve the summary of a project.
 
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -4836,6 +4836,20 @@ class HTTPRunDB(RunDBInterface):
             **kwargs,
         )
 
+    def get_project_summary(self, project="") -> mlrun.common.schemas.ProjectSummary:
+        """
+        Retrieve the summary of a project.
+
+        :param project: Project name for which the summary belongs.
+        :returns: A summary of the project.
+        """
+        project = project or config.default_project
+
+        endpoint_path = f"project-summaries/{project}"
+        error_message = f"Failed retrieving project summary for {project}"
+        response = self.api_call("GET", endpoint_path, error_message)
+        return mlrun.common.schemas.ProjectSummary(**response.json())
+
     @staticmethod
     def _parse_labels(
         labels: Optional[Union[str, dict[str, Optional[str]], list[str]]],
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -2952,6 +2952,9 @@ class SQLDB(DBInterface):
         dict[str, int],
         dict[str, int],
         dict[str, int],
+        dict[str, int],
+        dict[str, int],
+        dict[str, int],
     ]:
         results = await asyncio.gather(
             fastapi.concurrency.run_in_threadpool(
@@ -2974,6 +2977,10 @@ class SQLDB(DBInterface):
                 framework.db.session.run_function_with_new_db_session,
                 self._calculate_runs_counters,
             ),
+            fastapi.concurrency.run_in_threadpool(
+                framework.db.session.run_function_with_new_db_session,
+                self._calculate_alerts_activations_counters,
+            ),
         )
         (
             project_to_files_count,
@@ -2989,6 +2996,11 @@ class SQLDB(DBInterface):
                 project_to_recent_failed_runs_count,
                 project_to_running_runs_count,
             ),
+            (
+                project_to_endpoint_alerts_count,
+                project_to_job_alerts_count,
+                project_to_other_alerts_count,
+            ),
         ) = results
         return (
             project_to_files_count,
@@ -3000,6 +3012,9 @@ class SQLDB(DBInterface):
             project_to_recent_completed_runs_count,
             project_to_recent_failed_runs_count,
             project_to_running_runs_count,
+            project_to_endpoint_alerts_count,
+            project_to_job_alerts_count,
+            project_to_other_alerts_count,
         )
 
     @staticmethod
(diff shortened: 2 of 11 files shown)


server/py/framework/db/sqldb/db.py
```diff
@@ -3191,6 +3206,48 @@ def _calculate_runs_counters(
...
+        project_to_endpoint_alerts_count = collections.defaultdict(int)
+        project_to_job_alerts_count = collections.defaultdict(int)
+        project_to_other_alerts_count = collections.defaultdict(int)
+
+        last_day = datetime.now() - timedelta(hours=24)
+        alert_activations = self.list_alert_activations(
+            session=session,
+            projects_with_creation_time=projects_with_creation_time,
+            since=last_day,
+        )
```
liranbg: I suggest using group by kind and count over instead of counting using python

liranbg (changes requested):

server/py/framework/db/sqldb/db.py
```diff
@@ -3191,6 +3206,79 @@ def _calculate_runs_counters(
             project_to_running_runs_count,
         )
 
+    def _calculate_alerts_activations_counters(
```
```suggestion
    def _calculate_alert_activations_counters(
```

server/py/framework/db/sqldb/db.py
```diff
@@ -3191,6 +3206,79 @@ def _calculate_runs_counters(
...
+        project_to_job_alerts_count = collections.defaultdict(int)
+        project_to_other_alerts_count = collections.defaultdict(int)
+
+        project_filter_conditions = [
+            and_(
+                AlertActivation.project == project,
+                AlertActivation.activation_time > created,
+            )
+            for project, created in projects_with_creation_time
+        ]
```
rokatyy: This part (together with `.filter(or_(*project_filter_conditions))` can be moved to separate method and be used in both this funciton and list_alert_activations

server/py/framework/db/sqldb/db.py
```diff
@@ -3191,6 +3206,79 @@ def _calculate_runs_counters(
...
+
+        last_day = datetime.now() - timedelta(hours=24)
```
```suggestion
        last_day = mlrun.utils.datetime_now()- timedelta(hours=24)
```

server/py/framework/db/sqldb/db.py
```diff
@@ -3191,6 +3206,79 @@ def _calculate_runs_counters(
...
+                func.count(
+                    case(
+                        (
+                            ~AlertActivation.entity_kind.in_(
+                                [
+                                    mlrun.common.schemas.alert.EventEntityKind.MODEL_ENDPOINT_RESULT,
+                                    mlrun.common.schemas.alert.EventEntityKind.JOB,
+                                ]
+                            ),
+                            1,
```
rokatyy: I don't see any place where values from db request are put into variables which are returned. So I'd assume that system test will fail.

server/py/framework/db/sqldb/db.py
```diff
@@ -3191,6 +3206,79 @@ def _calculate_runs_counters(
...
+        session,
+    ) -> tuple[
+        dict[str, int],
+        dict[str, int],
+        dict[str, int],
+    ]:
+        projects_with_creation_time = self.list_projects(
+            session,
+            format_=mlrun.common.formatters.ProjectFormat.name_and_creation_time,
+        ).projects
```
rokatyy: It is not needed to request projects here, because we already do it here https://github.com/mlrun/mlrun/blob/8b5aec2a398611d99518c43d879cf88be95a8117/server/py/services/api/crud/projects.py#L440-L444. Format can be just changed from `name_only` to `name_and_creation_time` - this way instead of doing 2 request to db, we will do just one. And then projects with creation time can be passed to this function (the same as we do with `list_alert_activations`)

server/py/services/api/tests/unit/utils/projects/test_follower_member.py
```diff
@@ -466,7 +466,9 @@ async def test_list_project_summaries_fails_to_list_pipeline_runs(
         project,
     )
     framework.utils.singletons.db.get_db().get_project_resources_counters = (
-        unittest.mock.AsyncMock(return_value=tuple({project_name: i} for i in range(9)))
+        unittest.mock.AsyncMock(
+            return_value=tuple({project_name: i} for i in range(12))
+        )
```
rokatyy, line 474: why?

rokatyy (changes requested):

server/py/services/api/tests/unit/utils/projects/test_follower_member.py
Yacouby: because now get_project_resources_counters returns 12 counters and not 9

mlrun/db/httpdb.py
```diff
@@ -4767,6 +4767,20 @@ def paginated_list_alert_activations(
             **kwargs,
         )
 
+    def get_project_summary(self, project="") -> mlrun.common.schemas.ProjectSummary:
```
rokatyy:
```suggestion
    def get_project_summary(self, project: Optional[str] = None) -> mlrun.common.schemas.ProjectSummary:
```

server/py/framework/db/sqldb/db.py
```diff
@@ -3185,6 +3203,88 @@ def _calculate_runs_counters(
...
+        projects_with_creation_time: list[tuple[str, datetime]],
+    ) -> list[BinaryExpression]:
+        return [
+            and_(
+                AlertActivation.project == project,
+                AlertActivation.activation_time > created,
+            )
+            for project, created in projects_with_creation_time
+        ]
+
```
rokatyy: Please make it to get `query` as param and include ` query.filter(or_(*project_filter_conditions))` into this function.

server/py/framework/db/sqldb/db.py
```diff
@@ -3185,6 +3203,88 @@ def _calculate_runs_counters(
...
+            project_to_other_alerts_count[project] = other_counter
+
+        return (
+            project_to_endpoint_alerts_count,
+            project_to_job_alerts_count,
+            project_to_other_alerts_count,
+        )
+
+    @staticmethod
+    def _generate_project_filter_conditions(
```
liranbg: this is alert-activiation specific implementation, yet its name is very general
I suggest to move it to where it is being used or choose a proper function name

server/py/framework/db/sqldb/db.py
```diff
@@ -3185,6 +3203,88 @@ def _calculate_runs_counters(
...
+                            AlertActivation.entity_kind
+                            == mlrun.common.schemas.alert.EventEntityKind.JOB,
+                            1,
+                        )
+                    )
+                ).label("job_alerts_count"),
+                func.count(
+                    case(
+                        (
+                            ~AlertActivation.entity_kind.in_(
```
TomerShor: While this is correct, I think it's clearer using 
```suggestion
                            AlertActivation.entity_kind.not_in(
```

server/py/framework/db/sqldb/db.py
```diff
@@ -3185,6 +3203,88 @@ def _calculate_runs_counters(
...
+        project_to_job_alerts_count = collections.defaultdict(int)
+        project_to_other_alerts_count = collections.defaultdict(int)
+
+        project_filter_conditions = self._generate_project_filter_conditions(
+            projects_with_creation_time
+        )
+
+        last_day = mlrun.utils.datetime_now() - timedelta(hours=24)
+
+        query_results = (
```
TomerShor: Add a comment that explains this long query

TomerShor (changes requested):

server/py/framework/db/sqldb/db.py
```diff
@@ -3216,52 +3215,54 @@ def _calculate_alert_activations_counters(
...
+                            [
+                                mlrun.common.schemas.alert.EventEntityKind.MODEL_ENDPOINT_RESULT,
+                                mlrun.common.schemas.alert.EventEntityKind.JOB,
+                            ]
+                        ),
+                        1,
                     )
-                ).label("other_alerts_count"),
+                )
+            ).label("other_alerts_count"),
```
rokatyy, line 3253: Do we really need labels here?

rokatyy (approved):
LGTM 👍 

just minor thing

server/py/framework/db/sqldb/db.py
Yacouby: its just to improve readability, which can be helpful for debugging or any future enhancements where these labels might be referenced.

rokatyy: @Yacouby maybe worth leaving comment then so next developer don't remove it

TomerShor (approved):

PR #6851, opened 2024-12-05
[Projects] Improve performance during filtering by permissions

rokatyy:
Use set instead of list will allow to improve time complexity from:

Old: `O(m × n)`
New: `O(m + n)`

When n is number of all projects and m is number of allowed projects.

### DIFF
--- a/server/py/services/api/crud/projects.py
+++ b/server/py/services/api/crud/projects.py
@@ -354,11 +354,16 @@ class Projects(
             format_=mlrun.common.formatters.ProjectFormat.name_and_creation_time,
             **project_filters,
         )
-        allowed_project_names = await framework.utils.auth.verifier.AuthVerifier().filter_projects_by_permissions(
-            [project[0] for project in projects_output.projects],
-            auth_info,
-            action=action,
+
+        # Use a set to improve performance during filtering below
+        allowed_project_names = set(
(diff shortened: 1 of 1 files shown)


server/py/services/api/crud/projects.py
```diff
@@ -354,11 +354,16 @@ async def list_allowed_project_names_with_creation_time(
...
             **project_filters,
         )
-        allowed_project_names = await framework.utils.auth.verifier.AuthVerifier().filter_projects_by_permissions(
-            [project[0] for project in projects_output.projects],
-            auth_info,
-            action=action,
+
+        # Use a set to improve performance during filtering below
+        allowed_project_names = set(
+            await framework.utils.auth.verifier.AuthVerifier().filter_projects_by_permissions(
```
moranbental, line 360: Why not convert it to a set inside the `filter_projects_by_permissions` function? I think it would improve the performance of all the functions that call `filter_projects_by_permissions`

moranbental (changes requested):

server/py/services/api/crud/projects.py
rokatyy: @moranbental which for example? Another function that uses it is `list_allowed_project_names`, but we don't need any logic like this there because we return just list of str.

liranbg (approved):

PR #6852, opened 2024-12-05
[Microservices] Fix redundant DB session creation for forwarded requests

moranbental:
FastAPI dependencies automatically create DB sessions when entering an endpoint. These sessions are redundant when the request is going to be forwarded to services. 
To address this, before entering the endpoint, we check if the request should be forwarded by verifying whether the service that received the request has the corresponding path and method in its `method_routes`.

Also, add a PATCH request to the method_routes to allow PATCH requests

https://iguazio.atlassian.net/browse/ML-8391


server/py/framework/utils/clients/messaging.py
```diff
@@ -282,3 +289,20 @@ async def _resolve_request_kwargs_from_request(
 
         request_kwargs.update(**kwargs)
         return request_kwargs
+
+    @staticmethod
+    def _get_prefix_and_version(path: str):
+        match = PREFIX_GROUPING.match(path)
+        prefix = match.group(1)
+        version = match.group(2) or "v1"
+        return prefix, version
```
rokatyy: if doesn't match, match will be None, as a result `match.group(1)` will raise an error. is it expected?

server/py/framework/utils/clients/messaging.py
```diff
@@ -282,3 +289,20 @@ async def _resolve_request_kwargs_from_request(
...
+
+    def _prepare_request_data(self, method: str, path: str):
+        prefix, version = self._get_prefix_and_version(path)
+
+        # Remove the service and version prefix from the path
+        # The service prefix is to be replaced with the new service name
+        # The version will be re-added or default to v1 if not present
```
rokatyy: This comment should be part of  `_get_prefix_and_version`

server/py/framework/utils/clients/messaging.py
```diff
@@ -151,6 +145,19 @@ async def convert_requests_response_to_fastapi_response(
...
+    def should_forward_request(self, request: fastapi.Request) -> bool:
+        """
+        Checks whether the request should be forwarded to another service based on
+        the service and path being resolved.
+
+        :param request: The incoming FastAPI request.
+        :return: True if the request should be forwarded, False otherwise.
+        """
+        method = request.method
+        path = str(request.url.path)
```
rokatyy: isn't it already a str?

rokatyy (commented):
Looking very nice!

Couple of suggestions.

server/py/framework/utils/clients/messaging.py
moranbental: @rokatyy Actually, it is related to the `proxy_request` method, so I moved it there

liranbg (approved):

server/py/framework/utils/clients/messaging.py
```diff
@@ -282,3 +293,19 @@ async def _resolve_request_kwargs_from_request(
...
         return request_kwargs
+
+    @staticmethod
+    def _get_prefix_and_version(path: str):
+        match = PREFIX_GROUPING.match(path)
+        if not match:
+            raise ValueError(f"Invalid path format: {path}")
+
+        prefix = match.group(1)
+        version = match.group(2) or "v1"  # default to v1 if not present
```
rokatyy:
```suggestion
        # default to v1 if not present
        version = match.group(2) or "v1"
```

rokatyy (approved):
Minor nitpicking, otherwise approved 👍

server/py/framework/api/deps.py
```diff
@@ -25,14 +25,19 @@
...
-    try:
-        db_session = framework.db.session.create_session()
-        yield db_session
-    finally:
-        if db_session:
-            framework.db.session.close_session(db_session)
+def get_db_session(
+    request: Request,
+) -> typing.Generator[typing.Optional[Session], None, None]:
+    if not request.app.extra.get("mlrun_service").should_forward_request(request):
```
TomerShor: Add a comment on WHY you are checking this.

server/py/framework/service/__init__.py
```diff
@@ -349,6 +351,15 @@ async def _align_worker_state_with_chief_state(
             self._synchronize_with_chief_clusterization_spec.__name__
         )
 
+    def should_forward_request(self, request: fastapi.Request) -> bool:
```
TomerShor:
```suggestion
    def is_forwarded_request(self, request: fastapi.Request) -> bool:
```

server/py/framework/utils/clients/messaging.py
```diff
@@ -151,6 +149,19 @@ async def convert_requests_response_to_fastapi_response(
             media_type="application/json",
         )
 
+    def should_forward_request(self, request: fastapi.Request) -> bool:
```
TomerShor:
```suggestion
    def is_forwarded_request(self, request: fastapi.Request) -> bool:
```

TomerShor (changes requested):
Nice stuff.
Minor comments

PR #6853, opened 2024-12-05
[API] Optimize project resource verification prior deletion

liranbg:
in addition, drop obsoleted logs table
https://iguazio.atlassian.net/browse/ML-8570


server/py/framework/db/sqldb/db.py
```diff
@@ -3262,41 +3247,34 @@ def _get_project_record(
...
-        logs = self._list_logs(session, name)
-        self._verify_empty_list_of_project_related_resources(name, logs, "logs")
-        runs = self._find_runs(session, None, name, []).all()
-        self._verify_empty_list_of_project_related_resources(name, runs, "runs")
-        notifications = []
+        # it is enough to sample few resources, we do not need to retrieve all resources really.
+        resource_limit = 5
+        for resource_name, resource_list_function, kwargs in [
+            ("runs", self.list_runs, dict(sort=False, return_as_run_structs=False)),
+            ("artifacts", self._find_artifacts, dict(ids="*")),
```
quaark: Do these params matter?
`sort` definitely doesn't matter here, also `return_as_run_structs` doesn't matter as you don't care about the type of the returned resources only if the list is empty. and artifacts' `ids=*` is the same as `ids=None`.
Looks like you don't necessarily need any kwargs in this for loop, you only need resource name and method.

quaark (changes requested):

PR #6854, opened 2024-12-05
[Alert activations] Change name for alert_activations table from singular to plural

rokatyy:
Follow common conventions for table names.


yaelgen (approved):
LGTM

liranbg (changes requested):
for existing RCs you need to add rename for table in case you have table with singular

TomerShor (approved):

PR #6861, opened 2024-12-08
[Alerts] Fix resetting alert for case when full_object is None

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-8592

In case if alert was already reset and not activated after, full_object is None. So, when we do 

```
alert_state = {"full_object":None}
activation_id = alert_state.get("full_object", {}).get("last_activation_id")
```

we will get `AttributeError: 'NoneType' object has no attribute 'get'`


server/py/services/alerts/crud/alerts.py
```diff
@@ -421,7 +421,10 @@ def reset_alert(self, session: sqlalchemy.orm.Session, project: str, name: str):
                 session, alert.id
             )
             activation_time = alert_state.get("last_updated")
-            activation_id = alert_state.get("full_object", {}).get("last_activation_id")
+            # or {} is needed of the case if full_object is None
+            activation_id = (alert_state.get("full_object") or {}).get(
+                "last_activation_id"
+            )
```
TomerShor: if `alert_state` is None then getting `"last_updated"` will also raise the same exception.
You can instead do:
```python
            alert_state = alert_state or {}
            activation_time = alert_state.get("last_updated")
            activation_id = alert_state.get("full_object", {}).get("last_activation_id")
```

TomerShor (changes requested):

server/py/services/alerts/crud/alerts.py
rokatyy: @TomerShor the problem here is not that alert_state is empty but `alert_state["full_object""]` is None. 
This string `activation_id = alert_state.get("full_object", {}).get("last_activation_id")` you suggested is something we better avoid (see description of PR why).

Regarding empty `alert_state`, technically it's not possible to get `None` here as we use `.one()` expression https://github.com/mlrun/mlrun/blob/06c5da78dc4f539f85f9a4466387ee86b99f909a/server/py/framework/db/sqldb/db.py#L5925 to get record from db. But I agree that it's better to be on safe side, so I'll add a warning.

TomerShor (approved):
Great 👍

server/py/services/alerts/crud/alerts.py
TomerShor: It is confusing because looking at `get_alert_state_dict` it seems that you can get None, but I see why this is not the case.
😓

PR #6866, opened 2024-12-09
[Docs] Modify MonitoringApplicationContext docstring 

jillnogold:
Add details about the _init_ function

### DIFF
--- a/mlrun/model_monitoring/applications/context.py
+++ b/mlrun/model_monitoring/applications/context.py
@@ -84,7 +84,8 @@ class MonitoringApplicationContext:
         artifacts_logger: Optional[_ArtifactsLogger] = None,
     ) -> None:
         """
-        Initialize a :code:`MonitoringApplicationContext` object.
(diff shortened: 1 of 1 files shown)


assaf758 (approved):

mlrun/model_monitoring/applications/context.py
```diff
@@ -86,7 +86,8 @@ def __init__(
         feature_stats: Optional[FeatureStats] = None,
     ) -> None:
         """
-        Initialize a :code:`MonitoringApplicationContext` object.
+        The :code:`_init_` function initializes a :code:`MonitoringApplicationContext` object
+        and has the following attributes.
```
jond01: This is a _method_ (function of a class) and it's _`__init__`_. However, this explanation adds very little information IMO

jillnogold: The problem is that currently the params show but there's no context
image
Made your fixes. Thanks!

jond01 (approved):

PR #6868, opened 2024-12-09
[Alert activations] Add group by and aggregate by for alert activations

rokatyy:
Adds `AlertActivations` object to provide user ability to group and aggregate value. Inspired by the way these methods work in spark/pandas, so it be more intuitive for users who are familiar with them. 
 
Jira - https://iguazio.atlassian.net/browse/ML-8142

### DIFF
--- a/mlrun/common/schemas/alert.py
+++ b/mlrun/common/schemas/alert.py
@@ -12,8 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from collections import defaultdict
+from collections.abc import Iterator
 from datetime import datetime
-from typing import Annotated, Optional, Union
+from typing import Annotated, Any, Callable, Optional, Union
 
 import pydantic.v1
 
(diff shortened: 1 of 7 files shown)


mlrun/common/schemas/alert.py
```diff
@@ -216,3 +218,77 @@ class AlertActivation(pydantic.v1.BaseModel):
...
+        - If there's only one attribute, return the value directly (not a single-element tuple).
+        - If there are multiple attributes, return them as a tuple for grouping.
+
+        This ensures grouping behaves intuitively without redundant tuple representations.
+        """
+        if len(attributes) == 1:
+            # Avoid single-element tuple like (high,) when only one grouping attribute is used
+            return getattr(self, attributes[0])
+        # Otherwise, return a tuple of all specified attributes
+        return tuple(getattr(self, attr) for attr in attributes)
```
TomerShor, line 234: What about error handling? when the activation doesn't have one of the requested attribute?
Or maybe we are ok with `AttributeError`

mlrun/common/schemas/alert.py
```diff
@@ -216,3 +218,77 @@ class AlertActivation(pydantic.v1.BaseModel):
     number_of_events: int
     notifications: list[notification_objects.NotificationState]
     reset_time: Optional[datetime] = None
+
+    def group_key(self, attributes: list[str]) -> tuple:
```
TomerShor: Is the return value a tuple even when `len(attributes) == 1`?

TomerShor (changes requested):
That's a neat one :)

mlrun/common/schemas/alert.py
rokatyy: @TomerShor I think it's ok to raise attribute error in this case

TomerShor (approved):

PR #6876, opened 2024-12-10
[Chief] Do not fail MLRun chief on not getting the partition expression

rokatyy:
If partition interval isn't found, we want to determine the reason and raise the error.

### DIFF
--- a/server/py/framework/db/base.py
+++ b/server/py/framework/db/base.py
@@ -932,6 +932,13 @@ class DBInterface(ABC):
     ) -> str:
         pass
 
+    @staticmethod
+    def table_exist(
+        session,
+        table_name: str,
+    ) -> bool:
+        pass
+
     @abstractmethod
     def delete_alert(self, session, project: str, name: str):
         pass
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -5829,6 +5829,29 @@ class SQLDB(DBInterface):
         ).scalar()
 
     @staticmethod
+    def table_exist(
+        session: Session,
+        table_name: str,
+    ) -> bool:
+        """
+        Checks if a table exists in the current database schema.
+
+        :param session: SQLAlchemy session.
+        :param table_name: Name of the table to check.
+
+        :return: True if the table exists, False otherwise.
+        """
+        result = session.execute(
+            text(
+                "SELECT COUNT(*) "
+                "FROM information_schema.tables "
+                "WHERE TABLE_NAME = :table_name "
+            ),
+            {"table_name": table_name},
+        ).scalar()
+        return result > 0
+
+    @staticmethod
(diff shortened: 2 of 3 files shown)


server/py/services/api/main.py
```diff
@@ -547,21 +547,27 @@ def _start_periodic_partition_management(self):
...
+                    False,
+                    self._manage_partitions,
+                    table_name=table_name,
+                    retention_days=retention_days,
+                )
+            except mlrun.errors.MLRunValueError:
+                self._logger.warning(
+                    "Failed to start retention management for table",
+                    table_name=table_name,
+                )
```
TomerShor: What does it mean that there is no partition interval? Doesn't it mean that something is wrong?

rokatyy: @TomerShor it does mean, but if we raise error here, then chief will exit which sounds like not the best idea

TomerShor: But if we don't fail here:
- Something else will fail after MLRun starts
- We won't manage the partitions at all which is also not a good idea.

Can you say what can cause the partition interval to be missing (other than between RCs)?
Also, perhaps instead of ignoring we can use a default partition interval

rokatyy: @TomerShor
> Can you say what can cause the partition interval to be missing (other than between RCs)?

 if table is not there. For instance, because of the wrong configuration of mlrun. Or since we define partitioned tables in the config, I can imagine a situation where config from newer mlrun version was taken and imagine we have not only alert_activations, but some other table (`table_x` ) in retention_objects. But mlrun version is older and don't have this table yet. Chief will exit with error without this fix.

TomerShor: In that case I can argue the same thing regarding wrong configuration [here](https://github.com/mlrun/mlrun/blob/da480edda6b9a8a09e7211260b1f86e673a95038/mlrun/config.py#L1081-L1090).
Why would we fail in such cases and in other not?

I think its best if we can differentiate between an actual error and a misconfiguration.
I think the case where a table exists, its set in the `object_partitions` but it doesn't have a partition expression is an error we should not necessarily ignore 
If the table doesn't exist then maybe it's ok to ignore.

I understand that ignoring the exception fixes the bug, but we need to be understand what does it mean when we ignore it.

rokatyy: @TomerShor the problem in this specific case is that table doesn't exist, so we can't do any retention/partitioning on it. 

And it's not really a bug, it just the question of what we do in this case.

rokatyy: @TomerShor actually, maybe it would be better to raise an error here and make user change mlrun config (remove the wrong table) instead of ignoring it. wdyt?

TomerShor: I think that we should raise the error. An informative one though so it's clear what happened

TomerShor (approved):
LGTM, just updated the PR description accordingly

PR #6677, opened 2024-11-10
[Requirements] Add a lock-file for mlrun-api Python dependencies

jond01:
Use uv 0.5.5: https://docs.astral.sh/uv/.

Implements [ML-8383](https://iguazio.atlassian.net/browse/ML-8383).

Apart from locking the dependencies for reproducibility, it also:
* Significantly accelerates the `make api` **build time** to 30% of the time with pip (no cache).
* Reduces the **image size** to 68% the size with pip.

The `mlrun/mlrun-api` Docker image figures:

| tool                 | pip  | uv   |
|----------------------|------|------|
| build time (minutes) | 4:55 | 1:29 |
| size (GB)            | 4.37 | 2.96 |

From my local testing - the Docker image works perfectly fine.

[ML-4648]: https://iguazio.atlassian.net/browse/ML-4648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ML-8383]: https://iguazio.atlassian.net/browse/ML-8383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6886, opened 2024-12-11
[Images] Add a lock-file for mlrun/mlrun container image

liranbg:
Continuing the work started by https://github.com/mlrun/mlrun/pull/6677 - adding mlrun/mlrun image to the party.

This PR also brings
1. removes uv from the final image
2. Utilizing the docker host cache to optimize builds
3. Uses `--from` directive instead of ADD which has better caching optimization by docker

Next in line are mlrun/base, mlrun/test* images, mlrun/jupyter, mlrun/mlrun-gpu , mlrun/common

https://iguazio.atlassian.net/browse/ML-8382


dockerfiles/mlrun-api/Dockerfile
```diff
@@ -41,21 +44,29 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
...
 
-RUN sh /uv-installer.sh && rm /uv-installer.sh
+WORKDIR /mlrun
 
-ENV PATH="/root/.local/bin/:$PATH"
-ENV UV_SYSTEM_PYTHON=true
+RUN --mount=from=uv-image,source=/uv,target=/bin/uv \
+    --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=dockerfiles/mlrun-api/locked-requirements.txt,target=locked-requirements.txt \
+    uv pip sync --require-hashes locked-requirements.txt
```
TomerShor, line 54: if this ( + the UV image initialization) is going to be the same in all mlrun images, perhaps we can create a custom base image that will be used as a `FROM` for all of them?
not critical for this PR.

liranbg: that needs to have a bigger refactor but indeed could be optimized using bytecode compilation

Makefile
```diff
@@ -804,8 +807,22 @@ verify-uv-version:
...
+upgrade-mlrun-mlrun-deps-lock: verify-uv-version ## Upgrade mlrun-mlrun locked requirements file
+	uv pip compile \
+		requirements.txt \
+		extras-requirements.txt \
+		dockerfiles/mlrun/requirements.txt \
+		$(MLRUN_UV_UPGRADE_FLAG) \
+		--output-file dockerfiles/mlrun/locked-requirements.txt
+
+.PHONY: upgrade-mlrun-mlrun-deps-lock
+upgrade-mlrun-deps-lock: verify-uv-version ## Upgrade mlrun-* locked requirements file
```
jond01, line 826: The phony name doesn't match the target

dockerfiles/mlrun-api/Dockerfile
```diff
@@ -13,6 +13,9 @@
 # limitations under the License.
 
 ARG MLRUN_PYTHON_VERSION=3.9
+ARG MLRUN_UV_IMAGE=ghcr.io/astral-sh/uv:latest
```
jond01, line 16: Don't we want to pin it also in the Dockerfile? It's more robust than taking the latest version.

dockerfiles/mlrun/Dockerfile
```diff
@@ -21,14 +24,6 @@ RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get -y upgrade && \
     rm -rf /var/lib/apt/lists/*
 
-# Update base pip packages (usually pip, setuptools, wheel):
-RUN python -m pip  list --disable-pip-version-check --format freeze \
```
jond01, line 25: Don't we still want pip to be upgraded?

liranbg: happen on line 64

PR #6911, opened 2024-12-15
[Docs] Add artifacts to docs references section

assaf758:
[ML-8605](https://iguazio.atlassian.net/browse/ML-8605)

[ML-8605]: https://iguazio.atlassian.net/browse/ML-8605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/api/mlrun.artifacts/index.rst
+++ b/docs/api/mlrun.artifacts/index.rst
@@ -3,28 +3,8 @@
 mlrun.artifacts
 ===============
 
-.. automodule:: mlrun.artifacts
-   :members:
-   :show-inheritance:
-   :undoc-members:
+.. toctree::
+  :maxdepth: 1
+  :glob:
 
-.. automodule:: mlrun.artifacts.dataset
-   :members:
-   :show-inheritance:
-   :undoc-members:
-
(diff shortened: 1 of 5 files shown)


alxtkr77:
We need to make guidelines how to add things to documentation. One way is to use the index.rst, the other is using mlrun/artifacts/__init__.py::__all__ directive. I must say it is confusing

assaf758:
@rokatyy can you chime in on the correct way to do this?

rokatyy:
@assaf758 there were a lot of discussions regarding the best way to do this. For now please follow the way it is done here https://github.com/mlrun/mlrun/blob/046b46e4dba4352045085f3df829e11a609708c0/docs/api/mlrun.common/mlrun.common.schemas to keep the nested structure the same. In the future we will probably move to using autosummary, but for now we do everything manually.

rokatyy (changes requested):
Please move each module to a separate file as in the example above

alxtkr77 (approved):

PR #6923, opened 2024-12-16
[System id] Create system metadata table

Yacouby:
This PR introduces the `SystemMetadata` table. 
The table is designed to store system-level metadata (such as system_id that will be added in a following PR), and provides a flexible structure to accommodate various key-value pairs for system configurations.

Jira : 
https://iguazio.atlassian.net/browse/ML-8713


rokatyy (commented):
Some general questions regarding implementation. 

1. why do we need `id` if `key` is already supposed to be unique? 
2. If we do need `id`, I assume that it should be autoincrementable
3. why `key` is nullable?

Yacouby:
> 1. why do we need `id` if `key` is already supposed to be unique?
> 2. If we do need `id`, I assume that it should be autoincrementable
> 3. why `key` is nullable?

@rokatyy 
1. the id column serves as a primary key for efficient indexing and querying,  as integer-based are faster than string-based ones. also the id provides an internal identifier for each row, which IMO is more scalable than using the key itself.
2.  since id is defined as the primary key and is an integer, it is automatically auto-incremented. 
3. we can deploy the system without having any key in the table.

Yacouby:
> Some general questions regarding implementation.
> 
> 1. why do we need `id` if `key` is already supposed to be unique?
> 2. If we do need `id`, I assume that it should be autoincrementable
> 3. why `key` is nullable?

rokatyy:
@Yacouby 
1. Not sure how they will be used in indexing/querying (especially if we plan to always request by only `key` field), but in general I think it's alright to use both, taking into account that amount of data is very small anyway
2. 👍 
3. then we won't have any rows in the table, right? But if we add a record, data should not be empty

TomerShor (approved):

server/py/framework/db/sqldb/models.py
```diff
@@ -968,6 +968,17 @@ class ModelEndpoint(Base, mlrun.utils.db.HasStruct):
         def get_identifier_string(self) -> str:
             return f"{self.project}_{self.name}_{self.created}"
 
+    class SystemMetadata(Base, mlrun.utils.db.BaseModel):
+        __tablename__ = "system_metadata"
+        __table_args__ = (UniqueConstraint("key", name="_system_metadata_uc"),)
+
+        id = Column(Integer, primary_key=True)
+        key = Column(String(255, collation=SQLTypesUtil.collation()), nullable=False)
+        value = Column(String(255, collation=SQLTypesUtil.collation()))
```
rokatyy: In which cases value supposed to be nullable?

rokatyy (commented):
Last question, otherwise looks good

server/py/framework/db/sqldb/models.py
Yacouby: Since this table is generic, in future use we may want to allow certain keys to exist without an associated value. 
This provides flexibility to handle scenarios where a key acts as a flag, placeholder, or where the value is optional depending on the context..

rokatyy: @Yacouby then why do we need to keep key if value is not there?

Yacouby: on a second thought if adding a boolean key, then the value should be True/False, then anyway it should not be null.. updated

rokatyy (approved):

TomerShor (approved):

PR #6925, opened 2024-12-16
[Function] Fix update function flow

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-8495

Fixes a regression bug which causes empty `kind` field after function update.

### DIFF
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -2076,10 +2076,9 @@ class SQLDB(DBInterface):
             struct = function.struct
             for key, val in updates.items():
                 update_in(struct, key, val)
-            function.kind = struct.pop("kind", None)
+            function.kind = struct.pop("kind", None) if not function.kind else function.kind
             function.struct = struct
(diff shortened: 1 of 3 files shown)


server/py/framework/db/sqldb/db.py
```diff
@@ -2076,10 +2076,11 @@ def update_function(
             struct = function.struct
             for key, val in updates.items():
                 update_in(struct, key, val)
-            function.kind = struct.pop("kind", None)
+            function.kind = (
+                struct.pop("kind", None) if not function.kind else function.kind
+            )
             function.struct = struct
             self._upsert(session, [function])
-            return function.struct
```
TomerShor, line 2082: why not return it?

rokatyy: @TomerShor just use `get_function` then, it aligns flow with store_function which also doesn't return function.

moranbental (approved):
LGTM

liranbg (approved):

TomerShor (approved):

PR #6927, opened 2024-12-16
[Alerts] Make cache safer

rokatyy:
https://iguazio.atlassian.net/browse/ML-8703?quickReply=onIt

This error is really hard to reproduce (I caught it only once), it requires storing same alert at the same time. 
This PR ensures that no races happen.


liranbg (approved):
lgtm for the fix. worth moving to set to ensure cache invalidation of o(1)

PR #6934, opened 2024-12-17
[CI] Release tutorials add assets

daniels290813:
welcome.ipynb and README.md requires additional images, adding the assets folder from the `v3io/tutorials` repo.
Its suggest to use `git clone` and not wget in order to fetch the images folder


daniels290813:
plus, the `mlrun-release-stable` branch is out of sync, using a different branch didn't quite work, default branch is updated @alonmr FYI (if you recall we spoke about that)

PR #6936, opened 2024-12-17
[Alert activations] Fix partitioning for day and month

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-8768

the problem was that DAYOFMONTH and MONTH (probably, wasn't fully tested) partition expressions weren't entirely correct. As when we create partitions we create them with condition ` VALUES LESS THAN ({next_partition_value}`.

For `YEARWEEK` expression, the real generated value is:
```
mysql> select yearweek(now());
+-----------------+
| yearweek(now()) |
+-----------------+
|          202450 |
+-----------------+
```

When for dayofmonth():
```
mysql> select dayofmonth(now());
+-------------------+
| dayofmonth(now()) |
+-------------------+
|                17 |
+-------------------+
```

it is a single number, which means that the records will be always put to the oldest partition (because any day of the month 1-31 is always less than next_parititon_value).

This PR changes the expressions from dayofmonth or month to this for `DAY`:
```
mysql> select YEAR(NOW())*10000 + MONTH(NOW())*100 + DAY(NOW());
+---------------------------------------------------+
| YEAR(NOW())*10000 + MONTH(NOW())*100 + DAY(NOW()) |
+---------------------------------------------------+
|                                          20241217 |
+---------------------------------------------------+
```
And this for `month`:

```
SELECT YEAR('2025-01-01') * 100 + MONTH('2025-01-01');
+------------------------------------------------+
| YEAR('2025-01-01') * 100 + MONTH('2025-01-01') |
+------------------------------------------------+
|                                         202501 |
+------------------------------------------------+
```

(*) Also fixed PartitionInterval to be more general and do not hardcode `activation_time` as a column name.


mlrun/common/schemas/partition.py
```diff
@@ -120,11 +120,13 @@ def get_partition_name(self, current_datetime: datetime) -> str:
...
+    def get_partition_expression(self, column_name: str):
         if self == PartitionInterval.YEARWEEK:
-            return "YEARWEEK(activation_time, 1)"
-        else:
-            return f"{self}(activation_time)"
+            return f"YEARWEEK({column_name}, 1)"
+        elif self == PartitionInterval.DAY:
+            return f"YEAR({column_name}) * 10000 + MONTH({column_name}) * 100 + DAY({column_name})"
+        elif self == PartitionInterval.MONTH:
+            return f"YEAR({column_name}) * 100 + MONTH({column_name})"
```
TomerShor: I'd add a comment on why you do these multiplications

mlrun/common/schemas/partition.py
```diff
@@ -46,24 +46,24 @@ def as_duration(self) -> timedelta:
             return timedelta(weeks=1)
 
     @classmethod
-    def from_function(cls, partition_function: str):
+    def from_function(cls, partition_expression: str):
```
TomerShor: why not `expression` all the way then?
```suggestion
    def from_expression(cls, partition_expression: str):
```

server/py/services/api/migrations/versions/d78bd07756c1_fix_alert_activation_partition_.py
```diff
@@ -0,0 +1,80 @@
...
+            column_name="activation_time"
+        )
+
+        # Construct SQL for partitioning
+        partition_sql = f"""
+            ALTER TABLE alert_activations
+            PARTITION BY RANGE ({partition_expression}) (
+                PARTITION p{partition_name} VALUES LESS THAN ({partition_value})
+            );
+        """
```
TomerShor: For new systems (e.g. on the final 1.8.0) the partitioning will already be ok, since the original migration is fixed too.
Maybe we can validate this and skip this migration if it isn't needed.

TomerShor (changes requested):
Nice fix! some minor comments

mlrun/common/schemas/partition.py
rokatyy: @TomerShor good point, will rename it

server/py/services/api/migrations/versions/d78bd07756c1_fix_alert_activation_partition_.py
rokatyy: @TomerShor yeah, even for most of our systems (except those two configured with partition interval - mine and QA's) it will be alright. So in general I think it can be removed fully and just be well-communicated with QA team. wdyt?

TomerShor: These statements are true, right?
1. Default value is YEARWEEK, So most systems won't have an issue at all
2. We don't allow changing partition value unless starting mlrun from scratch

If that's the case I think we will be ok without it.

rokatyy: @TomerShor yes, that's absolutely right

okay, then removing the migration

mlrun/common/schemas/partition.py
```diff
@@ -46,24 +46,24 @@ def as_duration(self) -> timedelta:
...
     @classmethod
-    def from_function(cls, partition_function: str):
+    def from_expression(cls, partition_expression: str):
         """
         Returns the corresponding PartitionInterval for a given partition function,
         or None if the function is not mapped.
 
-        :param partition_function: The partition function to map to an interval.
+        :param partition_expression: The partition function to map to an interval.
         :return: PartitionInterval corresponding to the function, or None if no match is found.
```
TomerShor:
```suggestion
        :param partition_expression: The partition expression to map to an interval.
        :return: PartitionInterval corresponding to the expression, or None if no match is found.
```

TomerShor (approved):
LGTM

PR #6941, opened 2024-12-18
[Notification] Do not assume run labels

liranbg:
Run labels are set when run is either running or finished and can have a race condition with api running in the background sending notification when the run is failed or completed.

the race condition was originally happened because of session being reused (fixed on https://github.com/mlrun/mlrun/pull/6792) but the side effect here needs to be mitigated as well

https://iguazio.atlassian.net/browse/ML-8781


yaelgen (approved):
LGTM, maybe you can add a test case for a run that doesn't include metadata labels, to test this case

PR #6942, opened 2024-12-18
[System id] Add system id initialization

Yacouby:
This PR introduces the initialization and management of a unique system ID in the database.

Added logic to initialize and persist a unique 6-character base64-encoded system ID, following this order during initialization:

- Database value 
- Configuration
- Newly generated ID

Example of the database state after initializing the system id in a fresh deployment:

image

Jira :
https://iguazio.atlassian.net/browse/ML-8716

### DIFF
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -7086,8 +7086,8 @@ class SQLDB(DBInterface):
             return None
         return system_id_record.value
 
-    def create_system_id(self, session: Session, system_id: str):
-        logger.debug("Creating a new system id in DB", system_id=system_id)
+    def store_system_id(self, session: Session, system_id: str):
+        logger.debug("Storing a new system id in DB", system_id=system_id)
 
         system_id_record = SystemMetadata(key="system_id", value=system_id)
         self._upsert(session, [system_id_record])
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -78,6 +78,7 @@ from mlrun.utils import (
     validate_tag_name,
 )
 
+import framework.constants
 import framework.db.session
 import framework.utils.helpers
 from framework.db.base import DBInterface
@@ -7078,18 +7079,17 @@ class SQLDB(DBInterface):
     def get_system_id(self, session: Session) -> typing.Optional[str]:
         system_id_record = (
             self._query(session, SystemMetadata)
-            .filter(SystemMetadata.key == "system_id")
+            .filter(SystemMetadata.key == framework.constants.SYSTEM_ID_KEY)
             .one_or_none()
         )
-        if not system_id_record:
-            logger.debug("System id not found in DB")
-            return None
-        return system_id_record.value
+        return system_id_record.value if system_id_record else None
 
     def store_system_id(self, session: Session, system_id: str):
         logger.debug("Storing a new system id in DB", system_id=system_id)
 
-        system_id_record = SystemMetadata(key="system_id", value=system_id)
+        system_id_record = SystemMetadata(
+            key=framework.constants.SYSTEM_ID_KEY, value=system_id
+        )
         self._upsert(session, [system_id_record])
 
     # ---- Utils ----
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -115,6 +115,7 @@ from framework.db.sqldb.models import (
     ProjectSummary,
     Run,
     Schedule,
+    SystemMetadata,
     TimeWindowTracker,
     _labeled,
     _tagged,
@@ -7073,6 +7074,23 @@ class SQLDB(DBInterface):
             main_table_identifier=ModelEndpoint.uid if uids else None,
             main_table_identifier_values=uids,
         )
+
+    def get_system_id(self, session: Session) -> typing.Optional[str]:
+        system_id_record = (
+            self._query(session, SystemMetadata)
+            .filter(SystemMetadata.key == "system_id")
+            .one_or_none()
+        )
+        if not system_id_record:
+            logger.debug("System id not found in DB")
+            return None
+        return system_id_record.value
+
+    def create_system_id(self, session: Session, system_id: str):
+        logger.debug("Creating a new system id in DB", system_id=system_id)
+
+        system_id_record = SystemMetadata(key="system_id", value=system_id)
+        self._upsert(session, [system_id_record])
 
     # ---- Utils ----
     def delete_table_records(
--- a/server/py/services/api/initial_data.py
+++ b/server/py/services/api/initial_data.py
@@ -101,24 +101,20 @@ def init_data(
     logger.info("Creating initial data")
     config.httpdb.state = mlrun.common.schemas.APIStates.migrations_in_progress
 
-    if is_migration_from_scratch or is_migration_needed:
-        try:
-            _perform_schema_migrations(alembic_util)
-            init_db()
-            db_session = create_session()
-            try:
-                _add_initial_data(db_session)
-                _perform_data_migrations(db_session)
-            finally:
-                close_session(db_session)
-        except Exception:
-            state = mlrun.common.schemas.APIStates.migrations_failed
-            logger.warning("Migrations failed, changing API state", state=state)
-            config.httpdb.state = state
-            raise
-
     db_session = create_session()
     try:
+        if is_migration_from_scratch or is_migration_needed:
+            try:
+                _perform_schema_migrations(alembic_util)
+                init_db()
+                _add_initial_data(db_session)
+                _perform_data_migrations(db_session)
+            except Exception:
+                state = mlrun.common.schemas.APIStates.migrations_failed
+                logger.warning("Migrations failed, changing API state", state=state)
+                config.httpdb.state = state
+                raise
+
         # initialize system id
         _init_system_id(db_session)
     finally:
@@ -1003,9 +999,12 @@ def _create_project_summaries(db, db_session):
 
 
 def _init_system_id(db_session: sqlalchemy.orm.Session):
-    # Initializes a system id for MLRun deployment.
-    # The system id is first checked in the database. If it does not exist, the function checks for a configured id
-    # (from environment variables or config), and if neither is found, a new random one is generated and stored.
+    """
+    Initializes a system id for MLRun deployment.
+    The system id is first checked in the database. If it does not exist, the function checks for a configured id
+    (from environment variables or config), and if neither is found, a new random one is generated and stored.
+    """
+
     db = framework.db.sqldb.db.SQLDB()
 
     # check if a system id already exists in the database
@@ -1013,32 +1012,30 @@ def _init_system_id(db_session: sqlalchemy.orm.Session):
 
     if system_id is not None:
         logger.info("Existing system id found in the database", system_id=system_id)
-        return
+    else:
+        # check if the system id is already set in the config
+        system_id = _get_configured_system_id()
 
-    # check if the system id is configured via environment variables or configmap
-    system_id = _get_configured_system_id()
+        if system_id is not None:
+            logger.info("Using configured system id", system_id=system_id)
+            db.store_system_id(db_session, system_id)
+        else:
+            # if no system id is found, generate a new one
+            system_id = _generate_system_id()
+            db.store_system_id(db_session, system_id)
 
-    if system_id is not None:
-        logger.info("Using configured system id", system_id=system_id)
-        db.create_system_id(db_session, system_id)
-        return
-
-    # if no system id is found, generate a new one
-    system_id_str = _generate_system_id()
-    db.create_system_id(db_session, system_id_str)
+    # set the system id in mlrun config
+    mlrun.mlconf.system_id = system_id
 
 
 def _get_configured_system_id() -> typing.Optional[str]:
-    system_id = os.environ.get("SYSTEM_ID")
-    if system_id:
-        return system_id
     return mlrun.mlconf.system_id or None
 
 
 def _generate_system_id() -> str:
-    # Generate a random 32-bit unsigned integer and encode it as base64 string without padding
-    random_int = os.urandom(4)
-    base64_str = base64.b64encode(random_int).decode("utf-8").rstrip("=")
+    # Generate 4 random bytes and encode them as base64 string without padding
+    random_bytes = os.urandom(4)
+    base64_str = base64.b64encode(random_bytes).decode("utf-8").rstrip("=")
     return base64_str
 
 
--- a/server/py/services/api/initial_data.py
+++ b/server/py/services/api/initial_data.py
@@ -17,6 +17,7 @@ import datetime
 import json
 import os
 import pathlib
+import random
 import typing
 
 import dateutil.parser
@@ -1001,8 +1002,8 @@ def _create_project_summaries(db, db_session):
 def _init_system_id(db_session: sqlalchemy.orm.Session):
     """
     Initializes a system id for MLRun deployment.
-    The system id is first checked in the database. If it does not exist, the function checks for a configured id
-    (from environment variables or config), and if neither is found, a new random one is generated and stored.
+    The system id is first checked in the database. If it does not exist, the function checks if an id was set in the
+    config, and if neither is found, a new random one is generated and stored.
     """
 
     db = framework.db.sqldb.db.SQLDB()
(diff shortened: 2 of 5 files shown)


server/py/services/api/initial_data.py
```diff
@@ -115,6 +116,14 @@ def init_data(
             logger.warning("Migrations failed, changing API state", state=state)
             config.httpdb.state = state
             raise
+
+    db_session = create_session()
```
TomerShor: What about moving this to before the 'is_migration_needed', so we cam reuse the same session for both operations?

server/py/services/api/initial_data.py
```diff
@@ -993,6 +1002,46 @@ def _create_project_summaries(db, db_session):
     db._upsert(db_session, project_summaries, ignore=True)
 
 
+def _init_system_id(db_session: sqlalchemy.orm.Session):
+    # Initializes a system id for MLRun deployment.
+    # The system id is first checked in the database. If it does not exist, the function checks for a configured id
+    # (from environment variables or config), and if neither is found, a new random one is generated and stored.
```
TomerShor: Transform this into a proper docstring, using """ etc

server/py/services/api/initial_data.py
```diff
@@ -993,6 +1002,46 @@ def _create_project_summaries(db, db_session):
...
+        db.create_system_id(db_session, system_id)
+        return
+
+    # if no system id is found, generate a new one
+    system_id_str = _generate_system_id()
+    db.create_system_id(db_session, system_id_str)
+
+
+def _get_configured_system_id() -> typing.Optional[str]:
+    system_id = os.environ.get("SYSTEM_ID")
```
TomerShor: This env var is not something we support in MLRun.
Env vars are translated to the MLRun config [automatically](https://github.com/mlrun/mlrun/blob/94c3f48d248608872fde6a0e4882f6c5254bf9a4/mlrun/config.py#L1411-L1413), when they have a certain structure:
```
MLRUN_SYSTEM_ID
```
This will env var will be translated to `mlrun.mlconf.system_id`, so no need to manually check the env var here.

server/py/services/api/initial_data.py
```diff
@@ -993,6 +1002,46 @@ def _create_project_summaries(db, db_session):
...
+    db = framework.db.sqldb.db.SQLDB()
+
+    # check if a system id already exists in the database
+    system_id = db.get_system_id(db_session)
+
+    if system_id is not None:
+        logger.info("Existing system id found in the database", system_id=system_id)
+        return
+
+    # check if the system id is configured via environment variables or configmap
```
TomerShor: the configmap is just a way to set mlrun config via env vars.
You actually only check if the config is already set.

server/py/services/api/initial_data.py
```diff
@@ -993,6 +1002,46 @@ def _create_project_summaries(db, db_session):
     db._upsert(db_session, project_summaries, ignore=True)
 
 
+def _init_system_id(db_session: sqlalchemy.orm.Session):
```
TomerShor, line 1003: What's missing from this function is setting the system id in the config.
Currently you only ensure it is in the DB.
You want it to be available in `mlconf.system_id`, so you need to make sure that at you populate it with the resolved value (overwrite it if needed).

server/py/services/api/tests/unit/test_initial_data.py
```diff
@@ -375,6 +375,48 @@ def test_add_producer_uri_to_artifact():
...
+        ("configmap", "1111"),
+        # simulated mlconf system id
+        ("mlconf", "123"),
+    ],
+)
+def test_init_system_id(
+    system_id_source, expected_system_id, monkeypatch: pytest.MonkeyPatch
+):
+    if system_id_source == "configmap":
+        monkeypatch.setenv("SYSTEM_ID", "1111")
```
TomerShor: see my comments above on the envvar and configmap

TomerShor (changes requested):
Overall it's ok, but missing a key feature

server/py/framework/db/sqldb/db.py
```diff
@@ -7074,6 +7075,23 @@ def delete_model_endpoints(
...
+            self._query(session, SystemMetadata)
+            .filter(SystemMetadata.key == "system_id")
+            .one_or_none()
+        )
+        if not system_id_record:
+            logger.debug("System id not found in DB")
+            return None
+        return system_id_record.value
+
+    def create_system_id(self, session: Session, system_id: str):
```
```suggestion
    def store_system_id(self, session: Session, system_id: str):
    logger.debug("Storing a system id in DB", system_id=system_id)
```
as we already generated(=created) it here.

server/py/services/api/initial_data.py
```diff
@@ -993,6 +1002,46 @@ def _create_project_summaries(db, db_session):
...
+        return system_id
+    return mlrun.mlconf.system_id or None
+
+
+def _generate_system_id() -> str:
+    # Generate a random 32-bit unsigned integer and encode it as base64 string without padding
+    random_int = os.urandom(4)
+    base64_str = base64.b64encode(random_int).decode("utf-8").rstrip("=")
+    return base64_str
+
```
rokatyy: If you have a 4-byte string (like "0000"), the base64 encoding produces a 6-character string, and padding is not needed because it will never have any = at the end.

rokatyy (changes requested):

server/py/services/api/initial_data.py
```diff
@@ -993,6 +1002,46 @@ def _create_project_summaries(db, db_session):
...
+def _get_configured_system_id() -> typing.Optional[str]:
+    system_id = os.environ.get("SYSTEM_ID")
+    if system_id:
+        return system_id
+    return mlrun.mlconf.system_id or None
+
+
+def _generate_system_id() -> str:
+    # Generate a random 32-bit unsigned integer and encode it as base64 string without padding
+    random_int = os.urandom(4)
```
rokatyy: os.urandom(4) doesn't generate `a random 32-bit unsigned integer `, it generate a random mix of bytes
```
os.urandom(4)
Out[16]: b'?7\xe2x'
```

server/py/services/api/initial_data.py
Yacouby: ok so in this case the padding will not do anything.. and it will be consistent with the requirement for a 6-character string..

server/py/services/api/tests/unit/test_initial_data.py
```diff
@@ -375,6 +375,46 @@ def test_add_producer_uri_to_artifact():
...
+        ("random", None),
+        # when a system id is set in mlconf, it should be used
+        ("mlconf", "123"),
+    ],
+)
+def test_init_system_id(
+    system_id_source, expected_system_id, monkeypatch: pytest.MonkeyPatch
+):
+    if system_id_source == "mlconf":
+        monkeypatch.setattr(mlrun.mlconf, "system_id", "123")
```
TomerShor:
```suggestion
        monkeypatch.setattr(mlrun.mlconf, "system_id", expected_system_id)
```

server/py/framework/db/sqldb/db.py
```diff
@@ -7074,6 +7075,23 @@ def delete_model_endpoints(
             main_table_identifier_values=uids,
         )
 
+    def get_system_id(self, session: Session) -> typing.Optional[str]:
+        system_id_record = (
+            self._query(session, SystemMetadata)
+            .filter(SystemMetadata.key == "system_id")
```
liranbg: the "system_id" is repeatedly used here as a constant string. move to `server/py/framework/constants.py` and reuse

server/py/services/api/initial_data.py
```diff
@@ -993,6 +998,47 @@ def _create_project_summaries(db, db_session):
     db._upsert(db_session, project_summaries, ignore=True)
 
 
+def _init_system_id(db_session: sqlalchemy.orm.Session):
+    """
+    Initializes a system id for MLRun deployment.
+    The system id is first checked in the database. If it does not exist, the function checks for a configured id
+    (from environment variables or config), and if neither is found, a new random one is generated and stored.
```
TomerShor:
```suggestion
    The system id is first checked in the database. If it does not exist, the function checks if an id was 
    set in the config), and if neither is found, a new random one is generated and stored.
```

server/py/framework/db/sqldb/db.py
```diff
@@ -7074,6 +7075,23 @@ def delete_model_endpoints(
...
+            .one_or_none()
+        )
+        if not system_id_record:
+            logger.debug("System id not found in DB")
+            return None
+        return system_id_record.value
```
liranbg:
```suggestion
        return system_id_record.value if system_id_record else None
```

not sure the log needs to happen here or by its callee

server/py/services/api/initial_data.py
```diff
@@ -993,6 +998,47 @@ def _create_project_summaries(db, db_session):
...
+        # check if the system id is already set in the config
+        system_id = _get_configured_system_id()
+
+        if system_id is not None:
+            logger.info("Using configured system id", system_id=system_id)
+            db.store_system_id(db_session, system_id)
+        else:
+            # if no system id is found, generate a new one
+            system_id = _generate_system_id()
+            db.store_system_id(db_session, system_id)
```
TomerShor:
```suggestion
        if system_id is not None:
            logger.info("Using configured system id", system_id=system_id)
        else:
            # if no system id is found, generate a new one
            system_id = _generate_system_id()
        db.store_system_id(db_session, system_id)
```

server/py/services/api/initial_data.py
```diff
@@ -100,21 +101,25 @@ def init_data(
...
-            config.httpdb.state = state
-            raise
+            except Exception:
+                state = mlrun.common.schemas.APIStates.migrations_failed
+                logger.warning("Migrations failed, changing API state", state=state)
+                config.httpdb.state = state
+                raise
+
+        # initialize system id
+        _init_system_id(db_session)
```
liranbg, line 121: needs to be moved under `_add_initial_data`, no?

the flow is
"ensure db schema migration, initial the db connections and classes, add initial data, ensure data correctness (migrations)"

I think the `_init_system_id` can happen within `_add_initial_data`

server/py/services/api/initial_data.py
```diff
@@ -993,6 +998,47 @@ def _create_project_summaries(db, db_session):
...
+    system_id = db.get_system_id(db_session)
+
+    if system_id is not None:
+        logger.info("Existing system id found in the database", system_id=system_id)
+    else:
+        # check if the system id is already set in the config
+        system_id = _get_configured_system_id()
+
+        if system_id is not None:
+            logger.info("Using configured system id", system_id=system_id)
```
TomerShor: I'd move these 2 log lines to `debug`, and add an info log at the end :
```
logger.info("Initialized system ID", system_id=system_id)
```

TomerShor (changes requested):
Looking much better!
Minor comments

server/py/services/api/initial_data.py
liranbg:
```suggestion
        if system_id:
            logger.info("Using configured system id", system_id=system_id)
        else:
            # if no system id is found, generate a new one
            system_id = _generate_system_id()
        db.store_system_id(db_session, system_id)

```

server/py/services/api/initial_data.py
```diff
@@ -993,6 +998,47 @@ def _create_project_summaries(db, db_session):
...
+
+def _get_configured_system_id() -> typing.Optional[str]:
+    return mlrun.mlconf.system_id or None
+
+
+def _generate_system_id() -> str:
+    # Generate 4 random bytes and encode them as base64 string without padding
+    random_bytes = os.urandom(4)
+    base64_str = base64.b64encode(random_bytes).decode("utf-8").rstrip("=")
+    return base64_str
```
liranbg: is `+8boxw` a valid system id? because that is a possible generated base64

liranbg (changes requested):
overall flow is legit, some suggestions and comments to ensure correctness :)

server/py/services/api/initial_data.py
Yacouby: yeah thats right, i refactored it a bit and did not notice 👍

server/py/services/api/initial_data.py
rokatyy: @Yacouby but should they be random bytes? Either the requirement + task should be rephrased or logic should be aligned with the original idea of using integer.

server/py/services/api/initial_data.py
```diff
@@ -993,6 +998,47 @@ def _create_project_summaries(db, db_session):
...
+        if system_id is not None:
+            logger.info("Using configured system id", system_id=system_id)
+            db.store_system_id(db_session, system_id)
+        else:
+            # if no system id is found, generate a new one
+            system_id = _generate_system_id()
+            db.store_system_id(db_session, system_id)
+
+    # set the system id in mlrun config
+    mlrun.mlconf.system_id = system_id
```
rokatyy:
```suggestion
    if system_id is not None:
        logger.info("Existing system id found in the database", system_id=system_id)
        mlrun.mlconf.system_id = system_id
        return
  
    # check if the system id is already set in the config
    system_id = _get_configured_system_id()
  
    <Liran suggestion above>
    # set the system id in mlrun config
    mlrun.mlconf.system_id = system_id
```

This way we we improve readability and avoid unnecessary nested structure

server/py/services/api/initial_data.py
```diff
@@ -993,6 +998,47 @@ def _create_project_summaries(db, db_session):
...
+
+
+def _get_configured_system_id() -> typing.Optional[str]:
+    return mlrun.mlconf.system_id or None
```
rokatyy, line 1038: More a usability + validation question. 

if we say that system_id is base64 string of len==6, 

1. Should we enforce this and validate what user passed? 
2. Should we translate users input to base64(because current logic just takes it as is), or how else can we check that it's base64?

server/py/services/api/initial_data.py
Yacouby: thats not accurate, because we need this to be done after the migrations.. becuase in case we are in a flow of clean db, then the init_system_id will try to access the system_metadata table and it will not be there, causing a crash

server/py/services/api/initial_data.py
rokatyy: @liranbg raised a good point with possible `+` in the encoding, I believe that's the reason why originally it was only unsigned int - base64 encoding of strings 0000 to 9999 never includes non-alphabetic symbols because their binary representations map only to alphanumeric characters in the Base64 alphabet.

server/py/services/api/initial_data.py
Yacouby: I can use urlsafe_b64encode, in this case + will be replace with -, and / replace with _

server/py/services/api/initial_data.py
Yacouby: 1. the base64 string and len of 6 are used only when generating a new random one.. ( if it does not exist in the db )
2. users will not use this id, as it comes with MLRun deployment, and should be changed only by support ( by deleting it from the db which will cause generating a new one on next installation, or by config )

rokatyy: @Yacouby well, what happens if support mistakenly sets in to 7, 8 or 10 symbols?

TomerShor: @Yacouby @rokatyy i think the 32bit number was chosen more for its high probability for uniqueness, rather them it's actual length.
The custom ID is a nice-to-have and I don't think we should really limit it# perhaps maybe by coomin name length limitations (such as 63 chats for k8s resources).
We can bring @theSaarco into the discussion to validate the original meaning.

TomerShor: Even if we do decide to limit the size, it can be done in a followup PR

rokatyy (approved):

TomerShor (approved):
LGTM 👍

server/py/services/api/initial_data.py
```diff
@@ -993,6 +999,52 @@ def _create_project_summaries(db, db_session):
...
+
+
+def _generate_system_id() -> str:
+    # Generate a random 32-bit unsigned integer, convert it to 4 bytes, and encode as a Base64 string without padding
+    random_number = random.getrandbits(32)
+    random_bytes = random_number.to_bytes(4, "big")
+    base64_str = base64.b64encode(random_bytes).decode("utf-8").rstrip("=")
+    return base64_str
```
liranbg: instead of using base64 which yields unexpected (or undesired) characters, simply use random character from list of [a-z|A-Z] and numbers. perhaps you can generate and iterate until you find a match that satisfies k8s dns name regex - as we want the system id to later on translate to k8s resource names / other usages that might enforce naming constrains.

usage example - `"".join(random.choices("0123456789" + string.ascii_lowercase, k=N))`

then enforce it matching regex of https://github.com/mlrun/mlrun/blob/196234944cd6d95eeaa81e9654a099ff8a4638d1/mlrun/utils/regex.py#L26

liranbg (changes requested):

server/py/services/api/initial_data.py
theSaarco: @liranbg is right that base64 indeed has some issues with compatibility with DNS names. Another option would be to use the `altchars` parameter to replace `+` and `/` chars - see https://docs.python.org/3/library/base64.html#base64.b64encode , this parameter is actually there for this specific use-case. Usual replacements are to use `-` and `_` for this (but can use `.` as well).
There's also a `urlsafe_b64encode` method which basically does exactly that.
The caveat would be that the resulting string may still start with `_` or `-` which is invalid in many cases, so this string cannot be at the beginning of a name.

Yacouby: I used base64.urlsafe_b64encode to avoid issues with characters like + and / in DNS and Kubernetes names. Although the chance of starting with `-` or `_` is low, I included a check to further ensure the compatibility with the naming conventions.

server/py/services/api/initial_data.py
```diff
@@ -993,6 +999,57 @@ def _create_project_summaries(db, db_session):
...
+    # Generate a random 32-bit unsigned integer and encode as a URL-safe Base64 string without padding
+    while True:
+        random_number = random.getrandbits(32)
+        random_bytes = random_number.to_bytes(4, "big")
+        base64_str = base64.urlsafe_b64encode(random_bytes).decode("utf-8").rstrip("=")
+
+        # ensure the string does not start with '-' or '_', as these characters are not allowed at the start of DNS
+        # names or Kubernetes resource names
+        if not base64_str.startswith(("-", "_")):
+            return base64_str
```
TomerShor: @Yacouby this implementation is case sensitive, but for names it is better to have all lowercase so we won't have 2 ids that are different in case.

Also, underscore (`_`) is not allowed in kubernetes qualified name. You can use the regex Liran suggested - https://github.com/mlrun/mlrun/blob/196234944cd6d95eeaa81e9654a099ff8a4638d1/mlrun/utils/regex.py#L26

TomerShor (changes requested):

TomerShor (approved):
lgtm

liranbg (approved):

PR #6944, opened 2024-12-18
[Alert activation] Fix notification saving logic

rokatyy:
https://iguazio.atlassian.net/browse/ML-8683

Fixes the undefined behaviour bug which was caused by race condition where in one thread we were sending a notification and updating `alert` object with notification state and in another thread, using the information about notification from alert, we were saving the alert activation. New logic will not save notification when storing notification. Instead, notification state will be populated by notification pusher when notifications are sent.

### DIFF
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -6140,8 +6140,8 @@ class SQLDB(DBInterface):
 
         # we need to keep id to be able to update number_of_events for manual reset policy
         # and to update notification state when notification is sent
-        # activation time is retrieved to get the cut activation_time from db
-        # because 2024-12-18T16:06:09.083606+00:00 will be saved as 2024-12-18T16:06:09.084000+00:00
+        # NOTE: activation time is truncated to milliseconds when being saved to the database as we use mysql.DATETIME
+        # example: 2024-12-18T16:06:09.083606+00:00 will be saved as 2024-12-18T16:06:09.084000+00:00
         return self._upsert_object_and_flush_to_get_field(
             session, alert_activation_record, "id"
         )
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -43,6 +43,7 @@ from sqlalchemy import (
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.inspection import inspect
 from sqlalchemy.orm import Session, aliased
+from sqlalchemy.orm.attributes import flag_modified
 
 import mlrun
 import mlrun.common.constants as mlrun_constants
@@ -6107,13 +6108,9 @@ class SQLDB(DBInterface):
         session,
         alert_data: mlrun.common.schemas.AlertConfig,
         event_data: mlrun.common.schemas.Event,
-        notifications_states: list[mlrun.common.schemas.NotificationState],
-    ) -> Optional[str]:
+    ) -> int:
         extra_data = {
-            "notifications": [
-                notification.dict() for notification in notifications_states
-            ],
-            "criteria": alert_data.criteria.dict() if alert_data.criteria else None,
+            "criteria": alert_data.criteria.dict(),
         }
 
         # For JOB entities, construct entity_id as "name.uid" format
@@ -6136,17 +6133,18 @@ class SQLDB(DBInterface):
             data=extra_data,
         )
 
-        # if reset_policy is MANUAL, we need to keep id to be able to update number_of_events
-        # when the alert is reset
-        if alert_data.reset_policy == mlrun.common.schemas.alert.ResetPolicy.MANUAL:
-            return self._upsert_object_and_flush_to_get_field(
-                session, alert_activation_record, "id"
-            )
-        else:
-            # for auto reset policy reset_time is the same as the activation time
-            # for manual reset policy, we keep it empty until the alert is reset
+        # for auto reset policy reset_time is the same as the activation time
+        # for manual reset policy, we keep it empty until the alert is reset
+        if alert_data.reset_policy == mlrun.common.schemas.alert.ResetPolicy.AUTO:
             alert_activation_record.reset_time = alert_activation_record.activation_time
-            self._upsert(session, [alert_activation_record])
+
+        # we need to keep id to be able to update number_of_events for manual reset policy
+        # and to update notification state when notification is sent
+        # activation time is retrieved to get the cut activation_time from db
+        # because 2024-12-18T16:06:09.083606+00:00 will be saved as 2024-12-18T16:06:09.084000+00:00
+        return self._upsert_object_and_flush_to_get_field(
+            session, alert_activation_record, "id"
+        )
 
     def update_alert_activation(
         self,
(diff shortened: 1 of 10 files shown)


tests/integration/sdk_api/alerts/test_alerts.py
```diff
@@ -595,6 +595,46 @@ def test_alert_templates(self):
...
+            "Alert summary",
+            event_name,
+            reset_policy=mlrun.common.schemas.alert.ResetPolicy.AUTO,
+            notifications=[
+                mlrun.common.schemas.AlertNotification(
+                    notification=failure_webhook_notification
+                )
+            ],
+        )
+        for i in range(1, 4):
```
liranbg, line 628: why 4? magic?

server/py/framework/utils/notifications/notification_pusher.py
```diff
@@ -226,6 +257,74 @@ def _prepare_notification_args(
...
+            failed_count = status_info.get("failed_count", 0)
+
+            if errors:
+                if success_count == 0:
+                    error_message = (
+                        f"All {kind} notifications failed. Errors: {', '.join(errors)}"
+                    )
+                else:
+                    error_message = (
+                        f"Some {kind} notifications failed. Errors: {', '.join(errors)}"
```
liranbg, line 311: is Some equals to len(errors)?. also, this could be very long, isnt it? perhaps can be truncated ? or present first error only?

liranbg (commented):
minor

tests/integration/sdk_api/alerts/test_alerts.py
rokatyy: @liranbg technically it is three, just to ensure that bug doesn't reproduce and notification is updated very fast

server/py/framework/utils/notifications/notification_pusher.py
rokatyy: @liranbg since we use set, user will see only unique errors there. It is already truncated here https://github.com/mlrun/mlrun/blob/9ffd6dda1ee3b2d15852fb83e5addfc58e74e6d6/server/py/framework/utils/notifications/notification_pusher.py#L346

rokatyy: > is Some equals to len(errors)

it is equal to summary.failed

tests/integration/sdk_api/alerts/test_alerts.py
liranbg: how can you ensure _notification is updated very fast_ ?
perhaps on slow ci system the test would transiently fail? not sure I understand the scenario maybe could you elaborate on test docstring?

rokatyy: @liranbg we request alert_activation right after we posted an event on which this alert is created. And I expect to have notification state quite fast (it's not requirement, though a good thing) in activation, so in most cases it won't be really visible for user that notification is updated on the fly. Notification is send in async function which we not await for. 
> perhaps on slow ci system the test would transiently fail
 it may, then we might need to add sleep there
 
 I'll add a comment

server/py/framework/db/sqldb/db.py
```diff
@@ -6107,13 +6108,9 @@ def store_alert_activation(
...
         event_data: mlrun.common.schemas.Event,
-        notifications_states: list[mlrun.common.schemas.NotificationState],
-    ) -> Optional[str]:
+    ) -> int:
         extra_data = {
-            "notifications": [
-                notification.dict() for notification in notifications_states
-            ],
-            "criteria": alert_data.criteria.dict() if alert_data.criteria else None,
+            "criteria": alert_data.criteria.dict(),
```
TomerShor, line 6113: What if `criteria`is None?

server/py/framework/db/sqldb/db.py
```diff
@@ -6136,36 +6133,56 @@ def store_alert_activation(
...
             session, AlertActivation, id=activation_id, activation_time=activation_time
         )
         activation = query.one_or_none()
         if not activation:
-            raise mlrun.errors.MLRunNotFoundError(
-                f"Alert activation not found for id: {activation_id}"
-            )
+            # if the activation is not found, we try to find it again by id only
+            # (in case if something happened with activation time)
+            # usually it won't really happen, but just stay in safe side
```
```suggestion
            # if the activation is not found, we try to find it again by id only
            # (in case something happened with activation time)
            # usually it won't really happen, but just stay in safe side
```

What did you experience that made you add this defense?

server/py/framework/utils/notifications/notification_pusher.py
```diff
@@ -162,9 +165,37 @@ async def async_push():
...
+                try:
+                    self._update_alert_activation_notification_state(
+                        activation_id=activation_id,
+                        activation_time=activation_time,
+                        alert=alert,
+                    )
+                except Exception as e:
+                    logger.warning(
+                        "Failed to update alert activation state", error=str(e)
+                    )
```
TomerShor: Add a comment that you do it here to ensure that the notifications are updated with their actual status.

TomerShor (changes requested):

server/py/framework/db/sqldb/db.py
rokatyy: @TomerShor it can't be None

server/py/framework/db/sqldb/db.py
rokatyy: @TomerShor because I noticed that if I used non-normalised datatime, activation wasn't found. So, added this defense just in case.

server/py/framework/utils/notifications/notification_pusher.py
rokatyy: @TomerShor it is not really "ensure" -  I always save notification state only here. We don't save it anymore on initial storing of alert activation.

TomerShor: Right, so that's what I wanted to be in the comment.
To make it clear that the notifications are only saved here after their status is updated.

server/py/framework/rundb/sqldb.py
```diff
@@ -1304,6 +1305,22 @@ def list_alert_activations(
...
+    ):
+        # We run this function with a new session because it may run concurrently.
+        # Older sessions will not be able to see the changes made by this function until they are committed.
+        return self._transform_db_error(
+            framework.db.session.run_function_with_new_db_session,
+            services.alerts.crud.AlertActivation().update_alert_activation,
+            activation_id=activation_id,
+            activation_time=activation_time,
+            notifications_states=notifications_states,
+        )
```
TomerShor, line 1322: In what flow is this DB class instance called and not the `sqldb/db.py` class?
The notification pusher is on the server-side, so doesn't it has access to the sql db class?

server/py/services/alerts/crud/alerts.py
```diff
@@ -253,25 +253,26 @@ def _handle_notification(
         keep_cache = True
         active = False
         state["count"] += 1
-        logger.debug("Sending notifications for alert", name=alert.name)
```
TomerShor, line 256: This log line can be quite useful for debugging. Without it we don't have an indication that notifications are sent for a specific alert.
It should stay.

TomerShor (changes requested):

server/py/framework/rundb/sqldb.py
rokatyy: @TomerShor 
> The notification pusher is on the server-side, so doesn't it has access to the sql db class?

this one yes, but 1- it inherits the base class which can work in [both client and server](https://github.com/mlrun/mlrun/blob/da480edda6b9a8a09e7211260b1f86e673a95038/mlrun/utils/notifications/notification_pusher.py#L45-L70) and 2- I tried to use normal sql class and got circular imports problem 3- in all the places in notification pusher this proxy class is used

TomerShor (approved):

PR #6951, opened 2024-12-19
[Artifacts] Modify artifact deletion logic to reassign the "latest" tag

moranbental:
When deleting an artifact that holds the "latest" tag, the following rules should apply:

1. If the artifact’s iteration is 0 (it was created not during hyperparameter):

The "latest" tag should be moved to the most recent artifact that is marked as the best iteration (best_iteration=True), excluding the artifact being deleted.

2. If the artifact’s iteration is not 0 (it was created during hyperparameter):

First, check if there is another artifact with the same producer_id that has the "latest" tag in other iterations:
 - If another artifact with the "latest" tag exists in a different iteration, the "latest" tag should not be moved.
 - If no other "latest" tag exists, the "latest" tag should be moved to the most recent artifact from the previous run that is marked as the best iteration (best_iteration=True).

Additionally, as a result of [this PR](https://github.com/mlrun/mlrun/pull/6598), there is no longer any need to manually delete the tag from the Tag table, as artifact deletion now automatically deletes all associated tags due to the cascade deletion constraints.

**Note:** This PR does not cover moving the latest tag when deleting multiple artifact keys

https://iguazio.atlassian.net/browse/ML-8685


server/py/framework/db/sqldb/db.py
```diff
@@ -7143,3 +7146,101 @@ def _paginate_query(
...
+                ArtifactV2.Tag.name == mlrun.common.constants.RESERVED_TAG_NAME_LATEST,
+                ArtifactV2.Tag.project == object_record.project,
+                ArtifactV2.Tag.obj_id.in_(
+                    session.query(ArtifactV2.id).filter(
+                        ArtifactV2.producer_id == object_record.producer_id,
+                        ArtifactV2.iteration != object_record.iteration,
+                    )
+                ),
+            )
+            .first()
```
rokatyy: `.all()`?

server/py/framework/db/sqldb/db.py
```diff
@@ -7143,3 +7146,101 @@ def _paginate_query(
...
+            else:
+                logger.warning(
+                    "No more recent artifact found to move 'latest' tag",
+                    artifact_uid=object_record.uid,
+                )
+        else:
+            logger.debug(
+                "No 'latest' tag found for artifact",
+                artifact_uid=object_record.uid,
+            )
```
rokatyy: This check can be moved to the beginning and return there, and then you'll get rid of one nested level

server/py/services/api/tests/unit/db/test_artifacts.py
```diff
@@ -1040,6 +1040,203 @@ def test_delete_artifact_tag_filter(self):
...
+            project=project,
+            tag="v3",
+            best_iteration=True,
+        )
+
+        assert uid1 != uid2 != uid3
+        artifacts = self._db.list_artifacts(
+            self._db_session, name=artifact_key, project=project
+        )
+        assert len(artifacts) == 4
```
rokatyy, line 1100: Why 4? I'm not really familiar with how we log artifacts, may you add a comment explaining that?

rokatyy (commented):
Looking very nice, comments are really helpful 👍 

Some really minor questions

server/py/framework/db/sqldb/db.py
moranbental: @rokatyy we are using `first()` here because we are interested in checking whether any matching tags exist. When using `first()`, the search stops once the first match is found.

liranbg (approved):

server/py/framework/db/sqldb/db.py
```diff
@@ -7149,3 +7152,103 @@ def _paginate_query(
...
+            ArtifactV2.id != object_record.id,
+            ArtifactV2.project == object_record.project,
+            ArtifactV2.key == object_record.key,
+            ArtifactV2.best_iteration,
+        )
+
+        # Return the most recent artifact based on the update timestamp
+        return query.order_by(ArtifactV2.updated.desc()).first()
+
+    def _handle_latest_tag(self, session, object_record, commit):
```
```suggestion
    def _handle_latest_tag_deletion(self, session, object_record, commit):
```

server/py/framework/db/sqldb/db.py
TomerShor, line 1: Since all of these new functions are related to artifacts, move them below the other artifact-related functions and not at the bottom of the file.

server/py/framework/db/sqldb/db.py
```diff
@@ -7149,3 +7152,103 @@ def _paginate_query(
...
+            .filter(
+                ArtifactV2.Tag.obj_id == object_record.id,
+                ArtifactV2.Tag.name == mlrun.common.constants.RESERVED_TAG_NAME_LATEST,
+                ArtifactV2.Tag.project == object_record.project,
+            )
+            .one_or_none()
+        )
+
+    @staticmethod
+    def _find_other_latest_in_other_iterations(session, object_record):
```
TomerShor: 2 `other` is confusing
```suggestion
    def _find_latest_tag_in_other_iterations(session, object_record):
```

server/py/framework/db/sqldb/db.py
```diff
@@ -7149,3 +7152,103 @@ def _paginate_query(
...
+        """Check if there are other 'latest' tags in other iterations for the same producer_id."""
```
```suggestion
        """Check if other iterations for the same producer_id have the 'latest' tag."""
```

server/py/framework/db/sqldb/db.py
```diff
@@ -7149,3 +7152,103 @@ def _paginate_query(
...
+        # Check if we should check for other 'latest' tags (only when iteration != 0) for hyperparameters
+        if object_record.iteration != 0:
+            # Step 2: Check if there are other "latest" tags in the other iterations
+            other_latest = self._find_other_latest_in_other_iterations(
+                session, object_record
+            )
+
+            if other_latest:
+                logger.debug(
+                    "Another 'latest' tag exists in other iterations for the same producer_id. "
```
```suggestion
                    "'latest' tag exists in other iterations for the same producer_id. "
```

server/py/framework/db/sqldb/db.py
```diff
@@ -7149,3 +7152,103 @@ def _paginate_query(
...
+                    "Not moving the 'latest' tag",
+                    artifact_uid=object_record.uid,
+                    producer_id=object_record.producer_id,
+                )
+                return
+        else:
+            logger.debug(
+                "Skipping 'latest' tag check because the iteration is 0",
+                artifact_uid=object_record.uid,
+            )
```
TomerShor: redundant log?

server/py/framework/db/sqldb/db.py
TomerShor: Also, move this function above all of the `_find_...` functions as it calls them.

server/py/framework/db/sqldb/db.py
```diff
@@ -7149,3 +7152,103 @@ def _paginate_query(
...
+
+        # Step 3: Move the "latest" tag if no other "latest" exists in other iterations
+        most_recent_artifact = self._find_previous_most_recent_artifact(
+            session, object_record
+        )
+
+        if most_recent_artifact:
+            logger.debug(
+                "Moved 'latest' tag to the most recent artifact",
```
```suggestion
                "Moving 'latest' tag to the most recent artifact",
```

server/py/framework/db/sqldb/db.py
```diff
@@ -7149,3 +7152,103 @@ def _paginate_query(
...
+                    producer_id=object_record.producer_id,
+                )
+                return
+        else:
+            logger.debug(
+                "Skipping 'latest' tag check because the iteration is 0",
+                artifact_uid=object_record.uid,
+            )
+
+        # Step 3: Move the "latest" tag if no other "latest" exists in other iterations
```
```suggestion
        # Step 3: Move the "latest" tag to the most recently updated artifact with the same key
```

server/py/framework/db/sqldb/db.py
```diff
@@ -7149,3 +7152,103 @@ def _paginate_query(
...
+                "Moved 'latest' tag to the most recent artifact",
+                artifact_uid=most_recent_artifact.uid,
+            )
+            latest_tag.obj_id = most_recent_artifact.id
+            if commit:
+                session.commit()
+            # TODO: Handle the case when commit=False (e.g., deleting multiple artifact keys)
+        else:
+            logger.warning(
+                "No more recent artifact found to move 'latest' tag",
```
```suggestion
                "No recent artifact found to move 'latest' tag",
```

server/py/framework/db/sqldb/db.py
```diff
@@ -7149,3 +7152,103 @@ def _paginate_query(
...
+            session.query(ArtifactV2.Tag)
+            .filter(
+                ArtifactV2.Tag.obj_id != object_record.id,
+                ArtifactV2.Tag.name == mlrun.common.constants.RESERVED_TAG_NAME_LATEST,
+                ArtifactV2.Tag.project == object_record.project,
+                ArtifactV2.Tag.obj_name == object_record.key,
+                ArtifactV2.Tag.obj_id.in_(
+                    session.query(ArtifactV2.id).filter(
+                        ArtifactV2.producer_id == object_record.producer_id,
+                        ArtifactV2.iteration != object_record.iteration,
```
liranbg: add filter by project/key

server/py/framework/db/sqldb/db.py
theSaarco: This message is confusing - you're only skipping the tag check for other iterations. Even though it's just a debug message I don't see a reason to have this message.

TomerShor (changes requested):
Great stuff! Logic is sound, added some comments wrt readability.

server/py/framework/db/sqldb/db.py
```diff
@@ -7149,3 +7152,103 @@ def _paginate_query(
...
+                    )
+                ),
+            )
+            .first()
+        )
+
+    @staticmethod
+    def _find_previous_most_recent_artifact(session, object_record):
+        """Find the most recent artifact based on the update timestamp, excluding the current artifact."""
+        query = session.query(ArtifactV2).filter(
```
theSaarco: You don't really care for all the artifact object here, but just the `id`. So you should use the `with_entities(ArtifactV2.id)` option to get back just the `id` of the object, rather than the full object (which contains the entire JSON structure etc.). It will generate a small performance gain.

server/py/framework/db/sqldb/db.py
```diff
@@ -7149,3 +7152,103 @@ def _paginate_query(
...
+                )
+                return
+        else:
+            logger.debug(
+                "Skipping 'latest' tag check because the iteration is 0",
+                artifact_uid=object_record.uid,
+            )
+
+        # Step 3: Move the "latest" tag if no other "latest" exists in other iterations
+        most_recent_artifact = self._find_previous_most_recent_artifact(
```
theSaarco: As mentioned, you should just get the `id` here (if exists). I see that later you also log the `uid`, so you can get it back as well (or avoid that log).

server/py/framework/db/sqldb/db.py
```diff
@@ -7149,3 +7152,103 @@ def _paginate_query(
...
+                ArtifactV2.Tag.project == object_record.project,
+            )
+            .one_or_none()
+        )
+
+    @staticmethod
+    def _find_other_latest_in_other_iterations(session, object_record):
+        """Check if there are other 'latest' tags in other iterations for the same producer_id."""
+        return (
+            session.query(ArtifactV2.Tag)
```
theSaarco, line 7173: Here too, since you don't care about the full object but just the fact that it exists, you can use `with_entities` to avoid serializing the full JSON. You can just ask for the `id`.

theSaarco (changes requested):
Looks great. Have some suggestions that will give a (smallish, but important) performance boost.

server/py/framework/db/sqldb/db.py
rokatyy: @moranbental rename it then, because the name is confusing

server/py/framework/db/sqldb/db.py
moranbental: @TomerShor I think it is more like `_update_artifact_latest_tag`

TomerShor: still too general, as this function is a very specific flow that is only relevant to deletion.
I'll meet you in the middle with:
`_update_artifact_latest_tag_on_deletion`

server/py/framework/db/sqldb/db.py
moranbental: @TomerShor I have changed the function name to `_has_latest_tag_in_different_iterations` because it now returns True or False to indicate whether the tag exists

server/py/framework/db/sqldb/db.py
moranbental: Fixed

theSaarco (approved):
Looks good 👍

yaelgen (approved):

TomerShor (approved):

rokatyy (approved):

PR #6958, opened 2024-12-22
[Model Monitoring] Standardize urls for model monitoring

tomerm-iguazio:
[ML-8824](https://iguazio.atlassian.net/browse/ML-8824)

**Urls change under `f"projects/{project}/model-monitoring"`:**

POST /enable-model-monitoring -> /
PATCH /model-monitoring-controller -> /controller 
POST /deploy-histogram-data-drift-app -> /histogram-data-drift-app
DELETE /disable-model-monitoring -> / 
POST /set-model-monitoring-credentials -> /credentials

[ML-8824]: https://iguazio.atlassian.net/browse/ML-8824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/py/services/api/api/endpoints/model_monitoring.py
```diff
@@ -194,7 +151,7 @@ async def update_model_monitoring_controller(
     )
 
 
-@router.post("/deploy-histogram-data-drift-app")
+@router.post("/histogram-data-drift-app")
```
jond01: All the idempotent operations should be changed to PUT. I think it applies to all the POST endpoints you mentioned.

jond01 (commented):
I like it a lot!
Please see my comment, and tagging also @theSaarco to review it.

server/py/services/api/api/endpoints/model_monitoring.py
```diff
@@ -107,50 +107,7 @@ async def _common_parameters(
...
-        image=image,
-        base_period=base_period,
-        deploy_histogram_data_drift_app=deploy_histogram_data_drift_app,
-        rebuild_images=rebuild_images,
-        fetch_credentials_from_sys_config=fetch_credentials_from_sys_config,
-    )
-
-
-@router.patch("/model-monitoring-controller")
+@router.patch("/controller")
```
theSaarco, line 169: Did you delete the old API since we consider 1.8 to be a breaking change in MM? 
Normally we would keep the old endpoint, just mark it deprecated and remove in 1.10. If this was done intentionally, then let's discuss this to verify there are no concerns.

server/py/services/api/api/endpoints/model_monitoring.py
theSaarco: I tend to agree that these operations are better described as `PUT` operations, rather than `POST`, I'll modify the req accordingly.

theSaarco (changes requested):
We should discuss whether we want to break BC in this case and remove old endpoints, or keep them.

server/py/services/api/api/endpoints/model_monitoring.py
tomerm-iguazio: added deprecated warnings

jond01 (approved):

server/py/services/api/api/endpoints/model_monitoring.py
```diff
@@ -107,7 +107,15 @@ async def _common_parameters(
     )
 
 
-@router.post("/enable-model-monitoring")
+# TODO: remove /projects/{project}/model-monitoring/enable-model-monitoring in 1.8.0
+@router.post(
+    "/enable-model-monitoring",
+    deprecated=True,
+    description="/projects/{project}/model-monitoring/enable-model-monitoring "
+    "will be removed in 1.8.0, "
```
theSaarco, line 115:
```suggestion
    "will be removed in 1.10.0, "
```

server/py/services/api/api/endpoints/model_monitoring.py
```diff
@@ -107,7 +107,15 @@ async def _common_parameters(
     )
 
 
-@router.post("/enable-model-monitoring")
+# TODO: remove /projects/{project}/model-monitoring/enable-model-monitoring in 1.8.0
```
theSaarco, line 110:
```suggestion
# TODO: remove /projects/{project}/model-monitoring/enable-model-monitoring in 1.10.0
```

theSaarco (changes requested):
Looks good - quick fix needed (version #).

theSaarco (approved):
Looks good, as discussed.

server/py/services/api/api/endpoints/model_monitoring.py
tomerm-iguazio: As we talked, lets remove the deprecated paths in the end of 1.8.0.
I will open a task for it.

PR #6938, opened 2024-12-18
[Notifications] Send running notification from BE

roei3000b:
When pipeline start running, call the `push_notifications` endpoint so the notification will send from the BE.

https://iguazio.atlassian.net/browse/ML-8536
Example:

Running:
image

Error:
image

PR #6964, opened 2024-12-22
[Notifications] Fix bugs on push_pipeline_start_message

roei3000b:
Fix bug introduced on [this pr](https://github.com/mlrun/mlrun/pull/6938/files#diff-70248055cfe6255eee70163bf67189b832b6c469a72cdb6e26ec8d6f6816e856R651):
Before pushing start message on pipeline we should check that the context is not None.
For example:
```
project = mlrun.load_project(
            "hello",
            "git://github.com/roei3000b/mlrun_kfp_example.git",
            name="iguazio-academy",
            clone=True
        )

project.run(
            "train",
            watch=True,
            local=False,
        )

```

https://iguazio.atlassian.net/browse/ML-8816


mlrun/projects/pipelines.py
```diff
@@ -646,10 +646,14 @@ def run(
                     func_name=func.metadata.name,
                     exc_info=err_to_str(exc),
                 )
-        project.notifiers.push_pipeline_start_message(
-            project.metadata.name,
-            context.uid,
-        )
+
+        # TODO: we should check how can we get the run uid when we don't the the context (for example on
+        #  mlrun.load_project() and later call directly to project.run)
```
liranbg, line 651:
```suggestion
        # TODO: we should check how can we get the run uid when we don't have the context (for example on
        #  mlrun.load_project() and later call directly to project.run)
```

liranbg (commented):
can be in followup pr

PR #6972, opened 2024-12-23
[Alerts] Fix the alert state created field timestamp

Yacouby:
This PR addresses a bug where the created timestamp for new alert states was incorrectly set to the timestamp of the first alert state created in the database. 
The issue occurred because the default value for the created field in the AlertState SQLAlchemy model was being evaluated only once when the class was first loaded, rather than for each new instance causing the value to be the same for each newly added record.

Jira:
https://iguazio.atlassian.net/browse/ML-8378

### DIFF
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -6189,8 +6189,7 @@ class SQLDB(DBInterface):
             return state.to_dict()
 
     def create_alert_state(self, session, alert_record):
-        current_time = datetime.now(timezone.utc)
-        state = AlertState(count=0, parent_id=alert_record.id, created=current_time)
+        state = AlertState(count=0, parent_id=alert_record.id)
         self._upsert(session, [state])
 
     def delete_alert_notifications(
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -6189,7 +6189,8 @@ class SQLDB(DBInterface):
             return state.to_dict()
 
     def create_alert_state(self, session, alert_record):
-        state = AlertState(count=0, parent_id=alert_record.id)
+        current_time = datetime.now(timezone.utc)
+        state = AlertState(count=0, parent_id=alert_record.id, created=current_time)
         self._upsert(session, [state])
 
     def delete_alert_notifications(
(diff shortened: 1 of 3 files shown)


rokatyy (changes requested):
> The issue occurred because the default value for the created field in the AlertState SQLAlchemy model was being evaluated only once when the class was first loaded, rather than for each new instance.

Sounds like very nice catch! However this way of solving it is more a work-around than a solution. 

1. Why not changing default value to - `default=datetime.now` or `default=datetime.utcnow` (without calling it). I didn't check, but I assume it should solve issue?
2. I think that there are might be other places with this issue, worth checking and fixing them all(not necessarily within this ticket).

TomerShor:
@rokatyy you have the right idea.
@Yacouby will open a bug for fixing it for all of the models that use it, but as a current workaround for alerts specifically I think this will do.

rokatyy:
@TomerShor 
> but as a current workaround for alerts specifically I think this will do.

why not fixing it correctly for this field now?

Yacouby:
@rokatyy 
> 1. Why not changing default value to - `default=datetime.now` or `default=datetime.utcnow` (without calling it). I didn't check, but I assume it should solve issue?
Yes apparently this will solve the issue here

> 2. I think that there are might be other places with this issue, worth checking and fixing them all(not necessarily within this ticket).
You are right, there are several places that have this issue too that used the same workaround that i did.
I think it is better to fix all of them once on the same PR, since the change will require generating a new migration, so better do that one time. and also this will need removing the workarounds that were made and validating that everything is good after that.
So this might take more time than we think and the issue here should not wait for that, since it is kind of a blocker for the list alert activations flow, when using the `from_last_update` flag and deleting an alert and creating it again which will cause a wrong saved timestamps in the new created one and causing wrong output when returning the activations.

rokatyy:
@Yacouby 

No, Changing the default value in SQLAlchemy does not require migration because it only affects Python-side behaviour and does not change the database schema.

Yacouby:
@rokatyy Great point 👍  

Looks like this will be a simpler solution to the issue here, and indeed this change will not require migration, from SQLAlchemy doc:
"Column defaults can be client-side Python functions or constant values which are invoked by SQLAlchemy before data is passed to the database."
Since the datetime.utcnow default is handled by SQLAlchemy on the Python side, it won't affect the database schema or require any changes, making the migration process unnecessary.

TomerShor (approved):

PR #6973, opened 2024-12-23
[Packagers] Add `datetime.datetime` packager

jond01:
Fixes [ML-8817](https://iguazio.atlassian.net/browse/ML-8817).

[ML-8817]: https://iguazio.atlassian.net/browse/ML-8817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/execution.py
```diff
@@ -1306,4 +1307,8 @@ def _cast_result(value):
         return float(value)
     if isinstance(value, np.ndarray):
         return value.tolist()
+    if isinstance(value, datetime.datetime):
+        return value
+    if value is None:
+        return value
```
guy1992l, line 1313: you can't do it here as this function is used in `MLClientCtx.log_result`, it will break if the user will manually log a result as it will now be not JSON serializable.

mlrun/package/packagers/python_standard_library_packagers.py
```diff
@@ -628,9 +629,33 @@ def unpack_path(
...
 
+
+class DatetimePackager(DefaultPackager):
+    """
+    ``datetime.datetime`` packager.
+    """
+
+    PACKABLE_OBJECT_TYPE = datetime.datetime
+    DEFAULT_PACKING_ARTIFACT_TYPE = ArtifactType.RESULT
+    DEFAULT_UNPACKING_ARTIFACT_TYPE = ArtifactType.RESULT
```
guy1992l, line 645: You can't unpack a result as it is not a `DataItem`. You can simply delete this line.

mlrun/package/packagers/python_standard_library_packagers.py
```diff
@@ -628,9 +629,33 @@ def unpack_path(
...
+
+    def pack_result(self, obj: datetime.datetime, key: str) -> dict:
+        """
+        Pack the :code:`datetime.datetime` as a string result.
```
guy1992l, line 649:
```suggestion
        Pack the ``datetime.datetime`` as a string result.
```

tests/package/packagers_testers/python_standard_library_packagers_testers.py
```diff
@@ -47,17 +49,12 @@
 # ----------------------------------------------------------------------------------------------------------------------
 
 
-NoneType = type(None)  # TODO: Replace with types.NoneType from python 3.10
-
-
-def pack_none() -> NoneType:
+def pack_none() -> None:
```
guy1992l, line 52: type hinting `-> None` means it is not returning anything as `None` is a actually an instance of `NoneType`. So the hint is correct (see the TODO above about Python 3.10)

tests/package/packagers_testers/python_standard_library_packagers_testers.py
```diff
@@ -47,17 +49,12 @@
...
     return None
 
 
-def validate_none(result: NoneType) -> bool:
-    # TODO: None values should not be casted to strings when casted to results, once it is implemented in
-    #       'execution._cast_result`, change this validation to `return result is None`.
-    return result == "None"
```
guy1992l, line 60: Revert it as the cast result change will break stuff when users will manually log None or other non-serializable objects.

tests/package/packagers_testers/python_standard_library_packagers_testers.py
```diff
@@ -173,7 +170,7 @@ class FloatPackagerTester(PackagerTester):
 _BOOL_SAMPLE = True
 
 
-def pack_bool() -> float:
+def pack_bool() -> bool:
```
guy1992l, line 173: Thanks! :)

tests/package/packagers_testers/python_standard_library_packagers_testers.py
```diff
@@ -970,3 +967,53 @@ class PathPackagerTester(PackagerTester):
...
         ],
     ]
+
+
+# ----------------------------------------------------------------------------------------------------------------------
+# datetime packagers:
+# ----------------------------------------------------------------------------------------------------------------------
+
+
+_DATETIME_SAMPLE = datetime.datetime(2021, 3, 30, tzinfo=datetime.timezone.utc)
```
guy1992l, line 977: special date for you?

guy1992l (changes requested):
I think you were almost on point, but just missed the following detail: the packagers, like their name, are turning regular artifacts into packages they can later unpack automatically so the user will have automatic logging and "no hands" magic. But that is true for **ARTIFACTS**, not results. 

Results are JSON serializable and thus do not need special handling nor they can get one as they are passed via the `params` and do not turn into `DataItem`s. They are not artifacts so you can't inject `unpacking_instructions` to later look at and know how to unpack them. 

You can say you need the packagers to also try to cast values according to the type hinted in the function, which is cool and very easy to implement, but it needs implementation :)

tests/package/packagers_testers/python_standard_library_packagers_testers.py
jond01: I don't think so. We can discuss it, but please note that not returning anything from a function is equivalent to `return None`. It is indeed a special case where the type hint is not the class of the returned object.

tests/package/packagers_testers/python_standard_library_packagers_testers.py
jond01: Not sure I understand you here - this is just the test. Do you refer to the changes in `_cast_result`?
And `None` is JSON serializable:
```py
import json

json.dumps(None) == "null"
```

mlrun/execution.py
jond01: I agree about `datetime.datetime`, but `None` is JSON serializable

tests/package/packagers_testers/python_standard_library_packagers_testers.py
guy1992l: I guess you can test what happen if you pass:
```python
def some_handler(a):
    print(type(a))
    print(a)


my_mlrun_func.run(
    handler="some_handler",
    params={"a": None}
)
```
How will MLRun parse it? give it a try. If you get an actual `None` or `"None"`

tests/package/packagers_testers/python_standard_library_packagers_testers.py
guy1992l: well, I guess both are correct. I wen't on the pure logic approach I guess. If you think its better this way, let it be :)

jond01:
The original issue was fixed in a simpler way. Closing this for now.

PR #6975, opened 2024-12-24
[CI] Fix KFP dependencies for system tests

adamdelman:


liranbg (approved):
This is good enough to unblock system test.
That being said, we should limit the usage of kfp to a specific test suite so that all others wont interfere with kfp dependencies

PR #6976, opened 2024-12-24
[Client-spec] Add system id

Yacouby:
Jira:
https://iguazio.atlassian.net/browse/ML-8717

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -580,6 +580,7 @@ class HTTPRunDB(RunDBInterface):
                 or config.feature_store.default_targets
             )
             config.alerts.mode = server_cfg.get("alerts_mode") or config.alerts.mode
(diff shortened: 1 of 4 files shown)


mlrun/db/httpdb.py
```diff
@@ -580,6 +580,7 @@ def connect(self, secrets=None):
                 or config.feature_store.default_targets
             )
             config.alerts.mode = server_cfg.get("alerts_mode") or config.alerts.mode
+            config.system_id = server_cfg.get("system_id")
```
TomerShor: Why not?
```suggestion
            config.system_id = server_cfg.get("system_id") or config.system_id
```

Yacouby: @TomerShor Won't we always have it always in the server_cfg ?

liranbg (approved):

PR #6980, opened 2024-12-24
[Model Monitoring] Fix create model endpoint API 

davesh0812:


server/py/services/api/api/endpoints/model_endpoints.py
```diff
@@ -94,8 +94,7 @@ async def create_model_endpoint(
     if not model_endpoint.metadata.project or not model_endpoint.metadata.name:
         raise MLRunInvalidArgumentError("Model endpoint must have project and name")
 
-    return await run_in_threadpool(
-        services.api.crud.ModelEndpoints().create_model_endpoint,
```
gtopper, line 98: Yup, that wouldn't have worked, because `create_model_endpoint` is a coroutine.

davesh0812: But now it will :)

gtopper (approved):

assaf758 (approved):

PR #6981, opened 2024-12-24
[Object Counters] Introduce Feature Gates For The Various Object Counters [1.7.x]

quaark:

Our current counter logic is naive and can cause unwanted database locks which prevent backups from completing. We are working on a more robust counter system, though for now we are introducing feature gates to disable the problematic counter queries on very heavy systems.

Remove `_calculate_functions_counters` as it is not used anywhere and is private and internal.

### DIFF
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -127,7 +127,6 @@ default_config = {
                     "models": "enabled",
                     "runs": "enabled",
                     "pipelines": "enabled",
-                    "functions": "enabled",
                 },
             },
         },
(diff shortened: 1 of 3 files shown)


moranbental (approved):

mlrun/config.py
```diff
@@ -120,6 +120,14 @@
         "projects": {
             "summaries": {
                 "cache_interval": "30",
+                "feature_gates": {
```
theSaarco, line 123: Since you put feature gates on all types of counters, please also add one for `"functions"` (which is handled in `_calculate_functions_counters()`.

theSaarco (changes requested):
Looks good - missing one feature-gate, for completeness.

mlrun/config.py
quaark: I don't see anywhere that uses that method, but sure

theSaarco (approved):
sababa 👍

PR #6983, opened 2024-12-24
[Model Monitoring] Support V3IO datastore profile for the streams

jond01:
Implements [ML-8835](https://iguazio.atlassian.net/browse/ML-8835).

[ML-8835]: https://iguazio.atlassian.net/browse/ML-8835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Also fix https://iguazio.atlassian.net/browse/ML-8879


mlrun/common/model_monitoring/helpers.py
```diff
@@ -53,22 +51,6 @@ def parse_monitoring_stream_path(
         else:
             stream_uri += f"?topic=monitoring_stream_{project}_{function_name}"
 
-    elif stream_uri.startswith("v3io://") and mlrun.mlconf.is_ce_mode():
```
davesh0812, line 56: I agree that the http is not relevant any more... but lets raise an exception if the stream is not v3io or kafka

mlrun/model_monitoring/helpers.py
```diff
@@ -245,29 +254,40 @@ def get_tsdb_connection_string(
     )
 
 
-def _get_tsdb_profile(
-    project: str = "",
-    secret_provider: typing.Optional[typing.Callable[[str], str]] = None,
+def _get_profile(
+    project: str,
+    secret_provider: Optional[Callable[[str], str]],
+    profile_name_key: str,
```
davesh0812, line 260: I think this name could be more descriptive.
```suggestion
    secret_key: str,
```

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -290,8 +295,10 @@ def apply_and_create_stream_trigger(
         """
 
         # Get the stream path from the configuration
-        stream_path = services.api.crud.model_monitoring.get_stream_path(
-            project=self.project, function_name=function_name
+        stream_path = mlrun.model_monitoring.get_stream_path(
+            project=self.project,
+            function_name=function_name,
+            secret_provider=self._secret_provider,
```
davesh0812, line 301: I'm not sure if this is the relevant PR, but in the near future, we need to return the full profile to include the access_key or SSL (kafka).

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -1073,6 +1078,12 @@ def set_credentials(
                 or mlrun.mlconf.model_endpoint_monitoring.stream_connection
                 or _default_secrets_v3io
             )
+
+        if stream_profile_name:
+            # TODO: Add checks.
```
davesh0812, line 1083: ?

davesh0812 (approved):
LGTM.
Few comments.

After the Kafka PR, I think we should directly use the datastore profile to leverage the additional features it provides, such as the access key and SSL (for Kafka), and stop using the `get_stream_path` method, or update it accordingly.

mlrun/model_monitoring/helpers.py
jond01: Honestly, it was my first variable name. I think that `secret_key` is too generic. Maybe `profile_name_secret_key`? A bit long but can work

mlrun/common/model_monitoring/helpers.py
jond01: When this function is called the `stream_uri` is already set, so I think validating it in the first steps will give better coverage https://github.com/mlrun/mlrun/pull/6983#discussion_r1897325932

server/py/services/api/crud/model_monitoring/deployment.py
jond01: Once Kafka is also in, this step will validate that the given profile name is V3IO/Kafka.

PR #6985, opened 2024-12-24
[API] Fix service path resolution

liranbg:
the previous regex captured wrong paths, e.g causing `projects/fraud-demo-admin/nuclio/events-ingest/deploy` to be captured as alerts service falsely.

added a test to cover all endpoints and ensure routers with specific tags are redirected to alerts service

https://iguazio.atlassian.net/browse/ML-8793

### DIFF
--- a/server/py/framework/utils/clients/discovery.py
+++ b/server/py/framework/utils/clients/discovery.py
@@ -96,6 +96,7 @@ class Client(
             "put": [],
             "delete": [],
             "patch": [],
+            "head": [],
         }
         routes = self._service_routes(service_name)
         for methods, path in routes:
@@ -120,7 +121,7 @@ class Client(
         for service_name, service_instance in self.services.items():
             routes_patterns = service_instance.method_routes[method]
             for route_pattern in routes_patterns:
-                if route_pattern.match(path):
+                if route_pattern.fullmatch(path):
                     return service_name
         return None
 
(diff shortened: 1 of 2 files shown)


server/py/framework/utils/clients/discovery.py
```diff
@@ -96,6 +96,7 @@ def _resolve_service_method_routes(
             "put": [],
             "delete": [],
             "patch": [],
+            "head": [],
```
moranbental, line 99: Why do we need that here? I haven't seen a request type like that in our code

moranbental (changes requested):
I think it’s worth adding this log:
`logger.debug("Skipping DB session creation for forwarded request", method=request.method, path=request.url.path)
` here:
https://github.com/mlrun/mlrun/blob/449735b0a857bc03e3db1f9fe37063af1fc367d6/server/py/framework/api/deps.py#L42
That way, we can identify the problem next time if it happens.

server/py/framework/utils/clients/discovery.py
liranbg: needed for docs endpoint

liranbg:
> logger.debug("Skipping DB session creation for forwarded request", method=request.method, path=request.url.path)

this is too spammy and we dont have verbose log level

moranbental (approved):
LGTM

PR #6987, opened 2024-12-25
[Datastore] Support langchain vectordb loaders with undownloadable links

alxtkr77:
https://iguazio.atlassian.net/browse/ML-8873

### DIFF
--- a/mlrun/artifacts/document.py
+++ b/mlrun/artifacts/document.py
@@ -325,7 +325,7 @@ class DocumentArtifact(Artifact):
         """
 
         loader_spec = DocumentLoaderSpec.from_dict(self.spec.document_loader)
-        if loader_spec.mlrun_download_object and self.get_target_path():
+        if loader_spec.download_object and self.get_target_path():
             with tempfile.NamedTemporaryFile() as tmp_file:
                 mlrun.datastore.store_manager.object(
                     url=self.get_target_path()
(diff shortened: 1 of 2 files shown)


mlrun/artifacts/document.py
```diff
@@ -47,6 +47,7 @@ def __init__(
         self,
         loader_class_name: str = "langchain_community.document_loaders.TextLoader",
         src_name: str = "file_path",
+        mlrun_download_object: bool = False,
```
theSaarco: Now that I look at it, I actually think we can just call it `download_object` without the `mlrun` prefix. The reason I wanted the prefix was to separate this parameter (which is handled by MLRun) from the parameters passed to the underlying loader, since I thought we're using `**kwargs` to pass arbitrary params. However, since the loader params are anyway passed in a separate dictionary (the `kwargs` dict) then this extra separation is not needed.
Alternatively, maybe instead of a `kwargs` dict we want to use `**kwargs` and the user will just provide the loader params as 1st-level parameters to the class?

theSaarco (approved):
Looks good, though I have a general thought about how we pass the loader params through the loader-spec and how this affects the naming.

mlrun/artifacts/document.py
```diff
@@ -321,7 +325,7 @@ def to_langchain_documents(
         """
 
         loader_spec = DocumentLoaderSpec.from_dict(self.spec.document_loader)
-        if self.get_target_path():
+        if loader_spec.mlrun_download_object and self.get_target_path():
```
assaf758: mlrun_download_object => download_object

assaf758 (approved):

PR #7001, opened 2024-12-26
[Artifacts ] Modify del_artifacts logic to reassign the "latest" tag when deleting multiple artifacts keys

moranbental:
Change the process of deleting multiple artifacts to a simpler implementation that reassigns the "latest" tag when deleting multiple artifact keys. Limit the number of artifacts allowed for deletion in this manner (based on the number of artifacts returned by the query). If the user requests the deletion of more than 200 artifacts, fail the command and prompt them to use a more granular filter. This value is configurable via `mlrun.mlconf.artifacts.limits.max_deletions`

**The logic remains the same as before:** if the deletion of one artifact fails, the process continues with the next artifact, but now errors are logged. At the end, if not all artifacts are deleted, an error will be raised. This logic will probably need to be changed when enabling the deletion of artifact data for multiple artifact keys.

https://iguazio.atlassian.net/browse/ML-8821

### DIFF
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -967,7 +967,7 @@ class SQLDB(DBInterface):
                 f"is {max_deletions}. Refine the filter and try again with a smaller batch."
             )
 
-        logger.info(f"Deleting {total_artifacts} artifacts")
+        logger.info("Deleting artifacts", total_artifacts=total_artifacts)
 
         failed_deletions_count = 0
 
@@ -996,7 +996,7 @@ class SQLDB(DBInterface):
             raise mlrun.errors.MLRunInternalServerError(
                 f"Failed to delete {failed_deletions_count} artifacts"
             )
-        logger.info(f"Successfully deleted {total_artifacts} artifacts")
+        logger.info("Successfully deleted artifacts", total_artifacts=total_artifacts)
 
     def list_artifact_tags(
         self, session, project, category: mlrun.common.schemas.ArtifactCategories = None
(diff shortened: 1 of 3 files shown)


server/py/framework/db/sqldb/db.py
```diff
@@ -958,44 +958,45 @@ def del_artifacts(
...
+                f"is {max_deletions}. Refine the filter and try again with a smaller batch."
             )
-            if artifact_column_identifier is None:
-                # record was not found
-                continue
 
-            artifact_column_identifiers.setdefault(
-                artifact_column_identifier, []
-            ).append(column_value)
+        logger.info(f"Deleting {total_artifacts} artifacts")
```
liranbg:
```suggestion
        logger.info("Deleting artifacts", total_artifacts=total_artifacts)
```

server/py/framework/db/sqldb/db.py
```diff
@@ -958,44 +958,45 @@ def del_artifacts(
...
+                    err=err_to_str(exc),
+                )
+                failed_deletions_count += 1
+                continue
 
         if failed_deletions_count:
             raise mlrun.errors.MLRunInternalServerError(
                 f"Failed to delete {failed_deletions_count} artifacts"
             )
+        logger.info(f"Successfully deleted {total_artifacts} artifacts")
```
liranbg:
```suggestion
        logger.info("Successfully deleted artifacts",  total_artifacts=total_artifacts)
```

liranbg (changes requested):
well done, minor log convention fixes
in addition to that, worth adding UT coverage for multiple delete tags (if not exists), coverage for failing to deleting 1 artifacts while other succeeds, and coverage for having more artifacts that maximum allowed

moranbental:
> coverage for failing to deleting 1 artifacts while other succeeds

This is already covered in `test_delete_artifacts_failure`

liranbg (approved):

PR #7002, opened 2024-12-26
[Project Summaries] Improve Artifacts Count Query

quaark:
https://iguazio.atlassian.net/browse/ML-8892

Instead of querying all artifacts of a certain category and summing them in code, use the following query:
```
select project,kind,count(distinct artifacts_v2.key) from artifacts_v2 group by project,kind;
```

On a system with 15 million artifact this query would still take several minutes and not return in a reasonable time.
However, after creating the `project-kind-key` index on that system which took 15 minutes (for 15 million artifacts which is significantly more than we've seen in the field), the query now returns after 19 seconds.


server/py/framework/db/sqldb/db.py
```diff
@@ -3131,12 +3126,16 @@ async def get_project_resources_counters(
...
-            project_to_files_count,
+            project_to_artifact_category_count[
+                mlrun.common.schemas.ArtifactCategories.other
+            ],
             project_to_schedule_count,
             project_to_schedule_pending_jobs_count,
             project_to_schedule_pending_workflows_count,
             project_to_feature_set_count,
-            project_to_models_count,
+            project_to_artifact_category_count[
```
moranbental: why datasets, models, and documents be under the same counter?

quaark: That's a good question I don't have the answer for currently, but I asked @theSaarco the same, we decided not to change the current structure of the project summary for this version.

quaark: Also that's not exactly what we are counting:
models - > models
files -> any artifact that is NOT model/dataset/document - so `artifact` or `table` or `plot` etc.

moranbental: @quaark It’s worth adding a TODO/comment to explain that

liranbg (approved):
lgtm

moranbental (approved):
LGTM

server/py/framework/db/sqldb/db.py
```diff
@@ -3237,33 +3240,22 @@ def _calculate_feature_sets_counters(session) -> dict[str, int]:
...
-        return project_to_files_count
+    @staticmethod
+    def _calculate_artifact_counters_by_category(
+        session: Session,
+    ) -> dict[str, dict[str, int]]:
+        query = session.query(
+            ArtifactV2.project, ArtifactV2.kind, func.count(distinct(ArtifactV2.key))
+        ).group_by(ArtifactV2.project, ArtifactV2.kind)
+
+        project_to_artifact_category_count = {}
```
theSaarco: A better name would be `category_to_project_artifact_count` since it maps the category to projects.

theSaarco (approved):
LGTM. Have a bit of a nitpick comment about naming of object, but the logic looks solid.

server/py/framework/db/sqldb/db.py
TomerShor: Why not a defaultdict?

quaark: cos I don't know the default is for each category, the dictionary is: categories -> projects -> counters

PR #7010, opened 2024-12-26
[Docs] Add v1.7.2 to change log, additional doc updates [1.7.x]

jillnogold:
And add details about pre-emption and node selector. Remove refs to storey doc


rayaku (commented):
@jillnogold we need to remove support of Spark 3.5 .
It will be supported with IG 3.6.2, which is not released yet
@assaf758  , please keep me honest

Thanks!

assaf758:
> @jillnogold we need to remove support of Spark 3.5 . It will be supported with IG 3.6.2, which is not released yet @assaf758 , please keep me honest
> 
> Thanks!

@rayaku, the truth is in between  - while Spark upgrade is coming in 3.6.2, we did fix issues arising when mlrun working with spark 3.5 (see https://github.com/mlrun/mlrun/pull/6907).
In other words: 
1. we know that mlrun-1.7.1 will not work with 3.6.2 (due to spark upgrade)
2. mlrun-1.7.2 for all we know works 3.6.2/spark-3.5 but we didn't finalize the tests - so mlrun-1.7.3 might still be needed.

gilad-shaham (approved):

liranbg (approved):

PR #7011, opened 2024-12-26
[Model Monitoring] Add percentage based stream sampling to serving function

Eyal-Danieli:
Support down sampling of model events that will be pushed to the monitoring stream based on `sampling percentage` - [ML-8962](https://iguazio.atlassian.net/browse/ML-8962). 

Important notes:
- **`sampling_percentage` is provided via `set_tracking()` API**: The new value will be passed through the `set_tracking()` API. Note that the `sampling_percentage` can be any floating point number that is greater than 0 and less or equal to 100 (by default `sampling_percentage=100`). 
- **Store `sampling_percentage` in the mep schema**: Apart from the serving func, the new value will be stored in each mep that is served by this serving func. 
- **`estimated_prediction_count` - new TSDB field**: To be able to estimate the total amount of predictions when the `sampling_percentage!=100` a new field has been added to the tsdb `predictions` table. This field will be used in the UI to display an approximation of the actual prediction count - [ML-8931](https://iguazio.atlassian.net/browse/ML-8931).
- **`effective_sample_count` - new TSDB field**: To handle batch serving (and also fix [ML-8582](https://iguazio.atlassian.net/browse/ML-8285)), another new field has been added to the tsdb `predictions` table. This field represents the actual amount of events sampled in a single call. Meaning, that the `estimated_prediction_count` is basically `effective_sample_count*(100/sampling_percentage`)
- **How to display the estimated total amount of predictions?**: Following the last 2 points, to get the total estimated amount of predictions in a single time window, we will sum the `estimated_prediction_count` column, rather than just counting the amount of records directly as we did until now. 

[ML-5842]: https://iguazio.atlassian.net/browse/ML-5842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ML-8931]: https://iguazio.atlassian.net/browse/ML-8931?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ML-8582]: https://iguazio.atlassian.net/browse/ML-8582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ML-8962]: https://iguazio.atlassian.net/browse/ML-8962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/serving/v2_serving.py
```diff
@@ -537,4 +557,19 @@ def push(self, start, request, resp=None, op=None, error=None, partition_key=Non
...
+        """
+        Randomly selects indices of requests to sample based on the given percentage
+
+        :param num_of_reqs: Number of requests to select from
+        :param percentage: Sample percentage for each request
+        :return: A list containing the indices of the selected requests
+        """
+
+        return [
+            req for req in range(num_of_reqs) if random.random() < (percentage / 100)
```
davesh0812, line 554: Why not to define the percentage float to be between 0-1 in first place?

mlrun/serving/v2_serving.py
```diff
@@ -504,9 +504,28 @@ def push(self, start, request, resp=None, op=None, error=None, partition_key=Non
...
 
-        self._sample_iter = (self._sample_iter + 1) % self.stream_sample
-        if self.output_stream and self._sample_iter == 0:
+        if self.output_stream:
             microsec = (now_date() - start).microseconds
+            # sample the data based on the percentage
+
+            if self.sampling_percentage != 100:
+                # Randomly select a subset of the requests
+                num_of_inputs = len(request["inputs"])
```
davesh0812, line 515: The inputs can be list of features... so you need to check first if it is a nested list like we doing in  L-409 in the stream processing

mlrun/runtimes/nuclio/serving.py
```diff
@@ -309,21 +309,21 @@ def set_tracking(
         self,
         stream_path: Optional[str] = None,
         batch: Optional[int] = None,
-        sample: Optional[int] = None,
+        sampling_percentage: float = 100,
```
davesh0812, line 312: You sure it is ok to overwrite and not ti use the deprecation process?

tests/serving/test_serving.py
```diff
@@ -644,6 +645,40 @@ def test_function():
...
+            str(err.value)
+            == "`sampling_percentage` must be greater than 0 and less or equal to 100."
+        )
+
+    fn.set_tracking(
+        stream_path="dummy://", sampling_percentage=random_sample_percentage
+    )
+    server = fn.to_mock_server()
+    for i in range(500):
+        server.test("/v2/models/my/infer", testdata)
```
davesh0812, line 677: testdata = '{"inputs": [5]}'

You can change it to '{"inputs": [5, 7]}' and to test if the single infer is working as expected.

tests/system/model_monitoring/test_app.py
```diff
@@ -1608,3 +1608,163 @@ def test_histogram_app(self) -> None:
...
+        predictions_df_without_sample = predictions_df[
+            predictions_df["endpoint_id"] == ep_id_without_sample
+        ]
+
+        # Validate that the model endpoint without sampling includes all the data points
+        assert predictions_df_without_sample["effective_sample_count"].sum() == 1000
+        assert predictions_df_without_sample["estimated_prediction_count"].sum() == 1000
+        # As for the model endpoint with sampling, the effective sample count should be around 155
+        # corresponding to the 15.5% sampling. We will validate that it is not equal to 1000.
+        assert predictions_df_with_sample["effective_sample_count"].sum() != 1000
```
davesh0812, line 1770: Should we verify that the `estimated_prediction_count` falls within a reasonable range?

mlrun/serving/v2_serving.py
```diff
@@ -527,6 +546,7 @@ def push(self, start, request, resp=None, op=None, error=None, partition_key=Non
                         "metrics",
                     ]
                     data["values"] = self._batch
+                    data["effective_sample_count"] = len(request["inputs"])
```
davesh0812: In this case, I think you need to iterate over `self._batch` and sum up the `effective_sample_count` for each invocation. I'm not sure it was working well even before your change. :)

davesh0812 (commented):
Few comment :)

mlrun/serving/v2_serving.py
Eyal-Danieli: the req was 0-100 to clarify that it represents a percentage, not a rate (0–1)

mlrun/runtimes/nuclio/serving.py
Eyal-Danieli: We currently don't want to support the sample interval because it may cause issues with rebalancing and as far as I know it hasn't been maintained or tested.

tests/system/model_monitoring/test_app.py
Eyal-Danieli: Don't think it's necessary because it's a statistcal sampling so I don't want to risk in test failures in these cases. Theoretically, the test could also fail with an effective sample size of 1000, but the probability of that happening is almost impossible.

tests/serving/test_serving.py
Eyal-Danieli: I'll add more invocations with multiple features and verify that the estimated count is correct

mlrun/serving/v2_serving.py
Eyal-Danieli: fixed + identified a bug in the pre-defined batch param - https://iguazio.atlassian.net/browse/ML-8988

mlrun/serving/v2_serving.py
```diff
@@ -504,15 +504,43 @@ def push(self, start, request, resp=None, op=None, error=None, partition_key=Non
             self.output_stream.push([data], partition_key=partition_key)
             return
 
-        self._sample_iter = (self._sample_iter + 1) % self.stream_sample
-        if self.output_stream and self._sample_iter == 0:
+        if self.output_stream:
+            if not isinstance(request["inputs"][0], list):
```
davesh0812: I think this is the safer way to do this, think on a situation that one of the features is a list and the rest are not. 
```
request["inputs"] = (
    request["inputs"]
    if not any(not isinstance(feat, list) for feat in request["inputs"])
    else [request["inputs"]]
) 
```

davesh0812 (approved):
One last comment :)

assaf758 (approved):
Excellent work @Eyal-Danieli, both on quality and speed of execution!

PR #7016, opened 2024-12-27
[System id] Support syncing the system id in worker flow

Yacouby:
This PR implements the synchronization of system metadata from the database for the worker. 
The metadata is fetched and updated to ensure that the configuration values are accurately set and reflect the current system id stored in the database.

Jira:
https://iguazio.atlassian.net/browse/ML-8894


server/py/services/api/main.py
```diff
@@ -98,8 +99,33 @@ async def _move_service_to_online(self):
             services.api.initial_data.update_default_configuration_data()
             await self._start_periodic_functions()
 
+        # For the worker, fetch and sync the system metadata from the database to ensure that the config values are
+        # correctly set.
+        else:
+            await self._sync_system_metadata()
         await self._move_mounted_services_to_online()
 
+    async def _sync_system_metadata(self):
```
TomerShor: I don't think this should be an `async` function if you run everything in a threadpool.

TomerShor (approved):

quaark (approved):

PR #7029, opened 2024-12-30
[API] Ensure getting correct clusterization spec

liranbg:
When worker / alerts service comes up they try to clusterize with chief while getting the current state. In same edge cases it may be possible that the alerts service comes up before new-chief and it fires up a GET request to older one, seeing it in "online" mode and that kicks in the move_to_online functionality of alerts service that involves data changes.
Issue is, when upgrading from version < 1.8, alerts service cannot have any access to DB because .. no schemas migrations were done by new chief.
this fix intend to ensure that alerts coming up gets the clusterization spec from its ally chief by performing version checkups.

https://iguazio.atlassian.net/browse/ML-8919

### DIFF
--- a/server/py/framework/service/__init__.py
+++ b/server/py/framework/service/__init__.py
@@ -312,7 +312,7 @@ class Service(ABC):
         clusterization_spec: mlrun.common.schemas.ClusterizationSpec,
     ):
         # if clusterization_spec has different version (take out unstable)
-        # deny the response as it may come from a stable chief
+        # deny the response as it may come from an older, stable chief
         worker_version = semver.version.Version.parse(
             mlrun.utils.version.Version().get()["version"]
         )
--- a/server/py/framework/service/__init__.py
+++ b/server/py/framework/service/__init__.py
@@ -22,6 +22,7 @@ from abc import ABC, abstractmethod
 import fastapi
 import fastapi.concurrency
 import fastapi.exception_handlers
+import semver
 from dependency_injector import containers, providers
 
 import mlrun.common.schemas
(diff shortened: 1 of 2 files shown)


server/py/framework/utils/mlrunuvicorn.py
```diff
@@ -35,6 +35,7 @@ def run(httpdb_config, service_name="api"):
     uvicorn.run(
         f"services.{service_name}.daemon:app",
         host="0.0.0.0",
+        factory=True,
```
quaark, line 38: why?

quaark (commented):
Important to note, that the issue also affects workers today although we may have not noticed it due to no traffic during the upgrades.

server/py/framework/service/__init__.py
```diff
@@ -298,19 +299,34 @@ async def _synchronize_with_chief_clusterization_spec(
...
             )
-        else:
-            await self._align_worker_state_with_chief_state(clusterization_spec)
 
     async def _align_worker_state_with_chief_state(
         self,
         clusterization_spec: mlrun.common.schemas.ClusterizationSpec,
     ):
+        # if clusterization_spec has different version (take out unstable)
+        # deny the response as it may come from a stable chief
```
TomerShor:
```suggestion
        # deny the response as it may come from an older, stable chief
```

TomerShor (changes requested):

server/py/framework/utils/mlrunuvicorn.py
liranbg: this is relevant for when running service from pycharm
since .daemon:app is a function - it is implicitly mapped to be a factory and we get a warning log.
this change simply change the flag and set it explicitly.

PR #7030, opened 2024-12-30
[Alerts] Optimize cache and db calls

TomerShor:
In this PR:
- Initialize alerts and alert state caches when starting the service.
- Use the caches instead of calling the DB when we need an alert/state and already have its id at hand.
- Remove items from caches when deleting an alert.
- In `store_alert_state` - use a join on the alert id instead of doing 2 separate queries.
- Remove redundant alert state query when updating existing alert.
- Use `_upsert_object_and_flush_to_get_field` when creating an alert instead of committing and querying again.
- When querying alert state, use `one_or_none` instead of `one` for better error messages.
- Optimize filtering older events by using `filter` instead of removing elements from the event list while iterating on it.


moranbental (approved):

server/py/framework/db/sqldb/db.py
```diff
@@ -5838,18 +5838,17 @@ def store_alert(
...
         alert_record.full_object = alert.dict()
-        alert_state = self.get_alert_state(session, alert_record.id)
 
         self._delete_alert_notifications(session, alert.name, alert, alert.project)
         self._store_notifications(
             session,
             AlertConfig,
             alert.get_raw_notifications(),
-            alert_record.id,
+            str(alert_record.id),
```
rokatyy: why str? Id seems to be integer https://github.com/mlrun/mlrun/blob/93d2097e3dc64bbe1d3781c4414e9be8b7fdfcb3/server/py/framework/db/sqldb/models.py#L797

server/py/services/alerts/crud/alerts.py
```diff
@@ -306,13 +308,16 @@ def _get_alert_state_cached(cls):
             )
         return cls._alert_state_cache
 
-    @staticmethod
-    def _try_populate_event_cache(session: sqlalchemy.orm.Session):
+    def _try_populate_caches(self, session: sqlalchemy.orm.Session):
```
rokatyy, line 312:
```suggestion
    def _try_populate_cache(self, session: sqlalchemy.orm.Session):
```

server/py/services/alerts/crud/alerts.py
```diff
@@ -306,13 +308,16 @@ def _get_alert_state_cached(cls):
...
         for alert in framework.utils.singletons.db.get_db().get_all_alerts(session):
+            # Populate events cache
             for event_name in alert.trigger.events:
                 services.alerts.crud.Events().add_event_configuration(
                     alert.project, event_name, alert.id
                 )
+            # Populate the alert and alert state caches
+            self._get_alert_by_id_cached()(session, alert.id)
+            self._get_alert_state_cached()(session, alert.id)
```
rokatyy, line 321: Why do we need this and why on every iteration?

server/py/services/alerts/crud/alerts.py
```diff
@@ -398,19 +403,36 @@ def validate_alert_name(name: str) -> None:
...
+        def _is_valid_event(event):
             if isinstance(event, str):
                 event_time = datetime.datetime.fromisoformat(event)
             else:
                 event_time = event
-            if now > event_time + period:
-                events.remove(event)
+            return now <= event_time + period
+
+        obj["event_timestamps"] = list(filter(_is_valid_event, obj["event_timestamps"]))
```
rokatyy, line 420: Nice!

rokatyy (commented):
Couple of questions about caching, otherwise looks good

server/py/framework/db/sqldb/db.py
TomerShor: The id is an int, but `_store_notifications` expects a string argument.
So it's either casting to `str` or changing the function arg typing, which is also acceptable.

server/py/services/alerts/crud/alerts.py
TomerShor: My thought was that we anyway do this for loop for populating the events cache - so why not populating the other caches as well.
The initialization might be more costly, but later when endpoints are invoked we would already have the objects cached so no need to query the DB again.

server/py/services/alerts/crud/alerts.py
TomerShor: So my other comment - I populate 3 caches - events, alerts and alert states.

server/py/framework/db/sqldb/db.py
rokatyy: @TomerShor then typing should be changed

server/py/services/alerts/crud/alerts.py
rokatyy: @TomerShor but it just initialises it initially once, which means that every iteration just returns same object every time, doesn't it?

TomerShor: On the first iteration, the LRUCache itself is initialized with a registered callback.
When you call the cache with `(session, alert.id)`, it first searches for a relevant item in the cache. if not found - it triggers the callback.
The alert id is different in each iteration, so it populates the cache with the alert objects.

rokatyy (approved):
Looking good!

PR #7033, opened 2024-12-31
[Pipelines] Creating push notifications endpoint

roei3000b:
Create new endpoint for pushing notifications on pipelines (by pipeline id from kfp).
For that, create new notification pusher: `WorkflowNotificationPusher`:

1. Get the project, workflow_id, notifications and default_params.
2. Iterate over notifications and separate them to sync and async notifications.
3. Instead of checking should notify, always send the notifications.

On later PR we will use this endpoint for sending notifications on KFPRunner.

https://iguazio.atlassian.net/browse/ML-8868


mlrun/utils/notifications/notification_pusher.py
```diff
@@ -440,6 +446,162 @@ def _update_notification_status(
         )
 
 
+class WorkflowNotificationPusher(NotificationPusher):
```
liranbg: Workflow is a shared name for remote and kfp, perhaps rename this class to 
```suggestion
class KFPNotificationPusher(NotificationPusher):
```
as its usage is by kfp only, right?

liranbg (changes requested):
overall looking good. missing some UT coverage

mlrun/utils/notifications/notification_pusher.py
roei3000b: Yes

quaark (changes requested):
Shouldn't `WorkflowNotificationPusher` be in the server side notification pusher directory?
Eventually this package should be removed, no?
Otherwise looks good!

roei3000b:
> Shouldn't `WorkflowNotificationPusher` be in the server side notification pusher directory? Eventually this package should be removed, no? Otherwise looks good!

You right,  but I think we should do it later (I mean when we will move this package to the server, we also move the `KFPNotificationPusher`.

server/py/framework/utils/background_tasks/kinds.py
```diff
@@ -21,3 +21,4 @@ class BackgroundTaskKinds:
     function_deletion = "function.deletion.{0}"
     refresh_smtp_configuration = "smtp.configuration.refresh"
     push_notification = "notification.push.{0}.{1}"
+    push_pipeline_notification = "notification.push.{0}.{1}.{2}"
```
quaark: lets have some difference between `push_notification` and `push_pipeline_notification`, to more easily differentiate
```suggestion
    push_pipeline_notification = "notification.kfp.push.{0}.{1}.{2}"
```

quaark (changes requested):
One last thing

liranbg (approved):

quaark (approved):
lgtm

quaark (approved):

PR #7040, opened 2025-01-01
[Artifacts] Log custom object artifacts to database in model handler

yaelgen:
resolves: [ML-8673](https://iguazio.atlassian.net/browse/ML-8673)
Horovod training jobs fail due to missing artifacts in the database. The `_log_custom_objects` method currently logs custom object artifacts with `db_key=False`, preventing them from being stored in the database. Previously, artifacts were stored inline within the run object, but now only their URIs are saved, requiring database retrieval.

To address this, the PR updates `_log_custom_objects` to not pass `db_key=False` when logging custom object artifacts. This ensures that these artifacts are stored in the database and can be retrieved as needed. Since the custom artifacts are lightweight, storing them in the database has minimal impact on performance or storage.

We should discuss whether there is a reason to load all artifacts stored in the `artifact_uris` list on run object when calling `to_dict`, or if this behavior should be reconsidered.

[ML-8673]: https://iguazio.atlassian.net/browse/ML-8673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


moranbental (changes requested):
Looks good! Missing a test to cover that

liranbg (approved):

PR #7042, opened 2025-01-01
[Model Monitoring] Support Kafka source datastore profile

jond01:
Implements [ML-8918](https://iguazio.atlassian.net/browse/ML-8918).

[ML-8918]: https://iguazio.atlassian.net/browse/ML-8918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/model_monitoring/helpers.py
```diff
@@ -117,6 +117,7 @@ def get_stream_path(
     function_name: str = mm_constants.MonitoringFunctionNames.STREAM,
     stream_uri: Optional[str] = None,
     secret_provider: Optional[Callable[[str], str]] = None,
+    profile: Optional[mlrun.datastore.datastore_profile.DatastoreProfile] = None,
```
davesh0812, line 120: If I understand correctly, this is intended to be used by a client without access to the relevant secrets. I believe it would be better to pass only the profile name and retrieve the path directly from the profile.

mlrun/model_monitoring/helpers.py
```diff
@@ -126,20 +127,25 @@ def get_stream_path(
     :param stream_uri:          Stream URI. If provided, it will be used instead of the one from the project's secret.
     :param secret_provider:     Optional secret provider to get the connection string secret.
                                 If not set, the env vars are used.
+    :param profile:             Optional datastore profile of the stream (V3IO/KafkaSource profile).
     :return:                    Monitoring stream path to the relevant application.
```
davesh0812, line 131: I think we should open a bug or a separate issue to note that currently, we are only building the `stream_path` from the profile and not utilizing additional secrets like SSL, access_key, and others.

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -1048,10 +1045,72 @@ def check_if_credentials_are_set(
...
+            raise mlrun.errors.MLRunInvalidMMStoreTypeError(
+                f"The model monitoring stream profile is of an unexpected type: '{type(stream_profile)}'\n"
+                "Expects `DatastoreProfileV3io` or `DatastoreProfileKafkaSource`."
+            )
+
+    def _validate_kafka_stream(
+        self,
+        kafka_profile: mlrun.datastore.datastore_profile.DatastoreProfileKafkaSource,
+    ) -> None:
+        if kafka_profile.topics:
```
davesh0812, line 1080: Can you add a verification for the  kafka access?

server/py/services/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -874,18 +874,22 @@ def delete_model_endpoints_resources(
...
+            "Deleting model monitoring endpoints resources", project_name=project_name
         )
+        try:
+            stream_path = mlrun.model_monitoring.get_stream_path(
+                project=project_name,
+                secret_provider=services.api.crud.secrets.get_project_secret_provider(
+                    project=project_name
+                ),
+            )
+        except mlrun.errors.MLRunNotFoundError:
```
davesh0812, line 886: You are raising `MLRunValueError` from `get_stream_path`

mlrun/model_monitoring/helpers.py
```diff
@@ -126,20 +127,25 @@ def get_stream_path(
...
-
-    stream_uri = stream_uri or mlrun.get_secret_or_env(
-        key=mm_constants.ProjectSecretKeys.STREAM_PATH, secret_provider=secret_provider
-    )
+    elif isinstance(
+        profile, mlrun.datastore.datastore_profile.DatastoreProfileKafkaSource
+    ):
+        stream_uri = f"kafka://{profile.brokers[0]}"
+    else:
+        raise mlrun.errors.MLRunValueError(
```
davesh0812, line 145: please add to the function docs.

tests/system/model_monitoring/test_app.py
```diff
@@ -110,7 +112,7 @@ def custom_setup(cls, project_name: str) -> None:
             project_name, "./", allow_cross_project=True
         )
         project.set_model_monitoring_credentials(
-            stream_path=mlrun.mlconf.model_endpoint_monitoring.stream_connection,
+            stream_path=os.getenv("MLRUN_MODEL_ENDPOINT_MONITORING__STREAM_CONNECTION"),
```
davesh0812, line 115: I think we should change it to use profile as it the main flow now - relevant to all the MM tests.

mlrun/serving/server.py
```diff
@@ -72,14 +74,19 @@ def __init__(self, enabled: bool, parameters: dict, function_uri: str):
...
 
-            self.stream_uri = mlrun.model_monitoring.get_stream_path(project=project)
+            stream_args = parameters.get("stream_args", {})
+
+            if log_stream == DUMMY_STREAM:
+                self.stream_uri = DUMMY_STREAM
+            elif not stream_args.get("mock"):
+                self.stream_uri = mlrun.model_monitoring.get_stream_path(
+                    project=project
+                )
```
davesh0812: Why we need this change?

davesh0812 (commented):
Few comments

tests/system/model_monitoring/test_app.py
jond01: I definitely agree and intend to do it gradually after we support the TSDB profile.
Opened [ML-9020](https://iguazio.atlassian.net/browse/ML-9020).

server/py/services/api/crud/model_monitoring/deployment.py
jond01: Added

mlrun/model_monitoring/helpers.py
jond01: This argument is used mainly in the deployment flow, where we have the profile in our hands, so there's no reason to re-read it.

assaf758 (approved):
Nice!
Please handle the remaining unresolved comments in follow-up PR.

mlrun/model_monitoring/helpers.py
jond01: [ML-9031](https://iguazio.atlassian.net/browse/ML-9031)

PR #7043, opened 2025-01-01
[Notifications] Pushing KFP notifications from BE

roei3000b:
Pushing KFP notifications from BE using [push notifications endpoint](https://github.com/mlrun/mlrun/pull/7033).

1. Add the notifications to the `ProjectSpec`.
2. Create `push_pipeline_notification_kfp_runner` function that iterate other notifications and filter them by the run state.
3. Send the relevant notifications using `push_pipeline_notifications`

https://iguazio.atlassian.net/browse/ML-8869

### DIFF
--- a/mlrun/common/runtimes/constants.py
+++ b/mlrun/common/runtimes/constants.py
@@ -227,6 +227,8 @@ class RunStates:
             mlrun_pipelines.common.models.RunStatuses.canceling: RunStates.aborting,
             mlrun_pipelines.common.models.RunStatuses.skipped: RunStates.skipped,
             mlrun_pipelines.common.models.RunStatuses.runtime_state_unspecified: RunStates.unknown,
+            mlrun_pipelines.common.models.RunStatuses.error: RunStates.error,
+            mlrun_pipelines.common.models.RunStatuses.paused: RunStates.unknown,
         }[pipeline_run_status]
 
 
--- a/mlrun/common/runtimes/constants.py
+++ b/mlrun/common/runtimes/constants.py
@@ -214,6 +214,21 @@ class RunStates:
             RunStates.skipped: mlrun_pipelines.common.models.RunStatuses.skipped,
         }[run_state]
 
+    @staticmethod
+    def pipeline_run_status_to_run_state(pipeline_run_status):
+        if pipeline_run_status not in mlrun_pipelines.common.models.RunStatuses.all():
+            raise ValueError(f"Invalid pipeline run status: {pipeline_run_status}")
+        return {
+            mlrun_pipelines.common.models.RunStatuses.succeeded: RunStates.completed,
+            mlrun_pipelines.common.models.RunStatuses.failed: RunStates.error,
+            mlrun_pipelines.common.models.RunStatuses.running: RunStates.running,
+            mlrun_pipelines.common.models.RunStatuses.pending: RunStates.pending,
+            mlrun_pipelines.common.models.RunStatuses.canceled: RunStates.aborted,
+            mlrun_pipelines.common.models.RunStatuses.canceling: RunStates.aborting,
+            mlrun_pipelines.common.models.RunStatuses.skipped: RunStates.skipped,
+            mlrun_pipelines.common.models.RunStatuses.runtime_state_unspecified: RunStates.unknown,
+        }[pipeline_run_status]
+
 
 # TODO: remove this class in 1.9.0 - use only MlrunInternalLabels
 class RunLabels(enum.Enum):
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -787,13 +787,17 @@ class HTTPRunDB(RunDBInterface):
                 background_task.status.state
                 == mlrun.common.schemas.BackgroundTaskState.succeeded
             ):
-                logger.info("Notifications pushed", project=project, run_id=uid)
+                logger.info(
+                    "Notifications for the run have been pushed",
+                    project=project,
+                    run_id=uid,
+                )
             elif (
                 background_task.status.state
                 == mlrun.common.schemas.BackgroundTaskState.failed
             ):
                 logger.error(
-                    "Failed to push notifications",
+                    "Failed to push run notifications",
                     project=project,
                     run_id=uid,
                     error=background_task.status.error,
@@ -839,14 +843,16 @@ class HTTPRunDB(RunDBInterface):
                 == mlrun.common.schemas.BackgroundTaskState.succeeded
             ):
                 logger.info(
-                    "Notifications pushed", project=project, pipeline_id=pipeline_id
+                    "Pipeline notifications have been pushed",
+                    project=project,
+                    pipeline_id=pipeline_id,
                 )
             elif (
                 background_task.status.state
                 == mlrun.common.schemas.BackgroundTaskState.failed
             ):
                 logger.error(
-                    "Failed to push notifications",
+                    "Failed to push pipeline notifications",
                     project=project,
                     pipeline_id=pipeline_id,
                     error=background_task.status.error,
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -800,6 +800,9 @@ class HTTPRunDB(RunDBInterface):
         :param notifications: List of notifications to push.
         :returns: :py:class:`~mlrun.common.schemas.BackgroundTask`.
         """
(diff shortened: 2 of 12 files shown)


mlrun/common/runtimes/constants.py
```diff
@@ -214,6 +214,21 @@ def run_state_to_pipeline_run_status(run_state: str):
...
+        return {
+            mlrun_pipelines.common.models.RunStatuses.succeeded: RunStates.completed,
+            mlrun_pipelines.common.models.RunStatuses.failed: RunStates.error,
+            mlrun_pipelines.common.models.RunStatuses.running: RunStates.running,
+            mlrun_pipelines.common.models.RunStatuses.pending: RunStates.pending,
+            mlrun_pipelines.common.models.RunStatuses.canceled: RunStates.aborted,
+            mlrun_pipelines.common.models.RunStatuses.canceling: RunStates.aborting,
+            mlrun_pipelines.common.models.RunStatuses.skipped: RunStates.skipped,
+            mlrun_pipelines.common.models.RunStatuses.runtime_state_unspecified: RunStates.unknown,
+        }[pipeline_run_status]
```
moranbental, line 232: Is there a reason that `RunStatuses.error` and `RunStatuses.paused` are not included here? 
In case `pipeline_run_status` is `RunStatuses.error` or `RunStatuses.paused`, this code will fail

mlrun/db/httpdb.py
```diff
@@ -785,6 +785,37 @@ def push_run_notifications(
...
+        :param notifications: List of notifications to push.
+        :returns: :py:class:`~mlrun.common.schemas.BackgroundTask`.
+        """
+        project = project or config.default_project
+
+        response = self.api_call(
+            "POST",
+            path=f"projects/{project}/pipelines/{pipeline_id}/push-notifications",
+            error="Failed push notifications",
+            body=_as_json([notification.to_dict() for notification in notifications]),
```
moranbental, line 833: Validate that notifications is not None and is a list

mlrun/db/httpdb.py
```diff
@@ -785,6 +785,37 @@ def push_run_notifications(
...
+            timeout=timeout,
+        )
+        if response.status_code == http.HTTPStatus.ACCEPTED:
+            background_task = mlrun.common.schemas.BackgroundTask(**response.json())
+            return self._wait_for_background_task_to_reach_terminal_state(
```
moranbental: It’s worth adding a log if the background task fails

mlrun/projects/project.py
```diff
@@ -2670,6 +2673,36 @@ def push_run_notifications(
             timeout=timeout,
         )
 
+    def push_pipeline_notification_kfp_runner(
+        self,
+        pipeline_id: str,
+        current_run_state: "mlrun_pipelines.common.models.RunStatuse",
```
```suggestion
        current_run_state: mlrun_pipelines.common.models.RunStatuses,
```

moranbental (changes requested):

mlrun/db/httpdb.py
```diff
@@ -785,6 +785,43 @@ def push_run_notifications(
...
+        """
+        Push notifications for a pipeline.
+
+        :param pipeline_id: Unique ID of the pipeline(KFP).
+        :param project: Project that the run belongs to.
+        :param notifications: List of notifications to push.
+        :returns: :py:class:`~mlrun.common.schemas.BackgroundTask`.
+        """
+        if notifications is None or type(notifications) is not list:
+            raise MLRunInvalidArgumentError("notifications must be a list")
```
moranbental:
```suggestion
            raise MLRunInvalidArgumentError(""The 'notifications' parameter must be a list."")
```

mlrun/db/httpdb.py
moranbental: What I meant is that when the background task is in a terminal state, the status could also be `BackgroundTaskState.failed`, so in that case, it’s worth adding a log

moranbental (changes requested):
minor

mlrun/db/httpdb.py
```diff
@@ -785,6 +785,43 @@ def push_run_notifications(
...
+        response = self.api_call(
+            "POST",
+            path=f"projects/{project}/pipelines/{pipeline_id}/push-notifications",
+            error="Failed push notifications",
+            body=_as_json([notification.to_dict() for notification in notifications]),
+            timeout=timeout,
+        )
+        if response.status_code == http.HTTPStatus.ACCEPTED:
+            background_task = mlrun.common.schemas.BackgroundTask(**response.json())
+            return self._wait_for_background_task_to_reach_terminal_state(
```
moranbental: What I meant is that when the background task is in a terminal state, the status could also be BackgroundTaskState.failed, so in that case, it’s worth adding a log

moranbental (changes requested):

mlrun/projects/pipelines.py
```diff
@@ -647,13 +647,18 @@ def run(
                     exc_info=err_to_str(exc),
                 )
 
-        # TODO: we should check how can we get the run uid when we don't have the context (for example on
-        #  mlrun.load_project() and later call directly to project.run)
         if context:
             project.notifiers.push_pipeline_start_message(
                 project.metadata.name,
                 context.uid,
             )
```
quaark, line 662: do we still need this?
Also, how are you handling the ipython/console notifications here that don't get sent to the BE?

quaark (changes requested):
Looks good in general, but this PR will introduce a bug, where pipeline output won't be printed to the Jupyter output. This PR should handle that somehow, even if it's a temporary solution.

mlrun/db/httpdb.py
```diff
@@ -780,9 +780,78 @@ def push_run_notifications(
...
             background_task = mlrun.common.schemas.BackgroundTask(**response.json())
-            return self._wait_for_background_task_to_reach_terminal_state(
+            background_task = self._wait_for_background_task_to_reach_terminal_state(
                 background_task.metadata.name, project=project
             )
+            if (
+                background_task.status.state
+                == mlrun.common.schemas.BackgroundTaskState.succeeded
+            ):
+                logger.info("Notifications pushed", project=project, run_id=uid)
```
moranbental:
```suggestion
                logger.info("Notifications for the run have been pushed", project=project, run_id=uid)
```

mlrun/db/httpdb.py
```diff
@@ -780,9 +780,78 @@ def push_run_notifications(
...
+            elif (
+                background_task.status.state
+                == mlrun.common.schemas.BackgroundTaskState.failed
+            ):
+                logger.error(
+                    "Failed to push notifications",
```
moranbental:
```suggestion
                    "Failed to push run notifications",
```

mlrun/db/httpdb.py
```diff
@@ -780,9 +780,78 @@ def push_run_notifications(
...
+            background_task = mlrun.common.schemas.BackgroundTask(**response.json())
+            background_task = self._wait_for_background_task_to_reach_terminal_state(
+                background_task.metadata.name, project=project
+            )
+            if (
+                background_task.status.state
+                == mlrun.common.schemas.BackgroundTaskState.succeeded
+            ):
+                logger.info(
+                    "Notifications pushed", project=project, pipeline_id=pipeline_id
```
moranbental:
```suggestion
                    "Pipeline notifications have been pushed", project=project, pipeline_id=pipeline_id
```

mlrun/db/httpdb.py
```diff
@@ -780,9 +780,78 @@ def push_run_notifications(
...
+                )
+            elif (
+                background_task.status.state
+                == mlrun.common.schemas.BackgroundTaskState.failed
+            ):
+                logger.error(
+                    "Failed to push notifications",
```
moranbental:
```suggestion
                    "Failed to push pipeline notifications",
```

moranbental (changes requested):

mlrun/projects/pipelines.py
roei3000b: We use this for remote runner use case.
I know we talked about merge the two usecases such both of them will use the same endpoint but it is more complicated, so for the first phase I implement it as two flows.
About ipython/console I think I will solve it by also call to custom notification pusher and adding to project.notifiers only ipython/console notifications.

moranbental (approved):
LGTM

quaark (approved):

liranbg (approved):

PR #7044, opened 2025-01-02
[Artifacts] Retrieve Artifacts with "None" Tag When Listing Artifacts for Producer ID

moranbental:
When running a workflow that logs artifacts with no tag by default, the last artifact is stored with the "latest" tag, and older artifacts are stored with the "None" tag.

When attempting to retrieve the run, the function retrieves all the artifacts based on artifact_uri and stores them in the run. The issue arose when the artifact_uri included an artifact without a tag. In this case, we were using a JOIN on the Tag table, which did not retrieve all the artifacts that did not have a tag. To fix this, we are now using a LEFT OUTER JOIN to retrieve all artifacts.

Moreover, if the artifact_uri includes the UID, we can retrieve the artifact by its UID, ensuring that a unique artifact is fetched.

https://iguazio.atlassian.net/browse/ML-8571

### DIFF
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -1617,20 +1617,24 @@ class SQLDB(DBInterface):
         if producer_id:
             query = query.filter(ArtifactV2.producer_id == producer_id)
 
-        query = query.join(ArtifactV2.Tag, ArtifactV2.Tag.obj_id == ArtifactV2.id)
+        # To get all artifacts, even if they don't have tags
+        query = query.outerjoin(ArtifactV2.Tag, ArtifactV2.Tag.obj_id == ArtifactV2.id)
 
         tuples_filter = []
         for key, tag, iteration, uid in artifact_identifiers:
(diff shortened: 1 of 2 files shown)


server/py/framework/db/sqldb/db.py
```diff
@@ -1617,20 +1617,24 @@ def _find_artifacts_for_producer_id(
...
-            )
-            # Add UID filter only if UID is not None
+            base_filter = ArtifactV2.key == key
+
+            # Prioritize filtering by UID if provided
             if uid is not None:
                 base_filter = base_filter & (ArtifactV2.uid == uid)
+            else:
+                iteration = iteration or 0
+                tag = tag or mlrun.common.constants.RESERVED_TAG_NAME_LATEST
```
TomerShor: Won't this still filter out the artifacts with `None` tags?

TomerShor (approved):
LGTM

yaelgen (approved):
LGTM 🚀

PR #7046, opened 2025-01-02
[DB] Handling "Out of Range" Error When Committing Values Exceeding Column Limits

moranbental:
When committing to the database, if a field value exceeds the allowed range (for example, trying to commit a value of 144466666666666 to the current_page field, which is defined as an integer), the database fails with the following error:
```
DataError('(pymysql.err.DataError) (1264, "Out of range value for column \'current_page\' at row 1")').
```
In this case, an error is raised indicating that the column 'current_page' exceeds the allowed range.

https://iguazio.atlassian.net/browse/ML-8442

### DIFF
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -5002,17 +5002,23 @@ class SQLDB(DBInterface):
                     identifiers = ",".join(
                         object_.get_identifier_string() for object_ in objects
                     )
-
-                    mlrun_error = mlrun.errors.MLRunRuntimeError(
-                        f"Failed committing changes to DB. classes={classes} objects={identifiers}"
-                    )
-
                     # check if the error is a conflict error
                     if any([message in str(sql_err) for message in conflict_messages]):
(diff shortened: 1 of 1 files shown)


TomerShor (commented):
Overall looks ok, but what is the purpose here? Making it clearer to the user?
Not sure "column exceeds the allowed range" will be really more informative than "Out of range value for column". Or maybe just the column name was missing?

moranbental:
> Overall looks ok, but what is the purpose here? Making it clearer to the user? Not sure "column exceeds the allowed range" will be really more informative than "Out of range value for column". Or maybe just the column name was missing?

@TomerShor The previous error was
```
MLRunInternalServerError: list runs details: MLRunRuntimeError("Failed committing changes to DB. classes=['PaginationCache'] objects=f8e3927b9a68733eee9a984719ac7a2f82176708dfa39880fc2571e58d82c27a")
```

TomerShor (approved):
@moranbental OK than it is much better.

PR #7047, opened 2025-01-02
[Model Monitoring] Zip the serving spec and encode it to base64

davesh0812:
Fixs https://iguazio.atlassian.net/browse/ML-8960

### DIFF
--- a/server/py/services/api/crud/runtimes/nuclio/function.py
+++ b/server/py/services/api/crud/runtimes/nuclio/function.py
@@ -411,13 +411,14 @@ def _configure_serving_spec(
             can_pass_via_cm
             and serving_spec_len >= mlrun.mlconf.httpdb.nuclio.serving_spec_env_cutoff
         ):
-            if serving_spec_len >= SERVING_SPEC_MAX_LENGTH:
-                raise mlrun.errors.MLRunInvalidArgumentError(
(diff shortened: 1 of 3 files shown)


server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -417,14 +418,22 @@ def _configure_serving_spec(
                     + "in the returned json, and check if the function runs successfully. "
```
theSaarco, line 418: You should also change the check for maximal length that happens right before this - we want to limit to 5000 models, but this check should not happen here. In this location we can either relax the limit, or not do the check at all (no point in checking the length after the zip, since if after zip we have 1MB of data then we're way over the size limit, and the DB won't be able to store such an object).

theSaarco (changes requested):
Nice. Should also modify the size check.

server/py/services/api/crud/runtimes/nuclio/function.py
davesh0812: As discussed I change the limit to 10MB (only for sanity) & will implement the other limitation in https://iguazio.atlassian.net/browse/ML-8992

theSaarco (approved):
LGTM.

assaf758 (approved):

PR #7048, opened 2025-01-02
[Alerts] Improvements to increase stability and performance

rokatyy:
Within this pr:
1.There is a risk of encountering issues when storing alert configurations during an alert service restart. To address this, ensure that the state is always present when storing an alert configuration.
2. Added a with_state flag to enable fetching both the configuration and state in a single query.
3. Fixed the unique constraint for the alert_states table. Specifically, ensuring that parent_id is unique by adding a unique constraint on it and removing the unnecessary constraint.

jira -  https://iguazio.atlassian.net/browse/ML-8925


server/py/framework/db/sqldb/db.py
```diff
@@ -5866,10 +5868,15 @@ def store_alert(
         )
 
         self._upsert(session, [alert_record])
-        return self.get_alert_by_id(session, alert_record.id)
+        # in case if alert service was stopped when storing an alert
```
```suggestion
        # in case alert service was stopped while storing an alert, ensure that it has a state
```

server/py/services/alerts/crud/alerts.py
```diff
@@ -94,8 +94,12 @@ def store_alert(
...
         self._validate_and_mask_notifications(alert_data)
 
-        new_alert = framework.utils.singletons.db.get_db().store_alert(
-            session, alert_data
+        new_alert = (
+            framework.utils.singletons.db.get_db().store_alert(session, alert_data)
+            if existing_alert
+            else framework.utils.singletons.db.get_db().create_alert(
+                session, alert_data
+            )
```
TomerShor, line 98: Why? `store_alert` should handle that already.

server/py/services/alerts/crud/alerts.py
```diff
@@ -48,8 +48,8 @@ def store_alert(
     ):
         project = project or mlrun.mlconf.default_project
 
-        existing_alert = framework.utils.singletons.db.get_db().get_alert(
-            session, project, name
+        existing_alert, existing_alert_state = (
+            framework.utils.singletons.db.get_db().get_alert(session, project, name, with_state=True)
```
TomerShor, line 52: `existing_alert_state` is not used, so we don't need to query it.

TomerShor (changes requested):

server/py/services/alerts/crud/alerts.py
rokatyy: @TomerShor to avoid additional get_alert call when creating alert

server/py/services/alerts/crud/alerts.py
rokatyy: @TomerShor i forgot to change enrich_alert method to accept it. Now if alert_state exists, we will pass it to enrich_alert to avoid unnecessary call.

TomerShor: Actually, it can also be used here - https://github.com/mlrun/mlrun/pull/7048/files#diff-6809ef5c03708fc9e56be5eb145151e71bc24d022ce9c152293a21e244cdea6aR76-R78

server/py/services/alerts/crud/alerts.py
```diff
@@ -73,14 +73,14 @@ def store_alert(
...
-                else mlrun.common.schemas.AlertActiveState.INACTIVE
-            )
+            # Enrich the old alert with existing state
+            existing_alert.state = mlrun.common.schemas.AlertActiveState.INACTIVE
+            if existing_alert_state:
+                existing_alert.state = (
+                    mlrun.common.schemas.AlertActiveState.ACTIVE
+                    if existing_alert_state.to_dict()["active"]
+                    else mlrun.common.schemas.AlertActiveState.INACTIVE
+                )
```
```suggestion
            existing_alert.state = mlrun.common.schemas.AlertActiveState.INACTIVE
            if existing_alert_state and existing_alert_state.to_dict()["active"]:
                existing_alert.state = mlrun.common.schemas.AlertActiveState.ACTIVE
```

TomerShor (changes requested):
last thing, otherwise great

TomerShor (approved):

PR #7051, opened 2025-01-03
[Alerts] Fix test_job_failure_alert_sliding_window

rokatyy:
Added sleeps because there are might be delays in getting notifications + improved naming and added function deletion

### DIFF
--- a/tests/system/alerts/test_alerts.py
+++ b/tests/system/alerts/test_alerts.py
@@ -302,7 +302,7 @@ class TestAlerts(TestMLRunSystem):
             True,
             self._validate_notifications_on_nuclio,
             nuclio_function_url,
-            expected_notifications
+            expected_notifications,
         )
         mlrun.get_run_db().delete_function(
             name=function_name, project=self.project.name
--- a/tests/system/alerts/test_alerts.py
+++ b/tests/system/alerts/test_alerts.py
@@ -46,8 +46,9 @@ class TestAlerts(TestMLRunSystem):
         """
         validate that an alert is sent in case a job fails
         """
+        function_name = "test-func-job-failure-alert"
         self.project.set_function(
-            name="test-func",
+            name=function_name,
             func=str(self.assets_path / "function.py"),
             handler="handler",
             image="mlrun/mlrun" if self.image is None else self.image,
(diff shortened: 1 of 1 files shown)


TomerShor (changes requested):
Sleeps are tricky - if the test relies on transient / timed operations, we cannot guarantee that the sleep time will be sufficient for the system that runs that tests (which runs other tests in parallel), so we can end up increasing the sleeps more and more.

It is better to use a retry-until-successful mechanism if we want to wait for an operation to happen. This way it can be very quick or take time, but it is a more robust solution.

liranbg:
> Sleeps are tricky - if the test relies on transient / timed operations, we cannot guarantee that the sleep time will be sufficient for the system that runs that tests (which runs other tests in parallel), so we can end up increasing the sleeps more and more.
> 
> It is better to use a retry-until-successful mechanism if we want to wait for an operation to happen. This way it can be very quick or take time, but it is a more robust solution.

agree on that. you may use 60 seconds timeout but using retry until successful :)

rokatyy:
@TomerShor will rework with wait_until_successful. The only place where we need to keep sleep is where we make a check for empty notifications as we won't to ensure that no notification appeared within this time.

tests/system/alerts/test_alerts.py
```diff
@@ -419,14 +268,17 @@ def test_job_failure_alert_sliding_window(self):
...
         # Wait for more than two minutes to simulate a delay that is slightly longer than the alert period
         time.sleep(125)
 
         # this is the second failure
         with pytest.raises(Exception):
-            self.project.run_function("test-func")
+            self.project.run_function(function_name)
+
+        # wait since there is a might be a delay
+        time.sleep(20)
```
TomerShor: What delay might be here?

rokatyy: @TomerShor https://github.com/mlrun/mlrun/blob/5b496237e7443216e2aac17c279e2e5882e2eeb0/server/py/services/alerts/main.py#L500 which is 30s, so worst case is 30s

TomerShor: so why not `time.sleep(mlconf.alerts.events_generation_interval)`

rokatyy: @TomerShor not really sure if it really worth to sleep the max time every time. But alright, will do

TomerShor (approved):

PR #7058, opened 2025-01-05
[Pagination] Validate the Maximum Allowed Range for page and page_size

moranbental:
When trying to commit to the database with a page value of 144466666666666, the current_page field is defined as an integer. Since page and page_size are not validated using Pydantic, we need to validate these values before committing them to the database.

I added the validation here rather than at the endpoint level because, if the token already exists in `_create_or_update_pagination_cache_record`, we update current_page to be pagination_cache_record.current_page + 1 before committing to the database. We need to protect against this type of case.

We are limiting the page and page_size to a maximum of 1 million items. These limits can be configured using the following parameters:
```
mlconf.httpdb.pagination.page_limit 
mlconf.httpdb.pagination.page_size_limit
```

https://iguazio.atlassian.net/browse/ML-8442


server/py/framework/db/sqldb/db.py
```diff
@@ -7350,3 +7357,31 @@ def _paginate_query(
             query = query.limit(limit)
 
         return query
+
+    @staticmethod
+    def _validate_max_value(column: Column, value: int):
```
theSaarco:
```suggestion
    def _validate_integer_max_value(column: Column, value: int):
```

theSaarco (commented):
This works, and I'm all for having this infra for validating field limits in general, but why not do another check much earlier in the process? For example in the pagination crud as the request is received? This will allow setting reasonable limits that are related to pagination business logic - after all, I'm not sure having a page size of 2^32-1 really makes much sense for us.

liranbg (approved):

moranbental:
> This works, and I'm all for having this infra for validating field limits in general, but why not do another check much earlier in the process? For example in the pagination crud as the request is received? This will allow setting reasonable limits that are related to pagination business logic - after all, I'm not sure having a page size of 2^32-1 really makes much sense for us.

@theSaarco OK, what is the maximum valid value for page and page_size in relation to pagination business logic?

theSaarco:
@moranbental - we don't have specific limits defined right now (but as we do testing we probably would have better info on actual limits we want to impose). For now I would go for some sanity limits, for example - a page shouldn't be larger than 1M items, and same for number of pages - 1M pages seem more than enough for any reasonable use-case.
These limits should be configurable, of course.

theSaarco (approved):
👍 LGTM

PR #7061, opened 2025-01-05
[Datastore] Add TDEngine datastore profile

jond01:
Implements [ML-9002](https://iguazio.atlassian.net/browse/ML-9002).

[ML-9002]: https://iguazio.atlassian.net/browse/ML-9002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/datastore/test_datastore_profile.py
+++ b/tests/datastore/test_datastore_profile.py
@@ -13,15 +13,18 @@
 # limitations under the License.
 
 from collections.abc import Iterator
+from unittest.mock import patch
 
 import pytest
 
+import mlrun
 import mlrun.common.schemas
 import mlrun.errors
 from mlrun.datastore.datastore_profile import (
     DatastoreProfile2Json,
     DatastoreProfileKafkaTarget,
     DatastoreProfileV3io,
+    TDEngineDatastoreProfile,
     datastore_profile_read,
     register_temporary_client_datastore_profile,
     remove_temporary_client_datastore_profile,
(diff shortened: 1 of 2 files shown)


tests/datastore/test_datastore_profile.py
```diff
@@ -98,3 +103,54 @@ def test_from_public_json() -> None:
...
+                '{"password": "MTIzNA=="}',
+            )
+            profile_read = datastore_profile_read(f"ds://{profile_name}", project_name)
+
+        assert profile_read.type == "taosws", "Wrong profile type"
+        assert profile_read.password == "1234", "Wrong password"
+
+
+def test_datastore_type_map() -> None:
+    assert set(_DATASTORE_TYPE_TO_PROFILE_CLASS.values()) == set(
```
alxtkr77, line 150: Are you sure this test isn’t too strict? It could be broken simply by introducing an intermediate `DatastoreProfile` class. For instance, you might define `NumericDatastoreProfile(DatastoreProfile)`, then derive `DigitalSafeDataStoreProfile` from `NumericDatastoreProfile` and `AnalogSafeDatastoreProfile` directly from `DatastoreProfile`.

jond01: Yes, I agree that in this situation, we'll also have to take the deeper level of inheritance.
As we discussed orally, we will leave this for the future :)

assaf758 (approved):

PR #7062, opened 2025-01-05
[Serving] Limit the number of models to 5K per Router

davesh0812:
https://iguazio.atlassian.net/browse/ML-8992

Note: I removed `MAX_ALLOWED_STEPS` since we are only validating the number of models per router.

Add log when updating a model  (https://iguazio.atlassian.net/browse/ML-8388?focusedCommentId=164105)

### DIFF
--- a/mlrun/serving/states.py
+++ b/mlrun/serving/states.py
@@ -49,6 +49,8 @@ path_splitter = "/"
 previous_step = "$prev"
 queue_class_names = [">>", "$queue"]
 
+MAX_MODELS_PER_ROUTER = 5000
+
 
 class GraphError(Exception):
     """error in graph topology or configuration"""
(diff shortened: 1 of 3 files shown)


mlrun/serving/states.py
```diff
@@ -775,7 +775,13 @@ def add_route(
             2. Create a new model endpoint with the same name and set it to `latest`.
 
         """
-
+        if len(self.routes.keys()) >= MAX_MODELS_PER_ROUTER and key not in self.routes:
+            raise mlrun.errors.MLRunModelLimitExceededError(
+                f"Router cannot support more than {MAX_MODELS_PER_ROUTER} model endpoints. "
+                f"To add a new route, edit an existing one by passing the same key."
+            )
+        if key in self.routes:
```
assaf758, line 783: Minor optimization: if you do this check first, you can do the second check in the "else" clause withuot the "not in self.routes"

assaf758 (approved):

PR #7067, opened 2025-01-06
[Datastore] Fix Kafka source SASL configuration

TomerShor:
Fix how Kafka source configures SASL to align with how Nuclio expects [SASL configuration](https://github.com/nuclio/nuclio/blob/development/pkg/processor/trigger/kafka/types.go#L47).
Mainly - `enable` instead of `enabled`.

Additionally, add explicit `PLAIN` mechanism when using username-password SASL.

https://iguazio.atlassian.net/browse/ML-8968

### DIFF
--- a/mlrun/datastore/sources.py
+++ b/mlrun/datastore/sources.py
@@ -1089,9 +1089,10 @@ class KafkaSource(OnlineSource):
             attributes["partitions"] = partitions
         sasl = attributes.pop("sasl", {})
         if sasl_user and sasl_pass:
-            sasl["enabled"] = True
+            sasl["enable"] = True
             sasl["user"] = sasl_user
             sasl["password"] = sasl_pass
+            sasl["mechanism"] = sasl.get("mechanism", "PLAIN")
         if sasl:
             attributes["sasl"] = sasl
         super().__init__(attributes=attributes, **kwargs)
--- a/mlrun/datastore/sources.py
+++ b/mlrun/datastore/sources.py
@@ -1092,7 +1092,7 @@ class KafkaSource(OnlineSource):
             sasl["enable"] = True
             sasl["user"] = sasl_user
             sasl["password"] = sasl_pass
-            sasl["mechanism"] = sasl.get("mechanism", "PLAIN")
+            sasl["mechanism"] = "PLAIN"
         if sasl:
             attributes["sasl"] = sasl
         super().__init__(attributes=attributes, **kwargs)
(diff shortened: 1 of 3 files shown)


liranbg (approved):

jond01 (dismissed):
Most probably need to change "mechanism" also here:
https://github.com/mlrun/mlrun/blob/121c355835f42b9336ed298225240f5b30c78ffb/mlrun/datastore/datastore_profile.py#L213-L218

I think that the tests failure is related

---

Edit: Actually, I'm not sure it's required here. We need to see this works:
https://docs.mlrun.org/en/latest/feature-store/sources-targets.html#kafka-source
```py
profile = DatastoreProfileKafkaSource(
    name="profile-name", bootstrap_servers=["localhost:9093"], topic="topic_name"
)
source = KafkaSource(path="ds://profile-name")
# Add to Nuclio...
```

mlrun/datastore/sources.py
```diff
@@ -1089,9 +1089,10 @@ def __init__(
             attributes["partitions"] = partitions
         sasl = attributes.pop("sasl", {})
         if sasl_user and sasl_pass:
-            sasl["enabled"] = True
+            sasl["enable"] = True
             sasl["user"] = sasl_user
             sasl["password"] = sasl_pass
+            sasl["mechanism"] = sasl.get("mechanism", "PLAIN")
```
rokatyy: Since we are here, is it always in upper case? Maybe we can do `sasl.get("mechanism", "PLAIN").upper()`?

TomerShor: Actually, I think that in this case it must be `PLAIN` since it's a username-password authentication.
For other mechanisms such as `OAuth` / `SHA`,  are these values aren't relevant.
I'll update.

TomerShor:
> Most probably need to change "mechanism" also here:
> 
> https://github.com/mlrun/mlrun/blob/121c355835f42b9336ed298225240f5b30c78ffb/mlrun/datastore/datastore_profile.py#L213-L218
> 
> I think that the tests failure is related
> 
> Edit: Actually, I'm not sure it's required here. We need to see this works: https://docs.mlrun.org/en/latest/feature-store/sources-targets.html#kafka-source
> 
> ```python
> profile = DatastoreProfileKafkaSource(
>     name="profile-name", bootstrap_servers=["localhost:9093"], topic="topic_name"
> )
> source = KafkaSource(path="ds://profile-name")
> # Add to Nuclio...
> ```

@jond01 you're actually right that this section needs to be fixed too. And the mechanism just ensures that it's using the username-password authenitcation (aka `PLAIN`).

rokatyy (approved):

PR #7069, opened 2025-01-06
[Tests] Fix sleep interval

rokatyy:
title says it all

### DIFF
--- a/tests/system/alerts/test_alerts.py
+++ b/tests/system/alerts/test_alerts.py
@@ -279,7 +279,7 @@ class TestAlerts(TestMLRunSystem):
             self.project.run_function(function_name)
 
         # wait since there is a might be a delay
-        time.sleep(mlconf.alerts.events_generation_interval)
+        time.sleep(int(mlconf.alerts.events_generation_interval))
 
         # validate that no notifications were sent yet, as the two failures did not occur within the same period
         expected_notifications = []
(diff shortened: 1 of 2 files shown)


TomerShor (approved):

tests/system/alerts/test_alerts.py
```diff
@@ -279,7 +279,7 @@ def test_job_failure_alert_sliding_window(self):
             self.project.run_function(function_name)
 
         # wait since there is a might be a delay
-        time.sleep(mlconf.alerts.events_generation_interval)
+        time.sleep(int(mlconf.alerts.events_generation_interval))
```
moranbental: Why not change it here https://github.com/mlrun/mlrun/blob/b077c761e6c148f7ee4d1945b5aecbf824912e07/mlrun/config.py#L802 to int instead of string?

moranbental (changes requested):

tests/system/alerts/test_alerts.py
rokatyy: Done. Since it was introduced in 1.8.0, I don't see any risks of changing the type here.

moranbental (changes requested):
Great! also fix here - https://github.com/mlrun/mlrun/blob/b077c761e6c148f7ee4d1945b5aecbf824912e07/server/py/services/alerts/main.py#L506

rokatyy:
@moranbental I didn't fix it here on purpose. I think it's good to have int() just in case user mistakenly passed string

moranbental (approved):
LGTM

moranbental:
> @moranbental I didn't fix it here on purpose. I think it's good to have int() just in case user mistakenly passed string

Fine, although it is server configuration and not on the client side.

rokatyy:
@moranbental it can anyway be set by admin via config map

TomerShor (approved):

PR #7070, opened 2025-01-07
[Migrations] Kill foreign MySQL processes before migration

adamdelman:
[Migrations] Kill foreign MySQL processes before migration

Since unreaped MySQL client processes can hold locks and prevent the migration
from proceeding, we will kill all foreign processes prior to starting the migration.

https://iguazio.atlassian.net/browse/ML-8742


liranbg (approved):

moranbental (approved):
LGTM

PR #7072, opened 2025-01-07
[KFP] Delete pipeline runs and experiments concurrently

adamdelman:
[KFP] Delete pipeline runs and experiments concurrently
Since in cases where many pipeline runs exist a timeout can be raised,
we now delete the runs and experiments using a threadpool

https://iguazio.atlassian.net/browse/ML-8996


server/py/services/api/crud/pipelines.py
```diff
@@ -113,62 +114,85 @@ def delete_pipelines_runs(
             mlrun.utils.logger.debug(
                 "Detected pipeline runs for project, deleting them",
                 project_name=project_name,
-                pipeline_run_ids=[run["id"] for run in project_pipeline_runs],
+                pipeline_run_count=len(project_pipeline_runs),
```
liranbg, line 117: 🚀

liranbg (approved):

PR #7078, opened 2025-01-08
[Model Monitoring] Fix MM resources deletion

jond01:
Fixes [ML-9034](https://iguazio.atlassian.net/browse/ML-9034) and [ML-9129](https://iguazio.atlassian.net/browse/ML-9129).
The issue was that we relied on the datastore profiles in the MM project deletion process - but they have already been deleted.
I solved it by saving the relevant profiles' data in memory before the relevant records are deleted from the MySQL table.

In addition, delete the internal controller stream introduced in #7045.

I ran the `tests/system/model_monitoring/test_app.py::TestMonitoringAppFlow::test_app_flow[True]` system test with the following permutations:

1. TDEngine + Kafka
2. V3IO TSDB + V3IO stream
3. V3IO TSDB + Kafka

They all passed.

[ML-9034]: https://iguazio.atlassian.net/browse/ML-9034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ML-9129]: https://iguazio.atlassian.net/browse/ML-9129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/py/services/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -1267,3 +1268,117 @@ def _add_feature_stats(
...
+        self,
+    ) -> Optional[mlrun.datastore.datastore_profile.DatastoreProfile]:
+        return self._get_profile(
+            get_profile_function=mlrun.model_monitoring.helpers._get_tsdb_profile
+        )
+
+    def delete(self) -> None:
+        if not self._stream_profile or not self._tsdb_profile:
+            logger.debug(
+                "No model monitoring for the project",
```
Eyal-Danieli:
```suggestion
                "No model monitoring resources were found in this project",
```

server/py/services/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -1267,3 +1268,117 @@ def _add_feature_stats(
...
+                stream_profile=self._stream_profile,
+                model_monitoring_applications=self._model_monitoring_applications,
+                model_monitoring_access_key=self._model_monitoring_access_key,
+            )
+        except Exception as exc:
+            logger.warning(
+                "Failed to delete model monitoring resources",
+                project_name=self._project,
+            )
+            raise exc
```
Eyal-Danieli, line 1384: Are you sure that we don't break the delete project API if for some reason the mm resources were not deleted  as expected in this case? these resources can be stored on the user's side so what happens if the user has changed his kafka creds for example? we should catch the the exception and print it out but skip the rmm esources deletion in this case.

Eyal-Danieli (commented):
nice, my only concern is what happens if there is an error somewhere in the `ModelMonitoringResourcesDeleter` and it breaks the delete project API. We need to ensure that it doesn't break it even if there is an error.

server/py/services/api/crud/model_monitoring/model_endpoints.py
jond01: I agree it's a tricky point, but I kept the behavior here as it was.
I think it's better to start more strict and loosen this approach if we see fit.

We can check for specific errors and skip them - Kafka/TDEngine/V3IO connection errors, for example. However, we do not want to skip the deletion if our deletion code contains a logic error. This improvement can probably be implemented later.

server/py/services/api/crud/projects.py
```diff
@@ -248,23 +235,8 @@ def delete_project_resources(
...
-                project_name=name,
-                db_session=session,
-                model_monitoring_applications=model_monitoring_applications,
-                model_monitoring_access_key=model_monitoring_access_key,
-            )
-        except Exception as exc:
-            logger.warning(
-                "Failed to delete model monitoring resources", project_name=name
-            )
-            raise exc
```
jond01, line 267: The previous behavior

Eyal-Danieli: Alright I know that we are already catching specific errors for tsdb/kafka (such as `MLRunTSDBConnectionFailureError`) so I guess it's enough for now but we just need to be careful not to break the delete project API due to issues with the external resources

Eyal-Danieli (approved):
LGTM

PR #7079, opened 2025-01-08
[KFP] Flatten pipeline params if in list for retry

adamdelman:
[KFP] Flatten pipeline params if in list for retry
When retrying a KFP pipeline, we fetch the pipeline parameters from the previous run.
Due to an issue with the KFP server API, the pipeline parameters are returned as a list
containing a dictionary instead of a dictionary, this PR fixes that.

https://iguazio.atlassian.net/browse/ML-9038


pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/utils.py
```diff
@@ -94,12 +94,16 @@ def retry_run(
                     temp_file.write(pipeline_spec.workflow_manifest)
                     workflow_manifest_path = temp_file.name
 
+            pipeline_parameters = pipeline_spec.parameters
+            if isinstance(pipeline_parameters, list):
+                pipeline_parameters = pipeline_parameters[0]
+
```
liranbg, line 103: leave a comment on why/what you are doing here

adamdelman: done

liranbg (approved):

PR #7080, opened 2025-01-08
[Artifacts] Fix deleting previous artifacts tags

liranbg:
When two producers (aka runs) trying to log the same artifact, a race condition might happen when internal session can not see the deleted artifacts and assume it needs to add a new `latest` tag again causing multiple `latest` tags to be created. later on, fails when trying to resolve the artifact using store uri without `uid` (MultipleResultsFound exception).

This fix ensures the previously deletion tags are committed right before it tries to see if a new tag exists to create.

added session id to log as this critical part is crucial to understand when multiple processes trying to log artifact and the locking mechanism kicks in

https://iguazio.atlassian.net/browse/ML-8949

### DIFF
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -1095,26 +1095,76 @@ class SQLDB(DBInterface):
 
     def tag_artifacts(
         self,
-        session,
+        session: sqlalchemy.orm.Session,
         tag_name: str,
-        artifacts,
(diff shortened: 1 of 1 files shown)


moranbental (approved):
LGTM

quaark (approved):
I don't have any comments on the code itself.
I see not only the commit after the delete, but also you made sure the delete and get use the same session (as opposed to before).
If these 2 changes are what fixes the issue - then great

PR #7085, opened 2025-01-09
[Serving] Save serving spec as json when mlrun client < 1.8.0-rc20

davesh0812:
[ML-9051](https://iguazio.atlassian.net/browse/ML-9051)

[ML-9051]: https://iguazio.atlassian.net/browse/ML-9051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/py/services/api/crud/runtimes/nuclio/function.py
+++ b/server/py/services/api/crud/runtimes/nuclio/function.py
@@ -431,6 +431,15 @@ def _configure_serving_spec(
                 )
             else:
                 # TODO: remove in 1.10.0.
+                if (
+                    serving_spec_len >= SERVING_SPEC_MAX_LENGTH / 10
+                ):  # 1MB limitation as it were before the zip
+                    raise mlrun.errors.MLRunInvalidArgumentError(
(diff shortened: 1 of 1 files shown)


server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -418,11 +418,24 @@ def _configure_serving_spec(
...
+                >= semver.Version.parse("1.8.0-rc20")
+                or "unstable" in client_version
+            ):
+                # Compress and encode the serving spec
+                compressed_serving_spec = gzip.compress(serving_spec.encode("utf-8"))
+                encoded_serving_spec = base64.b64encode(compressed_serving_spec).decode(
+                    "utf-8"
+                )
+            else:
+                # TODO: remove in 1.10.0.
```
assaf758, line 433: in this case the config-map limit is 1MB, so I suggest you move the checking of serving_spec_len to be tailored for the zipped/non-zipped case.

assaf758 (approved):

PR #7089, opened 2025-01-12
[Notifications] Update notification state for running

roei3000b:
Change the logic regarding when we should update notification state in db:
Skip updating only if the following conditions met:

1. The run isn't in terminal state (=> "running").
2. The when contains only one state (which is the current state)

Doing it as a workaround for updating running notification state if the when contains only "running".

https://iguazio.atlassian.net/browse/ML-8630

### DIFF
--- a/mlrun/utils/notifications/notification_pusher.py
+++ b/mlrun/utils/notifications/notification_pusher.py
@@ -414,14 +414,14 @@ class NotificationPusher(_NotificationPusherBase):
     ):
         # Skip update the notification state if the following conditions are met:
         # 1. the run is not in a terminal state
-        # 2. the notification is not only for running state
+        # 2. the when contains only one state (which is the current state)
         # Skip updating because currently each notification has only one row in the db, even if it has multiple when.
         # This means that if the notification is updated to sent for running state for example, it will not send for
         # The terminal state
         # TODO: Change this behavior after implementing ML-8723
         if (
             run_state not in runtimes_constants.RunStates.terminal_states()
-            and notification.when != [runtimes_constants.RunStates.running]
+            and len(notification.when) > 1
         ):
             logger.debug(
                 "Skip updating notification status - run not in terminal state",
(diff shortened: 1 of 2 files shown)


mlrun/utils/notifications/notification_pusher.py
```diff
@@ -412,8 +412,17 @@ def _update_notification_status(
...
+        # Skip update notification state if the following conditions are met:
+        # 1. the run is not in a terminal state
+        # 2. the notification is not only for running state
+        # Skip updating because currently each notification has only one row in the db, even if it has multiple when.
+        # This means that if the notification is updated to sent for running state for example, it will not send for
+        # The terminal state
+        # TODO: Change this behavior after implementing ML-8723
+        if (
+            run_state not in runtimes_constants.RunStates.terminal_states()
+            and notification.when != ["running"]
```
liranbg:
```suggestion
            and notification.when != [mlrun.common.schemas.NotificationStatus.Running]
```

mlrun/utils/notifications/notification_pusher.py
```diff
@@ -412,8 +412,17 @@ def _update_notification_status(
         sent_time: typing.Optional[datetime.datetime] = None,
         reason: typing.Optional[str] = None,
     ):
-        if run_state not in runtimes_constants.RunStates.terminal_states():
-            # we want to update the notification status only if the run is in a terminal state for BC
+        # Skip update notification state if the following conditions are met:
```
liranbg:
```suggestion
        # Skip updating the notification state if the following conditions are met:
```

liranbg (commented):
very well. minor suggestion

mlrun/utils/notifications/notification_pusher.py
yaelgen: Maybe I would modify the condition to check if the `when` contains any terminal state instead of explicitly list of the running state, or check that the length of the list is one,  so it will be more consistent in case n new states are introduced in the future

yaelgen (approved):
LGTM, added a suggestion

PR #7090, opened 2025-01-12
[Alerts] Exclude updated field from alerts response for old client

Yacouby:
The updated field was added to 1.8.0 but clients using version 1.7.1 do not expect this field. 
As a result, calling `list_alerts_configs` or `get_alert_config` with a 1.7.1 client causes a TypeError due to the unexpected `updated` field being included in the response.

This PR introduces the logic to exclude the `updated` field from the alert responses when client version is < 1.8.0,
This was done by using FastAPI's `response_model_exclude_none=True` to ensure that `None` values are excluded from the response.

Jira:
https://iguazio.atlassian.net/browse/ML-8837

Also a minor refactor was done in this PR: changed the name of get_enriched_alert to get_alert ( as the older get_alert method was not used )

### DIFF
--- a/server/py/framework/routers/alerts.py
+++ b/server/py/framework/routers/alerts.py
@@ -70,9 +70,6 @@ async def get_alert(
         Provide[framework.service.ServiceContainer.service]
     ),
 ) -> mlrun.common.schemas.AlertConfig:
-    client_version = request.headers.get("x-mlrun-client-version")
-    exclude_updated = client_version < "1.8.0" if client_version else False
-
     return await service.handle_request(
         "get_alert",
         request,
@@ -80,7 +77,6 @@ async def get_alert(
         name,
         auth_info,
         db_session,
-        exclude_updated=exclude_updated,
     )
 
 
@@ -99,16 +95,12 @@ async def list_alerts(
         Provide[framework.service.ServiceContainer.service]
     ),
 ) -> list[mlrun.common.schemas.AlertConfig]:
-    client_version = request.headers.get("x-mlrun-client-version")
-    exclude_updated = client_version < "1.8.0" if client_version else False
-
     return await service.handle_request(
         "list_alerts",
         request,
         project,
         auth_info,
         db_session,
-        exclude_updated=exclude_updated,
     )
 
 
--- a/server/py/framework/routers/alerts.py
+++ b/server/py/framework/routers/alerts.py
@@ -57,6 +57,7 @@ async def store_alert(
 @router.get(
     "/{name}",
     response_model=mlrun.common.schemas.AlertConfig,
+    response_model_exclude_none=True,
 )
 @inject
 async def get_alert(
@@ -69,6 +70,9 @@ async def get_alert(
         Provide[framework.service.ServiceContainer.service]
     ),
 ) -> mlrun.common.schemas.AlertConfig:
+    client_version = request.headers.get("x-mlrun-client-version")
+    exclude_updated = client_version < "1.8.0" if client_version else False
+
     return await service.handle_request(
         "get_alert",
         request,
@@ -76,10 +80,15 @@ async def get_alert(
         name,
         auth_info,
         db_session,
+        exclude_updated=exclude_updated,
     )
 
 
-@router.get("", response_model=list[mlrun.common.schemas.AlertConfig])
+@router.get(
+    "",
+    response_model=list[mlrun.common.schemas.AlertConfig],
+    response_model_exclude_none=True,
+)
 @inject
 async def list_alerts(
     request: Request,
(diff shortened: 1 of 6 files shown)


server/py/framework/routers/alerts.py
```diff
@@ -69,17 +70,25 @@ async def get_alert(
         Provide[framework.service.ServiceContainer.service]
     ),
 ) -> mlrun.common.schemas.AlertConfig:
+    client_version = request.headers.get("x-mlrun-client-version")
+    exclude_updated = client_version < "1.8.0" if client_version else False
```
TomerShor: 1. This should be in `alerts/main` - The routers are only responsible for routing the request to the handlers, and the handlers have the logic. Note that this file is part of the `framework`, so should not include business logic.
2. This is duplicated in `list_alerts` , can be a common function.

server/py/services/alerts/tests/unit/crud/test_alerts.py
```diff
@@ -410,10 +410,65 @@ async def test_store_alert_update_time(
...
+        alert = services.alerts.crud.Alerts().get_alert(
             session=db,
             project=project,
             name=alert_name,
         )
         assert alert.updated is not None
         assert alert.updated > alert.created.replace(tzinfo=timezone.utc)
+
+    @pytest.mark.asyncio
+    @unittest.mock.patch.object(
```
TomerShor, line 431: import `unittest.mock` to remove PyCharm warnings

server/py/services/alerts/tests/unit/crud/test_alerts.py
```diff
@@ -410,10 +410,65 @@ async def test_store_alert_update_time(
...
+        framework.utils.singletons.db.SQLDB, "update_alert_activation"
+    )
+    @unittest.mock.patch.object(
+        services.alerts.crud.AlertActivation, "store_alert_activation"
+    )
```
TomerShor: instead of setting it separately in the test, do:
```suggestion
    @unittest.mock.patch.object(
        framework.utils.singletons.db.SQLDB, "update_alert_activation", return_value=None
    )
    @unittest.mock.patch.object(
        services.alerts.crud.AlertActivation, "store_alert_activation", return_value=None
    )
```

TomerShor (changes requested):
Looks good, have a couple of suggestions.

server/py/services/alerts/main.py
```diff
@@ -460,6 +474,13 @@ async def _move_service_to_online(self):
             services.alerts.initial_data.update_default_configuration_data(self._logger)
             await self._start_periodic_functions()
 
+    @staticmethod
+    def _should_exclude_updated(client_version: str):
+        # Determine whether the `updated` field should be excluded based on the client version.
```
TomerShor: that is self explanatory. It's better to comment the WHY, e.g - "The updated field was added in 1.8.0 and is not backwards complatible"

server/py/services/alerts/main.py
```diff
@@ -117,6 +118,10 @@ async def get_alert(
     ) -> mlrun.common.schemas.AlertConfig:
         # TODO: When alerts is a different service and not in Hydra mode, we need to send the request to the API and
         #  not access it directly (ML-8565)
+        exclude_updated = self._should_exclude_updated(
+            request.headers.get("x-mlrun-client-version")
+        )
+
```
TomerShor: The comment above it is related to `ensure_project`, not this new function.
Move the function above the call to the CRUD.

server/py/services/alerts/main.py
```diff
@@ -143,6 +152,10 @@ async def list_alerts(
         auth_info: mlrun.common.schemas.AuthInfo,
         db_session: sqlalchemy.orm.Session = None,
     ) -> list[mlrun.common.schemas.AlertConfig]:
+        exclude_updated = self._should_exclude_updated(
+            request.headers.get("x-mlrun-client-version")
+        )
```
TomerShor: same here.
If we don't have permissions on the project, it doesn't matter.

TomerShor (changes requested):

TomerShor (approved):

server/py/services/alerts/main.py
```diff
@@ -158,10 +166,14 @@ async def list_alerts(
             )
         )
 
+        exclude_updated = self._should_exclude_updated(
+            request.headers.get("x-mlrun-client-version")
+        )
```
TomerShor: Small comment post-approval:
Pass the request object and resolve the header inside the function, so the header name will be only in one place.

TomerShor (changes requested):

server/py/services/alerts/main.py
Yacouby: @TomerShor  Isn't it simpler and more efficient to pass the version to the function ? I mean no need to pass the whole request object to the method if you use only one field from it.

TomerShor: I just don't think the duplication of the string is a good idea

TomerShor (approved):

PR #7091, opened 2025-01-12
[Alerts] Add support for future schema fields

Yacouby:
This PR enhances the flexibility of the AlertConfig schema by:
1. Adding **kwargs to the AlertConfig client schema to ensure that future fields added to the schema won't cause errors.
2. Allowing extra fields in the server-side schema.

The client side fix can be a solution for: https://iguazio.atlassian.net/browse/ML-8837 with 1.7.2 client.

### DIFF
--- a/mlrun/alerts/alert.py
+++ b/mlrun/alerts/alert.py
@@ -123,7 +123,6 @@ class AlertConfig(ModelObj):
         :param count:          Internal counter of the alert (user should not supply it)
         :param updated:        The last update time of the alert (user should not supply it)
         """
-        super().__init__(**kwargs)
         self.project = project
         self.name = name
         self.description = description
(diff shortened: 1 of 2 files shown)


mlrun/alerts/alert.py
```diff
@@ -122,6 +123,7 @@ def __init__(
         :param count:          Internal counter of the alert (user should not supply it)
         :param updated:        The last update time of the alert (user should not supply it)
         """
+        super().__init__(**kwargs)
```
TomerShor: Won't this also panic on unexpected field?
Or does ModelObject also support random kwargs?

Yacouby: Indeed it will panic on unexpected field.. this was supposed to be a PR draft but I forgot to mark it.
I updated the PR and removed the super call

TomerShor (approved):

PR #7098, opened 2025-01-14
[KFP] Use correct image for workflow runner for old MLRun versions

adamdelman:
[KFP] Use correct image for workflow runner for old MLRun versions
For clients older than 1.8, use the mlrun-kfp image for the workflow runner

https://iguazio.atlassian.net/browse/ML-9088


server/py/services/api/utils/helpers.py
```diff
@@ -42,9 +42,12 @@ def resolve_client_default_kfp_image(
             # client version is not semver, pass
             pass
 
-    # client is olden than (<) 1.8, must use mlrun image for kfp
-    if must_use_mlrun_image:
-        return mlrun.mlconf.default_base_image
+    # client is older than (<) 1.8 and using kfp or remote engine, must use mlrun-kfp image for kfp
+    if must_use_mlrun_image and workflow_spec.engine in (
+        mlrun.common.schemas.workflow.EngineType.REMOTE,
+        mlrun.common.schemas.workflow.EngineType.KFP,
```
liranbg, line 72:
```suggestion
        mlrun.common.schemas.workflow.EngineType.KFP,
```

if you check for kfp,  than means that client-spec will return "mlrun/mlrun-kfp" for clients with 1.6.4
which will result with a bug

liranbg (changes requested):

liranbg (changes requested):
we need to make sure that
```
(
            "",
            "",
            mlrun.common.schemas.workflow.EngineType.REMOTE,
            "1.8.0",
            "mlrun/mlrun",
        ),
```
returns mlrun-kfp as well. the fact remote aliases to kfp needs to be reflected in the logic as well.

server/py/services/api/utils/helpers.py
```diff
@@ -17,39 +17,47 @@
...
+            parsed_version = semver.VersionInfo.parse(client_version)
+            if parsed_version < semver.VersionInfo(1, 8, 0):
+                pre_kfp_image_mlrun_version = True
         except ValueError:
-            # client version is not semver, pass
             pass
 
-    # client is olden than (<) 1.8, must use mlrun image for kfp
-    if must_use_mlrun_image:
+    engine = (
```
liranbg, line 57: please add a comment explaining the code flow and reasons behind the selection

liranbg (changes requested):

server/py/services/api/utils/helpers.py
adamdelman: done

liranbg (approved):

yaelgen (approved):
LGTM 🚀

PR #7101, opened 2025-01-15
[CI] Run CI against python 3.11

liranbg:
This PR adds Python 3.11 to the runtime matrix of which we run unit and integration test against. The idea is to ensure code added is compatible with Python 3.11.

In addition, since Python 3.11 would work with kfp 2 client and kfp 2 is yet to be ready, I have ignored few of the pipeline test files.

Dask module was required to be updated to support Python 3.11

Bokeh is no longer in used, therefor removed from dependencies

Code changes were required to ensure code is compatible with Python 3.11

https://iguazio.atlassian.net/browse/ML-8396


Makefile
```diff
@@ -510,22 +510,35 @@ test-dockerized: build-test ## Run mlrun tests in docker container
...
-test: clean ## Run mlrun tests
+test: ## Run mlrun tests
+	set -e ;\
+	COMMON_IGNORE_TEST_FLAGS=$$(echo "\
+		--ignore=tests/integration \
+		--ignore=tests/system \
+		--ignore=tests/rundb/test_httpdb.py \
+		--ignore=server/py/services/api/migrations \
+	");\
+	PER_PYTHON_VERSION_IGNORE_TEST_FLAGS=$(if $(filter $(MLRUN_PYTHON_VERSION),3.11),$$(echo "\
```
moranbental, line 528: Add a TODO here so we know that we need to run them when we move to 3.11

Makefile
```diff
@@ -510,22 +510,35 @@ test-dockerized: build-test ## Run mlrun tests in docker container
...
 		--rm \
 		--network='host' \
+		-e MLRUN_PYTHON_VERSION=$(MLRUN_PYTHON_VERSION) \
 		-v /tmp:/tmp \
 		-v /var/run/docker.sock:/var/run/docker.sock \
 		$(MLRUN_TEST_IMAGE_NAME_TAGGED) make test
 
 .PHONY: test
-test: clean ## Run mlrun tests
+test: ## Run mlrun tests
```
moranbental: The test is really confusing; maybe change this to `test-unit`
Also, why did you delete the `clean`?

moranbental (changes requested):
Looks good, just minor changes

Makefile
liranbg: welp, this is here since day0 we can add an alias to test-unit but that is mostly it

PR #7103, opened 2025-01-15
[Model Monitoring] Fix mep creation on graph step

royischoss:
https://iguazio.atlassian.net/browse/ML-8880
Fix allowing "to" for  "QueueStep", and "add_step" for "FlowStep"  method to get parameter of "model_endpoint_creation_strategy" allowing MEP creation when deploying a function using graph topology.
Note: if not provided "model_endpoint_creation_strategy" for those method the mep will not be created and monitoring will be disable.
Small fix for test_basic_model_monitoring


mlrun/serving/v2_serving.py
```diff
@@ -149,13 +149,21 @@ def post_init(self, mode="sync", **kwargs):
...
+            except mlrun.errors.MLRunNotFoundError:
+                logger.info(
+                    "Model Endpoint not found for this step we will not monitor this model",
+                    function_name=server.function_name,
+                    name=self.name,
+                )
+                self.model_endpoint, self.model_endpoint_uid = None, None
         self._model_logger = (
             _ModelLogPusher(self, self.context)
             if self.context and self.context.stream.enabled
```
davesh0812: To be sure that we push to the stream processing only if we have model endpoint 
```suggestion
            if self.context and self.context.stream.enabled and self.model_endpoint_uid
```

davesh0812 (commented):
Looks Good.
One extra comment :)

davesh0812 (approved):

PR #7104, opened 2025-01-15
[Helper] Fix UI links

rokatyy:
Clearer url building

Jira - https://iguazio.atlassian.net/browse/ML-9004

### DIFF
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -1044,26 +1044,35 @@ def get_project_url(project: str) -> str:
     """
     Generate the base URL for a given project.
 
-    Args:
-        project (str): The project name.
-
-    Returns:
-        str: The base URL for the project, or an empty string if the base URL is not resolved.
+    :param project: The project name.
+    :return: The base URL for the project, or an empty string if the base URL is not resolved.
     """
     if mlrun.mlconf.resolve_ui_url():
         return f"{mlrun.mlconf.resolve_ui_url()}/{mlrun.mlconf.ui.projects_prefix}/{project}"
     return ""
 
 
+def get_run_url(project: str, uid: str, name: str) -> str:
+    """
+    Generate the URL for a specific run.
+
+    :param project: The project name.
+    :param uid: The run UID.
+    :param name: The run name.
+    :return: The URL for the run, or an empty string if the base URL is not resolved.
+    """
+    runs_url = get_runs_url(project)
+    if not runs_url:
+        return ""
+    return f"{runs_url}/monitor-jobs/{name}/{uid}/overview"
+
+
 def get_runs_url(project: str) -> str:
     """
     Generate the URL for the runs of a given project.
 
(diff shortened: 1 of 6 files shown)


mlrun/utils/helpers.py
```diff
@@ -1040,34 +1040,98 @@ async def retry_until_successful_async(
...
+    """
+    Generate the base URL for a given project.
 
+    Args:
+        project (str): The project name.
 
-def get_model_endpoint_url(project, model_name, model_endpoint_id):
-    url = ""
+    Returns:
+        str: The base URL for the project, or an empty string if the base URL is not resolved.
```
TomerShor: Use reST docstrings and not Google format (for all docstrings below too)

mlrun/utils/helpers.py
```diff
@@ -1040,34 +1040,98 @@ async def retry_until_successful_async(
...
+        str: The URL for the runs, or an empty string if the base URL is not resolved.
+    """
+    base_url = get_project_url(project)
+    if not base_url:
+        return ""
+    return f"{base_url}/jobs"
+
+
+def get_model_endpoint_url(
+    project: str, model_name: Optional[str], model_endpoint_id: str
```
TomerShor: how come name is optional and id not? It doesn't make sense to use Id without name.
Also, if it's Optional, shouldn't it have a default value?

mlrun/utils/notifications/notification_pusher.py
```diff
@@ -628,7 +628,7 @@ def generate_start_message(
         if has_workflow_url:
             url = mlrun.utils.helpers.get_workflow_url(project, pipeline_id)
         else:
-            url = mlrun.utils.helpers.get_ui_url(project)
+            url = mlrun.utils.helpers.get_runs_url(project)
```
TomerShor, line 631: is this the multiple `runs` on purpose?

mlrun/utils/helpers.py
```diff
@@ -1040,34 +1040,98 @@ async def retry_until_successful_async(
...
+    base_url = get_project_url(project)
+    if not base_url:
+        return ""
+    url = f"{base_url}/jobs/monitor-workflows/workflow"
+    if id:
+        url += f"/{id}"
     return url
 
 
+def get_run_url(project: str, uid: str, name: str) -> str:
```
TomerShor: move next to `get_runs_url`

TomerShor (changes requested):

mlrun/utils/notifications/notification_pusher.py
rokatyy: @TomerShor I think so. The idea here is to just show a main screen if we don't have pipeline_id

mlrun/utils/helpers.py
rokatyy: fixed

TomerShor (approved):

PR #7105, opened 2025-01-15
[Notification] Improve notification validation to be done on client side

rokatyy:
The issue was that a server-side run could fail during the notification step because the notification wasn't properly validated (see screenshot). While validation was already implemented, it only occurred on the server side. This PR adds a validation call to the Notification object initialization and introduces a skip_empty_check flag. When set, this flag bypasses the emptiness check for required fields. This functionality is necessary to avoid failures when certain fields, which are expected to be populated server-side, are not provided initially.

Jira - https://iguazio.atlassian.net/browse/ML-8660

image

### DIFF
--- a/mlrun/model.py
+++ b/mlrun/model.py
@@ -768,15 +768,12 @@ class Notification(ModelObj):
         """
         Validates the notification parameters for the current notification instance.
 
-        Args:
-            default_notification_params (dict, optional): A dictionary of default notification
-                                                          parameters that can be used to enrich
-                                                          the existing params. Defaults to None.
-            skip_empty_check (bool, optional): A flag indicating whether to skip validation for
-                                                required fields that are empty. Defaults to False.
-                                                Set to True only on init of the object on the client side.
+        :param default_notification_params: A dictionary of default notification
+            parameters that can be used to enrich the existing params. Defaults to None.
+        :param skip_empty_check: A flag indicating whether to skip validation for
+            required fields that are empty. Defaults to False.
+            Set to True only on init of the object on the client side.
         """
-
         default_notification_params = default_notification_params or {}
         notification_type = mlrun.utils.notifications.NotificationTypes(self.kind)
         notification_class = notification_type.get_notification()
--- a/mlrun/model.py
+++ b/mlrun/model.py
@@ -760,7 +760,21 @@ class Notification(ModelObj):
                 "Notification params size exceeds max size of 1 MB"
             )
 
-    def validate_notification_params(self, default_notification_params=None):
+        self.validate_notification_params(skip_empty_check=True)
+
+    def validate_notification_params(self, default_notification_params=None, skip_empty_check=False):
+        """
+        Validates the notification parameters for the current notification instance.
+
+        Args:
+            default_notification_params (dict, optional): A dictionary of default notification
+                                                          parameters that can be used to enrich
+                                                          the existing params. Defaults to None.
+            skip_empty_check (bool, optional): A flag indicating whether to skip validation for
+                                                required fields that are empty. Defaults to False.
+                                                Set to True only on init of the object on the client side.
+        """
+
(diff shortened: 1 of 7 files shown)


mlrun/model.py
```diff
@@ -760,7 +760,23 @@ def validate_notification(self):
...
+        """
+        Validates the notification parameters for the current notification instance.
+
+        Args:
+            default_notification_params (dict, optional): A dictionary of default notification
+                                                          parameters that can be used to enrich
+                                                          the existing params. Defaults to None.
+            skip_empty_check (bool, optional): A flag indicating whether to skip validation for
+                                                required fields that are empty. Defaults to False.
+                                                Set to True only on init of the object on the client side.
```
moranbental: this is not our docstring format

mlrun/utils/notifications/notification/git.py
```diff
@@ -40,6 +40,9 @@ def validate_params(cls, params):
             or params.get("GIT_TOKEN", None)
             or params.get("GITHUB_TOKEN", None)
         )
+        if skip_empty_check:
+            return
```
moranbental, line 46: It's worth adding a debug log for skipping (also for mail, slack and webhook)

moranbental (changes requested):
minor

mlrun/utils/notifications/notification/git.py
rokatyy: @moranbental as this code is used in both client and server, I think that for client side it may be way to confusing to see these logs, wdyt?

moranbental: @rokatyy It can be a debug log instead of an info log, and we are only showing info logs to the client unless they change the severity to debug logs.

rokatyy: added

moranbental (approved):
LGTM

yaelgen (approved):
LGTM 🚀

mlrun/utils/notifications/notification/slack.py
```diff
@@ -36,7 +37,10 @@ class SlackNotification(NotificationBase):
     }
 
     @classmethod
-    def validate_params(cls, params):
+    def validate_params(cls, params, skip_empty_check=False):
+        if skip_empty_check:
+            mlrun.utils.logger.debug("Skipping emptiness check for notification params")
```
liranbg, line 42: I find this log line too verbose (spammy). Perhaps should be omitted completely (and for all notifications)

liranbg (commented):
I think that conditioned the validation on parameter is a bit complicated. overall, the flow is we should enrich, then validate and if validation fails, for whatever reason that is, it should happen at its time.
doing the validation client-side to make it better reflected to clients is legit (and done for mlrun on couple of flows) but we should not perform all validations of course, just what we know for sure (as we want the client to be less opinionated).

so, question is, can we have 2 functions, one called from client side and one on regular rather than having one function behaves differently given parameter?

rokatyy:
@liranbg 
> so, question is, can we have 2 functions, one called from client side and one on regular rather than having one function behaves differently given parameter?

of course we can, but the question is what is easier to support and how not to forget to add same validations(if required on both sides) to both functions. In general, when I was thinking about it I was considering two ways - the one I implemented it (which spits validation into emptiness check and correctness check of non-empty fields - the second can always be done on client side). The second approach is much more complicated, but more holistic. It's having the client-side notification entity(as we have for api gateways for instance). This way we could specify the list of required fields from client side and perform the validation of client-side fields fully.

mlrun/utils/notifications/notification/slack.py
rokatyy: I agree here, will remove 👍

rokatyy:
it was decided to proceed with second suggestion above. So closing the pr

PR #7113, opened 2025-01-16
[ML-7905] Move image pull secret enrichment to server side

adamdelman:
[ML-7905] Move image pull secret enrichment to server side

Disable setting the image pull secret on the client side, as this can change during runtime. Instead, do it in the ServerSideLauncher so we always get the up to date parameter.
Move the test `test_new_function_args_with_default_image_pull_secret` to `services.api.tests.unit.test_launcher` as enrichment as done in the server side launcher.
Remove the test `services.api.tests.unit.runtimes.test_spark.TestSpark3Runtime.test_deploy_with_image_pull_secret` as enrichment is no longer done on the client side.
Add `__deepcopy__` method to the `Config` class to prevent looping when accessing the `_missing` field.

https://iguazio.atlassian.net/browse/ML-7905

[ML-7905]: https://iguazio.atlassian.net/browse/ML-7905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/py/services/api/launcher.py
```diff
@@ -239,6 +240,15 @@ def _pre_run_node_selector_enrichement(self, runtime, run):
             run.spec.node_selector = resolved_node_selectors
         return run
 
+    def _pre_run_image_pull_secret_enrichment(self, runtime, run):
+        """
+        Enrich the run object with the project's image pull secret.
+        This ensures the image pull secret is correctly set on the run
+        while maintaining the runtime's integrity from system-specific project settings.
+        """
+        run.spec.image_pull_secret = mlrun.config.config.function.spec.image_pull_secret
```
```suggestion
        run.spec.image_pull_secret = mlrun.config.config.function.spec.image_pull_secret
```

the side effect is that if client set an explicit image_pull_secret you override it

liranbg (changes requested):

server/py/services/api/launcher.py
adamdelman: fixed

tests/runtimes/test_run.py
```diff
@@ -262,13 +262,12 @@ def test_with_limits():
 
 
 def test_new_function_args_with_default_image_pull_secret():
```
liranbg: the test isnt accurate anymore 
before: it verified the default mlconf kicked in
after: it verifies client override works

I think what needs to be done here is moving this test to when function is created at api side, it is enriched with the configuration

adamdelman: fixed

server/py/services/api/tests/unit/assets/sample_function.py
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+import mlrun
+
+mlrun.mlconf.feature_store.flush_interval = None
```
liranbg: you dont need that

liranbg (commented):
minor

liranbg (approved):

PR #7115, opened 2025-01-16
[Vectordb] Use 'db_key' instead of 'key' to indentify artifact

alxtkr77:
https://iguazio.atlassian.net/browse/ML-8939

### DIFF
--- a/mlrun/artifacts/document.py
+++ b/mlrun/artifacts/document.py
@@ -384,7 +384,7 @@ class DocumentArtifact(Artifact):
             metadata[self.METADATA_ORIGINAL_SOURCE_KEY] = self.spec.original_source
             metadata[self.METADATA_SOURCE_KEY] = self.get_source()
             metadata[self.METADATA_ARTIFACT_TAG] = self.tag or "latest"
-            metadata[self.METADATA_ARTIFACT_KEY] = self.key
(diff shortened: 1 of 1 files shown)


mlrun/artifacts/document.py
```diff
@@ -384,7 +384,7 @@ def to_langchain_documents(
             metadata[self.METADATA_ORIGINAL_SOURCE_KEY] = self.spec.original_source
             metadata[self.METADATA_SOURCE_KEY] = self.get_source()
             metadata[self.METADATA_ARTIFACT_TAG] = self.tag or "latest"
-            metadata[self.METADATA_ARTIFACT_KEY] = self.key
+            metadata[self.METADATA_ARTIFACT_KEY] = self.db_key
```
liranbg, line 387: @theSaarco why not resolving by uid?

alxtkr77: Our current method for updating a document artifact—adding or removing items from its collection—involves reading it from the database, modifying the collection list, and then writing the artifact back. However, this process causes the artifact’s UID to change with each update. Consequently, if you stored the artifact’s UID beforehand, it becomes invalid once the collection is updated by anyone else.

assaf758 (approved):

PR #7124, opened 2025-01-18
[KFP] Add one time KFP no compiper warning and bump pipeline-adapters version

adamdelman:


liranbg (changes requested):
run `make upgrade-mlrun-deps-lock` to ensure package is bumped on locked requirements files

PR #7131, opened 2025-01-20
[Migrations] Only kill foreign DB connections if they have a lock acquired

adamdelman:
https://iguazio.atlassian.net/browse/ML-9082


server/py/services/api/migrations/env.py
```diff
@@ -86,18 +86,40 @@ def run_migrations_online():
         )
 
     with connectable.connect() as connection:
-        # Kill all processes connected to the 'mlrun' database except the current connection
+        # Kill all processes connected to the 'mlrun' that have an acquired lock.
```
yaelgen:
```suggestion
        # Kill all processes connected to the 'mlrun' database that have an acquired lock.
```

server/py/services/api/migrations/env.py
```diff
@@ -86,18 +86,40 @@ def run_migrations_online():
...
         connection_ids = connection.execute(
             sqlalchemy.sql.text(
```
yaelgen, line 94: Maybe worth adding a comment explaining what the query does

yaelgen (changes requested):
LGTM, Added some suggestions.
Also, please add the ticket (https://iguazio.atlassian.net/browse/ML-9082) to the PR description

quaark (approved):
Looks good! Agree with @yaelgen's comments

liranbg (approved):

PR #7135, opened 2025-01-20
[Alerts] Event processing optimizations

TomerShor:
The current event cache is mapping `(project, event_kind) -> set(alert_configs)`. If all alerts are configured for the same event kind, we can possibly iterate over hundreds or more alerts, which some might be filtered out. 

This PR introduces a few improvements:
- Change event cache mapping to `(project, event_kind, entity_id) -> set(alert_configs)` - this will allow iterating on much less alerts. A caveat - this will need to be modified when we support multiple entities for a single alert config.
- Increase cache size to 10,000. When having more than 1,000 we will have a cache miss for each alert after 1000th alert.
- When storing alert state, use the parent alert if available instead of joining on alert.
- Rename `post_event` endpoint functions to `process_event` to reduce confusion.

https://iguazio.atlassian.net/browse/ML-7407


server/py/services/alerts/crud/events.py
```diff
@@ -42,8 +43,8 @@ def is_valid_event(project: str, event_data: mlrun.common.schemas.Event):
 
         return bool(event_data.is_valid())
 
-    def add_event_configuration(self, project, event_kind, alert_id):
-        self._cache.setdefault((project, event_kind), set()).add(alert_id)
+    def add_event_configuration(self, project, event_kind, alert_id, entity_id):
+        self._cache.setdefault((project, event_kind, entity_id), set()).add(alert_id)
 
     def remove_event_configuration(self, project, event_kind, alert_id):
         alerts = self._cache.get((project, event_kind), set())
```
rokatyy:
```suggestion
        alerts = self._cache.get((project, event_kind, entity_id), set())
```

rokatyy (approved):
LGTM

PR #7136, opened 2025-01-20
[Alerts] Add debug logs to process event flow

TomerShor:
To better understand https://iguazio.atlassian.net/browse/ML-7407 , add debug logs before and after operations in the process event flow.

This will be deleted later.

### DIFF
--- a/server/py/services/alerts/crud/alerts.py
+++ b/server/py/services/alerts/crud/alerts.py
@@ -193,6 +193,7 @@ class Alerts(
         if not self._event_entity_matches(alert.entities, event_data.entity):
             return
 
+        # TODO: Remove the logs in this function once the flow is stable
         log_kwargs = {
             "alert_id": alert_id,
             "alert_name": alert.name,
@@ -281,6 +282,7 @@ class Alerts(
         active = False
         state["count"] += 1
 
+        # TODO: Remove the logs in this function once the flow is stable
         log_kwargs = {
             "alert_id": alert.id,
             "alert_name": alert.name,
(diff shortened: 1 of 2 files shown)


server/py/services/alerts/crud/alerts.py
```diff
@@ -193,6 +193,18 @@ def process_event(
...
+            "alert_name": alert.name,
+            "event_kind": event_data.kind,
+            "entity": event_data.entity.ids[0],
+            "project": event_data.entity.project,
+            "session": session.hash_key,
+        }
+        logger.debug(
+            "Processing event",
+            **log_kwargs,
+        )
```
rokatyy, line 208: What do you think about wrapping these into `if logger.isEnabledFor(logging.DEBUG):` to mitigate performance impact?

server/py/services/alerts/crud/alerts.py
```diff
@@ -273,14 +304,15 @@ def _handle_notification(
             state["active"] = True
             state_obj["last_activation_id"] = activation_id
 
-        logger.debug("Sending notifications for alert", name=alert.name)
+        logger.debug("Sending notifications for alert", **log_kwargs)
```
rokatyy, line 309: Please add TODOs to remove these logs

server/py/services/alerts/crud/alerts.py
adamdelman: @rokatyy I'm not sure using `isEnabledFor` is relevant for this scenario, as building log_kwargs doesn't seem to be resource/compute intensive.

adamdelman (approved):
LGTM

server/py/services/alerts/crud/alerts.py
TomerShor: Also, if the log level is set to info or higher, we won't print the logs anyway

liranbg (approved):

PR #7138, opened 2025-01-20
[Runs] Add end time field to run object

Yacouby:
This PR introduces the addition of `end_time` field to the run object which enhances the tracking of run durations and improves the runs monitoring jobs ( will be introduced in a following PR ) to monitor by the termination time of the runs instead of their last update time.

- Added the `end_time` field to the runs DB table and created a schema migration to the new field.
- Updated the DB layer to support filtering by `end_time`, both as a start and end range.
- Extended the `list_runs` REST API to include optional filters for it.
- Updated httpdb to support these filters.
- Added the `end_time` field to the client-side ( under `RunStatus` ).

https://iguazio.atlassian.net/browse/ML-8008


server/py/services/api/tests/unit/api/test_runs.py
```diff
@@ -286,6 +286,9 @@ def test_list_runs_times_filters(db: Session, client: TestClient) -> None:
 
     run_1_update_time = datetime.now(timezone.utc)
 
+    time.sleep(0.1)
+    run_1_end_time = datetime.now(timezone.utc)
```
rokatyy: why not using timedelta instead?

server/py/framework/db/sqldb/models.py
```diff
@@ -349,6 +349,7 @@ class Run(Base, mlrun.utils.db.HasStruct):
         # TODO: change to JSON, see mlrun/common/schemas/function.py::FunctionState for reasoning
         body = Column(SQLTypesUtil.blob())
         start_time = Column(SQLTypesUtil.timestamp())
+        end_time = Column(SQLTypesUtil.timestamp())
```
rokatyy: why not datetime? I know that other fields in the table use it, but since it's new one and we are planning to move all things to datetime (because the latest year supported by ts is 2038 iirc)

rokatyy (changes requested):
Overall looks nice, couple of comment

server/py/framework/db/sqldb/models.py
TomerShor: That is correct - https://iguazio.atlassian.net/browse/ML-6921

rokatyy (approved):

TomerShor (approved):

PR #7142, opened 2025-01-21
[API] Make grpcio an explicit requirement

gtopper:
Since it's imported from mlrun-api server code.


liranbg (changes requested):
you would need to run the lock target to ensure the dependency is locked (and used de-facto)

dockerfiles/base/locked-requirements.txt
```diff
@@ -1628,15 +1628,48 @@ grpcio==1.70.0 \
...
     --hash=sha256:2c33bbdbe20188b2953f46f31af669263b6ee2a9b2d38fa0d36ee091532e21bf \
     --hash=sha256:53695f45da07437b7c344ee4ef60d370fd2850179f5a28bb26d8e2aa1102ec11
     # via google-api-core
-grpcio-status==1.62.3 ; python_full_version >= '3.11' \
+grpcio-status==1.62.3 ; python_full_version != '3.10.*' \
     --hash=sha256:289bdd7b2459794a12cf95dc0cb727bd4a1742c37bd823f760236c937e53a485 \
     --hash=sha256:f9049b762ba8de6b1086789d8315846e094edac2c50beaf462338b301a8fd4b8
     # via google-api-core
-grpcio-tools==1.48.2 ; python_full_version < '3.11' \
+grpcio-tools==1.30.0 ; python_full_version < '3.10' \
```
liranbg, line 1639: @gtopper is this expected?

dockerfiles/mlrun-api/requirements.txt
```diff
@@ -18,3 +18,4 @@ aiosmtplib~=3.0
 pydantic>=1,<2
 # ensure kfp is installed
 mlrun-pipelines-kfp-v1-8[kfp]~=0.3.5; python_version < "3.11"
+grpcio~=1.69.0
```
liranbg: why .69 and not 1.70?

liranbg (changes requested):

dockerfiles/mlrun-api/requirements.txt
gtopper: As per the lock file, which has been updated in the meantime. Will change to 1.70.0.

dockerfiles/base/locked-requirements.txt
gtopper: Apparently this happens when running `make upgrade-mlrun-deps-lock` on development.

liranbg (approved):

PR #7144, opened 2025-01-21
[Migrations] Don't fail migrations when killing transient DB connections

adamdelman:
When we kill DB connections holding locks on the mlrun DB, we might try to kill a process that has already been disconnected.
This PR ignores errors in these cases.


server/py/services/api/migrations/env.py
```diff
@@ -100,7 +101,10 @@ def run_migrations_online():
                 host=host,
                 db="mlrun",
             )
-            connection.execute(sqlalchemy.sql.text(f"KILL {connection_id};"))
+            try:
+                connection.execute(sqlalchemy.sql.text(f"KILL {connection_id};"))
+            except sqlalchemy.exc.DBAPIError:
```
liranbg: comment out or do error string search on exception to ignore cases when connection is gone?

liranbg (changes requested):

liranbg (approved):

PR #7146, opened 2025-01-22
[Model Monitoring] Delete controller internal stream on disable API

Eyal-Danieli:
Includes the following changes:

1 . Fix a bug in which the controller internal stream is not being deleted as expected when calling `disable_model_monitoring` API - https://iguazio.atlassian.net/browse/ML-8376.
2. Remove support for `rebuild_images()` in the `enable_model_monitoring()` API due to limitations with rebuilding nuclio functions that use Kafka topics. The expected approach of rebuilding the mm infra is to call `disable_model_monitoring()`, followed by `enable_model_monitoring()` with the new image
3. Fix `TestModelMonitoringInitialize` system test that validates the `disable/enable` APIs flow. Tested with both kafka and v3io resources.

### DIFF
--- a/server/py/services/api/crud/model_monitoring/deployment.py
+++ b/server/py/services/api/crud/model_monitoring/deployment.py
@@ -473,6 +473,7 @@ class MonitoringDeployment:
             function=function,
             function_name=mm_constants.MonitoringFunctionNames.STREAM,
             stream_args=config.model_endpoint_monitoring.serving_stream,
+            keep_stream_if_exists=True,
         )
 
         # Apply feature store run configurations on the serving function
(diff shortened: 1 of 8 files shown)


davesh0812 (approved):

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -289,9 +289,11 @@ def apply_and_create_stream_trigger(
...
         only with stream trigger(s).
 
-        :param function:      The serving function object that will be applied with the stream trigger.
-        :param function_name: The name of the function that be applied with the stream trigger.
-        :param stream_args:   Stream args from the config.
+        :param function:                            The serving function object that will be applied with the stream
+                                                    trigger.
+        :param function_name:                       The name of the function that be applied with the stream trigger.
+        :param stream_args:                         Stream args from the config.
+        :param ignore_stream_already_exists_failure: If True, skip the stream creation if it already exists.
```
assaf758:
```suggestion
        :param ignore_stream_already_exists_failure: If True, ignores TopicAlreadyExistsError error on MM-infra-functions deployment
```

assaf758 (approved):

PR #7147, opened 2025-01-22
[Project] Fix remove_function in project

rokatyy:
Mlrun never actually removes function from the project enitity

Jira - https://iguazio.atlassian.net/browse/ML-9121

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -2824,7 +2824,7 @@ class MlrunProject(ModelObj):
         """
         # TODO: Remove this in 1.10.0
         warnings.warn(
-            "'remove_function' is deprecated and will be removed in 1.9.0. "
+            "'remove_function' is deprecated and will be removed in 1.10.0. "
             "Please use `delete_function` instead.",
             FutureWarning,
         )
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -2822,7 +2822,7 @@ class MlrunProject(ModelObj):
 
         :param name:    name of the function (under the project)
         """
-        # TODO: Remove this in 1.9.0
+        # TODO: Remove this in 1.10.0
         warnings.warn(
             "'remove_function' is deprecated and will be removed in 1.9.0. "
             "Please use `delete_function` instead.",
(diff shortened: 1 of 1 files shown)


moranbental (approved):
LGTM

liranbg (changes requested):
while this is a destructive change - I suggest to do it on 2 phases - see below remove vs delete model monitoring function.

mlrun/projects/project.py
```diff
@@ -2822,6 +2822,23 @@ def remove_function(self, name):
 
         :param name:    name of the function (under the project)
         """
+        # TODO: Remove this in 1.9.0
+        warnings.warn(
+            "'remove_function' is deprecated and will be removed in 1.9.0. "
+            "Please use `delete_function` instead.",
+            FutureWarning,
+        )
+        self.spec.remove_function(name)
```
TomerShor: We keep deprecation notices for 2 versions, so this will be deleted in 1.10.0

TomerShor (changes requested):
LGTM.
I wonder if there

mlrun/projects/project.py
rokatyy: updated

TomerShor: In the warning too, not just the comment :)

TomerShor: And you can use the `@deprecated` decorator

liranbg (approved):

TomerShor (approved):

mlrun/projects/project.py
```diff
@@ -2817,13 +2818,31 @@ def _set_function(
...
+        category=FutureWarning,
+    )
     def remove_function(self, name):
         """remove the specified function from the project
 
         :param name:    name of the function (under the project)
         """
         self.spec.remove_function(name)
 
+    def delete_function(self, name, from_cache_only=False):
```
theSaarco, line 2835: Since this is replacing the existing `remove_function` we should maintain the same default logic, i.e. only delete from client-side project by default. So, I'd change `from_cache_only` to be `True` by default.
As a matter of fact, I think it's better since this is the new functionality to call this parameter `delete_from_db` and reverse its meaning (and of course have it `False` by default).

PR #7148, opened 2025-01-22
[Model Monitoring] Fix getting a model endpoint without a TSDB profile

jond01:
Fixes [ML-9049](https://iguazio.atlassian.net/browse/ML-9049).

This PR makes sure the following APIs work, even without a TSDB datastore profile credential:

* `get_model_endpoint` and `list_model_endpoints`, also with `tsdb_metrics=True`.
* `get_model_endpoint_monitoring_metrics` and `get_model_endpoint_monitoring_metrics_values`. I added a unit test.

Make sure to catch the correct `MLRunNotFoundError`.  
`MLRunInvalidMMStoreTypeError` is raised only when the datastore profile is of an invalid type, not when it is not set.

Set `tsdb_metrics=False` in the serving functions `get_model_endpoint`.

[ML-9049]: https://iguazio.atlassian.net/browse/ML-9049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/model_monitoring/db/tsdb/__init__.py
+++ b/mlrun/model_monitoring/db/tsdb/__init__.py
@@ -75,10 +75,11 @@ def get_tsdb_connector(
     :param secret_provider:         An optional secret provider to get the connection string secret.
     :param profile:                 An optional profile to initialize the TSDB connector from.
 
-    :return: `TSDBConnector` object. The main goal of this object is to handle different operations on the
+    :return: ``TSDBConnector`` object. The main goal of this object is to handle different operations on the
              TSDB connector such as updating drift metrics or write application record result.
-    :raise: `MLRunInvalidMMStoreTypeError` if the user didn't provide TSDB connection
-            or the provided TSDB connection is invalid.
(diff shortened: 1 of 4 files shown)


davesh0812 (commented):
Overall looks good.

I think that because all the endpoints are used by the UI we should catch both  `MLRunNotFoundError` and `MLRunInvalidMMStoreTypeError` and to return an empty list or a list without the tsdb data.

jond01:
> I think that because all the endpoints are used by the UI we should catch both `MLRunNotFoundError` and `MLRunInvalidMMStoreTypeError` and to return an empty list or a list without the tsdb data.

According to the implementation of `get_tsdb_connector`, `MLRunInvalidMMStoreTypeError` is raised only when a fatal condition occurs in which an unsupported profile type is configured for the TSDB. This should not happen there, as it will be blocked in the `set_model_monitoring_credentials` API.
So from my checks, catching only `MLRunNotFoundError` covers what we need.

assaf758 (approved):

PR #7159, opened 2025-01-26
[Project] Rename flag in delete_function

rokatyy:
After offline discussion, it was decided to rename this flag


theSaarco (approved):
LGTM 👍

PR #7160, opened 2025-01-26
[Automation] Add `--chart` CLI option to deployer

gtopper:


liranbg (approved):

automation/deployment/ce.py
```diff
@@ -92,6 +92,10 @@ def cli():
     "--mlrun-version",
     help="Version of mlrun to install. If not specified, will install the latest version",
 )
+@click.option(
+    "--chart-name",
```
liranbg: perhaps just `--chart` would be sufficient?

liranbg (approved):

PR #7161, opened 2025-01-26
[Alert activations] Add project validation when getting alert activation

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-9196

### DIFF
--- a/server/py/services/alerts/main.py
+++ b/server/py/services/alerts/main.py
@@ -481,12 +481,19 @@ class Service(framework.service.Service):
             db_session,
             activation_id,
         )
+        if alert_activation.project != project:
+            raise mlrun.errors.MLRunNotFoundError(
+                f"Alert activation not found. "
+                f"activation_id={activation_id}, "
+                f"project={project}"
+            )
         if name and alert_activation.name != name:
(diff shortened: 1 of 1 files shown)


server/py/services/alerts/main.py
```diff
@@ -481,12 +481,19 @@ async def get_alert_activation(
...
+                f"Alert activation not found. "
+                f"activation_id={activation_id}, "
+                f"project={project}"
+            )
         if name and alert_activation.name != name:
             raise mlrun.errors.MLRunNotFoundError(
                 f"Alert activation not found. "
                 f"activation_id={activation_id}, "
                 f"name={name}"
             )
```
TomerShor, line 495: reuse the same error message in both errors

TomerShor (changes requested):

server/py/services/alerts/main.py
rokatyy: @TomerShor what if name isn't passed?

TomerShor: Oh I missed the difference in project vs name 🤦

TomerShor (approved):

PR #7166, opened 2025-01-26
[Model Monitoring] Initialize the project object as part of the mm app pod initialization

davesh0812:
https://iguazio.atlassian.net/browse/ML-7407

1. Initialize the project object as part of the pod initialization
2. set `save=False` in order to avoid unnecessary PUT 
3. use `db.get_project` instead of `mlrun.load_project` (app & controller).

### DIFF
--- a/mlrun/serving/server.py
+++ b/mlrun/serving/server.py
@@ -553,7 +553,7 @@ class GraphContext:
     @property
     def project_obj(self):
         if not self._project_obj:
-            self._project_obj = mlrun.load_project(url=self.project, save=False)
+            self._project_obj = mlrun.get_run_db().get_project(name=self.project)
         return self._project_obj
 
     @property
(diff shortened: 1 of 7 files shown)


Eyal-Danieli (approved):
LGTM

mlrun/serving/server.py
```diff
@@ -544,11 +544,18 @@ def __init__(
...
+        self._project_obj = None
 
     @property
     def server(self):
         return self._server
 
+    @property
+    def project_obj(self):
+        if not self._project_obj:
+            self._project_obj = mlrun.load_project(url=self.project, save=False)
```
liranbg: load_project does more than just loading the project instance for project operations. e.g `project_setup.py` would be executed part of loading the project.
why not simply using httpdb to get the project and you will have the project instance at hand for project operations (like get function, store artifact, etc).

liranbg (changes requested):

assaf758 (approved):

liranbg (approved):

PR #7167, opened 2025-01-27
[Datastore] Fix s3 size/offset calculations

alxtkr77:
https://iguazio.atlassian.net/browse/ML-9001

### DIFF
--- a/mlrun/datastore/s3.py
+++ b/mlrun/datastore/s3.py
@@ -21,11 +21,18 @@ from fsspec.registry import get_filesystem_class
 
 import mlrun.errors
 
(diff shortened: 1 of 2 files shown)


mlrun/datastore/s3.py
```diff
@@ -21,12 +21,19 @@
...
 class S3Store(DataStore):
     using_bucket = True
 
+    @staticmethod
+    def get_range(size, offset):
+        byterange = f"bytes={offset}-"
+        if size:
+            byterange += str(offset + size - 1)
+        return byterange
+
```
liranbg: move under function starts with __

liranbg (approved):

assaf758 (approved):

PR #7170, opened 2025-01-27
[KFP] Return correct image for workflow runner

adamdelman:
[KFP] Return correct image for workflow runner

Separate the logic used by client-spec synchronization and getting the KFP workflow runner image logic,
so that the default mlrun image is always returned for client-spec and the KFP remote workflow runner
always gets the mlrun-kfp image.

https://iguazio.atlassian.net/browse/ML-9088
https://iguazio.atlassian.net/browse/ML-9198


server/py/services/api/utils/helpers.py
```diff
@@ -15,62 +15,26 @@
...
-        workflow_spec.engine.lower()
-        if (workflow_spec and workflow_spec.engine)
-        else mlrun.common.schemas.workflow.EngineType.LOCAL
+        image = project.spec.default_image
+    else:
+        image = mlrun.mlconf.kfp_image
+    logger.debug(
+        "Resolved KFP image for workflow",
+        project=project,
+        workflow_spec=workflow_spec,
```
liranbg: this is too verbose and spammy as spec could be large.

liranbg (approved):

PR #7171, opened 2025-01-27
[Runs] Update the run's end time when job is in terminal state

Yacouby:
Following the PR which only add the `end_time` field to the run object ( https://github.com/mlrun/mlrun/pull/7138 )

This PR ensures that the `end_time` is set when the job reaches a terminal state, and also modifies the `list_runs` functionality in two places :
1) sending notifications on terminal runs.
2) sending alerts notifications.

to filter the runs by their `end_time` instead of `last_update_time`

Resolves:
https://iguazio.atlassian.net/browse/ML-8008


server/py/framework/db/sqldb/helpers.py
```diff
@@ -45,6 +45,13 @@ def run_start_time(run):
     return parser.parse(ts)
 
 
+def run_end_time(run):
+    ts = get_in(run, "status.end_time", "")
+    if not ts:
+        return None
+    return parser.parse(ts)
```
TomerShor: This can also be shared with `run_start_time` since its exactly the same with a different string

TomerShor (changes requested):

TomerShor (approved):
LGTM 🚀

PR #7176, opened 2025-01-28
[Requirements] Aid correct locking of protobuf dependency

gtopper:
by explicitly specifying it in requirements.

Fixes [ML-9200](https://iguazio.atlassian.net/browse/ML-9200).

[ML-9200]: https://iguazio.atlassian.net/browse/ML-9200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


liranbg (changes requested):
discussed offline to see if there is a way to fix it via how `uv` locks the requirement files
as this requirement is not directly used by mlrun at all, yet the issue is that uv cannot determine correctly the required version per python version.

@jond01 maybe?

gtopper:
@liranbg, see #7183 for an alternative.

gtopper:
Going with the alternative.

PR #7178, opened 2025-01-28
[Model Monitoring] Fix version field resp

royischoss:
https://iguazio.atlassian.net/browse/ML-8609
Change version field in invoke response to model endpoint uid


mlrun/model_monitoring/stream_processing.py
```diff
@@ -392,7 +392,7 @@ def do(self, full_event):
         if not is_not_none(model, [EventFieldType.MODEL]):
             return None
 
-        version = full_event.body.get(EventFieldType.VERSION)
```
davesh0812, line 395: I think you can delete this and also to delete : 
```
versioned_model = f"{model}:{version}" if version else f"{model}:latest"
...
EventFieldType.MODEL: versioned_model,

```

please double check that we don't use  EventFieldType.MODEL in the next steps.

mlrun/serving/v2_serving.py
```diff
@@ -95,9 +95,6 @@ def predict(self, request):
         :param kwargs:     extra arguments (can be accessed using self.get_param(key))
         """
         self.name = name
-        self.version = ""
-        if name and ":" in name:
-            self.version = name.split(":", 1)[-1]
```
davesh0812, line 100: so maybe we can also to delete from the event
```
"version": self.model.version,
```

davesh0812 (commented):
Looks Good,
few comments

mlrun/model_monitoring/stream_processing.py
royischoss: I think that EventFieldType.MODEL can stay as it was entering the do method, and only remove EventFieldType.VERSIONED_MODEL.

mlrun/serving/v2_serving.py
royischoss: Already did :)

davesh0812 (approved):

PR #7179, opened 2025-01-28
[Model Monitoring] add `system_id` to model monitoring resources

Eyal-Danieli:
To better identify mm resources related to a specific system, we will append the `system_id` (was introduced in #6976 ) to the resource names. This change is currently relevant only to `kafka` topics and `tdengine` databases. With this approach, we can use the same resources across multiple systems, even when the project names are similar (resolves https://iguazio.atlassian.net/browse/ML-7804). 

For instance, we created a project named `test-mm-sys-id-v2` in two different systems, using the same `kafka` and `tdengine` resources. Assuming that the `system_id` of the first setup is `6wgc3w` and the `system_id` of the second setup is `rwmk0g`, we can now explore the resources in the shared `kafka` and `tdengine`:

**tdengine**
 2 databases - 1 for each system:
`mlrun_model_monitoring_6wgc3w`
`mlrun_model_monitoring_rwmk0g`

**kafka**
8 topics (including the histogram application) - 4 per each system:
```
' monitoring_stream_6wgc3w_test-mm-sys-id-v2_histogram-data-drift_v1',
 'monitoring_stream_6wgc3w_test-mm-sys-id-v2_model-monitoring-controller_v1',
 'monitoring_stream_6wgc3w_test-mm-sys-id-v2_model-monitoring-writer_v1',
 'monitoring_stream_6wgc3w_test-mm-sys-id-v2_v1',
 'monitoring_stream_rwmk0g_test-mm-sys-id-v2_histogram-data-drift_v1',
 'monitoring_stream_rwmk0g_test-mm-sys-id-v2_model-monitoring-controller_v1',
 'monitoring_stream_rwmk0g_test-mm-sys-id-v2_model-monitoring-writer_v1',
 'monitoring_stream_rwmk0g_test-mm-sys-id-v2_v1'
```

### DIFF
--- a/mlrun/common/model_monitoring/helpers.py
+++ b/mlrun/common/model_monitoring/helpers.py
@@ -47,9 +47,9 @@ def parse_monitoring_stream_path(
             function_name is None
             or function_name == mm_constants.MonitoringFunctionNames.STREAM
         ):
-            stream_uri += f"?topic=monitoring_stream_{project}_v1"
+            stream_uri += f"?topic=monitoring_stream_{project}_{mlrun.mlconf.system_id}_v1"
         else:
-            stream_uri += f"?topic=monitoring_stream_{project}_{function_name}_v1"
+            stream_uri += f"?topic=monitoring_stream_{project}_{mlrun.mlconf.system_id}_{function_name}_v1"
 
     return stream_uri
 
--- a/mlrun/common/model_monitoring/helpers.py
+++ b/mlrun/common/model_monitoring/helpers.py
@@ -47,7 +47,9 @@ def parse_monitoring_stream_path(
             function_name is None
             or function_name == mm_constants.MonitoringFunctionNames.STREAM
         ):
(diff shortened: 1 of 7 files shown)


mlrun/common/model_monitoring/helpers.py
```diff
@@ -47,9 +47,11 @@ def parse_monitoring_stream_path(
             function_name is None
             or function_name == mm_constants.MonitoringFunctionNames.STREAM
         ):
-            stream_uri += f"?topic=monitoring_stream_{project}_v1"
+            stream_uri += (
+                f"?topic=monitoring_stream_{mlrun.mlconf.system_id}_{project}_v1"
```
assaf758, line 51: seems we don't need the v1 anymore..
also - I tend to think it would be easier for us if v3io behave the same (testing-wise)

assaf758 (approved):

PR #7180, opened 2025-01-28
[Model Monitoring] Fix mm endpoints

davesh0812:
https://iguazio.atlassian.net/browse/ML-9212

pass `TestModelEndpointsOperations` sys test
pass test_app (without alerts)

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -3896,7 +3896,7 @@ class HTTPRunDB(RunDBInterface):
         function_name: Optional[str] = None,
         function_tag: Optional[str] = None,
         endpoint_id: Optional[str] = None,
-    ) -> mlrun.common.schemas.ModelEndpoint:
+    ) -> None:
         """
         Updates a model endpoint with the given attributes.
 
@@ -3921,7 +3921,7 @@ class HTTPRunDB(RunDBInterface):
             attributes_keys=attributes_keys,
             model_endpoint=model_endpoint,
         )
-        response = self.api_call(
+        uid = self.api_call(
             method=mlrun.common.types.HTTPMethod.PATCH,
             path=path,
             params={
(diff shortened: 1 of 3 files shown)


mlrun/db/httpdb.py
```diff
@@ -3896,7 +3896,7 @@ def patch_model_endpoint(
         function_name: Optional[str] = None,
         function_tag: Optional[str] = None,
         endpoint_id: Optional[str] = None,
-    ) -> mlrun.common.schemas.ModelEndpoint:
+    ) -> None:
```
Eyal-Danieli, line 3899: remove the 
`:return:                          The updated `ModelEndpoint` object.`
from the func docs

server/py/services/api/api/endpoints/model_endpoints.py
```diff
@@ -56,7 +56,7 @@ async def create_model_endpoint(
     creation_strategy: mm_constants.ModelEndpointCreationStrategy,
     auth_info: schemas.AuthInfo = Depends(framework.api.deps.authenticate_request),
     db_session: Session = Depends(framework.api.deps.get_db_session),
-) -> schemas.ModelEndpoint:
+) -> str:
```
Eyal-Danieli: change back to mep object

Eyal-Danieli (approved):

mlrun/db/httpdb.py
```diff
@@ -3921,16 +3920,20 @@ def patch_model_endpoint(
...
             method=mlrun.common.types.HTTPMethod.PATCH,
             path=path,
             params={
                 "attribute-key": attributes_keys,
             },
             body=model_endpoint.json(),
         )
-
-        return mlrun.common.schemas.ModelEndpoint(**response.json())
+        logger.info(
```
assaf758, line 3931: Do we really need a log of `before` and `after`?

davesh0812: what do you think to leave the before or the after log?
I added the after log because now we return None now

assaf758 (approved):

PR #7192, opened 2025-01-30
[AsyncHTTPClient] Enhance retry mechanism

liranbg:
1. Bump version to latest to accommodate latest fixes
2. Ensure iguazio http client retries are more relaxed

https://iguazio.atlassian.net/browse/ML-8375


server/py/framework/utils/clients/iguazio.py
```diff
@@ -1071,17 +1077,33 @@ async def _send_request_to_api_async(
...
+
+        # override with cherry-picked options
+        if retry_options_override:
+            if retry_options_override.blacklisted_methods is not None:
+                retry_options.blacklisted_methods = (
+                    retry_options_override.blacklisted_methods
+                )
+            retry_options._start_timeout = retry_options_override._start_timeout
+            retry_options._max_timeout = retry_options_override._max_timeout
+            retry_options._factor = retry_options_override._factor
```
TomerShor, line 1101: What if you only want to override one of these parameters?
Here you always take all 4 params

liranbg: then you change the logic as anyway there's a single usage of it - I thought about it but matter of fact, since it is requires more checkups and etc, and its initial values are given by default, ive decided to cut the changes here and give
either you provide all, or go home

TomerShor (approved):

PR #7197, opened 2025-01-30
[Artifacts] Ignore best iteration on latest reassignment after deletion

TomerShor:
When deleting a latest artifact, we move the latest tag to the previous most recent artifact with the same key.
However, if the previous most recent artifact is a part of a hyper param run, the other iterations don't get the latest tag, and remain untagged.

An example scenario where this can cause issues can be found in the ticket description: https://iguazio.atlassian.net/browse/ML-9066

In this PR, we move the latest tag to the previous most recent artifact OR artifacts, by ignoring the `best_iteration` flag when searching for the recent artifacts.
Additionally, if we find that the most recent artifact is a part of a hyper param run (iteration != 0) - we move the `latest` tag to ALL iterations of this run.


server/py/framework/db/sqldb/db.py
```diff
@@ -1987,20 +1997,32 @@ def _has_latest_tag_in_different_iterations(session, object_record):
         )
 
     @staticmethod
-    def _find_previous_most_recent_artifact_id(session, object_record):
-        """Find the most recent artifact id based on the update timestamp, excluding the current artifact."""
-        query = session.query(ArtifactV2.id).filter(
+    def _find_previous_most_recent_artifact_ids(session, object_record):
+        """Find the most recent artifact ids based on the update timestamp, excluding the current artifact."""
+        query = session.query(ArtifactV2).filter(
```
moranbental: Why retrieve the whole artifact? and not just the id and iteration?

moranbental (changes requested):
LGTM, just one suggestion

moranbental (approved):

server/py/framework/db/sqldb/db.py
```diff
@@ -1987,20 +1997,48 @@ def _has_latest_tag_in_different_iterations(session, object_record):
...
+                )
+                .outerjoin(ArtifactV2.Tag, ArtifactV2.Tag.obj_id == ArtifactV2.id)
+                .filter(
+                    or_(
+                        ArtifactV2.Tag.name.is_(None),
+                        ArtifactV2.Tag.name != mlrun.common.constants.RESERVED_TAG_NAME_LATEST,
+                    )
+                )
+            )
+            return [result[0] for result in query.all()]
```
moranbental: Move that to a separate function

moranbental (changes requested):

server/py/framework/db/sqldb/db.py
TomerShor: I think this is sill relevant for the `_find_previous_most_recent_artifact_ids` , so it makes sense in the same function

moranbental (approved):

PR #7198, opened 2025-01-30
[Runtime] Enrich image pull secrets for deployed functions

adamdelman:
https://iguazio.atlassian.net/browse/ML-9220


liranbg (commented):
make sure you add UT coverage

server/py/services/api/launcher.py
```diff
@@ -240,17 +244,25 @@ def _pre_run_node_selector_enrichement(self, runtime, run):
...
 
-    def _pre_run_image_pull_secret_enrichment(self, run):
+    def _pre_run_image_pull_secret_enrichment(
+        self,
+        runtime: "mlrun.runtimes.base.BaseRuntime",
+        run: mlrun.run.RunObject,
+    ):
         """
-        Enrich the run object with the project's image pull secret.
+        Enrich both the run object and runtime with the project's image pull secret.
```
liranbg: The reason we would not want to enrich runtime with image pull secret is that it is subjected to its environment - which means, if we move (export/import) the runtime between envs (imagine having staging + prod envs) then the configuration would persist.
what we do is enriching the run (and not the runtime) configuration.

if there is an issue with specific runtime (e.g nuclio) then we do the enrichment specifically for that runtime

liranbg (changes requested):

server/py/services/api/launcher.py
```diff
@@ -240,17 +244,25 @@ def _pre_run_node_selector_enrichement(self, runtime, run):
...
         This ensures the image pull secret is correctly set on the run,
         either from the run spec or from the MLRun config
         """
         existing_image_pull_secret = getattr(run.spec, "image_pull_secret", None)
+        if existing_image_pull_secret is not None:
```
liranbg: you may assume that `existing_image_pull_secret` is a string not a config object

server/py/services/api/launcher.py
```diff
@@ -240,17 +244,25 @@ def _pre_run_node_selector_enrichement(self, runtime, run):
...
+            existing_image_pull_secret = existing_image_pull_secret.default
         run.spec.image_pull_secret = (
             existing_image_pull_secret
-            or mlrun.config.config.function.spec.image_pull_secret
+            or mlrun.config.config.function.spec.image_pull_secret.default
         )
+
+        runtime.spec.image_pull_secret = run.spec.image_pull_secret
```
liranbg: if the run is enriched with the image pull secret, do you need it on the runtime as well?

liranbg (approved):

PR #7199, opened 2025-01-30
[Automation] Fix patch remote's wait for pod readiness

TomerShor:
Since some deployments' strategy is RollingUpdate (e.g. Worker), using `kubectl wait --for condition=Ready` sometimes times out because it waits for the terminating pod to be ready. 
This results with the script failing despite a successful patch, which is not a nice experience.

To mitigate this, we wait for condition with a smaller timeout (20s), and retry manually until the real timeout is passed (300s).


liranbg (commented):
🚀

liranbg (approved):

PR #7203, opened 2025-02-01
[Model Monitoring] Fix credentials exposure in YAML

alxtkr77:
https://iguazio.atlassian.net/browse/ML-7622

When building model monitoring graphs, we previously used TDEngine and Parquet targets with their corresponding private credentials encoded directly in the Python YAML code. To improve security, this PR implements datastore profiles instead of direct DSN names and credentials, preventing the exposure of sensitive authentication information.

### DIFF
--- a/mlrun/datastore/storeytargets.py
+++ b/mlrun/datastore/storeytargets.py
@@ -42,11 +42,10 @@ def get_url_and_storage_options(path, external_storage_options=None):
 
 
 class TDEngineStoreyTarget(storey.TDEngineTarget):
-    def __init__(self, *args, **kwargs):
-        path = kwargs.pop("url")
-        if path and path.startswith("ds://"):
+    def __init__(self, *args, url: str, **kwargs):
+        if url.startswith("ds://"):
             datastore_profile = (
-                mlrun.datastore.datastore_profile.datastore_profile_read(path)
+                mlrun.datastore.datastore_profile.datastore_profile_read(url)
             )
             if not isinstance(
                 datastore_profile,
@@ -56,9 +55,8 @@ class TDEngineStoreyTarget(storey.TDEngineTarget):
                     f"Unexpected datastore profile type:{datastore_profile.type}."
                     "Only TDEngineDatastoreProfile is supported"
                 )
-            path = datastore_profile.dsn()
-        kwargs["url"] = path
-        super().__init__(*args, **kwargs)
+            url = datastore_profile.dsn()
+        super().__init__(*args, url=url, **kwargs)
 
 
 class StoreyTargetUtils:
--- a/mlrun/datastore/storeytargets.py
+++ b/mlrun/datastore/storeytargets.py
@@ -43,7 +43,13 @@ def get_url_and_storage_options(path, external_storage_options=None):
 
 class TDEngineStoreyTarget(storey.TDEngineTarget):
     def __init__(self, *args, **kwargs):
-        kwargs["url"] = mlrun.model_monitoring.helpers.get_tsdb_connection_string()
+        path = kwargs.pop("url")
+        if path and path.startswith("ds://"):
+            datastore_profile = (
+                mlrun.datastore.datastore_profile.datastore_profile_read(path)
+            )
+            path = datastore_profile.dsn()
+        kwargs["url"] = path
         super().__init__(*args, **kwargs)
 
 
--- a/mlrun/datastore/storeytargets.py
+++ b/mlrun/datastore/storeytargets.py
@@ -48,6 +48,14 @@ class TDEngineStoreyTarget(storey.TDEngineTarget):
             datastore_profile = (
                 mlrun.datastore.datastore_profile.datastore_profile_read(path)
             )
+            if not isinstance(
+                datastore_profile,
+                mlrun.datastore.datastore_profile.TDEngineDatastoreProfile,
+            ):
+                raise ValueError(
+                    f"Unexpected datastore profile type:{datastore_profile.type}."
+                    "Only TDEngineDatastoreProfile is supported"
+                )
             path = datastore_profile.dsn()
         kwargs["url"] = path
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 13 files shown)


mlrun/datastore/storeytargets.py
```diff
@@ -43,7 +43,13 @@ def get_url_and_storage_options(path, external_storage_options=None):
 
 class TDEngineStoreyTarget(storey.TDEngineTarget):
     def __init__(self, *args, **kwargs):
-        kwargs["url"] = mlrun.model_monitoring.helpers.get_tsdb_connection_string()
+        path = kwargs.pop("url")
+        if path and path.startswith("ds://"):
+            datastore_profile = (
+                mlrun.datastore.datastore_profile.datastore_profile_read(path)
+            )
+            path = datastore_profile.dsn()
```
jond01: Please add an explicit check that the profile is of the expected type - `TDEngineDatastoreProfile`. Otherwise, an unexplained error will be raised, something like "no dsn method for object ...".

mlrun/model_monitoring/db/tsdb/__init__.py
```diff
@@ -92,7 +96,6 @@ def get_tsdb_connector(
         profile, mlrun.datastore.datastore_profile.TDEngineDatastoreProfile
     ):
         tsdb_connector_type = mlrun.common.schemas.model_monitoring.TSDBTarget.TDEngine
-        kwargs["connection_string"] = profile.dsn()
```
jond01, line 95: It's a good time to remove also L90:
```py
        kwargs["v3io_access_key"] = profile.v3io_access_key
```
And set the access key from the profile in `V3IOTSDBConnector`.

mlrun/model_monitoring/db/tsdb/__init__.py
```diff
@@ -29,7 +29,9 @@ class ObjectTSDBFactory(enum.Enum):
     v3io_tsdb = "v3io-tsdb"
     tdengine = "tdengine"
 
-    def to_tsdb_connector(self, project: str, **kwargs) -> TSDBConnector:
+    def to_tsdb_connector(
+        self, project: str, connection_profile, **kwargs
```
jond01: 1. Missing type hint
2. Missing docstring param
3. I think that `tsdb_profile` is a better name, even just `profile`.

tests/model_monitoring/db/tsdb/tdengine/test_tdengine_connector.py
```diff
@@ -43,9 +44,10 @@ def is_tdengine_defined() -> bool:
 def connector() -> Iterator[TDEngineConnector]:
     connection = taosws.connect()
```
jond01: Did it work for you?
@tomerm-iguazio found the issue in L44 where the `connection_string` is missing. It should be:
```py
    connection = taosws.connect(connection_string)
```

mlrun/projects/project.py
```diff
@@ -3782,6 +3782,24 @@ def set_model_monitoring_credentials(
...
+                profile = mlrun.datastore.datastore_profile.DatastoreProfileS3(
+                    name=mm_constants.DefaultProfileName.PARQUET,
+                    access_key_id=mlrun.get_secret_or_env("AWS_ACCESS_KEY_ID"),
+                    secret_key=mlrun.get_secret_or_env("AWS_SECRET_ACCESS_KEY"),
+                    endpoint_url=mlrun.get_secret_or_env("S3_ENDPOINT_URL"),
+                    force_non_anonymous=mlrun.get_secret_or_env("S3_NON_ANONYMOUS"),
+                    profile_name=mlrun.get_secret_or_env("AWS_PROFILE"),
+                    assume_role_arn=mlrun.get_secret_or_env("MLRUN_AWS_ROLE_ARN"),
+                )
+                self.register_datastore_profile(profile)
```
jond01: I don't really like this solution for two reasons:

1. You rely on the existence of the environment variables in the client's SDK environment. However, many times they are just not there.
2. `set_model_monitoring_credentials` should not create or register datastore profiles under the hood. The current creation and registration is a transitionary state, and will be removed soon.

mlrun/projects/project.py
```diff
@@ -3791,7 +3785,7 @@ def set_model_monitoring_credentials(
         db.set_model_monitoring_credentials(
             project=self.name,
             credentials={
-                "access_key": access_key,
+                "access_key": None,
```
jond01, line 3788: I would remove it altogether from the API endpoint:
https://github.com/mlrun/mlrun/blob/07e64f979bab02b31cfe72aa7eabb4d45d38f794/server/py/services/api/api/endpoints/model_monitoring.py#L372
But you can leave it to me - I'll do it in [ML-9020](https://iguazio.atlassian.net/browse/ML-9020).

mlrun/datastore/storeytargets.py
```diff
@@ -69,7 +83,12 @@ def process_args_and_kwargs(args, kwargs):
 
 class ParquetStoreyTarget(storey.ParquetTarget):
     def __init__(self, *args, **kwargs):
+        alt_key_name = kwargs.pop("alternative_v3io_access_key", None)
         args, kwargs = StoreyTargetUtils.process_args_and_kwargs(args, kwargs)
+        storage_options = kwargs.get("storage_options", {})
+        if storage_options and storage_options.get("v3io_access_key") and alt_key_name:
```
jond01, line 87: I wonder why you need this condition: `storage_options.get("v3io_access_key")`.
I think it should be ok to set the `v3io_access_key` parameter from `alt_key_name` without checking prior existence.

mlrun/datastore/storeytargets.py
```diff
@@ -43,7 +43,21 @@ def get_url_and_storage_options(path, external_storage_options=None):
 
 class TDEngineStoreyTarget(storey.TDEngineTarget):
     def __init__(self, *args, **kwargs):
```
jond01: It looks like you can safely change the signature to the more explicit form:
```suggestion
    def __init__(self, *args, url: str, **kwargs) -> None:
```

jond01 (approved):
Great job! I left some comments.

mlrun/datastore/storeytargets.py
alxtkr77: To handle non v3io parquet (CE for example)

mlrun/projects/project.py
alxtkr77: Go ahead :)

mlrun/datastore/storeytargets.py
jond01: I see, it's a nontrivial point actually

jond01 (approved):

PR #7209, opened 2025-02-02
[Model Monitoring] Controller fix- allow 'cron' event.trigger.type

royischoss:
https://iguazio.atlassian.net/browse/CEML-325
Fix adding 'cron' option for cron trigger type other than 'http' more over checking for all trigger kinds including stream to prevent wrong event handling.
includes small fix for test_inference_feature_set


davesh0812 (approved):
LGTM

PR #7210, opened 2025-02-02
[Security] Fix verify_label_key

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-9160

To fix the problem, we should parse the key properly and ensure that the check for reserved prefixes is done accurately. Instead of using string operations, we can split the key and check the prefix part explicitly. This will ensure that the reserved prefixes are correctly identified and handled.


liranbg (approved):

adamdelman (approved):
LGTM

yaelgen (approved):

PR #7211, opened 2025-02-02
[Logging] Use contextvars to get consistent request IDs when logging

adamdelman:
This allows us to get the request ID when running code both in threadpools and coroutines


mlrun/utils/logger.py
```diff
@@ -352,11 +356,20 @@ def bind(self, **kw_args):
...
         if kw_args:
             self._logger.log(
-                level, message, *args, exc_info=exc_info, extra={"with": kw_args}
+                level,
+                message,
+                *args,
+                exc_info=exc_info,
+                extra={
+                    "with": kw_args,
+                    "request_id": request_id,
```
liranbg:
```suggestion
                    "ctx": request_id,
```

request_id is just a usage of a context-based information, meaning, request id is usually for api request id. sometimes you have internal procedure running, and its context is not really a request_id.
`ctx` makes it consistent with other services and a bit more generic.

mlrun/utils/logger.py
```diff
@@ -352,11 +356,20 @@ def bind(self, **kw_args):
     def _update_bound_vars_and_log(
         self, level, message, *args, exc_info=None, **kw_args
     ):
+        request_id = request_id_var.get() or "N/A"
```
liranbg: can be left empty, I think
```suggestion
        request_id = request_id_var.get() or ""
```

mlrun/utils/logger.py
```diff
@@ -249,6 +250,9 @@ def _have_color_support(self):
         return stdout.isatty()
 
 
+request_id_var = contextvars.ContextVar("request_id", default=None)
```
liranbg: what does it store for background threads w/out any context of api requests?

liranbg (commented):
very much like it!
questions and suggestions

mlrun/utils/logger.py
adamdelman: The default would be None, we can add something to support that as well I guess

liranbg (approved):

PR #7216, opened 2025-02-02
[Pipelines] Add "Unknown" pipeline run status

TomerShor:
Pipelines v1 has an ["Unknown" pipeline run status](https://github.com/kubeflow/pipelines/blob/cd66b6965c1576d698fa6623cbe7c77cc7e8b5ed/backend/src/apiserver/model/run.go#L46), so add it to the supported statuses, which resolves to the "Unknown" MLRun RunStatus.

https://iguazio.atlassian.net/browse/ML-9231


adamdelman (commented):
LGTM

It is worth considering if this state should be supported in the pipelines retry flow.

liranbg (changes requested):
You will need to do it in 2 phases
1. Add code to mlrun-pipelines + merge it + release it
2. Bump requirement to use it + make the mlrun changes to use the new code


Also, when you change pipeline-adapters code, make sure to bump its version on its `setup.py`.

PR #7218, opened 2025-02-02
[Model Monitoring] Update Grafana to align with the new structure

Eyal-Danieli:
Update Grafana APIs and dashboards to support the new model monitoring schemas and resources. This PR includes the following changes:

- Implement two new APIs for querying model endpoint uids and metrics: `grafana_list_endpoints_uids` and `grafana_list_metrics`.
- Remove deprecated APIs: `individual_feature_analysis`, `overall_feature_analysis`, and `incoming_features`.
- Modify dashboards to align with the new v3io tsdb tables and add descriptions to each chart/KPI.

Below are examples of the updated dashboards:

image
image
image
image
image

https://iguazio.atlassian.net/browse/ML-9203


server/py/services/api/crud/model_monitoring/grafana.py
```diff
@@ -56,232 +52,57 @@ def grafana_list_projects(
...
+    project = query_parameters.get("project")
     if project:
         await framework.utils.auth.verifier.AuthVerifier().query_project_permissions(
             project,
             mlrun.common.schemas.AuthorizationAction.read,
             auth_info,
         )
-    endpoint_list = await run_in_threadpool(
-        services.api.crud.ModelEndpoints().list_model_endpoints,
+    endpoint_list = await services.api.crud.ModelEndpoints().list_model_endpoints(
```
davesh0812, line 78: You want to show all the meps or just the active?
If only the active (like the UI) you can use `latest_only=True`

server/py/services/api/crud/model_monitoring/grafana.py
```diff
@@ -291,122 +112,90 @@ async def grafana_overall_feature_analysis(
...
+        ):
+            continue
+        row = [
+            endpoint.metadata.uid,
+            endpoint.metadata.name,
+            endpoint.spec.function_name,
+            endpoint.spec.model_name,
+            endpoint.spec.model_class,
+            endpoint.status.error_count,
+            endpoint.status.result_status,
```
davesh0812, line 194: why to use this? and not to retrieve  it from frames?

server/py/services/api/crud/model_monitoring/grafana.py
Eyal-Danieli: that's a good question, I think that for now as you suggested we should display only the active meps

server/py/services/api/crud/model_monitoring/grafana.py
Eyal-Danieli: In grafana dashboard we can't use multiple datasources on the same chart. Now, in this case these values are being retrieved as part of a summary table of the model endpoints which is through `MLRun API` datasource and not `frames`. Unfortunately that also means that these values are not affect by the selected time range (we explain in the chart docs).

davesh0812 (approved):

assaf758 (approved):

PR #7221, opened 2025-02-03
[Config] Use orderdict to preserve key order in env parsing

Yacouby:
Using dict did not guarantee key order, which caused inconsistent behavior. Replacing it with OrderedDict ensures that environment variables are processed in the correct sequence.

Jira:
https://iguazio.atlassian.net/browse/ML-8460

### DIFF
--- a/mlrun/__init__.py
+++ b/mlrun/__init__.py
@@ -26,6 +26,7 @@ __all__ = [
     "VolumeMount",
 ]
 
+import collections
 from os import environ, path
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


moranbental (approved):
LGTM

liranbg (approved):

rokatyy (commented):
Does it mean that in client configuration clients should be very careful with configuration order?

Yacouby:
> Does it mean that in client configuration clients should be very careful with configuration order?

No, the function now ensures that specific important keys (defined in `mlconf.get_ordered_keys()`) are always processed first, while the rest of the environment variables maintain their original order. 
This guarantees that critical settings (like `MLRUN_HTTPDB__HTTP__VERIFY`) are applied before others, ensuring proper connection setup without requiring manual ordering by clients

Yacouby:
> Does it mean that in client configuration clients should be very careful with configuration order?

No, the function now ensures that specific important keys (defined in `mlconf.get_ordered_keys())` are always processed first, while the rest of the environment variables maintain their original order.

rokatyy (approved):

PR #7222, opened 2025-02-03
[Runs] Enhancements to preemption handling

yaelgen:
resolves: 
In this PR: https://iguazio.atlassian.net/browse/ML-8601
- Expanded test coverage for `enrich_function_preemption_spec`, ensuring expected behavior across different preemption modes.
- Introduced user warnings in `with_node_selection` when preemptible node selectors, tolerations, or affinities are set based on environment configurations, clarifying that they may be pruned at runtime.
- Modified `allow` mode behavior to retain preemptible node selectors and affinity/anti-affinity, ensuring that a combination of `allow` + preemptible selector explicitly enforces scheduling on spot nodes.

For convenience, I also added a Notion table summarizing how each preemption mode affects node selection, tolerations, and affinity handling.

The warning for the users:
image

### DIFF
--- a/mlrun/runtimes/pod.py
+++ b/mlrun/runtimes/pod.py
@@ -1207,9 +1207,9 @@ class KubeResource(BaseRuntime):
         """
         preemptible_tolerations = [
             k8s_client.V1Toleration(
-                key=toleration["key"],
-                value=toleration["value"],
-                effect=toleration["effect"],
+                key=toleration.get("key"),
+                value=toleration.get("value"),
+                effect=toleration.get("effect"),
             )
             for toleration in mlconf.get_preemptible_tolerations()
         ]
--- a/mlrun/runtimes/pod.py
+++ b/mlrun/runtimes/pod.py
@@ -17,6 +17,7 @@ import os
 import re
 import time
 import typing
+import warnings
 from collections.abc import Iterable
 from enum import Enum
 
@@ -720,12 +721,6 @@ class KubeResourceSpec(FunctionSpec):
                 affinity_field_name=affinity_field_name,
             )
 
-            # remove preemptible nodes constrain
-            self._prune_node_selector(
-                mlconf.get_preemptible_node_selector(),
-                node_selector_field_name=node_selector_field_name,
-            )
-
             # enrich with tolerations
             self._merge_tolerations(
                 generate_preemptible_tolerations(),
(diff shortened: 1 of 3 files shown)


mlrun/runtimes/pod.py
```diff
@@ -703,29 +704,8 @@ def enrich_function_preemption_spec(
                 ),
                 affinity_field_name=affinity_field_name,
             )
-        # purge any affinity / anti-affinity preemption related configuration and enrich with preemptible tolerations
+        # enrich with preemptible tolerations
```
TomerShor: can remove comment completely, it is duplicated 2 lines below.

mlrun/runtimes/pod.py
```diff
@@ -1201,6 +1181,91 @@ def with_requests(
...
+        preemptible_tolerations = [
+            k8s_client.V1Toleration(
+                key=toleration.get("key"),
+                value=toleration.get("value"),
+                effect=toleration.get("effect"),
+            )
+            for toleration in mlconf.get_preemptible_tolerations()
+        ]
+
+        transform_attribute_to_k8s_class_instance("tolerations", tolerations)
```
TomerShor: why is this needed and you can't just use `mlconf.get_preemptible_tolerations()`?

mlrun/runtimes/pod.py
```diff
@@ -1201,6 +1181,91 @@ def with_requests(
...
+                            message="The selected node affinity constraints may be adjusted at runtime"
+                        )
+
+    def raise_preemptible_warning(self, message: str) -> None:
+        warnings.warn(
+            f"Warning: {message} based on the preemptible node settings configured in your MLRun configuration. "
+            f"This adjustment depends on the function's preemption mode. "
+            f"The list of potential adjusted preemptible selectors can be viewed here: "
+            f"mlrun.mlconf.get_preemptible_node_selector() and mlrun.mlconf.get_preemptible_tolerations()."
+            f"For more information about preemptible mode, see:  "
```
TomerShor: see...?

TomerShor: If I use `with_node_selection` with mutliple values (e.g. setting NS, affinity and tolerations), I have a possibility to get this message many many times - once for each conflicted label, for each conflicted toleration, and for the affinity.

While this is probably an edge case, how about instead of printing the same warning every time, you accumulate a list of the conflicts, and warn it once (or at least, once for each type)?

TomerShor (changes requested):

mlrun/runtimes/pod.py
yaelgen: oopsi 👀

mlrun/runtimes/pod.py
yaelgen: The response of `mlconf.get_preemptible_tolerations()` is a dictionary, and I wanted to compare the `V1Toleration` objects, but maybe the `transform_attribute_to_k8s_class_instance("tolerations", tolerations) ` can be removed, since the` self.spec.tolerations` runs it prior to this function

liranbg (approved):

TomerShor (approved):
LGTM

PR #7224, opened 2025-02-03
[Docs] Add v1.8 to change log

jillnogold:
replaced by https://github.com/mlrun/mlrun/pull/7850


moranbental (changes requested):
please merge development to your branch

alonmr (changes requested):
you have many changes that are not related to the docs. your git branch is not aligned with development. easiest way to fix it will be to open a new branch from development and copy you changes there - then open a new PR.
you also need to verify that your local development branch is aligned. if that is not the case you might need help to reset it.
if you open a PR and the issue is still present we'll know that there is an issue

PR #7225, opened 2025-02-03
[Security] Fix ECR url detection to prevent substring matching issues

Yacouby:
Parse the registry URL and check the hostname instead of using substring matching to ensure accurate detection of ECR URLs and prevent potential security issues

Jira:
https://iguazio.atlassian.net/browse/ML-9145


adamdelman (approved):
LGTM

PR #7227, opened 2025-02-03
[Security] Enforce strict SSH host key verification

Yacouby:
Replaced WarningPolicy with RejectPolicy to enforce strict SSH host key verification and prevent potential man-in-the-middle attacks.
Jira:
https://iguazio.atlassian.net/browse/ML-9163

### DIFF
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -987,7 +987,7 @@ def fill_object_hash(object_dict, uid_property_name, tag=""):
     # originated from a datetime, for example), but since this is a one-way dump only for hash calculation,
     # it's valid here.
     data = json.dumps(object_dict, sort_keys=True, default=str).encode()
-    h = hashlib.sha1()
+    h = hashlib.sha256()
     h.update(data)
     uid = h.hexdigest()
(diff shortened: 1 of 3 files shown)


mlrun/utils/helpers.py
```diff
@@ -987,7 +987,7 @@ def fill_object_hash(object_dict, uid_property_name, tag=""):
     # originated from a datetime, for example), but since this is a one-way dump only for hash calculation,
     # it's valid here.
     data = json.dumps(object_dict, sort_keys=True, default=str).encode()
-    h = hashlib.sha1()
+    h = hashlib.sha256()
```
TomerShor: As we spoke offline, this can break compatibility as the hash is used to get functions/artifacts before updating them (in the `store` flows).
Changing the hashing will return a different UID for existing resources, so we might "miss" them and create new resources instead of update.

TomerShor (changes requested):

liranbg (approved):
Approving to quiet this one although it might be more of false positive for the reason that this script is used when deploying mlrun/ce during CI.

PR #7230, opened 2025-02-03
[Artifacts] Add Project-BestIteration-Updated Index for List Artifacts Query

quaark:
https://iguazio.atlassian.net/browse/ML-9189

The index is used to speed up the list artifact requests, most of which filter by best iteration and all of which always query by project and order by updated time.
Most queries also filter by kind, however to quote @theSaarco , adding the column to the index actually hurts performance:

> I actually got better results when creating an index that doesn't have the kind field in it. Reason is that the sorting of the index is done in the order of the fields, such that when asking for artifacts_v2.kind NOT IN ('model', 'dataset', 'document') - the DB needs to go through the entire index and filter out records with mismatching kinds. When I created an index just on project/bi/updated and ran the same query, it returned practically immediately within 0.01secs. This because it can go over the index in order and filter out anything with wrong kind, then stop once it gets enough records.
When asking for a specific kind (for example model) both indexes perform the same since the DB can still filter on the kind using the index and get back ordered results.

In addition, the index needs to be used explicitly in the query as MySQL doesn't automatically recognize this query should use the newly added index.
(For more info on the specific reasonings for this implementation - see the ticket)

Lastly, since the indexes are getting quite messy, added explanations and relevant tickets for all indexes except `idx_artifacts_name_uid_project` for which I haven't yet found the reason.


server/py/framework/db/sqldb/db.py
```diff
@@ -1515,7 +1515,7 @@ def _find_artifacts(
         query = session.query(ArtifactV2).with_entities(
             ArtifactV2.id,
             ArtifactV2.Tag.name,
-        )
+        ).with_hint(ArtifactV2, "USE INDEX idx_project_bi_updated")
```
theSaarco: We don't want to use this index for all queries, since this bypasses the DB optimizer - please see in the discussion in the Jira - 

> the challenge is to correctly tell the DB when to use these indexes. We need to be very careful not to interfere too much with the DB’s optimizer, and intervene only when we know we’re right.

This needs to be added only if doing pagination or ordering by updated. Possibly other conditions.

theSaarco (changes requested):
Sababa, need to be more specific about when to use this index.

theSaarco (approved):
Looks good. I actually expected a more "hard-coded" thing here, but no complaints :-)

liranbg (approved):

PR #7231, opened 2025-02-03
[Pipelines] Add `Unknown` to retryable statuses

TomerShor:
Currently we only retry on a stable pipeline status (a terminal state basically), but an "Unknown" state can possibly also be terminated, and there is no reason to not allow retrying on it.

This PR creates the `retryable_statuses` list, and a followup will use this list in [here](https://github.com/mlrun/mlrun/blob/916f8e006b04216883bae9f8487d223935eb7554/server/py/services/api/crud/pipelines.py#L287).

### DIFF
--- a/pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/models.py
+++ b/pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/models.py
@@ -111,3 +111,6 @@ class RunStatuses(StrEnum):
             for status in RunStatuses.all()
             if status not in RunStatuses.stable_statuses()
         ]
+
+    def retryable_statuses(self):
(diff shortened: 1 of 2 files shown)


liranbg (approved):

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/models.py
```diff
@@ -111,3 +111,6 @@ def transient_statuses():
             for status in RunStatuses.all()
             if status not in RunStatuses.stable_statuses()
         ]
+
+    def retryable_statuses(self):
+        return self.stable_statuses() + [RunStatuses.unknown]
```
TomerShor: Add unspecified status too

PR #7232, opened 2025-02-03
[Runs] Improve delete runs flow

TomerShor:
When deleting a job that has many runs (100s or 1000s), deleting each of the runs separately can cause stress on the DB even when done in batches.
For example, if more then 6 delete job requests arrive, with a batching of 10 we will get more than 60 concurrent select and deletion requests to the DB. The default MySQL QueuePool limit is 64, so we get the following error:
```
QueuePool limit of size 64 overflow 64 reached, connection timed out, timeout 30.00 (Background on this error at: https://sqlalche.me/e/14/3o7r)
```

In this PR, we introduce some improvements and optimizations to the `delete_runs` CRUD:
- Concurrently:
  - Delete each project's runs from the DB at once using a list of uids
  - Delete each project's runs' logs at once using a list of uids
- Use batched concurrency only for deleting each run's runtime resources
- Skip deletion of runtime resources if the time passed from the run's start time is more than the deletion grace period + 1 day. This heuristic should ensure the run are already deleted.
- Move some code around to reuse functionalities.

https://iguazio.atlassian.net/browse/ML-9155

### DIFF
--- a/server/py/services/api/crud/runs.py
+++ b/server/py/services/api/crud/runs.py
@@ -270,7 +270,13 @@ class Runs(
             )
             return
 
-        self._delete_run_resources(db_session, project, uid, run)
+        await run_in_threadpool(
+            self._delete_run_resources,
+            db_session,
+            project,
+            uid,
+            run,
+        )
 
         # get runtime kind for logging
         runtime_kind = (
@@ -331,12 +337,11 @@ class Runs(
         project_to_run_uids_to_delete = {}
 
         # Delete each run's resources asynchronously in batches
-        while runs_list:
+        for chunked_run_list in mlrun.utils.helpers.iterate_list_by_chunks(
+            runs_list, mlrun.mlconf.crud.runs.batch_delete_runs_chunk_size
+        ):
             tasks = []
-            runs_batch = runs_list[
-                : mlrun.mlconf.crud.runs.batch_delete_runs_chunk_size
-            ]
-            for run in runs_batch:
+            for run in chunked_run_list:
                 project_to_run_uids_to_delete.setdefault(run.project, []).append(
                     run.uid
                 )
@@ -365,8 +370,6 @@ class Runs(
                         project=project,
                         error=mlrun.errors.err_to_str(result),
                     )
-
-            runs_list = runs_list[mlrun.mlconf.crud.runs.batch_delete_runs_chunk_size :]
 
         # Delete each project runs in parallel, since log deletion doesn't support "*" projects
         if project_to_run_uids_to_delete:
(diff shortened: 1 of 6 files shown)


server/py/services/api/crud/runs.py
```diff
@@ -346,15 +328,27 @@ async def delete_runs(
...
         last_exception = None
+        project_to_run_uids_to_delete = {}
+
+        # Delete each run's resources asynchronously in batches
         while runs_list:
             tasks = []
-            for run in runs_list[: mlrun.mlconf.crud.runs.batch_delete_runs_chunk_size]:
+            runs_batch = runs_list[
+                : mlrun.mlconf.crud.runs.batch_delete_runs_chunk_size
+            ]
```
liranbg: see `iterate_list_by_chunks`

server/py/services/api/crud/runs.py
```diff
@@ -512,6 +521,79 @@ def _enrich_run_artifacts(
...
+                    force=True,
+                )
+
+    async def _delete_runs(self, db_session, project: str, uids: list[str]):
+        # Delete runs from DB
+        framework.utils.singletons.db.get_db().del_runs(
+            session=db_session,
+            project=project,
+            uids=uids,
+        )
```
liranbg: this one is blocking. run in thread

server/py/services/api/crud/runs.py
```diff
@@ -269,33 +270,14 @@ async def delete_run(
...
 
-        run_state = run.get("status", {}).get("state")
-        if (
-            run_state
-            in mlrun.common.runtimes.constants.RunStates.not_allowed_for_deletion_states()
-        ):
-            raise mlrun.errors.MLRunInvalidArgumentError(
-                f"Can not delete run in {run_state} state, consider aborting the run first"
-            )
+        self._delete_run_resources(db_session, project, uid, run)
```
liranbg: this function is async, make sure calls are running in threadpool

liranbg (changes requested):
very well optimized. some changes are needed

server/py/services/api/crud/runs.py
```diff
@@ -363,15 +362,28 @@ async def delete_runs(
...
+        # Delete each project runs in parallel, since log deletion doesn't support "*" projects
+        if project_to_run_uids_to_delete:
+            tasks = []
+            for project, run_uids_to_delete in project_to_run_uids_to_delete.items():
+                tasks.append(
+                    framework.db.session.run_function_with_new_db_session(
+                        self._delete_runs,
+                        project,
+                        run_uids_to_delete,
+                    )
```
liranbg, line 383: threadpool

liranbg (approved):

PR #7233, opened 2025-02-04
[Pipelines] Fix class method `retryable_statuses`

TomerShor:
It should be a `classmethod` to be available directly from the class.


liranbg (approved):

adamdelman (approved):
lgtm

PR #7235, opened 2025-02-04
[Tools] Make patch_remote.py docker image push concurrent

adamdelman:


automation/patch_igz/patch_remote.py
```diff
@@ -273,15 +274,19 @@ def _tag_images_for_multi_node_registries(self, built_images):
...
-            self._exec_local(
-                cmd=[
-                    "docker",
-                    "push",
-                    image,
-                ],
-                live=True,
-            )
+        with ThreadPoolExecutor(max_workers=len(built_images)) as executor:
+            futures = {
```
liranbg, line 278: just use list 🤷🏼

adamdelman: But then I can't print which image failed

PR #7237, opened 2025-02-04
[CE Deployer] Fix unsafe usage of subprocess.Popen

adamdelman:
https://iguazio.atlassian.net/browse/ML-9166


automation/deployment/deployer.py
```diff
@@ -752,28 +754,29 @@ def run_command(
...
-        command += " " + " ".join(args)
-
-    process = subprocess.Popen(
-        command,
-        stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
-        stdin=subprocess.PIPE,
-        shell=True,
-    )
+        command = [command] + args
```
TomerShor, line 760: While I see that `command` is only passed as a one-word string, perhaps best to verify it by splitting on space and taking the first item.

TomerShor (changes requested):
minor

TomerShor (approved):

liranbg (approved):

PR #7241, opened 2025-02-05
[Model Monitoring] Improve performance of MEP augmentation from V3IO TSDB

gtopper:
Reduces the delta between `list_model_endpoints(tsdb_metrics=False)` and `list_model_endpoints(tsdb_metrics=True)` from about 18 seconds to under 1 second.

[ML-9176](https://iguazio.atlassian.net/browse/ML-9176)

[ML-9176]: https://iguazio.atlassian.net/browse/ML-9176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -737,11 +741,74 @@ def get_avg_latency(
...
+            run_in_threadpool(self.get_avg_latency, endpoint_ids=uids),
+            run_in_threadpool(self.get_drift_status, endpoint_ids=uids),
+        ]
+
+        (
+            error_count_df,
+            last_request_df,
+            avg_latency_df,
+            drift_status_df,
+        ) = await asyncio.gather(*coroutines)
```
davesh0812, line 774: Can we put this in the base? to avoid code duplication.

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -935,3 +971,99 @@ def get_avg_latency(
...
+            last_request_res,
+            avg_latency_res,
+            drift_status_res,
+        ) = await asyncio.gather(*coroutines)
+
+        model_endpoint_objects_by_uid = {}
+        for model_endpoint_object in model_endpoint_objects:
+            model_endpoint_objects_by_uid[model_endpoint_object.metadata.uid] = (
+                model_endpoint_object
+            )
```
davesh0812: I think you don't need it, you doing the same in the first loop.

davesh0812 (commented):
Nice!!!

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
gtopper: Potentially, but I'm still looking into poor performance on the TDEngine side. Once that's settled, I'll look back to consider refactoring commonalities.

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
gtopper: Thanks, that was a refactoring accident.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
gtopper: I'm going to leave the TDEngine issue for another PR since it's not a regression here, but trying to share this bit of code turns out too ugly for too little benefit, so probably not worth the trouble (plus, TDEngine code might change soon anyway).

davesh0812 (approved):

PR #7243, opened 2025-02-05
[Notifications] Fix `add_notification` not taking effect on KFP engine workflows

quaark:
https://iguazio.atlassian.net/browse/ML-9097

The `project.notifiers` isn't aware of the project spec where the notifications are now saved, moreover `add_notification` didn't supply all the schema of the notification. Causing the new logic of calling MLRun API to ad-hoc send the notifications only works when supplying them fully with `project.run(..., notifications=[...])` which adds the notifications to the project spec.

Added `server_notifactions` member to `project.notifiers` and extended `add_notification`'s function signature to accept all notification attributes (the `when` will be all supported states if not given explicitly). When `add_notification` receives a server side notification it is added to that list. And finally when running the pipeline, the notifications get added to `project.notifiers.server_notifactions` via `add_notification` and ultimately get inserted in the project spec.

### DIFF
--- a/mlrun/common/runtimes/constants.py
+++ b/mlrun/common/runtimes/constants.py
@@ -195,6 +195,10 @@ class RunStates:
         ]
 
     @staticmethod
+    def notification_states():
+        return RunStates.terminal_states() + [RunStates.running]
+
+    @staticmethod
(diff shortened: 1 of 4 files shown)


TomerShor (approved):
Looks good to moi

PR #7244, opened 2025-02-05
[Model Monitoring] Fix allow 'evaluate' to work with mep names 

royischoss:
https://iguazio.atlassian.net/browse/ML-8915
Main changes: 
1) evaluate: endpoints input can be a string or list of strings stands for mep names will retrieve 'latest'
2) Change list_model_endpoints to filter by list of names preventing multiple queries


mlrun/model_monitoring/applications/base.py
```diff
@@ -145,6 +146,35 @@ def call_do_tracking(event: Optional[dict] = None):
         else:
             return call_do_tracking()
 
+    @staticmethod
+    def _handle_endpoints_type(project: str, endpoints) -> list[tuple[str, str]]:
```
davesh0812: add typing to all the params

mlrun/model_monitoring/applications/base.py
```diff
@@ -145,6 +146,35 @@ def call_do_tracking(event: Optional[dict] = None):
...
+                )
+                if endpoints_list:
+                    endpoints = [
+                        (endpoint.metadata.name, endpoint.metadata.uid)
+                        for endpoint in endpoints_list
+                    ]
+                else:
+                    raise mlrun.errors.MLRunNotFoundError(
+                        f"Did not find any model endpoint named ' {endpoints}'"
+                    )
```
davesh0812: I think you should check that the len of model endpoints are the same as the users list and that you got 1 mep per each , and if not to raise this exception.

server/py/services/api/tests/unit/db/test_model_endpoint.py
```diff
@@ -301,6 +301,21 @@ def test_list_filters(self) -> None:
...
+        ).endpoints
+        assert len(list_mep) == 1
+
+        list_mep = self._db.list_model_endpoints(
+            self._db_session,
+            project=model_endpoint.metadata.project,
+            names=["model-endpoint-1"],
+        ).endpoints
+        assert len(list_mep) == 2
+
```
davesh0812, line 328: please add another endpoint with diffrent name and try to get them both, or some new unit test that test your changes.

davesh0812 (commented):
Nice solution.
few comments

mlrun/model_monitoring/applications/base.py
royischoss: We decided to log warning if mep name not found and allow retrieving all meps of the same name - 'latest'

mlrun/model_monitoring/applications/base.py
```diff
@@ -137,14 +138,64 @@ def call_do_tracking(event: Optional[dict] = None):
...
+                    retrieve_ep_names = list(
+                        map(lambda endpoint: endpoint[0], list_endpoints_result)
+                    )
+                    missing = set(
+                        [endpoints] if isinstance(endpoints, str) else endpoints
+                    ) - set(retrieve_ep_names)
+                    if missing:
+                        logger.warning(
+                            f"Could not list endpoint named: {list(missing)} retrieve: {list_endpoints_result} as "
+                            f"endpoints"
```
davesh0812:
```suggestion
                        logger.warning(
                            "Could not list all the required endpoints.",  missing_endpoint=missing
                            )
```

davesh0812 (approved):

davesh0812 (approved):

PR #7246, opened 2025-02-05
[Config] Prioritize httpdb config first when reading env variables

Yacouby:
This change ensures that the `MLRUN_HTTPDB__HTTP__VERIFY` environment variable is processed first in the configuration, allowing SSL verification settings to be applied early in the configuration process.

The changes that were introduced in this PRs to fix the issue:
https://github.com/mlrun/mlrun/pull/7065
https://github.com/mlrun/mlrun/pull/7221

were specific to `set_env_from_file`, but that was not the flow that caused the bug here, so those changes were reverted, and since setting env from file flow calls the config reload then this case is also covered in this PR.

Jira:
https://iguazio.atlassian.net/browse/ML-8460

### DIFF
--- a/mlrun/__init__.py
+++ b/mlrun/__init__.py
@@ -26,7 +26,6 @@ __all__ = [
     "VolumeMount",
 ]
 
-import collections
 from os import environ, path
 from typing import Optional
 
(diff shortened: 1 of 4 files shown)


mlrun/__init__.py
```diff
@@ -215,40 +214,8 @@ def set_env_from_file(env_file: str, return_dict: bool = False) -> Optional[dict
...
-    The function ensures that environment variables defined in the `ordered_keys` list
-    are added to the result dictionary first. Any other environment variables from
-    `env_vars` are then added in the order they appear in the input dictionary.
-
-    :param env_vars: A dictionary where each key is the name of an environment variable (str),
-                      and each value is the corresponding environment variable value (str).
-    :return: A dictionary with the processed environment variables, ordered with the specific
-             keys first, followed by the rest in their original order.
-    """
-    ordered_keys = mlconf.get_ordered_keys()
```
TomerShor, line 240: It this function still needed?

Yacouby: I think yes because it is used in the system tests ( `tests/system/base.py` )

TomerShor: But there its also called right before `mlrun.mlconf.reload()`, which now handles it anyways

TomerShor (approved):

PR #7257, opened 2025-02-09
[Tests] Fix `TestBasics::test_basics`

gtopper:
Following breakage in #6905.


tests/system/base.py
```diff
@@ -370,19 +370,20 @@ def _verify_run_outputs(
         loss: typing.Optional[int] = None,
         best_iteration: typing.Optional[int] = None,
         iteration_results: bool = False,
+        fragment: str = "",
```
moranbental:
```suggestion
        iteration: typing.Optional[int] = None,
```
and If iteration is not None, it will prepend the value with a #

moranbental (changes requested):

moranbental (approved):
LGTM

PR #6796, opened 2024-11-27
[Model Monitoring] manage model endpoints in mlrun db

davesh0812:
implements https://iguazio.atlassian.net/browse/ML-7589

PR #7258, opened 2025-02-10
[Tests] Fix `TestMonitoredServings::test_different_kind_of_serving`

gtopper:
Following API change in #6796.


tests/system/model_monitoring/test_app.py
```diff
@@ -1250,15 +1250,9 @@ def test_different_kind_of_serving(self) -> None:
...
-                    ],
-                    feature_set_uri=endpoint[
-                        mm_constants.EventFieldType.FEATURE_SET_URI
-                    ],
-                    model_dict=self.router_models[
-                        endpoint[mm_constants.EventFieldType.MODEL].split(":")[0]
-                    ],
+                    model_name=endpoint.spec.model_name,
+                    feature_set_uri=endpoint.spec.monitoring_feature_set_uri,
+                    model_dict=self.router_models[endpoint.spec.model_name],
```
davesh0812:
```suggestion
                    model_dict=self.router_models[endpoint.metadata.name],
```

davesh0812 (changes requested):

tests/system/model_monitoring/test_app.py
gtopper: Why? The router models dict is keyed by the model name, not the endpoint name.

davesh0812: Nop it keyed by the model endpoint name.

tests/system/model_monitoring/test_app.py
```diff
@@ -1248,15 +1249,9 @@ def test_different_kind_of_serving(self) -> None:
...
-                    model_name=endpoint[mm_constants.EventFieldType.MODEL].split(":")[
-                        0
-                    ],
-                    feature_set_uri=endpoint[
-                        mm_constants.EventFieldType.FEATURE_SET_URI
-                    ],
-                    model_dict=self.router_models[
-                        endpoint[mm_constants.EventFieldType.MODEL].split(":")[0]
-                    ],
+                    model_name=endpoint.spec.model_name,
```
davesh0812: here also you need to pass the endpoint name, you can also change the name of the function param for clarity. 
```suggestion
                    model_name=endpoint.metadata.name,
```

davesh0812 (changes requested):

davesh0812 (approved):

PR #7261, opened 2025-02-11
[KFP] Resolve kfp fixed image

liranbg:
Ensure remote workflow pipeline has its client version fixed in so that when enrichment happen (later on) it won't set it to newer server version.
The flow is as such

1. workflow image is set (via `resolve_client_default_kfp_image`)
2. the remote workflow function is saved
3. its run is created
4. its image is enriched with server version

so, if client 1.7 is running, it needs the `mlrun/mlrun` with version 1.7 to ensure code is compatible with client env.
if client wants to get newer version of remote workflow, it needs to have newer client version.

Important to note, that on 1.8 we have removed `kfp` from `mlrun/mlrun` image and thus, for <1.8 we pick `mlrun/mlrun` and for >= 1.8 we pick `mlrun/mlrun-kfp`.

https://iguazio.atlassian.net/browse/ML-9292

### DIFF
--- a/server/py/services/api/utils/helpers.py
+++ b/server/py/services/api/utils/helpers.py
@@ -24,8 +24,8 @@ from mlrun.utils import logger
 
 
 def resolve_client_default_kfp_image(
-    project: typing.Optional[ProjectOut] = None,
-    workflow_spec: typing.Optional[WorkflowSpec] = None,
+    project: ProjectOut,
+    workflow_spec: WorkflowSpec,
     client_version: typing.Optional[str] = None,
 ) -> str:
     if workflow_spec and workflow_spec.image:
@@ -65,7 +65,7 @@ def resolve_client_default_kfp_image(
 
     logger.debug(
         "Resolved KFP image for workflow",
-        project=project,
+        project_name=project.metadata.name,
         client_version=client_version,
         workflow_spec_image=workflow_spec.image,
         project_spec_default_image=project.spec.default_image,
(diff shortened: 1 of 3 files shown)


server/py/services/api/utils/helpers.py
```diff
@@ -36,25 +36,39 @@ def resolve_client_default_kfp_image(
         must_use_mlrun_image = False
         if client_version and "unstable" not in client_version:
             try:
-                client_version = semver.Version.parse(client_version)
                 # client is olden than (<) 1.8, must use mlrun image for kfp
-                if client_version < semver.Version.parse("1.7.9999"):
+                if semver.Version.parse(client_version) < semver.Version.parse(
+                    "1.7.9999"
+                ):
```
TomerShor:
```suggestion
                if semver.Version.parse(client_version) < semver.Version.parse("1.8.0-rc0"):`
```

server/py/services/api/utils/helpers.py
```diff
@@ -36,25 +36,39 @@ def resolve_client_default_kfp_image(
...
             image = mlrun.mlconf.default_base_image
+            if ":" not in image:
+                # enrich the image with the client version to ensure that
+                # client < 1.8 will use the correct mlrun image and version.
+                # https://iguazio.atlassian.net/browse/ML-9292
+                enriched_image = mlrun.utils.enrich_image_url(
+                    image, client_version=client_version
+                )
+                logger.debug(
+                    "Ensure KFP image has fixed client version",
```
TomerShor:
```suggestion
                    "Ensuring KFP image has fixed client version",
```

TomerShor (changes requested):

TomerShor (approved):

PR #7250, opened 2025-02-05
[Model Monitoring] Remove deprecated set credentials arguments

jond01:
Implements [ML-9020](https://iguazio.atlassian.net/browse/ML-9020), [ML-9058](https://iguazio.atlassian.net/browse/ML-9058), and fixes [ML-9042](https://iguazio.atlassian.net/browse/ML-9042).

This PR breaks the `project.set_model_monitoring_credentials` API.
Previously, one could pass the deprecated `tsdb_connection` and `stream_path` arguments to this API, and the datastore profiles were registered as a part of this method.
Now, one has to take care of the profiles outside of this method and pass to this method only their names: `tsdb_profile_name` and `stream_profile_name`.

The way to enable MM infrastructure is now documented in this method:

```py
import mlrun
from mlrun.datastore.datastore_profile import DatastoreProfileV3io

project = mlrun.get_or_create_project("mm-infra-setup")

# Create and register TSDB profile
tsdb_profile = DatastoreProfileV3io(
    name="my-v3io-tsdb",
)
project.register_datastore_profile(tsdb_profile)

# Create and register stream profile
stream_profile = DatastoreProfileV3io(
    name="my-v3io-stream",
    v3io_access_key=mlrun.mlconf.get_v3io_access_key(),
)
project.register_datastore_profile(stream_profile)

# Set model monitoring credentials and enable the infrastructure
project.set_model_monitoring_credentials(
    tsdb_profile_name=tsdb_profile.name,
    stream_profile_name=stream_profile.name,
)
project.enable_model_monitoring()
```

To configure the MM datastore for running system tests - see the updated `tests/system/env-template.yml`.
A valid V3IO configuration can be:

```yml
# tests/system/env-template.yml
mlrun_model_monitoring_tsdb_profile:
  name: my-mm-tsdb
  type: v3io

mlrun_model_monitoring_stream_profile:
  name: my-mm-stream
  type: v3io
```

The TSDB can be changed to be a `TDEngineDatastoreProfile`, and the stream can be changed to be a `DatastoreProfileKafkaSource`.

Update the usage of `project.set_model_monitoring_credentials` in the tutorials and system tests - use the relevant datastore profiles, register them, and pass their names to this method.
Specifically for the system tests, add a helper MM system test class - `TestMLRunSystemModelMonitoring`, which parses the profile data into a datastore profile object.

The backward compatibility test failed due to:

- Delete the deprecated `POST /api/v1/projects/{project}/model-monitoring/set-model-monitoring-credentials`
[...]

PR #7265, opened 2025-02-11
[Tests] Fix MM env preparation for system tests

jond01:
Following #7250, the following workflow:
https://github.com/mlrun/mlrun/actions/workflows/system-tests-enterprise.yml
Fails on 1.7.x, e.g.: https://github.com/mlrun/mlrun/actions/runs/13231699922/job/36929984760#step:9:311

The reason is that the same development version of the preparation script is used for all the tested branches.

### DIFF
--- a/automation/system_test/prepare.py
+++ b/automation/system_test/prepare.py
@@ -399,16 +399,28 @@ class SystemTestPreparer:
         self._env_config["V3IO_API"] = f"https://{v3io_api_host}"
         self._env_config["MLRUN_DBPATH"] = f"https://{mlrun_api_url}"
 
-        self._env_config["mlrun_model_monitoring_tsdb_profile"] = {
-            "type": "v3io",
-            "name": "mm-tsdb-profile",
-            "v3io_access_key": None,
-        }
-        self._env_config["mlrun_model_monitoring_stream_profile"] = {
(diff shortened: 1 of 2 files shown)


automation/system_test/prepare.py
```diff
@@ -399,16 +399,28 @@ def _enrich_env(self):
...
-        }
+        # Since the prepare script is shared across branches, two MM configs are set.
+        # Remove the deprecated config when we stop testing 1.7.x.
+
+        # MM infra for < 1.8.0
+        self._env_config["MLRUN_MODEL_ENDPOINT_MONITORING__TSDB_CONNECTION"] = "v3io"
+        self._env_config["MLRUN_MODEL_ENDPOINT_MONITORING__STREAM_CONNECTION"] = "v3io"
+
+        # MM infra for >= 1.8.0
+        self._env_config["mlrun_model_monitoring_tsdb_profile"] = json.dumps(
```
liranbg, line 410: why not using capitalization if that is how, by convention, we set the env config?

jond01: Because it's not really an env var, but a datastore profile configuration/object for the system tests.
There is no matching configuration in `mlrun/config.py`, but it might be added in the future.

PR #7279, opened 2025-02-12
[Function] Show Warning When Inline Encoded Source Code is Over 10KB

quaark:
https://iguazio.atlassian.net/browse/ML-9127

When inline source code is over 10kb, the function may not work.
Raising a warning with a suggestion to use `with_source_archive` in this case.


moranbental (approved):
LGTM

liranbg (approved):

mlrun/utils/helpers.py
```diff
@@ -2129,3 +2129,16 @@ def as_dict(data: typing.Union[dict, str]) -> dict:
...
+
+def encode_user_code(
+    user_code: str, max_len_warning: typing.Optional[int] = None
+) -> str:
+    max_len_warning = max_len_warning or config.function.spec.source_code_max_bytes
+    encoded = base64.b64encode(user_code.encode("utf-8")).decode("utf-8")
+    if len(encoded) > max_len_warning:
+        logger.warning(
+            f"User code exceeds the maximum allowed size of {max_len_warning} bytes. "
+            "Consider using `with_source_archive`."
```
GiladShapira94: maybe worth to explain that by default the user code is not from the remote source (base64 embedded code) and if that issue occur the user should use remote source.
for example:
```
User code exceeds the maximum allowed size of {max_len_warning} bytes for non remote source. 
Consider using `with_source_archive` to add user code as a remote source to the function.
```

GiladShapira94: LGTM, the comment above is very minor

PR #7280, opened 2025-02-12
[System id] Generate system ID using only lowercase letters and digits

Yacouby:
This PR updates the system ID generation to ensure it follows stricter naming rules. 
The previous system ID generation involved using a URL-safe Base64 encoding, which could include special characters like `-` . 
The new approach generates a 6-character system ID using only lowercase letters (a-z) and digits (0-9), ensuring it remains alphanumeric and compatible with Kubernetes resource naming conventions.

Jira:
https://iguazio.atlassian.net/browse/ML-9239

### DIFF
--- a/server/py/services/api/initial_data.py
+++ b/server/py/services/api/initial_data.py
@@ -1140,17 +1140,7 @@ def _generate_system_id() -> str:
     valid_chars = string.ascii_lowercase + string.digits
     system_id_len = 6
 
-    while True:
-        system_id = "".join(random.choices(valid_chars, k=system_id_len))
-
-        # validate the result against the qualified_name regex
-        try:
-            mlrun.utils.helpers.verify_field_regex(
-                "system_id", system_id, mlrun.utils.regex.qualified_name
-            )
-            return system_id
-        except mlrun.errors.MLRunInvalidArgumentError:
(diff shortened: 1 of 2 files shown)


server/py/services/api/initial_data.py
```diff
@@ -1136,21 +1136,19 @@ def _get_configured_system_id() -> typing.Optional[str]:
...
         # validate the result against the qualified_name regex
         try:
             mlrun.utils.helpers.verify_field_regex(
-                "system_id", sanitized_str, mlrun.utils.regex.qualified_name
+                "system_id", system_id, mlrun.utils.regex.qualified_name
             )
-            return sanitized_str
+            return system_id
         except mlrun.errors.MLRunInvalidArgumentError:
             continue
```
TomerShor: Is this really still needed?
Alphanumeric-only will always pass this validation, no?

TomerShor (changes requested):

server/py/services/api/initial_data.py
Yacouby: you are correct, looks like the validation step is redundant and we can remove it.

TomerShor (approved):

PR #7287, opened 2025-02-13
[Model Monitoring] Fix the deletion of leftover mm resources when removing a model endpoint

Eyal-Danieli:
Fix an issue where during the deletion of a model endpoint (when not part of a project deletion cycle), some data is left behind, including the mm feature set and TSDB data associated with the deleted model endpoint. Note that the deletion of the TSDB records will be done in a new background task. 
This PR Includes:

- New `delete_tsdb_records` API in the TSDB connector: This API will be invoked when a model endpoint is deleted (e.g., during an `overwrite` creation strategy) to remove any associated TSDB records.
- New `delete_feature_sets` API in the sql db: This API enables the deletion of multiple feature sets and will be called when model endpoints are deleted. 

Resolves https://iguazio.atlassian.net/browse/ML-8787


server/py/services/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -810,16 +839,18 @@ def _delete_model_endpoint_monitoring_infra(self, uids: list[str], project: str)
...
-            logger.info("TSDB resources were not deleted")
+            tsdb_connector = mlrun.model_monitoring.get_tsdb_connector(
+                project=project,
+                secret_provider=services.api.crud.secrets.get_project_secret_provider(
+                    project=project
+                ),
+            )
+            for uid in uids:
+                # todo: optimize to delete all in one call
+                tsdb_connector.delete_tsdb_records(endpoint_id=uid)
```
liranbg: ⬆️

liranbg: @Eyal-Danieli is this deletion part of project deletion? regardless, worth putting effort into batch deletion (delete many uids at once / chunks) to ensure timeout wont expired per its operation

Eyal-Danieli: it's not part of the project deletion process, and we don’t expect it to be called with a large number of model endpoints (like in the project deletion API). 
Usually, we expect it to be used when overwriting an existing model endpoint (which requires deleting the previous one) or when exceeding the archive limit for the same model endpoint and then you need to delete older versions.

Eyal-Danieli: btw in project deletion process we delete tables and not specific records

Eyal-Danieli: @liranbg @davesh0812 updated to batch deletion. Please note the differences between `v3io` and `tdengine`:
1. `v3io`: due to filter limitations, we delete data in chunks, with a max of 200 meps per chunk.
2. `tdengine`: Since each mep is stored in its own table and we cannot delete multiple tables in a single query, we continue to drop these tables individually. However, querying the table names is now done in a single query. We retrieve all relevant table names and then proceed to delete each table one by one.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -266,6 +266,65 @@ def handle_model_error(
...
+                get_subtable_query = self.tables[table]._get_subtables_query_by_tag(
+                    filter_tag="endpoint_id", filter_values=endpoint_ids
+                )
+                subtables_result = self.connection.run(
+                    query=get_subtable_query,
+                    timeout=self._timeout,
+                    retries=self._retries,
+                )
+                subtables.extend([subtable[0] for subtable in subtables_result.data])
+        except Exception as e:
```
davesh0812, line 315: can we be more specific with the exception ?

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -266,6 +266,65 @@ def handle_model_error(
...
+            drop_statements.append(
+                self.tables[table].drop_subtable_query(subtable=subtable)
+            )
+        try:
+            self.connection.run(
+                statements=drop_statements,
+                timeout=self._timeout,
+                retries=self._retries,
+            )
+        except Exception as e:
```
davesh0812, line 336: same for here

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -428,6 +428,44 @@ def delete_tsdb_resources(self, table: Optional[str] = None):
...
+        if len(endpoint_ids) > V3IO_MEPS_LIMIT:
+            # Split the endpoint ids into chunks to avoid exceeding the v3io-engine filter-expression limit
+            endpoint_id_chunks = [
+                endpoint_ids[i : i + V3IO_MEPS_LIMIT]
+                for i in range(0, len(endpoint_ids), V3IO_MEPS_LIMIT)
+            ]
+        else:
+            endpoint_id_chunks = [endpoint_ids]
+
+        for endpoint_id_chunk in endpoint_id_chunks:
```
davesh0812:
```suggestion
    for i in range(0, len(endpoint_ids), V3IO_MEPS_LIMIT):
        endpoint_id_chunks = endpoint_ids[i : i + V3IO_MEPS_LIMIT]
```

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -428,6 +428,44 @@ def delete_tsdb_resources(self, table: Optional[str] = None):
...
+            filter_query = f"endpoint_id IN({str(endpoint_id_chunk)[1:-1]}) "
+            for table in tables:
+                try:
+                    self.frames_client.delete(
+                        backend=_TSDB_BE,
+                        table=self.tables[table],
+                        filter=filter_query,
+                        start="0",
+                    )
+                except Exception as e:
```
davesh0812, line 453: same for here

server/py/services/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -480,6 +503,16 @@ async def _archive_model_endpoint(
...
+                feature_set_uids = [
+                    f"{framework.db.sqldb.db.unversioned_tagged_object_uid_prefix}{uid}_"
+                    for uid in old_uids
+                ]
+                await run_in_threadpool(
+                    framework.utils.singletons.db.get_db().delete_feature_sets,
+                    session=db_session,
+                    project=model_endpoint.metadata.project,
+                    uids=feature_set_uids,
+                )
```
davesh0812: Maybe its better to put it inside `_delete_model_endpoint_monitoring_infra`, and to avoid code duplication

davesh0812 (commented):
Nice, few comments

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
Eyal-Danieli: just an update following our discussion - we want to keep this exception generic as it because we don't want to break the api of deleting infra if something has broken with removing the old data from the tsdb. The data is stored in the user's tdengine, and in the event of an error in this point, the mep will no longer be available through other APIs (because it has been deleted from mlrundb). As a result, outdated data for unsupported meps may remain in the user's tdengine. Please note that we do log the err as a warn and recommend that the user manually delete the data.

davesh0812 (approved):

assaf758 (approved):

PR #7288, opened 2025-02-13
[Helpers] Convert timezone aware datetime to UTC

Yacouby:
The `datetime_from_iso` function has been updated to ensure that timezone-aware datetime objects are properly converted to UTC when parsed. This guarantees that all datetime values are stored and processed consistently in UTC at the backend level.

This issue was not limited to the alert activations SDK but also affected other SDK resources with list functions. Previously, these functions did not explicitly ensure that the `since` and `until` datetimes were in UTC, which could lead to inconsistencies in how dates and times were interpreted. 

Jira:
https://iguazio.atlassian.net/browse/ML-9128

### DIFF
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -1336,7 +1336,7 @@ def datetime_to_iso(time_obj: Optional[datetime]) -> Optional[str]:
     if not time_obj:
         return
 
-    # if the datetime object has a timezone, convert it to UTC before formatting.
+    # if the datetime object has a timezone, convert it to UTC before formatting
     if time_obj.tzinfo:
         time_obj = time_obj.astimezone(timezone.utc)
     return time_obj.isoformat()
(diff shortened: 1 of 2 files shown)


rokatyy (approved):

mlrun/utils/helpers.py
```diff
@@ -1335,6 +1335,10 @@ def datetime_from_iso(time_str: str) -> Optional[datetime]:
 def datetime_to_iso(time_obj: Optional[datetime]) -> Optional[str]:
     if not time_obj:
         return
+
+    # if the datetime object has a timezone, convert it to UTC before formatting
+    if time_obj.tzinfo:
+        time_obj = time_obj.astimezone(timezone.utc)
```
liranbg: We can't just override given timezone as it has meaning from user pov.
what you want to do is align user input with how it is stored on backend. is that correct? or perhaps the time obj didnt have timezone and you wanted to ensure tz (therefor the if condition is wrong)
anyway, what should be done is you need to align the datetime on the backend side, not client side. as you have UI as client as well sending requests with different timezones.

liranbg (changes requested):

liranbg (approved):

PR #7294, opened 2025-02-16
[Artifacts] Handle Iterations When Ensuring the Latest Tag During Migration

moranbental:
This PR addresses the issue of correctly assigning the "latest" tag during data migration, specifically separating the logic for artifacts with iter == 0 and iter != 0. Previously, the "latest" tag assignment did not account for distinguishing between these two cases.

**_Changes:_**

To implement this, we have separated the artifacts into two sets of (project, key) tuples:
- One set for artifacts with iter == 0 and the "latest" tag.
- Another set for artifacts with iter > 0 and the "latest" tag.

During the migration process, when iterating over all artifacts that do not have the "latest" tag, the following checks are performed:

1. For artifacts with iter == 0:
Mark as "latest" only if the project's key combination does not already exist in the set of iter > 0 "latest" items.
Once marked as "latest", add the (project, key) to the iter == 0 set.

1. For artifacts with iter > 0:
Mark as "latest" only if the project's key combination does not already exist in the set of iter == 0 "latest" items.
After marking, add the (project, key) to the iter > 0 set.

_**Note:**_ 
In cases where the same project and key were created from both a hyper-params run and a single run, and the user removed the 'latest' tag from all items, we will assign the 'latest' tag to either the hyper-params items or the single run items. This will depend on which item we encounter first when iterating over the results, and it may not necessarily be really the latest.

https://iguazio.atlassian.net/browse/ML-9285

### DIFF
--- a/server/py/services/api/initial_data.py
+++ b/server/py/services/api/initial_data.py
@@ -1041,7 +1041,7 @@ def _ensure_latest_tag_for_artifacts(
     )
 
     # Collecting the two sets of (project, key) tuples
-    project_key_iter_0 = (
+    project_key_iter_0 = set(
         latest_with_iter_0.with_entities(
             subquery_filtered.c.project, subquery_filtered.c.key
         )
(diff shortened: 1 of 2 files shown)


liranbg (approved):
lgtm

PR #7297, opened 2025-02-16
[Model Monitoring] Allow adding outputs using 'add_model'

royischoss:
https://iguazio.atlassian.net/browse/ML-7578
Added param to `add_model` allowing to add output after training, allows the user to use `log_model` without outputs and add it later in the process.


mlrun/data_types/infer.py
```diff
@@ -75,6 +78,8 @@ def upsert_entity(name, value_type):
                 features[column].value_type = value_type
             else:
                 features[column] = {"name": column, "value_type": value_type}
+                if isinstance(features, ObjectList) and column in label_columns:
+                    features.move_to_end(column, last=False)
```
davesh0812: add inline comment that explain this

mlrun/data_types/infer.py
```diff
@@ -33,12 +34,14 @@ def infer_schema_from_df(
     entities,
     timestamp_key: Optional[str] = None,
     entity_columns=None,
+    label_columns: Optional[list[str]] = None,
```
davesh0812: please add this param as the last param for BC

davesh0812 (commented):
Nice, please also edit the logic as we talked

mlrun/runtimes/nuclio/serving.py
```diff
@@ -408,6 +409,8 @@ def add_model(
             * **archive**:
             1. If model endpoints with the same name exist, preserve them.
             2. Create a new model endpoint with the same name and set it to `latest`.
+        :param outputs: list of labels for the model, Note: please provide the label column, if provided will override
+                        outputs provided in log_model
```
GiladShapira94: maybe worth add more explicit warning like - 
`list of the model outputs (e.g. labels) ,if provided will override the outputs that been configured in the model artifact, please note that those outputs need to be more equal to the model serving function outputs`

davesh0812 (approved):

PR #7301, opened 2025-02-17
[Model Monitoring] Support datastore profiles in QueueStep

alxtkr77:
https://iguazio.atlassian.net/browse/ML-9238

### DIFF
--- a/mlrun/datastore/__init__.py
+++ b/mlrun/datastore/__init__.py
@@ -34,7 +34,10 @@ __all__ = [
     "VectorStoreCollection",
 ]
 
+from urllib.parse import urlparse
+
 import fsspec
+from mergedeep import merge
 
 import mlrun.datastore.wasbfs
 from mlrun.platforms.iguazio import (
@@ -95,6 +98,38 @@ def get_in_memory_items():
     return in_memory_store._items
 
 
+def debug_info(params, logger=None, msg="KUKAREKU=="):
+    import json
+    import traceback
+
+    """
+    Print debug information including message, parameters and stack trace to logger
+
+    Args:
+        msg: Debug message to log
+        params: Dictionary of parameters to log (can be nested)
+        logger: Logger instance to use for output
+
+        dd = {
+            "bootstrap_servers":self._brokers,
+            "options":self._producer_options
+        }
+        debug_info(dd)
+
+    """
+    # Get the current stack trace
+    stack = traceback.format_stack()[:-1]  # Exclude the current frame
+
+    # Create debug info dictionary
+    debug_data = {"message": msg, "parameters": params, "stack_trace": stack}
+    # Log as formatted JSON
+    res = json.dumps(debug_data, indent=2)
+    if logger:
+        logger.error(res)
+    else:
+        print(res)
+
+
 def get_stream_pusher(stream_path: str, **kwargs):
     """get a stream pusher object from URL.
 
@@ -106,23 +141,64 @@ def get_stream_pusher(stream_path: str, **kwargs):
 
     :param stream_path:        path/url of stream
     """
+    if stream_path.startswith("ds://"):
+        from mlrun.datastore.datastore_profile import (
+            DatastoreProfileKafkaSource,
+            DatastoreProfileKafkaTarget,
+            DatastoreProfileV3io,
+        )
 
-    kafka_brokers = get_kafka_brokers_from_dict(kwargs)
-    if stream_path.startswith("kafka://") or kafka_brokers:
-        topic, brokers = parse_kafka_url(stream_path, kafka_brokers)
-        return KafkaOutputStream(topic, brokers, kwargs.get("kafka_producer_options"))
-    elif stream_path.startswith("http://") or stream_path.startswith("https://"):
-        return HTTPOutputStream(stream_path=stream_path)
-    elif "://" not in stream_path:
-        return OutputStream(stream_path, **kwargs)
-    elif stream_path.startswith("v3io"):
-        endpoint, stream_path = parse_path(stream_path)
-        endpoint = kwargs.pop("endpoint", None) or endpoint
-        return OutputStream(stream_path, endpoint=endpoint, **kwargs)
-    elif stream_path.startswith("dummy://"):
-        return _DummyStream(**kwargs)
+        datastore_profile = mlrun.datastore.datastore_profile.datastore_profile_read(
+            stream_path
+        )
+        if isinstance(
+            datastore_profile,
+            (DatastoreProfileKafkaSource, DatastoreProfileKafkaTarget),
+        ):
+            attributes = datastore_profile.attributes()
+            brokers = attributes.pop("brokers", None)
+            # Override the topic with the one in the url (if any)
+            parsed_url = urlparse(stream_path)
+            topic = (
+                parsed_url.path.strip("/")
+                if parsed_url.path
+                else datastore_profile.topic()
+            )
+            dd = {
+                "attributes": attributes,
+                "brokers": brokers,
+                "profile_topics": datastore_profile.topic(),
+                "topic": topic,
+            }
+            debug_info(dd)
+            return KafkaOutputStream(topic, brokers, producer_options=attributes)
+
+        elif isinstance(datastore_profile, DatastoreProfileV3io):
+            parsed_url = urlparse(stream_path)
+            stream_path = datastore_profile.url(parsed_url.path)
+            endpoint, stream_path = parse_path(stream_path)
+            return OutputStream(stream_path, endpoint=endpoint, **kwargs)
+        else:
+            raise ValueError(f"unsupported stream path {type(datastore_profile)}")
     else:
-        raise ValueError(f"unsupported stream path {stream_path}")
+        kafka_brokers = get_kafka_brokers_from_dict(kwargs)
+        if stream_path.startswith("kafka://") or kafka_brokers:
+            topic, brokers = parse_kafka_url(stream_path, kafka_brokers)
+            return KafkaOutputStream(
+                topic, brokers, kwargs.get("kafka_producer_options")
+            )
+        elif stream_path.startswith("http://") or stream_path.startswith("https://"):
+            return HTTPOutputStream(stream_path=stream_path)
+        elif "://" not in stream_path:
+            return OutputStream(stream_path, **kwargs)
+        elif stream_path.startswith("v3io"):
+            endpoint, stream_path = parse_path(stream_path)
+            endpoint = kwargs.pop("endpoint", None) or endpoint
+            return OutputStream(stream_path, endpoint=endpoint, **kwargs)
+        elif stream_path.startswith("dummy://"):
+            return _DummyStream(**kwargs)
+        else:
+            raise ValueError(f"unsupported stream path {stream_path}")
 
 
 class _DummyStream:
--- a/mlrun/datastore/__init__.py
+++ b/mlrun/datastore/__init__.py
@@ -130,7 +130,7 @@ def get_stream_pusher(stream_path: str, **kwargs):
             topic = (
                 parsed_url.path.strip("/")
                 if parsed_url.path
-                else datastore_profile.topic()
+                else datastore_profile.get_topic()
             )
             return KafkaOutputStream(topic, brokers, producer_options=attributes)
 
--- a/mlrun/datastore/__init__.py
+++ b/mlrun/datastore/__init__.py
@@ -40,6 +40,11 @@ import fsspec
 from mergedeep import merge
 
 import mlrun.datastore.wasbfs
+from mlrun.datastore.datastore_profile import (
+    DatastoreProfileKafkaSource,
+    DatastoreProfileKafkaTarget,
+    DatastoreProfileV3io,
+)
 from mlrun.platforms.iguazio import (
     HTTPOutputStream,
     KafkaOutputStream,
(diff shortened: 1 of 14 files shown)


mlrun/datastore/datastore_profile.py
```diff
@@ -193,6 +196,10 @@ class DatastoreProfileKafkaSource(DatastoreProfile):
     kwargs_public: typing.Optional[dict]
     kwargs_private: typing.Optional[dict]
 
+    def get_topic(self):
+        topics = [self.topics] if isinstance(self.topics, str) else self.topics
+        return topics[0] if topics else None
```
jond01: Why did you choose the first topic? It seems arbitrary and can cause issues.

Missing type hint:

```suggestion
    def get_topic(self) -> typing.Optional[str]:
        topics = [self.topics] if isinstance(self.topics, str) else self.topics
        return topics[0] if topics else None
```

But actually `topics` is guaranteed to be non-empty, so the last `if else` is redundant, isn't it?

mlrun/datastore/datastore_profile.py
```diff
@@ -171,6 +171,9 @@ def __init__(self, **kwargs):
                 FutureWarning,
             )
 
+    def get_topic(self):
+        return self.topic
```
jond01: Type hint:

```suggestion
    def get_topic(self) -> str:
        return self.topic
```

mlrun/model_monitoring/stream_processing.py
```diff
@@ -259,39 +258,13 @@ def apply_parquet_target():
...
-                    "controller_stream_kafka",
-                    path=path,
-                    kafka_brokers=brokers,
-                    _sharding_func=ControllerEvent.ENDPOINT_ID,
-                    after="ForwardNOP",
-                )
+            graph.add_step(
+                ">>",
+                "controller_stream",
+                path=stream_uri,
```
jond01, line 264: Ideally, this function should not get the obsolete `stream_uri` argument.
When looking behind the scenes, this argument is initialized by calling:
`get_stream_path`
Which doesn't cover all the functionality provided by the profiles. For example, in Kafka, it takes only the first broker, even when there is more than one.

`mlrun.model_monitoring.helpers.get_output_stream` is the newer version that utilizes the full features of the stream's profile.

mlrun/datastore/__init__.py
```diff
@@ -106,23 +109,57 @@ def get_stream_pusher(stream_path: str, **kwargs):
...
-        endpoint = kwargs.pop("endpoint", None) or endpoint
-        return OutputStream(stream_path, endpoint=endpoint, **kwargs)
-    elif stream_path.startswith("dummy://"):
-        return _DummyStream(**kwargs)
+    if stream_path.startswith("ds://"):
+        from mlrun.datastore.datastore_profile import (
+            DatastoreProfileKafkaSource,
+            DatastoreProfileKafkaTarget,
+            DatastoreProfileV3io,
+        )
```
jond01: Why are these imports not at the top of the module?

mlrun/datastore/__init__.py
```diff
@@ -106,23 +109,57 @@ def get_stream_pusher(stream_path: str, **kwargs):
...
+
+        datastore_profile = mlrun.datastore.datastore_profile.datastore_profile_read(
+            stream_path
+        )
```
jond01, line 120: If we get here, the flow is:
1. We have the profile or read it
2. We generate `stream_uri` with it through `get_stream_path`
3. We call `get_stream_pusher` with `stream_uri`
4. We read the profile again

This duplicate profile reading can be spared

mlrun/datastore/datastore_profile.py
alxtkr77-creomagic: It's not guaranteed to be non empty. User can provide an empty list or empty string to the DataStore profile
As for the first topic - this is how its used down the road. We will have to review the multiple topics when refactoring the Kafka interfaces

mlrun/datastore/datastore_profile.py
alxtkr77-creomagic: will fix

mlrun/model_monitoring/stream_processing.py
alxtkr77-creomagic: Right, but we have to pass the stream_uri as a string, since the steps in the graph have to be serializable. So we cannot pass the profile as a parameter to the graph.add_step() function

mlrun/datastore/__init__.py
alxtkr77-creomagic: Will move

mlrun/datastore/__init__.py
alxtkr77: The stream_uri must be retained to maintain serialization compatibility, as it is embedded within the generated graph structures.

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -1189,7 +1182,14 @@ def _verify_kafka_access(
...
         try:
             # The following constructor attempts to establish a connection
-            consumer = kafka.KafkaConsumer(bootstrap_servers=kafka_brokers)
+            attributes = kafka_profile.attributes()
+            kafka_admin_client_kwargs = mlrun.datastore.utils.KafkaParameters(
+                attributes
+            ).consumer()
+
+            consumer = kafka.KafkaConsumer(
+                bootstrap_servers=kafka_brokers, **kafka_admin_client_kwargs
```
jond01:
```suggestion
            kafka_consumer_kwargs = mlrun.datastore.utils.KafkaParameters(
                attributes
            ).consumer()

            consumer = kafka.KafkaConsumer(
                bootstrap_servers=kafka_brokers, **kafka_consumer_kwargs
```

mlrun/datastore/utils.py
```diff
@@ -222,3 +222,51 @@ def validate_additional_filters(additional_filters):
...
+            "admin": self._kafka.KafkaAdminClient.DEFAULT_CONFIG,
+        }
+
+    def _get_config(self, client_type: str) -> dict:
+        res = {
+            k: self._kwargs[k]
+            for k in self._kwargs.keys() & self._client_configs[client_type].keys()
+        }
+        if "sasl" in self._kwargs:
+            sasl = self._kwargs["sasl"]
```
jond01: Just a suggestion:
```suggestion
        if sasl := self._kwargs.get("sasl"):
```

mlrun/datastore/utils.py
```diff
@@ -222,3 +222,51 @@ def validate_additional_filters(additional_filters):
             for sub_value in value:
                 if isinstance(sub_value, float) and math.isnan(sub_value):
                     raise mlrun.errors.MLRunInvalidArgumentError(nan_error_message)
+
+
+class KafkaParameters:
+    def __init__(self, kwargs: dict):
+        import kafka
```
jond01, line 229: Why is that here?
Same as https://github.com/mlrun/mlrun/pull/7301#discussion_r1961798326...

Please fix it

mlrun/datastore/sources.py
```diff
@@ -1202,16 +1199,9 @@ def create_topics(
...
-                    "sasl_mechanism": sasl["mechanism"],
-                    "sasl_plain_username": sasl["user"],
-                    "sasl_plain_password": sasl["password"],
-                }
-            )
+        kafka_admin_kwargs = mlrun.datastore.utils.KafkaParameters(
+            self.attributes
+        ).admin()
 
         kafka_admin = KafkaAdminClient(bootstrap_servers=brokers, **kafka_admin_kwargs)
```
jond01, line 1210: This could break when `brokers` is changed to `bootstrap_servers` in the profile's attributes

mlrun/datastore/utils.py
```diff
@@ -222,3 +222,51 @@ def validate_additional_filters(additional_filters):
...
+    def consumer(self) -> dict:
+        return self._get_config("consumer")
+
+    def producer(self) -> dict:
+        return self._get_config("producer")
+
+    def admin(self) -> dict:
+        return self._get_config("admin")
+
+    def sasl(self, usr=None, pwd=None) -> dict:
```
jond01: I would change it to be keyword only to avoid confusion:
```suggestion
    def sasl(self, *, usr=None, pwd=None) -> dict:
```
Missing type hints - similar to https://github.com/mlrun/mlrun/pull/7301#discussion_r1961754278 but for the inputs

mlrun/datastore/utils.py
```diff
@@ -222,3 +222,51 @@ def validate_additional_filters(additional_filters):
...
+        self._client_configs = {
+            "consumer": self._kafka.KafkaConsumer.DEFAULT_CONFIG,
+            "producer": self._kafka.KafkaProducer.DEFAULT_CONFIG,
+            "admin": self._kafka.KafkaAdminClient.DEFAULT_CONFIG,
+        }
+
+    def _get_config(self, client_type: str) -> dict:
+        res = {
+            k: self._kwargs[k]
+            for k in self._kwargs.keys() & self._client_configs[client_type].keys()
```
jond01, line 266: I don't like this key intersection, as it doesn't validate anything about `self._kwargs`.
For example - a typo in the passed kwargs will silently fail here.

mlrun/datastore/storeytargets.py
```diff
@@ -126,16 +128,18 @@ def __init__(self, *args, **kwargs):
...
                 mlrun.datastore.datastore_profile.datastore_profile_read(path)
             )
             attributes = merge(attributes, datastore_profile.attributes())
-            brokers = attributes.pop(
-                "brokers", attributes.pop("bootstrap_servers", None)
+            brokers = attributes.pop("brokers", None)
+            # Override the topic with the one in the url (if any)
+            parsed = urlparse(path)
+            topic = (
+                parsed.path.strip("/") if parsed.path else datastore_profile.get_topic()
```
jond01, line 148: Validate first that the `datastore_profile` is of the expected `KafkaTarget` type

mlrun/datastore/datastore_profile.py
```diff
@@ -193,6 +196,10 @@ class DatastoreProfileKafkaSource(DatastoreProfile):
     kwargs_public: typing.Optional[dict]
     kwargs_private: typing.Optional[dict]
 
+    def get_topic(self) -> str:
```
jond01: Should be changed to:
```suggestion
    def get_topic(self) -> typing.Optional[str]:
```
According to your comment https://github.com/mlrun/mlrun/pull/7301#discussion_r1966534591

mlrun/datastore/__init__.py
```diff
@@ -106,23 +114,51 @@ def get_stream_pusher(stream_path: str, **kwargs):
...
+            )
+            return KafkaOutputStream(topic, brokers, producer_options=attributes)
+
+        elif isinstance(datastore_profile, DatastoreProfileV3io):
+            parsed_url = urlparse(stream_path)
+            stream_path = datastore_profile.url(parsed_url.path)
+            endpoint, stream_path = parse_path(stream_path)
+            return OutputStream(stream_path, endpoint=endpoint, **kwargs)
+        else:
+            raise ValueError(f"unsupported stream path {type(datastore_profile)}")
```
jond01:
```suggestion
            raise ValueError(f"Unsupported datastore profile type: {type(datastore_profile)}")
```

mlrun/datastore/__init__.py
```diff
@@ -106,23 +114,51 @@ def get_stream_pusher(stream_path: str, **kwargs):
...
+            attributes = datastore_profile.attributes()
+            brokers = attributes.pop("brokers", None)
+            # Override the topic with the one in the url (if any)
+            parsed_url = urlparse(stream_path)
+            topic = (
+                parsed_url.path.strip("/")
+                if parsed_url.path
+                else datastore_profile.get_topic()
+            )
+            return KafkaOutputStream(topic, brokers, producer_options=attributes)
```
jond01: How does it work with SASL and `DatastoreProfileKafkaSource`?
`producer_options` are passed as is, I think that the compatibility translation is missing here

mlrun/datastore/utils.py
alxtkr77: Can you clarify the guidelines? I mean sometimes kafka is imported on the module level, but sometimes you put it inside class that uses it (server/py/services/api/crud/model_monitoring/deployment.py::_verify_kafka_access(), mlrun/platforms/iguazio.py::KafkaOutputStream::_lazy_init()). Just grep for 'import kafka'

mlrun/datastore/sources.py
alxtkr77: Should not, because bootstrap_servers should not be accessed directly, only using DatastoreProfileKafkaSource.attributes() interface

mlrun/datastore/utils.py
jond01: Sorry, I was wrong here.
`kafka` is not installed by default on the client side, so your inner import is actually a must.
In the example you gave it can be outside of the function, as `mlrun-api` installs `kafka`:
https://github.com/mlrun/mlrun/blob/2a960d0b8b731ddceeb42922b74b8a13759f618e/extras-requirements.txt#L37

mlrun/datastore/utils.py
alxtkr77: Hmm, good point. Let me clarify the attributes dictionary structure:

The attributes dictionary contains three categories:
1. Internal attributes used by KafkaOutputStream (e.g., "brokers", "topics", "sasl")
2. Common parameters passed to all Kafka clients (KafkaConsumer, KafkaProducer, KafkaAdminClient) such as 'ssl_check_hostname'
3. Client-specific parameters like 'metrics_sample_window_ms' which only apply to certain clients (e.g., KafkaConsumer)

All these different attribute types are currently maintained in a single dictionary. I'll implement a verification mechanism that validates the attribute list against the complete set of possible keys from all these categories.

mlrun/datastore/__init__.py
alxtkr77: good catch!

jond01 (approved):
LGTM, great work

PR #7302, opened 2025-02-17
[Python] Remove distutils references

liranbg:
distutils is deprecated since python 3.12
we currently get it by having setuptools but needlessly as this is a single functionality that can be part of mlrun utls
https://iguazio.atlassian.net/browse/ML-5438

### DIFF
--- a/tests/utils/test_helpers.py
+++ b/tests/utils/test_helpers.py
@@ -989,6 +989,40 @@ def test_get_pretty_types_names():
         assert pretty_result == expected
 
 
+@pytest.mark.parametrize(
+    "value, expected, exception",
+    [
+        # True values
+        ("y", True, does_not_raise()),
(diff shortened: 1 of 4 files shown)


tests/utils/test_helpers.py
```diff
@@ -989,6 +989,40 @@ def test_get_pretty_types_names():
         assert pretty_result == expected
 
 
+@pytest.mark.parametrize(
+    "value, expected, exception",
```
jond01, line 993: I like more the name `expectation` - as it's not necessarily an exception:
```suggestion
    "value, expected, expectation",
```

jond01 (approved):

tests/utils/test_helpers.py
liranbg: Understand the comment but No exception is still an exception :)

PR #7305, opened 2025-02-17
[Model Monitoring] Fix controller v3io stream committed offset

royischoss:
https://iguazio.atlassian.net/browse/ML-9344
Change `explicit_ack_mode` for v3io to react to engine type "sync"/"async" (as kafka source) when for "sync" engine we use nuclio default mode "disable" as for mm controller


server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -419,9 +419,15 @@ def _apply_and_create_v3io_source(
             != mm_constants.MonitoringFunctionNames.APPLICATION_CONTROLLER
             else mlrun.mlconf.get_v3io_access_key()
         )
+        explicit_ack_mode = (
+            "explicitOnly"
+            if function_name
+            != mm_constants.MonitoringFunctionNames.APPLICATION_CONTROLLER
+            else "disable"
+        )
```
gtopper:
```suggestion
```
It's unnecessary to set it to disable, and in fact we don't do it when `mlrun.mlconf.is_explicit_ack_enabled()` returns false. Would be enough to just change the condition:
```python
        if (
            mlrun.mlconf.is_explicit_ack_enabled()
            and function_name
            != mm_constants.MonitoringFunctionNames.APPLICATION_CONTROLLER
        ):
```
Disabling in two different ways is misleading.

gtopper (commented):
Very nice otherwise :)

mlrun/datastore/sources.py
```diff
@@ -966,6 +967,26 @@ def add_nuclio_trigger(self, function):
             "This source type is not supported with ingestion service yet"
         )
 
+    @staticmethod
+    def set_explicit_ack_mode(function: Function, **extra_arguments):
```
davesh0812:
```suggestion
    def set_explicit_ack_mode(function: Function, **extra_arguments) -> dict[str, Any]:
```

mlrun/datastore/sources.py
```diff
@@ -1028,15 +1049,8 @@ def add_nuclio_trigger(self, function):
...
-
         kwargs = {}
-        engine = "async"
-        if hasattr(function.spec, "graph") and function.spec.graph.engine:
-            engine = function.spec.graph.engine
-
-        if mlrun.mlconf.is_explicit_ack_enabled() and engine == "async":
-            kwargs["explicit_ack_mode"] = "explicitOnly"
-            kwargs["worker_allocation_mode"] = "static"
+        kwargs = self.set_explicit_ack_mode(function, **kwargs)
```
davesh0812: not sure why you need to send `kwargs` to the `set_explicit_ack_mode()`, its always an empty dict before

mlrun/datastore/sources.py
```diff
@@ -1118,20 +1132,11 @@ def add_nuclio_trigger(self, function):
...
-                "worker_allocation_mode", "static"
-            )
-        else:
-            extra_attributes["workerAllocationMode"] = extra_attributes.get(
-                "worker_allocation_mode", "pool"
-            )
+        extra_attributes = self.set_explicit_ack_mode(function, **extra_attributes)
+        explicit_ack_mode = extra_attributes.get("explicit_ack_mode")
+        if extra_attributes.get("workerAllocationMode") is None:
+            extra_attributes["workerAllocationMode"] = "pool"
```
davesh0812:
```suggestion
     extra_attributes["workerAllocationMode"] = extra_attributes.get("worker_allocation_mode", "pool")
```

davesh0812 (commented):
Nice :)
Few comments

davesh0812 (approved):

PR #7310, opened 2025-02-18
[Alerts] Add api to delete all alert configs and clean up cache

rokatyy:
Adds api to delete all alert configs and clean up cache, optimises `delete_project_alert_events`, fixes cache cleaning 

Jira - https://iguazio.atlassian.net/browse/ML-9367


server/py/framework/db/sqldb/db.py
```diff
@@ -5995,6 +5994,39 @@ def list_alerts(
             alerts.append(alert)
         return alerts
 
+    def list_and_delete_project_alerts(
```
TomerShor: Why not `delete_alerts`, and support `project = "*"`?
it's ok that a **delete** function will return the ids of the deleted resources.

server/py/framework/db/sqldb/db.py
```diff
@@ -5995,6 +5994,39 @@ def list_alerts(
...
+        for alert_id, alert_name in alerts:
+            self._delete_alert_notifications(
+                session=session,
+                name=alert_name,
+                alert_id=alert_id,
+                project=project,
+                commit=False,
+            )
+            # delete the alert config
+            self._delete(session, AlertConfig, id=alert_id)
```
liranbg: we need to support batch deletion, this is very excessive

server/py/framework/db/sqldb/db.py
```diff
@@ -5995,6 +5994,39 @@ def list_alerts(
...
+        project: str,
+    ) -> list[str]:
+        # We need to list ids and names here
+        # ids to reuse later for cleaning up the cache
+        # names to make notification deletion more efficient as notification table has index on name and parent_id
+        alerts = (
+            session.query(AlertConfig.id, AlertConfig.name)
+            .filter(AlertConfig.project == project)
+            .all()
+        )
```
liranbg: better of working in chunks and not loading everything at once (using limit + offset)

server/py/framework/utils/clients/chief.py
```diff
@@ -247,6 +247,16 @@ async def delete_alert(
             "DELETE", f"projects/{project}/alerts/{name}", request
         )
 
+    async def delete_alerts(
+        self, project: str, request: fastapi.Request
+    ) -> fastapi.Response:
+        """
+        Alerts are running only on chief
```
TomerShor, line 254: 👀 ?

server/py/services/api/crud/projects.py
```diff
@@ -211,6 +211,8 @@ def delete_project_resources(
         # actually forwarding the request and waiting for a response, since the project deletion flow is handled
         # by Chief only.
         services.alerts.crud.Events().delete_project_alert_events(name)
+        # only for the case when alerts service is running together with api
```
TomerShor: that's the TODO for the followup PR

TomerShor (approved):

TomerShor (changes requested):
approved by mistake

server/py/framework/db/sqldb/db.py
alonmr: perhaps use `_delete_multi_objects`

server/py/framework/utils/clients/chief.py
rokatyy: All methods here have this comment - it means that if alerts are running with api, they run only on chief

server/py/framework/db/sqldb/db.py
rokatyy: @alonmr In this case it won't help, especially that it's better to request data by indexes instead of just project

server/py/framework/db/sqldb/db.py
rokatyy: @TomerShor is this really a case that we remove everything from cluster? Or you mean to have it just in case?

server/py/framework/db/sqldb/db.py
```diff
@@ -5995,6 +5994,77 @@ def list_alerts(
...
+
+            # Step 4: Perform ORM-based deletion for alerts in the current chunk
+            alerts_to_delete = (
+                session.query(AlertConfig)
+                .filter(AlertConfig.id.in_(alert_ids_chunk))
+                .all()
+            )
+            for alert in alerts_to_delete:
+                # Deleting via ORM ensures cascading works
+                session.delete(alert)
```
alonmr: The for loop creates N delete statements to be executed once you commit. You may do it with a single statement instead.
```suggestion
     session.query(AlertConfig)
                .filter(AlertConfig.id.in_(alert_ids_chunk))
                .delete()
```

server/py/framework/db/sqldb/db.py
TomerShor: I meant having the option to pass `project="*"` if we want that in the future, not for our scenario.

And to rename the function to `delete_alerts`, because that's the actual purpose of this function. `list` is just another thing that happens here.

server/py/framework/db/sqldb/db.py
```diff
@@ -5995,6 +5994,77 @@ def list_alerts(
...
+
+            # Step 3: Extract alert IDs for deletion and delete alerts in this chunk
+            alert_ids_chunk = [alert_id for alert_id, _ in alerts]
+            alert_ids.extend(alert_ids_chunk)  # Collect all alert IDs to be deleted
+
+            # Step 4: Perform ORM-based deletion for alerts in the current chunk
+            alerts_to_delete = (
+                session.query(AlertConfig)
+                .filter(AlertConfig.id.in_(alert_ids_chunk))
+                .all()
```
TomerShor, line 6201: when you run `.all()`, isn't the session flushed / committed? causing the notifications to be deleted.

server/py/framework/db/sqldb/db.py
```diff
@@ -5995,6 +5994,77 @@ def list_alerts(
             alerts.append(alert)
         return alerts
 
+    def list_and_delete_project_alerts(
```
alonmr: I would just call it `delete_project_alerts` because you don't really list the alerts but just their IDs which is a legitimate "delete" op response

server/py/services/alerts/tests/unit/api/test_alerts.py
```diff
@@ -186,6 +186,18 @@ def test_get_alert_activations(
         result = client.get(f"projects/{project_name}/alerts/wrong-name/activations/1")
         assert result.status_code == 404
 
+    @unittest.mock.patch.object(services.alerts.crud.Alerts, "delete_alerts")
+    def test_delete_alerts(
```
TomerShor, line 190: well, this tests idempotency and not actually deleting the alerts.

server/py/framework/db/sqldb/db.py
```diff
@@ -5995,6 +5994,77 @@ def list_alerts(
...
+            # Notifications reference alerts via foreign keys (parent_id and name).
+            # Deleting them first prevents integrity errors.
+            for alert_id, alert_name in alerts:
+                self._delete_alert_notifications(
+                    session=session,
+                    name=alert_name,  # Index on name improves deletion efficiency
+                    alert_id=alert_id,
+                    project=project,
+                    commit=False,  # Defer commit for batch efficiency
+                )
```
TomerShor: Put comments above code and not inline, to align with our conventions

TomerShor (changes requested):
Looks great overall, minor comments

server/py/services/alerts/tests/unit/api/test_alerts.py
rokatyy: @TomerShor that's the point - just to test endpoints. As all the other tests in this suite. Actual deletion is checked below

alonmr (commented):
Looks great! Minor suggestions

server/py/framework/db/sqldb/db.py
rokatyy: @TomerShor you're right here! I'll go with the idea that Alon suggested below, which eliminates the need to list them

TomerShor (approved):

server/py/framework/db/sqldb/db.py
rokatyy: @alonmr actually it doesn't work as expected:

```>       cursor.execute(statement, parameters)
E       sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) FOREIGN KEY constraint failed
E       [SQL: DELETE FROM alert_configs WHERE alert_configs.id IN (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)]
E       [parameters: (1, 2, 3, 4, 5, 6, 7, 8, 9, 10)]
```

rokatyy: @alonmr @TomerShor I'm returning the old version of the code. Will glad to hear better suggestions :)

server/py/framework/db/sqldb/db.py
rokatyy: @TomerShor I'd add bare-minimum that is required for now as we are jumping in the last train in terms of release

server/py/framework/db/sqldb/db.py
TomerShor: What foreign key constraint fails? the notifications, alert states?

rokatyy: @TomerShor notifications

server/py/framework/db/base.py
```diff
@@ -904,6 +904,13 @@ def list_alerts(
     ) -> list[mlrun.common.schemas.AlertConfig]:
         pass
 
+    def delete_project_alerts(
```
moranbental, line 910: It should be an abstract method, like all the other functions

server/py/framework/db/sqldb/db.py
```diff
@@ -5995,6 +5994,78 @@ def list_alerts(
             alerts.append(alert)
         return alerts
 
+    def delete_project_alerts(
+        self, session, project: str, chunk_size: int = 100
```
moranbental: It’s worth adding `chunk_size` to the config.py

server/py/framework/db/sqldb/db.py
```diff
@@ -6342,25 +6413,16 @@ def create_alert_state(self, session, alert_id):
         state = AlertState(count=0, parent_id=alert_id)
         self._upsert(session, [state])
 
-    def delete_alert_notifications(
+    def _delete_alert_notifications(
         self,
         session,
+        name: typing.Optional[str],
```
```suggestion
        name: typing.Optional[str] = None,
```

server/py/framework/db/sqldb/db.py
```diff
@@ -5995,6 +5994,78 @@ def list_alerts(
             alerts.append(alert)
         return alerts
 
+    def delete_project_alerts(
```
moranbental, line 6144:
```suggestion
    def delete_alerts_and_notifications(
```
Because you are also deleting the notifications

server/py/framework/db/sqldb/db.py
```diff
@@ -5995,6 +5994,78 @@ def list_alerts(
...
+        self, session, project: str, chunk_size: int = 100
+    ) -> list[str]:
```
```suggestion
    ) -> typing.List[int]:
```

moranbental (changes requested):
Looks very good! minor..

server/py/framework/db/sqldb/db.py
rokatyy: @moranbental technically, I delete everything related to alerts here - configs, notifications. So, I will leave it as is

server/py/framework/db/sqldb/db.py
rokatyy: @moranbental everywhere in the code we use list.

moranbental: @rokatyy it was more for int instead of str

server/py/framework/db/base.py
moranbental: You missed that. still not abstract method

server/py/framework/db/sqldb/db.py
```diff
@@ -6137,6 +6141,81 @@ def list_alerts(
             alerts.append(alert)
         return alerts
 
+    def delete_project_alerts(
+        self,
+        session,
+        project: str,
+        chunk_size: int = mlrun.mlconf.alerts.chunk_size_during_project_deletion,
```
moranbental: We are not using the config in the function declaration because the value can change dynamically at runtime. use
`chunk_size = chunk_size or mlrun.mlconf.alerts.chunk_size_during_project_deletion`

moranbental (changes requested):

server/py/framework/db/sqldb/db.py
TomerShor: Why do we need to delete the notifications manually?
The alert config table has a `cascade` on the notification relationship, so once a config is deleted - all its notifications should be deleted too.

rokatyy: @TomerShor good question. It was in the old code, I just moved it around. Not sure what was the idea behind. Will remove it

server/py/framework/db/sqldb/db.py
```diff
@@ -6137,6 +6141,72 @@ def list_alerts(
...
+        :return: List of deleted alert IDs.
+        """
+        chunk_size = (
+            chunk_size or mlrun.mlconf.alerts.chunk_size_during_project_deletion
+        )
+        alert_ids = []
+        offset = 0
+
+        while True:
+            # Step 1: Retrieve alerts (IDs and names) for the given project in chunks
```
```suggestion
            # Step 1: Retrieve alerts (IDs) for the given project in chunks
```

moranbental (approved):

server/py/framework/db/sqldb/db.py
TomerShor: Will deleting with `AlertConfig.id in alert_ids` still fail now when you don't delete the notifications?

server/py/framework/db/sqldb/db.py
```diff
@@ -6137,6 +6141,72 @@ def list_alerts(
...
+        project: str,
+        chunk_size: typing.Optional[int] = None,
+    ) -> list[int]:
+        """
+        List all alert IDs associated with the specified project and delete them,
+        along with their related notifications, while ensuring foreign key constraints are respected.
+
+        Steps:
+        1. Retrieve all alerts (IDs and names) for the given project.
+        2. Delete related notifications first (since they have foreign key constraints).
```
TomerShor: not anymore

server/py/services/alerts/tests/unit/crud/test_alerts.py
```diff
@@ -483,3 +483,62 @@ async def test_get_alerts_exclude_updated(
...
+        assert len(alerts_after_deletion) == 0
+        for alert in alerts:
+            assert (
+                services.alerts.crud.Alerts()._get_alert_by_id_cached()(db, alert.id)
+                is None
+            )
+            with pytest.raises(mlrun.errors.MLRunNotFoundError):
+                assert services.alerts.crud.Alerts()._get_alert_state_cached()(
+                    db, alert.id
+                )
```
TomerShor, line 546: Also verify no notifications

TomerShor (changes requested):

server/py/framework/db/sqldb/db.py
rokatyy: @TomerShor yes

server/py/services/alerts/tests/unit/crud/test_alerts.py
rokatyy: @TomerShor we don't have method for that

server/py/framework/db/sqldb/db.py
alonmr: The problem is you don't have DB level cascade - see https://github.com/mlrun/mlrun/pull/6598
If you create a query to delete the notifications you can then delete the alerts with 2 queries instead of N but there is a race condition here so IMO leave it as is and add a TODO to fix the foreign key constraint

server/py/services/api/crud/projects.py
```diff
@@ -211,6 +211,8 @@ def delete_project_resources(
         # actually forwarding the request and waiting for a response, since the project deletion flow is handled
         # by Chief only.
         services.alerts.crud.Events().delete_project_alert_events(name)
```
alonmr, line 213: this line should be removed right?

server/py/services/api/tests/unit/api/test_projects.py
```diff
@@ -289,6 +309,34 @@ def _list_configmaps(*args, **kwargs):
...
+        assert (
+            services.alerts.crud.Alerts()._get_alert_state_cached()(db, alert_id)
+            is not None
+        )
+
+    # check that event cache is removed for project_to_remove, and it isn't for project_to_keep
+    found_project_to_keep = False
+    for key in services.alerts.crud.events.Events()._cache.keys():
+        assert key[0] != project_to_remove
+        found_project_to_keep |= key[0] == project_to_keep
```
alonmr: if you know how many events you have for this project you can keep a counter and be more accurate

alonmr (commented):
Very well! A few suggestions/comments but overall lgtm

server/py/services/api/tests/unit/api/test_projects.py
rokatyy: @alonmr which counters? this is just objects cache..

alonmr: I mean do something like
```
    project_to_keep_cached_events_count = 0
    for key in services.alerts.crud.events.Events()._cache.keys():
        assert key[0] != project_to_remove
        if key[0] == project_to_keep:
           project_to_keep_cached_events_count += 1

   assert project_to_keep_cached_events_count == expected_project_to_keep_cached_events_count
```

server/py/services/api/crud/projects.py
rokatyy: @alonmr no, it should remain - that's another cache and it's needed if alerts run together with api

server/py/services/api/tests/unit/api/test_projects.py
rokatyy: @alonmr done

server/py/services/alerts/crud/alerts.py
```diff
@@ -178,9 +178,27 @@ def delete_alert(
...
+        project = project or mlrun.mlconf.default_project
+        services.alerts.crud.Events().delete_project_alert_events(project)
+
+        alert_ids = framework.utils.singletons.db.get_db().delete_project_alerts(
+            session, project
+        )
+        if not alert_ids:
+            return
+
+        for alert_id in alert_ids:
```
liranbg: worth adding log for before operations (and after all). include information for debugging but make sure it is not spammy.
```suggestion
        project = project or mlrun.mlconf.default_project
        logger.debug("deleting project alert events", project=project, alert_name=name)
        services.alerts.crud.Events().delete_project_alert_events(project)
        logger.debug("deleting project alert", project=project)
        alert_ids = framework.utils.singletons.db.get_db().delete_project_alerts(
            session, project
        )
        if not alert_ids:
            return
        logger.debug("deleting project alert ids", alert_ids_length=len(alert_ids))
        for alert_id in alert_ids:
          ...
        logger.debug("successfully deleted project alert", project=project, alert_name=name)
        
```

liranbg (changes requested):
looking much better. last comment wrt logging experience

TomerShor (approved):
Other than Liran's suggestion - LGTM

server/py/framework/db/sqldb/db.py
```diff
@@ -6137,6 +6141,72 @@ def list_alerts(
...
+            for alert in alerts_to_delete:
+                # Deleting via ORM ensures cascading works
+                # TODO: fix the foreign key constraint by implementing db level cascade
+                session.delete(alert)
+
+            # Step 3: Commit all changes in one transaction for the current chunk
+            session.commit()
+
+            # Increment the offset to process the next chunk
+            offset += chunk_size
```
Yacouby: using the offset here could become inefficient for large datasets, so maybe a better approach here would be to use `order_by(AlertConfig.id)` and track the last processed id instead of relying on the offset.

Yacouby (commented):
LGTM, minor suggestion

Yacouby (approved):

liranbg (approved):

PR #7313, opened 2025-02-18
[Messaging] Sync inter-service communication

alonmr:
Extend the proxy messaging client to enable sync inter services communication.
This facilitates sending requests between internal services and in particular needed for the API to send a delete project alerts request when a project deletion job is in progress.
We aim for this implementation to be as simple as possible at this stage and it will be extended in future versions.

Blocked by https://github.com/mlrun/mlrun/pull/7310

https://iguazio.atlassian.net/browse/ML-9368


server/py/framework/utils/clients/messaging.py
```diff
@@ -135,6 +164,46 @@ async def send_request(
...
+        url: str,
+        headers: dict,
+        raise_on_failure: bool = True,
+        **kwargs,
+    ) -> requests.Response:
+        self._prepare_request_kwargs(headers=headers, kwargs=kwargs)
+        self._ensure_sync_retry_session()
+        kwargs_to_log = self._resolve_kwargs_to_log(kwargs)
+        logger.debug(
+            "Sending request to service",
```
TomerShor: Not critical, but do you want to differentiate this log in `sync` and `async` flows, for easier debugging?

server/py/framework/utils/clients/messaging.py
```diff
@@ -312,3 +411,10 @@ def _prepare_request_data(self, method: str, path: str):
         path = path.removeprefix(f"{prefix}/").removeprefix(f"{version}/")
         service_instance = self._discovery.resolve_service_by_request(method, path)
         return path, version, service_instance
+
+    @staticmethod
+    def _resolve_full_request_path(path, service_instance, version):
+        # The service and version prefixes have been removed from the path earlier in the process.
```
TomerShor, line 532: This happens in `_prepare_request_data`, which is currently only called in `proxy_request_to_service`. 
Will the new `delete` method also be called via `proxy_request_to_service`? I don't think so.

TomerShor (changes requested):

server/py/framework/utils/auth/verifier.py
```diff
@@ -245,6 +245,20 @@ async def authenticate_request(
...
+            "authorization",
+            "cookie",
+            mlrun.common.schemas.HeaderNames.projects_role,
+            "x-data-session-override",
+            "x-remote-user",
+            "X-V3io-session-key",
+            "X-V3io-Access-Key",
+        ]:
+            if header_value := request.headers.get(header):
+                auth_info.request_headers[header] = header_value
```
liranbg: why not taking them all and drop specifics (like host, etc)?

server/py/framework/utils/clients/messaging.py
```diff
@@ -217,6 +357,26 @@ async def _on_request_failure(
...
+        response_body: dict,
+        raise_on_failure: bool,
+        kwargs,
+    ):
+        log_kwargs = copy.deepcopy(kwargs)
+
+        # this can be big and spammy
+        log_kwargs.pop("json", None)
+        log_kwargs.update(
+            {"method": method, "url": url, "response_body": response_body}
```
liranbg: suggest to omit the response body

liranbg (changes requested):

server/py/framework/utils/clients/messaging.py
alonmr: No, `_prepare_request_data` is not used in sync communication because it assumes the path contains the service name and api version but I wanted to keep the sync client agnostic

server/py/framework/utils/auth/verifier.py
alonmr: Because I don't really know what to remove. I need to remove host, content-length, content-type and others? So if i miss something it might be destructive. I chose to go on the safe side but I agree there are more headers that we should keep. I'm ok with going either way I just think this is safer at this point in time.

server/py/framework/utils/clients/messaging.py
alonmr: Changed to log just the error details

server/py/framework/utils/clients/messaging.py
```diff
@@ -40,8 +41,106 @@ def __init__(self) -> None:
...
+    def send_sync_request(
+        self,
+        method: str,
+        path: str,
+        version: str = "v1",
+        headers: typing.Optional[dict] = None,
+        raise_on_failure: bool = True,
+        **kwargs,
+    ) -> requests.Response:
+        path = path.removeprefix("/")
```
moranbental, line 134: Why are you not removing here the prefix and version?

server/py/framework/utils/clients/messaging.py
```diff
@@ -179,10 +316,14 @@ async def _ensure_session(self):
             # by returning `True`, we tell the client the response is "legit" and so, it returns it to its callee.
             self._session.retry_options.evaluate_response_callback = lambda _: True
 
-    async def _ensure_retry_session(self):
+    def _ensure_retry_session(self):
```
moranbental, line 322: We are not really using this method. Maybe we can delete it along with the `_retry_session` object.

server/py/framework/utils/auth/verifier.py
liranbg: you need to remove `content-length` and set `x-forwarded-host` from the original `host`. you may see some of the functionality already implemented @ `_resolve_request_kwargs_from_request`

server/py/framework/utils/clients/messaging.py
alonmr: prefix and version are not expected to be part of the path here. version is a separate parameter and prefix is resolved by the discovery client

server/py/framework/utils/clients/messaging.py
alonmr: indeed this is hachana lemazgan I prefer to keep it just to outline how I think it should look like and to differentiate between sync / async / proxy sessions

moranbental (approved):
LGTM

liranbg (approved):

server/py/framework/utils/clients/messaging.py
```diff
@@ -40,8 +41,106 @@ def __init__(self) -> None:
         self._session: typing.Optional[mlrun.utils.AsyncClientWithRetry] = None
         # Retry session is for internal messaging
         self._retry_session: typing.Optional[mlrun.utils.AsyncClientWithRetry] = None
```
rokatyy: not really related, but it seems we don't use `_retry_session` anywhere.

Also, general question related to the bug from this pr https://github.com/mlrun/mlrun/pull/7324. Should `_sync_retry_session` be per thread as well?

server/py/services/api/crud/projects.py
```diff
@@ -203,16 +204,18 @@ def delete_project_resources(
...
-        # by Chief only.
+        # TODO: This should be refactored once we have a proper hydra implementation
+        # Delete alert's service resources
         services.alerts.crud.Events().delete_project_alert_events(name)
-        # TODO: this should run only for the case when alerts service is running together with api
-        services.alerts.crud.Alerts().delete_alerts(session=session, project=name)
+        if mlrun.mlconf.services.hydra.services == "*":
+            services.alerts.crud.Alerts().delete_alerts(session=session, project=name)
+        else:
+            messaging_client = framework.utils.clients.messaging.Client()
```
rokatyy: add comment of what this condition means

server/py/framework/utils/clients/messaging.py
```diff
@@ -312,3 +476,10 @@ def _prepare_request_data(self, method: str, path: str):
         path = path.removeprefix(f"{prefix}/").removeprefix(f"{version}/")
         service_instance = self._discovery.resolve_service_by_request(method, path)
         return path, version, service_instance
+
+    @staticmethod
+    def _resolve_full_request_path(path, service_instance, version):
+        # The service and version prefixes have been removed from the path earlier in the process.
+        # The service prefix will be replaced with the new service name, and the version will be re-added
+        # (or default to v1 if not present) during the final URL construction for the request.
+        return f"{service_instance.url}/{service_instance.name}/{version}/{path}"
```
rokatyy: how about using rstrip as we do here? https://github.com/mlrun/mlrun/blob/1b288a97709161603f2586ada547ac1db346ede4/mlrun/common/schemas/api_gateway.py#L106-L109

server/py/services/api/crud/projects.py
```diff
@@ -203,16 +204,18 @@ def delete_project_resources(
...
-        # TODO: Forward to alerts service
-        # The messaging client is async, and project deletion is sync.
-        # When deleting a project, we need to use a sync client to send the delete event request to the alerts service,
-        # or to Chief if in Hydra mode. (ML-8390)
-        # Until we implement the sync client, we can allow Chief to delete the project alerts itself, instead of
-        # actually forwarding the request and waiting for a response, since the project deletion flow is handled
-        # by Chief only.
+        # TODO: This should be refactored once we have a proper hydra implementation
+        # Delete alert's service resources
         services.alerts.crud.Events().delete_project_alert_events(name)
```
rokatyy: this can also be part of condition

server/py/framework/utils/clients/messaging.py
```diff
@@ -204,15 +345,32 @@ async def _on_request_failure(
...
+        raise_on_failure: bool,
+        kwargs,
+    ):
+        log_kwargs = copy.deepcopy(kwargs)
+
+        # this can be big and spammy
+        log_kwargs.pop("json", None)
+        log_kwargs.update(
+            {"method": method, "url": url, "error_details": error_details}
+        )
```
rokatyy: The difference here is only that async method requires `response_body = await response.json()` , 
suggestion - maybe most of the code can be moved to a common function?

server/py/framework/utils/clients/messaging.py
alonmr: wrt `_retry_session` @moranbental commented about it as well above and I explained why I added it there.
Regarding `_sync_retry_session` I believe so. I think if your fix works we should change the sync session as well

rokatyy: add comment that we don't use it and what is the future purpose of it

server/py/services/api/crud/projects.py
alonmr: why is this needed? It's done in `services.alerts.crud.Alerts().delete_alerts`
https://github.com/mlrun/mlrun/blob/development/server/py/services/alerts/crud/alerts.py#L191

rokatyy: @alonmr you're right, my bad, can be removed then :)

liranbg (approved):

rokatyy (approved):

PR #7318, opened 2025-02-19
[Model Monitoring] Remove deprecated constants and endpoints

jond01:
Implements [ML-8851](https://iguazio.atlassian.net/browse/ML-8851) and removes unused MM constants.

[ML-8851]: https://iguazio.atlassian.net/browse/ML-8851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/api/schemas/__init__.py
```diff
@@ -208,9 +208,6 @@ def __getattr__(self, attr):
 ModelEndpointMetadata = DeprecationHelper(mlrun.common.schemas.ModelEndpointMetadata)
 ModelEndpointSpec = DeprecationHelper(mlrun.common.schemas.ModelEndpointSpec)
 ModelEndpointStatus = DeprecationHelper(mlrun.common.schemas.ModelEndpointStatus)
-ModelMonitoringStoreKinds = DeprecationHelper(
```
alonmr, line 211: This may break upgrading to 1.8 from old versions. See https://github.com/mlrun/mlrun/pull/4798
I doubt we have anyone using mlrun 1.4 but this needs to be communicated / tested

alonmr (commented):
I have a few concerns about removing some of this.
For the constants - they were not deprecated correctly, it may break user code even if they were just imported and not used.
For the endpoints - although they were deprecated correctly, the deprecation is missing from the changelog where we aim to list all deprecations and removals for transparency with our users - https://docs.mlrun.org/en/latest/change-log/index.html#deprecations-and-removed-code
Please hold with these changes while we communicate these issues.

jond01:
> For the constants - they were not deprecated correctly, it may break user code even if they were just imported and not used.
> For the endpoints - although they were deprecated correctly, the deprecation is missing from the changelog where we aim to list all deprecations and removals for transparency with our users - https://docs.mlrun.org/en/latest/change-log/index.html#deprecations-and-removed-code

We do not aim for BC in MM between 1.7.0 and 1.8.0. This change is quite lightweight compared to previous changes introduced in 1.8.0.

alonmr:
> > For the constants - they were not deprecated correctly, it may break user code even if they were just imported and not used.
> > For the endpoints - although they were deprecated correctly, the deprecation is missing from the changelog where we aim to list all deprecations and removals for transparency with our users - https://docs.mlrun.org/en/latest/change-log/index.html#deprecations-and-removed-code
> 
> We do not aim for BC in MM between 1.7.0 and 1.8.0. This change is quite lightweight compared to previous changes introduced in 1.8.0.

I see, regarding the endpoints I checked and we intentionally don't document endpoints deprecation so that's ok to remove. Regarding the schemas changes we would like to keep track on things we remove between versions without warning - if there is anything you think should be communicated please add it at the last table on https://iguazio.atlassian.net/browse/ML-9365 ("Removed without warning"). If you think this is part of a bigger backwards incompatibility then I think we should add to the ticket that MM is no BC entirely to make sure it's documented.

mlrun/api/schemas/__init__.py
jond01: So yes, it seems irrelevant - we don't assume 1.4 --> 1.8 upgrade. Do I miss something?

assaf758: indeed, the model-monitoring for <1.8 is "tech-preview" and with no BC towards 1.8 (this was confirmed by product/customer-success from 1.8 get-go)

assaf758 (approved):

PR #7325, opened 2025-02-20
[Functions] Change tag computation when the function is unversioned

davesh0812:
In order to get the tag and the uid correctly when the retrieved function is unversioned , relevant to model endpoint.

related to https://iguazio.atlassian.net/browse/ML-9394


alonmr (approved):
Generally looks good. Though IIUC we have a similar logic (versioned objects) for feature sets and feature vectors. This fix might be needed there as well.

PR #7328, opened 2025-02-22
[SDK] Ensure list_artifacts work when `limit` is set

liranbg:
https://iguazio.atlassian.net/browse/ML-9412

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -5346,7 +5346,7 @@ class HTTPRunDB(RunDBInterface):
             )
         return None
 
-    def _resolve_page_params(self, params) -> dict:
+    def _resolve_page_params(self, params: typing.Optional[dict]) -> dict:
         """
         Resolve the page parameters, setting defaults where necessary.
         """
@@ -5362,6 +5362,15 @@ class HTTPRunDB(RunDBInterface):
                 # limit and page/page size are conflicting
                 page_params.pop("limit")
             page_params["page-size"] = page_size
+
+        # this may happen only when page-size was explicitly set along with limit
+        # this is to ensure we will not get stopped by API on similar below validation
+        # but rather simply fallback to use page-size.
+        if "page-size" in page_params and "limit" in page_params:
+            logger.warning(
+                "Both 'limit' and 'page-size' are provided, using 'page-size'."
+            )
+            page_params.pop("limit")
         return page_params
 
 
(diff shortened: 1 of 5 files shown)


mlrun/db/httpdb.py
```diff
@@ -5353,6 +5346,33 @@ def _wait_for_background_task_from_response(self, response):
...
+        Resolve the page parameters, setting defaults where necessary.
+        """
+        page_params = deepcopy(params) or {}
+        if page_params.get("page-token") is None and page_params.get("page") is None:
+            page_params["page"] = 1
+        if page_params.get("page-size") is None:
+            page_size = config.httpdb.pagination.default_page_size
+
+            if page_params.get("limit") is not None:
+                page_size = params["limit"]
```
TomerShor: why use `params` again?
```suggestion
            if limit := page_params.get("limit"):
                page_size = limit
```

TomerShor (commented):
Minor

mlrun/db/httpdb.py
liranbg: your suggestion wont catch limit=0

TomerShor: Right. 
But off-topic - limit=0 should be a valid input at all 👀

liranbg: not the client to decide :)

mlrun/db/httpdb.py
```diff
@@ -5353,6 +5346,33 @@ def _wait_for_background_task_from_response(self, response):
...
+            page_params["page-size"] = page_size
+
+        # this may happen only when page-size was explicitly set along with limit
+        # this is to ensure we will not get stopped by API on similar below validation
+        # but rather simply fallback to use page-size.
+        if page_params.get("page-size") and page_params.get("limit"):
+            logger.warning(
+                "Both 'limit' and 'page-size' are provided, using 'page-size'."
+            )
+            page_params.pop("limit")
```
alonmr, line 5373: IIUC `page-size` param is opaque to the users. So I think this warning is not informative for them and I would even take `limit` over `page-size` since that is what the user gave.

alonmr: Though I also think that letting the users change the page size to whatever they want is risky and there should be a limit on that

liranbg: page size is possible via sdk when you want to paginate over less/more than default.
overall, limit is more of "sql" language (and less what user should provide) - but we have that as part of the existing sdk.
if user provides both -> we take the page size as it means the user used the paginated sdk (meaning, on purpose) and provided limit as well (which is ambigous). to the point - I think what we want to educate the clients here are using page / size and not limits / offsets.

TomerShor (approved):
LGTM

alonmr (approved):

PR #7330, opened 2025-02-23
[Artifacts] Ensure order on list_artifacts

liranbg:
The list artifacts is compiled of 2 queries `SELECTS`.
- The `inner` one - which filters on the artifact table and returns only artifact id
- The `outer` one - which joins on the inner and returns the entire artifact row(s)

note: The reason we have both inner/outer is to optimize the amount of "data" being loading during the list_artifacts so we can scan the artifact table while keep memory footprint low

the bug is that the `outer` query loses the order during the `join` - the fix is to ensure the `order by` on the results

https://iguazio.atlassian.net/browse/ML-9408

### DIFF
--- a/server/py/services/api/tests/unit/crud/test_artifacts.py
+++ b/server/py/services/api/tests/unit/crud/test_artifacts.py
@@ -1,0 +1,96 @@
+# Copyright 2023 Iguazio
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+import time
+import uuid
+
+import sqlalchemy.orm
+
+import mlrun.common.schemas.artifact
+
+import services.api.crud
+
+
+class TestArtifacts:
+    def test_list_artifacts(
+        self,
+        db: sqlalchemy.orm.Session,
+    ):
+        tree, key = "tree", "key"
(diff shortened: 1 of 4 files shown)


server/py/services/api/tests/unit/db/test_artifacts.py
```diff
@@ -1557,12 +1557,16 @@ def test_list_artifact_for_tagging_fallback(self):
     @pytest.mark.parametrize("limit", [None, 6])
     def test_list_artifacts_returns_elements_by_order_updated_field(self, limit):
         project = "artifact_project"
+        artifact_kinds = ArtifactCategories.all()
```
TomerShor, line 1560: isn't this only implemented in https://github.com/mlrun/mlrun/pull/7328 ?

server/py/services/api/tests/unit/crud/test_artifacts.py
```diff
@@ -0,0 +1,96 @@
...
+
+import sqlalchemy.orm
+
+import mlrun.common.schemas.artifact
+
+import services.api.crud
+
+
+class TestArtifacts:
+    def test_list_artifacts(
```
TomerShor, line 26: why did you test it in the crud, and not only in the DB?

TomerShor (commented):
Looks good overall

server/py/services/api/tests/unit/db/test_artifacts.py
liranbg: yeah, I will wait with other pr to be merged as I worked on both in parallel. so that is a glitch

server/py/services/api/tests/unit/crud/test_artifacts.py
liranbg: I actually started from here then realized I need to work against mysql (thus integration test fix as well)
test itself has no harm, does it?

TomerShor: no harm, just wandered, because you are actually unit-testing the db functionality.

alonmr (approved):
lgtm 🔥

PR #7331, opened 2025-02-23
[Model Monitoring] Drop the TDEngine model monitoring database if it is empty

Eyal-Danieli:
Fix an issue in which the user has deleted all of his model monitoring resources but the database is still exist with no data. Once `delete_tsdb_resources()` is called, we try to delete the TDEngine mm db if it's empty by:
1 - Check if there are any remaining tables in the db by querying the `information_schema` stable metadata, specifically the `ins_tables` column ([TDengine Metadata - ins_tables](https://docs.tdengine.com/tdengine-reference/sql-manual/metadata/#ins_tables)).
2 - If no tables are found, drop the database.

https://iguazio.atlassian.net/browse/ML-9328


mlrun/model_monitoring/db/tsdb/tdengine/schemas.py
```diff
@@ -144,6 +144,12 @@ def _get_subtables_query(
             )
         return f"SELECT DISTINCT tbname FROM {self.database}.{self.super_table} WHERE {values};"
 
+    def drop_database_query(self) -> str:
```
assaf758: I wonder if it makes sense to have these methods in the TDEngineSchema and access them from the peri dictions super-table. Why not have them in the TDEngineConnector?

assaf758 (commented):
Looks good in general. 
Added one question inline.

assaf758 (approved):

PR #7336, opened 2025-02-23
[Nuclio] Fix python version enrichment and validation

TomerShor:
1. Nuclio removed support for Python 3.7 and 3.8 in version 1.14.1 .
2. Enrichment should be done to the defined default version.
3. The logic in `resolve_nuclio_runtime_python_image` is irrelevant since MLRun 1.5.0 , where python 3.7 support was removed.


server/py/services/api/tests/unit/crud/runtimes/nuclio/test_helpers.py
```diff
@@ -106,15 +106,15 @@ def test_resolve_work_dir_and_handler(handler, expected):
 @pytest.mark.parametrize(
     "mlrun_client_version,python_version,expected_runtime",
     [
-        ("1.3.0", "3.9.16", "python:3.9"),
-        ("1.3.0", "3.7.16", "python:3.7"),
+        ("1.9.0", "3.12.16", mlrun.mlconf.default_nuclio_runtime),
```
alonmr, line 109: shouldn't the expected be python 3.12?

alonmr (commented):
Looks good! Not sure about one test case below 👇

server/py/services/api/tests/unit/crud/runtimes/nuclio/test_helpers.py
TomerShor: It should, in the next version 😉 
This case doesn't go into [this condition](https://github.com/mlrun/mlrun/pull/7336/files#diff-c63a1c7dee3768ff7741fb563d0813efa6dda440864fc7d375036c136e1d651aR61-R65), so it returns the default version.
When the default will be 3.12, it will return it properly.

server/py/services/api/crud/runtimes/nuclio/helpers.py
```diff
@@ -39,26 +39,30 @@ def resolve_nuclio_runtime_python_image(
...
         return mlrun.mlconf.default_nuclio_runtime
 
-    # if mlrun version is older than 1.3.0 we need to use the previous default runtime which is python 3.7
+    # if mlrun version is older than 1.9.0 we need to use the previous default runtime which is python 3.9
     if semver.VersionInfo.parse(mlrun_client_version) < semver.VersionInfo.parse(
-        "1.3.0-X"
+        "1.9.0-X"
     ):
-        return "python:3.7"
+        return "python:3.9"
```
rokatyy, line 60: what if version is older than 1.3.0?

TomerShor: Python 3.7 is not supported anyway, so the "previous default" still applies

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -510,28 +510,27 @@ def _resolve_and_set_nuclio_runtime(
...
-    # For backwards compatibility, we need to adjust the runtime for old Nuclio versions
-    # TODO: remove in 1.8, default to 3.9
-    if services.api.crud.runtimes.nuclio.helpers.is_nuclio_version_in_range(
-        "0.0.0", "1.6.0"
+    if mlrun.runtimes.nuclio.function.validate_nuclio_version_compatibility(
+        "1.14.1",
     ) and nuclio_runtime in [
+        "python:3.6",
         "python:3.7",
         "python:3.8",
```
liranbg, line 524: worth doing the migration automatically -> `python:3.6` -> `default` as client using mlrun<>nuclio wont set the version explicitly but rather use nuclio which defaults its runtime to whatever legacy used to have.

TomerShor: you mean instead of failing, replace the version to the mlrun default version?
1. We only fail if the user set the python version excplicitly - automatically changing it might cause unwanted issues.
2. Having only a warning will not be visible, because this function is in the server.

liranbg: discussed offline - do the validation/enrichment only after setting the base image as if we set the base image to `mlrun/mlrun` there is no real meaning for the python version here.

liranbg (approved):

PR #7338, opened 2025-02-24
[Docs] Enhance Kubernetes install

jillnogold:
Update flags for NFS storage, M3/M4, and comment about Grafana statistics not working well


docs/install/kubernetes.md
```diff
@@ -138,10 +148,10 @@ Where:
 
 When the installation is complete, the helm command prints the URLs and ports of all the MLRun CE services.
 
-```{admonition} Note
-There is currently a known issue with installing the chart on Macs using Apple silicon (M1/M2). The current pipelines
-MySQL database fails to start. The workaround for now is to is to run this line `docker pull mysql:5.7 --platform linux/amd64` before installing the chart.
-in addition there is an issue with prometheus node selector the workaround for now is to opt out of kube-prometheus-stack by installing the chart with the `--set kube-prometheus-stack.enabled=false`.
+```{admonition} Known issue when installing the chart on Macs using Apple silicon (M1/M2/M3/M4):
```
liranbg:
```suggestion
```{admonition} Known issue when installing the chart on Macs using Apple silicon (arm based architecture):
```

liranbg (changes requested):
minor.
worth dropping the model name as it gets stale overtime

liranbg (approved):

PR #7340, opened 2025-02-24
[Functions] Add state column in DB table

alonmr:
The state column need was raised by model monitoring scale tests as it needs the function state to reflect on the model endpoints. The column eliminates the need to un-pickle the function blob which saves compute and memory consumption.

Note on a list query param - we want to allow specifying multiple states and the way to do so as a query param is `http://localhost:8000/items/?q=foo&q=bar` therefore instead of `q` being `states` it will be aliased as `state` respectively to "list runs" API.

https://iguazio.atlassian.net/browse/ML-9417


mlrun/db/httpdb.py
```diff
@@ -5135,6 +5138,7 @@ def _list_functions(
         format_: Optional[str] = None,
         since: Optional[datetime] = None,
         until: Optional[datetime] = None,
+        states: typing.Optional[list[mlrun.common.runtimes.constants.RunStates]] = None,
```
TomerShor: Why `RunStates` here, and everywhere else `FunctionState`?

TomerShor (changes requested):
Minor, otherwise LGTM

mlrun/db/httpdb.py
alonmr: opss copilot did me dirty

TomerShor (approved):

PR #7342, opened 2025-02-25
[KFP] Adhere to naming convention when retrying a pipeline

adamdelman:
Also, don't use the retry functionality in KFP server, instead always clone the run to retried to get better control of the run name.
https://iguazio.atlassian.net/browse/ML-9425


pipeline-adapters/mlrun-pipelines-kfp-v1-8/tests/utils.py
```diff
@@ -0,0 +1,39 @@
...
+        ("Retry of sample_run", "projectX", "projectX-Retry of sample_run"),
+        ("projectX-Retry of sample_run", "projectX", "projectX-Retry of sample_run"),
+        (
+            "  projectX-Retry of   sample_run  ",
+            "projectX",
+            "projectX-Retry of sample_run",
+        ),
+    ],
+)
+def test_normalize_retry_run(client, original_name, project, expected):
```
liranbg: make sure ci runs it thought

liranbg (changes requested):
bump pipeline version on setup.py

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/utils.py
```diff
@@ -101,10 +123,18 @@ def retry_run(
...
                 pipeline_parameters = pipeline_parameters[0]
 
+            desired_prefix = f"{project}-Retry of "
+            desired_prefix_lower = desired_prefix.lower()
+            current_name = run_details.name.strip()
+
+            if current_name.lower().startswith(desired_prefix_lower):
+                job_name = current_name
+            else:
+                job_name = self._normalize_retry_run(current_name, project)
```
theSaarco: This only handles the case where we re-run the pipeline (rather than do a retry). What happens if a "real" retry is executed? In that case we don't have any control over the name of the pipeline. I wonder if it might be better to actually always go through this flow, and have better control over the process. What say you? Are we losing anything by doing this flow instead of using the retry API?

adamdelman: @theSaarco Makes sense to me, I will make the appropriate changes.

PR #7343, opened 2025-02-25
[Workflows] Avoid sending client side completion notifications

Yacouby:
Ensure that the client side does not send redundant notifications when the run is completed. 
Previously, the client triggered a notification for both "running" and "completed" states, even though the server already handles the "completion" notifications. 
This fix updates the client to only send notifications when the run is in "running" state.

Jira:
https://iguazio.atlassian.net/browse/ML-8064

### DIFF
--- a/tests/system/projects/test_project.py
+++ b/tests/system/projects/test_project.py
@@ -1681,6 +1681,54 @@ class TestProject(TestMLRunSystem):
         project_from_db = self._run_db.get_project(name)
         assert project_from_db.source == source
 
+    def test_notifications_on_workflow_start_and_completion(self):
+        """
+        This test validates that notifications are sent correctly during the workflow run, when a notification with
+        when=["running", "completed"] is set.
+        one notification is sent when the workflow starts, and another is sent when the workflow completes successfully.
+        """
+        project_name = "my-project"
+        self.custom_project_names_to_delete.append(project_name)
+
+        project = self._load_remote_pipeline_project(name=project_name)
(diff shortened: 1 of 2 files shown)


liranbg (approved):
🚀

tests/system/projects/test_project.py
```diff
@@ -1681,6 +1681,54 @@ def test_remote_workflow_source_on_image(self):
...
+        time.sleep(35)
+
+        notifications = list(
+            notification_helpers.get_notifications_from_nuclio_and_reset_notification_cache(
+                nuclio_function_url
+            )
+        )
+
+        # Two notifications are expected: one from the client at the beginning when the workflow starts,
+        # and one from the server when the run completes.
```
liranbg, line 1727: why not doing so in a loop of waiting (with higher timeout (e.g 300 seconds)) and then exausht if you dont have len(notifications) of 2?

liranbg (commented):
idea to make test more stable

tests/system/projects/test_project.py
Yacouby: I don't think this is necessary because the test already ensures that both notifications have enough time to be processed. 
First, we wait for the pipeline to complete using `mlrun.wait_for_pipeline_completion(run_id)`, which guarantees that the "running" notification is sent at the start. Then, a 35-second sleep is added to allow the periodic monitor to detect the run completion and send the second notification.

liranbg: I see. alrighty.

PR #7346, opened 2025-02-25
[Model Monitoring] Fix model monitoring store prefixes CE

tomerm-iguazio:
1) Changed client spec to include mm store prefixes (default), so It will be possible to retrieve it from mlrun-api and override it in the client + mm pods (stream, controller, etc..).
The override is necessary to switch the paths from `v3io` to `s3` in CE.

2) Since we were using a futures pool, any raised error did not appear in the main thread;
we fixed this by writing an error log with traceback for every failure.

[ML-9423](https://iguazio.atlassian.net/browse/ML-9423)
[CEML-333](https://iguazio.atlassian.net/browse/CEML-333)

[ML-9423]: https://iguazio.atlassian.net/browse/ML-9423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CEML-333]: https://iguazio.atlassian.net/browse/CEML-333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Eyal-Danieli (approved):
LGTM

PR #7349, opened 2025-02-26
[Model Monitoring] Create TDEngine DB if not exists each time we create the infra STABLES

Eyal-Danieli:
Following #7331 we drop the TDEngine db if it's empty, which in other words means that we deleted a project with tsdb data and there are no more monitored projects in this db. Now, if the user would try to recreate a new project afterwards, he might face an issue which the DB doesn't exist (https://iguazio.atlassian.net/browse/ML-9436) because we use the same connection.
In this PR we address this issue by recreating the db if it doesn't exist each time we call `create_tables` instead of doing it once on the connection initialization. Note that this operation only occurs once during the project cycle when we set the credentials.

### DIFF
--- a/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
+++ b/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
@@ -75,13 +75,7 @@ class TDEngineConnector(TSDBConnector):
         """Establish a connection to the TSDB server."""
         logger.debug("Creating a new connection to TDEngine", project=self.project)
         conn = TDEngineConnection(self._tdengine_connection_profile.dsn())
-        conn.run(
-            statements=f"CREATE DATABASE IF NOT EXISTS {self.database}",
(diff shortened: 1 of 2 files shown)


davesh0812 (approved):

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -101,8 +95,27 @@ def _init_super_tables(self):
...
+    def _create_db_if_not_exists(self):
+        """Create the database if it does not exist."""
+        self.connection.prefix_statements = []
+        self.connection.run(
+            statements=f"CREATE DATABASE IF NOT EXISTS {self.database}",
+            timeout=self._timeout,
+            retries=self._retries,
+        )
+        self.connection.prefix_statements = [f"USE {self.database}"]
+        logger.debug(
```
assaf758, line 107: why do we need this log?

PR #7350, opened 2025-02-26
[Model Monitoring] Fix avoid `get_model_endpoint` call from serving

royischoss:
https://iguazio.atlassian.net/browse/ML-9265
Changes: 
1) `V2ModelServer` and `VotingEnsemble` avoid multiple API calls from `_lazy_init`, using class args for background task state, and model endpoints needed fields.
2) Change in the behavior of `create_model_endpoints` background task, divided to two parts: instructions, and the actual creation - allowing to pass mep uids to the serving classes
3) small changes in the instructions creation picking uid early for the mep while respecting creation_strategy
4) Checking background task status in   `BaseModelRouter::_update_background_task_state` handling every 15 seconds avoiding multiple api calls for each invoke


mlrun/serving/routers.py
```diff
@@ -195,6 +258,8 @@ def _resolve_route(self, body, urlpath):
 
     def _handle_event(self, event):
         name, route, subpath = self._resolve_route(event.body, event.path)
+        self._update_background_task_state(event)
```
davesh0812: I think all the the router uses `_pre_handle_event` so lets put `_update_background_task_state` only in there.. what do you think?

mlrun/serving/routers.py
```diff
@@ -160,6 +164,65 @@ def postprocess(self, event):
...
+                    logger.debug(
+                        f"Model endpoint creation task completed with state {background_task.status.state}"
+                    )
+                    self._background_task_terminate = True
+                else:  # in progress
+                    logger.debug(
+                        f"Model endpoint creation task is still in progress with the current state: "
+                        f"{background_task.status.state}. This event will not be monitored.",
+                        name=self.name,
+                        event_id=event_id,
```
davesh0812: lets remove this and add maybe timestamp that explain that all the event will not be monitoring from now to now+15sec

server/py/framework/db/sqldb/db.py
```diff
@@ -7563,8 +7563,12 @@ def list_model_endpoints(
         offset: typing.Optional[int] = None,
         limit: typing.Optional[int] = None,
         order_by: typing.Optional[str] = None,
-    ) -> mlrun.common.schemas.ModelEndpointList:
-        model_endpoints: list[mlrun.common.schemas.ModelEndpoint] = []
+        as_dict: bool = False,
+    ) -> Union[mlrun.common.schemas.ModelEndpointList, list[ModelEndpoint]]:
```
davesh0812: Your is call `as_dict` but you returning list?

server/py/services/api/api/endpoints/nuclio.py
```diff
@@ -271,19 +271,30 @@ async def deploy_function(
         background_tasks=[]
     )
     if function.get("kind") == mlrun.runtimes.RuntimeKinds.serving:
+        db_session = framework.db.session.create_session()
```
davesh0812: You can use here the `db_session` you got as param

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -1440,13 +1486,22 @@ def _extract_model_endpoints_from_function_graph(
...
+                str,
+            ]
+        ],
+        typing.Union[
+            mlrun.serving.states.RouterStep, mlrun.serving.states.RootFlowStep
+        ],
     ]:
+        endpoints_names_uid_map = {
+            endpoint.model_name: endpoint.uid for endpoint in model_endpoints
+        }
```
davesh0812: You should do as part of the list db function. Only in there the key needs to be - "<project_name>-<function_name>-<function_tag>-<name>".

davesh0812 (changes requested):
Nice. Few comments :)

davesh0812 (approved):

PR #7351, opened 2025-02-26
[Alerts] Adjust timing in sliding window alert system test

Yacouby:
This PR fixes a timing issue that occured in the `test_job_failure_alert_sliding_window` test to properly account for the periodic run monitor. 
The bug in the system test occurred because the test did not consider the delay introduced by the run monitor, which checks for job failures periodically. As a result, the second failure in the test was detected within the alert window, causing a  notification to be sent wrongly.

To fix this, the `time.sleep(mlconf.alerts.events_generation_interval)` was moved before the second failure, allowing the monitoring system sufficient time to detect the first failure. 
This ensures the alerting mechanism functions as expected and prevents false notifications.

Also, the alert period was minimized to 30 seconds to shorten the sleeps and the test duration.

### DIFF
--- a/tests/system/alerts/test_alerts.py
+++ b/tests/system/alerts/test_alerts.py
@@ -285,11 +285,8 @@ class TestAlerts(TestMLRunSystem):
             self.project.run_function(function_name)
 
         # wait for the periodic monitor runs function to run as it may take up to the maximum events_generation_interval
-        # to detect the event.
-        time.sleep(mlconf.alerts.events_generation_interval)
-
-        # Wait for more than half minute to simulate a delay that is slightly longer than the alert period
-        time.sleep(32)
+        # to detect the event + an extra 40s to simulate a delay that is slightly longer than the alert period
+        time.sleep(mlconf.alerts.events_generation_interval + 40)
 
         # this is the second failure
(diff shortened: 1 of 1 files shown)


tests/system/alerts/test_alerts.py
```diff
@@ -284,16 +284,17 @@ def test_job_failure_alert_sliding_window(self):
...
             self.project.run_function(function_name)
 
-        # Wait for more than two minutes to simulate a delay that is slightly longer than the alert period
-        time.sleep(125)
+        # wait for the periodic monitor runs function to run as it may take up to the maximum events_generation_interval
+        # to detect the event.
+        time.sleep(mlconf.alerts.events_generation_interval)
+
+        # Wait for more than half minute to simulate a delay that is slightly longer than the alert period
+        time.sleep(32)
```
TomerShor: Just sleep at once, and give it a bit more
```suggestion
        # wait for the periodic monitor runs function to run as it may take up to the maximum events_generation_interval
        # to detect the event + an extra 40s to simulate a delay that is slightly longer than the alert period
        time.sleep(mlconf.alerts.events_generation_interval + 40)
```

PR #7353, opened 2025-02-27
[API] Fix missing function status on list functions

alonmr:
Previous fix https://github.com/mlrun/mlrun/pull/7340
Related to https://github.com/mlrun/mlrun/pull/6783 and https://github.com/mlrun/mlrun/pull/6695


liranbg (changes requested):
Added UT to ensure next migration will fail on UT

liranbg (approved):

PR #7355, opened 2025-02-27
[Notifications] Increase timestamp precision to prevent skipped notifications

yaelgen:
resolves: [ML-9437](https://iguazio.atlassian.net/browse/ML-9432)

MySQL rounds `last_update_time` to milliseconds (using fsp=3), causing potential time gaps where notifications can be skipped if timestamps are rounded up.

The fix is to increase timestamp precision to fsp=6 (microseconds).
Added a migration to update the database schema, touches the schema for a few tables using `Datetime`.
This ensures all notifications are captured correctly in sliding window queries.

[ML-9437]: https://iguazio.atlassian.net/browse/ML-9437?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/py/services/api/main.py
```diff
@@ -773,6 +773,11 @@ def _push_terminal_run_notifications(
         # since the last time we pushed notifications.
         # On the first time we push notifications, we'll push notifications for all runs that are in a terminal state
         # and their notifications haven't been sent yet.
+        self._logger.debug(
+            "Checking notifications since last end time",
+            last_update_time=last_update_time,
+            now=datetime.datetime.now(),
```
liranbg: you dont need to log the `now` as the `now` is logged with the log line itself

server/py/services/api/main.py
```diff
@@ -782,6 +787,13 @@ def _push_terminal_run_notifications(
             with_notifications=True,
         )
 
+        if not len(runs):
+            self._logger.debug(
+                "No runs ended during the current window",
+                last_update_time=last_update_time,
+                now=datetime.datetime.now(),
```
liranbg: see above + `end_time_from=last_update_time,`

liranbg (commented):
minor

quaark (approved):

server/py/services/api/migrations/versions/b426a82845ea_datetime_fsp_to_6.py
adamdelman, line 1: I think that converting the datetime columns from timezone naive to timezone aware can cause problems on systems that are not in the UTC timezone. 
Did you account for this?

server/py/services/api/main.py
```diff
@@ -782,6 +786,12 @@ def _push_terminal_run_notifications(
             with_notifications=True,
         )
 
+        if not len(runs):
```
adamdelman, line 789: if not runs?

adamdelman (commented):
Overall looks good, one minor point to consider.

server/py/framework/utils/db/sql_types.py
```diff
@@ -29,7 +29,7 @@ class _Timestamp:
 
     class _Datetime:
         sqlite = sqlalchemy.DATETIME(timezone=True)
-        mysql = sqlalchemy.dialects.mysql.DATETIME(timezone=True, fsp=3)
+        mysql = sqlalchemy.dialects.mysql.DATETIME(timezone=True, fsp=6)
```
rokatyy: It will break part of alert activations logic [here](https://github.com/mlrun/mlrun/blob/9027287235ea58f0187a362d3b4718b50abe4795/server/py/framework/utils/notifications/notification_pusher.py#L192-L195) as we convert to mysql format with fsp=3. Also, higher precision will increase the size of data which is super crucial for alert activations. So at least in alert activations, I'd not change that

rokatyy (changes requested):
Having concerns about alert activations table..

server/py/framework/db/sqldb/models.py
```diff
@@ -862,7 +862,7 @@ class AlertActivation(Base, mlrun.utils.db.BaseModel):
         )
 
         id = Column(Integer, autoincrement=True)
-        activation_time = Column(SQLTypesUtil.datetime(), nullable=False)
+        activation_time = Column(SQLTypesUtil.datetime(fsp=3), nullable=False)
```
rokatyy, line 869: Please add comment that it should not be changed for future :) (and same for reset_time)

rokatyy (approved):
LGTM, just a minor thing

PR #7363, opened 2025-03-02
[Deprecations] Postpone deprecations to 1.9

alonmr:
We have identified deprecations late into release process and we don't want to remove them due to risk assessment.
Therefore we postpone them to 1.9

https://iguazio.atlassian.net/browse/ML-9365

### DIFF
--- a/server/py/framework/db/sqldb/models.py
+++ b/server/py/framework/db/sqldb/models.py
@@ -200,9 +200,7 @@ with warnings.catch_warnings():
     warnings.simplefilter("ignore")
 
     # deprecated, use ArtifactV2 instead
-    # TODO: remove in 1.8.0. Note that removing it will require upgrading mlrun in at least 2 steps:
-    #  1. upgrade to 1.6.x which will create the new table
-    #  2. upgrade to 1.7.x which will remove the old table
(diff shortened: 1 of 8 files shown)


server/py/framework/db/sqldb/models.py
```diff
@@ -200,9 +200,7 @@ class Notification(Base, mlrun.utils.db.BaseModel):
     warnings.simplefilter("ignore")
 
     # deprecated, use ArtifactV2 instead
-    # TODO: remove in 1.8.0. Note that removing it will require upgrading mlrun in at least 2 steps:
-    #  1. upgrade to 1.6.x which will create the new table
-    #  2. upgrade to 1.7.x which will remove the old table
```
TomerShor, line 205: The versions in the comment are not correct, but the 2-phase procedure will still be required once we remove the table.

alonmr: I think it is reflected in the new comment.
Once data migration v5 is removed,  we can assume no legacy artifacts pre upgrade. Then it will also be safe to remove the table in the same version.

liranbg (approved):

PR #7372, opened 2025-03-04
[Workflows] Fix set workflow to avoid looking for remote paths locally

Yacouby:
This fix ensures that `set_workflow` does not look for the workflow file locally when the project source points to a remote source (e.g., Git). 
Previously, the workflow file was always expected to exist locally, causing failures when using remote sources.

Jira:
https://iguazio.atlassian.net/browse/ML-9442

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -1412,7 +1412,9 @@ class MlrunProject(ModelObj):
         """
 
         # validate the provided workflow_path
-        self._validate_file_path(workflow_path, param_name="workflow_path")
+        self._validate_file_path(
+            workflow_path, param_name="workflow_path", engine=engine
+        )
 
         if engine and "local" in engine and schedule:
             raise ValueError("'schedule' argument is not supported for 'local' engine.")
@@ -5241,7 +5243,7 @@ class MlrunProject(ModelObj):
             if is_remote_enriched:
                 self.spec.repo.remotes[remote].set_url(clean_remote, enriched_remote)
 
-    def _validate_file_path(self, file_path: str, param_name: str):
+    def _validate_file_path(self, file_path: str, param_name: str, engine: str):
         """
         The function checks if the given file_path is a valid path.
         If the file_path is a relative path, it is completed by joining it with the self.spec.get_code_path()
@@ -5266,10 +5268,7 @@ class MlrunProject(ModelObj):
                 f"Invalid '{param_name}': '{file_path}'. Got a remote URL without a file suffix."
             )
 
-        _, in_context = self.get_item_absolute_path(file_path)
-
-        # if the file source is remote then skip the local file validation
-        if in_context and self.spec.source:
+        if engine and not engine.startswith("remote"):
             return
 
         code_path = self.spec.get_code_path()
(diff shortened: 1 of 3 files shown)


adamdelman (approved):
LGTM

alonmr (approved):

tests/system/projects/test_project.py
```diff
@@ -1729,6 +1729,20 @@ def test_notifications_on_workflow_start_and_completion(self):
             len(notifications) == 2
         ), f"Expected 2 notifications, got {len(notifications)}"
 
+    def test_set_and_run_remote_workflow(self):
```
alonmr: You need to create the project without copying the files.
1. Use `new_project` without source and then `project.set_source`
2. Verify that when using `kfp` engine this files with proper error message such as "File not found"

alonmr (changes requested):

tests/system/projects/test_project.py
```diff
@@ -1729,6 +1729,21 @@ def test_notifications_on_workflow_start_and_completion(self):
...
+        """
+        This test verifies that a workflow can be set and executed remotely when the project source
+        is a remote repository (e.g., a Git repository).
```
alonmr: We have tests for the current description so I think it needs to be elaborated
```suggestion
        This test verifies that a workflow can be set and executed remotely when the project source
        is a remote repository (e.g., a Git repository) and the function files are not present locally.
```

alonmr (approved):

mlrun/projects/project.py
```diff
@@ -5266,6 +5266,12 @@ def _validate_file_path(self, file_path: str, param_name: str):
                 f"Invalid '{param_name}': '{file_path}'. Got a remote URL without a file suffix."
             )
 
+        _, in_context = self.get_item_absolute_path(file_path)
```
liranbg: this function may raise `FileNotFoundError`. how can that not be a regression? what guards us from falling into that exception?

liranbg (changes requested):

mlrun/projects/project.py
alonmr: It raises file not found only when the file is an absolute path i.e. not in context. It would also raise `MLRunInvalidArgumentError` for the same reason below which is more elaborated.
I think we need to add a UT case on `test_set_workflow_path_validation` for absolute path that doesn't exist and make sure that the error is clear

tests/system/projects/test_project.py
```diff
@@ -1729,6 +1729,21 @@ def test_notifications_on_workflow_start_and_completion(self):
             len(notifications) == 2
         ), f"Expected 2 notifications, got {len(notifications)}"
 
+    def test_set_and_run_remote_workflow(self):
```
liranbg: remove from system test, need to be unit or removed completely

mlrun/projects/project.py
liranbg: This entirely can be removed. the validation should happen only if engine is provided and engine is NOT remote.
that means, if user provide `engine=remote` or `engine=remote:kfp` then you will skip the validation itself.

liranbg (changes requested):

liranbg (approved):

PR #7377, opened 2025-03-04
[Serving] Fix serialization and deployment of new model class

gtopper:


mlrun/serving/states.py
```diff
@@ -981,25 +981,32 @@ def __init__(
...
         super().__init__(
             *args,
             class_name="mlrun.serving.ModelRunner",
-            class_args=dict(runnables=self._models, model_selector=model_selector),
+            class_args=dict(model_selector=model_selector),
             **kwargs,
         )
 
-    def add_model(self, model: Model) -> None:
+    def add_model(self, model: Union[str, Model], **model_parameters) -> None:
```
davesh0812, line 991: please add docs

mlrun/serving/states.py
```diff
@@ -981,25 +981,32 @@ def __init__(
...
         """Add a Model to this ModelRunner."""
-        self._models.append(model)
+        models = self.class_args.get("models", [])
+        models.append((model, model_parameters))
+        self.class_args["models"] = models
```
davesh0812, line 1000: not sure you need this line... list is mutable

davesh0812 (commented):
Nice, I think we should create tiny sys test also.

mlrun/serving/states.py
gtopper: Yup, you need it in case the list didn't exist already.

gtopper:
Test WIP. On hold due to higher priority tasks.

tests/system/runtimes/test_nuclio.py
```diff
@@ -58,6 +59,29 @@ def test_deploy_function_with_error_handler(self):
...
+        graph = function.set_topology("flow", engine="async")
+        model_runner_step = ModelRunnerStep()
+        model_runner_step.add_model("DummyModel", name="my-model")
+
+        graph.to(model_runner_step)
+
+        self._logger.debug("Deploying nuclio function")
+        deployment = function.deploy()
+
+        assert deployment == function.get_url()  # check function url
```
davesh0812, line 83: please add invocaion, to see the full flow is working and also it is the only referance for now we have for understanding the new mechnisem.

davesh0812 (approved):

PR #7383, opened 2025-03-05
[System tests] Decouple dask system tests

alonmr:
We see `tests_dask_close` keeps failing due to dask pipeline that failed with image pull backoff and kfp keeps retrying it automatically.
The test checks if there is more than 1 pod in the dask cluster and it gets more because of that pipeline from another test.
The following fixes should resolve the issue:
* Remove all dask pipeline runs from kfp so it will stop retiring it.
* Change the dask cluster name between tests to decouple them.

### DIFF
--- a/tests/system/examples/dask/test_dask.py
+++ b/tests/system/examples/dask/test_dask.py
@@ -14,6 +14,8 @@
 #
 import datetime
 import os
+import random
+import string
 
 import kfp
 import kfp.compiler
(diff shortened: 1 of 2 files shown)


tests/system/examples/dask/test_dask.py
```diff
@@ -83,10 +73,12 @@ def test_dask(self):
         assert run_object.state() == "completed"
 
     def test_run_pipeline(self):
+        dask_function = self._generate_dask_function("my-dask-2")
```
liranbg: just randomize the name with 3-4 chars instead of relying on incrementing it manually

liranbg (changes requested):

liranbg (approved):

PR #7385, opened 2025-03-05
[ModelMonitoring] Support multiple workers for mm applications

alxtkr77:
https://iguazio.atlassian.net/browse/ML-9439

### DIFF
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -567,18 +567,18 @@ default_config = {
         },
         "application_stream_args": {
             "v3io": {
-                "shard_count": 1,
+                "shard_count": 8,
                 "retention_period_hours": 24,
-                "num_workers": 1,
-                "min_replicas": 1,
-                "max_replicas": 1,
+                "num_workers": 4,
+                "min_replicas": 8,
+                "max_replicas": 8,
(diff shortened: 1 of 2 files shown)


mlrun/config.py
```diff
@@ -567,18 +567,18 @@
...
             },
             "kafka": {
-                "partition_count": 1,
+                "partition_count": 16,
                 "replication_factor": 1,
-                "num_workers": 1,
+                "num_workers": 4,
                 "min_replicas": 1,
-                "max_replicas": 1,
+                "max_replicas": 8,
```
davesh0812: If I understand the requirement right... I think this should be  the config
```suggestion
                "shard_count": 4,
                "retention_period_hours": 24,
                "num_workers": 4,
                "min_replicas": 1,
                "max_replicas": 1,
            },
            "kafka": {
                "partition_count": 4,
                "replication_factor": 1,
                "num_workers": 4,
                "min_replicas": 1,
                "max_replicas": 1,
```

alxtkr77: Thanks! Will change as you suggest

assaf758 (approved):

PR #7386, opened 2025-03-05
[Tests] fix TestRecordResults::test_inference_feature_set and TestInferenceWithSpecialChars::test_inference_feature_set

davesh0812:


tests/system/model_monitoring/test_model_monitoring.py
```diff
@@ -1387,19 +1386,20 @@ def _train(cls) -> None:
             cls.y_train,  # pyright: ignore[reportGeneralTypeIssues]
         )
 
-    def _get_monitoring_feature_set(self) -> mlrun.feature_store.FeatureSet:
+    def _get_monitoring_feature_set(
+        self, endpoint_id: str
+    ) -> mlrun.feature_store.FeatureSet:
         model_endpoint = mlrun.get_run_db().get_model_endpoint(
```
Eyal-Danieli: A more proper approach would be passing the feature set uri when calling `_test_feature_names` from `test_inference_feature_set` (because you have the mep record at that point) and then you could retrieve it directly without getting the mep from the db again. Anyway it's not mandatory for now

tests/system/model_monitoring/test_model_monitoring.py
```diff
@@ -1432,19 +1432,18 @@ def test_inference_feature_set(self) -> None:
...
             ),
             model_endpoint_name=self.model_endpoint_name,
-            function_name=self.function_name,
             context=mlrun.get_or_create_ctx(name=f"{self.name_prefix}-context"),  # pyright: ignore[reportGeneralTypeIssues]
             infer_results_df=self.infer_results_df,
             # TODO: activate ad-hoc mode when ML-5792 is done
         )
 
-        self._test_feature_names()
+        self._test_feature_names(endpoint_id=model_endpoint.metadata.uid)
```
Eyal-Danieli: if you implement my above suggestion then it would be
```suggestion
        self._test_feature_names(endpoint_id=model_endpoint.metadata.uid, monitoring_feature_set_uri=model_endpoint.spec.monitoring_feature_set_uri)
```

Eyal-Danieli (approved):
approved with 1 suggestion for improvement

Eyal-Danieli (approved):

PR #7387, opened 2025-03-05
[Images] Bump python from 3.9.18 to 3.9.21 on mlrun/* images

liranbg:
remediating vulnerabilities and updating toolchain

https://iguazio.atlassian.net/browse/ML-9500


alonmr (approved):
I'm assuming python patch version is automatic on cache invalidation because we don't pin it in CI. It also depends on `gcr.io/iguazio/python` as base image.
Only thing that is not trivial is miniconda upgrade so consider it might cause backlash

liranbg:
> I'm assuming python patch version is automatic on cache invalidation because we don't pin it in CI. It also depends on `gcr.io/iguazio/python` as base image. Only thing that is not trivial is miniconda upgrade so consider it might cause backlash

indeed conda version has updated. this is required to ensure python latest patch.

PR #7389, opened 2025-03-05
[Notifications] Ensure run state update precedes end time update in run status

yaelgen:
resolves: https://iguazio.atlassian.net/browse/ML-9432

Some runs were observed where the end time was updated before the state, causing the initial update timestamp to be in the past. This led to runs falling outside the expected monitoring window. This fix ensures the state update occurs before setting the end time.


liranbg (approved):

quaark (approved):

alonmr (commented):
I see you removed the end time update from both the server and the client so I'm not sure where is it being updated at all?
And since you removed it from the client, will local runs be updated correctly?

alonmr (commented):
OK I checked `framework.db.sqldb.db.SQLDB.update_run` and I see the changes there so that answers my previous question and its definitely better place for this change.
My new question is why not set `last_update` to be the same as `end_time` ?

liranbg:
> OK I checked `framework.db.sqldb.db.SQLDB.update_run` and I see the changes there so that answers my previous question and its definitely better place for this change. My new question is why not set `last_update` to be the same as `end_time` ?

because run can be updated even after it has finished

alonmr:
> > OK I checked `framework.db.sqldb.db.SQLDB.update_run` and I see the changes there so that answers my previous question and its definitely better place for this change. My new question is why not set `last_update` to be the same as `end_time` ?
> 
> because run can be updated even after it has finished

I just think it's weird that the end time is X and the update time is X+1 but it's really not important.
I was referring to the single update that also ends the run and not the subsequent updates.
Either way lgtm

alonmr (approved):

PR #7406, opened 2025-03-06
[Model Monitoring]  Update list _mep (when latest_only=true) & raise explainable error when trying to delete model artifact which related to existed Model Endpoint

davesh0812:
resolves https://iguazio.atlassian.net/browse/ML-9507, https://iguazio.atlassian.net/browse/ML-9534

Summary of Changes:

1. `delete_artifact` – Added a validation step to check if the artifact is referenced by any model endpoints before removing its data.
2. `delete_artifacts` – No functional changes. However, when using this API, any artifacts currently used by model endpoints will not be removed along wuth its data.
3. `list_model_endpoints` – When passing `latest_only=True`, we now also verify whether the model endpoint has an associated function or if it is a batch endpoint.


server/py/framework/db/sqldb/models.py
```diff
@@ -935,16 +935,17 @@ class ModelEndpoint(Base, mlrun.utils.db.HasStruct):
         )
         function_id = Column(
             Integer,
-            ForeignKey("functions.id"),
+            ForeignKey("functions.id", ondelete="SET NULL"),
```
assaf758, line 938: Let's discuss if indeed "SET NULL" is the way to go..
Case in point: Deleting a model used by MEP might break the mm-apps if they use the model data for their work..

server/py/services/api/crud/artifacts.py
```diff
@@ -242,10 +243,46 @@ def delete_artifact(
...
+                db_session,
+                key,
+                tag,
+                project=project,
+                producer_id=producer_id,
+                object_uid=object_uid,
+                iter=iteration,
+            )
+        try:
+            return_val = framework.utils.singletons.db.get_db().del_artifact(
```
moranbental: You change the logic here. First, we delete the artifact data, then the artifact from the database. This is intentional because deleting the artifact data can fail. We proceed to delete the artifact from the database only if the deleting strategy is `data_optional`
It should be covered by a test, but the problem is that this test, `test_fails_deleting_artifact_data`, is not verifying that logic. It doesn't check that after `delete_artifact_data` fails, the artifact still exists in the db when the `deletion_strategy` is `data_force`

server/py/services/api/crud/artifacts.py
```diff
@@ -242,10 +243,46 @@ def delete_artifact(
...
+                session=db_session,
+                key=key,
+                tag=tag,
+                project=project,
+                uid=object_uid,
+                producer_id=producer_id,
+                iter=iteration,
+            )
+        except sqlalchemy.exc.IntegrityError as exc:
```
moranbental: Move the try-except block inside the `del_artifact` method. Also, you don't need to pass the `deletion_strategy` because once we reach the point of deleting the artifact from the db, the deletion strategy doesn't matter

moranbental (changes requested):

server/py/services/api/crud/artifacts.py
davesh0812: No problem, I reverted the order and added a check to ensure the artifact (model) can be safely removed before deleting its data.

server/py/framework/db/sqldb/db.py
```diff
@@ -1065,6 +1065,40 @@ def list_artifact_tags(
...
+        self,
+        session,
+        key,
+        tag="",
+        iter=None,
+        project="",
+        producer_id: Optional[str] = None,
+        uid: Optional[str] = None,
+    ):
+        db_artifact = self.read_artifact(
```
moranbental: In case the artifact is not found, it will raise an error here. However, it should continue as it does here - https://github.com/mlrun/mlrun/blob/64fa22942836aedc7e537a6e0650fb82120949bb/server/py/framework/db/sqldb/db.py#L5025

So, you need to add a try-except block here

server/py/framework/db/sqldb/db.py
```diff
@@ -1065,6 +1065,40 @@ def list_artifact_tags(
         # the query returns a list of tuples, we need to extract the tag from each tuple
         return [tag for (tag,) in query]
 
+    def is_artifact_can_be_removed(
```
moranbental: add docstring also here

server/py/framework/db/sqldb/db.py
```diff
@@ -1065,6 +1065,40 @@ def list_artifact_tags(
...
+        self,
+        session,
+        key,
+        tag="",
+        iter=None,
+        project="",
```
```suggestion
        key: str,
        tag: str = "",
        iter: Optional[int] = None,
        project: str = "",
```

server/py/framework/db/sqldb/db.py
```diff
@@ -1065,6 +1065,40 @@ def list_artifact_tags(
...
+            as_record=True,
+        )
+        dependent_endpoints_count = (
+            session.query(ModelEndpoint)
+            .filter(ModelEndpoint.model_id == db_artifact.id)
+            .count()
+        )
+        if dependent_endpoints_count:
+            raise mlrun.errors.MLRunConflictError(
+                f"Failed deleting artifact {key} in project {project}, tag {tag}"
```
moranbental, line 1122: Add also the `db_artifact.uid` in the error message, as this is the unique field

server/py/framework/db/base.py
```diff
@@ -288,6 +288,36 @@ def list_artifact_tags(
     ):
         return []
 
+    def is_artifact_can_be_removed(
+        self,
+        session,
+        key,
+        tag="",
+        iter=None,
+        project="",
```
moranbental:
```suggestion
        key: str,
        tag: str = "",
        iter: Optional[int] = None,
        project: str = "",
```

server/py/services/api/crud/artifacts.py
```diff
@@ -319,11 +369,12 @@ def _delete_artifact_data(
         ),
         secrets: typing.Optional[dict] = None,
         auth_info: mlrun.common.schemas.AuthInfo = mlrun.common.schemas.AuthInfo(),
+        artifact: typing.Optional[dict] = None,
     ):
         logger.debug("Deleting artifact data", project=project, key=key, tag=tag)
 
         try:
-            artifact = self.get_artifact(
+            artifact = artifact or self.get_artifact(
```
moranbental, line 343: There is no need to call `get_artifact` anymore, and it is no longer optional. Also, you no longer need to pass object_uid, producer_id, iteration, and tag

server/py/framework/db/sqldb/db.py
```diff
@@ -1065,6 +1065,40 @@ def list_artifact_tags(
...
+                f", iteration {iter}. The artifact is used by {dependent_endpoints_count} endpoints"
+            )
+        return mlrun.common.formatters.ArtifactFormat.format_obj(
+            db_artifact.full_object, mlrun.common.formatters.ArtifactFormat.full
```
moranbental: format can be minimal

moranbental (changes requested):
Looks good! missing some unit test for artifacts

moranbental (commented):
Why not implement it also for `delete_artifacts`?

server/py/services/api/crud/artifacts.py
moranbental: @davesh0812 I have updated the unit test accordingly - https://github.com/mlrun/mlrun/pull/7414

server/py/services/api/crud/artifacts.py
davesh0812: I added the `artifact` parameter as optional. Do you prefer it to be mandatory?
I know this method currently has only one usage, but I think it's better to keep it optional for flexibility and future use.

davesh0812:
> Why not implement it also for `delete_artifacts`? @moranbental 

As I mentioned in the description allready now - When using this API, any artifacts currently used by model endpoints will not be removed along with its data.

server/py/services/api/crud/artifacts.py
```diff
@@ -240,7 +241,18 @@ def delete_artifact(
...
+            key=key,
+            tag=tag,
+            iter=iteration,
+            project=project,
+            producer_id=producer_id,
+            object_uid=object_uid,
+        )
+        if not artifact:
+            return None
+        # artifact = None
```
assaf758: remove

assaf758 (approved):
Nice!
One minor comment

server/py/services/api/tests/unit/api/test_artifacts.py
```diff
@@ -1233,6 +1233,65 @@ def test_list_artifacts_partition_by(db: Session, unversioned_client: TestClient
     )
 
 
+def test_failed_to_delete_artifact_with_referenced_model_endpoint(
```
moranbental, line 1236: Why is the test here? You're testing the db, not the API. It should be in this file`server/py/services/api/tests/unit/db/test_artifacts.py`

server/py/services/api/tests/unit/api/test_artifacts.py
```diff
@@ -1233,6 +1233,65 @@ def test_list_artifacts_partition_by(db: Session, unversioned_client: TestClient
...
+        framework="xgboost",
+    )
+
+    response = unversioned_client.post(
+        STORE_API_ARTIFACTS_PATH.format(project=PROJECT, uid=UID, key="model", tag=TAG),
+        data=model_artifact.to_json(),
+    )
+    assert (
+        response.status_code == HTTPStatus.OK.value
+    ), f"Expected 200 OK when storing the artifact, got {response.status_code}: {response.text}"
```
moranbental: Ensure the artifact was created by getting the artifact

moranbental (changes requested):
minor

moranbental (approved):
LGTM

server/py/framework/db/sqldb/db.py
```diff
@@ -1071,6 +1071,62 @@ def list_artifact_tags(
         # the query returns a list of tuples, we need to extract the tag from each tuple
         return [tag for (tag,) in query]
 
+    def is_artifact_can_be_removed(
```
alonmr: name is meh
```suggestion
    def validate_artifact_removal_preconditions(
```

server/py/services/api/crud/artifacts.py
```diff
@@ -284,6 +296,45 @@ def delete_artifacts(
             db_session, name, project, tag, labels, producer_id=producer_id
         )
 
+    @classmethod
```
alonmr: why classmethod? I don't think this method is needed at all tbh just call `framework.utils.singletons.db.get_db().is_artifact_can_be_removed`

alonmr (changes requested):
Skimmed through have small comments. Can you explain what covers us in `del_artifacts` with meps?

davesh0812:
> Skimmed through have small comments. Can you explain what covers us in `del_artifacts` with meps?
At the moment, the `del_artifacts` function only removes artifacts from the database; it does not delete the associated artifact data.

As you can see [here](https://github.com/mlrun/mlrun/blob/868b5d2ffcfe6e6299fef4f4f10a152ea418d308/server/py/framework/db/sqldb/db.py#L1020), if an artifact is still referenced by a model endpoint, the deletion will raise an `IntegrityError`, and the artifact will not be removed.

If, in the future, this API is extended to delete the artifact data as well, you should include the validation I added here before attempting the deletion.
I add a TODO in the code.

server/py/services/api/crud/artifacts.py
```diff
@@ -279,6 +291,9 @@ def delete_artifacts(
         auth_info: mlrun.common.schemas.AuthInfo = mlrun.common.schemas.AuthInfo(),
         producer_id: typing.Optional[str] = None,
     ):
+        # TODO : If, in the future, this API is extended to delete the artifact data as well,
+        #  we should include the validation we added in validate_artifact_removal_preconditions
+        #  before attempting the deletion.
```
alonmr:
```suggestion
        # TODO : If, in the future, this API is extended to delete the artifact data as well,
        #  we should include the validation we added in validate_artifact_removal_preconditions
        #  before attempting the data deletion. Currently, deleting artifacts linked to model endpoints will fail with IntegrityError.
```

alonmr (approved):

alonmr (approved):

PR #7410, opened 2025-03-09
[Notifications] Fix stuck pending notifications

alonmr:
Some notifications may be stuck on pending and never be sent although the run has completed due to db session isolation level. Fix is to start a new session.
This also fixes a situation where the server sets a run to running although it is completed without removing the `end_time`. This can happen if the pod didn't terminate but the client updated the run to completed.

https://iguazio.atlassian.net/browse/ML-9432

### DIFF
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -304,8 +304,13 @@ class SQLDB(DBInterface):
             run.state in mlrun.common.runtimes.constants.RunStates.terminal_states()
             and not run.end_time
         ):
-            end_time = run_end_time(struct)
-            self._update_run_end_time(run, struct, now=end_time)
+            self._update_run_end_time(run, struct, now=datetime.now(timezone.utc))
+        elif (
+            run.state not in mlrun.common.runtimes.constants.RunStates.terminal_states()
+        ):
+            # Ensure end time is not set if the run is not in a terminal state
+            run.end_time = None
+            struct.setdefault("status", {}).pop("end_time", None)
 
         # Update the labels only if the run updates contains labels
         if run_labels(updates):
@@ -571,6 +576,13 @@ class SQLDB(DBInterface):
         ):
             end_time = run_end_time(run_data)
             self._update_run_end_time(run, run_data, now=end_time)
+        elif (
+            run.state not in mlrun.common.runtimes.constants.RunStates.terminal_states()
+        ):
+            # Ensure end time is not set if the run is not in a terminal state
+            run.end_time = None
+            run_data.setdefault("status", {}).pop("end_time", None)
+
         run.struct = run_data
 
     def _add_run_name_query(self, query, name):
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -300,17 +300,7 @@ class SQLDB(DBInterface):
         if start_time:
             run.start_time = start_time
 
-        if (
-            run.state in mlrun.common.runtimes.constants.RunStates.terminal_states()
-            and not run.end_time
-        ):
-            self._update_run_end_time(run, struct, now=datetime.now(timezone.utc))
-        elif (
-            run.state not in mlrun.common.runtimes.constants.RunStates.terminal_states()
-        ):
-            # Ensure end time is not set if the run is not in a terminal state
-            run.end_time = None
-            struct.setdefault("status", {}).pop("end_time", None)
+        self._update_run_end_time(run, struct)
 
         # Update the labels only if the run updates contains labels
         if run_labels(updates):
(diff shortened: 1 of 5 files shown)


server/py/framework/db/sqldb/db.py
```diff
@@ -571,6 +576,13 @@ def _enrich_run_model(self, now: datetime, run: Run, run_data: dict):
         ):
             end_time = run_end_time(run_data)
             self._update_run_end_time(run, run_data, now=end_time)
+        elif (
```
liranbg: looks code dup with above
```
f (
            run.state in mlrun.common.runtimes.constants.RunStates.terminal_states()
            and not run.end_time
        ):
            self._update_run_end_time(run, struct, now=datetime.now(timezone.utc))
        elif (
            run.state not in mlrun.common.runtimes.constants.RunStates.terminal_states()
        ):
            # Ensure end time is not set if the run is not in a terminal state
            run.end_time = None
            struct.setdefault("status", {}).pop("end_time", None)
```


can be moved out ?

server/py/framework/utils/time_window_tracker.py
```diff
@@ -141,3 +130,12 @@ async def run_with_time_window_tracker(
             cycle_tracker.update_window,
             now,
         )
+        ensure_window_update = False
```
liranbg, line 134: why was it removed from creating the window update?

liranbg (changes requested):

server/py/framework/db/sqldb/db.py
alonmr: It's similar but I kept this `end_time = run_end_time(run_data)` for "store run" because I think it might be useful to allow setting the end time by client input via store.
Can extract it with `end_time` parameter.

server/py/framework/utils/time_window_tracker.py
alonmr: the window update just above (lines 128-132). once done, there is no need to update again in the finally close. It's a insurance that window update will happen if the callback failed but it's not always wanted `_generate_events`

liranbg (approved):

server/py/framework/utils/time_window_tracker.py
quaark: Maybe add this as a comment so the reasoning here is clear?

server/py/framework/utils/time_window_tracker.py
```diff
@@ -141,3 +130,12 @@ async def run_with_time_window_tracker(
...
+    finally:
+        await run_in_threadpool(framework.db.close_session, db_session)
+        if ensure_window_update:
+            await run_in_threadpool(
+                framework.db.session.run_function_with_new_db_session,
+                cycle_tracker.update_window,
+                now,
+            )
```
quaark: I'd also add a comment here since its not trivially obvious why you are closing the session and immediately opening a new one

server/py/services/api/tests/unit/db/test_runs.py
```diff
@@ -425,6 +425,46 @@ def test_update_run_success(self):
...
+        project, name, uid, iteration, run = self._create_new_run(
+            state=mlrun.common.runtimes.constants.RunStates.completed
+        )
+        run = self._db.read_run(self._db_session, uid, project, iteration)
+
+        # Store completed expected to fill end time
+        initial_end_time = run["status"]["end_time"]
+        assert initial_end_time is not None
+
+        self._create_new_run(state=mlrun.common.runtimes.constants.RunStates.running)
```
quaark, line 439: From reading the rest of the test, I'm guessing this line isn't creating a new run but updating the same run, as you are querying the same uid and iteration of the old one and expecting the state to be running. Or am I missing something? If that is indeed the case, maybe use update here instead of create? Or add a comment explaining what's happening.

server/py/services/api/tests/unit/db/test_runs.py
```diff
@@ -425,6 +425,46 @@ def test_update_run_success(self):
             assert run["status"]["end_time"] is not None
             assert update_labels_mock.call_count == 0
 
+    def test_store_and_update_run_from_terminal_state_to_non_terminal_state(self):
```
quaark, line 428: I suggest adding to the test that 2 consecutive updates to `state=completed` keeps the `end_time` of the first update and the second update doesn't change the field.

quaark (changes requested):
Looks really good! A couple minor things - mainly code comments

server/py/services/api/tests/unit/db/test_runs.py
alonmr: deliberately using create to test the store flow as well. added a comment

server/py/framework/db/sqldb/db.py
```diff
@@ -589,6 +580,23 @@ def _ensure_run_name_on_update(run_record: Run, run_dict: dict):
                 "Changing name for an existing run is invalid"
             )
 
+    @staticmethod
+    def _update_run_end_time(run: Run, run_dict: dict, end_time: Optional[str] = None):
```
yaelgen, line 584: Please add a comment here that explains what is the case where we don't want to update the run's end time, cause the if statements here looks a bit confusing at first sight

yaelgen (changes requested):
LGTM, added a comment and agree with Adam on the test

quaark (approved):
🚀

yaelgen (approved):
🚀

PR #7411, opened 2025-03-09
[FeatureStore] Conclude deprecation of _get_offline_features() and _get_online_features_service() + fix system tests

alxtkr77:
- Checks if LangChain is not installed before running vectordb tests
- Fixes the alarm system tests to support the latest model monitoring datastore profile convention


tests/system/datastore/test_vectorstore.py
```diff
@@ -13,6 +13,8 @@
 # limitations under the License.
 #
 
+# Check if langchain_community is available
+import importlib.util
```
gtopper: Pretty sure the comment should be on the check and not the import.

mlrun/feature_store/api.py
```diff
@@ -91,7 +91,7 @@ def _features_to_vector_and_check_permissions(features, update_stats):
 
 
 def _get_offline_features(
```
assaf758: please move these private functions to feature_vector.py

assaf758 (changes requested):
looks good - one minor comment.

gtopper (changes requested):
Since this is no longer just about tests, please update the title and description accordingly.

PR #7413, opened 2025-03-10
[Model Monitoring] Support retrieving a single model endpoint by name, without the need for a id or function

Eyal-Danieli:
Following the changes introduced here https://github.com/mlrun/mlrun/pull/7341, we are no longer store the function name and tag if it's not connected to a serving function. This means that when a mep is generated through the batch infer process, the function details are not stored in the mep record. As a result, users cannot retrieve the mep based on these details and must provide either the endpoint id or the mep name.

In this PR, we ensure that if the user provides the mep name, the result will be returned only if a single record is found. If multiple records are found, the user will be asked to provide the endpoint id. 

Additionally, this PR updates the `get_or_create_model_endpoint` API to support the `feature_analysis` flag, and the batch inference tutorial has been aligned accordingly.

https://iguazio.atlassian.net/browse/ML-9517


server/py/framework/db/sqldb/db.py
```diff
@@ -5143,7 +5142,13 @@ def _get_mep_latest_instance(
...
             return query
 
-        return query.first()  # Use `.first()` instead of `.one_or_none()` for safety
+        if query.count() > 1:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "More than one instance found for the given Model Endpoint."
+                "Please provide endpoint_id or function name and tag to get a specific instance."
+            )
+
+        return query.one_or_none()
```
quaark, line 5151: I suggest not implementing like this, as you are running 2 queries against the DB which is less efficient. Since `one_or_none()` already implements this for you, you can use it and then compare with `None`. Or better yet use `one()` and catch the exception if it raises one, then you can raise your exception from it.

quaark: Actually I'm wrong as you want to support 0, 1 and an exception on more than 1. So ignore my comment :) 
Or you can list and do the count in python to still save the extra query

PR #7414, opened 2025-03-10
[Artifacts] Update unit test for artifact deletion data failure

moranbental:
This fix is based on the comment from this PR: https://github.com/mlrun/mlrun/pull/7406#discussion_r1986399428, which mentioned missing unit tests to ensure that any changes in the code will fail in the CI.

The fix involves updating the unit test to verify the following scenarios when deleting an artifact, and the deletion of artifact data fails:

- If the `deletion_strategy` is data_force, the artifact should still exist in the database.
- If the `deletion_strategy` is data_optional, the artifact should not exist in the database.

Also, in this PR, I updated the `patch_remote` script so that if patching fails to push images, the patch will fail.

### DIFF
--- a/automation/patch_igz/patch_remote.py
+++ b/automation/patch_igz/patch_remote.py
@@ -293,6 +293,7 @@ class MLRunPatcher:
                     future.result()
                 except Exception as exc:
                     logger.error(f"Error pushing image {image}: {exc}")
+                    raise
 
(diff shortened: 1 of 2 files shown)


alonmr (approved):
🔥

liranbg (approved):

PR #7415, opened 2025-03-10
[Project] Fix workflow sync functions engine test

alonmr:
Test should check if engine starts with remote and not equals to remote

https://iguazio.atlassian.net/browse/ML-9536


liranbg (approved):

GiladShapira94 (approved):

yaelgen (approved):
LGTM

PR #7417, opened 2025-03-10
[Golang] Use single `go.mod`

liranbg:
Simplifying the work against mlrun's go eco system


server/go/go.mod
```diff
@@ -1,18 +1,18 @@
-module github.com/mlrun/log-collector
+module github.com/mlrun
```
TomerShor, line 1: Shouldnt it be `github.com/mlrun/mlrun`?

liranbg: not really at this point

server/go/.golangci.yml
```diff
@@ -55,7 +55,9 @@ linters-settings:
   gci:
     sections:
       - standard
-      - prefix(github.com/mlrun/mlrun)
+      - prefix(github.com/mlrun/framework)
+      - prefix(github.com/mlrun/proto)
+      - prefix(github.com/mlrun/services)
```
alonmr, line 60: what's the reason to separate the import groups of internal mlrun packages?

alonmr (approved):
Looks good 👍 minor question

server/go/.golangci.yml
liranbg: just to be able to differentiate between the components as they suppose to be orthogonal

PR #7423, opened 2025-03-11
[Artifacts] Fix order by when listing artifacts

moranbental:
When listing artifacts, the object sorting is sometimes inconsistent. We sort the artifacts by the updated date in descending order. However, there was a bug where if the dates were very close to each other, they were not sorted correctly.

For example, `2025-03-07 21:01:33.773940` should come before `2025-03-07 21:01:33.773713`, but it was appearing in the opposite order.

The issue was that the sorting was based on the "updated" column, which was saved in the table as a timestamp with fsp=3. After sorting, we were returning the updated date from the "updated" field in the full_object, which was saved as a timestamp with fsp=6.

**Note**: In SQLite, the `updated` and `created` columns return timestamps with fsp=6.

Also, if the timestamps are the same, we need a secondary field to order by. That's why we are ordering by artifact ID.


server/py/framework/db/sqldb/db.py
```diff
@@ -857,6 +857,16 @@ def list_artifacts(
...
+            # These fields are saved in full_object as timestamps with fsp=6, while the corresponding columns
+            # in the database have fsp=3. Since 'ORDER BY' is applied to the column, we return the value from the column
+            # (not from the full_object) to ensure the ordering is correct.
+            # In SQLite, the updated and created columns return timestamps with fsp=6.
+            artifact_struct["metadata"]["updated"] = artifact.updated.strftime(
+                "%Y-%m-%d %H:%M:%S.%f"
+            )
+            artifact_struct["metadata"]["created"] = artifact.created.strftime(
+                "%Y-%m-%d %H:%M:%S.%f"
+            )
```
liranbg: why do you need to set the format explicitly? and also you omit the `timezone` by that.

liranbg (changes requested):

server/py/framework/db/sqldb/db.py
moranbental: Previously, the updated and created fields were returned as strings with a specific format (as they were saved in the full object). Now, the values that are returned are datetime objects. If I don't handle this, it will break compatibility

mlrun/utils/helpers.py
```diff
@@ -1371,6 +1371,20 @@ def has_timezone(timestamp):
...
+        dt = dt.replace(tzinfo=timezone.utc)
+
+    # Convert to UTC timezone
+    with_timezone = dt.astimezone(timezone.utc)
+
+    # Format as requested
+    formatted_time = with_timezone.strftime("%Y-%m-%d %H:%M:%S.%f%z")
+
+    # Replace the "+" sign with the ":"
+    return formatted_time[:-2] + ":" + formatted_time[-2:]
```
liranbg: on python 3.12 you could use `%:z`. so this thingy is temporary until python 3:12 become minimal version.
in addition to that, you dont replace the "+" - you add ":" between the hours and minutes of timezone

also, your function overrides datetime timezone. for when datetime is naive - I can understand. but why for other use cases?

liranbg (changes requested):

server/py/framework/db/sqldb/db.py
```diff
@@ -857,6 +857,17 @@ def list_artifacts(
         artifacts = ArtifactList()
         for artifact, artifact_tag in artifact_records:
             artifact_struct = artifact.full_object
+            # These fields are saved in full_object as timestamps with fsp=6, while the corresponding columns
+            # in the database have fsp=3. Since 'ORDER BY' is applied to the column, we return the value from the column
+            # (not from the full_object) to ensure the ordering is correct.
+            # In SQLite, the updated and created columns return timestamps with fsp=6.
```
alonmr, line 863:
```suggestion
            # In MySQL, the updated and created columns return timestamps with fsp=6.
```

server/py/framework/db/sqldb/db.py
```diff
@@ -857,6 +857,17 @@ def list_artifacts(
...
+            artifact_struct["metadata"]["updated"] = mlrun.utils.format_datetime(
```
alonmr: I believe we should do this for get artifact as well for consistency

moranbental: correct

liranbg (approved):

PR #7425, opened 2025-03-11
[Model Monitoring] Request only necessary metrics in `list_model_endpoints`

gtopper:
[ML-9527](https://iguazio.atlassian.net/browse/ML-9527)

Verified TSDB ✅ 
Verified TDEngine ✅ 

[ML-9527]: https://iguazio.atlassian.net/browse/ML-9527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/db/base.py
```diff
@@ -738,6 +738,7 @@ def list_model_endpoints(
         top_level: bool = False,
         uids: Optional[list[str]] = None,
         latest_only: bool = False,
+        metrics: Optional[list[str]] = None,
```
Eyal-Danieli: maybe instead of managing 2 different params, we can have only `tsdb_metrics` that for example can be either a list or a str. if it's `*` we can return all tsdb metrics, otherwise we return the provided metrics in the list (or a single metric if the user provides a str). If it's none (by default) we don't return any tsdb metrics.

mlrun/db/httpdb.py
```diff
@@ -3788,6 +3789,7 @@ def list_model_endpoints(
         :param top_level:       Whether to return only top level model endpoints.
         :param uids:            A list of unique ids to filter by.
         :param latest_only:     Whether to return only the latest model endpoint version.
+        :param metrics:         A list of metrics to add. Defaults to all metrics.
```
Eyal-Danieli: If we keep two different params, please add here that this is relevant only when `tsdb_metrics=True`

mlrun/projects/project.py
```diff
@@ -3833,6 +3834,7 @@ def list_model_endpoints(
         :param end:                       The end time to filter by. Corresponding to the `created` field.
         :param top_level: if true will return only routers and endpoint that are NOT children of any router
         :param uids: if passed will return a list `ModelEndpoint` object with uid in uids
+        :param metrics:             A list of metrics to add. Defaults to all metrics.
```
Eyal-Danieli: same here

Eyal-Danieli (commented):
nice, IMO we should consider combine `tsdb_metrics` and `metrics`. In addition, we should also adjust the `get_model_endpoint` to support different tsdb metrics as well to keep consistency.

mlrun/model_monitoring/controller.py
```diff
@@ -673,7 +673,9 @@ def push_regular_event_to_controller_stream(self) -> None:
         """
         logger.info("Starting monitoring controller chief")
         applications_names = []
-        endpoints = self.project_obj.list_model_endpoints(tsdb_metrics=True).endpoints
+        endpoints = self.project_obj.list_model_endpoints(
+            tsdb_metrics="last_request",
```
Eyal-Danieli, line 677: cool cool cool

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -1128,27 +1146,15 @@ def add_metric(
...
-            drift_status_res,
-        )
-
+        for metric_name, result in metric_name_to_result.items():
+            add_metric(
+                metric_name,
+                metric_name_to_function_and_column_name[metric_name][1],
+                result,
+            )
         self._enrich_mep_with_last_request(
```
Eyal-Danieli: unlike `tdengine`, here we always enrich the mep with the `last_request` value, even if that metric wasn't part of the provided `metrics`. I guess it's fine though we could be more consistent by adding an `if` clause to ensure the user actually requested the `last_request` metric before including it.

mlrun/projects/project.py
```diff
@@ -3795,7 +3795,7 @@ def list_model_endpoints(
         top_level: bool = False,
         uids: Optional[list[str]] = None,
         latest_only: bool = False,
-        tsdb_metrics: bool = True,
+        tsdb_metrics: Optional[str] = None,
```
Eyal-Danieli, line 3798: missing from the docstring (it was already missing before your PR, but you can copy and paste from what you’ve already written)

server/py/services/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -1365,6 +1368,9 @@ async def _add_basic_metrics(
                                         be filled with the relevant basic metrics.
         :param project:                The name of the project.
```
Eyal-Danieli:
```suggestion
```

Eyal-Danieli (commented):
few small comments but because of the API changes, you also need to go through the mm tests and validate that nothing breaks when we call `list_model_endpoints()`. For example I guess that `TestVotingModelMonitoring` will fail because it assumes that `tsdb_metrics=True`.

Eyal-Danieli (approved):

gtopper:
Changing it back to a draft pending:
* Conflict resolution with #7647 which will be merged first
* MM tests fixed and verified
* Potentially change to tsdb metrics being off by default (the status quo is a strange mix where they're off by default internally but on by default via SDK/API)

server/py/services/api/api/endpoints/model_endpoints.py
```diff
@@ -229,7 +229,7 @@ async def list_model_endpoints(
     start: Optional[datetime] = None,
     end: Optional[datetime] = None,
     top_level: bool = Query(False, alias="top-level"),
-    tsdb_metrics: bool = True,
+    tsdb_metrics: Optional[str] = None,
```
theSaarco, line 232:
```suggestion
    tsdb_metrics: Optional[str] = Query(None, alias="tsdb-metrics"),
```
The name of the parameter is invalid. You should not use underscores.

server/py/services/api/api/endpoints/model_endpoints.py
```diff
@@ -239,14 +239,15 @@ async def list_model_endpoints(
...
+    :param names:           The model endpoints names.
     :param model_name:      The model name.
     :param function_name:   The function name.
     :param function_tag:    The function tag.
     :param labels:          The labels of the model endpoint.
     :param start:           The start time to filter by.Corresponding to the `created` field.
     :param end:             The end time to filter by. Corresponding to the `created` field.
-    :param tsdb_metrics:    Whether to include metrics from the time series DB.
+    :param tsdb_metrics:    Comma separated list of metric names to add. Defaults to all available metrics. To disable,
+                            set to "false".
```
theSaarco, line 250: This is confusing logic, though I don't see a good way to have consistent logic with a single parameter.
I think it's better to be verbose here and separate to 2 params:
* `with-tsdb-metrics` - default to `True`.
* `tsdb-metrics` - if existing, this is the list of metrics. If `None` (the default) should behave according to the `with-tdsb-metrics` value, meaning either all metrics or no metrics.

server/py/services/api/api/endpoints/model_endpoints.py
```diff
@@ -229,7 +229,7 @@ async def list_model_endpoints(
     start: Optional[datetime] = None,
     end: Optional[datetime] = None,
     top_level: bool = Query(False, alias="top-level"),
-    tsdb_metrics: bool = True,
+    tsdb_metrics: Optional[str] = None,
     uids: list[str] = Query(None, alias="uid"),
     latest_only: bool = False,
```
theSaarco, line 234:
```suggestion
    latest_only: bool = Query(False, alias="latest-only"),
```

theSaarco (changes requested):
The API is not conforming with our standards for REST APIs, in terms of API param names. Also, I'm not very happy with the logic around `"false"` value.
Note that there are other parameters in the API that exist today and also are not conforming to our standards.

server/py/services/api/api/endpoints/model_endpoints.py
davesh0812: @theSaarco do we want to do some of this changes in 1.8.0 in order to avoid BC issues?

maybe we can changed `latest_only` ->  `latest-only`
and `tsdb_metrics` -> `with-tsdb-metrics`

gtopper:
Based on my discussion with @theSaarco, this PR won't make it into 1.8.0, and therefore we will need to respect BWC. We will add the list parameter as well as aliases.

gtopper:
Closing it for now. API realignment is happening in #7668. To be followed by [ML-9793](https://iguazio.atlassian.net/browse/ML-9793), the PR for which will draw on the code in this PR.

[ML-9793]: https://iguazio.atlassian.net/browse/ML-9793?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7431, opened 2025-03-12
[Model Monitoring] Fix limit controller unclosed windows 

royischoss:
https://iguazio.atlassian.net/browse/ML-9532
This fix includes a check in `MonitoringApplicationController::_should_monitor_endpoint` if the minimum last_analyzed is behind in time for more than five base periods -> we should push event in the stream allowing the controller to close the open windows, this prevent overhead when invoke is triggered after long time without any data in mm stream. 
Note: This change allow max of five opened windows, because of the parquet flash time we take in calculations we always keep the last windows open if the parquet flash time didn't passed.
Another small fix in `BASE_PERIOD_LOOKUP_TABLE` adding 2 min base period mapping to 2 min cron trigger


server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -69,7 +69,7 @@
     Path(mlrun.model_monitoring.applications.__file__).parent
     / "histogram_data_drift.py"
 )
-BASE_PERIOD_LOOKUP_TABLE = {1: 1, 20: 3, 60: 5, 120: 10, float("inf"): 20}
+BASE_PERIOD_LOOKUP_TABLE = {1: 1, 2: 2, 20: 3, 60: 5, 120: 10, float("inf"): 20}
```
davesh0812, line 72: 🍻

mlrun/model_monitoring/controller.py
```diff
@@ -265,9 +268,11 @@ def _get_model_monitoring_access_key() -> Optional[str]:
             access_key = mlrun.mlconf.get_v3io_access_key()
         return access_key
 
-    @staticmethod
     def _should_monitor_endpoint(
-        endpoint: mlrun.common.schemas.ModelEndpoint, application_names: set
+        self,
+        endpoint: mlrun.common.schemas.ModelEndpoint,
+        application_names: set,
+        base_period_minutes: int,
```
davesh0812, line 275: I know we should do it before, but please have full explanation on this functionality (when we should or not monitor)

davesh0812 (commented):
Minor

davesh0812 (approved):

PR #7434, opened 2025-03-12
[Tests] Split server and client unit test runners

tomerm-iguazio:
Split unit tests to different jobs, server and non server.

[ML-9561](https://iguazio.atlassian.net/browse/ML-9561 )
This PR is the first step of adding coverage to CI, issue: https://github.com/mlrun/mlrun/pull/7492

Coverage is slowing down unit tests. This PR aims to reduce the running time by parallelizing the executions.

[ML-9561]: https://iguazio.atlassian.net/browse/ML-9561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/ci.yaml
```diff
@@ -84,19 +84,51 @@ jobs:
...
         "
 
-  tests:
-    name: Run Dockerized Tests
+  tests-3_12:
+    name: Run Dockerized Tests 3.12
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: ["3.9", "3.12"]
```
liranbg: I understand you split the tests to client/server but why did you remove the matrix? you can decide to run the matrix ["client", "server"] X ["python:3.9", "python:3.12"] - wouldnt it be DRY?

tomerm-iguazio: I do not want to split also 3.12, it is not our bottleneck:
We are going to add coverage **only to 3.9**, so I will be a waste of runner to split 3.12 unit tests too.

alonmr: what's the reason to not run coverage on python 3.12?

.github/workflows/ci.yaml
```diff
@@ -84,19 +84,51 @@ jobs:
...
+
+    - name: Run Dockerized tests
+      run: |
+        MLRUN_DOCKER_REGISTRY=ghcr.io/ \
+        MLRUN_DOCKER_CACHE_FROM_TAG=${{ steps.docker_cache.outputs.tag }} \
+        MLRUN_PYTHON_VERSION=3.12 \
+        make test-dockerized
+
+  tests-non-server-3_9:
+    name: Run Non-Server Dockerized Tests
```
alonmr:
```suggestion
  tests-client-3_9:
    name: Run Client Dockerized Tests
```

.github/workflows/ci.yaml
```diff
@@ -118,9 +150,47 @@ jobs:
...
+    - name: Install automation scripts dependencies
+      run: pip install -r automation/requirements.txt
+
+    - name: Run Dockerized tests
+      #  including one test from /tests, due to ML-9563 issue.
+      run: |
+        MLRUN_DOCKER_REGISTRY=ghcr.io/ \
+        MLRUN_DOCKER_CACHE_FROM_TAG=${{ steps.docker_cache.outputs.tag }} \
+        MLRUN_PYTHON_VERSION=3.9 \
+        UNIT_TEST_PATH="tests/feature-store/test_common.py::test_parse_feature_string_with_dot_in_feature_set_name server/" \
```
alonmr: This looks like you chose some random test just to make it work but it doesn't address the root cause.
Can you explain exactly what is missing to make it work without running one client test? There must be some way to import the fixtures needed in the server from the client tests.

Makefile
```diff
@@ -524,11 +526,17 @@ test-dockerized: build-test ## Run mlrun tests in docker container
 		-e MLRUN_PYTHON_VERSION=$(MLRUN_PYTHON_VERSION) \
 		-v /tmp:/tmp \
 		-v /var/run/docker.sock:/var/run/docker.sock \
-		$(MLRUN_TEST_IMAGE_NAME_TAGGED) make test
+		$(MLRUN_TEST_IMAGE_NAME_TAGGED) make test  UNIT_TEST_IGNORE_PATH=$(UNIT_TEST_IGNORE_PATH) \
+		UNIT_TEST_PATH="$(UNIT_TEST_PATH)"
 
 .PHONY: test
 test: clean ## Run mlrun tests
 	# TODO: Remove ignored tests for Python 3.11 compatibility with KFP 2
```
alonmr: the comment is related to the lines below and now its confusing because you added lines in between

Makefile
```diff
@@ -98,6 +98,8 @@ endif
 
 # Change to `--upgrade-package <package-name>` to upgrade only a specific package
 MLRUN_UV_UPGRADE_FLAG ?= --upgrade
+UNIT_TEST_PATH?=./
+UNIT_TEST_IGNORE_PATH?=
```
alonmr:
```suggestion
UNIT_TESTS_PATH ?= ./
UNIT_TESTS_IGNORE_PATH ?=
```

alonmr (changes requested):
I agree that removing the python matrix is not the best solution here, it creates code duplication. You could probably resolve the command you want by the python version with bash. Anyway I think the PR needs a more elaborative description to explain the motivation behind the change.

.github/workflows/ci.yaml
tomerm-iguazio: there are some tests that run only in 3.9.
Please view makefile-> .PHONY: test

tomerm-iguazio:
https://github.com/mlrun/mlrun/pull/7434#discussion_r2016331686
answer: https://github.com/mlrun/mlrun/pull/7434#discussion_r2016331686

server/conftest.py
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from tests.common_fixtures import (  # noqa: F401  # Linter incorrectly flags this as unused
+    config_test_base,
+    rundb_mock,
```
alonmr: unrelated to this PR but rundb_mock should not be used in server tests so I'll check if we can remove it

server/conftest.py
alonmr, line 1: please move file to `server/py/conftest.py` we don't want python files outside of `py` folder

Makefile
```diff
@@ -524,12 +524,12 @@ test-dockerized: build-test ## Run mlrun tests in docker container
...
 		-v /var/run/docker.sock:/var/run/docker.sock \
-		$(MLRUN_TEST_IMAGE_NAME_TAGGED) make test
+		$(MLRUN_TEST_IMAGE_NAME_TAGGED) make test  UNIT_TESTS_IGNORE_PATH="$(UNIT_TESTS_IGNORE_PATH)" \
+		UNIT_TESTS_PATH="$(UNIT_TESTS_PATH)"
 
 .PHONY: test
 test: clean ## Run mlrun tests
-	# TODO: Remove ignored tests for Python 3.11 compatibility with KFP 2
-	set -e ;\
+#   TODO: Remove ignored tests for Python 3.11 compatibility with KFP 2
```
alonmr: I see you moved your changes below so this section can be left unchanged

alonmr (changes requested):
Looks great! minor comments

server/conftest.py
tomerm-iguazio: `test_deploy_function_with_image_pull_secret` failed without it...

tomerm-iguazio: you can see the logs in 
https://github.com/mlrun/mlrun/actions/runs/14110771331/job/39528580862

tomerm-iguazio:
```
___ ERROR at setup of test_new_function_args_with_default_image_pull_secret ____
file /mlrun/server/py/services/api/tests/unit/test_launcher.py, line 144
  def test_new_function_args_with_default_image_pull_secret(rundb_mock):
E       fixture 'rundb_mock' not found
>       available fixtures: _session_faker, alembic_config, alembic_engine, alembic_runner, anyio_backend, anyio_backend_name, anyio_backend_options, api_config_test, api_url, app, async_client, cache, capfd, capfdbinary, caplog, capsys, capsysbinary, client, config_test_base, db, doctest_namespace, event_loop, faker, free_tcp_port, free_tcp_port_factory, free_udp_port, free_udp_port_factory, httpserver, httpserver_ipv4, httpserver_ipv6, httpserver_listen_address, httpserver_ssl_context, iguazio_client, k8s_secrets_mock, kfp_client_mock, make_httpserver, make_httpserver_ipv4, make_httpserver_ipv6, mock_project_follower_iguazio_client, mocked_k8s_helper, monkeypatch, prefix, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, requests_mock, service_config_test, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory, unused_tcp_port, unused_tcp_port_factory, unused_udp_port, unused_udp_port_factory, unversioned_client
>       use 'pytest --fixtures [testpath]' for help on them.
```

alonmr: yes I saw the test I was saying that we should change the test to not use this fixture because its a client mock. but that is for another PR

alonmr (approved):

PR #7435, opened 2025-03-12
[Model Monitoring] Resolved a memory leak issue triggered by `list_model_endpoints` calls

davesh0812:
Explanation of Changes

The previous implementation used `joinedload()` to eagerly load related objects (`function`, `model`, and `tags`). While effective for small datasets, it caused high memory usage and potential OOM issues when dealing with large queries due to cartesian product effects and unnecessary data loading.

This PR replaces `joinedload()` with `selectinload()`, which issues separate SELECT statements to load related data. `selectinload()` avoids cartesian products and reduces memory footprint by loading only what's needed, making it more efficient for large datasets.

I also added `load_only()` to fetch just the required fields, further minimizing unnecessary data retrieval. 

performance : `list_model_endpoints`on 4500 meps takes ±8 sec from SDK
Resolved : https://iguazio.atlassian.net/browse/ML-9524


quaark (commented):
Great optimization!
Could you add a concise explanation in code comments as well?
These complicated ORM queries can often be hard to understand (I only understood after reading the PR description)

assaf758 (approved):

PR #7436, opened 2025-03-13
[Datastore] Rename `TDEngineDatastoreProfile` -> `DatastoreProfileTDEngine`

gtopper:
In line with all other datastore profiles other than `ConfigProfile`, which I wasn't sure about (@alxtkr77?).


mlrun/datastore/datastore_profile.py
```diff
@@ -477,7 +477,7 @@ def url(self, subpath):
         return f"webhdfs://{self.host}:{self.http_port}{subpath}"
 
 
-class TDEngineDatastoreProfile(DatastoreProfile):
+class DatastoreProfileTDEngine(DatastoreProfile):
```
jond01, line 480: From the [Zen of Python](https://peps.python.org/pep-0020/):

> Readability counts.

The existing name is readable and corresponds to how one would call this, while the other is not.
Therefore, I don't see the point in this change except that it aligns with the mis-convention in this module. The other classes should be changed in the proper order.
https://peps.python.org/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds

assaf758: Actually, I would have chosen a different name altogether if we were to reimplement this feature from scratch - Kafka is not a datastore etc  :-) 
Looking around at some code-bases and thinking about it, I do agree with Jonathan that writing the type of the thing (TDEngine) before the thing (DatastoreProfile) is the correct way to go.
But as to this suggested PR, most profiles are already out from 1.6/1.7, so for internal consistency and for practical reasons it makes sense to me to change the TDEngine profile to DatastoreProfileTDEngine.

assaf758 (approved):

PR #7449, opened 2025-03-16
[Tests] Fix vectordb and alarms system tests

alxtkr77:
- Checks if LangChain is not installed before running vectordb tests
- Fixes the alarm system tests to support the latest model monitoring datastore profile convention

### DIFF
--- a/tests/system/datastore/test_vectorstore.py
+++ b/tests/system/datastore/test_vectorstore.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 #
 
+# Check if langchain_community is available
+import importlib.util
 import os
 import random
 import string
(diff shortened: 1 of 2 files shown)


tests/system/datastore/test_vectorstore.py
```diff
@@ -29,6 +30,17 @@
 )
 from tests.system.base import TestMLRunSystem
 
+LANGCHAIN_AVAILABLE = (
+    importlib.util.find_spec("langchain") is not None
+    and importlib.util.find_spec("langchain_community") is not None
```
assaf758, line 35: Shouldn't be an OR?

assaf758 (approved):

PR #7453, opened 2025-03-17
[Runs] Use SQL NOW() function to calculate end time

alonmr:
https://iguazio.atlassian.net/browse/ML-9572

Using `NOW()` should help us mitigate a race condition where the push notifications cycle is listing runs and is not getting a runs that was just updated with end time. This should minimize the delta between the end time and the commit time.


server/py/framework/db/sqldb/db.py
```diff
@@ -594,9 +602,9 @@ def _update_run_end_time(run: Run, run_dict: dict, end_time: Optional[str] = Non
             and not run.end_time
         ):
             if end_time is None:
-                end_time = datetime.now(timezone.utc)
+                # Ensures fsp 6 for MySQL NOW() to includes microseconds
+                end_time = func.now(6)
```
TomerShor: Is this TZ-aware?

liranbg: NOW() follow its system timezone. mysql has utc by default. I dont think it is necessary to explicitly state `utc` - as for consistency with other datetime being generated automatically.

@alonmr can you use kwarg to set the fsp rather than magic number?

quaark (approved):

server/py/framework/db/sqldb/db.py
alonmr: it is UTC but it doesn't include the tzinfo so I'll add it manually with helpers we have

alonmr: @liranbg after I moved it to integ tests it turns out its not needed so I removed it
keep in mind it will not work for sqlite

liranbg (approved):

PR #7455, opened 2025-03-18
[Model Monitoring] Fix child function monitored when `set_tracking()` is applied

royischoss:
https://iguazio.atlassian.net/browse/ML-9153
Fix includes
1) Update spec.track_models when calling `ServingRuntime:set_tracking` and `ServingRuntime:add_child_function` Allowing to set the right monitoring regardless of the calling order. 
2) Fix checking background task state from V2ModelServer post_init if endpoint_type indicates it is not a child of a router ("flow")


mlrun/serving/v2_serving.py
```diff
@@ -145,8 +145,14 @@ def post_init(self, mode="sync", **kwargs):
                 feature.name for feature in self.model_spec.outputs
             ]
 
+        if (
+            kwargs.get("endpoint_type", mlrun.common.schemas.EndpointType.ROUTER)
```
davesh0812:
```suggestion
            kwargs.get("endpoint_type", mlrun.common.schemas.EndpointType.LEAF_EP)
```

davesh0812 (commented):
Can you please add some basic unit test?

davesh0812 (approved):

PR #7457, opened 2025-03-18
[Model Monitoring] Mask out v3io access key from functions YAML

alxtkr77:
https://iguazio.atlassian.net/browse/ML-9443

### DIFF
--- a/server/py/services/api/crud/model_monitoring/deployment.py
+++ b/server/py/services/api/crud/model_monitoring/deployment.py
@@ -422,7 +422,10 @@ class MonitoringDeployment:
             else mlrun.mlconf.get_v3io_access_key()
         )
         kwargs = {
-            "access_key": access_key,
+            # ML-9443: Removing the access key from parameters.add_v3io_stream_trigger() automatically sets it to
+            # mlrun.model.Credentials.generate_access_key, causing Nuclio to replace the key with its V3IO_ACCESS_KEY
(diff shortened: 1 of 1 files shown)


server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -422,7 +422,10 @@ def _apply_and_create_v3io_source(
             else mlrun.mlconf.get_v3io_access_key()
         )
         kwargs = {
-            "access_key": access_key,
+            # ML-9443: Removing the access key from parameters.add_v3io_stream_trigger() automatically sets it to
+            # mlrun.model.Credentials.generate_access_key, causing Nuclio to replace the key with its V3IO_ACCESS_KEY
+            # environment variable. This change prevents the access key from being displayed in the ML function YAML.
+            # "access_key": access_key,
```
TomerShor: Why leave it as commented out and not just remove it?

I also think the `kwargs` dict is unnecessary here, you can simply pass the arguments to the function below, as it is not used again later: 
```python
        # Generate V3IO stream trigger
        function.add_v3io_stream_trigger(
            stream_path=stream_path,
            name=f"monitoring_{function_name}_trigger",
            worker_allocation_mode="static",
            max_workers=stream_args.v3io.num_workers,
        )
```

TomerShor (changes requested):

TomerShor (approved):

assaf758 (approved):

PR #7460, opened 2025-03-18
[Alerts] Add limit to list alert configs

TomerShor:
When we have many (~20k) alert configs in the system, listing them all causes the API behave unexpectedly.
While we will want to implement pagination for alert configs, as a current mitigation this PR adds a default limit to the list alert configs endpoint - 2k .

To avoid adding a query param to the endpoint that will be deprecated when pagination is implemented, the new param `page-size` is basically implementing the `limit` param.
On the client-side, we still use `limit` and `offset` arguments, that are translated to `page-size` and `offset`.

The 2k limit is configurable via the mlrun config.

Additionally, change the `list_alert_configs` response to return a dictionary that contains a list, instead of only a list, to prepare for pagination.

https://iguazio.atlassian.net/browse/ML-9576

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -4826,6 +4826,7 @@ class HTTPRunDB(RunDBInterface):
 
         :param project: The project name.
         :param limit: The maximum number of alerts to return.
+            Defaults to `mlconf.alerts.default_list_alert_configs_limit` if not provided.
         :param offset: The number of alerts to skip.
 
         :returns: All the alerts objects of the project.
@@ -4834,7 +4835,7 @@ class HTTPRunDB(RunDBInterface):
         endpoint_path = f"projects/{project}/alerts"
         error_message = f"get alerts {project}/alerts"
         params = {}
-        # TODO: remove limit and offset when pagination is implemented
+        # TODO: Deprecate limit and offset when pagination is implemented
         if limit:
             params["page-size"] = limit
         if offset:
@@ -4843,7 +4844,7 @@ class HTTPRunDB(RunDBInterface):
             "GET", endpoint_path, error_message, params=params
         ).json()
         results = []
-        for item in response:
+        for item in response.get("alerts", []):
             results.append(AlertConfig(**item))
         return results
 
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -4818,18 +4818,30 @@ class HTTPRunDB(RunDBInterface):
         response = self.api_call("GET", endpoint_path, error_message)
         return AlertConfig.from_dict(response.json())
 
-    def list_alerts_configs(self, project="") -> list[AlertConfig]:
+    def list_alerts_configs(
+        self, project="", limit: Optional[int] = None, offset: Optional[int] = None
+    ) -> list[AlertConfig]:
         """
         Retrieve list of alerts of a project.
 
         :param project: The project name.
+        :param limit: The maximum number of alerts to return.
+        :param offset: The number of alerts to skip.
 
         :returns: All the alerts objects of the project.
         """
         project = project or config.default_project
         endpoint_path = f"projects/{project}/alerts"
         error_message = f"get alerts {project}/alerts"
-        response = self.api_call("GET", endpoint_path, error_message).json()
+        params = {}
+        # TODO: remove limit and offset when pagination is implemented
+        if limit:
+            params["page-size"] = limit
+        if offset:
+            params["offset"] = offset
+        response = self.api_call(
+            "GET", endpoint_path, error_message, params=params
+        ).json()
         results = []
         for item in response:
             results.append(AlertConfig(**item))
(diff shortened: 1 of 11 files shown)


mlrun/db/httpdb.py
```diff
@@ -4818,18 +4818,30 @@ def get_alert_config(self, alert_name: str, project="") -> AlertConfig:
...
+        :param offset: The number of alerts to skip.
 
         :returns: All the alerts objects of the project.
         """
         project = project or config.default_project
         endpoint_path = f"projects/{project}/alerts"
         error_message = f"get alerts {project}/alerts"
-        response = self.api_call("GET", endpoint_path, error_message).json()
+        params = {}
+        # TODO: remove limit and offset when pagination is implemented
```
alonmr:
```suggestion
        # TODO: Deprecate limit and offset when pagination is implemented
```

mlrun/db/httpdb.py
```diff
@@ -4818,18 +4818,30 @@ def get_alert_config(self, alert_name: str, project="") -> AlertConfig:
...
 
-    def list_alerts_configs(self, project="") -> list[AlertConfig]:
+    def list_alerts_configs(
+        self, project="", limit: Optional[int] = None, offset: Optional[int] = None
+    ) -> list[AlertConfig]:
         """
         Retrieve list of alerts of a project.
 
         :param project: The project name.
+        :param limit: The maximum number of alerts to return.
```
alonmr, line 4828:
```suggestion
        :param limit: The maximum number of alerts to return. Defaults to mlconf.alerts...
```

alonmr (commented):
Nice! Minor comments

mlrun/projects/project.py
```diff
@@ -5034,14 +5034,19 @@ def get_alert_config(self, alert_name: str) -> AlertConfig:
...
         return db.get_alert_config(alert_name, self.metadata.name)
 
-    def list_alerts_configs(self) -> list[AlertConfig]:
+    def list_alerts_configs(
+        self, limit: Optional[int] = None, offset: Optional[int] = None
+    ) -> list[AlertConfig]:
         """
         Retrieve list of alerts of a project.
 
+        :param limit: The maximum number of alerts to return.
```
moranbental, line 5043:
```suggestion
        :param limit: The maximum number of alerts to return. Defaults to mlconf.alerts...
```

server/py/services/alerts/main.py
```diff
@@ -145,6 +146,8 @@ async def list_alerts(
         self,
         request: fastapi.Request,
         project: str,
+        page_size: int,
```
moranbental:
```suggestion
        page_size: typing.Optional[int]
```

server/py/framework/db/sqldb/db.py
```diff
@@ -6256,6 +6258,8 @@ def list_alerts(
         ).add_entity(AlertState)
 
         query = self._filter_query_by_resource_project(query, AlertConfig, project)
+        query = self._paginate_query(query, offset, limit)
```
moranbental, line 6270: If you are using a limit, it should be ordered by something

moranbental (changes requested):
Why not implement pagination?

server/py/framework/db/sqldb/db.py
TomerShor: Adding ordering by id, as created/updated are not saved as a db column.

TomerShor:
> Why not implement pagination?

@moranbental This is needed for 1.8.0 and pagination will not be implemented in the UI or tested properly in the timeline.

server/py/services/alerts/tests/unit/api/test_alerts.py
```diff
@@ -198,6 +198,52 @@ def test_delete_alerts(
...
+            assert resp.status_code == HTTPStatus.OK.value
+
+        for params, expected_length in [
+            # limit to 5
+            ({"page-size": 5}, 5),
+            # limit to 5, offset 2
+            ({"page-size": 5, "offset": 2}, 5),
+            # limit to 5, offset 6 - not enough alerts
+            ({"page-size": 5, "offset": 8}, 2),
+            # no limit at all
```
yaelgen:
```suggestion
            # no explicit limit, should return default limit (8)
```

server/py/services/alerts/tests/unit/api/test_alerts.py
```diff
@@ -198,6 +198,52 @@ def test_delete_alerts(
...
+                    project=project
+                ),
+            )
+            resp = client.put(
+                STORE_ALERTS_PATH.format(project=project, name=alert_name),
+                json=alert_config.dict(),
+            )
+            assert resp.status_code == HTTPStatus.OK.value
+
+        for params, expected_length in [
```
yaelgen: I would suggest using a fixture and parametrize for this instead of a for loop

yaelgen (approved):
LGTM, added a suggestion

moranbental (approved):
LGTM

PR #7041, opened 2025-01-01
[Model Monitoring] Limit archived models to 5 and retrieve all MEPs from v3io-tsdb if the count exceeds 200.

davesh0812:
Fix for [ML-8886](https://iguazio.atlassian.net/browse/ML-8886)

1. Limit the number of archived endpoint versions to the 5 most recently created.  
2. When querying v3io-tsdb for more than 200 model endpoints, automatically ignore the filter and retrieve all model endpoints from the tsdb table.

[ML-8886]: https://iguazio.atlassian.net/browse/ML-8886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7463, opened 2025-03-19
[Model Monitoring] Fix querying `last_request` with 150+ model endpoints

Eyal-Danieli:
Following #7459 , we are using `V3IOClient` to retrieve the `last_request` values from the KV table. When providing more than 177 model endpoints, the filter expression becomes too large for the query. This PR addresses the issue using the same mechanism introduced here - #7041 , meaning if we query more than 150 model endpoints then we will query the last request for all model endpoints in a single query. 

https://iguazio.atlassian.net/browse/ML-9602

### DIFF
--- a/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
+++ b/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
@@ -916,23 +916,6 @@ class V3IOTSDBConnector(TSDBConnector):
 
         return last_request_timestamps
 
-        # try:
-        #     res = self.v3io_client.kv.new_cursor(
-        #         container=self.container,
-        #         table_path=self.last_request_table,
-        #         filter_expression=filter_expression,
-        #     ).all()
-        # except Exception as e:
-        #     logger.warning(
-        #         "Failed to get last request timestamp from V3IO KV table.",
-        #         err=mlrun.errors.err_to_str(e),
-        #         project=self.project,
-        #         table=self.last_request_table,
-        #     )
-        #     return {}
-
-        # return {d["__name"]: d["last_request_timestamp"] for d in res}
-
     def get_drift_status(
         self,
         endpoint_ids: Union[str, list[str]],
--- a/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
+++ b/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
@@ -33,8 +33,12 @@ _TSDB_BE = "tsdb"
 _TSDB_RATE = "1/s"
 _CONTAINER = "users"
 
-V3IO_FRAMESD_MEPS_LIMIT = 200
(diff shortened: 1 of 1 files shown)


royischoss (approved):
LGTM

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -880,33 +885,50 @@ def get_last_request(
...
-                filter_expression=filter_expression,
-            ).all()
+            if len(endpoint_ids) > V3IO_CLIENT_MEPS_LIMIT:
+                logger.warning(
+                    "The number of endpoint ids exceeds the v3io-engine filter-expression limit, "
+                    "retrieving last request for all the model endpoints from the KV table.",
+                    limit=V3IO_CLIENT_MEPS_LIMIT,
+                    amount=len(endpoint_ids),
+                )
+
```
assaf758, line 903: This behavior is by-design, so no reason to treat it as `warning` severity (morel like `debug`).
But I actually think we should remove this log altogether - the flow here is very regular, the controller and UI will use it all the time, and it does not add useful info (but it does spam the logs)

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -880,33 +885,50 @@ def get_last_request(
...
+                res = self.v3io_client.kv.new_cursor(
+                    container=self.container,
+                    table_path=self.last_request_table,
+                ).all()
+                last_request_timestamps.update(
+                    {d["__name"]: d["last_request_timestamp"] for d in res}
+                )
+            else:
+                filter_expression = " OR ".join(
```
assaf758, line 912: most of the code is shared between the `if` and `else` clauses - only filter_expression manipulation should remain in the if-else clause

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -33,7 +33,12 @@
 _TSDB_RATE = "1/s"
 _CONTAINER = "users"
 
-V3IO_MEPS_LIMIT = 200
+V3IO_FRAMESD_MEPS_LIMIT = (
+    200  # Maximum number of model endpoints per single request when using V3IO Frames
+)
+V3IO_CLIENT_MEPS_LIMIT = (
+    150  # Maximum number of model endpoints per single request when using V3IO Client
+)
```
assaf758, line 41: Suggest to rename to make it clear :
```suggestion
V3IO_GET_ITEMS_FILTER_MEP_KEYS_MAX_NUMBER = (
    150  # Maximum number of MEPs per request when used as filter-expression in v3io nosql table (filter expression max size is 8192 bytes) 
)
```

assaf758 (approved):
Approved.
I do have some comments for follow-up PR.

PR #7468, opened 2025-03-19
[Model Monitoring] Fix connection between offline model endpoint and function 

Eyal-Danieli:
Following #7341 , each model endpoint is connected to a function name and a tag. However, this was only applied to serving functions, where the hashing key is based on the `unversioned` tag along with the actual tag. As a result, creating a model endpoint through an offline process, such as `batch inference`, would result in a model endpoint without an associated function, causing various issues both in the UI and logic (such as https://iguazio.atlassian.net/browse/ML-9517, https://iguazio.atlassian.net/browse/ML-9492).
This PR resolves the issue by ensuring that a mep is always connected to the function, even for offline jobs. In such cases, the hashing key is not required when querying the function from the db.
In addition, in this PR we also updates the `get_or_create_model_endpoint` API to support `feature_analysis` flag and aligns the `batch-infer` tutorial accordingly.


mlrun/model_monitoring/api.py
```diff
@@ -50,25 +50,26 @@
 
 def get_or_create_model_endpoint(
     project: str,
+    model_endpoint_name: str,
     model_path: str = "",
-    model_endpoint_name: str = "",
     endpoint_id: str = "",
     function_name: str = "",
-    function_tag: str = "latest",
+    function_tag: str = "" or "latest",
```
jond01: Isn't it just `"latest"`? 🤔

daniels290813 (approved):

mlrun/model_monitoring/api.py
```diff
@@ -50,25 +50,26 @@
 
 def get_or_create_model_endpoint(
     project: str,
+    model_endpoint_name: str,
```
davesh0812, line 53: you changed the order.. it can create BC issues

davesh0812 (commented):
Nice, minor comment

mlrun/model_monitoring/api.py
Eyal-Danieli: it's a change in the order of the params because this keyword argument is now mandatory + ATM this api is only in internal use

davesh0812 (approved):

PR #7471, opened 2025-03-20
[Application] Don't log query params in reverse proxy

TomerShor:
Remove query params from log to prevent log forgery / injection.

Resolves https://github.com/mlrun/mlrun/security/code-scanning/38

**Note:** All lines were "changed" because formatting replaced 4 spaces with tabs, as is the convention in Go.

### DIFF
--- a/mlrun/runtimes/nuclio/application/reverse_proxy.go
+++ b/mlrun/runtimes/nuclio/application/reverse_proxy.go
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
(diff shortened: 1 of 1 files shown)


alonmr (commented):
different suggestion - what if we don't log the query params? currently the default log level is "info" anyway

TomerShor:
> different suggestion - what if we don't log the query params? currently the default log level is "info" anyway

@alonmr I suppose I'm ok with this option. Also the default container for `kubectl logs` is the sidecar and not the reverse proxy so even better.

alonmr (approved):

PR #7475, opened 2025-03-20
[Tests] Edit test_drift_detection_alert test

tomerm-iguazio:
`test_drift_detection_alert` test failed only in CI because it can not generate alerts properly.
The alert configs that returns from `create_model_monitoring_alert_configs` are empty.
I suggest that this might be due to not finding the generated endpoints, and we may need to extend the sleep time before creating the alerts.

Added prints for debugging.

**When running against the CI setup from desktop, the exception does not reproduce.**

https://github.com/mlrun/mlrun/actions/runs/13878681953/job/38834452186

error_log:
```
>       self.project.store_alert_config(alert_data[0])
E       IndexError: list index out of range

tests/system/alerts/test_alerts.py:415: IndexError
```


mlrun/projects/project.py
TomerShor, line 1: Please remove the changes from this file. This is bad practice due to:
1. This is not adding logs only for the test, this is adding it to the SDK
2. Generally - We don't use `print`, but we use the `logger` to print logs.

TomerShor (changes requested):

PR #7476, opened 2025-03-20
[SDK] Bump max python version and add kfp v2 requirement [feature/python-3-12]

alonmr:
https://iguazio.atlassian.net/browse/ML-9618
https://iguazio.atlassian.net/browse/ML-9619


setup.py
```diff
@@ -67,7 +67,7 @@ def version():
         "machine-learning",
         "experiment-tracking",
     ],
-    python_requires=">=3.9, <3.12",
+    python_requires=">=3.9, <3.13",
```
jond01, line 70: The PR title says "[SDK] Bump min python version ..." - should be **max**, right?

alonmr: right 👍

PR #7477, opened 2025-03-20
[Model Monitoring] Fix avoid sending the controller duplicate events

royischoss:
https://iguazio.atlassian.net/browse/ML-9634
Change in `MonitoringApplicationController:_should_monitor_endpoint` condition by checking if `last_request` , and `last_analyzed` timestamps were already sent as a closing window event if so we avoid of sending a duplicate event, lowering the load on the controller stream

Controller performance:
1000 Model endpoints:
with apps
 > 2025-03-20 14:27:48,119 [info] Starting monitoring controller chief
 > 2025-03-20 14:29:46,034 [info] Finish analyze for: {"timestamp":"2025-03-20T14:27:31.541999+00:00"}
 118 seconds


mlrun/common/schemas/model_monitoring/constants.py
```diff
@@ -282,6 +282,7 @@ class FileTargetKind:
     ERRORS = "errors"
     STATS = "stats"
     LAST_REQUEST = "last_request"
+    LAST_ANALYZED = "last_analyzed"
```
davesh0812: Not sure this is the place to put this one :)

mlrun/model_monitoring/controller.py
```diff
@@ -300,17 +316,33 @@ def _should_monitor_endpoint(
                 project=endpoint.metadata.project,
                 endpoint_id=endpoint.metadata.uid,
             ) as batch_window_generator:
-                base_period_seconds = base_period_minutes * _SECONDS_IN_MINUTE
-                if application_names != batch_window_generator.get_application_list():
+                current_time = mlrun.utils.datetime_now()
+                if (
+                    # Different application names, or last analyzed never updated
+                    application_names != batch_window_generator.get_application_list()
+                    or not batch_window_generator.get_min_last_analyzed()
```
davesh0812: please run `batch_window_generator.get_min_last_analyzed()` once in this method, to be sure we always work on the same one, like we are dong with the `current_time`

mlrun/model_monitoring/controller.py
```diff
@@ -281,11 +285,23 @@ def _should_monitor_endpoint(
             2.  first request exists
             3.  last request exists
             4.  endpoint_type is not ROUTER
-        if the four above conditions apply we require one of the three conditions to monitor:
+        if the four above conditions apply we require one of the two condition monitor:
             1.  never monitored the one of the endpoint applications meaning min_last_analyzed is None
-            2.  last request has a higher timestamp than the min_last_analyzed timestamp
-            3.  We didn't analyze one of the application for over than _MAX_OPEN_WINDOWS_ALLOWED windows
+            2.  min_last_analyzed stands in the condition for sending NOP event and this the first time regular event
+            is sent with the current last_request per endpoint.
```
davesh0812:
```suggestion
            2.  min_last_analyzed stands in the condition for sending NOP event and this the first time regular event
            is sent with the combination of  current last_request  & current last_analyzed  per endpoint.
```

mlrun/model_monitoring/db/_schedules.py
```diff
@@ -149,6 +162,80 @@ def get_min_timestamp(self) -> Optional[int]:
...
+        if endpoint_uid in self._schedules:
+            self._schedules[endpoint_uid][FileTargetKind.LAST_ANALYZED] = timestamp
+        else:
+            self._schedules[endpoint_uid] = {
+                FileTargetKind.LAST_ANALYZED: timestamp,
+                FileTargetKind.LAST_REQUEST: None,
+            }
+
+    def get_endpoint_last_analyzed(self, endpoint_uid: str) -> Optional[int]:
+        if endpoint_uid in self._schedules:
```
davesh0812, line 207:
```suggestion
        self._check_open_schedules()
        if endpoint_uid in self._schedules:
```

mlrun/model_monitoring/db/_schedules.py
```diff
@@ -149,6 +162,80 @@ def get_min_timestamp(self) -> Optional[int]:
...
 
 
+class ModelMonitoringSchedulesFileChief(ModelMonitoringSchedulesFileBase):
+    def __init__(self, project: str) -> None:
+        """
+        Initialize applications monitoring schedules file object.
+        The JSON file stores a dictionary of registered application name as key and Unix timestamp as value.
+        When working with the schedules data, use this class as a context manager to read and write the data.
+
+        :param project:     The project name.
```
davesh0812: please update

mlrun/model_monitoring/db/_schedules.py
```diff
@@ -149,6 +162,80 @@ def get_min_timestamp(self) -> Optional[int]:
...
+
+    def update_endpoint_last_analyzed(self, endpoint_uid: str, timestamp: int) -> None:
+        self._check_open_schedules()
+        if endpoint_uid in self._schedules:
+            self._schedules[endpoint_uid][FileTargetKind.LAST_ANALYZED] = timestamp
+        else:
+            self._schedules[endpoint_uid] = {
+                FileTargetKind.LAST_ANALYZED: timestamp,
+                FileTargetKind.LAST_REQUEST: None,
+            }
```
davesh0812: Why not to create one method to update both? 
Do we have a scenario which in we update only one of them?

server/py/services/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -837,7 +838,11 @@ def _delete_model_endpoint_monitoring_infra(
         for uid in uids:
             ModelMonitoringCurrentStatsFile(project=project, endpoint_id=uid).delete()
             ModelMonitoringDriftMeasuresFile(project=project, endpoint_id=uid).delete()
-            ModelMonitoringSchedulesFile(project=project, endpoint_id=uid).delete()
+            ModelMonitoringSchedulesFileEndpoint(
+                project=project, endpoint_id=uid
+            ).delete()
+
+        ModelMonitoringSchedulesFileChief(project=project).delete()
```
davesh0812: Not sure this is the right place to delete it… We’re currently calling this function on each MEP deletion.
I think it would be better to handle the deletion inside the `delete_model_endpoint_monitoring_resources` method, since that only runs when the user deletes the project.
Alternatively, you could consider deleting it when calling `disable_model_monitoring`, but that approach is a bit riskier.

davesh0812 (commented):
Few comments

mlrun/model_monitoring/db/_schedules.py
```diff
@@ -13,36 +13,29 @@
...
 from types import TracebackType
 from typing import Final, Optional
 
 import botocore.exceptions
 
-import mlrun.common.schemas
 import mlrun.errors
 import mlrun.model_monitoring.helpers
+from mlrun.common.schemas import ModelEndpoint
+from mlrun.common.schemas.model_monitoring.constants import FileTargetKind
```
jond01: See pt. 13 in the [CONTRIBUTING.md](https://github.com/mlrun/mlrun/blob/development/CONTRIBUTING.md#python-code-conventions) guide

mlrun/model_monitoring/db/_schedules.py
```diff
@@ -13,36 +13,29 @@
...
+        self._item = None
 
-        :param project:     The project name.
-        :param endpoint_id: The endpoint ID.
-        """
-        # `self._item` is the persistent version of the monitoring schedules.
-        self._item = mlrun.model_monitoring.helpers.get_monitoring_schedules_data(
-            project=project, endpoint_id=endpoint_id
-        )
+    def _post_init(self):
```
jond01: Frankly, I do not like the `_post_init` style. It is confusing, somewhat ugly, and in most cases, unnecessary.
In addition, it's a bad practice to define attributes outside of the constructor:
https://stackoverflow.com/a/38378757/27658872

mlrun/model_monitoring/db/_schedules.py
```diff
@@ -13,36 +13,29 @@
...
     INITIAL_CONTENT = json.dumps(DEFAULT_SCHEDULES)
     ENCODING = "utf-8"
 
-    def __init__(self, project: str, endpoint_id: str) -> None:
-        """
-        Initialize applications monitoring schedules file object.
-        The JSON file stores a dictionary of registered application name as key and Unix timestamp as value.
-        When working with the schedules data, use this class as a context manager to read and write the data.
+    def __init__(self):
+        self._item = None
```
jond01: What's the point of that attribute initialization? You initialize it afterward.
This is a "code smell" IMO - see the previous comment

mlrun/model_monitoring/db/_schedules.py
```diff
@@ -13,36 +13,29 @@
...
-import mlrun.common.schemas
 import mlrun.errors
 import mlrun.model_monitoring.helpers
+from mlrun.common.schemas import ModelEndpoint
+from mlrun.common.schemas.model_monitoring.constants import ScheduleChiefFields
 from mlrun.utils import logger
 
 
-class ModelMonitoringSchedulesFile(AbstractContextManager):
+class ModelMonitoringSchedulesFileBase(AbstractContextManager, ABC):
```
jond01, line 29: If you don't declare any abstract methods in this class - it's not truly an ABC:

```suggestion
class ModelMonitoringSchedulesFileBase(AbstractContextManager):
```

mlrun/model_monitoring/db/_schedules.py
```diff
@@ -149,6 +162,79 @@ def get_min_timestamp(self) -> Optional[int]:
...
+                last_request
+            )
+            self._schedules[endpoint_uid][ScheduleChiefFields.LAST_ANALYZED] = (
+                last_analyzed
+            )
+        else:
+            self._schedules[endpoint_uid] = {
+                ScheduleChiefFields.LAST_REQUEST: last_request,
+                ScheduleChiefFields.LAST_ANALYZED: last_analyzed,
+            }
```
jond01: Isn't it pretty much the same? You can stay with just the else block, if I get it right.

jond01 (changes requested):
I didn't review it all yet, but I have some comments that need to be addressed

mlrun/common/schemas/model_monitoring/constants.py
```diff
@@ -289,6 +289,11 @@ class ModelMonitoringMode(StrEnum):
     disabled = "disabled"
 
 
+class ScheduleChiefFields(MonitoringStrEnum):
```
jond01: If you don't use `.list()`, the following is clearer:
```suggestion
class ScheduleChiefFields(StrEnum):
```

mlrun/model_monitoring/db/_schedules.py
royischoss: Left overs thanks

mlrun/model_monitoring/helpers.py
```diff
@@ -586,16 +586,43 @@ def _get_monitoring_schedules_folder_path(project: str) -> str:
...
+        _get_monitoring_schedules_file_chief_path(project=project)
+    )
+
+
+def _get_monitoring_schedules_file_chief_path(
+    *,
+    project: str,
+) -> str:
+    return os.path.join(
+        _get_monitoring_schedules_folder_path(project), f"{project}.json"
```
jond01, line 627: Is there a way to differentiate between a project file and a MEP file in the case project_name == mep_uid?
There's a potential (although very unlikely) collision here

mlrun/model_monitoring/db/_schedules.py
```diff
@@ -149,6 +165,68 @@ def get_min_timestamp(self) -> Optional[int]:
         return min(self._schedules.values(), default=None)
 
 
+class ModelMonitoringSchedulesFileChief(ModelMonitoringSchedulesFileBase):
+    def __init__(self, project: str) -> None:
+        """
+        Initialize applications monitoring schedules chief file object.
+        The JSON file stores a dictionary of registered model endpoints uid as key and point to a dictionary of
+        "last_request" and "last_analyzed" mapped to two Unix timestamp as values.
```
jond01: minor
```suggestion
        "last_request" and "last_analyzed" mapped to two Unix timestamps as values.
```

mlrun/model_monitoring/controller.py
```diff
@@ -300,26 +313,38 @@ def _should_monitor_endpoint(
...
-                if application_names != batch_window_generator.get_application_list():
+                current_time = mlrun.utils.datetime_now()
+                current_min_last_analyzed = (
+                    batch_window_generator.get_min_last_analyzed()
+                )
+                if (
+                    # Different application names, or last analyzed never updated
+                    application_names != batch_window_generator.get_application_list()
+                    or not current_min_last_analyzed
+                ):
```
davesh0812: We should first check whether there are any applications before sending events to the worker.  
Regarding the `or not current_min_last_analyzed` condition, this may result in sending unnecessary events. When `current_min_last_analyzed` is `None`, it indicates there are no applications in the file. If applications do exist in the app list, they will already be handled by the first condition.

```suggestion
                if (
                    # Different application names, or last analyzed never updated
                    application_names != batch_window_generator.get_application_list()
                   and application_names
                ):
```

mlrun/model_monitoring/db/_schedules.py
```diff
@@ -149,6 +165,68 @@ def get_min_timestamp(self) -> Optional[int]:
...
+    def update_endpoint_timestamps(
+        self, endpoint_uid: str, last_request: int, last_analyzed: int
+    ) -> None:
+        self._check_open_schedules()
+        self._schedules[endpoint_uid] = {
+            schemas.model_monitoring.constants.ScheduleChiefFields.LAST_REQUEST: last_request,
+            schemas.model_monitoring.constants.ScheduleChiefFields.LAST_ANALYZED: last_analyzed,
+        }
+
+    def get_endpoint_last_analyzed(self, endpoint_uid: str) -> Optional[int]:
```
jond01, line 205: Could you add some info about this "last analyzed" in the docstrings? If it's not app-specific, is it the oldest time? Something else?

mlrun/model_monitoring/controller.py
```diff
@@ -612,14 +658,16 @@ def endpoint_to_regular_event(
...
         if self._should_monitor_endpoint(
             endpoint,
             set(applications_names),
             policy.get(ControllerEventEndpointPolicy.BASE_PERIOD, 10),
+            schedule_file,
         ):
-            logger.info(
-                "Regular event is being pushed to controller stream for model endpoint",
+            logger.debug(
+                "endpoint data is being prepared for regular event",
```
jond01: minor
```suggestion
                "Endpoint data is being prepared for regular event",
```

mlrun/model_monitoring/controller.py
```diff
@@ -281,11 +282,23 @@ def _should_monitor_endpoint(
             2.  first request exists
             3.  last request exists
             4.  endpoint_type is not ROUTER
-        if the four above conditions apply we require one of the three conditions to monitor:
+        if the four above conditions apply we require one of the two condition monitor:
             1.  never monitored the one of the endpoint applications meaning min_last_analyzed is None
-            2.  last request has a higher timestamp than the min_last_analyzed timestamp
-            3.  We didn't analyze one of the application for over than _MAX_OPEN_WINDOWS_ALLOWED windows
+            2.  min_last_analyzed stands in the condition for sending NOP event and this the first time regular event
+            is sent with the combination of  current last_request  & current last_analyzed  per endpoint.
```
jond01, line 288: Is it the answer to my question? 🤔

mlrun/model_monitoring/controller.py
```diff
@@ -281,11 +282,23 @@ def _should_monitor_endpoint(
...
         """
+        last_timestamp_sent = schedules_file.get_endpoint_last_request(
+            endpoint.metadata.uid
+        )
+        last_analyzed_sent = schedules_file.get_endpoint_last_analyzed(
+            endpoint.metadata.uid
+        )
+        logger.debug(
+            "Chief Should monitor:",
```
jond01: minor
```suggestion
            "Chief Should monitor",
```

mlrun/model_monitoring/controller.py
```diff
@@ -300,26 +313,42 @@ def _should_monitor_endpoint(
                 project=endpoint.metadata.project,
                 endpoint_id=endpoint.metadata.uid,
             ) as batch_window_generator:
-                base_period_seconds = base_period_minutes * _SECONDS_IN_MINUTE
-                if application_names != batch_window_generator.get_application_list():
+                current_time = mlrun.utils.datetime_now()
+                current_min_last_analyzed = (
+                    batch_window_generator.get_min_last_analyzed()
```
jond01, line 318: What is the initial value?

jond01 (approved):
Looks good! I left some questions :)

jond01 (approved):

mlrun/model_monitoring/db/_schedules.py
royischoss: I will add a comment yes it is the oldest last_analyze we sent an event to the controller stream

mlrun/model_monitoring/controller.py
royischoss: Yeah while we are writing to file the oldest combined with the last request to prevent duplication, making sure one of the values changed

mlrun/model_monitoring/controller.py
royischoss: this values goes to the nop event check so we want them stable and not read from file in each check

mlrun/model_monitoring/helpers.py
royischoss: I think it very unlikely maybe in the future we will add a prefix or suffix but still the chance is there

davesh0812 (approved):
Nice!

PR #7483, opened 2025-03-23
[Labels] Validate labels

moranbental:
Adding validation to verify that labels are a dictionary of strings.

Additionally, the maximum length for the label key is 255 characters.
[Here](https://github.com/mlrun/mlrun/pull/7149), we validate the label value, but not the label key.
This causes the error: `Failed committing changes to DB`.

These bugs are relevant for all resources that have labels stored in a different table:

- Artifact
- Run
- Schedule
- Function
- Project
- Feature
- Entity
- FeatureSet
- FeatureVector
- ModelEndpoint

https://iguazio.atlassian.net/browse/ML-9632
https://iguazio.atlassian.net/browse/ML-9599

### DIFF
--- a/server/py/framework/db/sqldb/helpers.py
+++ b/server/py/framework/db/sqldb/helpers.py
@@ -63,7 +63,7 @@ def run_state(run):
     )
 
 
-def update_labels(obj, labels: dict):
+def update_labels(obj, labels: dict[str, str]):
     old = {label.name: label for label in obj.labels}
     obj.labels.clear()
     for name, value in labels.items():
--- a/server/py/framework/db/sqldb/helpers.py
+++ b/server/py/framework/db/sqldb/helpers.py
@@ -63,7 +63,7 @@ def run_state(run):
     )
 
 
-def update_labels(obj, labels: dict[str, str]):
+def update_labels(obj, labels: dict[str, typing.Union[str, int]]):
     if not isinstance(labels, dict):
         raise mlrun.errors.MLRunInvalidArgumentError("Labels must be a dictionary.")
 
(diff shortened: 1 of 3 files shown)


server/py/framework/db/sqldb/helpers.py
```diff
@@ -67,11 +67,11 @@ def update_labels(obj, labels: dict):
...
     for name, value in labels.items():
+        error_msg = f" of `{name}` label is too long. Maximum allowed length is {max_str_length} characters."
+        if len(str(name)) > max_str_length:
+            raise mlrun.errors.MLRunInvalidArgumentError("Name" + error_msg)
         if len(str(value)) > max_str_length:
-            raise mlrun.errors.MLRunInvalidArgumentError(
-                f"Value of `{name}` label is too long. "
-                f"Maximum allowed length is {max_str_length} characters."
-            )
+            raise mlrun.errors.MLRunInvalidArgumentError("Value" + error_msg)
```
TomerShor: While this works, I'd use another helper function like this:
```python
def validate_label_length(label_type: str, label_name: str, label_value: str, max_length: int):
    """Validates the length of a label name or value and raises an error if it exceeds max_length."""
    if len(label_value) > max_length:
            f"{label_type} of `{label_name}` label is too long. Maximum allowed length is {max_length} characters."
        )
```
Then call twice:
```python
    validate_label_length("Name", name, name, max_str_length)
    validate_label_length("Value", name, value, max_str_length)
```

server/py/framework/db/sqldb/helpers.py
```diff
@@ -176,6 +175,24 @@ def ensure_max_length(string: str):
     return string
 
 
+def _validate_label_length(label_type: str, label_name: str, validate_element: str):
+    """Validates the length of a label name or value and raises an error if it exceeds max_length."""
+    if len(validate_element) > max_str_length:
+        raise mlrun.errors.MLRunInvalidArgumentError(
+            f"{label_type} of `{label_name}` label is too long. Maximum allowed length is {max_str_length} characters."
+        )
```
TomerShor: move below `_validate_label`

TomerShor (changes requested):
Looks good, minor comment

liranbg (approved):

TomerShor (approved):

server/py/framework/db/sqldb/helpers.py
```diff
@@ -176,6 +175,31 @@ def ensure_max_length(string: str):
...
+            "The value in the label must be a string or an integer."
+        )
+
+    value = str(value)
+
+    _validate_label_length(label_type="Name", label_name=name, validate_element=name)
+    _validate_label_length(label_type="Value", label_name=name, validate_element=value)
+
+
+def _validate_label_length(label_type: str, label_name: str, validate_element: str):
```
rokatyy, line 195: validate_element can be int. That's why it was `if len(str(value)) > max_str_length:`

moranbental: @rokatyy No, it cannot be because I am doing that: `value = str(value)` before

rokatyy: @moranbental right, missed that 👍

PR #7490, opened 2025-03-24
[Linting] Enforce fully qualified imports

adamdelman:


.github/workflows/enforce_imports.yml
```diff
@@ -0,0 +1,45 @@
...
+        cache: pip
+
+    - name: Get Modified Files
+      run: |
+        BASE_BRANCH=${{ github.event.pull_request.base.ref }}
+        git diff --name-only $(git merge-base HEAD $BASE_BRANCH) > modified_files.txt
+
+    - name: Run Import Check
+      run: |
+        python automation/scripts/check_fully_qualified_imports.py $(cat modified_files.txt)
```
jond01, line 53: We should allow such imports (e.g. `from mlrun import get_or_create_project`) in tests, docs and examples

jond01:
> _ No description provided. _

Please add a description and a link to the Jira ticket if there's one

.github/workflows/enforce_imports.yml
```diff
@@ -0,0 +1,53 @@
...
+      - '!docs/**'
+      - '!examples/**'
+      - '!tests/**'
+      - '!automation/**'
+jobs:
+  check_imports:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.11"]
```
jond01, line 30: Irrelevant here

.github/workflows/enforce_imports.yml
```diff
@@ -0,0 +1,53 @@
...
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Set up python ${{ matrix.python-version }}
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+        cache: pip
```
jond01, line 39: Cache is irrelevant - you don't need any dependencies to run the script

.github/workflows/enforce_imports.yml
```diff
@@ -0,0 +1,53 @@
...
+        git diff --name-only $(git merge-base HEAD "$BASE_BRANCH") \
+          -- . \
+          ':(exclude)docs/**' \
+          ':(exclude)examples/**' \
+          ':(exclude)tests/**' \
+          ':(exclude)automation/**' \
+          > modified_files.txt
+    - name: Run Import Check
+      run: |
+        python automation/scripts/check_fully_qualified_imports.py $(cat modified_files.txt)
```
jond01, line 53: Missing newline at the EOF, we should probably check that in lint...

.github/workflows/enforce_imports.yml
```diff
@@ -0,0 +1,53 @@
...
+    - name: Get Modified Files
+      run: |
+        BASE_BRANCH=${{ github.event.pull_request.base.ref }}
+        git diff --name-only $(git merge-base HEAD "$BASE_BRANCH") \
+          -- . \
+          ':(exclude)docs/**' \
+          ':(exclude)examples/**' \
+          ':(exclude)tests/**' \
+          ':(exclude)automation/**' \
+          > modified_files.txt
```
jond01, line 50: Please follow the standard way of saving step outputs in a GitHub workflow step with `$GITHUB_OUTPUT`:
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-defining-outputs-for-a-job
See also other workflows in the repo

.github/workflows/enforce_imports.yml
jond01, line 53: We'd better check only the changed lines and not whole files.
Such a check will make the first repo-wide changes impractical - consider #7619, for example.
It's not their job to fix it, and they probably won't add it to their already extensive changes. What do you plan in that aspect?

jond01 (changes requested):

jond01:
> adamdelman wants to merge 8 commits into `mlrun:1.9.x` from `adamdelman:adamd/enforce_imports`

Why isn't the base branch `mlrun:development`?

PR #7495, opened 2025-03-25
[DB] Fix bug when object sorting is sometimes inconsistent

moranbental:
When listing runs, the object sorting is sometimes inconsistent if the dates are very close to each other.

For runs:
The issue was that the sorting was based on the "start_time" column, which was saved in the table as a timestamp with fsp=3. After sorting, we were returning the start_time date from the "start_time" field in the struct, which was saved as a timestamp with fsp=6.

**Note**: In SQLite, the start_time column returns timestamps with fsp=6.

Also, if the timestamps are the same, we need a secondary field to order by. That's why we are ordering by ID.

I also fixed that for functions and alert activations.

**Note**: The function is now returning the field as a string, not as a datetime

related pr for artifacts - https://github.com/mlrun/mlrun/pull/7423
related CR - https://iguazio.atlassian.net/browse/ML-9670


server/py/services/api/tests/unit/db/test_runs.py
```diff
@@ -593,6 +594,44 @@ def test_list_runs_with_same_names(self):
         )
         assert len(runs) == 4
 
+    def test_list_runs_orders_by_id_when_start_time_is_identical(self):
+        # this test is verified that when start_time date is identical, runs should be ordered by run id
```
rokatyy:
```suggestion
        # this test verifies that when start_time date is identical, runs should be ordered by run id
```

rokatyy (approved):
LGTM

server/py/framework/db/sqldb/db.py
```diff
@@ -525,10 +526,17 @@ def del_runs(
...
         self, run: Run, run_struct: dict, with_notifications: bool
     ):
+        status = run_struct.setdefault("status", {})
+
+        # These fields are saved in struct as timestamps with fsp=6, while the corresponding columns
+        # in the database have fsp=3. Since 'ORDER BY' is applied to the column, we return the value from
+        # the column (not from the struct) to ensure the ordering is correct.
+        # In SQLite, the start_time and updated columns return timestamps with fsp=6.
+        status["start_time"] = self._add_utc_timezone(run.start_time).isoformat()
+        status["last_update"] = self._add_utc_timezone(run.updated).isoformat()
```
liranbg: if run.start_time / run.updated is None, it will raise an exception.
I suggest to condition that on whether it is set
```suggestion
for status_field, struct_field in [("end_time", "end_time"), ("start_time", "start_time"), ("last_update", "update")]:
        if field_value := getattr(run, struct_field, None):
          status[status_field] = self._add_utc_timezone(field_value).isoformat()
```

server/py/framework/db/sqldb/db.py
```diff
@@ -2368,6 +2376,15 @@ def list_functions(
...
         ):
             function_dict = function.struct
+
+            # updated field is saved in struct as timestamps with fsp=6, while the corresponding column
+            # in the database have fsp=3. Since 'ORDER BY' is applied to the column, we return the value from
+            # the column (not from the struct) to ensure the ordering is correct.
+            # In SQLite, the updated column return timestamps with fsp=6.
+            # TODO: The function currently returns the field as a datetime. It should return it as a string
+            #  to align with other resources.
+            function_dict["metadata"]["updated"] = function.updated
```
liranbg: you may return it as string

server/py/framework/db/sqldb/db.py
```diff
@@ -2565,6 +2582,14 @@ def _get_function(
...
             function.setdefault("status", {})
             function["status"]["state"] = obj.state
+
+            # updated field is saved in struct as timestamps with fsp=6, while the corresponding column
+            # in the database have fsp=3. Since 'ORDER BY' is applied to the column, we return the value from
+            # the column (not from the struct) to ensure the ordering is correct.
+            # In SQLite, the updated column return timestamps with fsp=6.
+            # TODO: The function currently returns the field as a datetime. It should return it as a string
+            #  to align with other resources.
+            function["metadata"]["updated"] = obj.updated
```
liranbg: return as string as well

server/py/framework/db/sqldb/db.py
```diff
@@ -525,10 +526,20 @@ def del_runs(
...
-        if run.end_time:
-            run_struct.setdefault("status", {})["end_time"] = self._add_utc_timezone(
-                run.end_time
-            ).isoformat()
+        status = run_struct.setdefault("status", {})
+
+        # These fields are saved in struct as timestamps with fsp=6, while the corresponding columns
+        # in the database have fsp=3. Since 'ORDER BY' is applied to the column, we return the value from
+        # the column (not from the struct) to ensure the ordering is correct.
+        # In SQLite, the start_time and updated columns return timestamps with fsp=6.
```
alonmr, line 534: Comment is not accurate because end time is a datetime and is fsp 6
```suggestion
        # Return the value from the column to ensure the ordering is correct since the sort is done on the table columns and
        # timestamps are being saved with fsp=3 while struct fields are fsp=6.
        # In SQLite, the start_time and updated columns return timestamps with fsp=6.
```

server/py/services/api/tests/unit/db/test_functions.py
```diff
@@ -522,7 +522,7 @@ def test_list_functions_with_time_filters(self):
 
         # extract the updated time of the functions
         function_times = [
-            function["metadata"]["updated"]
+            datetime.datetime.fromisoformat(function["metadata"]["updated"])
```
alonmr, line 525: Just verifying - this does not break BC because the response schema converts it to string anyway?

server/py/services/api/tests/unit/db/test_functions.py
```diff
@@ -706,6 +706,47 @@ def test_list_functions_returns_elements_by_order_updated_field(self):
                 function_name == expected_name
             ), f"Expected {expected_name}, got {function_name}"
 
+    def test_list_functions_orders_by_id_when_updated_is_identical(self):
+        # this test is verified that when updated date is identical, functions should be ordered by function id
```
alonmr, line 710:
```suggestion
        # this test verifies that when updated date is identical, functions should be ordered by function id
```

server/py/services/api/tests/unit/db/test_functions.py
```diff
@@ -706,6 +706,47 @@ def test_list_functions_returns_elements_by_order_updated_field(self):
...
+            self._db.store_function(
+                self._db_session,
+                function.to_dict(),
+                function.metadata.name,
+                versioned=False,
+                tag=tag,
+            )
+
+            # Set the same `updated` timestamp for all functions
+            db_artifact = self._db._query(
```
alonmr, line 726:
```suggestion
            db_function = self._db._query(
```

liranbg (approved):

alonmr (changes requested):
Nice! Some small comments

PR #7501, opened 2025-03-25
[Model Monitoring] Query the TSDB to check for new data on non-batch model endpoints

Eyal-Danieli:
Includes the following changes:
- The controller now query the TSDB for new data on non-batch model endpoints, rather than checking parquet files. For batch model endpoints, the controller will continue to check the parquet files.
- disable `record_results` on existing non-batch model endpoints in order to avoid offline batching on a real time monitored mep, which could cause issues with time windows.
- Fix a bug where the controller incorrectly checked the `endpoint_policy` key instead of the `endpoint_type` key to identify non-batch meps.
- Remove an initiation gap between the `first_request` and `last_request` in the batch EP, which was implemented prior to NOP event support and led to displaying an incorrect value for `last_prediction` in the UI.

https://iguazio.atlassian.net/browse/ML-9644


mlrun/model_monitoring/controller.py
```diff
@@ -281,9 +281,12 @@ def __init__(self) -> None:
                 mlrun.platforms.iguazio.KafkaOutputStream,
             ],
         ] = {}
-        self.feature_sets: OrderedDict[str, mlrun.common.schemas.FeatureSet] = (
+        self.feature_sets: OrderedDict[str, mlrun.feature_store.FeatureSet] = (
```
royischoss, line 284: 🥇

mlrun/model_monitoring/controller.py
```diff
@@ -477,16 +480,9 @@ def model_endpoint_process(
             ]
 
             not_batch_endpoint = (
-                event[ControllerEvent.ENDPOINT_POLICY] != EndpointType.BATCH_EP
+                event[ControllerEvent.ENDPOINT_TYPE] != EndpointType.BATCH_EP
```
royischoss, line 483: 👑

mlrun/model_monitoring/controller.py
```diff
@@ -511,13 +507,39 @@ def model_endpoint_process(
...
+                                endpoint_id=endpoint_id,
+                            )
+                            if prediction_metric.data:
+                                data_in_window = True
+                        else:
+                            if endpoint_id not in self.feature_sets:
+                                self.feature_sets[endpoint_id] = fstore.get_feature_set(
+                                    event[ControllerEvent.FEATURE_SET_URI]
+                                )
+                            self.feature_sets.move_to_end(endpoint_id, last=False)
```
royischoss, line 525: I wonder if it better to introduce a method for getting the feature set, might add a small pr after yours is in. Let me know what you think.

tests/system/model_monitoring/test_model_monitoring.py
```diff
@@ -1191,41 +1191,92 @@ def test_batch_drift(self):
...
+
+        # by default, the model endpoint is created with the node type
+        assert (
+            model_endpoint_non_batch.metadata.endpoint_type
+            == mlrun.common.schemas.model_monitoring.EndpointType.NODE_EP
+        )
+
         # Wait for the controller, app and writer to complete
-        sleep(130)
+        sleep(180)
```
royischoss, line 1238: I think we should add constants or comment on why we are waiting this time - parquet flash NOP event etc.

royischoss (approved):
Hey LGTM I will add the small comments on a follow up small PR

mlrun/model_monitoring/controller.py
```diff
@@ -511,13 +507,39 @@ def model_endpoint_process(
...
+                        data_in_window = False
+                        if not_batch_endpoint:
+                            # Serving endpoint - get the relevant window data from the TSDB
+                            prediction_metric = self.tsdb_connector.read_predictions(
+                                start=start_infer_time,
+                                end=end_infer_time,
+                                endpoint_id=endpoint_id,
+                            )
+                            if prediction_metric.data:
+                                data_in_window = True
```
davesh0812, line 519: Can we cache it across application? maybe to use dictionary that the key is the `str({start_time}-{base period})`, same for the check on the df?

mlrun/model_monitoring/controller.py
```diff
@@ -511,13 +507,39 @@ def model_endpoint_process(
...
-                            start_time=start_infer_time,
-                            end_time=end_infer_time,
-                            time_column=mm_constants.EventFieldType.TIMESTAMP,
-                            storage_options=self.storage_options,
-                        )
-                        if len(df) == 0:
+                        data_in_window = False
+                        if not_batch_endpoint:
+                            # Serving endpoint - get the relevant window data from the TSDB
+                            prediction_metric = self.tsdb_connector.read_predictions(
```
davesh0812, line 513: Do you have an idea how much time `read_predictions()` takes?

davesh0812 (commented):
Looks Good, I have one extra suggestion to improve the performance :)

PR #7419, opened 2025-03-11
[Model Monitoring] Remove parallelization of V3IO TSDB queries

gtopper:
As it puts too much pressure on framesd, as per [ML-9527](https://iguazio.atlassian.net/browse/ML-9527).

[ML-9527]: https://iguazio.atlassian.net/browse/ML-9527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7505, opened 2025-03-26
[Model Monitoring] Add missing `run_in_threadpool`

gtopper:
[ML-9651](https://iguazio.atlassian.net/browse/ML-9651)

[ML-9651]: https://iguazio.atlassian.net/browse/ML-9651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
+++ b/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
@@ -1090,9 +1090,9 @@ class V3IOTSDBConnector(TSDBConnector):
         Fetch basic metrics from V3IO TSDB and add them to MEP objects.
 
         :param model_endpoint_objects: A list of `ModelEndpoint` objects that will
-                                        be filled with the relevant basic metrics.
+                                       be filled with the relevant basic metrics.
         :param project:                The name of the project.
-        :param run_in_threadpool:      Has no effect.
+        :param run_in_threadpool:      A function that runs another function in a thread pool.
 
         :return: A list of `ModelEndpointMonitoringMetric` objects.
         """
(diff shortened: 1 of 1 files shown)


davesh0812 (commented):
Are you sure https://github.com/mlrun/mlrun/pull/7419 is relevant? 
Why not to revert it?

If I remember correctly we solved [ML-9527](https://iguazio.atlassian.net/browse/ML-9527) with the KV no?

gtopper:
I don't think it matters much either way (whether we parallelize it or not). [ML-9527](https://iguazio.atlassian.net/browse/ML-9527) was caused by reading 24h of raw data because the relevant pre-aggregation wasn't defined on the table.

[ML-9527]: https://iguazio.atlassian.net/browse/ML-9527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

gtopper:
@davesh0812 See #7506 and pick your poison :)

davesh0812:
I'll leave it to @assaf758  to decide. :)

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -1104,9 +1104,15 @@ async def add_basic_metrics(
...
-        drift_status_res = self.get_drift_status(endpoint_ids=uids, get_raw=True)
+        error_count_res = await run_in_threadpool(
+            self.get_error_count, endpoint_ids=uids, get_raw=True
+        )
+        avg_latency_res = await run_in_threadpool(
+            self.get_avg_latency, endpoint_ids=uids, get_raw=True
+        )
+        drift_status_res = await run_in_threadpool(
+            self.get_drift_status, endpoint_ids=uids, get_raw=True
+        )
```
liranbg, line 1115: why not move it to a function and run that function only? saving the overhead of moving those 3 altogether to a thread?

assaf758:
let's go with the safer approach of #7505, where we don't run them in parallel (to make sure we are of the hook on https://iguazio.atlassian.net/browse/ML-9651).
Also, we probably need a follow-up PR for putting `_enrich_mep_with_last_request` in a threadpool as well, I think it makes sense to have one thread for all 4 metrics as @liranbg suggested

assaf758 (approved):

PR #7509, opened 2025-03-27
[Nuclio] Mask trigger sensitive fields

rokatyy:
The logic is that we send raw data to nuclio and after sending it, mask the fields by the same algorithms it is done in nuclio. 

List of fields to mask: https://github.com/nuclio/nuclio/blob/01b66be40410cb9827b3de0eb7c8628ade18911b/pkg/platformconfig/types.go#L319-L333. 

Jira - https://iguazio.atlassian.net/browse/ML-9443

image

### DIFF
--- a/mlrun/runtimes/nuclio/function.py
+++ b/mlrun/runtimes/nuclio/function.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import asyncio
+import copy
 import json
 import typing
 import warnings
@@ -49,6 +50,19 @@ from mlrun.runtimes.pod import KubeResource, KubeResourceSpec
 from mlrun.runtimes.utils import get_item_name, log_std
 from mlrun.utils import get_in, logger, update_in
 from mlrun_pipelines.common.ops import deploy_op
+
+SENSITIVE_PATHS_IN_TRIGGER_CONFIG = {
+    "password",
+    "secret",
+    "attributes/password",
+    "attributes/accesskeyid",
+    "attributes/secretaccesskey",
+    "attributes/cacert",
+    "attributes/accesskey",
+    "attributes/accesscertificate",
+    "attributes/sasl/password",
+    "attributes/sasl/oauth/clientsecret",
+}
 
 
 def validate_nuclio_version_compatibility(*min_versions):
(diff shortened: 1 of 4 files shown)


mlrun/runtimes/nuclio/function.py
```diff
@@ -274,6 +288,37 @@ def pre_deploy_validation(self):
...
             mlrun.utils.validate_tag_name(self.metadata.tag, "function.metadata.tag")
 
+    def mask_sensitive_data_in_config(self):
+        if not self.spec.config:
+            return {}
+
+        raw_config = copy.deepcopy(self.spec.config)
+
+        for key, value in self.spec.config.items():
+            if key.startswith("spec.triggers"):
```
alonmr, line 298: Since you enlightened us with the list in https://github.com/nuclio/nuclio/blob/01b66be40410cb9827b3de0eb7c8628ade18911b/pkg/platformconfig/types.go#L305

what about the other fields in the list? should we consider them as well?

mlrun/runtimes/nuclio/function.py
```diff
@@ -274,6 +288,37 @@ def pre_deploy_validation(self):
...
+                trigger_name = key.split(".")[-1]
+
+                for path in SENSITIVE_PATHS_IN_TRIGGER_CONFIG:
+                    # Handle nested keys
+                    nested_keys = path.split("/")
+                    target = value
+                    for sub_key in nested_keys[:-1]:
+                        target = target.get(sub_key, {})
```
alonmr, line 306: you may use `mlrun.utils.helpers.get_in` if you change the `/` notation to `.`

mlrun/runtimes/nuclio/function.py
```diff
@@ -274,6 +288,37 @@ def pre_deploy_validation(self):
...
+                    if last_key in target:
+                        sensitive_field = target[last_key]
+                        if sensitive_field.startswith(
+                            mlrun.model.Credentials.secret_reference_prefix
+                        ):
+                            # already masked
+                            continue
+                        target[last_key] = (
+                            f"{mlrun.model.Credentials.secret_reference_prefix}/spec/triggers/{trigger_name}/{path}"
+                        )
```
alonmr, line 318: Is this the same masking that nuclio does? meaning if I redeploy a function which is already masked and nuclio has a secret for the password will it work?

alonmr (commented):
Great stuff! Small comment/questions

mlrun/runtimes/nuclio/function.py
rokatyy: @alonmr we need `/` format below anyway, so I'd stick with it

mlrun/runtimes/nuclio/function.py
rokatyy: @alonmr yes, nuclio understands if field is already masked

mlrun/runtimes/nuclio/function.py
TomerShor: @rokatyy the list can also change, as it accepts custom sensitive fields as an input to the platformconfig.

rokatyy: @TomerShor yes, list can change, however we don't have a sync mechanism for now between mlrun and nuclio, so we had either this option or an option of duplicating secret creation mechanism in mlrun.

So it was decided to go with this option, assuming that fields won't change and masking is always enabled.

rokatyy: @alonmr from all other params, I'd say it would be also nice to support codeentryattributes masking (because I found places where we fill them in with sensitive data), volumes in mlrun are already masked from what I understand. 

And annotations are up to us, but can be added as well

liranbg: @rokatyy this sounds like a change request we can add

alonmr (approved):

PR #7515, opened 2025-03-27
[Serving] Fix step name  in graph using `set_flow`  

royischoss:
https://iguazio.atlassian.net/browse/ML-9540
Fix adding logic to retrieve step name while calling the  `params_to_step` method, adding unit test.


mlrun/serving/states.py
```diff
@@ -1646,7 +1646,7 @@ def _insert_error_step(self, name, step):
 
     def set_flow(
         self,
-        steps: list[Union[str, StepToDict, dict[str, Any]]],
+        steps: list[Union[str, dict[str, Any], BaseStep]],
```
gtopper: `to()` expects a `class_name: Union[str, StepToDict]`, so changing it to `BaseStep` here causes a type inconsistency.

mlrun/serving/states.py
```diff
@@ -1661,7 +1661,7 @@ def set_flow(
             if isinstance(next_step, dict):
                 step = step.to(**next_step)
             else:
-                step = step.to(next_step)
+                step = step.to(next_step, name=next_step.name)
```
gtopper: Doesn't it already take the name from `step.name` internally?

tests/serving/test_flow.py
```diff
@@ -429,3 +430,31 @@ def test_set_flow():
...
+def test_set_flow_names(
+    steps: list[Union[TaskStep, QueueStep, dict[str, Any]]],
+):
+    fn = mlrun.new_function("tests", kind="serving")
+    graph = fn.set_topology("flow", engine="sync")
+    graph.set_flow(steps=steps, force=True)
+    if isinstance(steps[0], dict):
+        assert list(graph.to_dict()["steps"].keys()) == [step["name"] for step in steps]
+    else:
+        assert list(graph.to_dict()["steps"].keys()) == [step.name for step in steps]
```
gtopper:
```suggestion
    expected = [step["name"] for step in steps] if isinstance(steps[0], dict) else [step.name for step in steps]
    assert list(graph.to_dict()["steps"].keys()) == expected
```

mlrun/serving/states.py
royischoss: it didn't this way for every step kind we give a name

royischoss: There was a problem that some steps didn't got the name field in to_dict method and that was the root cause of the bug

mlrun/serving/states.py
```diff
@@ -412,7 +412,7 @@ def to(
 
     def set_flow(
         self,
-        steps: list[Union[str, StepToDict, dict[str, Any]]],
+        steps: list[Union[str, dict[str, Any], StepToDict]],
```
gtopper: I doubt this change is on purpose. Please revert.

mlrun/serving/states.py
```diff
@@ -1646,7 +1646,7 @@ def _insert_error_step(self, name, step):
...
```
gtopper: I doubt this change is on purpose. Please revert.

mlrun/serving/states.py
gtopper: Ok, but then the bug is _not_ fixed if the user uses `.to()` directly and expects the name to be picked up from the step.

royischoss: There are some steps we really demands a name, there was a change in to_dict of BaseStep so it won't returns name field, I didn't want to touch this behaviour

gtopper (approved):

PR #7523, opened 2025-03-30
[Requirements-automated] Upgrade lock files [1.9.x]

iguazio-cicd:
Update the uv generated lock files.
Generated by the run:
https://github.com/mlrun/mlrun/actions/runs/14153958066


jond01:
@liranbg this workflow was not designed to work against multiple branches, as it creates its changes in the same branch:

https://github.com/mlrun/mlrun/blob/72cb53c400c7955944bf533cca783c09b73445ae/.github/workflows/upgrade-lock.yaml#L47

So we can run it one by one or slightly adapt it. Do you think it will be a common case?

liranbg:
> @liranbg this workflow was not designed to work against multiple branches, as it creates its changes in the same branch:
> 
> https://github.com/mlrun/mlrun/blob/72cb53c400c7955944bf533cca783c09b73445ae/.github/workflows/upgrade-lock.yaml#L47
> 
> So we can run it one by one or slightly adapt it. Do you think it will be a common case?

indeed it is.

PR #7527, opened 2025-03-30
[Tests] Limit vizro version due to pydantic upgrade

alonmr:
https://github.com/mckinsey/vizro/blob/main/vizro-core/CHANGELOG.md#0132--2025-02-04

This change breaks our applications so pin until fixed

### DIFF
--- a/tests/system/runtimes/test_application.py
+++ b/tests/system/runtimes/test_application.py
@@ -187,7 +187,7 @@ class TestApplicationRuntime(tests.system.base.TestMLRunSystem):
         function = self.project.set_function(
             name=name,
             kind="application",
-            requirements=["vizro", "gunicorn", "Werkzeug==2.2.2"],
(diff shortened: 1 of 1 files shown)


tests/system/runtimes/test_application.py
```diff
@@ -187,7 +187,7 @@ def _create_vizro_application(
         function = self.project.set_function(
             name=name,
             kind="application",
-            requirements=["vizro", "gunicorn", "Werkzeug==2.2.2"],
+            requirements=["vizro<0.1.32", "gunicorn", "Werkzeug==2.2.2"],
```
TomerShor, line 190: Add a TODO to remove it once we upgrade pydatic

alonmr: I'm not sure if it's related to our pydantic version or the actual usage of the vizro application.
I'm also not sure if we should remove it because that's not the point of the test

moranbental (approved):
LGTM

PR #7555, opened 2025-04-01
[Images] Bump mlrun-gpu to python 3.11 and cuda 12.8.1

alonmr:
Part of mlrun migration to python 3.11
Maintains the previous mlrun-gpu image with python 3.9 for backwards compatibility with tag suffix -py39

In new image:
* Specify cudnn instead of cudnn8 - installs latest cudnn (9.x)
* Use symlink `/usr/local/cuda/lib64` instead of specifying the cuda version (available on both cuda images used here)

https://iguazio.atlassian.net/browse/ML-9668


Makefile
```diff
@@ -52,7 +52,7 @@ MLRUN_RELEASE_NOTES_OUTPUT_FILE ?=
 MLRUN_SYSTEM_TESTS_CLEAN_RESOURCES ?= true
 MLRUN_SYSTEM_TEST_MARKERS ?=
 MLRUN_SYSTEM_TESTS_GITHUB_RUN_URL ?=
-MLRUN_GPU_CUDA_VERSION ?= 11.8.0-cudnn8-devel-ubuntu22.04
+MLRUN_GPU_CUDA_VERSION ?= 12.8.1-cudnn-devel-ubuntu22.04
```
liranbg, line 55: we need to document that cudnn / gpu cuda version is updated. 

as you can see https://docs.nvidia.com/deploy/cuda-compatibility/index.html#cuda-11-and-later-defaults-to-minor-version-compatibility

as you can see, cuda >= 12.8 requires driver >=525.60.13
and cuda >= 11.8 requires driver >= 450.80.02*

liranbg (approved):

liranbg (approved):

guy1992l:
This is a welcome change BUT it might break some demos :) we'll wait and see.

PR #7556, opened 2025-04-01
[Model Monitoring] Fix adding `endpoint_uid` to application artifact key

royischoss:
https://iguazio.atlassian.net/browse/ML-9647
`MonitoringApplicationContext::log_artifact:log_dataset` change concat key with endpoint uid allowing unique name and prevent overriding the artifact data by application.


tests/model_monitoring/test_applications/test_app_context.py
```diff
@@ -27,7 +27,7 @@
 from mlrun.serving import GraphContext, GraphServer
 
 
-@pytest.mark.parametrize("method", ["log_artifact", "log_dataset", "log_model"])
+@pytest.mark.parametrize("method", ["log_dataset", "log_model"])
 def test_log_object_signature(method: str) -> None:
```
jond01: 1. Please adapt the test, don't remove it... there's a reason for it.
2. The change is relevant also to other artifact types - so it should be applied to `log_dataset` and `log_model` as well

jond01 (changes requested):

tests/model_monitoring/test_applications/test_app_context.py
royischoss: What is the reason to connect the two?

jond01: It's explained in the following line (L32).
Anyway - if you remove a test, please justify it. Here there's no reason

mlrun/model_monitoring/applications/context.py
```diff
@@ -202,7 +202,9 @@ def _get_default_labels(self) -> dict[str, str]:
 
     def _add_default_labels(self, labels: Optional[dict[str, str]]) -> dict[str, str]:
         """Add the default labels to logged artifacts labels"""
-        return (labels or {}) | self._default_labels
+        return (labels or {}) | {
+            key: value for key, value in self._default_labels.items() if value
```
liranbg: `value is None` maybe? perhaps empty/0 is legit?

tests/model_monitoring/test_applications/test_app_context.py
royischoss: For "log_dataset", "log_model" I will take another look. Do we need log_model in MonitoringApplicationContext?

jond01: Why not? At this point, I don't think we want to change this API

tests/model_monitoring/test_applications/test_app_context.py
```diff
@@ -30,9 +30,13 @@
...
     """Future-proof the `log_x` method of MM app context with respect to the project object"""
-    assert inspect.signature(
+    monitoring_parameters = inspect.signature(
         getattr(MonitoringApplicationContext, method)
-    ) == inspect.signature(getattr(MlrunProject, method))
+    ).parameters.values()
+    project_parameters = inspect.signature(
+        getattr(MlrunProject, method)
+    ).parameters.values()
+    assert all(item in monitoring_parameters for item in project_parameters)
```
jond01: I think that the following will work:
```suggestion
    assert project_parameters <= monitoring_parameters
```

tests/model_monitoring/test_applications/test_app_context.py
davesh0812: I talked with @jond01 and we agree to remove `log_model`, I know it already in from 1.7, but we dont see any relevant scenario to use it and we dont have BC in MM, so lets delete it.

mlrun/model_monitoring/applications/context.py
```diff
@@ -329,13 +330,23 @@ def log_artifact(
...
+        :param unique_per_endpoint: by default True, we will log different artifact for each model endpoint,
+        set to False without changing item key will cause artifact override
         """
         labels = self._add_default_labels(labels)
+        # By default, we want to log different artifact for each model endpoint
+        endpoint_id = labels.get(mlrun_constants.MLRunInternalLabels.endpoint_id, "")
+        if unique_per_endpoint and isinstance(item, str):
+            item = f"{item}-{endpoint_id}" if endpoint_id else item
+        elif unique_per_endpoint:  # isinstance(item, str) is false
+            item.key = f"{item.key}-{endpoint_id}" if endpoint_id else item.key
```
jond01, line 347: Nice 👍 
Maybe change the comment to `isinstance(item, Artifact)`?

jond01 (approved):
LGTM

tests/model_monitoring/test_applications/test_app_context.py
royischoss: Worked nice with keys :)

liranbg (approved):

PR #7558, opened 2025-04-01
[Requirements] Change uv locking to non-universal

jond01:
This PR demonstrates non-universal locking for the mlrun-api lock file.
The advantage of this change is that it allows upgrading only a specific package with the following flag:
https://github.com/mlrun/mlrun/blob/e368ffc06d4f67b7b7853793034022c6127058a8/Makefile#L99-L100

For example, running the following works as expected:
```sh
MLRUN_UV_UPGRADE_FLAG="-P multidict" make upgrade-mlrun-api-deps-lock
```

This cannot be accomplished today, due to https://github.com/astral-sh/uv/issues/6158 (https://github.com/astral-sh/uv/issues/9990).

The downside is that each Python version requires a different lock file, which adds to the complexity.
For example - the mlrun Dockerfile will need two lock files in MLRun 1.9 - one for Python 3.9, and the other for Python 3.11.
Today, both versions are supported in the same lock file.

Note: this PR changes only mlrun-api for now until we decide on the desired direction.


adamdelman:
Overall LGTM, can you please add some more text explaining what the universal flag does and what exactly changed in that regard?

PR #7561, opened 2025-04-01
[MPIJob] Modify server side polling for job startup

Yacouby:
This PR refactors how MPIJob runs are handled by eliminating unnecessary server-side waiting for the job to start. Instead of blocking the request until the job transitions to a running state, the server now:
- Submits the MPIJob
- Fetches the launcher pod status once
- Sets the run state to "running" if the launcher pod has started, otherwise "pending".

This ensures the client is responsible for polling and detecting any failures

Jira:
https://iguazio.atlassian.net/browse/ML-5339

### DIFF
--- a/server/py/services/api/runtime_handlers/mpijob/abstract.py
+++ b/server/py/services/api/runtime_handlers/mpijob/abstract.py
@@ -63,17 +63,26 @@ class AbstractMPIJobRuntimeHandler(KubeRuntimeHandler, abc.ABC):
             state = self._crd_state_to_run_state(status)
             launcher, _ = self._get_launcher(meta.name, meta.namespace)
             execution.set_hostname(launcher)
-            txt = f"MpiJob {meta.name} launcher pod {launcher} is in state {state}"
-            logger.info(txt)
+            status_text = (
+                f"MpiJob {meta.name} launcher pod {launcher} is in state {state}"
+            )
+            logger.info(
+                "MpiJob launcher pod state update",
+                name=meta.name,
+                launcher=launcher,
+                state=state,
+            )
         else:
             # no state yet, assume pending
             state = mlrun.run.RunStatuses.pending
-            txt = f"MpiJob {meta.name} pending - awaiting launcher pod startup"
-            logger.info(txt)
+            status_text = f"MpiJob {meta.name} pending - awaiting launcher pod startup"
+            logger.info(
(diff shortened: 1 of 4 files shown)


server/py/services/api/runtime_handlers/mpijob/abstract.py
```diff
@@ -43,44 +47,36 @@ def run(
             runtime.store_run(run)
 
         meta = self._get_meta(runtime, run, True)
+        self._meta = meta
```
TomerShor: The runtime handler class handles multiple runs of the same runtime. 
Here you "attach" the class metadata to a specific run. Why is this needed?

Yacouby: to use it in `_ensure_run_state` and get the launcher pod which needs `meta.name` & `meta.namespace` to retrieve it

TomerShor: But what if you have 2 (or more) MPIJob runs, each have different name, and the run monitor gets only one of them at a time - the wrong launcher pod might be retrieved since you only save 1 metadata object on the runtime handler.

TomerShor (approved):
LGTM

server/py/services/api/runtime_handlers/mpijob/abstract.py
```diff
@@ -50,37 +52,28 @@ def run(
...
+            state = self._crd_state_to_run_state(status)
+            launcher, _ = self._get_launcher(meta.name, meta.namespace)
+            execution.set_hostname(launcher)
+            txt = f"MpiJob {meta.name} launcher pod {launcher} is in state {state}"
+            logger.info(txt)
+        else:
+            # no state yet, assume pending
+            state = mlrun.run.RunStatuses.pending
+            txt = f"MpiJob {meta.name} pending - awaiting launcher pod startup"
+            logger.info(txt)
```
liranbg:
```suggestion
            logger.info("Waiting for MpiJob launcher pod to start", name=meta.name, state=state)
```
structured log is preferred, giving more organized and context-based log lines, same goes to line above

liranbg (changes requested):
Overall looks neat, small comment wrt logging convention

yaelgen (approved):
LGTM 🚀

liranbg (approved):

PR #7572, opened 2025-04-02
[DB] Fix Handling of Timestamps Without Milliseconds

moranbental:
When using `isoformat()` and there are no milliseconds, by default, isoformat() will convert the timestamp to include only seconds, excluding milliseconds/microseconds. 
This PR fixes cases where there are no milliseconds in the `run.start_time`, `run.end_time`, or `function.updated` timestamps

https://iguazio.atlassian.net/browse/ML-9708


adamdelman (approved):
LGTM

alonmr (commented):
is it also needed in https://github.com/mlrun/mlrun/blob/development/mlrun/lists.py#L208 ?

alonmr (approved):

liranbg (approved):

PR #7580, opened 2025-04-02
[Runtimes] Refactor preemption mode enrichment logic to stateless utility function

yaelgen:
This PR refactors the preemption mode enrichment mechanism by moving the logic from the function specification phase to the runtime execution phase. The change introduces a cleaner separation between user-defined configuration and runtime-resolved scheduling behavior.

**Behavioral Changes**
Function Object (FunctionSpec)
Only stores the `preemption_mode` value (prevent, allow, constrain, or none).
Does not store or apply enriched `tolerations`, `affinity`, or `node_selector` values derived from the preemption mode.

Run Object (RunSpec)
Now explicitly includes tolerations, affinity, and node_selector fields.
During run creation, the server-side launcher enriches these fields based on the function’s `preemption_mode`.
This enrichment occurs similarly to how node selectors are merged and takes precedence over user-provided values and project/service level  defaults.

**Changes Introduced**
- Refactored `enrich_preemption_mode` into a stateless utility that operates on the runtime scheduling spec at execution time.
- FunctionSpec no longer mutates tolerations, affinity, or node selectors as a result of setting preemption_mode.
- Updated relevant unit tests to reflect and verify the new enrichment flow.
- Applied and tested the new logic to the following runtimes: `KubeJob`, `Databricks`, `MPI` and `Nuclio`

**To Be Handled in Follow-Up PRs**
- Add full system test coverage for enriched scheduling behavior.
- Extend the new enrichment logic to the following runtimes: `Spark`, `Dask`, `KFP` and `Kaniko` 

https://iguazio.atlassian.net/browse/ML-9745


mlrun/k8s_utils.py
```diff
@@ -228,3 +230,332 @@ def handle_invalid(message):
...
+# ) -> typing.Optional[kubernetes.client.V1Affinity]:
+#     if not affinity:
+#         return None
+#     if (
+#         not affinity.node_affinity
+#         and not affinity.pod_affinity
+#         and not affinity.pod_anti_affinity
+#     ):
+#         return None
+#     return affinity
```
liranbg: what is it used for?

mlrun/k8s_utils.py
```diff
@@ -228,3 +230,332 @@ def handle_invalid(message):
...
+
+
+def _clear_affinity_if_initialized_but_empty(
```
liranbg: not sure I understand the correlation of the function name and its logic

mlrun/k8s_utils.py
```diff
@@ -228,3 +230,332 @@ def handle_invalid(message):
...
+                        node_selector_terms=new_node_selector_terms
+                    )
+                )
+        # if both preferred and new required are empty, clean node_affinity
+        if (
+            not node_affinity.preferred_during_scheduling_ignored_during_execution
+            and not new_required_during_scheduling_ignored_during_execution
+        ):
+            setattr(affinity, "node_affinity", None)
+            # self.affinity.node_affinity = None
```
liranbg: can be deleted?

mlrun/k8s_utils.py
```diff
@@ -228,3 +230,332 @@ def handle_invalid(message):
...
+            return
+
+        _initialize_affinity(affinity=affinity)
+        _initialize_node_affinity(affinity=affinity)
+
+        # fmt: off
```
liranbg: why is it needed?

liranbg (changes requested):
I see some code is duplicating pod.py wrt preemption mode handling - worth removing what is not needed and keep code DRY

yaelgen:
> I see some code is duplicating pod.py wrt preemption mode handling - worth removing what is not needed and keep code DRY

The code dupe with `pod.py` file will be removed in Spark's follow up PR, since spark is still using it for the driver and executor

server/py/services/api/launcher.py
```diff
@@ -227,21 +227,72 @@ def _pre_run_node_selector_enrichement(
...
-        return run
+            if project:
+                project_node_selector = project.spec.default_function_node_selector
+                resolved_node_selectors = mlrun.runtimes.utils.resolve_node_selectors(
+                    project_node_selector, run.spec.node_selector
+                )
+                mlrun.k8s_utils.validate_node_selectors(resolved_node_selectors)
+                run.spec.node_selector = resolved_node_selectors
+
+    def _enrich_run_with_preemption_mode(
```
alonmr, line 258: this method assumes that `_enrich_run_with_project_node_selector` was called before it. this could easily break so instead of assuming, I would actually call it inside `_enrich_run_with_preemption_mode`

server/py/services/api/runtime_handlers/mpijob/v1.py
```diff
@@ -124,10 +124,10 @@ def _configure_pod_template(
                 run.spec.node_selector
             ),
             "spec.affinity": mlrun.runtimes.pod.get_sanitized_attribute(
-                runtime.spec, "affinity"
+                run.spec, "affinity"
```
alonmr, line 127: can you assume this for old clients?

alonmr (commented):
Very well clean code 🙏 
Minor comments 👇

mlrun/k8s_utils.py
```diff
@@ -228,3 +230,322 @@ def handle_invalid(message):
...
+
+    tolerations = _merge_tolerations(tolerations, preemptible_tolerations)
+    return node_selector, tolerations, affinity
+
+
+def _merge_tolerations(
+    existing: list[kubernetes.client.V1Toleration],
+    to_add: list[kubernetes.client.V1Toleration],
+) -> list[kubernetes.client.V1Toleration]:
+    merged = existing.copy()
```
quaark: from looking at your uses of this method, it seems inefficient to copy and return the `existing` list. 
I'd just merge the tolerations from `to_add` into the original `existing` as a reference and no need to return (even though you can return it)

mlrun/k8s_utils.py
```diff
@@ -228,3 +230,322 @@ def handle_invalid(message):
...
+        return
+
+    mlrun.utils.logger.debug("Pruning node selectors", node_selector=node_selector)
+    for key, value in node_selector.items():
+        if value:
+            spec_value = enriched_node_selector.get(key)
+            if spec_value and spec_value == value:
+                enriched_node_selector.pop(key)
+
+    return enriched_node_selector
```
quaark:
```suggestion
    return {
        key: value
        for key, value in enriched_node_selector.items()
        if node_selector.get(key) != value
    }
```
Your code skips pruning if `node_selector[key]` is `None` (even if `enriched_node_selector[key]` is also `None` and matching) - not sure if that was intentional. If it is then my suggestion is:
```suggestion
    return {
        key: value
        for key, value in enriched_node_selector.items()
        if not node_selector.get(key) or node_selector.get(key) != value
    }
```
And you should add a comment explaining why. If not, then use the first suggestion.

mlrun/k8s_utils.py
```diff
@@ -228,3 +230,322 @@ def handle_invalid(message):
...
+                node_affinity.required_during_scheduling_ignored_during_execution
+            )
+            new_node_selector_terms = (
+                _prune_node_selector_requirements_from_node_selector_terms(
+                    node_selector_terms=node_selector.node_selector_terms,
+                    requirements_to_prune=node_selector_requirements,
+                )
+            )
+            # check whether there are node selector terms to add to the new list of required terms
+            if len(new_node_selector_terms) > 0:
```
quaark:
```suggestion
            if new_node_selector_terms:
```

mlrun/k8s_utils.py
```diff
@@ -228,3 +230,322 @@ def handle_invalid(message):
...
+                    kubernetes.client.V1NodeSelector(
+                        node_selector_terms=new_node_selector_terms
+                    )
+                )
+        # if both preferred and new required are empty, clean node_affinity
+        if (
+            not node_affinity.preferred_during_scheduling_ignored_during_execution
+            and not new_required_during_scheduling_ignored_during_execution
+        ):
+            setattr(affinity, "node_affinity", None)
```
quaark: Is there a reason not to do this here?
```suggestion
            affinity.node_affinity = None
```

mlrun/k8s_utils.py
```diff
@@ -228,3 +230,322 @@ def handle_invalid(message):
...
+    :return: A new list of V1NodeSelectorTerm objects with the specified requirements pruned.
+    """
+    pruned_terms = []
+
+    for term in node_selector_terms:
+        remaining_requirements = [
+            expr
+            for expr in term.match_expressions or []
+            if not any(expr == r for r in requirements_to_prune)
+        ]
```
quaark:
```suggestion
        remaining_requirements = [
            expr
            for expr in term.match_expressions or []
            if expr not in requirements_to_prune
        ]
```
Or better yet:
```suggestion
        remaining_requirements = set(term.match_expressions or []) - set(requirements_to_prune)
```

quaark (changes requested):
Looks good - a few suggestions for efficiency and readability

server/py/services/api/runtime_handlers/mpijob/v1.py
yaelgen: If I'm not missing anything, It is not any different then the node selector change, it passes through the launcher which sets the affinity and toleration on the run object and access it from here.

mlrun/k8s_utils.py
```diff
@@ -228,3 +230,319 @@ def handle_invalid(message):
...
+    preemptible_tolerations = generate_preemptible_tolerations()
+
+    # Map preemption modes to handler functions
+    mode_handlers = {
+        mlrun.common.schemas.PreemptionModes.prevent.value: _handle_prevent_mode,
+        mlrun.common.schemas.PreemptionModes.constrain.value: _handle_constrain_mode,
+        mlrun.common.schemas.PreemptionModes.allow.value: _handle_allow_mode,
+    }
+
+    handler = mode_handlers.get(preemption_mode)
```
rokatyy: how about a separate function for this:

```
def _get_mode_handler(mode: str):
    return {
        PreemptionModes.prevent: _handle_prevent_mode,
        PreemptionModes.constrain: _handle_constrain_mode,
        PreemptionModes.allow: _handle_allow_mode,
    }.get(mode)

```

rokatyy (approved):
Minor comment, otherwise looks good to me.

please also update the description, because the changes include nuclio :)

mlrun/execution.py
```diff
@@ -227,6 +229,16 @@ def node_selector(self):
         """Dictionary with node selectors (read-only)"""
         return deepcopy(self._node_selector)
 
+    @property
```
adamdelman, line 232: Missing return type annotations?

mlrun/k8s_utils.py
```diff
@@ -228,3 +230,320 @@ def handle_invalid(message):
...
+            kubernetes.client.V1NodeSelector(
+                node_selector_terms=generate_preemptible_nodes_anti_affinity_terms()
+            ),
+            affinity,
+        )
+
+    return node_selector, tolerations, affinity
+
+
+def _handle_constrain_mode(
```
adamdelman, line 338: more type annotations :-)

server/py/services/api/tests/unit/runtimes/test_kubejob.py
```diff
@@ -37,6 +38,20 @@
 from services.api.tests.unit.runtimes.base import TestRuntimeBase
 
 
+def create_node_affinity_with_terms(terms):
```
adamdelman: more type annotations :-)

adamdelman (changes requested):
overall looks good, minor stuff

mlrun/k8s_utils.py
```diff
@@ -228,3 +230,344 @@ def handle_invalid(message):
...
+) -> tuple[
+    dict[str, str],
+    list[kubernetes.client.V1Toleration],
+    typing.Optional[kubernetes.client.V1Affinity],
+]:
+    def _prune_tolerations(to_remove: list[kubernetes.client.V1Toleration]):
+        return [t for t in tolerations if t not in to_remove]
+
+    # Ensure no preemptible node tolerations
+    tolerations = _prune_tolerations(preemptible_tolerations)
```
quaark: I don't see where in this method you reuse `_prune_tolerations`, so why not just set the `tolerations` variable to that list comprehension?

quaark (commented):
Very minor

quaark (approved):
🚀

liranbg (approved):

PR #7595, opened 2025-04-03
[Pipelines] Filter pipelines for the project summary [1.8.x]

moranbental:
Include only pipelines created in the past 3 days when calculating the projects summary.
https://iguazio.atlassian.net/browse/ML-9716

### DIFF
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -1155,7 +1155,7 @@ def get_kfp_list_runs_filter(
             {
                 "key": "created_at",
                 "op": kfp_less_than_or_equal_op,
-                "string_value": end_date,
+                "timestamp_value": end_date,
             }
         )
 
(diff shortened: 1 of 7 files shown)


mlrun/utils/helpers.py
```diff
@@ -81,6 +82,11 @@
 DEFAULT_TIME_PARTITIONS = ["year", "month", "day", "hour"]
 DEFAULT_TIME_PARTITIONING_GRANULARITY = "hour"
 
+# KFP filter operation codes
+LESS_THAN_OR_EQUAL_OP = 7  # '<='
+GREATER_THAN_OR_EQUAL_OP = 5  # '>='
+SUBSTRING_OP = 9  # Substring match
```
```suggestion
KFP_LESS_THAN_OR_EQUAL_OP = 7  # '<='
KFP_GREATER_THAN_OR_EQUAL_OP = 5  # '>='
KFP_SUBSTRING_OP = 9  # Substring match
```

liranbg: also why constant? are they really being commonly used? if used by single function - leave it as is

liranbg (dismissed):

liranbg (approved):

theSaarco (approved):
Looks very good, and a very needed fix. I would consider enlarging the window to 3 days, to minimize the risk of miscounting long-running pipelines.

PR #7604, opened 2025-04-03
[Serving] Explicitly set async as the default since it aligns with the actual default behavior for flow topology

davesh0812:
When configuring a stream trigger for a serving function, we check the engine and determine whether to set explicit ack accordingly. To ensure this works correctly, the engine must be explicitly set to async by default.

### DIFF
--- a/mlrun/runtimes/nuclio/serving.py
+++ b/mlrun/runtimes/nuclio/serving.py
@@ -272,7 +272,7 @@ class ServingRuntime(RemoteRuntime):
 
           flow   - workflow (DAG) with a chain of states
                    flow supports both "sync" and "async" engines, with "async" being the default.
-                   branches are not allowed in sync mode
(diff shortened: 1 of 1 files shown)


mlrun/runtimes/nuclio/serving.py
```diff
@@ -271,7 +271,8 @@ def set_topology(
                    can specify special router class and router arguments
 
           flow   - workflow (DAG) with a chain of states
-                   flow support "sync" and "async" engines, branches are not allowed in sync mode
+                   flow supports both "sync" and "async" engines, with "async" being the default.
+                   branches are not allowed in sync mode
```
gtopper:
```suggestion
                   Branches are not allowed in sync mode.
```

gtopper (approved):

royischoss (approved):
LGTM

gtopper (approved):

PR #7617, opened 2025-04-06
[Model Monitoring] Avoid replacing event object as it interferes with offset commit management

gtopper:
[ML-9660](https://iguazio.atlassian.net/browse/ML-9660)

[ML-9660]: https://iguazio.atlassian.net/browse/ML-9660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/model_monitoring/stream_processing.py
+++ b/mlrun/model_monitoring/stream_processing.py
@@ -15,8 +15,6 @@
 import datetime
 import typing
 
-import storey
-
 import mlrun
 import mlrun.common.model_monitoring.helpers
 import mlrun.feature_store as fstore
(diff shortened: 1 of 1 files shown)


mlrun/model_monitoring/stream_processing.py
```diff
@@ -478,8 +477,9 @@ def do(self, full_event):
 
         # Create a storey event object with list of events, based on endpoint_id which will be used
         # in the upcoming steps
-        storey_event = storey.Event(body=events, key=endpoint_id)
-        return storey_event
+        full_event.key = endpoint_id
```
davesh0812, line 480: What will happen on the next step? IIRC it `storey.FlatMap`... all the new events will have the same id? Is it ok?

gtopper: The same key. The flatten with the same endpoint ID is the cause of [ML-9667](https://iguazio.atlassian.net/browse/ML-9667), but this PR does not aim to fix it.

davesh0812 (approved):

PR #7625, opened 2025-04-06
[Python] Bump jupyter image to python 3.11 [1.9.x]

moranbental:
Updated the default Conda environment to use Python 3.11.

Switched the base image to `quay.io/jupyter/scipy-notebook:python-3.11` instead of `quay.io/mlrun/scipy-notebook:python-3.11`, since the original was only used to avoid DockerHub's pull rate limits.

Created a new Conda environment named `mlrun` and set it as the default environment.
https://iguazio.atlassian.net/browse/ML-9596


.github/workflows/build-workflow-matrix.json
```diff
@@ -44,6 +44,11 @@
     "python-version":"3.9",
     "include-suffix":""
   },
+  {
+    "image-name":"jupyter",
+    "python-version":"3.11",
+    "include-suffix":""
+  },
```
alonmr: no need to maintain 2 jupyter images. can remove the 3.9 above

dockerfiles/jupyter/Dockerfile
```diff
@@ -13,12 +13,12 @@
...
 #
 
-ARG MLRUN_PYTHON_VERSION=3.9
+ARG MLRUN_PYTHON_VERSION=3.11
 ARG MLRUN_UV_IMAGE=ghcr.io/astral-sh/uv:latest
 
 FROM ${MLRUN_UV_IMAGE} AS uv-image
 
-FROM quay.io/mlrun/jupyter-scipy-notebook:python-${MLRUN_PYTHON_VERSION}
+FROM quay.io/jupyter/scipy-notebook:python-${MLRUN_PYTHON_VERSION}
```
alonmr, line 21: we should push the base image to `quay.io/mlrun` to make sure it doesn't change under our feet and to avoid rate limiting

dockerfiles/jupyter/Dockerfile
```diff
@@ -58,6 +70,8 @@ COPY --chown=$NB_UID:$NB_GID ./dockerfiles/jupyter/mlce-start.sh /usr/local/bin/
 
 ENV UV_SYSTEM_PYTHON=true UV_LINK_MODE=copy
 
+USER root
```
alonmr: why do you need root here?

alonmr (commented):
Very nice! small comments

dockerfiles/jupyter/Dockerfile
moranbental: I added comment

liranbg (approved):

PR #7631, opened 2025-04-06
[Nuclio] Update client-side with server-side spec changes after deployment

Yacouby:
This PR fixes an issue where replica values (`min_replicas` and `max_replicas`) in the function spec were incorrect after deployment.

The cause of this is that after deploying a function, we call `_enrich_command_from_status`, which saved the outdated client-side function object. This overwrote server-side updates, causing inconsistencies in the function's spec.

So we now update the function’s spec from the server-side deployment response (similar to how we already update the status). This ensures that all server-side changes, including replica settings, are correctly reflected on the client side.

Jira:
https://iguazio.atlassian.net/browse/ML-9725

### DIFF
--- a/mlrun/runtimes/nuclio/function.py
+++ b/mlrun/runtimes/nuclio/function.py
@@ -652,7 +652,7 @@ class RemoteRuntime(KubeResource):
         logger.info("Starting remote function deploy")
         data = db.deploy_nuclio_function(func=self, builder_env=builder_env)
         self.status = data["data"].get("status")
-        self.spec = merge(self.spec.to_dict(), data["data"].get("spec"))
+        self.spec = merge(self.spec.to_dict(), data["data"].get("spec") or {})
 
         self._update_credentials_from_remote_build(data["data"])
 
--- a/mlrun/runtimes/nuclio/function.py
+++ b/mlrun/runtimes/nuclio/function.py
@@ -27,7 +27,6 @@ import requests
 import semver
 from aiohttp.client import ClientSession
 from kubernetes import client
-from mergedeep import merge
 from nuclio.deploy import find_dashboard_url, get_deploy_status
 from nuclio.triggers import V3IOStreamTrigger
 
@@ -652,7 +651,10 @@ class RemoteRuntime(KubeResource):
         logger.info("Starting remote function deploy")
         data = db.deploy_nuclio_function(func=self, builder_env=builder_env)
         self.status = data["data"].get("status")
-        self.spec = merge(self.spec.to_dict(), data["data"].get("spec") or {})
+
+        # Extract the spec to avoid overwriting server-side updates during the later save in
+        # _enrich_command_from_status.
+        self.spec = data["data"].get("spec")
 
         self._update_credentials_from_remote_build(data["data"])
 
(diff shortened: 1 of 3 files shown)


tests/system/model_monitoring/test_app.py
```diff
@@ -823,13 +824,26 @@ def _deploy_monitoring_infra(self) -> None:
...
 
+    @staticmethod
+    def _assert_replicas(fn):
+        """
+        Validate that the 'min_replicas' and 'max_replicas' values in the function's spec are correct after deployment.
+        This check ensures that the replica settings, which are modified on the server side during deployment, are
+        properly reflected on the client side.
+        """
+        assert fn.spec.min_replicas == 1
+        assert fn.spec.max_replicas == 1
```
TomerShor: move to after the test that calls it

TomerShor (changes requested):
Minor comment

mlrun/runtimes/nuclio/function.py
```diff
@@ -655,6 +656,8 @@ def deploy(
         logger.info("Starting remote function deploy")
         data = db.deploy_nuclio_function(func=self, builder_env=builder_env)
         self.status = data["data"].get("status")
+        self.spec = merge(self.spec.to_dict(), data["data"].get("spec") or {})
```
rokatyy: this should strictly be done after `  self._update_credentials_from_remote_build(data["data"])`, where we clean up sensitive data

rokatyy (changes requested):

mlrun/runtimes/nuclio/function.py
rokatyy: also, worth adding comment explaining what problem we are trying to solve here (smth about that we do .save() later and it may lead to rewriting values)

rokatyy (commented):
Also, thinking about it, I see another issue worth raising. Namely, this introduces a significant behavior change. Previously, we only saved to the DB the configurations explicitly set by the user. Now, in addition to that, we’re also saving values enriched by Nuclio. This could lead to those values remaining in the config even after the defaults are changed in either Nuclio or MLRun.

mlrun/runtimes/nuclio/function.py
TomerShor: It is better to clean up sensitive data after the merge, so the merge won't re-add these fields.

alonmr (commented):
What's the reason for doing merge? why not do a full refresh like in `mlrun.model.RunObject.refresh` ?
Edit: here you don't really need to get it from the DB but I don't understand which fields are in the client spec and not in the server spec.

mlrun/runtimes/nuclio/function.py
Yacouby: if we do the merge after, we might overwrite what _`update_credentials_from_remote_build ` did,
so if we do it before the update credentials will clean up the merged env.

Yacouby:
> Also, thinking about it, I see another issue worth raising. Namely, this introduces a significant behavior change. Previously, we only saved to the DB the configurations explicitly set by the user. Now, in addition to that, we’re also saving values enriched by Nuclio. This could lead to those values remaining in the config even after the defaults are changed in either Nuclio or MLRun.

@rokatyy The function returned from the deployment process does not directly save any values enriched by Nuclio. While we invoke the `deploy_config` function on the Nuclio side, the changes made during the deployment (such as adding ingress settings or modifying the service type) are temporary. These changes only affect the deployment configuration passed to Nuclio at runtime. Once the function is deployed, these modifications are applied within the Nuclio environment, but they do not permanently alter the underlying function spec.

mlrun/runtimes/nuclio/function.py
rokatyy: after online discussion, we have check that it fine, so closing this thread

rokatyy (approved):

TomerShor (approved):

PR #7637, opened 2025-04-07
[API] Upgrade mlrun-api to python 3.11 [1.9.x]

alonmr:
https://iguazio.atlassian.net/browse/ML-9582


liranbg (approved):

moranbental (approved):
LGTM

Makefile
```diff
@@ -30,7 +30,7 @@ MLRUN_NO_CACHE ?=
 MLRUN_ML_DOCKER_IMAGE_NAME_PREFIX ?= ml-
 # do not specify the patch version so that we can easily upgrade it when needed - it is determined by the base image
 # mainly used for mlrun, base and mlrun-gpu. mlrun API version >= 1.3.0 should always have python 3.9
-MLRUN_PYTHON_VERSION ?= 3.9
+MLRUN_PYTHON_VERSION ?= 3.11
```
jond01, line 33: I suggest using this Python version in the `uv pip compile` targets for the relevant images - with `mlrun/mlrun-api` as the leading example.

The following unchanged setting in `pyproject.toml` should define the minimal Python version:
https://docs.astral.sh/uv/reference/settings/#pip_python-version
https://github.com/mlrun/mlrun/blob/f9dd4303ba335deb5a14df4ff2b442b59d09a100/pyproject.toml#L37-L38
(Note that the comment should be changed there)

But it can be overridden with this flag:
https://docs.astral.sh/uv/reference/cli/#uv-pip-compile--python-version
https://github.com/mlrun/mlrun/blob/f9dd4303ba335deb5a14df4ff2b442b59d09a100/Makefile#L884-L891

Obviously this comment is irrelevant for images that must support Python 3.9 as well, such as `mlrun/mlrun`.

dockerfiles/mlrun-api/locked-requirements.txt
jond01, line 1: As explained in the previous comment, this lock file still contains many unused specifications.
For example:
https://github.com/mlrun/mlrun/blob/f96ab2f1f393ba0963d314f612afb9a37f2d88c2/dockerfiles/mlrun-api/locked-requirements.txt#L658
It is unchanged here but can be removed with the proper resolution.

Makefile
```diff
@@ -947,6 +947,7 @@ upgrade-mlrun-kfp-deps-lock: verify-uv-version ## Upgrade mlrun-kfp locked requi
 	uv pip compile \
 		requirements.txt \
 		dockerfiles/mlrun-kfp/requirements.txt \
+		--python-version 3.9 \
```
jond01, line 950: Why is that needed?

dockerfiles/test/requirements.txt
```diff
@@ -1,3 +1 @@
-# This is here because it is yet GAed and therfor do not appear on and of MLRun's
-# requirement files.
-mlrun-pipelines-kfp-v2~=0.4.2; python_version >= "3.11"
+mlrun-pipelines-kfp-v1-8[kfp]~=0.4.0; python_version < "3.11"
```
jond01: Missing newline

Probably this requirement is worth an explanation comment

Makefile
alonmr: mlrun/mlrun-kfp runs only with python 3.9

jond01 (approved):

liranbg (approved):

PR #7659, opened 2025-04-09
[Artifacts] Ensure 'latest' tag appears first when listing artifacts [1.8.x]

moranbental:
When listing artifacts, if an artifact has multiple tags, the tags should be ordered with the 'latest' tag first, followed by the rest of the tags ordered by tag ID in descending order.


yaelgen (approved):
LGTM

quaark (approved):

PR #6615, opened 2024-10-31
[Microservices] Project structure refactor

alonmr:
## ⚠️  Project structure change ⚠️

Project restructure preparing the ground for microservices infrastructure implementation.
The new structure is:
```
mlrun/ # Client side code
server/
       |______ py/
       |       |__ services/
       |           |_______ api/
       |                   |___ test/
       |                   |___ .... # Business logic
       |           |_______ alerts/
       |           |_______ ...
       |______ go/
       |       |__ go.work
       |       |__ mlrun-go/
       |          |________common/
       |          |________framework/
       |       |__ services/
       |           |_______ log-collector/ # pkg
       |       |_____ Makefile
       |______ common/
       |       |_____ proto/
       |       |_____ builders/
       |       |_____ Makefile
```

## Environment changes you need after pulling ❗ 
On PyCharm: Right click on `server/py` folder ➜ mark directory as ➜ source root
For other IDEs: Configure `PYTHONPATH=<project-root>:<project-root>/server/py`

Influenced by https://github.com/mlrun/mlrun/pull/5284

PR #7670, opened 2025-04-10
[Tests] Fix feature store system tests

gtopper:
Fixes the following system tests that broke following #6615:
```
TestFeatureStore::test_stream_source
TestFeatureStore::test_preview_saves_changes
TestFeatureStore::test_deploy_ingestion_service_with_different_targets
```


tests/system/feature_store/assets/sample_function.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2023 Iguazio
```
gtopper, line 1: This file is copied, so the copyright year is correct.

Eyal-Danieli (approved):

PR #7682, opened 2025-04-20
[Artifacts] Fix artifact listing order by error with partition_by and limit [1.8.x]

moranbental:
This PR fixes a query error that occurred when listing artifacts with both `partition_by` and `limit`, but without filtering by `tag`.

Although the tag was joined in a subquery, the outer ORDER BY clause referred to artifacts_v2_tags.name, which wasn't available in that scope — leading to the error:
`Unknown column 'artifacts_v2_tags.name' in 'order clause'`

The fix ensures that aliases are used properly so columns referenced in ORDER BY are available in the final query.
https://iguazio.atlassian.net/browse/ML-9849


server/py/framework/db/sqldb/db.py
```diff
@@ -1658,14 +1658,14 @@ def _find_artifacts(
...
-        # apply all filters and limits
+        # Create a subquery that selects only the artifact IDs along with tag metadata.
+        # The tag name and tag ID are explicitly aliased as 'name' and 'tag_id' so they can be
+        # referenced in window functions, ordering, and outer queries (especially for sorting by tag).
         query = session.query(ArtifactV2).with_entities(
             ArtifactV2.id,
-            ArtifactV2.Tag.name,
-            # Include tag ID (as 'tag_id') to enable sorting by tag creation order DESC.
-            # The alias is required to reference it later in subqueries and outer queries.
+            ArtifactV2.Tag.name.label(tag_name_alias),
```
adamdelman, line 1668: Is there any reason we use labels here? can't we just use the columns directly?

liranbg (approved):

adamdelman (commented):
One minor question

server/py/framework/db/sqldb/db.py
moranbental: @adamdelman Yes, we define labels because after the joins, we need to order by the result of the join — specifically, `anon_1.tag_id` and `anon_1.name`. We define labels so we can reference them in the order_by clause. This is necessary because SQLAlchemy does not allow direct references to aliased columns in order_by

PR #7695, opened 2025-04-21
[Alerts] Replace slack webhook with dummy URL

Yacouby:
The `test_alert_operations` integration test has been failing intermittently in CI due to timeouts when attempting to send Slack notifications.
Logs show that the notification was triggered and entered the async push phase, but the process hung without completing or logging a success message. 
This suggests the failure happened during the HTTP request to the Slack webhook which is casued by the test using a real Slack webhook URL, making it dependent on network stability and Slack’s availability, which introduces flakiness in our tests.

Jira:
https://iguazio.atlassian.net/browse/ML-9786


alonmr (changes requested):
There are more places where we use `https://hooks.slack.com/services/`
lets change them as well

tests/integration/sdk_api/alerts/test_alerts.py
```diff
@@ -692,9 +692,6 @@ def _create_alerts_test(self, project_name, alert1, alert2):
                     "severity": "warning",
                     "when": ["now"],
                     "condition": "failed",
-                    "secret_params": {
-                        "webhook": "https://hooks.slack.com/services/",
-                    },
```
TomerShor, line 697: change the notification name to `git` too while you're at it (instead of `slack_jobs`)

TomerShor (changes requested):

TomerShor (approved):

PR #7702, opened 2025-04-22
[Bugfix] Properly serialize datetime when sending webbook notifications [1.8.x]

adamdelman:
https://iguazio.atlassian.net/browse/ML-9856

### DIFF
--- a/mlrun/utils/notifications/notification/webhook.py
+++ b/mlrun/utils/notifications/notification/webhook.py
@@ -16,6 +16,7 @@ import re
 import typing
 
 import aiohttp
+import orjson
 
 import mlrun.common.schemas
 import mlrun.lists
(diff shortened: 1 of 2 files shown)


quaark (changes requested):
nice! If we are going with this approach, maybe lets separate the building of the request body from the `push` method to a different method, and then you may add some unit test for that method with a few offending fields making sure the request body gets serialized as you expect it to.

liranbg:
> nice! If we are going with this approach, maybe lets separate the building of the request body from the `push` method to a different method, and then you may add some unit test for that method with a few offending fields making sure the request body gets serialized as you expect it to.

id actually UT the push of webkhook notification class, no need a different method / testing orjson

liranbg (approved):

PR #7729, opened 2025-04-27
[Bugfix] Don't double serialize json body in webhook notification body [1.8.x]

adamdelman:
Fix double‐serialization in webhook notifications
Use aiohttp’s json_serialize hook with orjson instead of manually dumping the payload, and update tests to assert on dicts rather than serialized strings.

https://iguazio.atlassian.net/browse/ML-9879

### DIFF
--- a/tests/utils/test_notifications.py
+++ b/tests/utils/test_notifications.py
@@ -1651,75 +1651,19 @@ class DummySession:
         self,
         method: str,
         str_or_url: StrOrURL,
-        *,
-        params: Query = None,
-        data: Any = None,
-        json: Any = None,
-        cookies: Optional[LooseCookies] = None,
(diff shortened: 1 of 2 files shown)


tests/utils/test_notifications.py
```diff
@@ -1649,21 +1638,96 @@ async def put(
...
+            ssl=ssl,
+        )
+        return DummyResponse()
+
+    async def _request(
+        self,
+        method: str,
+        str_or_url: StrOrURL,
+        *,
+        params: Query = None,
```
alonmr: why not just use kwargs on everything below this one?

alonmr (approved):
Looks good! Minor comment and please add some description to explain the problem and solution

alonmr (approved):

PR #7730, opened 2025-04-27
[Artifacts] Delete objects in batches to avoid deletion timeouts

Yacouby:
Batch the deletion of objects (artifacts, functions, runs, etc.) to avoid timeouts and improve performance when deleting a large number of records.
Previously, deletions could get stuck or fail when handling many objects at once.

Jira:
https://iguazio.atlassian.net/browse/ML-9550


server/py/framework/db/sqldb/db.py
```diff
@@ -3113,17 +3113,69 @@ def skip_deletion():
...
-        log_kwargs = {
-            "deletions_count": deletions_count,
-            "main_table": main_table,
-            "project": project,
-            "main_table_identifier": main_table_identifier,
-        }
-        logger.debug("Removed rows from table", **log_kwargs)
-        session.commit()
-        return deletions_count
+        total_deleted = SQLDB._delete_table_in_batches(
```
TomerShor: use `self._delete...`, and remove the `staticmethod` from this function if needed.

server/py/framework/db/sqldb/db.py
```diff
@@ -3113,17 +3113,69 @@ def skip_deletion():
...
+            session, main_table, where_clause
+        )
+        # redundant log ?
```
TomerShor: Not redundant because you want to know that the operation has ended (replace with "info")

TomerShor (changes requested):

server/py/framework/db/sqldb/db.py
```diff
@@ -3113,17 +3113,65 @@ def skip_deletion():
...
+        """
+        Delete rows from a table in batches based on ID ordering.
+        :param session: SQLAlchemy session.
+        :param table: SQLAlchemy ORM model/table to delete from.
+        :param where_clause: SQLAlchemy WHERE clause.
+        :return: Total number of deleted rows.
+        """
+        last_id = 0
+        total_deleted = 0
+        batch_size = mlrun.mlconf.artifacts.limits.deletion_batch_size
```
TomerShor: This function does not only delete artifacts, but other objects too, so i think the config is misplaced.
Since this is mostly called when deleting project resources, perhaps it can be moved to:
```
mlrun.mlconf.httpdb.projects.resource_deletion_batch_size
```

server/py/framework/db/sqldb/db.py
```diff
@@ -3113,17 +3113,65 @@ def skip_deletion():
...
+
+        while True:
+            ids_to_delete = (
+                session.query(table.id)
+                .filter(where_clause, table.id > last_id)
```
TomerShor, line 3146: is `table.id > last_id` needed if we already delete all of the ids until (and including) `last_id` ?

server/py/services/api/tests/unit/db/test_artifacts.py
```diff
@@ -1282,6 +1283,55 @@ def test_delete_artifact_with_latest_tag_and_iteration_not_0(self):
...
+        assert len(artifacts) == 15
+
+        # Set small batch size to force batch deletion
+        mlrun.mlconf.artifacts.limits.deletion_batch_size = 5
+
+        where_clause = ArtifactV2.project == project
+
+        with unittest.mock.patch.object(
+            self._db_session, "execute", wraps=self._db_session.execute
+        ) as mock_execute:
```
TomerShor, line 1312: nice touch with the `wraps` 👌

TomerShor (changes requested):

server/py/framework/db/sqldb/db.py
Yacouby: the `table.id > last_id` condition ensures we progress between batches by fetching only rows with higher IDs. Without it, we might re-fetch deleted rows or get stuck in an infinite loop.

TomerShor (approved):
LGTM

server/py/framework/db/sqldb/db.py
```diff
@@ -3113,17 +3113,65 @@ def skip_deletion():
...
+            result = session.execute(delete_stmt)
+            session.commit()
+
+            last_id = id_values[-1]
+            total_deleted += result.rowcount
+
+            logger.debug(
+                "Deleted batch from table",
+                batch_size=len(id_values),
+                total_deleted=total_deleted,
```
moranbental, line 3171: For debugging 
```suggestion
                total_deleted=total_deleted,
                last_id= last_id,
```

moranbental (approved):
Looks great! Just one suggestion

TomerShor (approved):

PR #7732, opened 2025-04-28
[Requirements] Uphold v3io-frames to protobuf constraint for python 3.9

alonmr:
old v3io frames require protobuf 3.20 as does kfp
this change forces protobuf 3.20 when using python 3.9 to stay consistent with mlrun/mlrun-kfp image


liranbg (changes requested):
Kfp is an optional requirement of mlrun. Move it accordingly

alonmr:
> Kfp is an optional requirement of mlrun. Move it accordingly

This change is not due to kfp it's due to v3io-frames. The mention to kfp is just to make the python 3.9 client env consistent with mlrun-kfp image. The other option is to bump v3io-frames but that means we won't be able to install mlrun client on mlrun/mlrun-kfp without customizations.

PR #7743, opened 2025-04-29
[Requirements] Fix package test by restricting grpcio-tools version

gtopper:
Which is resolved to the oldest version possible by pip 25.1.


alonmr:
IMO we should limit pip until they resolve the issue

gtopper:
As per above, #7742 was merged instead.

PR #7762, opened 2025-05-04
[API] Remove default project behavior

Yacouby:
This PR addresses the removal of implicit `default_project` usage from the server side, in alignment with the goal of enforcing explicit project handling across the system.

Key Changes:

- Error Handling: Introduced a new error class `MLRunMissingProjectError` to clearly signal when a required project parameter is missing.
- Function Signatures: Removed default values for the project parameter in relevant DB functions. Calls to these functions now explicitly require a project value. An appropriate error is raised if the project is not provided.
- Test Updates: Updated unit and integration tests that previously relied on implicit `default_project` behavior by explicitly setting the project in each case.
- Code Readability: Refactored affected function calls to use keyword arguments instead of positional ones where appropriate, improving clarity and reducing the chance of future mistakes.

Jira:
https://iguazio.atlassian.net/browse/ML-9837


quaark (changes requested):
Great stuff!
One thing only - I'd add a couple of UTs in the db, and crud and api that purposefully give `None` projects and expect the new exception.

server/py/services/api/tests/unit/api/test_submit.py
```diff
@@ -282,10 +282,6 @@ def test_submit_job_service_accounts(
     project = "my-proj1"
     services.api.tests.unit.api.utils.create_project(client, project)
 
-    # must set the default project since new_function creates the function object and ignores the project parameter.
-    # Instead, the function always gets the default project name. This may be a bug, need to check.
```
TomerShor, line 286: Did you check that the project is not ignored?

server/py/services/api/tests/unit/crud/test_functions.py
```diff
@@ -98,3 +98,35 @@ def test_update_functions_with_api_gateway_url(db: sqlalchemy.orm.Session):
...
+    with pytest.raises(mlrun.errors.MLRunMissingProjectError):
+        services.api.crud.Functions().get_function(
+            db, name="name", project=None, tag="latest"
+        )
+
+    # list with missing project should raise error
+    with pytest.raises(mlrun.errors.MLRunMissingProjectError):
+        services.api.crud.Functions().list_functions(
+            db, name="name", project=None, tag="latest"
+        )
```
TomerShor: validate you can get the function with the valid projects

server/py/services/api/tests/unit/crud/test_runs.py
```diff
@@ -282,6 +282,42 @@ def test_run_abortion_failure(self, db: sqlalchemy.orm.Session):
         assert run["status"]["state"] == mlrun.common.runtimes.constants.RunStates.error
         assert run["status"]["error"] == "Failed to abort run, error: BOOM"
 
+    @pytest.mark.asyncio
+    async def test_store_and_get_run_missing_project(self, db: sqlalchemy.orm.Session):
```
TomerShor: Why async?

server/py/services/api/tests/unit/db/test_artifacts.py
```diff
@@ -43,6 +43,7 @@
 
 class TestArtifacts(TestDatabaseBase):
     def test_list_artifact_name_filter(self):
+        project = "some-project"
```
TomerShor: Since this file is using a test class, you can define the project name as a class variable and access it in all tests using `self.project` instead of defining it again in every test.

server/py/services/api/tests/unit/db/test_functions.py
```diff
@@ -29,21 +29,30 @@
 
 class TestFunctions(TestDatabaseBase):
     def test_store_function_default_to_latest(self):
+        project = "some-project"
```
TomerShor: same in this class.

TomerShor (changes requested):
Looks great overall, have some minor improvement comments

server/py/services/api/tests/unit/api/test_submit.py
Yacouby: yes, it is not ignored

server/py/services/api/tests/unit/crud/test_runs.py
Yacouby: by mistake, removed

mlrun/errors.py
```diff
@@ -230,6 +230,10 @@ class MLRunTSDBConnectionFailureError(MLRunHTTPStatusError, ValueError):
     error_status_code = HTTPStatus.BAD_REQUEST.value
 
 
+class MLRunMissingProjectError(MLRunBadRequestError):
+    pass
```
moranbental: instead of specifying the message when raising this error, you can define a default message in the exception class itself
```suggestion
class MLRunMissingProjectError(MLRunBadRequestError):
    def __init__(self, message=None):
        default_message = "Project must be provided"
        super().__init__(message or default_message)
```

and call it as MLRunMissingProjectError()

moranbental (changes requested):
Looks good just one suggestion

TomerShor (approved):
Good stuff 👌

moranbental (approved):

PR #7766, opened 2025-05-04
[Serving] Change `ModelRunnerStep:add_model`, added model endpoints creation

royischoss:
https://iguazio.atlassian.net/browse/ML-9890
Updated `ModelRunnerStep:add_model`  SDK
Updated Model Endpoints creation in deployment extracting models from `ModelRunnerStep`


mlrun/serving/states.py
```diff
@@ -1005,24 +1024,91 @@ def __init__(
...
+            2. Create a new model endpoint with the same name and set it to `latest`.
+        :param inputs:              list of the model inputs (e.g. features) ,if provided will override the inputs that
+                                    been configured in the model artifact, please note that those inputs need to be
+                                    equal in length and order to the inputs that Model predict method expects
+        :param outputs:             list of the model outputs (e.g. labels) ,if provided will override the outputs that
+                                    been configured in the model artifact, please note that those outputs need to be
+                                    equal to the model serving function outputs (length, and order)
+        :param input_path:          input path inside the user event, expect scopes to be defined by dot notation
+                                    (e.g "inputs.my_model_inputs"). expects list or dictionary type object in path.
+        :param override:            bool allow override existing model.
```
davesh0812: existing model on this ModelRunnerStep right?

mlrun/serving/states.py
```diff
@@ -1005,24 +1024,91 @@ def __init__(
             **kwargs,
         )
 
-    def add_model(self, model: Union[str, Model], **model_parameters) -> None:
+    def add_model(
+        self,
+        endpoint_name: str,
+        model_class: Union[str, Model],
+        model_artifact: Optional[Union[str, mlrun.artifacts.Artifact]] = None,
```
```suggestion
        model_artifact: Optional[Union[str, mlrun.artifacts.ModelArtifact]] = None,
```

mlrun/serving/states.py
```diff
@@ -1255,6 +1341,18 @@ def add_step(
...
+        if isinstance(class_name, ModelRunnerStep) or class_name == "ModelRunnerStep":
+            if self.parent is not None:
+                root = self.parent
+            else:
+                root = self
+            for model_endpoint in step.class_args["models"].keys():
+                if model_endpoint in root.model_endpoints:
+                    raise GraphError(
+                        "ModelRunnerStep points to existed model in graph using the same model_endpoint name"
+                    )
```
davesh0812: Mashu Kaza
```suggestion
                    raise GraphError(
                        "The graph is already contains model endpoint named {model_endpoint}"
                    )
```

mlrun/serving/states.py
```diff
@@ -1255,6 +1341,18 @@ def add_step(
...
+                else:
+                    root.update_model_endpoint(model_endpoint)
```
davesh0812: You should update the model endpoints of the graph only after you add the ModelRunnerStep to the graph.
So you need to do it after the for loop and to add all together

mlrun/serving/states.py
```diff
@@ -1678,6 +1776,34 @@ class RootFlowStep(FlowStep):
...
+    @property
+    def model_endpoints(self) -> list[str]:
+        return self._models
+
+    @model_endpoints.setter
+    def model_endpoints(self, models: list[str]):
+        self._models = models
+
+    def update_model_endpoint(self, model_endpoint_name: str):
+        self._models.append(model_endpoint_name)
```
davesh0812: Relevant to the prev comment
```suggestion
    def update_model_endpoint(self, model_endpoint_names: list):
        self._models.extend(model_endpoint_names)
```

mlrun/common/schemas/model_monitoring/constants.py
```diff
@@ -446,6 +446,17 @@ class ModelEndpointMonitoringMetricType(StrEnum):
...
 
+class ServingMonitoringData(StrEnum):
+    MONITORING_DATA = "monitoring_data"
+    INPUTS = "inputs"
+    OUTPUTS = "outputs"
+    INPUT_PATH = "input_path"
+    CREATION_STRATEGY = "creation_strategy"
+    LABELS = "labels"
+    MODEL_PATH = "model_path"
+    MODEL_ENDPOINT_UID = "model_endpoint_uid"
```
davesh0812: move to mlrun/common/schemas/serving.py and split as we spoke

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -1779,6 +1799,75 @@ def _get_trigger_frequency(base_period: int) -> int:
...
+                            sampling_percentage=sampling_percentage,
+                            uid=uid,
+                            label_names=monitoring_data[endpoint_name].get(
+                                mm_constants.ServingMonitoringData.OUTPUTS
+                            ),
+                            model_path=monitoring_data[endpoint_name].get(
+                                mm_constants.ServingMonitoringData.MODEL_PATH, ""
+                            ),
+                        ),
+                        monitoring_data[endpoint_name].get("creation_strategy"),
```
davesh0812: use constants

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -1779,6 +1799,75 @@ def _get_trigger_frequency(base_period: int) -> int:
...
+        ]
+    ]:
+        model_endpoints_instructions = []
+        monitoring_data = model_runner.class_args.get(
+            mm_constants.ServingMonitoringData.MONITORING_DATA, {}
+        )
+        for endpoint_name, (
+            model_class,
+            model_parameter,
+        ) in model_runner.class_args.get("models", {}).items():
```
davesh0812: I think you can run over monitoring_data.items()

davesh0812 (commented):
Looks Good.
Review the core code for now (without tests)

mlrun/serving/states.py
```diff
@@ -1678,6 +1776,34 @@ class RootFlowStep(FlowStep):
     kind = "root"
     _dict_fields = ["steps", "engine", "final_step", "on_error"]
```
theSaarco: You need to add the new `model_endpoints` member here as well, so it will be serialized to-dict and back.

mlrun/serving/states.py
theSaarco: Why not call this `add_model_endpoints`? After all, this doesn't really update any existing MEPs, it just adds more.

mlrun/serving/states.py
```diff
@@ -402,6 +403,24 @@ def to(
             class_args=class_args,
             model_endpoint_creation_strategy=model_endpoint_creation_strategy,
         )
+
+        if isinstance(class_name, ModelRunnerStep) or class_name == "ModelRunnerStep":
+            if self.parent is not None:
+                root = self.parent
+            else:
+                root = self
+            if not isinstance(root, RootFlowStep):
```
theSaarco: This is a bit strange to me - what if the graph has root -> step1 -> step2, and you do `step2.to()` with a model runner step? The parent will be step1 which is not a root step, but this is still a valid operation - you just need to go back to the actual root of the graph.

mlrun/serving/states.py
```diff
@@ -402,6 +403,24 @@ def to(
...
+                raise GraphError(
+                    "ModelRunnerStep should be added to 'Flow' topology graph only"
+                )
+            for model_endpoint in step.class_args["models"].keys():
+                if model_endpoint in root.model_endpoints:
+                    raise GraphError(
+                        "ModelRunnerStep points to existed model in graph using the same model_endpoint name"
```
theSaarco:
```suggestion
                        f"It is invalid to add a model endpoint with an already existing name - {model_endpoint}"
```

mlrun/serving/states.py
```diff
@@ -402,6 +403,24 @@ def to(
...
         )
+
+        if isinstance(class_name, ModelRunnerStep) or class_name == "ModelRunnerStep":
+            if self.parent is not None:
+                root = self.parent
+            else:
+                root = self
+            if not isinstance(root, RootFlowStep):
+                raise GraphError(
+                    "ModelRunnerStep should be added to 'Flow' topology graph only"
```
theSaarco:
```suggestion
                    "ModelRunnerStep can be added to 'Flow' topology graph only"
```

mlrun/serving/states.py
```diff
@@ -1005,24 +1024,91 @@ def __init__(
...
+            2. Create a new model endpoint entry and set it as `latest`.
+            * **inplace** (default):
+            1. If model endpoints with the same name exist, update the `latest` entry.
+            2. Otherwise, create a new entry.
+            * **archive**:
+            1. If model endpoints with the same name exist, preserve them.
+            2. Create a new model endpoint with the same name and set it to `latest`.
+        :param inputs:              list of the model inputs (e.g. features) ,if provided will override the inputs that
+                                    been configured in the model artifact, please note that those inputs need to be
+                                    equal in length and order to the inputs that Model predict method expects
```
theSaarco:
```suggestion
                                    equal in length and order to the inputs that model_class predict method expects
```

mlrun/serving/states.py
```diff
@@ -446,6 +451,25 @@ def set_flow(
...
 
+    def verify_model_runner_step(self, step: "ModelRunnerStep"):
+        if self.parent is not None:
+            root = self.parent
+        else:
+            root = self
+        if not isinstance(root, RootFlowStep):
+            raise GraphError(
+                "ModelRunnerStep should be added to 'Flow' topology graph only"
+            )
```
theSaarco: I'm not sure if this check is right - what if we have a graph root -> step1 -> step2, and you do `step2.to()` with a MRS. The parent is step1 which is not a `RootFlowStep`, but the graph is still a flow graph. 
I think you need to go all the way to the root step to do this validation, not just look for 1st level parent.

mlrun/serving/states.py
```diff
@@ -1005,24 +1029,92 @@ def __init__(
...
+        if endpoint_name in models and not override:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"Model with name {endpoint_name} already exists in this ModelRunnerStep."
+            )
+
+        model_parameters["name"] = endpoint_name
+        monitoring_data = self.class_args.get(
+            schemas.ModelRunnerStepData.MONITORING_DATA, {}
+        )
+        model_class = model_class.__class__ if isinstance(model_class, Model) else model_class
```
theSaarco: When you do this, you lose any parameters already configured in the `Model` instance passed, all you maintain is the class name and the user needs to pass any relevant parameter in the `model_parameters` dict. This is not what users would expect from passing a class instance.

theSaarco (changes requested):
Nice. Have some comments and fixes needed.

mlrun/serving/states.py
royischoss: Removed this line I will make sure that providing initialized object survive serialization (with all class variable) as we expect

mlrun/serving/states.py
```diff
@@ -446,6 +451,30 @@ def set_flow(
     def supports_termination(self):
         return False
 
+    def verify_model_runner_step(self, step: "ModelRunnerStep"):
```
davesh0812, line 453: please add documentation

mlrun/serving/states.py
```diff
@@ -1005,24 +1034,87 @@ def __init__(
             **kwargs,
         )
 
-    def add_model(self, model: Union[str, Model], **model_parameters) -> None:
+    def add_model(
+        self,
+        endpoint_name: str,
+        model_class: str,
```
davesh0812, line 1045: please add here TODO and mention the story ticket number

mlrun/serving/states.py
```diff
@@ -1676,7 +1770,35 @@ class RootFlowStep(FlowStep):
...
+            name,
+            steps,
+            after,
+            engine,
+            final_step,
+        )
+        self._models = []
+
+    @property
+    def model_endpoints(self) -> list[str]:
```
```suggestion
    def model_endpoints_names(self) -> list[str]:
```

mlrun/serving/states.py
```diff
@@ -446,6 +451,30 @@ def set_flow(
     def supports_termination(self):
         return False
 
+    def verify_model_runner_step(self, step: "ModelRunnerStep"):
+        current_node = self
+        while current_node.parent is not None:
+            current_node = self.parent
+
+        root = current_node
```
```suggestion
        root = self.parent
        while current_node.parent is not None:
            root = self.parent
```

mlrun/serving/states.py
```diff
@@ -446,6 +451,30 @@ def set_flow(
...
+        if not isinstance(root, RootFlowStep) or not isinstance(self, RootFlowStep):
```
```suggestion
        if not isinstance(root, RootFlowStep):
```

mlrun/serving/states.py
```diff
@@ -1255,6 +1347,8 @@ def add_step(
             model_endpoint_creation_strategy=model_endpoint_creation_strategy,
             class_args=class_args,
         )
+        if isinstance(class_name, ModelRunnerStep) or class_name == "ModelRunnerStep":
```
davesh0812: I think you can move this check to be the first thing you do inside `verify_model_runner_step` to avoid code duplication

mlrun/serving/states.py
```diff
@@ -402,6 +403,10 @@ def to(
             class_args=class_args,
             model_endpoint_creation_strategy=model_endpoint_creation_strategy,
         )
+
+        if isinstance(class_name, ModelRunnerStep) or class_name == "ModelRunnerStep":
```
davesh0812: same here

mlrun/serving/states.py
```diff
@@ -446,6 +451,30 @@ def set_flow(
...
+        if not isinstance(root, RootFlowStep) or not isinstance(self, RootFlowStep):
+            raise GraphError(
+                "ModelRunnerStep can be added to 'Flow' topology graph only"
+            )
+        step_model_endpoints = list(
+            step.class_args[schemas.ModelRunnerStepData.MODELS].keys()
+        )
+        # Get all model_endpoints names that are in both lists
+        common_endpoints_names = list(
+            set(root.model_endpoints) & set(step_model_endpoints)
```
davesh0812: tilll

davesh0812 (commented):
Few more comments

davesh0812 (approved):
LGTM

mlrun/serving/states.py
```diff
@@ -446,6 +450,36 @@ def set_flow(
...
+        """
+        Verify ModelRunnerStep, can be part of Flow graph and models can not repeat in graph.
+        :param step: ModelRunnerStep to verify
+        """
+        if not isinstance(step, ModelRunnerStep):
+            return
+
+        root = self
+        while root.parent is not None:
+            root = self.parent
```
theSaarco:
```suggestion
            root = root.parent
```
Infinite loop otherwise... 😄

tests/system/model_monitoring/test_model_monitoring.py
```diff
@@ -573,6 +574,44 @@ def test_mep_with_model(self):
...
+        function.save(versioned=False)
+        graph = function.set_topology("flow", engine="async")
+        model_runner_step = mlrun.serving.states.ModelRunnerStep(name="model-runner")
+        model_runner_step.add_model(
+            model_class="IncModel", endpoint_name="my-model-1", inc=1
+        )
+        model_runner_step.add_model(
+            model_class="IncModel", endpoint_name="my-model-2", inc=2
+        )
+        graph.to(model_runner_step, "runner").respond()
```
theSaarco: Please add a small step before that, maybe even a step that does nothing, just to verify that the root-finding iteration really works (it would catch the bug you currently have there).

theSaarco (changes requested):
Looks good, but there's a bug that needs fixing first.

mlrun/serving/states.py
royischoss: good catch thanks

theSaarco (approved):
LGTM.

PR #7781, opened 2025-05-05
[Docs] Fix TDengine and Kafka for model monitoring in K8s install

jillnogold:


docs/install/kubernetes.md
```diff
@@ -177,10 +177,27 @@ python -m ipykernel install --user --name myenv --display-name "Python (myenv)"
...
+    port=6041,
+    user="username",
+    password="<tdengine-password>",
+)
+project.register_datastore_profile(tsdb_profile)
+
+# Create and register stream profile
+stream_profile = DatastoreProfileKafkaSource(
+    name="my-kafka",
+    brokers=["kafka://kafka-stream:9092"],
```
jond01: That's not how a broker URI looks like

docs/install/kubernetes.md
```diff
@@ -177,10 +177,27 @@ python -m ipykernel install --user --name myenv --display-name "Python (myenv)"
 TDengine and Kafka are part of the default CE installations. These are the default TDengine and Kafka installation values. It's recommended to change the user/password.
 
 ```py
-stream_path = "kafka://kafka-stream:9092"
-tsdb_connection = "taosws://root:taosdata@tdengine-tsdb:6041"
+# Create and register TSDB profile
+tsdb_profile = DatastoreProfileTDEngine(
+    name="my-tdengine",
+    host="<taosws://root:taosdata@tdengine-tsdb>",
```
jond01: Same

jond01 (changes requested):

jond01 (approved):
Looks fine, thanks.
Since I guess we set the default values in the CE helm chart, we can probably use them here. But it can be left for the future

alonmr (approved):

PR #7802, opened 2025-05-08
[Docs] add logo

gilad-shaham:


docs/requirements.txt
```diff
@@ -15,3 +15,4 @@ sphinx_design~=0.4.1
 sphinx-reredirects~=0.1.2
 sphinx-version-warning~=1.1
 sphinxcontrib-mermaid~=0.9.2
+sphinx-image-inverter~=1.1, >=1.1.3
```
rokatyy:
```suggestion
sphinx-image-inverter~=1.1.3
```

docs/conf.py
```diff
@@ -191,6 +192,8 @@ def current_version() -> str:
 
 smartquotes = False
 
+inverter_all = False
+
```
rokatyy, line 197: what it does? Could you please add a comment above

gilad-shaham: disables by default inverting all images, so only the ones we explicitly set will do that

rokatyy (approved):

jillnogold (approved):

PR #7804, opened 2025-05-08
[Github] Unpinned tag for a non-immutable Action

moranbental:
Fix security vulnerabilities

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,9 +43,9 @@ jobs:
       matrix:
         python-version: ["3.9"]
     steps:
-    - uses: actions/checkout@v4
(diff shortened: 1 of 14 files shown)


.github/workflows/ci.yaml
```diff
@@ -43,9 +43,9 @@ jobs:
       matrix:
         python-version: ["3.9"]
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # 4.2.2
```
liranbg: anything starts with `action/` is considered safe as it originates from github themselves. therefor no need to pin to hash. can be left with tag. that would save us troubles updating steps

liranbg (changes requested):

PR #7668, opened 2025-04-10
[Model Monitoring] Align query parameter names with convention

gtopper:
[ML-9792](https://iguazio.atlassian.net/browse/ML-9792)

[ML-9792]: https://iguazio.atlassian.net/browse/ML-9792?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7811, opened 2025-05-11
[Model Monitoring] controller get last_requests directly from tsdb_connector

davesh0812:
1. controller get last_requests directly from tsdb_connector
2. change default of tsdb-metrics to false (SDK)

https://iguazio.atlassian.net/browse/ML-9865


mlrun/model_monitoring/controller.py
```diff
@@ -721,16 +728,20 @@ def push_regular_event_to_controller_stream(self) -> None:
...
-                        self.endpoint_to_regular_event,
-                        endpoint,
-                        policy,
-                        set(applications_names),
-                        schedule_file,
-                    ): endpoint
-                    for endpoint in endpoints
-                }
+                for endpoint in endpoints:
+                    last_request = last_request_dict.pop(endpoint.metadata.uid, None)
```
assaf758: why pop() and not a simple last_request_dict.get(endpoint.metadata.uid)?

mlrun/db/httpdb.py
```diff
@@ -3888,8 +3888,8 @@ def patch_model_endpoint(
         attributes_keys = list(attributes.keys())
         attributes["name"] = name
         attributes["project"] = project
-        attributes["function-name"] = function_name or None
-        attributes["function-tag"] = function_tag or None
+        attributes["function_name"] = function_name or None
```
assaf758, line 3891: is this related?

davesh0812: No, but it is new bug that were introduced [here](https://github.com/mlrun/mlrun/pull/7668), and also was BP to 1.8.x

mlrun/model_monitoring/controller.py
davesh0812: It can be get as well… but why does it matter?

assaf758 (approved):

PR #7812, opened 2025-05-11
[SDK] Remove default project behavior

Yacouby:
Following the previous PR: https://github.com/mlrun/mlrun/pull/7762, which handles the removal of the default project from the server side, 
This PR removes all usage of `default_project` from the SDK, in alignment with the goal of enforcing explicit project handling across the system.

Key Changes:

- Configuration Update:  Added new `mlconf.active_project` to replace internal fallback usage of `default_project` in SDK flows.
- Code Refactor: Eliminated all silent fallbacks to a default project in SDK flows to enhance transparency and enforce explicit project handling. Replaced all uses of `mlconf.default_project` with `mlconf.active_project` across HTTPDB functions to align with the updated configuration.
- Test Updates: Refactored SDK-related integration tests that relied on `default_project`.

Jira:
https://iguazio.atlassian.net/browse/ML-9836


mlrun/__init__.py
```diff
@@ -120,7 +120,7 @@ def set_environment(
     :param mock_functions: set to True to create local/mock functions instead of real containers,
                            set to "auto" to auto determine based on the presence of k8s/Nuclio
     :returns:
-        default project name
+        default active project name
```
TomerShor:
```suggestion
       active project name
```

mlrun/common/helpers.py
```diff
@@ -14,9 +14,9 @@
 
 
 def parse_versioned_object_uri(
-    uri: str, default_project: str = ""
+    uri: str, active_project: str = ""
 ) -> tuple[str, str, str, str]:
-    project = default_project
+    project = active_project
```
TomerShor: In this case it should remain `default_project`, because the helper function defaults to the project in the argument if not found in the URI. Unrelated to the config's `default_project`.

mlrun/config.py
```diff
@@ -95,6 +95,7 @@
     # template for project default image name. Parameter {name} will be replaced with project name
     "default_project_image_name": ".mlrun-project-image-{name}",
     "default_project": "default",  # default project name
+    "active_project": "active",  # active project name
```
TomerShor: 1. Shouldn't `default_project` be removed?
2. `"active_project": "",`

mlrun/run.py
```diff
@@ -464,7 +464,7 @@ def new_function(
            f = new_function().run(task, handler=myfunction)
 
     :param name:     function name
-    :param project:  function project (none for 'default')
+    :param project:  function project (none for 'active')
```
TomerShor:
```suggestion
    :param project:  function project (none for the active project)
```

mlrun/run.py
```diff
@@ -640,7 +640,7 @@ def code_to_function(
     Learn more about :doc:`../../concepts/functions-overview`
 
     :param name:         function name, typically best to use hyphen-case
-    :param project:      project used to namespace the function, defaults to 'default'
+    :param project:      project used to namespace the function, defaults to 'active'
```
TomerShor:
```suggestion
    :param project:      project used to namespace the function, defaults to the active project
```

server/py/services/api/api/endpoints/projects.py
```diff
@@ -464,7 +464,7 @@ async def load_project(
     )
 
     # We must create the project before we run the remote load_project function because
-    # we want this function will be running under the project itself instead of the default project.
+    # we want this function will be running under the project itself instead of the active project.
```
TomerShor: This comment is not relevant anymore because neither the default and active project are relevant in the server now.

server/py/services/api/launcher.py
```diff
@@ -326,7 +326,7 @@ def enrich_runtime(
 
         # ensure the runtime has a project before we enrich it with the project's spec
         runtime.metadata.project = (
-            project_name or runtime.metadata.project or mlrun.mlconf.default_project
+            project_name or runtime.metadata.project or mlrun.mlconf.active_project
```
TomerShor: also here, IIUC the active project is valid in the client, but the server shouldn't enrich to it (needs an explicit project)

server/py/services/api/tests/unit/runtimes/test_spark.py
```diff
@@ -1139,7 +1139,7 @@ def test_get_offline_features(
                 target=ParquetTarget(),
             )
 
-        self.project = "default"
+        self.project = "active"
```
TomerShor: this is the result of mass replace(?), a project named "active" is not very clear (in tests too).

mlrun/db/httpdb.py
```diff
@@ -88,9 +88,9 @@ class HTTPRunDB(RunDBInterface):
 
     - Every object in MLRun exists in the context of a project (except projects themselves). When referencing an object
       through any API, a project name must be provided. The default for most APIs is for an empty project name, which
-      will be replaced by the name of the default project (usually ``default``). Therefore, if performing an API to
+      will be replaced by the name of the active project (usually ``active``). Therefore, if performing an API to
```
```suggestion
      will be replaced by the name of the active project (if set). Therefore, if performing an API to
```

mlrun/db/httpdb.py
```diff
@@ -88,9 +88,9 @@ class HTTPRunDB(RunDBInterface):
...
       list functions, for example, and not providing a project name - the result will not be functions from all
-      projects but rather from the ``default`` project.
+      projects but rather from the ``active`` project.
```
```suggestion
      projects but rather from the active project.
```

TomerShor (changes requested):
A general comment - we don't want to have a project named "active" instead of "default". Active project is a concept, not an actual project called "active".
This is also relevant to all tests changed in this PR - we don't want to create a project called "active" so the tests won't break. If you want to the tests to rely on the enrichment, a project should be created and set as active.

mlrun/config.py
Yacouby: 1. will be removed in a separate PR, since there is another separate PR handling for removing the default project for pipelines ( https://github.com/mlrun/mlrun/pull/7836 )

mlrun/feature_store/feature_set.py
```diff
@@ -414,7 +414,7 @@ def uri(self):
     def fullname(self) -> str:
         """full name in the form ``{project}/{name}[:{tag}]``"""
         fullname = (
-            f"{self._metadata.project or mlconf.default_project}/{self._metadata.name}"
+            f"{self._metadata.project or mlconf.active_project}/{self._metadata.name}"
```
moranbental: If there is no project and active_project is empty, we end up with "/name". 
Maybe it should return "name" without the slash.
Just like we check for the tag and add it only if it exists, maybe we should do the same for the project, add it only if it exists.

mlrun/feature_store/feature_vector.py
```diff
@@ -333,9 +333,7 @@ def status(self, status):
     @property
     def uri(self):
         """fully qualified feature vector uri"""
-        uri = (
-            f"{self._metadata.project or mlconf.default_project}/{self._metadata.name}"
-        )
+        uri = f"{self._metadata.project or mlconf.active_project}/{self._metadata.name}"
```
moranbental: maybe same here

moranbental (changes requested):
Looks good, just one comment.

TomerShor (approved):
LGTM

moranbental (approved):
LGTM

PR #7814, opened 2025-05-11
[Pipelines] Fix pipeline steps python version resolution

alonmr:
Since the workflow runner runs with python 3.9 for KFP dependencies, we need to explicitly specify the client python version. It is done with setting `MLRUN_PYTHON_VERSION` env var.
This PR also refactors some label enrichments for conventions.

https://iguazio.atlassian.net/browse/ML-9728


server/py/services/api/crud/workflows.py
```diff
@@ -234,6 +232,19 @@ def _label_run_object(
...
+        mlrun.runtimes.utils.enrich_run_labels(
+            labels,
+            [mlrun_constants.MLRunInternalLabels.owner],
+        )
+        client_python_version = runner.metadata.labels.get(
+            mlrun_constants.MLRunInternalLabels.client_python_version
+        )
+        # The runner always runs with python 3.9 therefore we need to explicitly specify the user client python version
+        if client_python_version:
+            runner.set_env("MLRUN_PYTHON_VERSION", client_python_version)
```
liranbg: this is true until we have kfp 2.0 client

mlrun/runtimes/utils.py
```diff
@@ -433,18 +433,35 @@ def enrich_function_from_dict(function, function_dict):
...
+    :return: The enriched labels dict
+    """
+    # Merge the labels with the labels enrichment extension
     labels_enrichment = {
-        RunLabels.owner: os.environ.get("V3IO_USERNAME") or getpass.getuser(),
+        mlrun_constants.MLRunInternalLabels.owner: os.environ.get("V3IO_USERNAME")
+        or getpass.getuser(),
         # TODO: remove this in 1.9.0
-        RunLabels.v3io_user: os.environ.get("V3IO_USERNAME"),
+        mlrun_constants.MLRunInternalLabels.v3io_user: os.environ.get("V3IO_USERNAME"),
```
liranbg: can we delete it on 1.10?

liranbg (commented):
minor comment

mlrun/runtimes/utils.py
alonmr: it is used here https://github.com/mlrun/mlrun/blob/development/pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/ops.py#L664
we can remove it but we'll need to verify pipeline flows

adamdelman (approved):
LGTM 👍

PR #7819, opened 2025-05-11
Revert "[Artifacts] Remove deprecated limit parameter when listing artifacts"

moranbental:
This reverts commit 940cf5ad2f58e4daf5a7aec1fe028a7998a70590  as the `limit` parameter in `list_artifacts` was deprecated in version 1.8.0 and should be removed in 1.11.0, not in 1.10.0.
I also fixed the deprecation version msg from 1.10 to 1.11


mlrun/db/httpdb.py
```diff
@@ -5096,6 +5100,13 @@ def _list_artifacts(
         project = project or config.default_project
         labels = self._parse_labels(labels)
 
+        if limit:
+            # TODO: Remove this in 1.10.0
```
liranbg: change to 1.11

liranbg (changes requested):

Yacouby (approved):

PR #7820, opened 2025-05-11
Revert "[SDK] Remove unused deprecated methods (#7727)"

moranbental:
This reverts commit 7cb8d3bdf07bb183c80fb54df4b1de776a26202b as it was deprecated in version 1.8.0 and should be removed in 1.11.0, not in 1.10.0.
I also fixed the deprecation version msg from 1.10 to 1.11

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -2861,10 +2861,10 @@ class MlrunProject(ModelObj):
 
         self.spec.set_function(name, function_object, func)
 
-    # TODO: Remove this in 1.10.0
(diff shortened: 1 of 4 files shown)


mlrun/projects/project.py
```diff
@@ -2860,6 +2861,20 @@ def _set_function(
 
         self.spec.set_function(name, function_object, func)
 
+    # TODO: Remove this in 1.10.0
+    @deprecated.deprecated(
```
liranbg: change to 1.11

Yacouby (approved):

PR #7818, opened 2025-05-11
[Deprecations] Fix version in deprecations msg [1.8.x]

moranbental:
Version 1.9 is not part of the deprecation flow.
Anything deprecated in 1.7 will be removed in 1.10 (since the version sequence is 1.7 → 1.8 → 1.10).
Anything deprecated in 1.8 will be removed in 1.11 (since the version sequence is 1.8 → 1.10 → 1.11).

https://iguazio.atlassian.net/browse/ML-9365

PR #7834, opened 2025-05-12
[Requirements] Bump mlrun-pipelines-kfp-common [1.8.x]

moranbental:
Bump the version following the changes in this pr - https://github.com/mlrun/mlrun/pull/7818


liranbg (changes requested):
need to run make command to bump locks

PR #7859, opened 2025-05-13
[Datastore] Remove obsolete parameters

alxtkr77:
https://iguazio.atlassian.net/browse/ML-9706

### DIFF
--- a/docs/feature-store/sources-targets.md
+++ b/docs/feature-store/sources-targets.md
@@ -65,7 +65,7 @@ Support for Confluent Kafka is currently in Tech Preview status.
 
 ```python
 profile = DatastoreProfileKafkaSource(
-    name="profile-name", bootstrap_servers="localhost", topic="topic_name"
+    name="profile-name", brokers="localhost", topic="topic_name"
 )
 target = KafkaSource(path="ds://profile-name")
 ```
@@ -203,14 +203,14 @@ NFS, S3, Azure blob storage, Redis, SQL, and on Iguazio DB/FS.
 
 ```python
 profile = DatastoreProfileKafkaTarget(
-    name="profile-name", bootstrap_servers="localhost", topic="topic_name"
+    name="profile-name", brokers="localhost", topic="topic_name"
 )
 target = KafkaTarget(path="ds://profile-name")
 ```
 
 `DatastoreProfileKafkaTarget` class parameters:
 - `name` &mdash; Name of the profile
-- `bootstrap_servers` &mdash; A string representing the 'bootstrap servers' for Kafka. These are the initial contact points you use to discover the full set of servers in the Kafka cluster, typically provided in the format `host1:port1,host2:port2,...`.
+- `brokers` &mdash; A string representing the 'bootstrap servers' for Kafka. These are the initial contact points you use to discover the full set of servers in the Kafka cluster, typically provided in the format `host1:port1,host2:port2,...`.
 - `topic` &mdash; A string that denotes the Kafka topic to which data is sent or from which data is received.
 - `kwargs_public` &mdash; This is a dictionary (`Dict`) meant to hold a collection of key-value pairs that could represent settings or configurations deemed public. These pairs are subsequently passed as parameters to the underlying `kafka.KafkaConsumer()` constructor. The default value for `kwargs_public` is `None`.
 - `kwargs_private` &mdash; This dictionary (`Dict`) is designed to store key-value pairs, typically representing configurations that are of a private or sensitive nature. These pairs are also passed as parameters to the underlying `kafka.KafkaConsumer()` constructor. It defaults to `None`.
--- a/docs/feature-store/sources-targets.md
+++ b/docs/feature-store/sources-targets.md
@@ -294,7 +294,7 @@ For example: `rediss://localhost:6379` creates a Redis target, where:
    - The server location is localhost port 6379.
 - If the path parameter is not set, it tries to fetch it from the MLRUN_REDIS__URL environment variable.
 - You cannot pass the username/password as part of the URL. If you want to provide the username/password, use secrets as:
-`<prefix_>REDIS_USER <prefix_>REDIS_PASSWORD` where \<prefix> is the optional RedisNoSqlTarget `credentials_prefix` parameter.
+`REDIS_USER REDIS_PASSWORD`.
 - Two types of Redis servers are supported: StandAlone and Cluster (no need to specify the server type in the config).
 - A feature set supports one online target only. Therefore `RedisNoSqlTarget` and `NoSqlTarget` cannot be used as two targets of the same feature set.
     
--- a/mlrun/datastore/datastore_profile.py
+++ b/mlrun/datastore/datastore_profile.py
@@ -146,14 +146,6 @@ class DatastoreProfileKafkaTarget(DatastoreProfile):
     kwargs_public: typing.Optional[dict]
     kwargs_private: typing.Optional[dict]
 
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
-
-        if not self.brokers:
-            raise mlrun.errors.MLRunInvalidArgumentError(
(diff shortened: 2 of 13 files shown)


mlrun/datastore/datastore_profile.py
```diff
@@ -142,7 +141,6 @@ def attributes(self):
 class DatastoreProfileKafkaTarget(DatastoreProfile):
     type: str = pydantic.v1.Field("kafka_target")
     _private_attributes = "kwargs_private"
-    bootstrap_servers: typing.Optional[str] = None
     brokers: typing.Optional[str] = None
```
moranbental:
```suggestion
    brokers: str = None
```

mlrun/datastore/datastore_profile.py
```diff
@@ -248,18 +231,7 @@ class DatastoreProfileS3(DatastoreProfile):
...
     access_key_id: typing.Optional[str] = None
     secret_key: typing.Optional[str] = None
-    bucket: typing.Optional[str] = None
-
-    @pydantic.v1.validator("bucket")
-    @classmethod
-    def check_bucket(cls, v):
-        if not v:
-            warnings.warn(
-                "The 'bucket' attribute will be mandatory starting from version 1.9",
```
moranbental: From what I see, this line has existed since version 1.7.0, so we need to update the message to:
`The 'bucket' attribute will be mandatory starting from version 1.10.`
This is because it actually becomes mandatory in 1.10.
We should create PRs for both versions 1.8.0 and 1.9.0 to update the message accordingly.

mlrun/datastore/datastore_profile.py
```diff
@@ -353,18 +325,7 @@ class DatastoreProfileGCS(DatastoreProfile):
...
     credentials_path: typing.Optional[str] = None  # path to file.
     gcp_credentials: typing.Optional[typing.Union[str, dict]] = None
-    bucket: typing.Optional[str] = None
-
-    @pydantic.v1.validator("bucket")
-    @classmethod
-    def check_bucket(cls, v):
-        if not v:
-            warnings.warn(
-                "The 'bucket' attribute will be mandatory starting from version 1.9",
```
moranbental: same here

mlrun/datastore/datastore_profile.py
```diff
@@ -410,18 +371,7 @@ class DatastoreProfileAzureBlob(DatastoreProfile):
...
     sas_token: typing.Optional[str] = None
     credential: typing.Optional[str] = None
-    container: typing.Optional[str] = None
-
-    @pydantic.v1.validator("container")
-    @classmethod
-    def check_container(cls, v):
-        if not v:
-            warnings.warn(
-                "The 'container' attribute will be mandatory starting from version 1.9",
```
moranbental: same here

moranbental (changes requested):

assaf758 (approved):

docs/feature-store/sources-targets.md
```diff
@@ -203,14 +203,14 @@ NFS, S3, Azure blob storage, Redis, SQL, and on Iguazio DB/FS.
 
 ```python
 profile = DatastoreProfileKafkaTarget(
```
moranbental, line 205: This item is missing from the documentation for version 1.8.0, only `bootstrap_servers` in `mlrun.datastore.targets.KafkaTarget` is documented. Please make sure to include this one as well in the ticket https://iguazio.atlassian.net/browse/ML-10042.

mlrun/datastore/targets.py
```diff
@@ -1213,7 +1193,6 @@ def __init__(
         flush_after_seconds: Optional[int] = None,
         storage_options: Optional[dict[str, str]] = None,
         schema: Optional[dict[str, Any]] = None,
-        credentials_prefix=None,
```
moranbental, line 1216: Why did you remove that? There was no deprecation warning for this value, and deleting it will break compatibility

mlrun/datastore/targets.py
```diff
@@ -1488,7 +1466,7 @@ class RedisNoSqlTarget(NoSqlBaseTarget):
     writer_step_name = "RedisNoSqlTarget"
 
     @staticmethod
-    def get_server_endpoint(path, credentials_prefix=None):
+    def get_server_endpoint(path):
```
moranbental, line 1469: Same here, if you want to remove this, you need to add a deprecation warning and remove it in 1.12

mlrun/model.py
```diff
@@ -2155,7 +2155,6 @@ class DataSource(ModelObj):
         "max_age",
         "start_time",
         "end_time",
-        "credentials_prefix",
```
moranbental, line 2158: Same here, a deprecation warning is needed first if users can use this parameter

mlrun/model.py
```diff
@@ -2303,7 +2299,6 @@ class DataTarget(DataTargetBase):
         "key_bucketing_number",
         "partition_cols",
         "time_partitioning_granularity",
-        "credentials_prefix",
```
moranbental, line 2306: same

mlrun/secrets.py
```diff
@@ -150,7 +150,6 @@ def get_secret_or_env(
     key: str,
     secret_provider: Union[dict, SecretsStore, Callable, None] = None,
     default: Optional[str] = None,
-    prefix: Optional[str] = None,
```
moranbental: same

moranbental (changes requested):

mlrun/model.py
alxtkr77: in class BaseStoreTarget: 
`        if credentials_prefix:
            warnings.warn(
                "The 'credentials_prefix' parameter is deprecated in 1.7.0 and will be removed in "
                "1.10.0. Please use datastore profiles instead.",
                FutureWarning,
            )
`
The rest are residuals

docs/feature-store/sources-targets.md
alxtkr77: Its there: 
`credentials_prefix  in mlrun.datastore.targets.BaseStoreTarget resolved in [[Datastore] Remove obsolete parameters by alxtkr77 · Pull Request #7859 · mlrun/mlrun](https://github.com/mlrun/mlrun/pull/7859)`

mlrun/datastore/targets.py
alxtkr77: Its derived from BaseStoreTarget

mlrun/datastore/targets.py
alxtkr77: Its derived from BaseStoreTarget

mlrun/model.py
alxtkr77: Its derived from BaseStoreTarget

mlrun/secrets.py
alxtkr77: Its derived from BaseStoreTarget

docs/feature-store/sources-targets.md
moranbental: but this one is - 
`bootstrap_servers`  in `mlrun.datastore.datastore_profile.DatastoreProfileKafkaTarget`

mlrun/datastore/targets.py
moranbental: But there wasn't a deprecation warning before, so I think it's breaking compatibility

theSaarco: I think this is ok, since it calls `super().__init__` with this parameter, so the user would get the deprecation warning when calling this API with this parameter.

mlrun/secrets.py
theSaarco: Here it's definitely not derived from anything - this `get_secret_or_env` method is a top-level method that is part of the sdk and called directly by the user. If a parameter of this method is deprecated, we need to follow the regular procedure regarding this specific API.

moranbental (approved):
LGTM, Don’t forget to document everything you removed

jillnogold (approved):

PR #7860, opened 2025-05-13
[ModelMonitoring] Remove deprecations

alxtkr77:
https://iguazio.atlassian.net/browse/ML-9706


mlrun/runtimes/nuclio/serving.py
```diff
@@ -207,7 +201,6 @@ def __init__(
         self.graph_initializer = graph_initializer
         self.error_stream = error_stream
         self.track_models = track_models
-        self.tracking_policy = tracking_policy
```
Eyal-Danieli, line 210: also remove the same line from the `GraphServer` in `mlrun/serving/server.py` (L132)

tests/system/model_monitoring/test_app.py
```diff
@@ -241,7 +218,6 @@ def _test_v3io_records(
         last_request: typing.Optional[datetime] = None,
         error_count: typing.Optional[float] = None,
     ) -> None:
-        cls._test_parquet(ep_id, inputs, outputs)
```
Eyal-Danieli, line 244: remove `inputs` and `outputs` params from func signature (and adjust the callers as well)

Eyal-Danieli (commented):
cool, 2 small comments

moranbental (approved):

tests/system/model_monitoring/test_app.py
```diff
@@ -209,39 +209,13 @@ def _test_predictions_table(cls, ep_id: str, should_be_empty: bool = False) -> N
...
     @classmethod
     def _test_v3io_records(
         cls,
         ep_id: str,
-        inputs: set[str],
-        outputs: set[str],
         last_request: typing.Optional[datetime] = None,
         error_count: typing.Optional[float] = None,
     ) -> None:
-        cls._test_parquet(ep_id, inputs, outputs)
```
Eyal-Danieli, line 244: See L838, remove `inputs` and `outputs` from the caller

Eyal-Danieli (commented):
looks good! You just missed removing one line from the test.

Eyal-Danieli (approved):

mlrun/runtimes/nuclio/serving.py
```diff
@@ -132,7 +127,6 @@ def __init__(
         graph_initializer=None,
         error_stream=None,
         track_models=None,
-        tracking_policy=None,
```
moranbental, line 135: Why did you remove this from the class constructor? It wasn't deprecated..

mlrun/serving/server.py
```diff
@@ -108,7 +108,6 @@ def __init__(
         graph_initializer=None,
         error_stream=None,
         track_models=None,
-        tracking_policy=None,
```
moranbental, line 111: same here

moranbental (commented):
post review pr comments

mlrun/serving/server.py
alxtkr77: The `TrackingPolicy` class is deprecated from version 1.7.0 and is not used anymore

moranbental: @alxtkr77 The TrackingPolicy class is deprecated, but this field in GraphServer is not. I'm not sure you can delete it unless it was explicitly deprecated before

mlrun/runtimes/nuclio/serving.py
alxtkr77: While the tracking_policy parameter was never deprecated, the only way users could use it was by creating a TrackingPolicy class instance, which has deprecation messages in its constructor. So any user who actually used the parameter should have received the deprecation message.

mlrun/serving/server.py
alxtkr77: While the tracking_policy parameter was never deprecated, the only way users could use it was by creating a TrackingPolicy class instance, which has deprecation messages in its constructor. So any user who actually used the parameter should have received the deprecation message.

PR #7874, opened 2025-05-14
[Deprecations] Mark rebuild_images query parameter as deprecated [1.9.x]

moranbental:
backport https://github.com/mlrun/mlrun/pull/7868


liranbg (approved):

Yacouby (commented):
What about the suggestion that was in the log that you deleted ? calling disable and enable model monitoring ? it is not relevant anymore ?

moranbental:
> What about the suggestion that was in the log that you deleted ? calling disable and enable model monitoring ? it is not relevant anymore ?

@Yacouby We are on the server side, and the deprecation message is only available in Swagger, where we don’t really know what is enabled or disabled. That’s why I changed the flow to first use a DELETE request, followed by an PUT request.

PR #7901, opened 2025-05-15
[KFP] Fix Pipeline Completed Notification

quaark:
https://iguazio.atlassian.net/browse/ML-9880

It seems that when the notification tries to `kfp_client.wait_for_run_completion` on the completed pipeline, the first request returns still with `Running`, since the timeout and the interval were both 5, it didn't get a second chance to try again.
Increasing the timeout indeed fixed the issue and in the subsequent requests `Succeeded` was returned.
This change is safe since:
1. The notification is sent in a background task
2. The pipeline should be Succeeded already and will return shortly (this may be a kfp2 hiccough)

----
Additional fix: chmod the schemas after compilation so `patch_env` may build a working api image.


liranbg (approved):

server/common/Makefile
```diff
@@ -69,6 +69,7 @@ compile-schemas-python:
 	@echo Fixing Python GRPC Dir and Imports
 	@sh hack/scripts/fix_python_grpc_imports.sh
 	@cp -RL proto/build/py/* ../py/schemas/proto
+	@chmod -R 755 ../py/schemas/proto
```
adamdelman, line 72: Maybe this should be in a separate PR?

adamdelman (changes requested):
LGTM, minor suggestion

server/common/Makefile
quaark: @liranbg asked me to add this btw :)

liranbg (approved):

PR #7906, opened 2025-05-15
[Tutorials] fix GenAI and MLRun tutorials

GiladShapira94:
Fix - https://iguazio.atlassian.net/browse/ML-9990
This demo includes - 
1. Align project names across notebooks
2. Fix the GenAi monitoring application
3. Remove the need and the example of non local Milvus installation


docs/tutorials/src/data-prep.py
```diff
@@ -1,10 +1,13 @@
+
 import pandas as pd
 from sklearn.datasets import load_breast_cancer
 
+import mlrun
```
gilad-shaham: Does it require this import?

docs/tutorials/genai-03-vector-db.ipynb
```diff
@@ -421,20 +396,13 @@
...
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": []
   }
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "mlrun-base",
+   "display_name": "tutorial-llm",
```
gilad-shaham: We should leave the kernel name as-is

docs/tutorials/genai_01_basic_tutorial.ipynb
```diff
@@ -136,55 +128,33 @@
     "    name=\"fetch-vectordb-data\",\n",
     "    func=\"src/fetch-vectordb-data.py\",\n",
     "    kind=\"job\",\n",
-    "    # image=\"gcr.io/iguazio/mlrun-genai/mlrun-llm-demo-data:1.7.0\",\n",
     "    image=\"edmondg/mlrun-llm-demo-data-new:1.8.0rc30\",\n",
```
gilad-shaham: There are conflicts, I think around this, there was a fix where the image names were changed which is not in this PR. Also image name with 1.7.0 appears somewhere here and I remember this was fixed

gilad-shaham (changes requested):

docs/tutorials/src/data-prep.py
GiladShapira94: removed

docs/tutorials/genai-03-vector-db.ipynb
GiladShapira94: fixed

PR #7917, opened 2025-05-17
[Docker] Compile Python bytecode in `uv pip install` / `sync`

jond01:
Back in #6677, I replaced `pip install` with `uv pip install`/`uv pip sync`. We had some unexplained 68% reduction in the `mlrun-api` image size.
After talking with uv authors at PyCon US 2025, I understood this "improvement" in the image size is due to missing bytecode compiled files (`.pyc` in `__pycache__`).

The change in this PR is setting the `UV_COMPILE_BYTECODE=1` env var, as recommended in Docker files:
https://docs.astral.sh/uv/guides/integration/docker/#compiling-bytecode

The previous behavior compiled the `.py` files on the go when they were first called.

See also:

* https://docs.astral.sh/uv/pip/compatibility/#bytecode-compilation
* https://docs.astral.sh/uv/configuration/environment/#uv_compile_bytecode


liranbg (approved):
💯

PR #7938, opened 2025-05-20
[Application] Change default image to be resolve by client python version

alonmr:
https://iguazio.atlassian.net/browse/ML-10060


liranbg (approved):

mlrun/config.py
```diff
@@ -286,7 +286,7 @@
             "remote": "mlrun/mlrun",
             "dask": "mlrun/ml-base",
             "mpijob": "mlrun/mlrun",
-            "application": "python:3.9",
+            "application": "python:3.11",
```
liranbg: need to condition base on client version

liranbg (changes requested):

liranbg (approved):

PR #7940, opened 2025-05-20
[Deprecations] Remove `uid` param in GET and DELETE artifact

moranbental:
deprecated in 1.7.0 removed in 1.10 
1. DELETE `/projects/{project}/artifacts/{key:path}?uid`  - Use `object-uid` instead
2. GET `/projects/{project}/artifacts/{key:path}?uid` - Use `object-uid` instead
https://iguazio.atlassian.net/browse/ML-10046


moranbental:
It fails in BC because we're deleting the UID in the query

liranbg (approved):

PR #7946, opened 2025-05-21
[Artifacts] LLM prompt artifact DB&APIs

davesh0812:
1. **Parent Artifact Support** – Introduced the ability to associate a child artifact with its parent artifact via `parent_uri`, including filtering capabilities through the API for querying artifacts by their parent.

2. **LLMPromptArtifact Class** – Added a new artifact type to represent LLM prompts, encapsulating prompt content, metadata, and generation configuration for structured prompt management.

3. **SDK Enhancements** – Extended the MLRun SDK to support logging and listing LLM prompt artifacts, by using exist APIs.

https://iguazio.atlassian.net/browse/ML-9966


mlrun/artifacts/llm_prompt.py
```diff
@@ -0,0 +1,130 @@
...
+        if prompt_string and len(prompt_string) > MAX_PROMPT_LENGTH:
+            logger.info("prompt_string is too long, creating a temp file")
+            temp_file = tempfile.NamedTemporaryFile(
+                delete=False, mode="w", suffix=".txt"
+            )
+            temp_file.write(prompt_string)
+            temp_file.flush()
+            temp_file.close()
+            prompt_path = temp_file.name
+            prompt_string = None
```
liranbg: you may use `with ...` to flush and close it for you

mlrun/utils/helpers.py
```diff
@@ -131,7 +134,7 @@ def get_artifact_target(item: dict, project=None):
     kind = item.get("kind")
     uid = item["metadata"].get("uid")
 
-    if kind in {"dataset", "model", "artifact"} and db_key:
+    if kind in {"dataset", "model", "artifact", "llm-prompt"} and db_key:
```
liranbg: change to constants please

server/py/framework/db/sqldb/db.py
```diff
@@ -1781,6 +1840,19 @@ def _find_artifacts(
                 partition_order,
                 with_tagged=True,
             )
+        if parent_key or parent_tag:
+            ref_alias = aliased(ArtifactV2)
+
+            # Join on reference_artifact_id -> ArtifactV2.id
+            query = query.outerjoin(ref_alias, ArtifactV2.parent_id == ref_alias.id)
```
liranbg: I suggest to revisit the outerjoin here as it will multiple the artifacts table twice. perhaps you can use specific columns on retrieval so outerjoin wont create 2 full artifact tables

tests/system/api/test_artifacts.py
```diff
@@ -26,7 +26,7 @@
 
 @TestMLRunSystem.skip_test_if_env_not_configured
 class TestAPIArtifacts(TestMLRunSystem):
-    project_name = "db-system-test-project"
+    project_name = "db-system-test-project-3"
```
liranbg:
```suggestion
    project_name = "test-project-artifacts"
```
perhaps better?

tests/system/api/test_artifacts.py
```diff
@@ -90,3 +90,37 @@ def test_verify_artifact_tag_in_output(self):
...
+    def test_llm_prompt_artifact_with_model(self):
+        # Log a model artifact
+        model_name = "model"
+        model_file = "./assets/model.pkl"
+        model = self.project.log_model(
+            model_name,
+            model_file=model_file,
+            upload=True,
+        )
+        # Create a project and log an LLM prompt artifact
```
liranbg: you dont create a project

liranbg (changes requested):
premature review with some suggestions and corrections
main concern is db query against big artifacts table where outjoin will duplicate the table to scan for parents and etc. I suggest to build the query query generated by sqlalchemy and run it manually against big artifacts table and see what can be optimized and apply it in code.

server/py/framework/db/sqldb/db.py
davesh0812: Good catch... for sure it suppose to be inner join

mlrun/artifacts/llm_prompt.py
```diff
@@ -0,0 +1,133 @@
...
+    def __init__(
+        self,
+        target_path=None,
+        viewer=None,
+        is_inline=False,
+        format=None,
+        size=None,
+        db_key=None,
+        extra_data=None,
+        body=None,
```
theSaarco: What is the meaning of getting a `body` here? We are putting the prompt string in the body, so what will this mean?

mlrun/artifacts/llm_prompt.py
```diff
@@ -0,0 +1,136 @@
...
+        parent_uri: typing.Optional[str] = None,
+        prompt_string: typing.Optional[str] = None,
+        prompt_path: typing.Optional[str] = None,
+        prompt_legend: typing.Optional[dict] = None,
+        generation_configuration: typing.Optional[dict] = None,
+        description: typing.Optional[str] = None,
+    ):
+        if prompt_string and prompt_path:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "cannot specify prompt_string and prompt_path together"
```
theSaarco:
```suggestion
                "cannot specify both prompt_string and prompt_path"
```

mlrun/artifacts/llm_prompt.py
```diff
@@ -0,0 +1,136 @@
...
+# limitations under the License.
+import tempfile
+import typing
+
+import mlrun
+import mlrun.artifacts.model as model_artifact
+from mlrun.artifacts import Artifact, ArtifactSpec
+from mlrun.utils import StorePrefix, logger
+
+MAX_PROMPT_LENGTH = 1000
```
theSaarco: I prefer 1024...

mlrun/artifacts/llm_prompt.py
```diff
@@ -0,0 +1,136 @@
...
+        super().__init__(
+            src_path=prompt_path,
+            target_path=target_path,
+            viewer=viewer,
+            is_inline=is_inline,
+            format=format,
+            size=size,
+            db_key=db_key,
+            extra_data=extra_data,
+            body=body,
```
theSaarco: If we're not getting a body, you can put here the `prompt_string`, and then you won't need to change the `get_body` implementation (or add the `self.prompt_string` parameter).

mlrun/artifacts/llm_prompt.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class LLMPromptArtifact(Artifact):
+    """
+    LLM Prompt Artifact
+
+    This artifact is used to store and manage LLM prompts.
+    Stores the prompt string/path and a link to the related model artifact.
+    """
+
```
theSaarco, line 78: Why not generate an `__init__` implementation with the relevant parameters?

mlrun/execution.py
```diff
@@ -893,6 +895,94 @@ def log_model(
...
+            else model_artifact,
+            generation_configuration=generation_configuration,
+            target_path=target_path,
+            description=description,
+        )
+        llm_prompt = LLMPromptArtifact(
+            key=key,
+            spec=llm_prompt_spec,
+            **kwargs,
+        )
```
theSaarco: If you implement `LLMPromptArtifact.__init__` then you don't need to do this step of generating the spec first, but rather just call that method.

mlrun/execution.py
```diff
@@ -893,6 +895,94 @@ def log_model(
...
+            project.log_llm_prompt(
+                key="summarization-prompt",
+                prompt_path="prompts/summarize.txt",
+                labels=["task=summarization"],
+            )
+
+        :param key: Unique name of the artifact.
+        :param prompt_string: Raw prompt text as a string. Cannot be used with `prompt_path`.
+        :param prompt_path: Path to a file containing the prompt content. Cannot be used with `prompt_string`.
+        :param prompt_legend: A dictionary that maps variables used in the prompt to their expected context keys.
```
theSaarco: hmm... I was actually thinking that this will indicate per template parameter what the type of the object is. I'm not sure this is really meaningful - mapping to context keys. Let's discuss this.

mlrun/artifacts/llm_prompt.py
```diff
@@ -0,0 +1,136 @@
...
+        size=None,
+        db_key=None,
+        extra_data=None,
+        body=None,
+        unpackaging_instructions: typing.Optional[dict] = None,
+        parent_uri: typing.Optional[str] = None,
+        prompt_string: typing.Optional[str] = None,
+        prompt_path: typing.Optional[str] = None,
+        prompt_legend: typing.Optional[dict] = None,
+        generation_configuration: typing.Optional[dict] = None,
```
theSaarco: I don't like the name "generation configuration" - this is not the config of the generation of the prompt, but rather this is expected to be the model configuration to be used. So why not call this `model_configuration`?

mlrun/artifacts/base.py
```diff
@@ -106,6 +106,8 @@ class ArtifactSpec(ModelObj):
         "extra_data",
         "unpackaging_instructions",
         "producer",
+        "parent_uri",
+        "have_children",
```
theSaarco:
```suggestion
        "has_children",
```

mlrun/artifacts/base.py
```diff
@@ -138,6 +141,8 @@ def __init__(
         self.db_key = db_key
         self.extra_data = extra_data or {}
         self.unpackaging_instructions = unpackaging_instructions
+        self.parent_uri = parent_uri
+        self.have_children = None
```
theSaarco: What does `None` mean in this context? Is it the same as `False`? Then why not use `False`? Adding the 3rd option of using `None` to a boolean is logically confusing.

mlrun/db/httpdb.py
```diff
@@ -1253,6 +1254,8 @@ def list_artifacts(
                 "results", tag="*", project="iris", labels=["uploaded", "type=binary"]
             )
 
+        :param parent: The parent artifact uri. This is used to filter artifacts that
+                        are children of a specific artifact.
```
theSaarco: Is it just the URI? I see in the Crud you allow a store URI or just a string which is the `key:tag`. This should be noted.

theSaarco (changes requested):
Overall looks very good, some changes are needed.

mlrun/execution.py
davesh0812: Maybe this better : 
```
        :param prompt_legend: A dictionary where each key is a placeholder in the prompt (e.g., {user_name}) 
                              and the value is a description or explanation of what that placeholder represents.
                              Useful for documenting and clarifying dynamic parts of the prompt.
```

mlrun/artifacts/llm_prompt.py
```diff
@@ -0,0 +1,156 @@
...
+        target_path: Optional[str] = None,
+        **kwargs,
+    ):
+        if prompt_string and prompt_path:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "cannot specify both prompt_string and prompt_path"
+            )
+
+        if prompt_string and len(prompt_string) > MAX_PROMPT_LENGTH:
+            logger.info("prompt_string is too long, creating a temp file")
```
liranbg: is it relevant for client? perhaps you can just debug it and not info

mlrun/artifacts/llm_prompt.py
```diff
@@ -0,0 +1,156 @@
...
+            with tempfile.NamedTemporaryFile(
+                delete=False, mode="w", suffix=".txt"
```
liranbg: when do you actually delete the file? comment it

mlrun/execution.py
```diff
@@ -893,6 +894,92 @@ def log_model(
...
+        as a string or by referencing a file path. Optionally, you can link the prompt to a parent model artifact and
+        provide metadata like a prompt legend (e.g., input variable mapping) and generation configuration.
+
+        Examples::
+
+            # Log an inline prompt
+            context.log_llm_prompt(
+                key="qa-prompt",
+                prompt_string="Q: {question}\nA:",
+                model_artifact="model:latest",
```
liranbg: I suggest encouraging clients using artifact uri (in here you can provide a reference from code to use "get_model" or so with key "model" and tag="latest") wdyt?

tests/system/api/assets/model.pkl
liranbg, line 1: needed?

mlrun/common/schemas/artifact.py
```diff
@@ -26,6 +26,7 @@ class ArtifactCategories(mlrun.common.types.StrEnum):
     model = "model"
     dataset = "dataset"
     document = "document"
+    llm_prompt = "llm-prompt"
```
liranbg, line 29: @theSaarco  should `llm_prompt` be considered as `other`? if no then need to add this kind to `from_kind` function so project counters wont consider it as "other"

liranbg (changes requested):
few last tweaks

mlrun/common/schemas/artifact.py
theSaarco: It shouldn't be considered `other` - it has its own category. Indeed, `from_kind` should be modified.

mlrun/execution.py
```diff
@@ -893,6 +894,92 @@ def log_model(
...
+        :param labels: Labels to tag the artifact (e.g., list or dict of key-value pairs).
+        :param upload: Whether to upload the artifact to the store (defaults to True).
+        :param kwargs: Additional fields to pass to the `LLMPromptArtifact` constructor.
+
+        :returns: The logged `LLMPromptArtifact` object.
+        """
+
+        llm_prompt = LLMPromptArtifact(
+            key=key,
+            project=self._project or "",
```
theSaarco: Is this line needed? The project is deduced anyway when calling `log_artifact`.

mlrun/projects/project.py
```diff
@@ -1881,6 +1882,93 @@ def log_model(
...
+            LLMPromptArtifact,
+            self.log_artifact(
+                llm_prompt,
+                artifact_path=artifact_path,
+                tag=tag,
+                upload=upload,
+                labels=labels,
+            ),
+        )
+        self._update_run()
```
theSaarco: Remove this line, it's a copy-paste error from the context flow.

mlrun/projects/project.py
```diff
@@ -4595,6 +4683,130 @@ def paginated_list_models(
...
+        :param model: Return prompts associated with this model (can be `Artifact` URI or `Artifact` object).
+        :param format_: Output format: one of `full`, `metadata`, or `name_only`.
+
+        :returns: A list of filtered `LLMPromptArtifact` objects matching the given parameters.
+        """
+        db = mlrun.db.get_run_db(secrets=self._secrets)
+        return db.list_artifacts(
+            name,
+            self.metadata.name,
+            tag,
```
theSaarco: Would be easier for the reader to make these named parameters.

server/py/framework/db/sqldb/db.py
```diff
@@ -818,6 +819,27 @@ def create_artifact(
...
+            _, uri = mlrun.datastore.parse_store_uri(parent_uri)
+            project, key, iteration, tag, tree, uid = parse_artifact_uri(uri)
+            parent_db_artifact = self.read_artifact(
+                session=session,
+                key=key,
+                tag=tag,
+                iter=iteration,
+                producer_id=tree,
+                uid=uid,
+                project=project,
```
theSaarco, line 833: hmm... This made me think - we don't block cross-project reference here, and this is potentially a security breach where a prompt artifact actually points at a model from another project. This retrieval here is done without any checking that the user is allowed to access the project pointed at by the uri.
For now I think you need to block cross-project references (in the sdk and in the server side), we need to consider what it means to allow cross-project references, and whether it makes sense.

theSaarco (changes requested):
There are some changes needed.
The big one is that we don't verify that model referenced is actually from a project that is reachable by the user. In general, users can point at models belonging to any project whatsoever. For now this needs to be blocked (project needs to be the same).

mlrun/execution.py
davesh0812: Yep I need to provide some project name when creating `ArtifactMetadata`

mlrun/projects/project.py
davesh0812: No problem

mlrun/artifacts/llm_prompt.py
davesh0812: I added a deletion when logging the artifact

tests/system/api/assets/model.pkl
davesh0812: Yep for the sys test, I need to log model

mlrun/execution.py
davesh0812: `get_model` also need `model_uri` so I just wll cahnge it to :
`model_artifact=model,  # model is `ModelArtifact``

mlrun/artifacts/llm_prompt.py
```diff
@@ -0,0 +1,164 @@
...
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "cannot specify both prompt_string and prompt_path"
+            )
+
+        super().__init__(
+            src_path=prompt_path,
+            target_path=target_path,
+            parent_uri=model_artifact.uri
+            if isinstance(model_artifact, model_art.ModelArtifact)
+            else model_artifact,
```
theSaarco, line 55: If it's a string then you will need to verify here that it's a valid artifact uri - currently other places in your code assumes the uri is a store uri, so need to verify that users understand that this is the expected format, or also enable other identifications. For example, I can pass key:latest or key^some-uuid as a valid artifact uri, but this is not a store uri and also it doesn't contain a project in it.
To clarify - I think such a uri should be considered valid.

mlrun/common/helpers.py
```diff
@@ -44,3 +45,28 @@ def generate_api_gateway_name(project: str, name: str) -> str:
...
+    expected_parent_uri: str,
+) -> None:
+    """
+    Check if the artifact's parent URI is valid and under the same project.
+    :param artifact_project:     Artifact project name
+    :param expected_parent_uri:  Expected parent URI of the artifact
+    :raise: MLRunInvalidArgumentError if the parent URI is invalid or not under the same project
+    """
+    # check if the parent_uri is a valid artifact uri and it is under the same project
+    if mlrun.datastore.is_store_uri(expected_parent_uri):
```
theSaarco: This is where you assume the uri is a store uri. This is currently not said anywhere (and as mentioned - I think it should be expanded a bit).

server/py/services/api/api/endpoints/artifacts_v2.py
```diff
@@ -62,6 +63,13 @@ async def create_artifact(
             auth_info,
         )
     )
+
+    if artifact.spec.parent_uri:
+        run_in_threadpool(
```
theSaarco: Does this need to run in threadpool? This check just does string matching.

server/py/services/api/crud/artifacts.py
```diff
@@ -168,6 +170,14 @@ def list_artifacts(
     ) -> list:
         if labels is None:
             labels = []
+        if mlrun.datastore.is_store_uri(parent):
+            _, uri = mlrun.datastore.parse_store_uri(parent)
+            _, parent_key, _, parent_tag, _, _ = parse_artifact_uri(uri)
```
theSaarco: This is surprising logic to me - if the user provided a full URI, then I'd expect you to just return prompts for that specific artifact instance. The way you parse it here, the user may be surprised because if providing a URI that has a key and a uid (for example), you will only filter on the key since there is no tag in the URI. This will return much more results than what the user expected.
My expectation is that if a URI was used, then get the specific instance the URI points at. If a string was provided then get instance based on key and tag.

theSaarco (changes requested):
So now I think there are gaps in the way artifact URIs are handled, are those store URIs or just artifact URIs? Also, the list-artifacts filter per parent model also has issues in parsing the parent URI.

mlrun/artifacts/llm_prompt.py
davesh0812: I will add TODO

theSaarco (approved):
LGTM 👍

mlrun/artifacts/base.py
```diff
@@ -451,6 +459,10 @@ def _upload_file(
             url=target_path or self.spec.target_path
         ).upload(source_path)
 
+        if self._src_is_temp and os.path.exists(self.spec.src_path):
+            # delete the temporary file if it was created for the upload
+            os.remove(self.spec.src_path)
```
moranbental, line 464: What if it fails to remove the file? Add a try except block and maybe log the problem

mlrun/artifacts/llm_prompt.py
```diff
@@ -0,0 +1,166 @@
...
+        prompt_path: Optional[str] = None,
+        prompt_legend: Optional[dict] = None,
+        model_configuration: Optional[dict] = None,
+        description: Optional[str] = None,
+        target_path: Optional[str] = None,
+        **kwargs,
+    ):
+        if prompt_string and prompt_path:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "cannot specify both prompt_string and prompt_path"
```
moranbental:
```suggestion
                "Cannot specify both 'prompt_string' and 'prompt_path'"
```

mlrun/artifacts/llm_prompt.py
```diff
@@ -0,0 +1,166 @@
...
+            model_art.ModelArtifact, str
+        ] = None,  # TODO support partial model uri
+        prompt_string: Optional[str] = None,
+        prompt_path: Optional[str] = None,
+        prompt_legend: Optional[dict] = None,
+        model_configuration: Optional[dict] = None,
+        description: Optional[str] = None,
+        target_path=None,
+        **kwargs,
+    ):
```
moranbental, line 96: In 1.10, we deprecated the key, project, and target_path parameters in the Artifact constructor, so I'm not sure you can add them here.

mlrun/artifacts/llm_prompt.py
```diff
@@ -0,0 +1,166 @@
...
+            return self.spec._model_artifact
+        return None
+
+    def read_prompt(self) -> str:
+        """
+        Read the prompt string from the artifact.
+        """
+        if self.spec.prompt_string:
+            return self.spec.prompt_string
+        if self.spec.target_path:
```
moranbental, line 142: What if both prompt_string and target_path are None? You're not returning anything in that case

mlrun/artifacts/llm_prompt.py
```diff
@@ -0,0 +1,166 @@
...
+            with mlrun.datastore.store_manager.object(url=self.spec.target_path).open(
+                mode="r"
+            ) as p_file:
+                return p_file.read()
+
+    def before_log(self):
```
moranbental, line 148: Who is calling this function?

mlrun/execution.py
```diff
@@ -893,6 +894,93 @@ def log_model(
...
+                key="qa-prompt",
+                prompt_string="Q: {question}\nA:",
+                model_artifact=model,  # model is `ModelArtifact`
+                prompt_legend={"question": "user_input"},
+                model_configuration={"temperature": 0.7, "max_tokens": 128},
+                tag="latest",
+            )
+
+            # Log from a file
+            project.log_llm_prompt(
```
moranbental:
```suggestion
            context.log_llm_prompt(
```

mlrun/execution.py
```diff
@@ -893,6 +894,93 @@ def log_model(
...
+        :param labels: Labels to tag the artifact (e.g., list or dict of key-value pairs).
+        :param upload: Whether to upload the artifact to the store (defaults to True).
+        :param kwargs: Additional fields to pass to the `LLMPromptArtifact` constructor.
+
+        :returns: The logged `LLMPromptArtifact` object.
+        """
+
+        llm_prompt = LLMPromptArtifact(
+            key=key,
+            project=self._project or "",
```
moranbental: use the property
```suggestion
            project=self.project or "",
```

mlrun/artifacts/llm_prompt.py
```diff
@@ -0,0 +1,166 @@
...
+
+class LLMPromptArtifact(Artifact):
+    """
+    LLM Prompt Artifact
+
+    This artifact is used to store and manage LLM prompts.
+    Stores the prompt string/path and a link to the related model artifact.
+    """
+
+    kind = "llm-prompt"
```
moranbental: llm_prompt
```suggestion
    kind = mlrun.common.schemas.ArtifactCategories.llm_prompt
```

mlrun/artifacts/manager.py
```diff
@@ -262,6 +265,10 @@ def log_artifact(
         item.iter = producer.iteration
         project = project or producer.project
         item.project = project
+        if item.spec.parent_uri:
+            mlrun.artifacts.helpers.check_artifact_parent(
+                artifact_project=item.project, expected_parent_uri=item.spec.parent_uri
```
moranbental, line 270: if project and producer.project are None?

mlrun/projects/project.py
```diff
@@ -4595,6 +4686,130 @@ def paginated_list_models(
...
+
+            # Get prompts using a name wildcard
+            prompts = project.list_llm_prompts(name="~chat")
+
+        :param name: Name of the prompt artifact. Prefix with '~' for wildcard search (case-insensitive).
+        :param tag: Filter artifacts by this tag (e.g., 'latest', 'prod').
+        :param labels: Filter by label key-value pairs or key presence. Can be:
+                       - Dict: {"key": "value"}, {"key": None}
+                       - List: ["key=value", "key"]
+                       - Comma-separated string: "key=value,key2"
```
```suggestion
        :param labels: Filter LLM prompts artifacts by label key-value pairs or key existence. This can be provided as:

                       - A dictionary in the format `{"label": "value"}` to match specific label key-value pairs,
                         or `{"label": None}` to check for key existence.
                       - A list of strings formatted as `"label=value"` to match specific label key-value pairs,
                         or just `"label"` for key existence.
                       - A comma-separated string formatted as `"label1=value1,label2"` to match entities with
                         the specified key-value pairs or key existence.
```

mlrun/projects/project.py
```diff
@@ -4595,6 +4686,130 @@ def paginated_list_models(
...
+        :param since: (Unused)
+        :param until: (Unused)
+        :param iter: Retrieve a specific iteration. Use `0` for root; `None` for all.
+        :param best_iteration: Return artifacts from the best iteration (used with hyperparameter runs).
+        :param tree: Filter by artifact tree ID (e.g., for lineage filtering).
+        :param limit: (Deprecated) Max number of results to return.
```
moranbental: Why are you adding limit if it is deprecated?

mlrun/projects/project.py
```diff
@@ -4595,6 +4686,130 @@ def paginated_list_models(
...
+            prompts = project.list_llm_prompts(name="~chat")
+
+        :param name: Name of the prompt artifact. Prefix with '~' for wildcard search (case-insensitive).
+        :param tag: Filter artifacts by this tag (e.g., 'latest', 'prod').
+        :param labels: Filter by label key-value pairs or key presence. Can be:
+                       - Dict: {"key": "value"}, {"key": None}
+                       - List: ["key=value", "key"]
+                       - Comma-separated string: "key=value,key2"
+        :param since: (Unused)
+        :param until: (Unused)
```
moranbental: Why is it unused? It's actually used in httpdb.list_artifacts

mlrun/projects/project.py
```diff
@@ -4595,6 +4686,130 @@ def paginated_list_models(
...
+        :param iter: Retrieve a specific iteration. Use `0` for root; `None` for all.
+        :param best_iteration: Return artifacts from the best iteration (used with hyperparameter runs).
```
```suggestion
        :param best_iteration: Returns the artifact which belongs to the best iteration of a given run, in the case of
            artifacts generated from a hyper-param run. If only a single iteration exists, will return the artifact
            from that iteration. If using ``best_iter``, the ``iter`` parameter must not be used.
```

mlrun/projects/project.py
```diff
@@ -4595,6 +4686,130 @@ def paginated_list_models(
...
+        :param tree: Filter by artifact tree ID (e.g., for lineage filtering).
+        :param limit: (Deprecated) Max number of results to return.
+        :param model: Return prompts associated with this model (can be `Artifact` URI or `Artifact` object).
+        :param format_: Output format: one of `full`, `metadata`, or `name_only`.
```
```suggestion
        :param format_: The format in which to return the artifacts. Default is 'full'.
```

mlrun/projects/project.py
```diff
@@ -4595,6 +4686,130 @@ def paginated_list_models(
...
+        since=None,
+        until=None,
+        iter: Optional[int] = None,
+        best_iteration: bool = False,
+        tree: Optional[str] = None,
+        limit: Optional[int] = None,
+        model: Optional[Union[str, Artifact]] = None,
+        format_: Optional[
+            mlrun.common.formatters.ArtifactFormat
+        ] = mlrun.common.formatters.ArtifactFormat.full,
```
moranbental, line 4693: what about all the partition_by fields?

server/py/framework/db/sqldb/db.py
```diff
@@ -1781,6 +1836,49 @@ def _find_artifacts(
                 partition_order,
                 with_tagged=True,
             )
+        if parent_uri:
```
moranbental, line 1839: Why not move all the relevant parent related code to a separate function?

server/py/services/api/api/endpoints/artifacts_v2.py
```diff
@@ -161,6 +173,7 @@ async def list_artifacts(
     tree: Optional[str] = None,
     producer_uri: Optional[str] = None,
     best_iteration: bool = Query(False, alias="best-iteration"),
+    parent: Optional[str] = Query(None, alias="parent"),
```
moranbental: Why do you need the alias?

tests/system/api/test_artifacts.py
```diff
@@ -90,3 +90,32 @@ def test_verify_artifact_tag_in_output(self):
         assert "v2" in outputs_uri, "Expected 'v2' tag in outputs_uri"
 
         mlrun.get_dataitem(output_uri)
+
+    def test_llm_prompt_artifact(self):
```
moranbental, line 94: Not sure why it should be a system test

moranbental (changes requested):

mlrun/artifacts/llm_prompt.py
davesh0812: Why not? I am not pass it to the `super.__init__()`

mlrun/artifacts/llm_prompt.py
davesh0812: So it is a None..

mlrun/artifacts/llm_prompt.py
davesh0812: The ArtifactManger before it logs an artifact..

mlrun/artifacts/manager.py
davesh0812: This is the last point you adding the project to artifact no?
So if now the project are not align it will fail as expected

tests/system/api/test_artifacts.py
davesh0812: For now I will it as sys test we can change it later

server/py/framework/db/sqldb/db.py
davesh0812: I can do it in follow up PR

mlrun/projects/project.py
davesh0812: I use the list_model as a reference... but no problem will add it

mlrun/artifacts/llm_prompt.py
moranbental: so return None explicitly and change the method signature to be optional str

mlrun/artifacts/llm_prompt.py
moranbental: call `super().before_log()` in this function

mlrun/artifacts/manager.py
moranbental: add unit test to verify that

tests/system/api/test_artifacts.py
moranbental: If you don't change this now, no one will do it later...

mlrun/artifacts/base.py
davesh0812: Will add it in follow up PR

mlrun/artifacts/manager.py
davesh0812: `test_unauthorised_model` this one is not good?
You want the special case when `project==None`?

mlrun/projects/project.py
```diff
@@ -4595,6 +4677,156 @@ def paginated_list_models(
...
+               - A dictionary in the format `{"label": "value"}` to match specific label key-value pairs,
+                 or `{"label": None}` to check for key existence.
+               - A list of strings formatted as `"label=value"` to match specific label key-value pairs,
+                 or just `"label"` for key existence.
+               - A comma-separated string formatted as `"label1=value1,label2"` to match entities with
+                 the specified key-value pairs or key existence.
+        :param since: Return artifacts updated after this date (as datetime object).
+        :param until: Return artifacts updated before this date (as datetime object).
+        :param iter: Retrieve a specific iteration. Use `0` for root; `None` for all.
+        Suggested change
```
moranbental: Suggested change ?

mlrun/projects/project.py
```diff
@@ -4595,6 +4677,156 @@ def paginated_list_models(
...
+        :param best_iteration: Returns the artifact which belongs to the best iteration of a given run, in the case of
+            artifacts generated from a hyper-param run. If only a single iteration exists, will return the artifact
+            from that iteration. If using ``best_iter``, the ``iter`` parameter must not be used.
+        :param tree: Filter by artifact tree ID (e.g., for lineage filtering).
+        :param model: Return prompts associated with this model (can be `Artifact` URI or `Artifact` object).
+        Suggested change
```
moranbental: same

mlrun/artifacts/manager.py
moranbental: when producer.project and project are none

moranbental (changes requested):

mlrun/artifacts/manager.py
davesh0812: Is it a valid option? The artifact will store with no project specify?

moranbental (approved):
LGTM, but all the changes should be made in the next PR

PR #7959, opened 2025-05-21
[Model Monitoring] Add `function-summaries` API

Eyal-Danieli:
Add a new API named `function-summaries` (`GET /api/v1/projects/{project-name}/model-monitoring/function-summaries`). The new API returns details and statistics (including the number of detections and possible detections) for model monitoring functions (both application-level and infrastructure) within a project.. 
As part of implementing this new API, the following changes were made:
- Adding label `infra` to operational mm functions (e.g. model-monitoring-writer) and label `application` to application mm functions. 
- Converting the `endpoint_name` into a label in V3IO TSDB. This label is effectively correlated with `endpoint_id`, which helps reduce object count and improves query performance.
- Introducing a new related SDK method under the Project API `project.get_monitoring_function_summaries()`

--- 
**Function Summaries API**

**Path** : `GET /api/v1/projects/{project}/model-monitoring/function-summaries`:
**Parameters**:

- `start`: The query start time to get stats. The string format will be similar to the metric-values API, meaning Unix epoch time format (e.g. `1746693591046`).
- `end`: the query end time to get stats.
- `names`: Function names to filter by. 
- `labels`: Labels to filter by. 
- `include_stats`: If true, include statistics (and #meps if supported) over the requested period. By default, this value is True.
- `include_infra`: if true, include mm infra functions. By default, this value is True. 

**Response**: A list of all the model monitoring functions with their summaries. Note that the response includes
        both monitoring application real time functions and monitoring infrastructure functions.

- `name`: The function name.
- `application_class`: The application class.
- `updated_time`: The function last modification time.
- `base_period`: The base period of the function. 
- `status`: The function status, e.g. ready, failed, etc. 
- `stats`: statistics over the requested period, including:
    - `detections` - total number of detected drift results.
    - potential detections - total number of possible detected drift results.
    - `lag` (future) - total number of unprocessed messages. 
    - `committed_offset` (future) - the latest offset to be processed (agg by shards).

https://iguazio.atlassian.net/browse/ML-9947


server/py/services/api/api/endpoints/model_monitoring.py
```diff
@@ -327,3 +331,101 @@ def set_model_monitoring_credentials(
...
+            raise mlrun.errors.MLRunInvalidArgumentTypeError(
+                "Custom start and end times must contain the timezone."
+            )
+        if start > end:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "The start time must precede the end time."
+            )
+    else:
+        raise mlrun.errors.MLRunInvalidArgumentError(
+            "Provided only one of start time, end time. Please provide both or neither."
```
gtopper, line 390: Why though? And why is this check in a function called `_verify_authorization`?

server/py/services/api/api/endpoints/model_monitoring.py
```diff
@@ -327,3 +331,101 @@ def set_model_monitoring_credentials(
...
+async def get_model_monitoring_function_summaries(
+    commons: Annotated[_FunctionSummariesParams, Depends(_common_function_parameters)],
+    names: Optional[list[str]] = Query(None, alias="name"),
+    labels: list[str] = Query([], alias="label"),
+    include_stats: bool = Query(True, alias="include_stats"),
+    include_infra: bool = Query(True, alias="include_infra"),
+) -> list[mlrun.common.schemas.model_monitoring.FunctionSummary]:
+    """Get monitoring function summaries for the specified project.
+
+    :param commons:                     The common parameters of the request.
```
gtopper:
```suggestion
    :param commons: The common parameters of the request.
```

mlrun/common/schemas/model_monitoring/functions.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        Create a FunctionSummary instance from a function object.
+        """
+
+        return cls(
+            type=func_type,
+            name=func.metadata.name,
+            application_class=func.spec.graph.steps.get(
+                "PushToMonitoringWriter", {}
+            ).get("after", [None])[0]
```
theSaarco: This is a pretty complex logic to retrieve the class, please add a comment or something explaining it.
Also, why isn't this relevant to infra apps?

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -741,16 +750,203 @@ def _create_tsdb_tables(
...
+        :param labels:           List of labels to filter the response. Default is None.
+        :param include_stats:    If True, the function will include the statistics of the monitoring applications.
+                                 Currently, the statistics include the number of detections and possible detections.
+        :param include_infra:    If True, include the model monitoring infrastructure functions in the response.
+
+        :return:                A list of FunctionSummary objects, each representing a model monitoring function.
+        """
+
+        # Enrich response with infra functions
+        functino_summaries_list, base_period = self._enrich_function_summary_with_infra(
```
gtopper:
```suggestion
        function_summaries_list, base_period = self._enrich_function_summary_with_infra(
```
Unless it's meant to be in Italian 🇮🇹

gtopper: Also in 7 other places.

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -741,16 +750,203 @@ def _create_tsdb_tables(
...
+            ]
+        if not mm_functions:
+            logger.info("No model monitoring applications found")
+            return functino_summaries_list
+
+        detection_stats_dict = {}
+        if include_stats:
+            # enrich func stats with #detections and #possible_detections
+            start = start or (mlrun.utils.datetime_now() - timedelta(hours=24))
+            end = end or mlrun.utils.datetime_now()
```
gtopper:
```suggestion
            now = mlrun.utils.datetime_now()
            start = start or (now - timedelta(hours=24))
            end = end or now
```

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -741,16 +750,203 @@ def _create_tsdb_tables(
...
+                base_period = self._get_base_period(controller_func=controller_func)
+            except mlrun.errors.MLRunNotFoundError:
+                logger.info(
+                    "Model monitoring controller function not found, "
+                    "probably you need to re-enable model monitoring",
+                    project=self.project,
+                )
+        return function_summaries_list, base_period
+
+    def _enrich_function_summary_with_applications(
```
gtopper: I don't think there's a reason to 'enrich' the list (really append to it). This function can just return a list with the applications and leave it to the caller to concatenate it to any other items.

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -741,16 +750,203 @@ def _create_tsdb_tables(
...
+                    ),
+                }
+            functino_summaries_list.append(function_summary)
+        return functino_summaries_list
+
+    @staticmethod
+    def _get_base_period(controller_func: dict[str, typing.Any]) -> int:
+        base_period = 0
+        for env in controller_func["spec"]["env"]:
+            if env["name"] == "batch_intervals_dict":
```
gtopper: Seems like `"batch_intervals_dict"` has a constant so better use that for consistency.

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -741,16 +750,203 @@ def _create_tsdb_tables(
...
+                controller_func = services.api.crud.Functions().get_function(
+                    db_session=self.db_session,
+                    name=mm_constants.MonitoringFunctionNames.APPLICATION_CONTROLLER,
+                    project=self.project,
+                )
+                base_period = self._get_base_period(controller_func=controller_func)
+            except mlrun.errors.MLRunNotFoundError:
+                logger.info(
+                    "Model monitoring controller function not found, "
+                    "probably you need to re-enable model monitoring",
```
gtopper:
```suggestion
                    "Model monitoring controller function not found. "
                    "Try to re-enable model monitoring.",
```

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -741,16 +750,203 @@ def _create_tsdb_tables(
...
+            # getting the base period from the controller function
+            try:
+                controller_func = services.api.crud.Functions().get_function(
+                    db_session=self.db_session,
+                    name=mm_constants.MonitoringFunctionNames.APPLICATION_CONTROLLER,
+                    project=self.project,
+                )
+                base_period = self._get_base_period(controller_func=controller_func)
+            except mlrun.errors.MLRunNotFoundError:
+                logger.info(
```
gtopper, line 871: Are we sure we want to log it and ignore this error?

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -741,16 +750,203 @@ def _create_tsdb_tables(
...
+    ) -> tuple[list[mlrun.common.schemas.model_monitoring.FunctionSummary], int]:
+        """
+        Enrich the function summaries list with the model monitoring infrastructure functions.
+        In addition, it returns the base period of the controller function.
+
+        :param include_infra: If True, include the model monitoring infrastructure functions in the response.
+
+        :return: Returns a tuple containing a list of FunctionSummary objects and the base period of the controller
+                 function. If `include_infra` is False, the list will be empty.
+
```
gtopper, line 837:
```suggestion
```

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -741,16 +750,203 @@ def _create_tsdb_tables(
...
+        self,
+        include_infra: bool = True,
+    ) -> tuple[list[mlrun.common.schemas.model_monitoring.FunctionSummary], int]:
+        """
+        Enrich the function summaries list with the model monitoring infrastructure functions.
+        In addition, it returns the base period of the controller function.
+
+        :param include_infra: If True, include the model monitoring infrastructure functions in the response.
+
+        :return: Returns a tuple containing a list of FunctionSummary objects and the base period of the controller
```
gtopper:
```suggestion
        :return: a tuple containing a list of FunctionSummary objects and the base period of the controller
```

mlrun/common/schemas/model_monitoring/functions.py
```diff
@@ -0,0 +1,92 @@
...
+            if func_type == ModelMonitoringFunctionsType.APPLICATION
+            else "",
+            updated_time=func.metadata.updated,
+            status=func.status.state,
+            base_period=base_period,
+            stats=stats,
+        )
```
gtopper: I wonder if we need this if it's not used and a function object can be turned into a dict anyway.

The downside of having it is of course that it duplicates the logic in `from_dict`.

mlrun/common/schemas/model_monitoring/functions.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @classmethod
+    def from_dict(
```
gtopper: Maybe rename it to avoid misunderstanding?
```suggestion
    def from_function_dict(
```

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -694,22 +695,24 @@ def _get_v3io_frames_client(
...
+        if isinstance(filter_values, str):
+            return f"{filter_key}=='{filter_values}'"
+        elif isinstance(filter_values, list):
+            if len(filter_values) > V3IO_FRAMESD_MEPS_LIMIT:
                 logger.info(
-                    "The number of endpoint ids exceeds the v3io-engine filter-expression limit, "
-                    "retrieving all the model endpoints from the db.",
+                    "The number of filter values exceeds the v3io-engine filter-expression limit, "
+                    "retrieving all the values from the db.",
                     limit=V3IO_FRAMESD_MEPS_LIMIT,
```
gtopper: We will probably want to know what field this is
```suggestion
                    "The number of filter values exceeds the v3io-engine filter-expression limit, "
                    "retrieving all the values from the db.",
                    filter_key=filter_key,
                    limit=V3IO_FRAMESD_MEPS_LIMIT,
```

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -694,22 +695,24 @@ def _get_v3io_frames_client(
...
-                    amount=len(endpoint_id),
+                    amount=len(filter_values),
                 )
                 return None
-            return f"endpoint_id IN({str(endpoint_id)[1:-1]}) "
+            return f"{filter_key} IN({str(filter_values)[1:-1]}) "
         else:
             raise mlrun.errors.MLRunInvalidArgumentError(
-                f"Invalid 'endpoint_id' filter: must be a string or a list, endpoint_id: {endpoint_id}"
+                f"Invalid filter key {filter_key}: must be a string or a list, filter values: {filter_values}"
```
gtopper: Maybe mention the offending type, i.e. `type(filter_values).__name__`

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -1177,3 +1196,70 @@ def _enrich_mep_with_last_request(
...
+        start = start or (mlrun.utils.datetime_now() - timedelta(hours=24))
+        end = end or mlrun.utils.datetime_now()
+        filter_query = ""
+        if endpoint_ids:
+            filter_query = self._generate_filter_query(
+                filter_key=mm_schemas.ApplicationEvent.ENDPOINT_ID,
+                filter_values=endpoint_ids,
+            )
+        if application_names:
+            filter_query = self._generate_filter_query(
```
gtopper: This can override the query for `endpoint_ids`, won't it? Shouldn't we merge the two?

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -1177,3 +1196,70 @@ def _enrich_mep_with_last_request(
...
+        if df.empty:
+            return {}
+        else:
+            # convert application name to lower case
+            df[mm_schemas.ApplicationEvent.APPLICATION_NAME] = df[
+                mm_schemas.ApplicationEvent.APPLICATION_NAME
+            ].str.lower()
+
+            # convert result status to numerical values
+            # df[mm_schemas.ResultData.RESULT_STATUS] = df[mm_schemas.ResultData.RESULT_STATUS].astype(str).astype(int)
```
gtopper: Should probably either comment it back in or remove it.

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -741,16 +750,203 @@ def _create_tsdb_tables(
...
+        if infra_only:
+            # Model monitoring infrastructure functions
+            labels.append(
+                f"{mm_constants.ModelMonitoringInfraLabel.KEY}={mm_constants.ModelMonitoringInfraLabel.VAL}"
+            )
+        else:
+            # Model monitoring applications
+            labels.append(
+                f"{mm_constants.ModelMonitoringAppLabel.KEY}={mm_constants.ModelMonitoringAppLabel.VAL}"
+            )
```
gtopper: Looks like this will return only non-infra functions if `infra_only` is false?

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -741,16 +750,203 @@ def _create_tsdb_tables(
...
+            functino_summaries_list=functino_summaries_list,
+            base_period=base_period,
+            start=start,
+            end=end,
+            names=names,
+            labels=labels,
+            include_stats=include_stats,
+        )
+
+    def _enrich_function_summary_with_infra(
```
gtopper: This is definitely not enriching anything :). It returns a list of function summaries.

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -1177,3 +1196,70 @@ def _enrich_mep_with_last_request(
             mep.status.last_request = last_request_dictionary.get(
                 uid, mep.status.last_request
             )
+
+    def read_results_by_status(
```
gtopper, line 1202: For this function and maybe a few others, I'd consider creating a few unit tests.

server/py/services/api/api/endpoints/model_monitoring.py
```diff
@@ -327,3 +331,101 @@ def set_model_monitoring_credentials(
...
+    start: Optional[datetime] = None,
+    end: Optional[datetime] = None,
+) -> _FunctionSummariesParams:
+    """
+    Verify authorization and return common parameters.
+
+    :param project:         Project name.
+    :param auth_info:       The auth info of the request.
+    :param db_session:      A session that manages the current dialog with the database.
+    :returns:          A `_CommonParameters` object that contains the input data.
```
gtopper:
```suggestion
    :return:          A `_FunctionSummariesParams` object that contains the input data.
```

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -1177,3 +1196,70 @@ def _enrich_mep_with_last_request(
...
+        self,
+        start: Optional[Union[datetime, str]] = None,
+        end: Optional[Union[datetime, str]] = None,
+        endpoint_ids: Optional[Union[str, list[str]]] = None,
+        application_names: Optional[Union[str, list[str]]] = None,
+        result_status_list: Optional[list[int]] = None,
+    ) -> dict[tuple[str, int], int]:
+        start = start or (mlrun.utils.datetime_now() - timedelta(hours=24))
+        end = end or mlrun.utils.datetime_now()
```
```suggestion
        now = mlrun.utils.datetime_now()
        start = start or (now - timedelta(hours=24))
        end = end or now
```

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -328,6 +328,29 @@ def get_avg_latency(
...
 
+    @abstractmethod
+    def read_results_by_status(
+        self,
+        start: Optional[Union[datetime, str]] = None,
+        end: Optional[Union[datetime, str]] = None,
+        endpoint_ids: Optional[Union[str, list[str]]] = None,
+        application_names: Optional[Union[str, list[str]]] = None,
+        result_status_list: Optional[list[int]] = None,
+    ) -> dict[tuple[str, int], int]:
```
gtopper, line 339: Maybe add a `:return:` to the docstring to spell out what's being returned?

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -741,16 +750,203 @@ def _create_tsdb_tables(
...
+        mm_functions = self.list_model_monitoring_functions(
+            labels=labels, format_=mlrun.common.formatters.FunctionFormat.minimal
         )
+        if names:
+            mm_functions = [
+                fn for fn in mm_functions if fn["metadata"]["name"] in names
+            ]
+        if not mm_functions:
+            logger.info("No model monitoring applications found")
+            return functino_summaries_list
```
gtopper: It makes more logical sense to switch the ordering here.
```suggestion
        if not mm_functions:
            logger.info("No model monitoring applications found")
            return function_summaries_list

        if names:
            mm_functions = [
                fn for fn in mm_functions if fn["metadata"]["name"] in names
            ]
```

server/py/services/api/api/endpoints/model_monitoring.py
```diff
@@ -327,3 +331,101 @@ def set_model_monitoring_credentials(
...
+    )
+
+
+@router.get("/function-summaries")
+async def get_model_monitoring_function_summaries(
+    commons: Annotated[_FunctionSummariesParams, Depends(_common_function_parameters)],
+    names: Optional[list[str]] = Query(None, alias="name"),
+    labels: list[str] = Query([], alias="label"),
+    include_stats: bool = Query(True, alias="include_stats"),
+    include_infra: bool = Query(True, alias="include_infra"),
```
gtopper:
```suggestion
    include_stats: bool = Query(True, alias="include-stats"),
    include_infra: bool = Query(True, alias="include-infra"),
```

server/py/services/api/api/endpoints/model_monitoring.py
```diff
@@ -327,3 +331,101 @@ def set_model_monitoring_credentials(
...
+        client_version=client_version,
+        action=mlrun.common.schemas.AuthorizationAction.read,
+    )
+    if start is None and end is None:
+        end = mlrun.utils.helpers.datetime_now()
+        start = end - timedelta(days=1)
+    elif start is not None and end is not None:
+        if start.tzinfo is None or end.tzinfo is None:
+            raise mlrun.errors.MLRunInvalidArgumentTypeError(
+                "Custom start and end times must contain the timezone."
```
gtopper, line 381: Maybe we can add this check on the client side as well.

gtopper (commented):
Long review for a long PR :). Looks good on the whole, though.

mlrun/common/schemas/model_monitoring/functions.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        Create a FunctionSummary instance from a dictionary.
+        """
+
+        return cls(
+            type=func_type,
+            name=func_dict["metadata"]["name"],
+            application_class=""
+            if func_type != ModelMonitoringFunctionsType.APPLICATION
+            else func_dict["spec"]["graph"]["steps"]["PushToMonitoringWriter"]["after"][
```
theSaarco, line 59: Are we guaranteed that all these hierarchies exist in the dict provided?

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -328,6 +328,29 @@ def get_avg_latency(
...
+        """
+        Read results status from the TSDB and return a dictionary of results statuses by application name.
+
+        :param start:             The start time in which to read the results. By default, the last 24 hours
+                                  are read.
+        :param end:               The end time in which to read the results.
```
theSaarco:
```suggestion
        :param end:               The end time in which to read the results. Default is the current time (now).
```

server/py/services/api/api/endpoints/model_monitoring.py
theSaarco: Agree on 1st question - this doesn't make sense. It's perfectly valid for example to provide a different `start` (for example a week ago) and use the default `end` (now).

server/py/services/api/api/endpoints/model_monitoring.py
```diff
@@ -327,3 +331,101 @@ def set_model_monitoring_credentials(
...
+        end=end,
+    )
+
+
+@router.get("/function-summaries")
+async def get_model_monitoring_function_summaries(
+    commons: Annotated[_FunctionSummariesParams, Depends(_common_function_parameters)],
+    names: Optional[list[str]] = Query(None, alias="name"),
+    labels: list[str] = Query([], alias="label"),
+    include_stats: bool = Query(True, alias="include_stats"),
```
theSaarco: Shouldn't the default be `False`?

server/py/services/api/api/endpoints/model_monitoring.py
```diff
@@ -327,3 +331,101 @@ def set_model_monitoring_credentials(
...
+    include_infra: bool = Query(True, alias="include_infra"),
+) -> list[mlrun.common.schemas.model_monitoring.FunctionSummary]:
+    """Get monitoring function summaries for the specified project.
+
+    :param commons:                     The common parameters of the request.
+    :param names: List of function names to filter by (optional).
+    :param labels: Labels to filter by (optional).
+    :param include_stats: Whether to include statistics in the response (default is False).
```
theSaarco: Here you say it's `False`...

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -741,16 +750,203 @@ def _create_tsdb_tables(
...
+        :return: Returns a tuple containing a list of FunctionSummary objects and the base period of the controller
+                 function. If `include_infra` is False, the list will be empty.
+
+        """
+        function_summaries_list = []
+        base_period = 0
+        if include_infra:
+            infra_mm_functions = self.list_model_monitoring_functions(
+                format_=mlrun.common.formatters.FunctionFormat.full, infra_only=True
+            )
```
theSaarco: This seems to be conflicting logic - `include_infra` means to get infra+user-apps, while the list here asks for `infra_only`, so will only return infra functions.

theSaarco (changes requested):
Generally looks good, with some questions/comments.

server/py/services/api/crud/model_monitoring/deployment.py
Eyal-Danieli: Grazie

server/py/services/api/api/endpoints/model_monitoring.py
Eyal-Danieli: It's not in `_verify_authorization`, but in `_common_function_parameters` right after `_verify_authorization`. I tried to keep consistency with similar validation we do in the `metrics-values` API (as part of `_get_metrics_values_params`) but yes in this case, the user should also be able to provide just one of `start_time` or `end_time`, and we’ll use a default value for the other. I'll update it.

server/py/services/api/crud/model_monitoring/deployment.py
Eyal-Danieli: not sure what is better because if both `start` and `end` are provided then there is no reason to call `mlrun.utils.datetime_now()` which is the case in my approach ("short circuit evaluation").

server/py/services/api/crud/model_monitoring/deployment.py
Eyal-Danieli: yes, because this isn’t a real issue. If the user disables mm after data already exists, we remove the controller func, even if application functions remain. In that case, there is no `base period` but the missing controller isn’t an error, it’s expected behavior.

mlrun/common/schemas/model_monitoring/functions.py
Eyal-Danieli: The reason we have it it's because it can be useful for the client side. When user for example call `get_function` or `list_functions` he gets runtimes. With this method, he could easily convert that runtime into a `FunctionSummary` object.

gtopper: In that case, can we just implement it with `to_dict` and `from_function_dict`? To avoid duplicating the logic.

Eyal-Danieli: amm yep we can do that so I'll remove that `from_func`

server/py/services/api/crud/model_monitoring/deployment.py
Eyal-Danieli: I'll rename the param to `function_type` for better clarity. This param will accept values defined in `ModelMonitoringFunctionsType`, which currently includes "application" and "infra". By default this value will be set to `application`.

mlrun/common/schemas/model_monitoring/functions.py
Eyal-Danieli: Because infra apps don't include this step, they don't have an actual "application class". Anyway, following @gtopper review we removed that part (https://github.com/mlrun/mlrun/pull/7959#discussion_r2109125308).

mlrun/common/schemas/model_monitoring/functions.py
Eyal-Danieli: All the mm apps share a similar graph structure, and this is the only place where we can retrieve the original app class name.

server/py/services/api/api/endpoints/model_monitoring.py
Eyal-Danieli: amm in my opinion, it makes more sense for the API (the main client of this API is the UI) to include the statistics by default. However, for the SDK, I believe it's better to keep this option disabled  by default, and only return the tsdb stats if the user explicitly requests them.
If we have to keep consistency here, I'll change it to `True` on the SDK side.

server/py/services/api/crud/model_monitoring/deployment.py
Eyal-Danieli: in this `_get_function_summary_infra` we return `base_period` and list of infra funcs. It has nothing to do with the user-apps. If `include_infra` is False, it returns just the base period (and an empty list). Perhaps we should choose a clearer name, such as `enrich_with_infra` .

server/py/services/api/crud/model_monitoring/deployment.py
gtopper: The point isn't to save on cycles, but to ensure that they refer to the same time point, so that they are exactly 24 hours apart if neither was provided.

mlrun/common/schemas/model_monitoring/functions.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class FunctionsType(enum.Enum):
+    APPLICATION = "application"
+    INFRA = "infra"
+
+
+class FunctionSummary(BaseModel):
+    """
+    Function summary model. Includes metadata about the function, such as its name, but also statistical metrics such
```
gtopper:
```suggestion
    Function summary model. Includes metadata about the function, such as its name, as well as statistical metrics such
```

mlrun/db/httpdb.py
```diff
@@ -4118,6 +4119,52 @@ def set_model_monitoring_credentials(
...
+        response = self.api_call(
+            method=mlrun.common.types.HTTPMethod.GET,
+            path=path,
+            params={
+                "start": datetime_to_iso(start),
+                "end": datetime_to_iso(end),
+                "name": names,
+                "label": labels,
+                "include_stats": include_stats,
+                "include_infra": include_infra,
```
gtopper: Doesn't seem to me like this should work. Are we missing test coverage?
```suggestion
                "include-stats": include_stats,
                "include-infra": include_infra,
```

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -328,6 +328,37 @@ def get_avg_latency(
...
+
+        :param start:             The start time in which to read the results. By default, the last 24 hours
+                                  are read.
+        :param end:               The end time in which to read the results. Default is the current time (now).
+        :param endpoint_ids:      Optional list of endpoint ids to filter the results by. By default, all
+                                  endpoint ids are included.
+        :param application_names: Optional list of application names to filter the results by. By default, all
+                                  application are included.
+        :param result_status_list: Optional list of result statuses to filter the results by. By default, all
+                                  result statuses are included.
```
gtopper: Parameter list needs horizontal realignment.

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -694,22 +695,26 @@ def _get_v3io_frames_client(
...
+                    "The number of filter values exceeds the v3io-engine filter-expression limit, "
+                    "retrieving all the values from the db.",
+                    filter_key=filter_key,
                     limit=V3IO_FRAMESD_MEPS_LIMIT,
-                    amount=len(endpoint_id),
+                    amount=len(filter_values),
                 )
                 return None
-            return f"endpoint_id IN({str(endpoint_id)[1:-1]}) "
+            return f"{filter_key} IN({str(filter_values)[1:-1]}) "
```
```suggestion
            return f"{filter_key} IN ({str(filter_values)[1:-1]}) "
```

gtopper: Also feeling a bit uneasy about relying on Python string representation of a list of strings and how it accidentally happens to map to V3IO filter expression syntax.

I guess the alternative is something like
```python
",".join([f"'{val}'" for val in filter_values])
```

tests/model_monitoring/test_stores/test_v3io.py
```diff
@@ -302,3 +388,25 @@ def test_normalize_dict_for_v3io_frames(
     input_event: dict[str, Any], expected_output: dict[str, Any]
 ) -> None:
     assert _normalize_dict_for_v3io_frames(input_event) == expected_output
+
+
+@pytest.mark.usefixtures("_mock_frames_client_extended")
+def test_read_results_by_status():
```
gtopper, line 394: Great, but maybe we should test some of these parameters as well?
```python
        start: Optional[Union[datetime, str]] = None,
        end: Optional[Union[datetime, str]] = None,
        endpoint_ids: Optional[Union[str, list[str]]] = None,
        application_names: Optional[Union[str, list[str]]] = None,
```

server/py/services/api/api/endpoints/model_endpoints.py
```diff
@@ -572,21 +572,15 @@ async def _get_metrics_values_params(
...
+    end = end or mlrun.utils.helpers.datetime_now()
+    start = start or (end - timedelta(days=1))
+    if start.tzinfo is None or end.tzinfo is None:
+        raise mlrun.errors.MLRunInvalidArgumentTypeError(
+            "Custom start and end times must contain the timezone."
+        )
+    if start > end:
         raise mlrun.errors.MLRunInvalidArgumentError(
-            "Provided only one of start time, end time. Please provide both or neither."
+            "The start time must precede the end time."
```
gtopper: This error message might be a bit confusing if the user provides a future start time and no end time.

gtopper (commented):
Mostly minor comments (all but the `api_call` one).

tests/model_monitoring/test_stores/test_v3io.py
Eyal-Danieli: these params are considered in the DB query so testing them requires access to a real db to verify that the filtering logic returns the expected results. We can do that but not as part of a unit test.

server/py/services/api/api/endpoints/model_endpoints.py
Eyal-Danieli: I took it from other existing API but I'll try to make it more clear

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -694,22 +695,26 @@ def _get_v3io_frames_client(
...
+                    "The number of filter values exceeds the v3io-engine filter-expression limit, "
+                    "retrieving all the values from the db.",
+                    filter_key=filter_key,
                     limit=V3IO_FRAMESD_MEPS_LIMIT,
-                    amount=len(endpoint_id),
+                    amount=len(filter_values),
                 )
                 return None
-            return f"endpoint_id IN({str(endpoint_id)[1:-1]}) "
+            return f"{filter_key} IN ({', '.join(repr(v) for v in filter_values)}) "
```
gtopper, line 713:
```suggestion
            return f"{filter_key} IN ({','.join(repr(v) for v in filter_values)}) "
```

gtopper (approved):
LGTM ✅

theSaarco (approved):
LGTM

mlrun/common/schemas/model_monitoring/functions.py
```diff
@@ -0,0 +1,66 @@
...
+    updated_time: datetime
+    status: Optional[str] = None
+    base_period: Optional[int] = None
+    stats: Optional[dict] = None
+
+    @classmethod
+    def from_function_dict(
+        cls,
+        func_dict: dict,
+        func_type=FunctionsType.APPLICATION,
```
assaf758, line 46: typing hint is missing

mlrun/common/schemas/model_monitoring/functions.py
```diff
@@ -0,0 +1,66 @@
...
+        base_period: Optional[int] = None,
```
assaf758, line 47: does it make sense to not have base_period for a function?

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -328,6 +328,36 @@ def get_avg_latency(
         If an endpoint has not been invoked within the specified time range, it will not appear in the result.
         """
 
+    @abstractmethod
+    def read_results_by_status(
```
assaf758, line 332:
```suggestion
    def count_results_by_status(
```

assaf758 (approved):
Nice!
Some minor comments included for the follow-up pr.

mlrun/common/schemas/model_monitoring/functions.py
Eyal-Danieli: yes, for example when the mm is disabled

PR #7960, opened 2025-05-22
[Requirements] Bump KFP common

moranbental:
according these changes—https://github.com/mlrun/mlrun/pull/7942
https://iguazio.atlassian.net/browse/ML-10051


liranbg (changes requested):
run lock to update requirement locks

moranbental:
> run lock to update requirement locks

I always forget this

PR #7877, opened 2025-05-14
[Model Monitoring] Refactor server endpoint deprecations

gtopper:
Relates to [ML-9792](https://iguazio.atlassian.net/browse/ML-9792).

[ML-9792]: https://iguazio.atlassian.net/browse/ML-9792?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #7962, opened 2025-05-22
[Model Monitoring] Fix `list_model_endpoints`'s `latest-only` parameter

gtopper:
Following #7668 and #7877.

### DIFF
--- a/server/py/services/api/api/endpoints/model_endpoints.py
+++ b/server/py/services/api/api/endpoints/model_endpoints.py
@@ -245,7 +245,7 @@ async def list_model_endpoints(
     tsdb_metrics: bool = Query(True, alias="tsdb-metrics"),
     metric_list: Optional[list[str]] = Query(None, alias="metric"),
     uids: list[str] = Query(None, alias="uid"),
-    latest_only: bool = Query(False, alias="latest-only-old"),
+    latest_only: bool = Query(False, alias="latest-only"),
     auth_info: schemas.AuthInfo = Depends(framework.api.deps.authenticate_request),
     db_session: Session = Depends(deps.get_db_session),
 ) -> schemas.ModelEndpointList:


server/py/services/api/api/endpoints/model_endpoints.py
```diff
@@ -245,7 +245,7 @@ async def list_model_endpoints(
     tsdb_metrics: bool = Query(True, alias="tsdb-metrics"),
     metric_list: Optional[list[str]] = Query(None, alias="metric"),
     uids: list[str] = Query(None, alias="uid"),
-    latest_only: bool = Query(False, alias="latest-only-old"),
+    latest_only: bool = Query(False, alias="latest-only"),
```
moranbental, line 248: You are the BC here, since the query parameter name has changed

moranbental (changes requested):

server/py/services/api/api/endpoints/model_endpoints.py
gtopper: This change is for 1.8.x as well, so no deprecation.

gtopper: Created #7969.

gtopper: And #7970.

theSaarco (approved):
👍 
@moranbental - this is fine since we had BC code and we remove it as it's not needed (parameter introduced at 1.8)

server/py/services/api/api/endpoints/model_endpoints.py
moranbental: OK, your PR was against the development branch, so it was a bit confusing.

moranbental (approved):
LGTM

PR #7968, opened 2025-05-22
[Notifications] Fix push notifications session [1.8.x]

liranbg:
Refactors the KFPNotificationPusher to receive a database session, enabling proper access to workflow steps. Also modifies Slack notification to use runtime constants.

addressing an issue where run retrieval for workflow step run is failing to be retrieved due to changes in pod hostname conventions between KFP versions

https://iguazio.atlassian.net/browse/ML-10086

### DIFF
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -2124,9 +2124,7 @@ class Workflow:
 
                 _run = _runs[0]
             except IndexError:
-                logger.warning(
-                    "No run found for step", step=_step.to_dict()
-                )
+                logger.warning("No run found for step", step=_step.to_dict())
                 _run = {
                     "metadata": {
                         "name": _step.display_name,
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -2088,7 +2088,9 @@ def join_urls(base_url: Optional[str], path: Optional[str]) -> str:
 
 class Workflow:
     @staticmethod
-    def get_workflow_steps(db, workflow_id: str, project: str) -> list:
+    def get_workflow_steps(
+        db: "mlrun.db.RunDBInterface", workflow_id: str, project: str
+    ) -> list:
         steps = []
 
         def _add_run_step(_step: mlrun_pipelines.models.PipelineStep):
@@ -2106,8 +2108,8 @@ class Workflow:
                     node_name_initials, node_name_generated_id = _step.node_name.rsplit(
                         "-", 1
                     )
-                    # x-y, Z, N -> x-y-Z-N
-                    # runner_pod = x-y-Z-N
+                    # compile the expected runner pod hostname as per kfp >= 2.4
+                    # x-y, Z, N -> runner_pod = x-y-Z-N
                     runner_pod_value = "-".join(
                         [node_name_initials, _step.display_name, node_name_generated_id]
                     )
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -2088,9 +2088,7 @@ def join_urls(base_url: Optional[str], path: Optional[str]) -> str:
 
 class Workflow:
     @staticmethod
-    def get_workflow_steps(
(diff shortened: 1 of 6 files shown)


alonmr (approved):

alonmr (approved):

mlrun/utils/helpers.py
```diff
@@ -2088,22 +2088,47 @@ def join_urls(base_url: Optional[str], path: Optional[str]) -> str:
 
 class Workflow:
     @staticmethod
-    def get_workflow_steps(workflow_id: str, project: str) -> list:
+    def get_workflow_steps(db, workflow_id: str, project: str) -> list:
```
adamdelman: It's not clear to me why we need the DB session here, is the problem related to the session isolation level?
Or would an explicit commit in the the session affecting the runs or something like a `session.refresh(_runs)`
Also, please add type annotations.

mlrun/utils/helpers.py
```diff
@@ -2088,22 +2088,47 @@ def join_urls(base_url: Optional[str], path: Optional[str]) -> str:
...
                     project=project,
                     labels=f"{mlrun_constants.MLRunInternalLabels.runner_pod}={_step.node_name}",
-                )[0]
+                )
+                if not _runs:
+                    # x-y-N -> x-y, N
+                    node_name_initials, node_name_generated_id = _step.node_name.rsplit(
+                        "-", 1
+                    )
+                    # x-y, Z, N -> x-y-Z-N
```
adamdelman: These comments are a bit unclear.

server/py/framework/utils/notifications/notification_pusher.py
```diff
@@ -365,12 +368,19 @@ def _update_notification_status(
...
     def __init__(
         self,
+        db_session: sqlalchemy.orm.Session,
         project: str,
         workflow_id: str,
         notifications: list[mlrun.common.schemas.Notification],
         default_params: typing.Optional[dict] = None,
     ):
         self._project = project
+
```
adamdelman, line 378: I think it would be best to put this in a contextvars.ContextVar storage

server/py/framework/utils/notifications/notification_pusher.py
```diff
@@ -398,26 +408,28 @@ def __init__(
                 )
 
     def push(self, sync_push_callback=None, async_push_callback=None):
-        def sync_push():
+        def sync_push(runs_):
```
adamdelman: type annotation

server/py/framework/utils/notifications/notification_pusher.py
```diff
@@ -398,26 +408,28 @@ def __init__(
...
+                        runs_,
                     )
                 except Exception as exc:
                     logger.warning(
                         "Failed to push notification sync",
                         error=mlrun.errors.err_to_str(exc),
                     )
 
-        async def async_push():
+        async def async_push(runs_):
```
adamdelman: type annotation

adamdelman (changes requested):
Overall looks good, some issues need to be addressed.

mlrun/utils/helpers.py
liranbg: type annotation 👍🏼 
about the problem - the function used to initialize the db access via a global session which could potentially not be able to see recent run update changes as it may have been created a bit before. that is not the RCA for this issue but rather a high-potential one. in addition, the reason we initialized the db access at all is due to a code-locality bug whereas notification pusher classes are still co-located between client and server whereas the db could be once the actual db session and sometimes http-like client. anyway, to utilize the session created for pushing the notification, I have simply added the argument to ensure we have "fresh" session and moved the `get_run_db()` for client-side function callee.

server/py/framework/utils/notifications/notification_pusher.py
liranbg: not for 1.8 and perhaps not even after - lets discuss offline

PR #8014, opened 2025-05-28
[Deprecations] Remove `labels` param in `list_schedules`

moranbental:
`labels` in `GET /projects/{project}/schedules?labels` is deprecated in 1.7.0
Use `label` instead which is an array of strings
https://iguazio.atlassian.net/browse/ML-10045


moranbental:
It fails on BC, which is expected because we are removing the query parameter from the endpoint.

alonmr (approved):

PR #8015, opened 2025-05-28
[Runs] Add retry run spec and validation [feature/retry-job]

alonmr:
1st PR for the retry jobs feature:
1. Adds retry spec to run object
2. Adds spec validation in the launchers

Notes:
* Only supported for kubejobs
* Actual retry implementation will be added separately

https://iguazio.atlassian.net/browse/ML-10112


requirements.txt
```diff
@@ -42,5 +42,6 @@ orjson>=3.9.15, <4
 mlrun-pipelines-kfp-common~=0.5.5
 mlrun-pipelines-kfp-v1-8~=0.5.4
 docstring_parser~=0.16
+timelength~=1.1
```
liranbg: bump to 3.x if you're at it

mlrun/config.py
```diff
@@ -264,6 +264,12 @@
                     "executing": "24h",
                 }
             },
+            "retry": {
+                "backoff": {
+                    "default_base_delay": "30s",
```
liranbg, line 269: is the `base` redundant?

mlrun/launcher/base.py
```diff
@@ -187,13 +189,45 @@ def _validate_run_params(self, parameters: dict[str, Any]):
...
+    def _validate_run_single_param(cls, param_name: str, param_value: int):
         # verify that integer parameters don't exceed a int64
         if isinstance(param_value, int) and abs(param_value) >= 2**63:
             raise mlrun.errors.MLRunInvalidArgumentError(
                 f"Parameter {param_name} value {param_value} exceeds int64"
             )
 
+    @staticmethod
+    def _validate_retry(runtime_kind: str, retry: Optional["mlrun.model.Retry"]):
+        if retry is None or not retry.count:
```
liranbg: why is the retry on base and not on server side launcher? because local=True kind of jobs?

liranbg (commented):
very well. minor comments

mlrun/config.py
alonmr: it's base delay because the actual delay is `f(base_delay, attempt)`

mlrun/launcher/base.py
alonmr: good point, perhaps better to move to server side and it will be ignored for local=True runs

liranbg (approved):

requirements.txt
alonmr: I reverted the bump because they change the error handling and it requires some changes.
Moved the requirement back to server side so no effect on client
I liked the 1.x implementation better, perhaps worth upgrading to 2.x

PR #8016, opened 2025-05-28
[Dask] Raise on dask cluster start timeout

Yacouby:
This PR fixes an issue where system tests using an invalid Dask image were incorrectly passing.

- Added a check to raise `MLRunRuntimeError` if the background task for starting the Dask cluster does not reach a terminal state before the timeout expires.
- This ensures that deployment failures (e.g., due to an invalid image) are properly detected and surfaced.
- Reduced the Dask cluster startup timeout from 10 minutes to 5 minutes to shorten feedback loops and avoid unnecessarily long waits during failures.

Jira :
https://iguazio.atlassian.net/browse/ML-10125

### DIFF
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -234,7 +234,10 @@ default_config = {
                 "model_endpoint_creation": "600",
                 "model_endpoint_tsdb_leftovers": "900",
             },
-            "runtimes": {"dask": "600"},
+            "runtimes": {
+                "dask": "600",
+                "dask_cluster_start": "300",
+            },
             "push_notifications": "60",
         },
     },
--- a/mlrun/runtimes/daskjob.py
+++ b/mlrun/runtimes/daskjob.py
@@ -255,7 +255,11 @@ class DaskCluster(KubejobRuntime):
         background_task = db.start_function(func_url=self._function_uri())
         if watch:
             now = datetime.datetime.utcnow()
-            timeout = now + datetime.timedelta(minutes=10)
+            timeout = now + datetime.timedelta(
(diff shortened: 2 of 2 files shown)


mlrun/runtimes/daskjob.py
```diff
@@ -282,6 +286,9 @@ def _start(self, watch=True):
                         return
                 time.sleep(5)
                 now = datetime.datetime.utcnow()
+            raise mlrun.errors.MLRunRuntimeError(
```
TomerShor:
```suggestion
            raise mlrun.errors.MLRunTimeoutError(
```

mlrun/config.py
```diff
@@ -234,7 +234,10 @@
                 "model_endpoint_creation": "600",
                 "model_endpoint_tsdb_leftovers": "900",
             },
-            "runtimes": {"dask": "600"},
+            "runtimes": {
+                "dask": "600",
+                "dask_cluster_start": "300",
```
TomerShor, line 239: Is 5 minutes a reasonable timeout for dask functions?
Can you check for average start time (without errors) and what does it depend on?

TomerShor (changes requested):

mlrun/config.py
Yacouby: A 5-minute timeout is generally reasonable for most environments. Typically, Dask clusters start within seconds to a few minutes.
However, startup time depends on factors like dask cluster size and resources (larger clusters with more workers or GPUs take longer) and docker image pull time.

TomerShor (approved):

PR #8018, opened 2025-05-29
[Datastore] Make TDEngine connector thread safe

alxtkr77:
https://iguazio.atlassian.net/browse/ML-10131

### DIFF
--- a/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connection.py
+++ b/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connection.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import traceback
 from collections.abc import Callable
 from enum import Enum
 from typing import Any, Final, Optional, Union
@@ -135,34 +134,6 @@ class Statement:
         statement.bind_param(bind_params)
         statement.add_batch()
         return statement
-
-
-def _run(connection_string, prefix_statements, q, statements, query):
-    try:
-        conn = taosws.connect(connection_string)
-
-        for statement in prefix_statements + statements:
-            if isinstance(statement, Statement):
-                prepared_statement = statement.prepare(conn.statement())
-                prepared_statement.execute()
-            else:
-                conn.execute(statement)
-
-        if not query:
-            q.put(None)
-            return
-
-        res = conn.query(query)
-
-        # taosws.TaosField is not serializable
-        fields = [
-            Field(field.name(), field.type(), field.bytes()) for field in res.fields
-        ]
-
-        q.put(QueryResult(list(res), fields))
-    except Exception as e:
-        tb = traceback.format_exc()
-        q.put(ErrorResult(tb, e))
 
 
 class TDEngineConnection:
--- a/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connection.py
+++ b/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connection.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 import time
 from collections.abc import Callable
 from enum import Enum
@@ -20,6 +19,7 @@ from typing import Any, Final, Optional, Union
 import taosws
 from taosws import TaosStmt
 
+import mlrun
 from mlrun.utils import logger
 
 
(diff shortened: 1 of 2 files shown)


gtopper (commented):
Looks good. It may or may not fix the issue but it makes sense. Only concern I have is we might end up opening too many connections if the thread pool is large. That's probably not the case, but it's possible. We may also actually increase the likelihood of a connection error if we have more connections (especially if concurrency wasn't the cause of the error in the ticket).

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connection.py
```diff
@@ -137,40 +139,123 @@ def prepare(self, statement: TaosStmt) -> TaosStmt:
...
+            "closed",
+            "refused",
+            "unreachable",
+            "disconnected",
+            "connection",
+            "network",
+            "socket",
+            "lost",
+            "dropped",
+            "reset",
```
gtopper: I think this is way too likely to yield false positives as well as false negatives.

gtopper: There's also a `taosws.ConnectionError` which we should probably assume is a connection error regardless of the error message.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connection.py
```diff
@@ -181,33 +266,38 @@ def run(
...
-                    self._conn.execute(statement)
-                except taosws.Error as e:
-                    raise TDEngineError(
-                        f"Failed to run statement `{statement}`: {e}"
-                    ) from e
+                operation_name += f" `{statement}`"
 
+            self._execute_with_retry(self._execute_statement, operation_name, statement)
+
+        # Execute query if provided
```
gtopper:
```suggestion
```

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connection.py
```diff
@@ -181,33 +266,38 @@ def run(
...
         return QueryResult(list(res), fields)
+
+    def close(self):
+        """Close the connection."""
+        try:
+            if hasattr(self, "_conn") and self._conn:
+                self._conn.close()
+                logger.debug("TDEngine connection closed")
+        except Exception as e:
+            logger.warning(f"Error closing TDEngine connection: {e}")
```
gtopper: Better use `err_to_str` from `mlrun.errors`.

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connection.py
```diff
@@ -181,33 +266,38 @@ def run(
...
         fields = [
             Field(field.name(), field.type(), field.bytes()) for field in res.fields
         ]
 
         return QueryResult(list(res), fields)
+
+    def close(self):
+        """Close the connection."""
+        try:
+            if hasattr(self, "_conn") and self._conn:
```
gtopper: This if checks for things that cannot happen.
```suggestion
```

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connection.py
alxtkr77: So you propose always retry except the taosws.ConnectionError ?

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connection.py
alxtkr77: Maybe leave only 'if self._conn:'. It is a good practice to check things even if you think its not supposed to be

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connection.py
gtopper: Always retry on connection error, and then also retry on other specific errors that we think are related (like "channel closed"). I don't think we should use a list of keywords (where does this list even come from?).

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connection.py
gtopper: Then the number of things you might check for is unlimited though. E.g. verify the type, in case `_conn` is a string or an interger.

alxtkr77: disagree

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connection.py
alxtkr77: This list was generated by great ChatGPT and myself :) If you don't have an alternative list, I think we should just retry on any error, which makes sense since today any error is unrecoverable.

gtopper (approved):

PR #8023, opened 2025-06-01
[Docs] Update release month in change log [1.8.x]

jillnogold:


alonmr:
the base branch of your branch is 1.8 but it should be development. you need to `git rebase development` but you might hit conflicts so I suggest opening a new branch (checkout from development) and cherry-pick the commit there

PR #8029, opened 2025-06-03
[Model Monitoring] Connect `ModelRunnerStep` to Model Monitoring

royischoss:
https://iguazio.atlassian.net/browse/ML-9891
Contains:
1. Adding system steps to server graph allowing pre-processing to the response from `ModelRunnerStep` and adding mm-stream target, full cycle was tested and added to test_app_flow using `ModelRunnerStep`
2. Adding unit tests checking functionality using to_mock_server 
3. Adding system test allowing to check `test_app_flow` and `TestBatchServingWithSampling::test_serving` using `ModelRunnerStep`


mlrun/serving/server.py
```diff
@@ -360,14 +368,124 @@ def add_system_steps_to_graph(graph: RootFlowStep):
     return graph
 
 
+def apply_monitoring_general_steps(
```
theSaarco: Why `apply` rather than `add`?
Also, is this expected to be idempotent? If I already deployed this graph and got all these steps in the graph, running it again will add those steps again?

mlrun/serving/server.py
```diff
@@ -360,14 +368,124 @@ def add_system_steps_to_graph(graph: RootFlowStep):
...
+            raise GraphError(
+                "Expected model runner step to be followed by error raiser step"
+            )
+    return graph
+
+
+def add_system_steps_to_graph(
+    project: str, graph: RootFlowStep, track_models: bool, context, serving_spec
+) -> RootFlowStep:
+    graph = add_error_raiser_step(graph)
```
theSaarco: What if the graph has no MRS steps? You would still add the MM steps which will be doing pretty much nothing. Also, in both `add_error_raiser_step` and in `add_monitoring_pre_process_steps` you iterate over all the graph, looking for MRS steps. It may be a better approach to split this:
1. Look for MRS steps in the graph. If none exist, no need for further processing.
2. Call `add_error_raiser_step` with this list of steps.
3. Call `add_monitoring_pre_process_steps` with the list of steps.

mlrun/serving/server.py
```diff
@@ -360,14 +368,124 @@ def add_system_steps_to_graph(graph: RootFlowStep):
...
+            sharding_func=mlrun.common.schemas.model_monitoring.constants.StreamProcessingEvent.ENDPOINT_ID,
+            after="filter_none_2",
+        )
+    return graph, monitor_flow_step
+
+
+def add_monitoring_pre_process_steps(
+    project: str, graph: RootFlowStep, context, serving_spec
+) -> RootFlowStep:
+    graph, monitor_flow_step = apply_monitoring_general_steps(
```
theSaarco: Why call this here and not separately by the `add_system_steps_to_graph` method?

mlrun/serving/server.py
```diff
@@ -360,14 +368,124 @@ def add_system_steps_to_graph(graph: RootFlowStep):
...
+        project, graph, context, serving_spec
+    )
+    model_runner_steps_names = {
+        step.name: step
+        for step in graph.steps.values()
+        if isinstance(step, mlrun.serving.states.ModelRunnerStep)
```
theSaarco: It would be best if we generalize this check - rather than testing for a specific class, have this class be a subclass of some `MonitoredStep` class (or Mixin), and check if it's an instance of that class. This will make it easier to extend the MM functionality to various other steps in the future (we do want to enable monitoring on a wide range of steps). Let's discuss this.

mlrun/serving/system_steps.py
```diff
@@ -0,0 +1,439 @@
...
+        self.context = copy(context)
+        self.model_endpoints: dict[str, [dict[str, str]]] = {}
+        self.labels: dict[str, dict[str, str]] = {}
+        self.input_path: dict[str, str] = {}
+        self.result_path: dict[str, str] = {}
+        self.output_schema: dict[str, Union[list[str], str]] = {}
+
+        server: mlrun.serving.GraphServer = getattr(
+            self.context, "_server", None
+        ) or getattr(self.context, "server", None)
```
theSaarco: Why do we need this elaborate code to get the server from the context? Is this because the `context` object may be different in various scenarios?

theSaarco: Also, there's no check for a situation where `server` is `None`. I understand this is not expected, but the next line would explode in that case, it's probably better to do a sanity test and show an indicative error here.

mlrun/serving/system_steps.py
```diff
@@ -0,0 +1,439 @@
...
+                        monitoring_data.get(model, {}).get(
+                            mlrun.common.schemas.MonitoringData.RESULT_PATH
+                        )
+                    )
+                    self.output_schema[model] = monitoring_data.get(model, {}).get(
+                        mlrun.common.schemas.MonitoringData.OUTPUTS
+                    )
+                    self.models_uri[model] = monitoring_data.get(model, {}).get(
+                        mlrun.common.schemas.MonitoringData.MODEL_PATH
+                    )
```
theSaarco: Isn't it better to actually have this logic localized in the `ModelRunnerStep` class? Why do this here?
This also goes back to my previous comment on creating a `MonitoredStep` base class - I would expect this logic to sit there, where it can be easily overridden if needed by child-classes (that are not MRS). The way it's currently written, when expanding to other classes you'll need to add an if-else here to handle each case separately, rather than have the logic in the class itself.

mlrun/serving/system_steps.py
```diff
@@ -0,0 +1,439 @@
...
+                    else None,
+                    mm_schemas.StreamProcessingEvent.METRICS: event.body[
+                        mm_schemas.StreamProcessingEvent.METRICS
+                    ]
+                    if mm_schemas.StreamProcessingEvent.METRICS in event.body
+                    else None,
+                }
+            )
+        event.body = monitoring_event_list
+        logger.info("monitoring pre processor ended", event=event)
```
theSaarco: Why print the entire event in `info` log? Seems a bit too much.

mlrun/serving/system_steps.py
```diff
@@ -0,0 +1,439 @@
...
+    def do(self, event):
+        if (self.context and self.context.is_mock) or self.context is None:
+            return event
+        if (
+            self._background_task_status
+            == mlrun.common.schemas.BackgroundTaskState.running
+            and (
+                self._background_task_check_timestamp is None
+                or mlrun.utils.now_date() - self._background_task_check_timestamp
+                >= timedelta(seconds=15)
```
theSaarco: Can we make this a configurable parameter? Allowing users/admins to modify the frequency? Even an MLRun config value would be better than a fixed value that we cannot change in the field.

mlrun/serving/system_steps.py
```diff
@@ -0,0 +1,439 @@
...
+        return event
+
+    def _background_task_succeeded(
+        self, background_task: mlrun.common.schemas.BackgroundTask
+    ):
+        logger.info(
+            "Checking model endpoint creation task status",
+            task_name=self.server.model_endpoint_creation_task_name,
+        )
+        self._background_task_status = background_task.status.state
```
theSaarco: Since this method also updates the status in the class, maybe change its name - I was surprised this line is here since I thought this just does the check itself and return the status.

mlrun/serving/system_steps.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+class SamplingStep(storey.MapClass):
+    def __init__(
+        self,
+        sampling_percentage: float,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.sampling_percentage = sampling_percentage
```
theSaarco: Validate that `sampling_percentage` is between 0 and 100.

mlrun/serving/system_steps.py
```diff
@@ -0,0 +1,439 @@
...
+        return (
+            background_task.status.state
+            == mlrun.common.schemas.BackgroundTaskState.succeeded
+        )
+
+
+class SamplingStep(storey.MapClass):
+    def __init__(
+        self,
+        sampling_percentage: float,
```
theSaarco: Why do we need float here?

mlrun/serving/system_steps.py
```diff
@@ -0,0 +1,439 @@
...
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        self.sampling_percentage = sampling_percentage
+
+    def do(self, event):
+        logger.info(
+            "sampling step runs",
+            event=event,
```
theSaarco, line 339: Why print the entire event in `info` logs?

mlrun/serving/system_steps.py
```diff
@@ -0,0 +1,439 @@
...
+
+            if isinstance(request["outputs"], list):
+                event[mm_schemas.StreamProcessingEvent.RESPONSE]["outputs"] = [
+                    request["outputs"][i] for i in sampled_requests_indices
+                ]
+        event[mm_schemas.EventFieldType.SAMPLING_PERCENTAGE] = self.sampling_percentage
+        event[mm_schemas.EventFieldType.EFFECTIVE_SAMPLE_COUNT] = len(
+            event.get(mm_schemas.StreamProcessingEvent.REQUEST, {}).get("inputs", [])
+        )
+        logger.info("sampling step ended", event=event)
```
theSaarco: And again...
You need to cleanup the logs in general.

theSaarco (changes requested):
Reviewed (didn't go over the tests yet). 
All in all looks good, I have some suggestions for code restructuring to support future expansion of MM to other classes.

mlrun/serving/system_steps.py
royischoss: `set_tracking` gets a float and in `V2ModelServer` we also added hint of float kept it the same

mlrun/serving/system_steps.py
royischoss: I made the change outside and kept the name

mlrun/serving/server.py
royischoss: Made some changes think it now doing more efficient work there

mlrun/serving/server.py
royischoss: Made this change as suggested

mlrun/serving/server.py
royischoss: Added a field to RootFlow suggesting a ModelRunnerStep was added to graph, collecting the ModelRunnerSteps once and than adding the connections needed when adding error raiser and monitoring flow

mlrun/serving/server.py
royischoss: changed

mlrun/serving/system_steps.py
royischoss: Made the change we talked about

mlrun/serving/system_steps.py
royischoss: Will do

mlrun/serving/system_steps.py
royischoss: left overs from V2ModelServer chaged it

mlrun/serving/server.py
```diff
@@ -360,14 +370,127 @@ def add_system_steps_to_graph(graph: RootFlowStep):
...
+        # Connect each model runner to the monitoring step:
+        for step_name, step in monitored_steps.items():
+            if isinstance(monitor_flow_step.after, list):
+                monitor_flow_step.after.append(step_name)
+            elif isinstance(monitor_flow_step.after, str):
+                monitor_flow_step.after = [monitor_flow_step.after, step_name]
+            else:
+                raise GraphError(
+                    "Expected model runner step to be followed by error raiser step"
+                )
```
davesh0812: Not sure I understand the error msg here

mlrun/serving/server.py
```diff
@@ -360,14 +370,127 @@ def add_system_steps_to_graph(graph: RootFlowStep):
...
+    if track_models:
+        graph, monitor_flow_step = add_monitoring_general_steps(
+            project, graph, context, serving_spec
+        )
+        # Connect each model runner to the monitoring step:
+        for step_name, step in monitored_steps.items():
+            if isinstance(monitor_flow_step.after, list):
+                monitor_flow_step.after.append(step_name)
+            elif isinstance(monitor_flow_step.after, str):
+                monitor_flow_step.after = [monitor_flow_step.after, step_name]
```
davesh0812:
```suggestion
            elif monitor_flow_step.after and isinstance(monitor_flow_step.after, str):
                monitor_flow_step.after = [monitor_flow_step.after, step_name]
            else:
                monitor_flow_step.after = [step_name]
```

mlrun/serving/server.py
```diff
@@ -360,14 +370,127 @@ def add_system_steps_to_graph(graph: RootFlowStep):
     return graph
 
 
+def add_monitoring_general_steps(
```
davesh0812, line 382: Please add to the description a graph plot only of the sys steps

mlrun/serving/states.py
```diff
@@ -452,11 +454,52 @@ def set_flow(
...
+                            "Cannot add router step containing the same model endpoint name as "
+                            "ModelRunnerStep model"
+                        )
+                    models.append(route.name)
+            else:
+                if step.name in root.model_endpoints_names:
+                    raise GraphError(
+                        "Cannot add router step or task step containing the same model endpoint name as "
+                        "ModelRunnerStep model"
+                    )
```
davesh0812: please edit :)

mlrun/serving/states.py
```diff
@@ -452,11 +454,52 @@ def set_flow(
...
+        """
+        if (
+            isinstance(step, TaskStep)
+            and not isinstance(step, ModelRunnerStep)
+            and model_endpoint_creation_strategy
+            != schemas.ModelEndpointCreationStrategy.SKIP
+        ):
+            root = self
+            while root.parent is not None:
+                root = root.parent
```
davesh0812: please create private method for this, you using also in  `verify_model_runner_step`

mlrun/serving/states.py
```diff
@@ -1906,6 +2069,17 @@ def model_endpoints_names(self, models: list[str]):
     def extend_model_endpoints_names(self, model_endpoints_names: list):
         self._models.extend(model_endpoints_names)
 
+    @property
+    def model_endpoints_routes_names(self) -> list[str]:
```
davesh0812, line 2083: Why we need to specify this and not to use `model_endpoints_names` for all the kind of endpoint?

mlrun/serving/system_steps.py
davesh0812: @royischoss still relevant

davesh0812 (commented):
Look Good,
I mange to go over only partially.
Still need to look on the system steps and the tests.

mlrun/serving/system_steps.py
royischoss: We hint it as float in every code assurances, I also make cast before providing to the step, you want me to give it Optional ?

mlrun/serving/server.py
royischoss: Me too :)

mlrun/config.py
```diff
@@ -233,6 +233,7 @@
                 "delete_function": "900",
                 "model_endpoint_creation": "600",
                 "model_endpoint_tsdb_leftovers": "900",
+                "model_endpoint_creation_check": "15",
```
theSaarco: Not the right place for this. This block of configurations relates to bg-tasks timeouts and is used on the server-side. This period relates to the serving steps on the client side, and should probably be somewhere under the `model_endpoint_monitoring` section of the config, and should be called something like `model_endpoint_creation_check_period`

mlrun/runtimes/nuclio/serving.py
```diff
@@ -769,6 +766,18 @@ def to_mock_server(
             monitoring_mock=self.spec.track_models,
         )
 
+        if (
+            isinstance(self.spec.graph, RootFlowStep)
+            and self.spec.graph.include_model_runner
```
theSaarco: Why add this as a member of the graph? Can't you just calculate it on-the-fly when deploying the monitoring steps? If you use a member variable then you need to make sure you update it whenever someone adds or removes steps from the graph, even if doing through some hacky ways (such as replacing the step in the graph list of steps).
Also, since you changed the decision on monitored steps to be more generic and not rely specifically on MRS steps, calling this `include_model_runner` is a very bad name - it should look for generally monitored steps. But, as noted above, I don't see a lot of value in actually storing this in the graph as a member.

mlrun/serving/server.py
```diff
@@ -360,14 +370,127 @@ def add_system_steps_to_graph(graph: RootFlowStep):
...
+
+
+def add_system_steps_to_graph(
+    project: str, graph: RootFlowStep, track_models: bool, context, serving_spec
+) -> RootFlowStep:
+    monitored_steps = {
+        step.name: step
+        for step in graph.steps.values()
+        if isinstance(step, mlrun.serving.MonitoredStep)
+    }
```
theSaarco: Basically this is exactly the check that you need to replace the `include_model_runner` member - just put this in a function and call it prior to adding the system steps...

mlrun/serving/states.py
```diff
@@ -1057,16 +1111,55 @@ class ModelRunner(storey.ParallelExecution):
...
+
+        return event
 
     def select_runnables(self, event):
         models = cast(list[Model], self.runnables)
         return self.model_selector.select(event, models)
 
 
-class ModelRunnerStep(TaskStep, StepToDict):
+class MonitoredStep(TaskStep, StepToDict):
```
theSaarco: I would make this abstract class so users cannot use it directly.

mlrun/serving/states.py
```diff
@@ -1057,16 +1111,55 @@ class ModelRunner(storey.ParallelExecution):
...
+        super().__init__(*args, name=name, **kwargs)
+        self.raise_exception = raise_exception
+        self._monitoring_data = None
+
+    def load_monitoring_data(self):
+        """
+        Child class must override load_monitoring_data() method and provide meaningful data-structure to the
+        pre-process step in monitoring flow.
+        """
+        self._monitoring_data = {}
```
theSaarco: So, why not returning `NotImplemented` if you must override?

mlrun/serving/states.py
```diff
@@ -1057,16 +1111,55 @@ class ModelRunner(storey.ParallelExecution):
...
+class MonitoredStep(TaskStep, StepToDict):
+    kind = "monitored"
+    _dict_fields = TaskStep._dict_fields + ["raise_exception"]
+
+    def __init__(self, *args, name: str, raise_exception=True, **kwargs):
+        super().__init__(*args, name=name, **kwargs)
+        self.raise_exception = raise_exception
+        self._monitoring_data = None
+
+    def load_monitoring_data(self):
```
theSaarco: Load from where? Maybe change name to `calculate_monitoring_data` or something.
Also, IIUC, this is mostly an internal function called by the `monitoring_data` property, so prefix its name with `_`.

mlrun/serving/states.py
```diff
@@ -1057,16 +1111,55 @@ class ModelRunner(storey.ParallelExecution):
...
+
+    @property
+    def monitoring_data(self) -> dict:
+        if self._monitoring_data is None:
+            self.load_monitoring_data()
+        return self._monitoring_data
+
+    @monitoring_data.setter
+    def monitoring_data(self, data: dict) -> None:
+        self._monitoring_data = data
```
theSaarco: I didn't understand at first why we need a setter for this property (I thought it's only calculated by the `load_monitoring_data` method), but then I saw that the MRS actually modified this property directly, and therefore probably needs this.
The question is whether we expect users to use this setter directly, or do we expect them to only subclass the `load_monitoring_data` - we document that they need to do the latter, but our own code does that + something else. So, it may be very confusing to users wanting to implement extensions.

mlrun/serving/states.py
```diff
@@ -1057,16 +1111,55 @@ class ModelRunner(storey.ParallelExecution):
...
+        """
+        Child class must override load_monitoring_data() method and provide meaningful data-structure to the
+        pre-process step in monitoring flow.
+        """
+        self._monitoring_data = {}
+
+    @property
+    def monitoring_data(self) -> dict:
+        if self._monitoring_data is None:
+            self.load_monitoring_data()
```
theSaarco: What happens if `load_monitoring_data` was already called (for example, someone asked for the `monitoring_data` property, and then the class instance was changed? For example, someone added a model to the MRS? While caching this information is great for performance, you need to make sure the cached data is invalidated on any changes that affect the result.
Alternatively, calculate it on-demand (depending on how many times this is expected to be called and how complicated it is to do this calculation).

mlrun/serving/states.py
```diff
@@ -1057,16 +1111,55 @@ class ModelRunner(storey.ParallelExecution):
...
+
+    def __init__(self, *args, name: str, raise_exception=True, **kwargs):
+        super().__init__(*args, name=name, **kwargs)
+        self.raise_exception = raise_exception
+        self._monitoring_data = None
+
+    def load_monitoring_data(self):
+        """
+        Child class must override load_monitoring_data() method and provide meaningful data-structure to the
+        pre-process step in monitoring flow.
```
theSaarco: If we expect users to ever subclass this, this documentation must be much more elaborate explaining exactly what the structure of the returned data must be, what fields etc.
Also, if the expectation is that this method will update the `monitoring_data` property (as the MRS does), then this needs to be stated. I was actually expecting this to return the data to the caller.

theSaarco (changes requested):
I think the implementation of the monitoring data retrieval is confusing and will be hard for users to subclass.

mlrun/serving/states.py
royischoss: I will remove it, at first I kept it in _dict_fields and a setter is mandatory

mlrun/serving/states.py
royischoss: We need it cause I am not checking for duplication in routes and I must differ between route and model runner in the addition of a model.

mlrun/config.py
royischoss: Fixed as mentioned

mlrun/runtimes/nuclio/serving.py
royischoss: Removed added two methods one collecting the MonitoredSteps and one (for avoiding repeatedly) running on all steps in graph returns for the first monitoredStep

mlrun/serving/states.py
royischoss: Made the change as mentioned

mlrun/serving/states.py
royischoss: made the change as mentioned

mlrun/serving/states.py
royischoss: Made the change as mentioned

mlrun/serving/states.py
royischoss: Now it is a param for the method and we also works on copy to prevent changing user code and class_args

mlrun/serving/states.py
royischoss: changed it with small caching inside class_args for outputs if there is an override we will calculate again for the same model name. because add_model should be called with new params and we override the value in the class args dict

mlrun/serving/states.py
royischoss: added more elaborate description, I will seat with you and Jill and add documentation for this usages because there are some implementation details we must provide for the new serving

mlrun/serving/server.py
```diff
@@ -335,34 +343,147 @@ def wait_for_completion(self):
         return self.graph.wait_for_completion()
 
 
-def add_system_steps_to_graph(graph: RootFlowStep):
+def add_error_raiser_step(
+    graph: RootFlowStep, model_runner_steps: dict[str, MonitoredStep]
```
theSaarco: Instead of `model_runner_steps`, should rename to `monitored_steps` since we generalized it.

mlrun/serving/system_steps.py
```diff
@@ -0,0 +1,391 @@
...
+            background_task.status.state
+            in mlrun.common.schemas.BackgroundTaskState.terminal_states()
+        ):
+            logger.info(
+                f"Model endpoint creation task completed with state {background_task.status.state}"
+            )
+        else:  # in progress
+            logger.info(
+                f"Model endpoint creation task is still in progress with the current state: "
+                f"{background_task.status.state}. Events will not be monitored for the next 15 seconds",
```
theSaarco:
```suggestion
                f"{background_task.status.state}. Events will not be monitored for the next {mlrun.mlconf.model_endpoint_monitoring.model_endpoint_creation_check_period} seconds",
```

mlrun/serving/system_steps.py
```diff
@@ -0,0 +1,391 @@
...
+                return None
+        elif (
+            self._background_task_status
+            == mlrun.common.schemas.BackgroundTaskState.failed
+        ):
+            return None
+        return event
+
+    def _background_task_succeeded(
+        self, background_task: mlrun.common.schemas.BackgroundTask
```
theSaarco: Nitpicking: you don't need all the `BackgroundTask` object here, just its state.
And, TBH, this method mainly just logs stuff, it could functionally have been replaced with just doing the last `return` statement.

tests/system/model_monitoring/test_app.py
```diff
@@ -519,20 +520,46 @@ def _add_error_alert(self) -> None:
...
+                input_path="inputs",
+                result_path="outputs",
+            )
+            graph = serving_fn.set_topology("flow", engine="async")
+            graph.to(model_runner_step).respond()
+        else:
+            serving_fn = typing.cast(
+                mlrun.runtimes.nuclio.serving.ServingRuntime,
+                mlrun.import_function(
+                    "hub://v2_model_server",
```
theSaarco, line 554: This code is rather outdated, and should be replaced with simply using the `V2ModelServer` class (and in the future the `Model` class, of course). Don't expect this to be changed as part of this PR, but we should get rid of using this code.

mlrun/serving/states.py
```diff
@@ -1057,16 +1103,61 @@ class ModelRunner(storey.ParallelExecution):
...
+            "inputs" - inputs features
+            "outputs" - output schema expected
+            "input_path" - the path where inputs are
+            "result_path" - the path where results are
+            "creation_strategy" - model endpoint creation strategy
+            "labels" - model endpoint labels
+            "model_endpoint_uid" - model endpoint uid (added in deployment)
+            "model_class" - the model class
+        }
+        """
```
theSaarco: The HTML documentation does not include docstrings from internal methods, and they need to be manually marked for inclusion, otherwise they won't appear in the MLRun docs. 
See for example here: https://github.com/mlrun/mlrun/blob/c0d9796be7ec6c611eadf6c4eab969dd1adf3d79/docs/api/mlrun.feature_store/index.rst?plain=1#L16

theSaarco (approved):
Looks good. I have some small comments (that I think should be addressed) but approving this PR for now. Either do relevant fixes before completing this PR or on the next one.

mlrun/serving/server.py
royischoss: added

mlrun/serving/system_steps.py
```diff
@@ -0,0 +1,391 @@
...
+                outputs = self.transpose_by_key(result, output_schema)
+            elif len(result) == 1:
+                outputs = (
+                    result[output_schema[0]]
+                    if output_schema
+                    else list(result.values())[0]
+                )
+            else:
+                outputs = []
+            if not output_schema:
```
assaf758, line 65: I'm not sure if we want to support not providing order…
But if we do, If I understand correctly, this log will be written on every event being sent to the stream pod and will spam the logs…

assaf758 (approved):
Very impressive and substantial work!
minor question/comment inside.

mlrun/serving/system_steps.py
royischoss: This log will apply only when user provides outputs as dict instance and not provided any output schema in the model artifact or in `ModelRunnerStep::add_model`, so in every other sequence that are the common one (list outputs, provided output/labels in model artifact) we won't write this log

PR #8031, opened 2025-06-03
[Deprecations] Initial deprecation of mlrun/ml-base image

Yacouby:
This PR deprecates the use of the `mlrun/ml-base` image and transitions to using `mlrun/mlrun` instead. 

The following changes were made:
- Added Dask-related packages to the `requirements.txt` of the `mlrun/mlrun` image.
- Replaced `mlrun/ml-base` with `mlrun/mlrun` for the dask_kfp_image in `config.py`.
- Added a deprecation warning to the client side when using ml-base image and to  the server when encountering the `mlrun/ml-base` image, and replaced its enrichment logic to use the `mlrun/mlrun` image instead.
- Updated system tests using ml-base to use mlrun image instead.

Note: Removing the image, its associated Dockerfile and update any relevant CI/CD rules will be done in a separate PR.

Jira:
https://iguazio.atlassian.net/browse/ML-10057


mlrun/utils/helpers.py
```diff
@@ -912,6 +912,21 @@ def enrich_image_url(
...
+            if mlrun.utils.helpers.validate_component_version_compatibility(
+                "mlrun-client", "1.10.0", mlrun_client_version=tag
+            ):
+                warnings.warn(
+                    "'mlrun/ml-base' image is deprecated in 1.10.0 and will be removed in 1.12.0, "
+                    "please use 'mlrun/mlrun' instead.",
+                    # TODO: Remove this in 1.12.0
+                    FutureWarning,
+                )
+                image_url = image_url.replace("mlrun/ml-base", "mlrun/mlrun")
```
TomerShor, line 926: So we are replacing the image under the hood here. Do we want to do it even when an ml-base image from an earlier version?

Yacouby: no, thats why i checked if the version is >= `1.10.0` then the replacement is done, otherwise it should stay as is.
Thats because ml-base image from earlier versions will not include all the dask packages for example.

mlrun/launcher/client.py
```diff
@@ -60,6 +61,15 @@ def prepare_image_for_deploy(runtime: "mlrun.runtimes.BaseRuntime"):
         ):
             image = mlrun.mlconf.function_defaults.image_by_kind.to_dict()[runtime.kind]
 
+        # Warn if user explicitly set the deprecated mlrun/ml-base image
+        if image and "mlrun/ml-base" in image:
+            warnings.warn(
+                "'mlrun/ml-base' image is deprecated in 1.10.0 and will be removed in 1.12.0, "
+                "please use 'mlrun/mlrun' instead.",
```
alonmr:
```suggestion
                "use 'mlrun/mlrun' instead.",
```

mlrun/runtimes/daskjob.py
```diff
@@ -192,7 +192,7 @@ def __init__(self, spec=None, metadata=None):
         super().__init__(spec, metadata)
         self._cluster = None
         self.use_remote = not mlrun.k8s_utils.is_running_inside_kubernetes_cluster()
-        self.spec.build.base_image = self.spec.build.base_image or "daskdev/dask:latest"
+        self.spec.build.base_image = self.spec.build.base_image or "mlrun/mlrun"
```
alonmr:
```suggestion
        self.spec.build.base_image = self.spec.build.base_image or mlconf.default_base_image
```

mlrun/utils/helpers.py
```diff
@@ -912,6 +912,21 @@ def enrich_image_url(
...
     is_mlrun_image = image_url.startswith("mlrun/") or "/mlrun/" in image_url
 
+    if is_mlrun_image and "mlrun/ml-base" in image_url:
+        if tag:
+            if mlrun.utils.helpers.validate_component_version_compatibility(
+                "mlrun-client", "1.10.0", mlrun_client_version=tag
+            ):
+                warnings.warn(
+                    "'mlrun/ml-base' image is deprecated in 1.10.0 and will be removed in 1.12.0, "
+                    "please use 'mlrun/mlrun' instead.",
```
alonmr:
```suggestion
                    "use 'mlrun/mlrun' instead.",
```

server/py/services/api/utils/builder.py
```diff
@@ -840,7 +840,6 @@ def is_mlrun_image(base_image):
     mlrun_images = [
         "mlrun/mlrun",
         "mlrun/mlrun-gpu",
-        "mlrun/ml-base",
```
alonmr: shouldnt remove this until the image is really removed

mlrun/launcher/client.py
alonmr: you also need to add that for clients >= 1.10 we replace it to mlrun/mlrun

alonmr (commented):
Nice job! A few comments

server/py/services/api/utils/builder.py
Yacouby: 👍  will add it to this PR : https://github.com/mlrun/mlrun/pull/8040

mlrun/utils/helpers.py
```diff
@@ -912,6 +912,21 @@ def enrich_image_url(
     # it's an mlrun image if the repository is mlrun
     is_mlrun_image = image_url.startswith("mlrun/") or "/mlrun/" in image_url
 
+    if is_mlrun_image and "mlrun/ml-base" in image_url:
+        if tag:
+            if mlrun.utils.helpers.validate_component_version_compatibility(
```
TomerShor, line 917: use `framework.utils.helpers.validate_client_version`

mlrun/utils/helpers.py
TomerShor: I understand 👍

TomerShor (changes requested):

mlrun/utils/helpers.py
Yacouby: MLRun server may import mlrun but not the way around

Yacouby: thats the error that i get from the linter once i try to call the method you suggested

TomerShor: OK 👍

liranbg (approved):

TomerShor (approved):

PR #8043, opened 2025-06-06
[Serving] Implement serving graph as a job

gtopper:
[ML-10065](https://iguazio.atlassian.net/browse/ML-10065)

This allows the user to define a serving function, customize its graph, turn it into a job using the `to_job()` method, and then run/deploy the job.

[ML-10065]: https://iguazio.atlassian.net/browse/ML-10065?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/runtimes/nuclio/serving.py
```diff
@@ -812,3 +821,31 @@ def _set_as_mock(self, enable):
...
+        """
+        Convert this ServingRuntime to a KubejobRuntime, so that the graph can be run as a standalone job.
+
+        :param target_mapping: Set this to map targets in this ServingRuntime to other targets that will run as
+          part of the standalone job. For example, you may wish to replace a ParquetTarget in order to change its path.
+        """
+        serving_job = self.copy()
+        if target_mapping:
+            for from_name, (to_name, to_step) in target_mapping.items():
+                to_step = to_step.copy()
```
theSaarco: Would be better to verify that `to_step` is indeed a target step.

mlrun/runtimes/nuclio/serving.py
```diff
@@ -812,3 +821,31 @@ def _set_as_mock(self, enable):
...
+    ) -> KubejobRuntime:
+        """
+        Convert this ServingRuntime to a KubejobRuntime, so that the graph can be run as a standalone job.
+
+        :param target_mapping: Set this to map targets in this ServingRuntime to other targets that will run as
+          part of the standalone job. For example, you may wish to replace a ParquetTarget in order to change its path.
+        """
+        serving_job = self.copy()
+        if target_mapping:
+            for from_name, (to_name, to_step) in target_mapping.items():
```
theSaarco: So you're assuming that `target_mapping` is a dict `str -> tuple(name, step)`? This doesn't match the definition of the object in the function signature. Also, not sure why you need the `to_name` in this case? The thinking was that the step provided will have the same target name as the one it replaces.

mlrun/runtimes/nuclio/serving.py
```diff
@@ -812,3 +821,31 @@ def _set_as_mock(self, enable):
...
+                to_step = to_step.copy()
+                to_step.after = serving_job.spec.graph.steps[from_name].after
+                to_step.responder = serving_job.spec.graph.steps[from_name].responder
+                del serving_job.spec.graph.steps[from_name]
+                serving_job.spec.graph.steps[to_name] = to_step
+
+        spec = KubeResourceSpec(
+            serving_spec=serving_job._get_serving_spec(),
+            default_handler="mlrun.serving.server.execute_graph",
+        )
```
theSaarco: Are there no other parameters that need copying in this case? I would think most of these parameters need copying as well (resources, mounts etc.).

mlrun/serving/server.py
```diff
@@ -405,6 +411,80 @@ def v2_serving_init(context, namespace=None):
...
+            await run(data)
+
+    termination_result = server.wait_for_completion()
+    if asyncio.iscoroutine(termination_result):
+        termination_result = await termination_result
+
+    return responses, termination_result
+
+
+def execute_graph(
```
theSaarco, line 482: Should add documentation here, since this is the handler that will be called, so the parameters need to be documented.

mlrun/runtimes/nuclio/serving.py
```diff
@@ -709,6 +713,7 @@ def _get_serving_spec(self):
             "track_models": self.spec.track_models,
             "default_content_type": self.spec.default_content_type,
             "model_endpoint_creation_task_name": self.spec.model_endpoint_creation_task_name,
+            "filename": getattr(self.spec, "filename", None),
```
theSaarco, line 710: Why is this needed? I see in this PR that it's used in tests to run the job locally with custom classes, and it can work with `code_to_function` but how will this work with `project.set_function()` (which is what users are expected to be using now)?

mlrun/serving/server.py
```diff
@@ -405,6 +411,80 @@ def v2_serving_init(context, namespace=None):
...
+        batch_size = len(df)
+
+    batch = []
+    for index, row in df.iterrows():
+        data = row.to_dict()
+        if batching:
+            batch.append(data)
+            if len(batch) == batch_size:
+                await run(batch)
+                batch = []
```
theSaarco, line 468: You don't handle the case of a last batch that's smaller than `batch_size`. Should add an additional `run(batch)` after the loop if `batch` is not empty.

server/py/services/api/tests/unit/runtimes/test_serving.py
```diff
@@ -165,23 +160,6 @@ def _assert_deploy_spec_has_secrets_config(
                     },
                 },
             ]
-            if use_config_map:
```
theSaarco, line 168: Why was this removed? Has this changed due to this PR?

tests/launcher/test_local.py
```diff
@@ -187,3 +191,75 @@ def func_b():
...
+    )
+    graph = function.set_topology("flow", engine="async")
+
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        graph.to(name="increaser", class_name="SepalLengthIncreaser")
+        graph.to(name="parquet", class_name="storey.ParquetTarget", path=tmp_dir)
+
+        target_mapping = None
+        if with_target_mapping:
+            new_target = params_to_step(
```
theSaarco: This is probably not how users will actually do this - I would expect them to use the MLRun classes, such as `mlrun.datastore.ParquetTarget` directly, rather than use this internal `params_to_step` function.

theSaarco (changes requested):
Looks very good, but some changes are still needed, and I have some questions about some aspects of the implementation.

tests/launcher/test_local.py
gtopper: Then they'll need to pass this as a string, same as we do everywhere else (i.e. in `step.to()`). That is to say, we can call `params_to_step` internally for the user.

server/py/services/api/tests/unit/runtimes/test_serving.py
gtopper: The config map is now added in the server side launcher, and therefore doesn't show up in the result for the purpose of this test. Do you have a suggestion how these tests might be modified to work with the later addition of the config map? I tried, but couldn't figure out a way in reasonable time.

mlrun/runtimes/nuclio/serving.py
gtopper: In serving, we don't actually have a distinction between target and non-target steps. That only exists in feature store.

mlrun/runtimes/nuclio/serving.py
gtopper: Removed `to_name`.

> The thinking was that the step provided will have the same target name as the one it replaces.

Regarding this one, should we enforce that? I.e. fail if the step's name (as defined in step.name) differs from the original name?

This could prevent the pitfall of breaking steps that come after the target, which storey supports in principle, but is not a typical use case and is not even very well-defined in terms of what events a target step might output and with what guarantees.

mlrun/serving/server.py
gtopper: Fixed and added a test.

mlrun/runtimes/nuclio/serving.py
gtopper: Added all the relevant ones (please check).

mlrun/runtimes/nuclio/serving.py
theSaarco: I think we should enforce. Failing early at this point will save trouble later. So, verify that the step sent as target has the same name as the one being replaced.

server/py/services/api/tests/unit/runtimes/test_serving.py
theSaarco: This is a unit test, so everything here is a bit complex, mocking flows in various ways. If the functionality is now in the launcher then I can't really point at an easy fix to make sure we go through this flow.

mlrun/runtimes/nuclio/serving.py
gtopper: Done.

mlrun/serving/server.py
gtopper: Done.

mlrun/serving/server.py
```diff
@@ -405,6 +411,94 @@ def v2_serving_init(context, namespace=None):
     _set_callbacks(server, context)
 
 
+async def async_execute_graph(
+    context: MLClientCtx,
+    data: DataItem,
+    batching: bool,
+    batch_size: Optional[int],
+) -> (list[Any], Any):
```
gtopper: I'm thinking we should probably remove the termination result, because it's never been documented or tested via mlrun, and managing termination result over a split graph is also something I'm not sure we can currently due from mlrun.

Basically, it's always been a feature of storey that's used only for testing.

theSaarco: In this case there's no consideration of split graph (for now), so technically this challenge doesn't exist here.
As for the actual feature and whether we should expose it - unless we can tell for sure that nobody uses it, it's probably better to keep this here. In most cases it would simply be `None` IIUC.

tests/launcher/test_local.py
theSaarco: This really goes back to the question of how target steps are represented in the graph today. The users can create class instances of those targets, or we can indeed call `params_to_step` for them. My concern is mostly that `params_to_step` is not widely documented and is mostly an internal function, so it's better if we do this for the users like we do in other places.

mlrun/serving/server.py
gtopper: By split graph, I mean nonlinear, not multiple functions. And yes, we can tell for sure that no one uses it because it's never been exposed.

theSaarco: ok, so let's discuss this. Returning the termination result was something that was brought up in the design of this feature as something that should be exposed. If there's nobody using it at all and it's not in use in MLRun (vs. directly with storey) then I agree this can be removed.

gtopper: Removed termination result.

tests/launcher/test_local.py
gtopper: We discussed this with @assaf758 and @royischoss and decided to leave target_mapping out of this requirement for the time being. Removed from PR.

mlrun/runtimes/nuclio/serving.py
gtopper: Will follow up on it in [another story](https://iguazio.atlassian.net/browse/ML-10214) / PR, as discussed today.

theSaarco (approved):
LGTM.

PR #8051, opened 2025-06-08
[Artifacts] Allow partial filtering over parent name and parent tag

davesh0812:
Implements  - https://iguazio.atlassian.net/browse/ML-10123


moranbental (approved):
LGTM

PR #8059, opened 2025-06-08
[API/Kubejob] Retry jobs controller [feature/retry-job]

alonmr:
## Periodic controller for jobs retries

* Update run state to "pendingRetry" on client execution
* Fallback to API monitoring for run updates
* Periodic submit retry job
* Use cache on run uids to ensure retry exactly once

https://iguazio.atlassian.net/browse/ML-10113


mlrun/common/runtimes/constants.py
```diff
@@ -185,12 +188,19 @@ def error_and_abortion_states():
...
+            | set(RunStates.error_and_abortion_states())
+        )
+
     @staticmethod
     def not_allowed_for_deletion_states():
         return [
             RunStates.running,
             RunStates.pending,
-            # TODO: add aborting state once we have it
+            # TODO: add aborting?
```
liranbg: I think can be removed

mlrun/config.py
```diff
@@ -118,6 +118,12 @@
             # max number of parallel abort run jobs in runs monitoring
             "concurrent_abort_stale_runs_workers": 10,
             "list_runs_time_period_in_days": 7,  # days
+            "retry": {
+                # periodic job for triggering retries interval in seconds
+                "interval": "30",
+                # runs limit to fetch for retrying
+                "runs_limit": 1000,
```
liranbg:
```suggestion
                "fetch_runs_limit": 1000,
```

mlrun/execution.py
```diff
@@ -450,10 +454,11 @@ def from_dict(
         if start:
             start = parser.parse(start) if isinstance(start, str) else start
             self._start_time = start
-        self._state = "running"
+        self._state = mlrun.common.runtimes.constants.RunStates.running
```
liranbg, line 457: can we distinguish between the state updates on local vs by server? I wonder because this functionality perhaps should move to server side? to launcher? this function remains here means it may run on both local and remote and I am not sure the retry is relevant at all to local runs

mlrun/execution.py
```diff
@@ -1147,9 +1155,15 @@ def set_state(
         updates = {"status.last_update": now_date().isoformat()}
 
         if error is not None:
-            self._state = "error"
+            new_state = mlrun.common.runtimes.constants.RunStates.error
```
liranbg: final_state / desired_state / override_state

mlrun/launcher/base.py
```diff
@@ -359,7 +360,10 @@ def _enrich_run(
             | state_thresholds
         )
         run.spec.state_thresholds = state_thresholds or run.spec.state_thresholds
-        run.spec.retry = retry
+        run.spec.retry = retry or run.spec.retry
+        if run.status.state == mlrun.common.runtimes.constants.RunStates.pending_retry:
+            run.status.state = mlrun.common.runtimes.constants.RunStates.running
+            run.status.retry_count = (run.status.retry_count or 0) + 1
```
liranbg:
```suggestion
            run.status.retry_count = run.status.retry_count or 0 # in case it is none
            run.status.retry_count += 1 # increment by one
```

server/py/services/api/main.py
```diff
@@ -193,6 +196,7 @@ async def _start_periodic_functions(self):
             self._start_periodic_project_summaries_calculation()
         self._start_periodic_partition_management()
         self._start_periodic_refresh_smtp_configuration()
+        self._start_periodic_retry_jobs()
```
liranbg: wrap in feature gate

server/py/services/api/main.py
```diff
@@ -911,6 +926,124 @@ async def _stop_logs_for_runs(self, runs: list, chunk_size: int = 10):
...
+        try:
+            runs = await fastapi.concurrency.run_in_threadpool(
+                get_db().list_runs,
+                db_session,
+                project="*",
+                states=[mlrun.common.runtimes.constants.RunStates.pending_retry],
+                limit=mlconf.monitoring.runs.retry.runs_limit,
+            )
+            if not runs:
+                self._logger.debug("No runs to retry")
```
liranbg: too verbose.

server/py/services/api/main.py
```diff
@@ -911,6 +926,124 @@ async def _stop_logs_for_runs(self, runs: list, chunk_size: int = 10):
...
+        ) * (run.status.retry_count + 1)
+        delta = (
+            datetime.datetime.fromisoformat(run.status.end_time)
+            + datetime.timedelta(seconds=delay)
+            - datetime.datetime.now(datetime.timezone.utc)
+        )
+        call_after_seconds = max(delta.total_seconds(), 0)
+
+        # Submit the job with the calculated delay
+        loop.call_later(
```
liranbg, line 1038: log before submitting

liranbg (changes requested):
overall looking very neat. minor comments for conventions / code-locality improvement

mlrun/runtimes/base.py
```diff
@@ -600,12 +601,28 @@ def _update_run_state(
...
+        ):
+            new_state = mlrun.common.runtimes.constants.RunStates.error
+            status_text = None
+            max_retries = get_in(resp, "spec.retry.count", 0)
+            retry_count = get_in(resp, "status.retry_count", 0) or 0
+            if max_retries and retry_count < max_retries:
+                new_state = mlrun.common.runtimes.constants.RunStates.pending_retry
+                status_text = f"Run failed attempt {retry_count + 1} of {max_retries}"
+            elif max_retries and retry_count >= max_retries:
+                status_text = f"Run failed after {retry_count} attempts"
```
rokatyy: minor and feel free to ignore, but seems to be a bit more human-readable this way
```suggestion
            if max_retries:
              if retry_count < max_retries:
                  new_state = run_states.pending_retry
                  status_text = f"Run failed attempt {retry_count + 1} of {max_retries}"
              else:
                  status_text = f"Run failed after {retry_count} attempts"
```

server/py/services/api/main.py
```diff
@@ -911,6 +926,124 @@ async def _stop_logs_for_runs(self, runs: list, chunk_size: int = 10):
...
+                run = mlrun.RunObject.from_dict(run_dict)
+                if run.metadata.uid in self._retry_in_progress_run_uids:
+                    self._logger.debug(
+                        "Run is already being retried, skipping",
+                        run_uid=run.metadata.uid,
+                    )
+                    continue
+
+                    # retry_count may be None on the first attempt
+                run.status.retry_count = run.status.retry_count or 0
```
rokatyy:
```suggestion
                # retry_count may be None on the first attempt
                run.status.retry_count = run.status.retry_count or 0
```

server/py/services/api/main.py
```diff
@@ -911,6 +926,124 @@ async def _stop_logs_for_runs(self, runs: list, chunk_size: int = 10):
...
+                exceptions = await asyncio.gather(*futures, return_exceptions=True)
+                for exception in exceptions:
+                    if isinstance(exception, Exception):
+                        self._logger.warning(
+                            "Failed task in retry job",
+                            exc=err_to_str(exception),
+                        )
+
+        finally:
+            await fastapi.concurrency.run_in_threadpool(close_session, db_session)
```
rokatyy: what about exception handling before `finally`? maybe now exceptions are not likely to happen, but later with changing of this method, adding a small exception log might be very helpful for debug purposes

server/py/services/api/main.py
```diff
@@ -911,6 +926,124 @@ async def _stop_logs_for_runs(self, runs: list, chunk_size: int = 10):
...
+
+    def _submit_run_for_retry(self, run: mlrun.RunObject):
+        self._retry_in_progress_run_uids.add(run.metadata.uid)
+        loop = asyncio.get_event_loop()
```
rokatyy: from python 3.10 [asyncio](https://docs.python.org/3/library/asyncio-eventloop.html) recommends using `get_running_loop()` instead of `get_event_loop()`

server/py/services/api/main.py
```diff
@@ -911,6 +926,124 @@ async def _stop_logs_for_runs(self, runs: list, chunk_size: int = 10):
...
+
+        # Calculate the delay based on the retry policy
+        delay = framework.utils.helpers.time_string_to_seconds(
+            run.spec.retry.backoff.base_delay,
+            mlrun.mlconf.function.spec.retry.backoff.min_base_delay,
+        ) * (run.status.retry_count + 1)
+        delta = (
+            datetime.datetime.fromisoformat(run.status.end_time)
```
rokatyy, line 1024: Does `run.status.end_time` always include timezone information? If not, `fromisoformat` gives datetime without ts, and then when calculating delta code may crash:

image

server/py/services/api/runtime_handlers/base.py
```diff
@@ -1751,6 +1751,15 @@ def _ensure_run_state(
             elif run_state == RunStates.error:
                 # Try resolving the error reason
                 reason, message = self._resolve_container_error_status(runtime_resource)
+                # Should run be retried
+                retry_spec = run.get("spec", {}).get("retry")
+                max_retries = retry_spec.get("count") if retry_spec else -1
+                retry_count = run.get("status", {}).get("retry_count", 0) or 0
```
rokatyy, line 1758:
```suggestion
                retry_count = run.get("status", {}).get("retry_count", 0)
```

rokatyy (changes requested):
Some suggestions

mlrun/execution.py
alonmr: Retry spec is validated and allowed only for kubejobs in the launcher. It is safe to have it here. Although this is by bad design, we should refactor it someday but not relevant for current changes.

mlrun/execution.py
alonmr: changed it to `state`

server/py/services/api/runtime_handlers/base.py
alonmr: if `retry_count` is `None` in the dict `run.get("status", {}).get("retry_count", 0)` will return `None`

rokatyy: @alonmr make sense. Then I suggest adding a comment with this information to avoid the same confusion in the future

server/py/services/api/runtime_handlers/base.py
```diff
@@ -1751,6 +1751,15 @@ def _ensure_run_state(
...
+                if retry_count < max_retries:
+                    run_state = RunStates.pending_retry
```
yaelgen, line 1760: It might help to log the transition to pendingRetry explicitly here (with run UID and retry count)

alonmr:
@CodiumAI-Agent /review

server/py/services/api/runtime_handlers/base.py
alonmr: There is a log a few lines below :)

server/py/services/api/main.py
```diff
@@ -911,6 +927,136 @@ async def _stop_logs_for_runs(self, runs: list, chunk_size: int = 10):
...
+        """
+        self._logger.debug("Retrying jobs with retry policy configured")
+        db_session = await fastapi.concurrency.run_in_threadpool(create_session)
+        try:
+            runs = await fastapi.concurrency.run_in_threadpool(
+                get_db().list_runs,
+                db_session,
+                project="*",
+                states=[mlrun.common.runtimes.constants.RunStates.pending_retry],
+                limit=mlconf.monitoring.runs.retry.fetch_runs_limit,
```
liranbg, line 943: just to prevent potential starvation - since you fetch only subset of the potential pending retry runs - perhaps you can order by update and take the earliest - (FIFO)

server/py/services/api/main.py
```diff
@@ -911,6 +927,136 @@ async def _stop_logs_for_runs(self, runs: list, chunk_size: int = 10):
...
+                # retry_count may be None on the first attempt
+                run.status.retry_count = run.status.retry_count or 0
+                # sanity
+                if not run.status.retry_count < run.spec.retry.count:
+                    self._logger.debug(
+                        "Run has reached max retry count, skipping",
+                        run_uid=run.metadata.uid,
+                        retry_count=run.status.retry_count,
+                        max_retry_count=run.spec.retry.count,
+                    )
```
liranbg:
```suggestion
                    self._logger.warn(
                        "Run has reached max retry count, skipping",
                        run_uid=run.metadata.uid,
                        retry_count=run.status.retry_count,
                        max_retry_count=run.spec.retry.count,
                    )
```

server/py/services/api/main.py
```diff
@@ -911,6 +927,136 @@ async def _stop_logs_for_runs(self, runs: list, chunk_size: int = 10):
...
+
+                try:
+                    self._submit_run_for_retry(run)
+                except Exception as exc:
+                    self._logger.warning(
+                        "Failed retrying run",
+                        run_uid=run.metadata.uid,
+                        exc=err_to_str(exc),
+                        traceback=traceback.format_exc(),
+                    )
```
liranbg, line 994: so my suggestion above says FIFO - question if - if you forever fail to retry run - what would be its verdict? maybe we should ahve some guardrails such as
if I try to retry a run from "x hours ago / days ago" perhaps that run is stale and should move to "failed" permanent  ?

server/py/services/api/main.py
liranbg: end_time should be set by api only. I suggest assuming utc if not set.

liranbg (commented):
good stuff! some suggestions and fixes below

server/py/services/api/main.py
alonmr: I'll add it as user story

server/py/services/api/main.py
alonmr: I'm not sure if sorting will ensure no starvation because the latest updated could have a long retry delay while the N+1 run needs to be triggered now.
So I think we should use pagination

mlrun/__main__.py
```diff
@@ -252,7 +252,13 @@ def run(
     config = environ.get("MLRUN_EXEC_CONFIG")
     if from_env and config:
         config = json.loads(config)
-        runobj = RunTemplate.from_dict(config)
+        # If run is a retry we need to maintain the run status therefore using RunObject instead of RunTemplate
+        retry_count = config.get("status", {}).get("retry_count")
+        if retry_count:
+            print(f"Retrying run - attempt: {retry_count + 1}")
```
liranbg: print -> log

PR #8063, opened 2025-06-08
[Model Monitoring] Connect evaluate to the writer

jond01:
Implements [ML-10152](https://iguazio.atlassian.net/browse/ML-10152).

This PR connects the model monitoring application `evaluate` (and the underlying job) to the relevant databases.

When the `write_output` argument is set to `True`, the job sends the results to the writer.
The stream profile and secrets are fetched from the environment variables and k8s secrets in a remote job, or can be passed with the `stream_profile` parameter when running locally.

Note: the credentials and infrastructure resources have to be set and enabled for this to work.

Upcoming PRs will improve the support with regard to more sophisticated aspects and error handling.

The tests and the extended system test cover the new functionality.

[ML-10152]: https://iguazio.atlassian.net/browse/ML-10152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/model_monitoring/applications/_application_steps.py
+++ b/mlrun/model_monitoring/applications/_application_steps.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import collections
-import json
 import traceback
 from collections import OrderedDict
 from datetime import datetime
@@ -23,10 +22,12 @@ import mlrun.common.schemas
 import mlrun.common.schemas.alert as alert_objects
 import mlrun.common.schemas.model_monitoring.constants as mm_constants
 import mlrun.model_monitoring.helpers
+import mlrun.platforms.iguazio
 from mlrun.serving import GraphContext
 from mlrun.serving.utils import StepToDict
 from mlrun.utils import logger
 
+from .base import _serialize_context_and_result
 from .context import MonitoringApplicationContext
 from .results import (
     ModelMonitoringApplicationMetric,
@@ -45,7 +46,7 @@ class _PushToMonitoringWriter(StepToDict):
         :param project: Project name.
         """
         self.project = project
-        self.output_stream = None
+        self._output_stream = None
 
     def do(
         self,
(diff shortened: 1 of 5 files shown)


mlrun/model_monitoring/applications/_application_steps.py
```diff
@@ -65,48 +66,31 @@ def do(
...
-            self.output_stream.push(
-                [writer_event], partition_key=application_context.endpoint_id
-            )
-            logger.debug("Pushed data to output stream successfully")
 
-    def _lazy_init(self):
-        if self.output_stream is None:
-            self.output_stream = mlrun.model_monitoring.helpers.get_output_stream(
+        writer_events = [
+            _serialize_context_and_result(context=application_context, result=result)
```
assaf758, line 72: Very nice!
I didn't realize we have this optimization opportunity, I believe this batching will improve the writer throughput considerably (together with storey target)

mlrun/model_monitoring/applications/base.py
```diff
@@ -470,6 +582,11 @@ def evaluate(
                                   ..., (\\operatorname{start} +
                                   m\\cdot\\operatorname{base\\_period}, \\operatorname{end}]`,
                                   where :math:`m` is some positive integer.
+        :param write_output:      Whether to write the results and metrics to the time-series DB. Can be ``True`` only
+                                  if ``endpoints`` are passed.
+                                  Note: the model monitoring infrastructure must be up for the writing to work.
+        :param stream_profile:    The stream profile, relevant only when running locally (``run_local=True``) and
+                                  ``write_output`` is ``True``.
```
assaf758:
```suggestion
        :param stream_profile:    The stream datasotre-profile, should be provided only if running locally and writing the metrics to time-series DB (``run_local=True`` and ``write_output=True`` respectevly). For reference on profile to use, check how model-monitoring infra was enabled (enable_model_monitoring)
```

mlrun/model_monitoring/applications/base.py
```diff
@@ -507,10 +624,25 @@ def evaluate(
...
+                if stream_profile:
+                    if not run_local:
+                        raise mlrun.errors.MLRunValueError(
+                            "Passing a `stream_profile` is relevant only when running locally"
+                        )
+                    if not write_output:
+                        raise mlrun.errors.MLRunValueError(
+                            "Passing a `stream_profile` is relevant only when writing the outputs"
+                        )
+                params["stream_profile"] = stream_profile
```
assaf758, line 640: I think you also need to check the case where stream_profile was not provided but run_local == True ?

mlrun/model_monitoring/applications/base.py
```diff
@@ -127,53 +211,78 @@ def _handler(
...
+        stream_profile: Optional[ds_profile.DatastoreProfile] = None,
     ):
         """
         A custom handler that wraps the application's logic implemented in
         :py:meth:`~mlrun.model_monitoring.applications.ModelMonitoringApplicationBase.do_tracking`
         for an MLRun job.
         This method should not be called directly.
         """
+
+        if write_output and (
```
assaf758, line 224: why not check this in evaluate() - as the other checks

mlrun/model_monitoring/applications/base.py
```diff
@@ -127,53 +211,78 @@ def _handler(
         start: Optional[str] = None,
         end: Optional[str] = None,
         base_period: Optional[int] = None,
+        write_output: bool = False,
+        stream_profile: Optional[ds_profile.DatastoreProfile] = None,
```
assaf758, line 215: with what value of stream_profile the _handler is called in run_local = false?

assaf758 (commented):
Very nice!
Some comments inline.

mlrun/model_monitoring/applications/base.py
```diff
@@ -127,53 +211,78 @@ def _handler(
...
-                            mm_constants.ApplicationEvent.ENDPOINT_ID: endpoint_id,
-                            mm_constants.ApplicationEvent.START_INFER_TIME: window_start,
-                            mm_constants.ApplicationEvent.END_INFER_TIME: window_end,
-                        }
-                    )
-                    result_key = (
-                        f"{endpoint_name}-{endpoint_id}_{window_start.isoformat()}_{window_end.isoformat()}"
-                        if window_start and window_end
-                        else f"{endpoint_name}-{endpoint_id}"
+        with self._push_to_writer(
```
assaf758, line 238:
```suggestion
        with ModelMonitoringApplicationBase._push_to_writer(
```

mlrun/model_monitoring/applications/_application_steps.py
jond01: When working in Kafka this change is not so helpful, because down the road, the Kafka producer sends the events one by one from the application:
https://github.com/mlrun/mlrun/blob/60b9b831a10ce9a97e50385cddbd7943046f9000/mlrun/platforms/iguazio.py#L244

With V3IO stream, there should indeed be a gain

mlrun/model_monitoring/applications/base.py
jond01: It is checked in `evaluate` in L642.
The check here is for the workflows that use
```py
app_job = app.to_job()
app_job.run(..., write_output=True)
```
In this case, we also want to forbid writing to the DBs when not working with the endpoint data.

mlrun/model_monitoring/applications/base.py
jond01: In a remote run, `stream_profile` is simply `None`. The profile is fetched from the env vars + k8s secrets

mlrun/model_monitoring/applications/base.py
jond01: Any reason for this change other than style?
Regarding style, I prefer using `self.` instead of the full class name when possible

mlrun/model_monitoring/applications/base.py
jond01: I preferred not blocking this, as users may use `register_temporary_client_datastore_profile` instead of passing the profile directly

mlrun/model_monitoring/applications/_application_steps.py
assaf758: Strange that the kafka target works this way, @gtopper do you remember why?

mlrun/model_monitoring/applications/base.py
assaf758: Well, my motivation to this is code-readability (where the peer developer assumes the object is provided when he sees a method called with "self.")
Not critical though.

assaf758 (approved):

PR #8081, opened 2025-06-10
[Nuclio] Fix spec enrichment if base_spec is not empty

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-10206

Test doesn't work without fix. In general, it is better to use `set_function` in tests than `code_to_function` because their behaviour is different. 

The root cause is that if base_spec is not empty, then we do not get a basic config - https://github.com/nuclio/nuclio-jupyter/blob/21b122bf97b86e1926fd0a74b951f377da42f8bb/nuclio/config.py#L37-L54 and as a result, we fail on not having labels key here https://github.com/nuclio/nuclio-jupyter/blob/21b122bf97b86e1926fd0a74b951f377da42f8bb/nuclio/config.py#L391.

tested on a system:
image
image

### DIFF
--- a/server/py/services/api/crud/runtimes/nuclio/function.py
+++ b/server/py/services/api/crud/runtimes/nuclio/function.py
@@ -340,7 +340,7 @@ def _compile_function_config(
             if not mlrun.utils.get_in(config, key):
                 mlrun.utils.update_in(config, key, {})
 
-        if not mlrun.utils.get_in(config,     "spec.handler"):
+        if not mlrun.utils.get_in(config, "spec.handler"):
             # if handler was not set, we set it to the default value
             mlrun.utils.update_in(config, "spec.handler", handler or "main:handler")
         config = nuclio.config.extend_config(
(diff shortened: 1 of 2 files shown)


server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -331,12 +331,21 @@ def _compile_function_config(
...
             config = nuclio.config.new_config()
-            mlrun.utils.update_in(config, "spec.handler", handler or "main:handler")
 
+        # if base_spec comes from client side, make sure that labels and annotations exists in dictionary
+        for key in [
+            "metadata.labels",
+            "metadata.annotations",
+        ]:
+            if not mlrun.utils.get_in(config, key):
+                mlrun.utils.update_in(config, key, {})
```
TomerShor: what about handling it in `_set_function_metadata` and calling it above the if statement (line 321)?

TomerShor (approved):

PR #8085, opened 2025-06-11
[Deprecations] Change functionality of `allow_cross_project` in `mlrun.load_project`

moranbental:
Project name differs from the name specified in the context's project YAML.
This functionality is no longer supported. If you want to enable this behavior, you can take one of the following actions:

1. Set `allow_cross_project=True` when loading the project.
2. Delete the existing project YAML, or ensure its name field matches the actual project name.
3. Use a different project context directory.

Previously, when `allow_cross_project` was not set (None), it implicitly behaved as if it were True, with a warning. Now, if you want this functionality, you must explicitly set `allow_cross_project=True`.

Also, added support for `allow_cross_project` in the CLI with default as true.

https://iguazio.atlassian.net/browse/ML-10062

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -757,14 +757,7 @@ def _project_instance_from_struct(struct, name, allow_cross_project):
             "3. Use different project context dir."
         )
 
-        if allow_cross_project is None:
-            # TODO: Remove this warning in version 1.10.0 and also fix cli to support allow_cross_project
-            warnings.warn(
-                f"Project {name=} is different than specified on the context's project yaml. "
-                "This behavior is deprecated and will not be supported from version 1.10.0."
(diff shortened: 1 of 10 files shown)


mlrun/projects/project.py
```diff
@@ -757,14 +757,7 @@ def _project_instance_from_struct(struct, name, allow_cross_project):
             "3. Use different project context dir."
         )
 
-        if allow_cross_project is None:
-            # TODO: Remove this warning in version 1.10.0 and also fix cli to support allow_cross_project
```
liranbg, line 761: need to add support for cli

liranbg (changes requested):

liranbg (approved):

PR #8114, opened 2025-06-15
[Config] Move db init to end of config population

Yacouby:
This PR improves the configuration loading flow by deferring the DB connection until after the configuration is fully populated.

Additionally, it imports `get_secret_or_env` explicitly in `httpdb.py` to resolve a circular import error that occurred during early mlrun imports.

Jira:
https://iguazio.atlassian.net/browse/ML-10104


tests/test_config.py
```diff
@@ -503,7 +503,9 @@ def test_get_default_function_node_selector():
     assert mlrun.mlconf.get_default_function_node_selector() == {}
 
 
-def test_setting_dbpath_trigger_connect(requests_mock: requests_mock_package.Mocker):
+def test_setting_dbpath_does_not_trigger_connect(
```
liranbg: why? this is a regression isnt it

Yacouby: That's actually the point of this fix.
Previously, setting `mlconf.dbpath` would trigger an early DB connection via `get_run_db()`, which led to unwanted side effects during mlrun import and configuration setup.
With the new behavior, setting `mlconf.dbpath` only updates the configuration and the connection happens at the end of config population.

mlrun/config.py
```diff
@@ -1436,6 +1429,11 @@ def _do_populate(env=None, skip_errors=False):
     _configure_ssl_verification(config.httpdb.http.verify)
     _validate_config(config)
 
+    if config.dbpath:
+        from mlrun.db import get_run_db
+
+        get_run_db(config.dbpath, force_reconnect=True)
```
TomerShor: Keep the comment:
```
# when dbpath is set we want to connect to it which will sync configuration from it to the client
```

TomerShor (changes requested):
Minor

TomerShor (approved):
LGTM

tests/test_config.py
alonmr: this test is now redundant. the test below verifies the same thing

alonmr (changes requested):
Very nice.
IIUC the old test can be removed.
Also we need to update the docs:
If user changes the dbpath, they have to reload the config.

mlrun/config.py
```diff
@@ -1252,15 +1252,8 @@ def dbpath(self):
     def dbpath(self, value):
         self._dbpath = value
         if value:
-            # importing here to avoid circular dependency
-            import mlrun.db
-
-            # It ensures that SSL verification is set before establishing a connection
             _configure_ssl_verification(self.httpdb.http.verify)
```
moranbental: I don't think we need to call _configure_ssl_verification here anymore, since it was only added due to the call to get_run_db, which you've now removed.
That also means we no longer need the getter and setter for dbpath, as they don't really do anything.
You can also remove this functionality as it's no longer needed:
https://github.com/mlrun/mlrun/blob/6299569344e93820c49d8bcbbfc2e8488d1d7878/mlrun/config.py#L894

moranbental (changes requested):

mlrun/config.py
Yacouby: I think you meant removing this:
https://github.com/mlrun/mlrun/blob/3ecfa8e446258dcb00b3f148eef2c58a8a2d200b/mlrun/config.py#L899-L901

and not the whole setattr

Yacouby:
> Reference

Opened a ticket for the doc update: https://iguazio.atlassian.net/browse/ML-10249

Yacouby:
> Very nice. IIUC the old test can be removed. Also we need to update the docs: If user changes the dbpath, they have to reload the config.

@alonmr Opened a ticket for the doc update: https://iguazio.atlassian.net/browse/ML-10249

liranbg (approved):

moranbental (approved):
Great job!

liranbg (approved):

alonmr (approved):

PR #8121, opened 2025-06-16
[Deprecation] Warn on usage of deprecated ml-base image in both client and runtime

Yacouby:
Previously, the warning was only issued in the client side during `prepare_image_for_deploy()`, which could be skipped if the image was set manually after function creation. 
To ensure users are always warned, the check was added also to `_validate_runtime()` in the runtime launcher, which is invoked just before function execution.

Jira:
https://iguazio.atlassian.net/browse/ML-10241


mlrun/launcher/base.py
```diff
@@ -148,6 +148,8 @@ def _validate_runtime(
         self._validate_run_params(run.spec.parameters)
         self._validate_output_path(runtime, run)
 
+        mlrun.utils.helpers.warn_on_deprecated_image(runtime.spec.image)
```
TomerShor: since this is now called in 2 places, is there a flow where a user will get the warning twice?

mlrun/utils/helpers.py
```diff
@@ -2121,6 +2121,32 @@ def join_urls(base_url: Optional[str], path: Optional[str]) -> str:
     return f"{base_url.rstrip('/')}/{path.lstrip('/')}" if path else base_url
 
 
+def warn_on_deprecated_image(image: Optional[str]):
+    """
+    Warn if the provided image is the deprecated 'mlrun/ml-base' image.
+    This image is deprecated as of 1.10.0 and will be removed in 1.12.0.
+    """
+    if image and "mlrun/ml-base" in image:
```
alonmr:
```suggestion
    deprecated_images = ["mlrun/ml-base"]
    if image and any(image in deprecated_image for deprecated_image in deprecated_images):
```

mlrun/utils/helpers.py
```diff
@@ -2121,6 +2121,32 @@ def join_urls(base_url: Optional[str], path: Optional[str]) -> str:
...
+        )
+        message = (
+            "'mlrun/ml-base' image is deprecated in 1.10.0 and will be removed in 1.12.0. "
+            "Use 'mlrun/mlrun' instead."
+        )
+        if auto_replaced:
+            message += (
+                " Since your client version is >= 1.10.0, the image will be automatically "
+                "replaced with mlrun/mlrun."
+            )
```
alonmr: I don't think the message is accurate.
> and will be removed in 1.12.0.

what will be removed?

> Since your client version is >= 1.10.0

Of course the client is 1.10 otherwise it would not have this code.

The message should be something like:
"'mlrun/ml-base' image is deprecated in 1.10.0 and will be replaced by `mlrun/mlrun`. This behavior will be removed in 1.12"

mlrun/launcher/base.py
alonmr: should also validate the base image

alonmr (changes requested):
Some small comments

mlrun/launcher/base.py
Yacouby: Yes, if you call `code_to_function` with passing the `ml-base` image to it, and then running the function, you will get the warning twice.
I think it is acceptable in this case since it's a deprecation warning.
We could add the warning only when running the function, but then in `code_to_function` we would not get it, which is not ideal IMO, since in some flows, the user might rely on `code_to_function` and not actually run the function - so they would miss the warning entirely in this case.

mlrun/utils/helpers.py
Yacouby: @alonmr do you mean that there is no need for the check that is done with `auto_replaced` ? and always show the deprecation message

alonmr: yup :)

mlrun/launcher/base.py
```diff
@@ -148,6 +148,11 @@ def _validate_runtime(
         self._validate_run_params(run.spec.parameters)
         self._validate_output_path(runtime, run)
 
+        deprecated_image = "mlrun/ml-base"
+        images = [runtime.spec.image, getattr(runtime.spec.build, "base_image", None)]
+        if deprecated_image in images:
+            mlrun.utils.helpers.warn_on_deprecated_image(deprecated_image)
```
alonmr: `warn_on_deprecated_image` should get the user image but you passed `deprecated_image = "mlrun/ml-base"`

alonmr:
```suggestion
        for image in [runtime.spec.image, getattr(runtime.spec.build, "base_image", None)]
            mlrun.utils.helpers.warn_on_deprecated_image(image)
```

Yacouby: but in this case we might get two same warnings if both images are defined

Yacouby: nvm, looks like the `warnings` module will only show one warning in this case and ignore duplications of it.

alonmr (approved):

PR #8112, opened 2025-06-15
[Model Monitoring] Extend `project-summary` API

danielperezz:
https://iguazio.atlassian.net/browse/ML-9950?linkSource=email

Extended the project-summary with these new items:
- `running_model_monitoring_functions` (int): number of running mm apps. Will be taken from the functions table. 
- `failed_model_monitoring_functions` (int): number of mm apps that are currently in failed state. Will be taken from the functions table.
- `real_time_model_endpoint_count` (int): number of active real time model endpoints in the project. Will be taken from the meps table. 
- `batch_model_endpoint_count` (int): number of active batch model endpoints in the project. Will be taken from the meps table.
 both by name and cross-project.

PR #8132, opened 2025-06-17
[Monitoring Applications] Extend `project-summary` API

danielperezz:
Follow-up to PR: https://github.com/mlrun/mlrun/pull/8112

Includes moving counters computation logic to the SQL side and adding a suitable index.


server/py/services/api/migrations/versions/692f2f936258_idx_project_state.py
```diff
@@ -0,0 +1,43 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_index(
+        "idx_project_state", "functions", ["project", "state"], unique=False
+    )
+    # ### end Alembic commands ###
```
liranbg, line 37: @theSaarco this is needed fyi

server/py/framework/db/sqldb/db.py
```diff
@@ -3901,16 +3901,38 @@ def _calculate_mm_functions_counters(
...
+                case(
+                    (
+                        subq.c.function_state
+                        == mlrun.common.schemas.FunctionState.error,
+                        1,
+                    )
+                )
+            ).label("failed_count"),
+        ).group_by(subq.c.project)
+        results = query.all()
```
quaark: Query looks good!
For these kinds of complicated queries I'd probably add a unit test to make sure it behaves how you expect it to. Fill the inmemory/file sqlite db with some data and then run the method to see you get the expected numbers

server/py/framework/db/sqldb/db.py
```diff
@@ -3919,17 +3941,20 @@ def _calculate_mm_functions_counters(
...
     def _calculate_mep_counters(session) -> tuple[dict[str, int], dict[str, int]]:
-        query = session.query(ModelEndpoint.project, ModelEndpoint.endpoint_type)
+        query = session.query(
+            ModelEndpoint.project,
+            func.count().label("mep_count"),
+            func.count(
+                case((ModelEndpoint.endpoint_type == EndpointType.BATCH_EP, 1))
+            ).label("batch_count"),
+        ).group_by(ModelEndpoint.project)
+        results = query.all()
```
quaark: This query is simpler, but a unit test for it as well won't hurt

quaark (changes requested):

server/py/framework/db/sqldb/db.py
```diff
@@ -3901,16 +3901,38 @@ def _calculate_mm_functions_counters(
         )  # filter duplications
```
theSaarco, line 3931: Not related to this PR directly, but why are we joining with the tags table? Is there a requirement that the function must have a tag?

server/py/framework/db/sqldb/db.py
```diff
@@ -3901,16 +3901,38 @@ def _calculate_mm_functions_counters(
...
+            func.count(
+                case(
+                    (
+                        subq.c.function_state
+                        == mlrun.common.schemas.FunctionState.error,
+                        1,
+                    )
+                )
+            ).label("failed_count"),
+        ).group_by(subq.c.project)
```
theSaarco: Why do you need a subquery? Can't this be done directly on the original query and add the group-by clause to it with the counters?
Also, to make it faster (and since you're adding a project+state index anyway), it might be faster to execute the query if you add a filter of `state in (error, ready)`.
And, finally, using `case` is nice, but it might be clearer if all you do is `group by project,state` and select a simple `count(*)`, and then in the python code extract the values according to state.

theSaarco (commented):
Looks good, I think the query can be made simpler.

quaark (approved):
Other than @theSaarco 's suggestions, looks good

server/py/framework/db/sqldb/db.py
```diff
@@ -3894,45 +3894,42 @@ def _calculate_feature_sets_counters(session) -> dict[str, int]:
...
-                case(
-                    (
-                        subq.c.function_state
-                        == mlrun.common.schemas.FunctionState.error,
-                        1,
-                    )
-                )
-            ).label("failed_count"),
-        ).group_by(subq.c.project)
+        labels = label_set(labels)
```
theSaarco: This is now redundant.

theSaarco (approved):
👍 Almost there - need to remove a redundant line. 
I'm approving, but please make sure to clean it.

liranbg (approved):
🚀

Eyal-Danieli (approved):

server/py/framework/db/sqldb/db.py
```diff
@@ -3946,17 +3964,22 @@ def _calculate_mm_functions_counters(
...
 
         project_to_real_time_mep_count = {}
         project_to_batch_mep_count = {}
-        for project, endpoint_type in query.all():
-            project_to_real_time_mep_count.setdefault(project, 0)
-            project_to_batch_mep_count.setdefault(project, 0)
+        for project, endpoint_type, count in results:
             if endpoint_type == EndpointType.BATCH_EP:
-                project_to_batch_mep_count[project] += 1
+                project_to_batch_mep_count[project] = count
```
assaf758, line 3978: why don't you aggregate as in the real_time case?

assaf758 (approved):

PR #8140, opened 2025-06-18
[Go] Fix security vulnerabilities

alonmr:
Raised by `govulncheck`
* GO-2025-3751
* GO-2025-3750
* GO-2025-3749
* GO-2025-3563


server/go/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/mlrun
 
-go 1.23.8
+go 1.24.4
```
liranbg: is it required? we're jumping between odd usually

PR #8143, opened 2025-06-18
[Workflows] Add allow cross project flag to workflow runner 

yaelgen:
This PR sets allow_cross_project=True when calling `load_project` and `get_or_create_project` in the workflow-runner logic. This change prevents failures when the project name in the project.yaml differs from the one passed to the function, which is a common scenario in remote workflows.

Without this flag, the remote runner fail with a project name mismatch error, as shown in the attached example. We already rely on this behavior in our system tests for remote workflows, where the project name is intentionally different.

image

https://iguazio.atlassian.net/browse/ML-10297


tests/system/projects/test_project.py
```diff
@@ -630,7 +630,7 @@ def test_remote_pipeline_with_workflow_runner_node_selector(self):
 
         # Test scheduled workflow
         schedule = "0 0 30 2 *"
-        run_id = project.run(
+        project.run(
```
moranbental, line 633: Why is the run_id removed? I see it was used for system test - why is it no longer needed?

mlrun/projects/pipelines.py
```diff
@@ -1153,6 +1153,7 @@ def load_and_run_workflow(
     project = mlrun.get_or_create_project(
         context=project_context or f"./{project_name}",
         name=project_name,
+        allow_cross_project=True,
```
moranbental, line 1156: why not adding a new parameter of `allow_cross_project` to `load_and_run_workflow`?
@liranbg Or maybe that means we cannot deprecate it because users won’t be able to prepare their code for the deprecation in version 1.8?

moranbental (changes requested):

tests/system/projects/test_project.py
yaelgen: Not related, I debugged the test, and turns out it is failing since the scheduled run does not return the id. 
It is needed but need to find another way to test it.

yaelgen: Not related, I debugged the test, and turns out it is failing since the scheduled run does not return the id.
It is needed but need to find another way to test it. will do in a follow-up,

mlrun/projects/pipelines.py
yaelgen: The users already opt into allow_cross_project during project creation if needed. The workflow-runner is an internal mechanism that simply reloads the same project. It’s not really a user-facing flow, and users aren’t meant to be aware of what it does—so exposing flexibility here doesn’t provide any real benefit and only adds unnecessary complexity.

If we did want to make it configurable, the only places to expose it would be through project.run()—which doesn't make sense since the user shouldn't need to be aware of it—or by checking it ourselves before creating the runner. But since the project might have been created earlier and not in the same process, we’d need to fetch it and inspect its configuration, which also feels like overhead.
Also BC wise, we can't count on old clients to pass this param for runners that may use newer versions, it will always fail for them. 

Do you see any harm in keeping allow_cross_project=True unconditionally in this internal flow?

liranbg: I dont see why it should be configurable tbh and it is pretty much a safe-guard for existing projects from failing. I agree with @yaelgen here.
while we want to deprecate the behavior, it is crucial not to fail existing workflows and code but to put the safeguards on the entry of "new code"

liranbg (approved):

tests/system/projects/test_project.py
moranbental: > Not related, I debugged the test, and turns out it is failing since the scheduled run does not return the id. It is needed but need to find another way to test it. will do in a follow-up,

So please mention that in the PR description, as this is not really related

PR #8144, opened 2025-06-18
[Requirements] Add nuclio-requirements.txt

rokatyy:
Since nuclio has specific requirement for msgpack and nuclio-sdk-py, `mlrun` image should include these requirements as nuclio loads them only from local wheels and doesn't go to pypi.

Requirements in nuclio: https://github.com/nuclio/nuclio/blob/development/pkg/processor/runtime/python/py/requirements/common.txt#L4


nuclio-requirements.txt
```diff
@@ -0,0 +1,2 @@
+nuclio-sdk==0.5.15
+msgpack==1.1.0
```
liranbg, line 5: why not ~ it? at least on patch version

nuclio-requirements.txt
jond01, line 1: Can you please add a link to the relevant requirements in Nuclio?

nuclio-requirements.txt
rokatyy: @liranbg because in nuclio it's not `~`, so if the correct version is not in the wheel, then kaniko build will fail

nuclio-requirements.txt
rokatyy: done (added to description)

jond01: Thanks, please also add it as a comment in the file for future maintainers 🙂

rokatyy: added ✅

nuclio-requirements.txt
liranbg: fair. what's the impact if you have newer nuclio version with 1.1.1? that means nuclio <> mlrun relationship is stricter now
I wonder if we have another option here to mitigate the relationship. perhaps fixing the issue on nuclio to be less strict and more lean.

and we should have `storey` not pin `nuclio-sdk` as it should not "demand" the sdk but rather inherit the version it is given with

rokatyy:
Closing, as it was decided on another solution https://github.com/nuclio/nuclio/pull/3666

PR #8146, opened 2025-06-18
[Docs] Update the documentation by adding a link to the commit activity icon

Olexandr88:
Знімок екрана 2025-06-18 о 18 12 29

If I understand correctly, this link should be here
https://docs.mlrun.org/en/stable/concepts/alerts-notifications.html


liranbg:
Thanks @Olexandr88 - 
I am not sure the pr description matches the pr title / changes. Please so update it or correct me, until then I will merge it as it make sense to make the activity icon refer users to mlrun commit  backlog

Olexandr88:
Thanks for the help, what do you think about this link? is it correct?
[Alerts and notifications:](https://docs.mlrun.org/en/stable/concepts/model-monitoring.html)

I think it should be replaced with this

https://docs.mlrun.org/en/stable/concepts/alerts-notifications.html

jillnogold:
@Olexandr88 
That is correct.
And Model monitoring should point to: https://docs.mlrun.org/en/stable/concepts/model-monitoring.html

Olexandr88:
I understood

PR #8148, opened 2025-06-18
[Scripts] Fix mlrun-kfp using patch remote script 

yaelgen:
Set `MLRUN_KFP_IMAGE` env var in mlrun-api patch to ensure workflow runner pulls the image from the internal registry.

reverts: https://github.com/mlrun/mlrun/pull/8124

### DIFF
--- a/automation/patch_igz/patch_remote.py
+++ b/automation/patch_igz/patch_remote.py
@@ -45,7 +45,6 @@ class Constants:
     mlrun_kfp = "mlrun-kfp"
     log_collector = "log-collector"
     default_namespace = "default-tenant"
-    workflow_controller = "workflow-controller"
     alerts = "mlrun-alerts"
     targets_to_image_name = {
         api: api_container,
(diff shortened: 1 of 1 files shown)


adamdelman (approved):
lagtam

automation/patch_igz/patch_remote.py
```diff
@@ -234,6 +227,25 @@ def _make_targets(
...
+                # so that workflow pods will use the correct KFP image from the internal registry.
+                _, overwrite_registry = self._resolve_overwrite_registry()
+                kfp_image_uri = (
+                    f"{overwrite_registry}/mlrun/{Constants.mlrun_kfp}:{image_tag}"
+                )
+
+                mlrun_api_container = self._deploy_patch["mlrun_api"]["spec"][
+                    "template"
+                ]["spec"]["containers"][0]
+                mlrun_api_container.setdefault("env", []).append(
```
liranbg: Many patches would cause endless amount of envvars. Perhaps you can iterate over and set the value if it exists or append if not

liranbg (changes requested):

liranbg (approved):

PR #8149, opened 2025-06-18
[Pipelines] Fix pipeline arg validation

adamdelman:
https://iguazio.atlassian.net/browse/ML-10089


alonmr (changes requested):
bump patch version

liranbg (approved):

PR #8156, opened 2025-06-19
[Tests] Fix `TestMonitoredServings::test_different_kind_of_serving`

royischoss:
Fixed test_different_kind_of_serving system test. Increase timeout post inference + added thread safe behavior inside of `_test_endpoint`


tests/system/model_monitoring/test_app.py
```diff
@@ -70,6 +71,8 @@
 )
 from .assets.custom_evidently_app import DemoEvidentlyMonitoringApp
 
+lock = threading.Lock()
```
gtopper: Should probably be a class member with a more specific name, or a parameter to `_test_endpoint`.

tests/system/model_monitoring/test_app.py
```diff
@@ -1231,41 +1234,42 @@ def _test_endpoint(
...
-                {"inputs": data_point},
-            ),
-        )
-        if endpoint_name == "img_one_to_one":
-            data_point = data_point[0]
-        serving_fn.invoke(
-            f"v2/models/{endpoint_name}/infer",
-            json.dumps({"inputs": [data_point, data_point]}),
-        )
+        with lock:
```
gtopper: But really, wouldn't it be simpler to forgo the lock and just run the first section in serial, sleep, and run the second section in serial as well? Why parallelize something that we don't want running in parallel?

royischoss: I tried to keep it similar to the original, I saw performance difference between a simple loop to the way we run it now with the locks. I guess the sleep is utilize better

gtopper: Not sure what you mean by a simple loop? It would be two loops – loop one, sleep, loop two. Is that what you tried?

royischoss: nop I have put one loop running _test_endpoint you probably right two loops will do the trick, but this is not the only usage of the method

royischoss: I will try to make it two simple loops

gtopper (approved):

PR #8157, opened 2025-06-19
[API] Paginate list runs on retry jobs periodic [feature/retry-job]

alonmr:
https://iguazio.atlassian.net/browse/ML-10253

Having weird issues with the unit tests not being able to see the runs with new session for a second after commit.
Also tried auxiliary function to poll the runs every 0.1 seconds but it returns immediately even with new session on a different thread.

### DIFF
--- a/server/py/services/api/main.py
+++ b/server/py/services/api/main.py
@@ -934,6 +934,7 @@ class Service(framework.service.Service):
         """
         self._logger.debug("Retrying jobs with retry policy configured")
         db_session = await fastapi.concurrency.run_in_threadpool(create_session)
+        fetch_runs_limit = int(mlconf.monitoring.runs.retry.fetch_runs_limit)
         try:
             offset = 0
             while runs := await fastapi.concurrency.run_in_threadpool(
@@ -941,7 +942,7 @@ class Service(framework.service.Service):
                 db_session,
                 project="*",
                 states=[mlrun.common.runtimes.constants.RunStates.pending_retry],
-                limit=int(mlconf.monitoring.runs.retry.fetch_runs_limit),
+                limit=fetch_runs_limit,
                 offset=offset,
             ):
                 self._logger.debug(
--- a/server/py/services/api/main.py
+++ b/server/py/services/api/main.py
@@ -935,72 +935,72 @@ class Service(framework.service.Service):
         self._logger.debug("Retrying jobs with retry policy configured")
         db_session = await fastapi.concurrency.run_in_threadpool(create_session)
         try:
-            runs = await fastapi.concurrency.run_in_threadpool(
+            offset = 0
+            while runs := await fastapi.concurrency.run_in_threadpool(
                 get_db().list_runs,
                 db_session,
                 project="*",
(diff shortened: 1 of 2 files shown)


server/py/services/api/tests/unit/test_service.py
```diff
@@ -81,6 +82,34 @@ async def test_retried_jobs_cache(self, db: Session):
...
+        run = self._generate_retry_job(
+            uid=running_run_uid, state=mlrun.common.runtimes.constants.RunStates.running
+        )
+        mlrun.db.get_run_db().store_run(
+            struct=run, uid=running_run_uid, project=self._project
+        )
+
+        # There seems to be some OS level caching that prevents the new db session from seeing the new runs immediately.
+        # Also, waiting less than a second does not work.
+        await asyncio.sleep(1)
```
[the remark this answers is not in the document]
alonmr: this doesnt work

server/py/services/api/tests/unit/test_service.py
```diff
@@ -81,6 +82,34 @@ async def test_retried_jobs_cache(self, db: Session):
...
 
+    async def test_retry_job_paginated_list_runs(self, db: Session):
+        mlrun.mlconf.monitoring.runs.retry.fetch_runs_limit = 3
+        mlrun.mlconf.function.spec.retry.backoff.min_base_delay = "0s"
+        run_uids = [str(uuid.uuid4()) for _ in range(10)]
+        for run_uid in run_uids:
+            run = self._generate_retry_job(uid=run_uid)
+            mlrun.db.get_run_db().store_run(
+                struct=run, uid=run_uid, project=self._project
+            )
```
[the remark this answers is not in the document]
alonmr: we have a fixture to reset it

server/py/services/api/main.py
```diff
@@ -934,74 +934,75 @@ async def _retry_jobs(self):
...
-                            project=run.metadata.project,
+                limit=fetch_runs_limit,
+                offset=offset,
+            ):
+                self._logger.debug(
+                    "Found runs to retry", runs_count=len(runs), offset=offset
+                )
+                offset = offset + len(runs)
+
+                futures = []
```
liranbg, line 953: use semaphore. retrying 1k runs in parallel would lead to 1k requests to update in parallel. perhaps chunk their async by something configurable would be better.

liranbg (commented):
very well, minor suggestion

server/py/services/api/main.py
alonmr: This code is just for sanity.  This should never happen:
```
# sanity
if not run.status.retry_count < run.spec.retry.count:
```
because if this is true then the run should already be in error state by runs monitoring or client error handling code

alonmr: client - https://github.com/mlrun/mlrun/blob/feature/retry-job/mlrun/execution.py#L1161
server - https://github.com/mlrun/mlrun/blob/feature/retry-job/server/py/services/api/runtime_handlers/base.py#L1759

PR #8159, opened 2025-06-19
[RemoteModel] Add ModelProvider infrastructure + openaiprovider

tomerm-iguazio:
1. **Added Infrastructure for model providers**
    a) Add `ModelProvider` abstract class, that determines the interface of all model providers + `AsyncModelProvider`.
    b) Add `BaseRemoteClient` Abstract class: , for common functions between datastore and ModelProvider.
    c) Support `ModelProvider` in `StoreManager`, and wrap it with `get_model_provider`  in mlrun `run.py`.

2. **Created OpenAI implementation**
   a) Add `OpenAIProvider` class that managing synchronous  and asynchronous openai clients, and basic and custom  
        invocations.
   b) Add `DatastoreProfileOpenAI` class, for credentials handling across the project.
   
3. **Updated Model class** to work with `ModelProvider` by default, supporting `LLMPromptArtifact` and `ModelArtifact` usage.

4. **Implemented synchronous test cases for OpenAI integration:**
    a) Unit tests - Check that OpenAI Model Provider was created with all user parameters (Optional). 
    b) Regular integration tests- check default operation `chat.completions.create.`
    c) Comprehensive Integration tests, including building a `ModelArtifact` And `LLmPromptArtifact`, with server  
         mocking.
    d) Full system tests, including the same functionality of the Comprehensive Integration tests, with function deployment.

5. **Bugfix** including `LLMPromptArtifact` in `is_artifact` function.

[ML-10254](https://iguazio.atlassian.net/browse/ML-10254)

[ML-10254]: https://iguazio.atlassian.net/browse/ML-10254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/datastore/datastore.py
```diff
@@ -256,5 +228,125 @@ def get_or_create_store(
...
     def reset_secrets(self):
         self._secrets = {}
+
+    def get_or_create_model_provider(
+        self,
+        url,
+        secrets: Optional[dict] = None,
+        project_name="",
+        default_invoke_kwargs: Optional[dict] = None,
+    ) -> (ModelProvider, str, str):
```
theSaarco: What are those other two `str` return values? Looks like you only return the `ModelProvider`.

mlrun/datastore/datastore.py
```diff
@@ -256,5 +228,125 @@ def get_or_create_store(
...
+        url,
+        secrets: Optional[dict] = None,
+        project_name="",
+        subpath: Optional[str] = None,
+    ):
+        datastore_profile = datastore_profile_read(url, project_name, secrets)
+        if secrets and datastore_profile.secrets():
+            secrets = merge(secrets, datastore_profile.secrets())
+        else:
+            secrets = secrets or datastore_profile.secrets()
```
theSaarco:
```suggestion
        secrets = merge(secrets or {}, datastore_profile.secrets() or {})
```

mlrun/datastore/abstract_base.py
```diff
@@ -0,0 +1,74 @@
+# Copyright 2025 Iguazio
```
alxtkr77: Not a good name for the file. There is nothing abstract inside. Maybe remote_client.py?

mlrun/datastore/abstract_base.py
```diff
@@ -0,0 +1,74 @@
...
+        hostname_index_in_netloc = lower_netloc.index(str(lower_hostname))
+        endpoint = netloc[
+            hostname_index_in_netloc : hostname_index_in_netloc + len(lower_hostname)
+        ]
+    if parsed_url.port:
+        endpoint += f":{parsed_url.port}"
+    return schema, endpoint, parsed_url
+
+
+class BaseRemoteClient(ABC):
```
alxtkr77: Why inherit from ABC? No abstract interfaces defined. Maybe just move it to utils.py as well?

mlrun/datastore/abstract_base.py
```diff
@@ -0,0 +1,74 @@
...
+# limitations under the License.
+
+from abc import ABC
+from typing import Optional
+from urllib.parse import urlparse
+
+import mlrun
+
+
+def parse_url(url):
```
alxtkr77: I would move parse_url() to mlrun/datastore/utils.py

mlrun/datastore/abstract_base.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class BaseRemoteClient(ABC):
+    def __init__(self, parent, name, kind, endpoint="", secrets: Optional[dict] = None):
+        self._parent = parent
+        self.name = name
+        self.kind = kind
+        self.endpoint = endpoint
+        self._secrets = secrets or {}
+        self.secret_pfx = ""
```
alxtkr77: Why is there secret_pfx ? Who changes it? If someone does, I think there should be a setter. Otherwise it should be removed

mlrun/datastore/base.py
```diff
@@ -45,16 +46,18 @@ def __repr__(self):
...
 
-class DataStore:
+class DataStore(BaseRemoteClient):
     using_bucket = False
 
     def __init__(self, parent, name, kind, endpoint="", secrets: Optional[dict] = None):
+        super().__init__(
+            parent=parent, name=name, kind=kind, endpoint=endpoint, secrets=secrets
+        )
         self._parent = parent
```
alxtkr77: probably all this should be removed as well, since they are set in the RemoteClient

mlrun/datastore/model_providers.py
```diff
@@ -0,0 +1,229 @@
...
+
+import mlrun
+from mlrun.datastore.abstract_base import (
+    BaseRemoteClient,
+)
+
+T = TypeVar("T")
+
+
+class ModelProvider(BaseRemoteClient, ABC):
```
alxtkr77: Why inherit ABC? No abstract interfaces defined, and also BaseRemoteClient inherits from ABC

mlrun/datastore/model_providers.py
```diff
@@ -0,0 +1,229 @@
...
+    def model(self):
+        return None
+
+    def get_invoke_kwargs(self, invoke_kwargs):
+        kwargs = self.default_invoke_kwargs.copy()
+        kwargs.update(invoke_kwargs)
+        return kwargs
+
+
+class AsyncModelProvider(ModelProvider, ABC):
```
alxtkr77: I don't think its a good inheritance - this implies AsyncModelProvider is as  ModelProvider, which is actually SyncModelProvider (it has invoke()). So either create a base class ModelProvider that has no invoke and then inherit from it SyncModelProvider and AsyncModelProvider, or move async interfaces to the ModelProvider so it would support both sync and async (which I think is better)

mlrun/artifacts/base.py
```diff
@@ -942,3 +940,11 @@ def fill_artifact_object_hash(object_dict, iteration=None, producer_id=None):
             object_dict["spec"][key] = value
 
     return uid
+
+
+def verify_target_artifact(artifact: Artifact):
```
theSaarco:
```suggestion
def verify_target_path(artifact: Artifact):
```

mlrun/datastore/datastore.py
```diff
@@ -175,12 +148,21 @@ def get_store_artifact(
...
-            raise mlrun.errors.MLRunInvalidArgumentError(
-                f"Resource {url} does not have a valid/persistent offline target"
-            )
+        if not allow_empty_resources:
+            if isinstance(resource, LLMPromptArtifact):
+                if not resource.spec.model_uri:
+                    raise mlrun.errors.MLRunInvalidArgumentError(
+                        f"LLMPromptArtifact {url} does not contain model artifact uri"
+                    )
+            if not target and not (
```
theSaarco: Shouldn't there be an `else` here? In this case `LLMPromptArtifact` would still be tested, and since it has no `target` and it's not ` ModelArtifact` this will explode.

mlrun/config.py
```diff
@@ -184,6 +184,7 @@
         "url": "",
     },
     "v3io_framesd": "http://framesd:8080",
+    "model_providers": {"openai_default_model": "gpt-4"},
```
theSaarco:
```suggestion
    "model_providers": {
        "openai_default_model": "gpt-4"
    },
```
To be consistent with the rest of the file.

mlrun/datastore/datastore.py
```diff
@@ -256,5 +238,105 @@ def get_or_create_store(
...
+        if secrets and datastore_profile.secrets():
+            secrets = merge(secrets, datastore_profile.secrets())
+        else:
+            secrets = secrets or datastore_profile.secrets()
+        url = datastore_profile.url(subpath)
+        schema, endpoint, parsed_url = parse_url(url)
+        subpath = parsed_url.path
+        return secrets, url, schema, endpoint, parsed_url, subpath
+
+    def set(self, secrets=None, db=None):
```
theSaarco: set what?
I would split this to two methods -
`add_secrets(self, secrets: dict[str, str]=None)` - because you're not really setting them, you're adding to the existing secrets.
`set_db(db=None)` - configure the db.

mlrun/datastore/model_providers.py
```diff
@@ -0,0 +1,229 @@
...
+    async def async_invoke(self, prompt: str, **invoke_kwargs) -> Awaitable[str]:
+        raise NotImplementedError("async_invoke is not implemented")
+
+
+class OpenAIProvider(AsyncModelProvider):
+    def __init__(
+        self,
+        parent,
+        name,
+        kind,
```
theSaarco: What if kind here is not `openai`? Does it even make sense? Maybe should validate this here.

mlrun/datastore/datastore.py
```diff
@@ -256,5 +238,105 @@ def get_or_create_store(
...
+
+        model_provider = model_provider_class(
+            parent=self,
+            name=key,
+            kind=schema,
+            endpoint=endpoint,
+            secrets=secrets,
+            default_invoke_kwargs=default_invoke_kwargs,
+        )
+        return model_provider
```
theSaarco, line 334: The `StoreManager` class has a cache of datastores that it holds, so that if you ask for the same store again it doesn't have to be re-created. There is some logic around it, for example it only stores in cache if running on client side etc. (see `get_or_create_store` for that). 
Shouldn't we do the same for model providers? Does it even make sense? It might not be a very useful idea due to the fact that the `invoke_kwargs` may be very different, but it's something to think about.

mlrun/datastore/model_providers.py
```diff
@@ -0,0 +1,229 @@
+# Copyright 2025 Iguazio
```
theSaarco: Why not put `model_providers` as a folder under `datastore`, and then have a file for the base classes, and another one for `openai`? First, I'd like to separate them from the "regular" datastores, and second, there's no reason to put all model providers in a single file.

mlrun/datastore/model_providers.py
theSaarco: Another approach would be to use mixins for that, and have sync invoke as one mixin and async invoke as another, and each model-provider can extend one of them or both (or none, which will make it not very much interesting).

mlrun/datastore/model_providers.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def load_client(self) -> None:
+        try:
+            from openai import OpenAI, AsyncOpenAI  # noqa
+
+            self._client = OpenAI(**self.options)
+            self._default_operation = self.client.chat.completions.create
+
+            self._async_client = AsyncOpenAI(**self.options)
+            self._default_async_operation = self.async_client.chat.completions.create
```
theSaarco: I seem to recall we said we'll just have a single async client (instead of having both), and just implement the sync methods by calling the async + doing an `await` on that.

mlrun/datastore/model_providers.py
```diff
@@ -0,0 +1,229 @@
...
+        return messages, invoke_kwargs
+
+    async def async_invoke(self, prompt: str, **invoke_kwargs) -> Awaitable[str]:
+        messages, invoke_kwargs = self.get_messages_parameter(
+            prompt=prompt, **invoke_kwargs
+        )
+        response = await self._default_async_operation(
+            model=self.endpoint, messages=messages, **invoke_kwargs
+        )
+        return response.choices[0].message.content
```
theSaarco: Why doing an `await` in the `async_invoke` method? You return the actual content after waiting for the response, so what makes this async?

mlrun/serving/states.py
```diff
@@ -1014,15 +1016,32 @@ def __init__(
         if artifact_uri is not None and not isinstance(artifact_uri, str):
             raise MLRunInvalidArgumentError("'artifact_uri' argument must be a string")
         self.artifact_uri = artifact_uri
+        self.invocation_artifact: Optional[LLMPromptArtifact] = None
+        self.model_artifact: Optional[ModelArtifact] = None
+        self.model: Optional[ModelProvider] = None
```
theSaarco: Why not `self.model_provider`?

tests/datastore/test_base.py
```diff
@@ -203,6 +204,18 @@ def test_schema_to_store(schemas, expected_class, expected):
         assert all(store == expected_class for store in stores)
 
 
+@pytest.mark.parametrize(
+    "schemas,expected_class,expected",
+    [
+        (["openai"], OpenAIProvider, does_not_raise()),
```
theSaarco, line 210: Add a negative test - the infra already covers it.

tests/integration/model_providers/openai/test_openai.py
```diff
@@ -0,0 +1,255 @@
...
+            )
+        return body
+
+    def enrich_prompt(self, body) -> str:
+        if isinstance(self.invocation_artifact, mlrun.artifacts.LLMPromptArtifact):
+            prompt_template = self.invocation_artifact.spec.prompt_string
+            needed_params = ["question", "depth_level", "persona", "tone"]
+            sub_dict = {k: body[k] for k in needed_params if k in body}
+            return prompt_template.format(**sub_dict)
+        return body["prompt"]
```
theSaarco: As discussed - this will become mostly not needed once the enrichment is implemented in the `Model` class. Maybe add a `TODO` here so we don't forget.

theSaarco (changes requested):
Overall looks good - please see my comments.

mlrun/datastore/model_providers.py
tomerm-iguazio: I do not think it's a good idea for few reasons:

1) The memory usage of holding openai sdk client is minimal (less than 0.1 KB).
2) it would require us to call `async_invoke` from the `invoke` method by starting a new event loop using asyncio.run(self.async_invoke(...)). This introduces overhead and may lead to issues in environments where an event loop is already running.
3)  I want to give the user the option to access both the client and async_client for more customized usagee.

mlrun/datastore/abstract_base.py
tomerm-iguazio: I’d like this to be in a separate PR:

https://github.com/mlrun/mlrun/pull/8231

mlrun/datastore/base.py
tomerm-iguazio: Removed duplicates, tnx!

(subpath, filesystem, etc... stil relevant).

mlrun/datastore/model_providers.py
tomerm-iguazio: Changed to single ModelProvider class, as @alxtkr77 suggests, tnx!

mlrun/datastore/datastore.py
tomerm-iguazio: Related to our discussion tomorrow — happy to adjust this based on what we align on during the meeting.

mlrun/datastore/datastore.py
tomerm-iguazio: This is existing functionality — I’ll move it back to its original location to avoid it being flagged as a change. However, given its widespread usage, I don’t think we should modify it.

mlrun/datastore/model_providers.py
tomerm-iguazio: fixed, tnx!

mlrun/datastore/abstract_base.py
tomerm-iguazio: decided to keep it because of backwards compatibility, anyway this parameter in use in `_get_parent_secret` in `BaseRemoteClient`, so we need it in `BaseRemoteClient`.

mlrun/datastore/datastore.py
tomerm-iguazio: fixed

mlrun/datastore/datastore.py
tomerm-iguazio: disabled cache in model provider.
tnx!

alxtkr77 (approved):

theSaarco (approved):
LGTM

tests/system/model_monitoring/assets/models.py
```diff
@@ -83,23 +83,13 @@ async def predict_async(self, body):
...
 class MyRemoteModel(mlrun.serving.states.Model):
-    def __init__(self, name, raise_exception, artifact_uri, **kwargs):
-        super().__init__(
-            name=name,
-            raise_exception=raise_exception,
-            artifact_uri=artifact_uri,
-            **kwargs,
-        )
-        self.artifact = None
+    execution_mechanism = "naive"
```
assaf758: execution_mechanism is no longer a class attribute, it is a prameter of add_model

tomerm-iguazio: Already added as part of the tests, now removed from the model classes:)

assaf758 (approved):

PR #8172, opened 2025-06-22
[Model Serving] Move execution mechanism from `Model` to `add_model`

gtopper:
For better ease-of-use and flexibility.

~Pending https://github.com/mlrun/storey/pull/577.~

[ML-10259](https://iguazio.atlassian.net/browse/ML-10259)

~Current CI failure is due to the name parameter disappearing during serialization (shown in #8244), following merge of #8183.~

[ML-10259]: https://iguazio.atlassian.net/browse/ML-10259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/serving/states.py
```diff
@@ -1208,6 +1209,7 @@ def add_model(
         self,
         endpoint_name: str,
         model_class: str,
+        execution_mechanism: str,
```
theSaarco: Why not provide a default value here? This will force any user adding any model to provide a mechanism. It could default to `naive`, or any other mechanism that we decide makes sense by default.

mlrun/serving/states.py
```diff
@@ -1225,6 +1227,24 @@ def add_model(
...
+        * "asyncio" – To run in an asyncio task. This is appropriate for I/O tasks that use asyncio, allowing the event
+            loop to continue running while waiting for a response.
+        * "shared_executor" – Reuses an external executor (typically managed by the flow or context) to execute the
+            runnable. Should be used only if you have multiply `ParallelExecution` in the same flow and especially
+            useful when:
+            - You want to share a heavy resource like a large model loaded onto a GPU.
+            - You want to centralize task scheduling or coordination for multiple lightweight tasks.
+            - You aim to minimize overhead from creating new executors or processes/threads per runnable.
+          The runnable is expected to be pre-initialized and reused across events, enabling efficient use of memory and
+          hardware accelerators.
```
theSaarco: Where's `naive`?

tests/serving/test_async_flow.py
```diff
@@ -513,18 +569,24 @@ def test_model_runner_with_selector(execution_mechanism: str):
...
 
     function = mlrun.new_function("tests", kind="serving")
     graph = function.set_topology("flow", engine="async")
     model_runner_step = ModelRunnerStep(
         name="my_model_runner",
     )
-    model_runner_step.add_model(endpoint_name=m1.name, model_class=m1)
-    model_runner_step.add_model(endpoint_name=m2.name, model_class=m2)
+    model_runner_step.add_model(
+        endpoint_name=m1.name, model_class=m1, execution_mechanism="naive"
```
theSaarco, line 585: Can you do `model_class=m1`, i.e. passing the actual object? The function hint doesn't say that (it says just `str`).

mlrun/serving/states.py
```diff
@@ -1208,6 +1209,7 @@ def add_model(
         self,
         endpoint_name: str,
         model_class: str,
```
theSaarco: The tests, as well as the docstring of the class, indicate that you can pass a model instance here. However, this only says `str`, and the code itself actually has a TODO to allow passing the instance. So, what's the status? Is it supported or not?

theSaarco (changes requested):
If we're adding parameters, I would at least add a default value to make it easier for the user.
Also, I have some questions on how the `model_class` parameter is handled.

tests/serving/test_async_flow.py
gtopper: From what I remember, it works when run locally, but will fail if needs to be serialized. We don't want the user to do this.

mlrun/serving/states.py
gtopper: Support was removed, I guess not thoroughly enough.

mlrun/serving/states.py
gtopper: I don't think there is a default value we can choose without creating a pitfall for the user, though.

mlrun/serving/states.py
gtopper: @royischoss has #8183 open to take care of this.

tests/serving/test_async_flow.py
gtopper: @royischoss has #8183 open to take care of this.

mlrun/serving/states.py
gtopper: From our discussion on Monday, it sounds like there is a tendency to continue requiring the user to specify this, because it's important, and there is no default that would be appropriate for a large majority of cases.

dockerfiles/gpu/locked-requirements.txt
```diff
@@ -3812,8 +3812,8 @@ starlette==0.46.2 \
     --hash=sha256:595633ce89f8ffa71a015caed34a5b2dc1c0cdb3f0f1fbd1e69339cf2abeec35 \
     --hash=sha256:7f7361f34eed179294600af672f565727419830b54b7b084efe44bb82d2fccd5
     # via fastapi
-storey==1.10.4 \
-    --hash=sha256:4917c4d56c5e26663015206a923eb5f3d0f6a577d23c6110e3af7d4249f34f8f
+storey==1.10.6 \
```
davesh0812, line 3815: please bump to storey==1.10.7 :)

requirements.txt
```diff
@@ -31,7 +31,7 @@ dependency-injector~=4.41
 # should be identical to gcs and s3fs.
 fsspec>=2023.9.2, <2024.7
 v3iofs~=0.1.17
-storey==1.10.4
+storey~=1.10.6
```
davesh0812: please bump to storey~=1.10.7

tests/test_requirements.py
```diff
@@ -122,7 +122,7 @@ def test_requirement_specifiers_convention():
     ignored_invalid_map = {
         # See comment near requirement for why we're limiting to patch changes only for all of these
         "aiobotocore": {">=2.5.0,<2.16"},
-        "storey": {"==1.10.4"},
+        "storey": {"~=1.10.6"},
```
davesh0812:
```suggestion
        "storey": {"~=1.10.7"},
```

tests/serving/test_async_flow.py
royischoss: [#8183](https://github.com/mlrun/mlrun/pull/8183) was merged so either way should work

mlrun/serving/states.py
royischoss: https://github.com/mlrun/mlrun/pull/8183 was merged so Model  can be provided

royischoss (approved):
LGTM - description should be changed, does test_deploy_function_with_model_runner passes?

mlrun/serving/states.py
```diff
@@ -1221,6 +1222,7 @@ def add_model(
         self,
         endpoint_name: str,
         model_class: Union[str, Model],
+        execution_mechanism: str,
```
```suggestion
        execution_mechanism: Union[str, ParallelExecutionMechanisms],
```

davesh0812 (approved):
LGTM, please make sure that the relevant sys test are passing

gtopper:
Passed `test_deploy_function_with_model_runner` and `test_app_flow`. Will merge once CI passes.

theSaarco (approved):
👍

PR #8183, opened 2025-06-23
[Serving] Allow `ModelRunnerStep:add_model` to get model_class as `Model` object

royischoss:
https://iguazio.atlassian.net/browse/ML-9924
Contains:
1. fix for `ModelRunnerStep:add_model` allowing object and retrieving needed information.
2. fix for `ModelRunnerStep:init_object`
3. fix for `ModelObj:from_dict` allowing initialization using params in the init


gtopper:
I guess we should have a system test that covers it? Especially given that the issue is with serialization.

mlrun/model.py
```diff
@@ -221,15 +221,26 @@ def _apply_enrichment_before_to_dict_completion(
...
         if not fields:
             fields = list(inspect.signature(cls.__init__).parameters.keys())
-        new_obj = cls()
+
+        if init_with_params:
+            kwargs = {field: struct.get(field, None) for field in fields}
+            kwargs.pop("self", None)
+            new_obj = cls(**kwargs)
+        else:
+            new_obj = cls()
```
davesh0812: Need to consider 3 main points : 
1. What happen when user define `_dict_fields` and passes `init_with_params=True` - I think you should use the signature when you initialized the object.  
2. What happens if `**kwargs` is part of the `cls` signature.
3. Why not to delete the fields you already use for the initialization from the struct?

assaf758: Do we really need to introduce the new `init_with_params` parameter? I mean, why not do as we do below, and check if `struct` exist in order to decide what is needed?

royischoss: I will add a new suggestion maybe separate fields and parameters in the initialization. @davesh0812 about number 3 I want to prevent lost of logic the user might did on his object. meaning even if I pass a parameter in the initialization the user might have change it in many way in his own code, I want to take the updated one

tests/serving/test_async_flow.py
```diff
@@ -161,8 +161,8 @@ def test_push_error():
 class MyModel(Model):
     execution_mechanism = "naive"
 
-    def __init__(self, *args, inc: int, gpu_number: Optional[int] = None, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, name: str, inc: int, gpu_number: Optional[int] = None, **kwargs):
```
davesh0812: please check it also when the user Model don't have explicit `name` param

mlrun/serving/states.py
```diff
@@ -1280,7 +1291,15 @@ def add_model(
...
-        models[endpoint_name] = (model_class, model_parameters)
+        model_class_dict = (
+            model_class.to_dict() if isinstance(model_class, Model) else None
+        )
+        model_class = (
+            model_class
+            if isinstance(model_class, str)
+            else model_class.__class__.__name__
+        )
+        models[endpoint_name] = (model_class, model_parameters, model_class_dict)
```
davesh0812: As we agree lets try to use treat model_class_dict and model_parameters in the same way.
Also you can get the model_class_dict before the first if statement (L-1277) and not to change it,

davesh0812 (changes requested):

mlrun/serving/states.py
```diff
@@ -1251,8 +1261,15 @@ def add_model(
         :param override:            bool allow override existing model on the current ModelRunnerStep.
         :param model_parameters:    Parameters for model instantiation
         """
-        # TODO allow model_class as Model object as part of ML-9924
-        model_parameters = model_parameters or {}
+
+        if not isinstance(model_class, str) and model_parameters:
```
assaf758:
```suggestion
        if isinstance(model_class, Model) and model_parameters:
```

mlrun/serving/states.py
```diff
@@ -1251,8 +1261,15 @@ def add_model(
...
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "Cannot provide as model initialized model object with model_parameters."
```
assaf758:
```suggestion
                "Cannot provide a model object as argument to `model_class` and also provide `model_parameters`."
```

mlrun/model.py
```diff
@@ -221,15 +221,28 @@ def _apply_enrichment_before_to_dict_completion(
...
+        struct=None,
+        fields=None,
+        deprecated_fields: Optional[dict] = None,
+        init_with_params: bool = False,
     ):
         """create an object from a python dictionary"""
         struct = {} if struct is None else struct
         deprecated_fields = deprecated_fields or {}
-        fields = fields or cls._dict_fields
+        fields = fields or copy(cls._dict_fields)
```
assaf758: Better do a copy anyway (so as not to change the `fields` dict arg)

gtopper: Maybe I'm missing something, but it doesn't seem like any copy is necessary here, because the `fields` list is never modified?

davesh0812 (approved):

PR #8186, opened 2025-06-23
[Functions] Handle null function state in build status

Yacouby:
This PR introduces a mitigation for cases where a function's `status.state` is null in the DB, which causes encoding issues when querying the `/build/status` endpoint.

To ensure a valid and consistent API response, a new function state `initialized` was added, so when a function has no state, the build status response returns `initialized`, preventing encoding errors in the response headers.

A more robust solution would be to ensure a default state is set during function storage (store_function), which requires more work and can be handled later.

Jira:
https://iguazio.atlassian.net/browse/ML-10175

### DIFF
--- a/server/py/services/api/api/endpoints/functions.py
+++ b/server/py/services/api/api/endpoints/functions.py
@@ -458,7 +458,9 @@ def _handle_job_deploy_status(
 ):
     # job deploy status
     response_headers = {}
-    function_state = get_in(fn, "status.state", "")
+    function_state = (
+        get_in(fn, "status.state", "") or mlrun.common.schemas.FunctionState.initialized
+    )
     pod = get_in(fn, "status.build_pod", "")
     image = get_in(fn, "spec.build.image", "")
     out = b""
--- a/mlrun/common/schemas/function.py
+++ b/mlrun/common/schemas/function.py
@@ -47,6 +47,7 @@ class FunctionState:
 
     # for pipeline steps
     skipped = "skipped"
+    initialized = "initialized"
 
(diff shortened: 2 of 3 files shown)


liranbg (approved):

server/py/services/api/api/endpoints/functions.py
```diff
@@ -458,7 +458,9 @@ def _handle_job_deploy_status(
 ):
     # job deploy status
     response_headers = {}
-    function_state = get_in(fn, "status.state", "")
+    function_state = (
+        get_in(fn, "status.state", "") or mlrun.common.schemas.FunctionState.initialized
```
TomerShor, line 462:
```suggestion
        get_in(fn, "status.state", mlrun.common.schemas.FunctionState.initialized)
```

TomerShor (changes requested):

server/py/services/api/api/endpoints/functions.py
Yacouby: That won’t work as expected because `get_in` returns the actual value even if it’s None.. so if `status.state` exists but is null, the default won’t be used—you’ll still get None.

TomerShor: Then essentially the default will never be used?
Sounds like a bug to me

Yacouby: The current behavior of `get_in` only applies the default if a key along the path is missing — not if its value is None.

TomerShor (approved):

PR #8195, opened 2025-06-24
[Images] Fix ml-base image replacement logic to support rc tags

Yacouby:
Jira :
https://iguazio.atlassian.net/browse/ML-10275


TomerShor (approved):
LGTM

PR #8196, opened 2025-06-24
[Pipelines] Add RerunRunner and basic rerun_workflow handler

yaelgen:
ticket: https://iguazio.atlassian.net/browse/ML-10218

* Introduce a new `RerunRunner` class for scheduling retry jobs of failed remote KFP workflows.
* Add a minimal `rerun_workflow` handler that calls the `retry_pipeline`, records the new pipeline ID in the context, and waits for its completion.
* **Note**: this is purely an infrastructure phase—no end-to-end flow or tests yet.

**Next Steps**

1. Add this runner into the `/retry-workflow` API endpoint.
2. Implement full notification handling to match the standard `remote:kfp` flow.
3. Add unit and system tests covering the retry lifecycle and notification behavior.


mlrun/common/schemas/workflow.py
```diff
@@ -46,6 +46,13 @@ class WorkflowRequest(pydantic.v1.BaseModel):
     notifications: typing.Optional[list[Notification]] = None
 
 
+class RerunWorkflowRequest(pydantic.v1.BaseModel):
+    run_name: typing.Optional[str] = (None,)
+    rerun_id: typing.Optional[str] = (None,)
```
liranbg: why not `run_id` to align with `retry_pipeline` parameter names

mlrun/projects/pipelines.py
```diff
@@ -1070,6 +1070,40 @@ def github_webhook(request):
...
+
+        # wait for pipeline completion so monitor will push terminal notifications
+        wait_for_pipeline_completion(
+            new_pipeline_id,
+            project=project_name,
+        )
+
+    # Temporary exception
+    except Exception as exc:
+        context.logger.error("Failed to rerun workflow", exc=exc)
```
liranbg: mlrun err to str

server/py/services/api/crud/workflows.py
```diff
@@ -627,3 +636,91 @@ def get_workflow_id(
...
+            source=project.spec.source,
+            project_name=project.metadata.name,
+            save=False,
+            handler="mlrun.projects.rerun_workflow",
+            parameters={
+                "run_uid": rerun_request.rerun_id,
+                "project_name": project.metadata.name,
+            },
+            notifications=notifications,
+            run_name=f"rerun-{rerun_request.run_name}-{rerun_request.rerun_id}",
```
liranbg: could it be too long? I suggest to remote the rerun id from the name - adding perhaps an increment of the reruns could be an option?

liranbg (commented):
preliminary review

server/py/services/api/crud/workflows.py
yaelgen: It is too long for sure, I wanted to run it and see what fits better (probably without the id, cause it will be on the labels)

liranbg (approved):

PR #8199, opened 2025-06-25
[Serving] Improvements related to getting background task status

gtopper:
* Increase log level from debug to info and error.
* Add `get_project_background_task` stub in `RunDBInterface`.

### DIFF
--- a/mlrun/serving/routers.py
+++ b/mlrun/serving/routers.py
@@ -185,12 +185,12 @@ class BaseModelRouter(RouterToDict):
                     background_task.status.state
                     in mlrun.common.schemas.BackgroundTaskState.terminal_states()
                 ):
-                    logger.debug(
+                    logger.info(
                         f"Model endpoint creation task completed with state {background_task.status.state}"
                     )
                     self._background_task_terminate = True
                 else:  # in progress
-                    logger.debug(
+                    logger.info(
                         f"Model endpoint creation task is still in progress with the current state: "
                         f"{background_task.status.state}. Events will not be monitored for the next 15 seconds",
                         name=self.name,
@@ -198,7 +198,7 @@ class BaseModelRouter(RouterToDict):
                     )
                 return background_task.status.state
             else:
-                logger.debug(
+                logger.error(
                     "Model endpoint creation task name not provided",
                 )
         elif self.context.monitoring_mock:
--- a/mlrun/serving/routers.py
+++ b/mlrun/serving/routers.py
@@ -86,8 +86,8 @@ class BaseModelRouter(RouterToDict):
     @property
     def background_task_reached_terminal_state(self):
         return (
-            self._background_task_check_timestamp
-            and self._background_task_check_timestamp != "running"
+            self._background_task_current_state
+            and self._background_task_current_state != "running"
         )
 
     def parse_event(self, event):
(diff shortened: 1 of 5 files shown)


mlrun/serving/routers.py
```diff
@@ -185,20 +185,20 @@ def _get_background_task_status(
                     background_task.status.state
                     in mlrun.common.schemas.BackgroundTaskState.terminal_states()
                 ):
-                    logger.debug(
+                    logger.info(
```
royischoss, line 194: you want make it info? I think we did it as debug so we won't spam the log although it is every 15 seconds.

mlrun/serving/routers.py
```diff
@@ -185,20 +185,20 @@ def _get_background_task_status(
...
                         f"Model endpoint creation task completed with state {background_task.status.state}"
                     )
                     self._background_task_terminate = True
                 else:  # in progress
-                    logger.debug(
+                    logger.info(
                         f"Model endpoint creation task is still in progress with the current state: "
                         f"{background_task.status.state}. Events will not be monitored for the next 15 seconds",
```
royischoss: If we already changing something here I added in my pr a config constant for this check
`mlrun.mlconf.model_endpoint_monitoring.model_endpoint_creation_check_period` we can add it here in the comment and in the check when calling this method

royischoss (commented):
Two small comments probably the log level is fine but if we open the time period we check the background task for user configuration it might spam the log. see in comments

mlrun/serving/routers.py
gtopper: Pretty sure it should only be printed once, when the terminal state is reached?

mlrun/serving/routers.py
```diff
@@ -80,10 +80,16 @@ def __init__(
...
         self._background_task_check_timestamp = None
-        self._background_task_terminate = False
         self._background_task_current_state = None
         self.kwargs = kwargs
 
+    @property
+    def background_task_reached_terminal_state(self):
+        return (
+            self._background_task_check_timestamp
+            and self._background_task_check_timestamp != "running"
```
royischoss:
```suggestion
            and self._background_task_current_state != "running"
```

mlrun/serving/system_steps.py
```diff
@@ -314,7 +314,8 @@ def _log_background_task_state(
         else:  # in progress
             logger.info(
                 f"Model endpoint creation task is still in progress with the current state: "
-                f"{background_task_state}. Events will not be monitored for the next 15 seconds",
+                f"{background_task_state}. Events will not be monitored for the next "
+                f"{mlrun.mlconf.model_endpoint_monitoring.model_endpoint_creation_check_period} seconds",
```
royischoss, line 318: 👑

mlrun/serving/routers.py
```diff
@@ -80,10 +80,16 @@ def __init__(
...
         self._result_path = result_path
         self._background_task_check_timestamp = None
-        self._background_task_terminate = False
         self._background_task_current_state = None
         self.kwargs = kwargs
 
+    @property
+    def background_task_reached_terminal_state(self):
+        return (
+            self._background_task_check_timestamp
```
royischoss:
```suggestion
            self._background_task_current_state
```

royischoss (commented):
a small one

royischoss (approved):
LGTM 🔥

PR #8202, opened 2025-06-25
[Package] Improve the build

jond01:
Continue the work from #8100:

- Add the required uv version and upgrade uv
  https://docs.astral.sh/uv/reference/settings/#required-version
  Remove the `verify-uv-version` target.
- Package build:
  - Remove the deprecated tests_require package input
    https://setuptools.pypa.io/en/latest/deprecated/changed_keywords.html#new-and-changed-setup-keywords
  - Remove the deprecated license classifier 
    https://packaging.python.org/en/latest/specifications/pyproject-toml/#classifiers

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -144,15 +144,8 @@ jobs:
     needs: [ prepare-inputs, trigger-and-wait-for-ui-image-building, trigger-and-wait-for-mlrun-image-building ]
     steps:
       - uses: actions/checkout@v4
-      - name: Get uv version
-        id: uv-version
-        run: |
-          version=$(grep "^MLRUN_UV_VERSION ?= " Makefile | cut -d' ' -f3)
-          echo "version=$(version)" >> $GITHUB_OUTPUT
       - name: Set up uv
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yaml
```diff
@@ -144,21 +144,12 @@ jobs:
...
       - name: Set up uv
         uses: astral-sh/setup-uv@bd01e18f51369d5a26f1651c3cb451d3417e3bba # 6.3.1
-        with:
-          version: ${{ steps.uv-version.outputs.version }}
       - name: Build & push to PyPI
         if: github.event.inputs.skip_publish_pypi != 'true'
         env:
           MLRUN_VERSION: ${{ needs.prepare-inputs.outputs.version }}
-          UV_PUBLISH_USERNAME: ${{ secrets.PYPI_USERNAME }}
-          UV_PUBLISH_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
```
liranbg: I think it can be on a separated PR and have this one merged until it resolves internally.

liranbg (changes requested):

.github/workflows/release.yaml
jond01: 9004e46

liranbg (approved):

PR #8203, opened 2025-06-25
[Frameworks] Fixes for ML-10270 and ML-10333

guy1992l:
* Optimizer is returned and not stored.
* Model file renamed also in OS level.

Tickets:
* https://iguazio.atlassian.net/browse/ML-10270?isEligibleForUserSurvey=true
* https://iguazio.atlassian.net/browse/ML-10333


yonishelach (approved):
Looking good

PR #8210, opened 2025-06-25
[Model endpoints] Fix model endpoint uid parse as hex back from db

royischoss:
fix https://iguazio.atlassian.net/browse/ML-10292
parse uid as hex back from DB


liranbg (changes requested):
I suggest to add a ut coverage to ensure this wont break again

moranbental (commented):
what about this place - https://github.com/mlrun/mlrun/blob/5e2f42164a334a2e0396236e8d33879c299f4641/server/py/framework/db/sqldb/db.py#L8010 from this pr - https://github.com/mlrun/mlrun/commit/5e2f42164a334a2e0396236e8d33879c299f4641

assaf758 (approved):

PR #8211, opened 2025-06-25
[Datastore] Rename store_manager to remote_item_manager

tomerm-iguazio:
For a more generic approach with remote items, the functions, class names, and modules in store_manager need to be decoupled from the datastore-specific syntax.

[ML-10347](https://iguazio.atlassian.net/browse/ML-10347)

[ML-10347]: https://iguazio.atlassian.net/browse/ML-10347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/datastore/remote_manager.py
+++ b/mlrun/datastore/remote_manager.py
@@ -116,7 +116,7 @@ def uri_to_ipython(link):
     return schema_to_store(schema).uri_to_ipython(endpoint, parsed_url.path)
 
 
-class StoreManager:
+class RemoteClientManager:
     def __init__(self, secrets=None, db=None):
         self._stores = {}
         self._secrets = secrets or {}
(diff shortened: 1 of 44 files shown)


moranbental (commented):
So in this PR, you basically changed the class name from `StoreManager` to `RemoteItemManager`, and I think this breaks compatibility without proper deprecation. Until now, users could access `mlrun.datastore.store_manager`, but it no longer exists — they now need to use `mlrun.datastore.remote_item_manager`.

What was the reason or purpose behind this change?

mlrun/datastore/remote_manager.py
```diff
@@ -116,7 +116,7 @@ def uri_to_ipython(link):
     return schema_to_store(schema).uri_to_ipython(endpoint, parsed_url.path)
 
 
-class StoreManager:
+class RemoteItemManager:
```
liranbg, line 119:
```suggestion
class RemoteItemManager:
  ...
  

# class name is preserved for bc
StoreManager = RemoteItemManager
```

liranbg: it would make sense to retain class names when classes are accessible to end-users

PR #8206, opened 2025-06-25
[Job] Fix retry status message [feature/retry-job]

alonmr:
Fix status text to reflect correct attempts count
Add tests (system test was mistakenly removed by another PR)

PR #8213, opened 2025-06-26
[Job] Fix reached max attempts test condition [feature/retry-job]

alonmr:
Added the +1 to retries in https://github.com/mlrun/mlrun/pull/8206 which broke the logic and moved the run to error before the last attempt


Yacouby (approved):
LGTM

PR #8216, opened 2025-06-26
[Launcher] Strip run start time upon retry [feature/retry-job]

alonmr:
Strip the start time to make sure we update to start time of current attempt


mlrun/launcher/base.py
```diff
@@ -370,6 +370,8 @@ def _enrich_run(
             run.status.state = mlrun.common.runtimes.constants.RunStates.running
             run.status.retry_count = run.status.retry_count or 0  # in case it is none
             run.status.retry_count += 1  # increment by one
+            # TODO: Maintain start time of each retry ML-10169
```
Yacouby, line 373: is this #TODO needed ?

alonmr: I added it to make it clear that this is the place where the change is needed and also I'm not sure it will be done in 1.10 but would definitely be nice to have

Yacouby (approved):

PR #8218, opened 2025-06-26
[Model Monitoring] Support all model endpoints in `evaluate`

jond01:
Implements [ML-10264](https://iguazio.atlassian.net/browse/ML-10264).

Add the `endpoints="all"` option in `app.evaluate`.

- Support this option also when running the job directly.
- Breaking change: when someone wants to run on a model endpoint by its name, they could previously pass `endpoints="<mep-name>"`. They will now have to wrap it in a list `endpoints=["<mep-name>"]`.
- Update the docstrings and add informative error messages.

[ML-10264]: https://iguazio.atlassian.net/browse/ML-10264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/model_monitoring/applications/base.py
```diff
@@ -223,7 +223,9 @@ def _handler(
         context: "mlrun.MLClientCtx",
         sample_data: Optional[pd.DataFrame] = None,
         reference_data: Optional[pd.DataFrame] = None,
-        endpoints: Optional[Union[list[tuple[str, str]], list[str], str]] = None,
+        endpoints: Union[
+            list[tuple[str, str]], list[list[str]], list[str], Literal["all"], None
```
assaf758, line 227: What is this new parameter type `list[list[str]]`, when would it be used?
Also I don't see it's covered in `_handle_endpoints_type_evaluate` and in the tests...

jond01: It's not really new - it was missing.
When running a remote job, Python tuples are serialized to lists, and in particular, `list[tuple[str, str]]` becomes `list[list[str]]`.
For example, `endpoints = [("name", "uid")]` will be passed as is in a local job run, but in a remote run, it will be passed as the JSON serializable `[["name", "uid"]]`.

Regarding testing, the unit tests cover a part of the local case, and the system tests cover both local and remote runs.

assaf758 (approved):

PR #8219, opened 2025-06-26
[Publishing] Use PyPI trusted publishers for `mlrun`

jond01:
Remove PyPI username and password - use trusted publishers instead.
This PR continues the work from #8202.

See:
- https://docs.astral.sh/uv/guides/package/#publishing-your-package
- https://docs.pypi.org/trusted-publishers/adding-a-publisher/
- https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#defining-a-workflow-job-environment

TODO (together with @nashpaz123):
- [ ] Configure the trusted publishers on PyPI

After it is merged, we most probably can remove the `PYPI_USERNAME` and `PYPI_PASSWORD` secrets from GitHub and revoke the PyPI token.


jond01:
We could not set it in the relevant PyPI account. Closing for now.

PR #8230, opened 2025-06-29
[Logs] Handle log collection per attempt [feature/retry-job]

alonmr:
Send a start logs request per each attempt differentiating the logs by including the attempt in the "run uid" forwarded to the log collector.
Small gap that this solution has is if the log collector is down for a long period of time, we might skip collecting logs for attempts.

https://iguazio.atlassian.net/browse/ML-10346

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -608,7 +608,7 @@ class HTTPRunDB(RunDBInterface):
         error = f"store log {project}/{uid}"
         self.api_call("POST", path, error, params, body)
 
-    def get_log(self, uid, project="", offset=0, size=None):
+    def get_log(self, uid, project="", offset=0, size=None, attempt=None):
         """Retrieve 1 MB data of log.
 
         :param uid: Log unique ID
@@ -616,6 +616,7 @@ class HTTPRunDB(RunDBInterface):
         :param offset: Retrieve partial log, get up to ``size`` bytes starting at offset ``offset``
             from beginning of log (must be >= 0)
         :param size: If set to ``-1`` will retrieve and print all data to end of the log by chunks of 1MB each.
+        :param attempt: For retriable runs, the attempt number to retrieve the log for. 0 is the initial attempt.
         :returns: The following objects:
 
             - state - The state of the runtime object which generates this log, if it exists. In case no known state
@@ -636,6 +637,8 @@ class HTTPRunDB(RunDBInterface):
             return state, offset
 
         params = {"offset": offset, "size": size}
+        if attempt:
+            params["attempt"] = attempt
         path = self._path_of("logs", project, uid)
         error = f"get log {project}/{uid}"
         resp = self.api_call("GET", path, error, params=params)
@@ -658,7 +661,7 @@ class HTTPRunDB(RunDBInterface):
         resp = self.api_call("GET", path, error)
         return resp.json()["size"]
 
-    def watch_log(self, uid, project="", watch=True, offset=0):
+    def watch_log(self, uid, project="", watch=True, offset=0, attempt=None):
         """Retrieve logs of a running process by chunks of 1MB, and watch the progress of the execution until it
         completes. This method will print out the logs and continue to periodically poll for, and print,
         new logs as long as the state of the runtime which generates this log is either ``pending`` or ``running``.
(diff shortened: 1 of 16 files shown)


mlrun/db/httpdb.py
```diff
@@ -668,10 +672,11 @@ def watch_log(self, uid, project="", watch=True, offset=0):
         :param watch: If set to ``True`` will continue tracking the log as described above. Otherwise this function
             is practically equivalent to the :py:func:`~get_log` function.
         :param offset: Minimal offset in the log to watch.
+        :param attempt: For retriable runs, the attempt number to retrieve the log for. 0 is the initial attempt.
```
Yacouby: why here 0 is the initial attempt, and in httpdb 1 is the initial ?

server/py/services/api/launcher.py
```diff
@@ -234,9 +233,38 @@ def _enrich_run(
...
         return self._pre_run_scheduling_constraints_enrichment(runtime, run)
 
+    @staticmethod
+    def _handle_retry(run: mlrun.run.RunObject):
+        if run.status.state != mlrun.common.runtimes.constants.RunStates.pending_retry:
+            return
+
+        run.status.state = mlrun.common.runtimes.constants.RunStates.running
+        run.status.retry_count = run.status.retry_count or 0  # in case it is none
+        run.status.retry_count += 1  # increment by one
```
Yacouby: comment should be placed above the code line

Yacouby (changes requested):
couple of questions:
- If multiple retries are triggered close together, `_start_log_for_run() ` might be invoked multiple times for the same `run_uid-attempt-N`, potentially causing duplicate `start_logs()` calls
- Would it make sense to support attempt="all" in `get_logs()` to concatenate logs from all attempts? This could improve usability for who want a complete view of the run's output without specifying each attempt manually

server/py/framework/rundb/sqldb.py
```diff
@@ -73,15 +73,9 @@ def store_log(self, uid, project="", body=b"", append=False):
...
-        # TODO: this is method which is not being called through the API (only through the SDK), but due to changes in
-        #  the API we changed the get_log method to async so we cannot call it here, and in this PR we won't change the
-        #  SDK to run async, we will use the legacy method for now, and later when we will have a better solution
-        #  we will change it.
+    def get_log(self, uid, project="", offset=0, size=0, attempt=None):
         raise NotImplementedError(
-            "This should be changed to async call, if you are running in the API, use `services.api.crud.get_log`"
-            " method directly instead and not through the get_db().get_log() method. "
-            "This will be removed in 1.5.0",
+            "Use `services.api.crud.get_log`method directly instead and not through the get_db().get_log() method. "
```
TomerShor:
```suggestion
            "Use `services.api.crud.get_log` method directly instead, and not through the get_db().get_log() method. "
```

alonmr:
> * If multiple retries are triggered close together, `_start_log_for_run() ` might be invoked multiple times for the same `run_uid-attempt-N`, potentially causing duplicate `start_logs()` calls
> * Would it make sense to support attempt="all" in `get_logs()` to concatenate logs from all attempts? This could improve usability for who want a complete view of the run's output without specifying each attempt manually

1. `start_logs` is idempotent by design so that's ok
2. maybe in the future because its not trivial, since we use offset and size and if got EOF you will need to fill the response from the next attempt logs until the log size reaches the requested size

server/py/services/api/main.py
```diff
@@ -509,9 +510,15 @@ async def _start_log_for_run(
                     # runtimes that the user will create with hundreds of resources (e.g mpi job can have multiple
                     # workers which aren't really important for log collection
                     with_main_runtime_resource_label_selector=True,
+                    retry_count=retry_count,
                 )
+                logs_run_uid = run_uid
+                if retry_count:
+                    # Adding the attempt number to the run uid since the log collector does not support multiple pods
+                    # per run uid. This separates the attempts so that each attempt has its own logs file.
+                    logs_run_uid = f"{run_uid}-attempt-{int(retry_count)+1}"
```
TomerShor, line 520: Why +1 ?

TomerShor (changes requested):

server/py/services/api/main.py
alonmr: added a comment

Yacouby (approved):

TomerShor (approved):
LGTM

liranbg (approved):

PR #8231, opened 2025-06-29
[Datastore] remove secret pfx

tomerm-iguazio:
secret_pfx is no longer in use.

Following @alxtkr77 comment


moranbental (commented):
I didn't see that the `secret_pfx` field had been deprecated before

PR #8235, opened 2025-06-30
[Serving] Added special plot for `ModelRunnerStep`

royischoss:
https://iguazio.atlassian.net/browse/ML-9897
Includes:
1. Adding track_models flag to `RootFlowStep`
2. Update this flag in `set_tracking`
3. Adding `_add_graphviz_model_runner`

image


assaf758 (approved):

mlrun/runtimes/nuclio/serving.py
```diff
@@ -331,6 +331,8 @@ def set_tracking(
         """
         # Applying model monitoring configurations
         self.spec.track_models = enable_tracking
+        if self.spec.graph and isinstance(self.spec.graph, RootFlowStep):
+            self.spec.graph.track_models = enable_tracking
```
davesh0812, line 336: what about the child function graph?

royischoss: Seems that child function graph is None in this stage so probably the copy is closer to the deployment and that way the above update should do the trick

davesh0812: Sounds good.. please test it :)

royischoss: I have checked the child function can have it own graph, covered it from both time lines, tracking->add_child function and add_child_function->set_tracking, using set_topology otherwise we take the parent graph

assaf758:
I thought maybe it would be clearer to the user that he has 22 models and model-monitoring if we would change the display to the below:
```
22 -> 22M
M -> MM
```
what do you think?

royischoss:
I can check if adding models next to number is too much, neither way documentation should be added. I will plot the different suggestions and we will decide

davesh0812 (approved):

PR #8237, opened 2025-06-30
[Bugfix] MLRun K8S client raises an SSL error

adamdelman:
## Fix SSL-retry regression from [#8220](https://github.com/mlrun/mlrun/pull/8220)

* **Dev deps**
  * add `testcontainers[k3s]`

* **`K8sHelper`**
  * support providing `kube_config_path`

* **Integration tests** (K3s)
  * invalid CA → expect `ApiException` with `urllib3.exceptions.SSLError`
  * valid CA → `list_pods()` works
  
  https://iguazio.atlassian.net/browse/ML-10374
https://iguazio.atlassian.net/browse/ML-9573


server/py/services/api/tests/integration/__init__.py
liranbg, line 1: add path to to `make test-integration` and ignore on `make test`

liranbg (changes requested):
verify against an iguazio system

server/py/services/api/tests/integration/k8s/conftest.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+@pytest.fixture(scope="session")
+def k3s():
+    logging.getLogger("urllib3").setLevel(logging.WARNING)
+    logging.getLogger("testcontainers").setLevel(logging.DEBUG)
+
+    # Ensure that the testcontainers library can find the Docker socket when running inside a Docker container.
+    # This env var is parsed by testcontainers at import time.
+    os.environ["TESTCONTAINERS_HOST_OVERRIDE"] = "host.docker.internal"
```
liranbg, line 36: you only need to do it when running in a container.
but, when running on your local env (e.g mac) - is it required as well?

liranbg (commented):
question

server/py/services/api/tests/integration/k8s/conftest.py
adamdelman: It is not required, but has no effect as this is the default on docker desktop as well
I'd rather not start checking platform and doing different stuff per platform

quaark (approved):

PR #8238, opened 2025-06-30
[Docs] Add details about run-id when creating alerts

jillnogold:
[ML-10367](https://iguazio.atlassian.net/browse/ML-10367)

[ML-10367]: https://iguazio.atlassian.net/browse/ML-10367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/concepts/alerts.md
```diff
@@ -56,12 +56,21 @@ See {ref}`model-monitoring-overview` for more details on drift and performance.
 ## Creating an alert
 When creating an alert you can select an event type for a specific model, for example `data_drift_suspected` or any of the predefined events above.
 You can optionally specify the frequency of the alert through the criteria field in the configuration (how many times in what time window, etc.).
```
Yacouby:
```suggestion
You can optionally specify the frequency of the alert using the criteria field, which controls how many times the event must occur in a given time window to trigger the alert.
```

docs/concepts/alerts.md
```diff
@@ -56,12 +56,21 @@ See {ref}`model-monitoring-overview` for more details on drift and performance.
 ## Creating an alert
 When creating an alert you can select an event type for a specific model, for example `data_drift_suspected` or any of the predefined events above.
 You can optionally specify the frequency of the alert through the criteria field in the configuration (how many times in what time window, etc.). 
-If not specified, it uses the default.
+If not specified, it uses the default. Alerts track the runs by name, not by UID. You can configure Git, Slack, and webhook notifications for the alert.
```
Yacouby:
```suggestion
If criteria is not specified, the default is count=1 and period=None. This means the alert will trigger immediately upon the first matching event.
You can configure Slack, Git, or webhook notifications for the alert.
```

docs/concepts/alerts.md
```diff
@@ -56,12 +56,21 @@ See {ref}`model-monitoring-overview` for more details on drift and performance.
 ## Creating an alert
 When creating an alert you can select an event type for a specific model, for example `data_drift_suspected` or any of the predefined events above.
 You can optionally specify the frequency of the alert through the criteria field in the configuration (how many times in what time window, etc.). 
-If not specified, it uses the default.
+If not specified, it uses the default. Alerts track the runs by name, not by UID. You can configure Git, Slack, and webhook notifications for the alert. 
+
```
Yacouby:
```suggestion
Note on run identification:
Alerts track job runs by name (run.metadata.name), not by the unique run UID. The run name can either be set explicitly or automatically generated when a job is executed. 
You can access the run name from the result of the run_function call:
For example:
run = project.run_function("my-function", handler="handler", local=True)
run_id = run.metadata.name

```

docs/concepts/alerts.md
```diff
@@ -56,12 +56,21 @@ See {ref}`model-monitoring-overview` for more details on drift and performance.
...
+For alerts on model endpoints, see [Creating a model monitoring alert](#creating-a-model-monitoring-alert).
+
+This example illustrates creating an alert with a Slack notification for a job failure with defined criteria. 
+This example uses `run_id`. You can set it to the run’s name (run.metadata.name), which is assigned when you run a job function, and then
+access it from the run object returned by `run_function`.
+For example:
+```
+run = project.run_function("my-function", handler="handler", local=True)
+run_id = run.metadata.name
+```
```
Yacouby:
```suggestion
This example uses `run_id`. You can set it to the run’s name (run.metadata.name), which is assigned when you run a job function.
```

Yacouby (changes requested):

docs/concepts/alerts.md
```diff
@@ -55,13 +55,23 @@ See {ref}`model-monitoring-overview` for more details on drift and performance.
 
 ## Creating an alert
 When creating an alert you can select an event type for a specific model, for example `data_drift_suspected` or any of the predefined events above.
-You can optionally specify the frequency of the alert through the criteria field in the configuration (how many times in what time window, etc.). 
-If not specified, it uses the default.
+You can optionally specify the frequency of the alert using the criteria field, which controls the threshold number of events in a given time window that triggers the alert.
+If criteria is not specified, the default is `count=1` and `period=None`, so thatß the alert triggers immediately upon the first matching event.
```
Yacouby:
```suggestion
If criteria is not specified, the default is `count=1` and `period=None`. This means that the alert will trigger immediately upon the first matching event.
```

docs/concepts/alerts.md
```diff
@@ -55,13 +55,23 @@ See {ref}`model-monitoring-overview` for more details on drift and performance.
...
+You can configure Slack, Git, or webhook notifications for the alert.
+``` {Admonition} Note on run identification
+Alerts track job runs by name (`run.metadata.name`), not by the unique run UID. The run name can either be set explicitly or automatically generated when a job is executed. 
+You can access the run name from the result of the `run_function call`, for example:
```
Yacouby:
```suggestion
You can access the run name from the result of the `run_function` call, for example:
```

docs/concepts/alerts.md
```diff
@@ -1,7 +1,7 @@
 (alerts)=
 # Alerts 
 
-Alerts are a generic mechanism that allows you to define specific rules using events, such as: an event happens N times within a specified time period. They can also be used for model monitoring where the basic notification is just for jobs and workflows.
+Alerts are a generic mechanism that allows you to define specific rules using events, such as: an event happens N times within a specified time period. They can also be used for model monitoring (as opposed to notifications that are ony used with jobs and workflows).
```
Yacouby: I would modify this explanation a bit for something like :
```suggestion
Alerts mechanism provide a flexible way to detect and respond to important system events - such as job failures or model drift. You can define alerts using conditions like “event X happens N times in T minutes,” and attach notifications to be triggered when the alert is activated. 
```

Yacouby (changes requested):
more changes for clarity

docs/concepts/alerts.md
jillnogold: call is not part of the actual code, so not inside '...'

docs/concepts/alerts.md
jillnogold: reworded

docs/concepts/alerts.md
jillnogold: reworded

docs/concepts/alerts.md
Yacouby: yes thats exactly what I suggested here..

docs/concepts/alerts-notifications.md
```diff
@@ -1,7 +1,7 @@
 (alerts-notifications)=
 # Alerts and notifications 
 
-{ref}`notifications` are used to notify you of the system events, for example, start/finish of processes, and error messages; {ref}`alerts` are a generic mechanism that allows you to define more specific rules, such as an event happens N times within a time period, and can also be used for model monitoring where the basic notification is just for jobs and workflows.
+{ref}`notifications` are used to notify you of system events on jobs (both scheduled and not), for example, start/finish of processes, and error messages; {ref}`alerts` are a generic mechanism that allows you to define more specific rules, such as an event X happens N times in T minutes, and can also be used for model monitoring.
```
Yacouby: what notifications do we have on not scheduled jobs ? this is not so clear..
also we do not have start notifications on regular jobs, we have it on workflows.
i suggest to refine this explanation

Yacouby (changes requested):

Yacouby (approved):

alonmr (approved):

PR #8241, opened 2025-06-30
[Model Monitoring] Add `function-summary` API

Eyal-Danieli:
Following #7959 , this PR introduces a new API: `function-summary` (`GET /api/v1/projects/{project}/model-monitoring/function-summary/{function_name}`), which returns metadata and statistics for a specific monitoring application. This API is similar to the existing `function-summaries` API but additionally includes:
- The number of processed model endpoints
- The latest results and metrics of the application

Please note that this PR currently supports only `TDEngine`. Support for `V3IO` TSDB will be added in a separate PR.

JIRA: https://iguazio.atlassian.net/browse/ML-9951


mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -825,6 +825,163 @@ def count_results_by_status(
...
+    def count_processed_model_endpoints(
+        self,
+        start: Optional[Union[datetime, str]] = None,
+        end: Optional[Union[datetime, str]] = None,
+        application_names: Optional[Union[str, list[str]]] = None,
+    ) -> dict:
+        filter_query = ""
+        now = mlrun.utils.datetime_now()
+        start = start or (now - timedelta(hours=24))
+        end = end or now
```
yaelgen: maybe move this logic to a shared helper since it appears in multiple places, also worth considering to normalize to UTC

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -825,6 +825,163 @@ def count_results_by_status(
...
+            for _, row in df_results.iterrows():
+                metric_list.append(
+                    {
+                        "type": "result",
+                        "time": row[mm_schemas.WriterEvent.END_INFER_TIME],
+                        "name": row[mm_schemas.ResultData.RESULT_NAME],
+                        "kind": row[mm_schemas.ResultData.RESULT_KIND],
+                        "status": row[mm_schemas.ResultData.RESULT_STATUS],
+                        "value": row[mm_schemas.ResultData.RESULT_VALUE],
+                    }
```
yaelgen: The metric/result dicts returned have an implicit schema, maybe we can consider defining a pydantic model or type alias to make the structure explicit?

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -825,6 +825,163 @@ def count_results_by_status(
...
+            )
+            for _, row in df_metrics.iterrows():
+                metric_list.append(
+                    {
+                        "type": "metric",
+                        "time": row[mm_schemas.WriterEvent.END_INFER_TIME],
+                        "name": row[mm_schemas.MetricData.METRIC_NAME],
+                        "value": row[mm_schemas.MetricData.METRIC_VALUE],
+                    }
+                )
```
yaelgen: The two blocks for results and metrics are quite similar, consider extracting to a small helper

yaelgen (changes requested):
preliminary review, added some suggestions

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
Eyal-Danieli: right, so I took an existing helper func named `_get_start_end` which is stored in `mlrun/model_monitoring/db/tsdb/base.py` and moved it to `mlrun.model_monitoring.helpers` along with some adjustments (https://github.com/mlrun/mlrun/pull/8241/commits/965ccb8514a9280804826fd132d280c37a94c461). 
As for the timezone, if the user doesn’t provide `start` or `end`, we default to `mlrun.utils.datetime_min()` and `mlrun.utils.datetime_now()`, which return UTC-normalized datetimes by default.

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -358,6 +358,57 @@ def count_results_by_status(
                  }
         """
 
+    @abstractmethod
+    def count_processed_model_endpoints(
+        self,
+        start: Optional[Union[datetime, str]] = None,
+        end: Optional[Union[datetime, str]] = None,
+        application_names: Optional[Union[str, list[str]]] = None,
```
theSaarco, line 366: Shouldn't this be scoped by a project as well (and have a `project` parameter)? For example, if I'm calling this without any `application_names`, will I get all MEPs from all projects?
Also, are application names unique across projects? Do they include the project name in them? Because if not, then they won't be unique and you have to have a project scope for this.

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -358,6 +358,57 @@ def count_results_by_status(
...
+        :return:                   The count of processed model endpoints.
+        """
+
+    @abstractmethod
+    def calculate_latest_metrics(
+        self,
+        start: Optional[Union[datetime, str]] = None,
+        end: Optional[Union[datetime, str]] = None,
+        application_names: Optional[Union[str, list[str]]] = None,
+    ) -> list[dict]:
```
theSaarco: Same question about project name.

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -358,6 +358,57 @@ def count_results_by_status(
...
+        :param application_names:  A list of application names to filter the results by. If not provided, all
+                                   applications are included.
+        :return:                   A list of dictionaries containing the latest metrics and results for each
+                                   application.
+                                   example::
+                                   [
+                                       {
+                                           "type": "metric",
+                                           "time": "2025-06-29 13:36:37 +00:00",
+                                           "name": "hellinger_mean",
```
theSaarco: Is the name just the name of the metric? If so, then you have to also say what app this came from, otherwise it may not be unique.

theSaarco (changes requested):
Generally look good, I think some of your APIs miss the project scope/parameter.

mlrun/model_monitoring/db/tsdb/base.py
Eyal-Danieli: The connector itself is already in a project scope  https://github.com/mlrun/mlrun/blob/965ccb8514a9280804826fd132d280c37a94c461/mlrun/model_monitoring/db/tsdb/base.py#L48

mlrun/model_monitoring/db/tsdb/base.py
Eyal-Danieli: same answer :)

tests/model_monitoring/test_helpers.py
```diff
@@ -607,3 +608,55 @@ def test_get_output_stream_unsupported() -> None:
...
+    start, end = get_start_end(
+        start=now - datetime.timedelta(seconds=10),
+        end=None,
+        delta=datetime.timedelta(seconds=1),
+    )
+
+    assert start == now - datetime.timedelta(seconds=10)
+    assert end == start + datetime.timedelta(seconds=1)
+
+    # Test when start is bigger than end
```
gtopper:
```suggestion
    # Test when start time is later than end time
```

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -827,10 +843,58 @@ def function_summaries(
...
+        start: typing.Optional[datetime] = None,
+        end: typing.Optional[datetime] = None,
+        include_latest_metrics: bool = False,
+    ) -> mlrun.common.schemas.model_monitoring.FunctionSummary:
+        """
+        Retrieve a single model monitoring function summary by its name.
+        :param name:  The name of the model monitoring function to retrieve.
+        :param start: The start time of the statistics of the monitoring application. If not set, the default is
+                      24 hours ago.
+        :param include_latest_metrics: Whether to include the latest metrics in the response (default is False).
```
gtopper, line 865: I guess the indentation here should be aligned as we do elsewhere.

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -358,6 +358,57 @@ def count_results_by_status(
...
+    def count_processed_model_endpoints(
+        self,
+        start: Optional[Union[datetime, str]] = None,
+        end: Optional[Union[datetime, str]] = None,
+        application_names: Optional[Union[str, list[str]]] = None,
+    ) -> dict[str, int]:
+        """
+        Count the number of processed model endpoints within a given time range for specific applications.
+        :param start:              The start time of the query. Last 24 hours is used by default.
+        :param end:                The end time of the query. Now is used by default.
```
gtopper:
```suggestion
        :param end:                The end time of the query. The current time is used by default.
```

tests/model_monitoring/test_helpers.py
```diff
@@ -607,3 +608,55 @@ def test_get_output_stream_unsupported() -> None:
...
+    with pytest.raises(
+        mlrun.errors.MLRunInvalidArgumentError,
+    ) as exc:
+        get_start_end(
+            start=now + datetime.timedelta(seconds=10),
+            end=now,
+        )
+        assert "The start time must be before the end time" in str(exc.value)
```
gtopper: The assert isn't actually getting run at all the way it's currently written. And if an exception were not raised, you'd get this error:
```
self = <ExceptionInfo for raises contextmanager>

    @property
    def value(self) -> E:
        """The exception value."""
        assert (
>           self._excinfo is not None
        ), ".value can only be used after the context manager exits"
E       AssertionError: .value can only be used after the context manager exits
```
The fix:
```suggestion
    with pytest.raises(
        mlrun.errors.MLRunInvalidArgumentError,
        match="The start time must be before the end time"
    ):
        get_start_end(
            start=now + datetime.timedelta(seconds=10),
            end=now,
        )
```

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -825,13 +825,166 @@ def count_results_by_status(
...
+        filter_query = ""
+        start, end = get_start_end(start=start, end=end, delta=timedelta(hours=24))
+
+        if application_names:
+            filter_query = self._generate_filter_query(
+                filter_column=mm_schemas.WriterEvent.APPLICATION_NAME,
+                filter_values=application_names,
+            )
+
+        def _get_application_endpoints_records(super_table: str) -> pd.DataFrame:
```
gtopper: There's no need to name it with an underscore if it's in local scope.

tests/model_monitoring/db/tsdb/tdengine/test_tdengine_connector.py
```diff
@@ -140,6 +140,22 @@ def test_write_application_event(
...
+    assert processed_model_endpoints == {"my_app": 2}
+
+    # calculate latest metrics
+    latest_metrics = connector.calculate_latest_metrics(
+        start=start_infer_time, end=end_infer_time, application_names="my_app"
+    )
+
+    assert len(latest_metrics) == 2
+    first_metric = latest_metrics[0]
+    assert first_metric["status"] == 2
```
gtopper: This one failed when I ran the test. Mentioning it here for tracking.

mlrun/model_monitoring/db/tsdb/base.py
Eyal-Danieli: added https://github.com/mlrun/mlrun/pull/8241/commits/5ee73f01527b20f57290bc1df90c2fa9ac51cbec#diff-9a03f226df7e3e7b37034e20143348ae1bc3ce0f2f650703280fd83036c01f75R340

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -825,13 +825,166 @@ def count_results_by_status(
...
+            return {}
+        grouped_df = combined_df.groupby(
+            mm_schemas.WriterEvent.APPLICATION_NAME
+        ).count()
+
+        # Convert DataFrame to a dictionary
+        return {
+            app_name: int(row[mm_schemas.WriterEvent.ENDPOINT_ID])
+            for app_name, row in grouped_df.iterrows()
+        }
```
gtopper: Can be simplified
```suggestion
        return grouped_df[mm_schemas.WriterEvent.ENDPOINT_ID].to_dict()
```

tests/model_monitoring/db/tsdb/tdengine/test_tdengine_connector.py
```diff
@@ -140,6 +140,22 @@ def test_write_application_event(
...
+    )
+    assert processed_model_endpoints == {"my_app": 2}
+
+    # calculate latest metrics
+    latest_metrics = connector.calculate_latest_metrics(
+        start=start_infer_time, end=end_infer_time, application_names="my_app"
+    )
+
+    assert len(latest_metrics) == 2
+    first_metric = latest_metrics[0]
```
gtopper, line 155: Any reason not to check the second metric too?

Eyal-Danieli: It didn’t seem very important, but I added it just in case - https://github.com/mlrun/mlrun/pull/8241/commits/796ca69c836f703ff0e44cedd3072ca340430c93

gtopper (approved):
LGTM

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -825,13 +825,187 @@ def count_results_by_status(
...
+        start, end = get_start_end(start=start, end=end, delta=timedelta(hours=24))
+
+        if application_names:
+            filter_query = self._generate_filter_query(
+                filter_column=mm_schemas.WriterEvent.APPLICATION_NAME,
+                filter_values=application_names,
+            )
+
+        def get_latest_metrics_records(
+            type: Literal["metrics", "results"],
```
yaelgen: suggest renaming `type` param to something like `record_type` to avoid shadowing python built-in type

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -825,13 +825,187 @@ def count_results_by_status(
...
+                agg_column = mm_schemas.MetricData.METRIC_VALUE
+
+            return self._get_records(
+                table=table,
+                start=start,
+                end=end,
+                columns=columns,
+                filter_query=filter_query,
+                timestamp_column=mm_schemas.WriterEvent.END_INFER_TIME,
+                group_by=columns[1:],  # without timestamp column
```
yaelgen:
```suggestion
                timestamp_column=mm_schemas.WriterEvent.END_INFER_TIME,
                # Aggregate per application/metric pair regardless of timestamp
                group_by=columns[1:], 
```

yaelgen (changes requested):
Looks great, minor comments

yaelgen (approved):
LGTM

mlrun/model_monitoring/db/tsdb/base.py
```diff
@@ -358,6 +358,60 @@ def count_results_by_status(
...
+        :param end:                The end time of the query. The current time is used by default.
+        :param application_names:  A list of application names to filter the results by. If not provided, all
+                                   applications are included.
+        :return:                   A list containing the latest metrics and results for each application.
+                                   example::
+                                   [
+                                       {
+                                           "type": "metric",
+                                           "time": "2025-06-29 13:36:37 +00:00",
+                                           "application_name": "histogram-data-drift",
```
theSaarco: So we decided to take this out 😄

theSaarco (approved):
LGTM. Just a small leftover to fix the docstring.

assaf758 (approved):

PR #8243, opened 2025-06-30
[DB] Fix return uid value from db as hex

royischoss:
Fix return value from `process_result_value`

### DIFF
--- a/mlrun/db/sql_types.py
+++ b/mlrun/db/sql_types.py
@@ -150,7 +150,7 @@ class UuidType(TypeDecorator):
 
     def process_result_value(
         self, value: Optional[Union[uuid.UUID, bytes, str]], dialect: Dialect
-    ) -> Optional[uuid.UUID]:
+    ) -> Optional[str]:
         if value is None:
             return None
         return value.hex if isinstance(value, uuid.UUID) else value
--- a/mlrun/db/sql_types.py
+++ b/mlrun/db/sql_types.py
@@ -153,7 +153,7 @@ class UuidType(TypeDecorator):
     ) -> Optional[uuid.UUID]:
         if value is None:
             return None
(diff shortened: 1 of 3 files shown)


davesh0812 (approved):
Looks Good!

mlrun/db/sql_types.py
```diff
@@ -153,7 +153,7 @@ def process_result_value(
     ) -> Optional[uuid.UUID]:
```
davesh0812: need to change this to `Optional[str]`

server/py/services/api/tests/unit/db/test_model_endpoint.py
```diff
@@ -213,10 +213,10 @@ def test_batch_insert_and_update(self) -> None:
             self._db_session,
             "project-1",
             {
-                uuid.UUID("5cfeed66-72cc4d978ff9b7b06ebe77f2"): {
+                uuid.UUID("5cfeed6672cc4d978ff9b7b06ebe77f2").hex: {
```
gtopper, line 216: Isn't this just
```suggestion
                "5cfeed6672cc4d978ff9b7b06ebe77f2": {
```
?

server/py/services/api/tests/unit/db/test_model_endpoint.py
```diff
@@ -213,10 +213,10 @@ def test_batch_insert_and_update(self) -> None:
...
                     "monitoring_mode": ModelMonitoringMode.disabled
                 },
-                uuid.UUID("2127986e91f544af9be31250295f03b6"): {
+                uuid.UUID("2127986e91f544af9be31250295f03b6").hex: {
```
gtopper, line 219: And here as well.

davesh0812 (approved):

PR #8244, opened 2025-06-30
[Tests] Following #8183, model name disappears after serialization

gtopper:


tests/serving/test_tracking.py
```diff
@@ -261,6 +261,13 @@ async def predict_async(self, body):
         return self.predict(body)
 
 
+class MyModelSelector(ModelSelector):
+    def select(
+        self, event, available_models: list[Model]
+    ) -> Union[list[str], list[Model]]:
+        return ["MyModel"]
```
royischoss, line 268: This is the class name not the model name.

royischoss, line 268:
```suggestion
        return ["my_model"]
```

gtopper:
Fixed by @royischoss in #8247.

PR #8248, opened 2025-07-01
[Launcher] Skip retried runs that were aborted or deleted

Yacouby:
When a run is submitted to the event loop for retry, it may be aborted (or aborted and deleted) by the user before the retry is actually triggered. In such cases, the retry logic still attempts to launch the run, since the task was already scheduled.
This fix adds a safeguard in the launcher to check if the run was aborted or deleted before launching it. If the run is no longer valid (e.g., aborted), the launcher skips it, ensuring that aborted runs are not retried.

Jira:
https://iguazio.atlassian.net/browse/ML-10355

### DIFF
--- a/server/py/services/api/launcher.py
+++ b/server/py/services/api/launcher.py
@@ -175,11 +175,6 @@ class ServerSideLauncher(launcher.BaseLauncher):
             try:
                 # Skip retried run if it was aborted or deleted
                 if self._should_skip_run(run):
-                    mlrun.utils.logger.info(
-                        "The pending retry run was aborted or deleted, skipping launch",
-                        uid=run.metadata.uid,
-                        project=run.metadata.project,
-                    )
                     run.status.state = mlrun.common.runtimes.constants.RunStates.aborted
 
                 else:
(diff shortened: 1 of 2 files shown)


server/py/services/api/launcher.py
```diff
@@ -173,10 +173,20 @@ def launch(
...
-                    runtime.kind
-                )
-                runtime_handler.run(runtime, run, execution)
+                # Skip retried run if it was aborted or deleted
+                if self._should_skip_run(run):
+                    mlrun.utils.logger.info(
+                        "The pending retry run was aborted or deleted, skipping launch",
+                        uid=run.metadata.uid,
+                        project=run.metadata.project,
+                    )
```
TomerShor: Move the log into `_should_skip_run`, where you know the actual reason for skipping - then log for each case, instead of `aborted OR deleted`.

server/py/services/api/launcher.py
```diff
@@ -173,10 +173,20 @@ def launch(
...
+                    run.status.state = mlrun.common.runtimes.constants.RunStates.aborted
```
TomerShor, line 178: why do you set it to aborted here?
Does it mean it's aborting the retry?

TomerShor (changes requested):

adamdelman (approved):

server/py/services/api/launcher.py
Yacouby: yes, since the retry should not take place in this case

TomerShor (approved):

PR #8249, opened 2025-07-01
[Tests] Fix `test_app_flow` system test with `with_training_set=False`

jond01:
Fixes [ML-10387](https://iguazio.atlassian.net/browse/ML-10387).

Port #8122 to dev and add the required change in `_test_function_summaries`.

I checked, and both

- `test_app_flow[True-False]`
- `test_app_flow[False-False]`

passed (`with_training_set=False` and `with_model_runner` either `True` or `False`).

[ML-10387]: https://iguazio.atlassian.net/browse/ML-10387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/system/model_monitoring/test_app.py
+++ b/tests/system/model_monitoring/test_app.py
@@ -107,7 +107,6 @@ _DefaultDataDriftAppData = _AppData(
 class _V3IORecordsChecker:
     project_name: str
     _logger: Logger
-    apps_data: list[_AppData]
     app_interval: int
     mm_tsdb_profile: DatastoreProfile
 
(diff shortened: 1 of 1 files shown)


davesh0812:
Maybe it's worth to add an assertion to check that the expected amount of monitoring application are deployed?

assaf758 (approved):

jond01:
> Maybe it's worth to add an assertion to check that the expected amount of monitoring application are deployed?

https://github.com/mlrun/mlrun/blob/5cb5543bac08449d2a1e2d3d6479bb2c4a15de8c/tests/system/model_monitoring/test_app.py#L136

https://github.com/mlrun/mlrun/blob/5cb5543bac08449d2a1e2d3d6479bb2c4a15de8c/tests/system/model_monitoring/test_app.py#L268

We have the above assertions for the results and metrics in the TSDB. If the apps aren't deployed, they won't have TSDB records, and the assertions will fail

PR #8256, opened 2025-07-02
[Serving] Enable support for shared models within function graphs

davesh0812:
### This PR introduces:

1. **New SDK Methods**:

   * `RootFlowStep.add_shared_model`: Add a shared model to the graph, making it accessible to all `ModelRunnerStep` instances.
   * `RootFlowStep.set_shared_models_resources`: Configure resource limits (CPU/GPU/memory) for shared models in the graph.
   * `ModelRunnerStep.add_shared_model_proxy`: Add a proxy model to a `ModelRunnerStep`, referencing a shared model already defined in the graph.

2. **Updated SDK**:

   * `ModelRunnerStep.add_model`: Now supports passing an LLM prompt as a model artifact.

3. **MEP Linking**:

   * Model endpoint objects (MEPs) can now reference `LLMPromptArtifact` as their model.

4. **Validation Enhancement** (bug fix):

   * When logging an `LLMPromptArtifact`, the user must now specify either `prompt_string` or `prompt_path`. Providing both or neither will raise an error.

---

**Related Jira Tickets**:

* [ML-9967](https://iguazio.atlassian.net/browse/ML-9967)
* [ML-10250](https://iguazio.atlassian.net/browse/ML-10250)

[ML-9967]: https://iguazio.atlassian.net/browse/ML-9967?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ML-10250]: https://iguazio.atlassian.net/browse/ML-10250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/serving/states.py
```diff
@@ -1218,6 +1243,89 @@ def __init__(
         self.raise_exception = raise_exception
         self.shape = "folder"
 
+    def add_shared_model_proxy(
```
gtopper, line 1315: I still think in terms of naming, it probably makes sense to talk about a shared model, or a model proxy, but not a shared model proxy. Wdyt?

davesh0812: @theSaarco Wdyt? I remember you asked me to use this name..

mlrun/execution.py
```diff
@@ -953,6 +953,11 @@ def log_llm_prompt(
         :returns: The logged `LLMPromptArtifact` object.
         """
 
+        if not prompt_string and not prompt_path:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "Either 'prompt_string' or 'prompt_path' must be provided"
+            )
```
moranbental, line 967: Why not move it to the `LLMPromptArtifact` constructor?

mlrun/projects/project.py
```diff
@@ -1940,6 +1940,11 @@ def log_llm_prompt(
         :returns: The logged `LLMPromptArtifact` object.
         """
 
+        if not prompt_string and not prompt_path:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "Either 'prompt_string' or 'prompt_path' must be provided"
+            )
```
moranbental, line 1948: same

moranbental (changes requested):

mlrun/execution.py
davesh0812: Because it will fail when we trying to initialized it after the serialization with `from_dict`.

mlrun/projects/project.py
davesh0812: same :)

mlrun/serving/states.py
```diff
@@ -528,6 +529,24 @@ def verify_model_runner_step(
             raise GraphError(
                 f"The graph already contains the model endpoints named - {common_endpoints_names}."
             )
+
+        step_shared_model_endpoints_names = [
+            step.class_args.get(schemas.ModelRunnerStepData.MODELS, {})
+            .get(name, ["", {}])[1]
```
royischoss: Maybe adding a constant so the code will self explain we are looking for model params

mlrun/serving/states.py
```diff
@@ -528,6 +529,24 @@ def verify_model_runner_step(
...
+            .get("shared_runnable_name")
```
royischoss: also here better use a const

moranbental (approved):

mlrun/serving/states.py
```diff
@@ -1218,12 +1243,95 @@ def __init__(
...
+            model_artifact.uri
+            if isinstance(model_artifact, mlrun.artifacts.Artifact)
+            else model_artifact
+        )
+        model_class = Model(
+            name=model_artifact_uri,
+            shared_runnable_name=shared_model_name,
+        )
+
+        root = self._extract_root_step()  # todo validation on shared models
```
royischoss: this is a problem checking it here because you might not connected the step to the graph yet

mlrun/serving/states.py
```diff
@@ -1298,9 +1405,16 @@ def add_model(
...
+            model_artifact,
+            ModelArtifact,
         ):
             outputs = [feature.name for feature in model_artifact.spec.outputs]
+        elif outputs is None and isinstance(
+            model_artifact,
+            LLMPromptArtifact,
+        ):
+            _model_artifact = model_artifact.model_artifact
+            outputs = _model_artifact.spec.outputs or []
```
royischoss: outputs = [feature.name for feature in _model_artifact.spec.outputs]

mlrun/serving/states.py
```diff
@@ -1748,7 +1872,8 @@ def init_object(self, context, namespace, mode="sync", reset=False, **extra_kwar
         self._insert_all_error_handlers()
         self.check_and_process_graph()
 
-        for step in self._steps.values():
+        for step in self.steps.values():
+            print(step.name)
```
royischoss: remove print

royischoss (commented):
Hey looks good some small comments

mlrun/serving/states.py
theSaarco: Since we have a "shared model" then this is a proxy to that shared model, which makes this a shared model proxy. I actually think it's a good name, even though it wasn't my proposal 🤓

gtopper: Maybe I've been reading it wrong, as shared (model proxy) rather than (shared model) proxy.

server/py/framework/db/sqldb/db.py
```diff
@@ -6291,7 +6294,9 @@ def _fill_model_endpoint_with_model_data(
                 [tag.name for tag in model_tags] if model_tags else []
             )
             model_artifact_uri = mlrun.datastore.get_store_uri(
-                kind=mlrun.utils.helpers.StorePrefix.Model,
+                kind=mlrun.utils.helpers.StorePrefix.Model
+                if model.kind == mlrun.common.schemas.ArtifactCategories.model
```
theSaarco: Nitpicking - using `ArtifactCategories.model` here is not perfect, since you're comparing a `kind` to a category. Since you have the model at hand you can just check if it's `isinstance(ModelArtifact)` or compare its kind to `mlrun.artifacts.ModelArtifact.kind` - it will be cleaner.

mlrun/serving/states.py
```diff
@@ -1218,12 +1243,90 @@ def __init__(
         self.raise_exception = raise_exception
         self.shape = "folder"
 
+    def add_shared_model_proxy(
+        self,
+        endpoint_name: str,
```
theSaarco, line 1317:
```suggestion
        name: str,
```
I think that using `endpoint_name` is confusing, since it's basically the `name` of the model (or rather, of the proxy). For example, if I don't do MM at all then I won't really have a MEP so I might ask myself why I need to define an endpoint name?

I do understand that this is also the same in `add_model`, and I have the same question about that...

mlrun/serving/states.py
```diff
@@ -1218,12 +1243,90 @@ def __init__(
...
+        shared_model_name: str,
+        model_artifact: Optional[Union[str, ModelArtifact, LLMPromptArtifact]] = None,
+        labels: Optional[Union[list[str], dict[str, str]]] = None,
+        creation_strategy: Optional[
+            schemas.ModelEndpointCreationStrategy
+        ] = schemas.ModelEndpointCreationStrategy.INPLACE,
```
theSaarco:
```suggestion
        model_endpoint_creation_strategy: Optional[
            schemas.ModelEndpointCreationStrategy
        ] = schemas.ModelEndpointCreationStrategy.INPLACE,
```
Since this method deals with adding a model, it should be clear that the creation strategy is not for the model, but for the MEP.

mlrun/serving/states.py
```diff
@@ -1218,12 +1243,90 @@ def __init__(
         self.raise_exception = raise_exception
         self.shape = "folder"
 
+    def add_shared_model_proxy(
+        self,
+        endpoint_name: str,
+        shared_model_name: str,
+        model_artifact: Optional[Union[str, ModelArtifact, LLMPromptArtifact]] = None,
```
theSaarco: This is not according to the design:
1. Per the design, this needs to be a `LLMPromptArtifact`. What's the meaning of having a `ModelArtifact` pointing to another model? The use-case we had in mind was to have the LLM-prompt pointing at the shared model itself while specifying a specific prompt for it.
2. Also, the design actually didn't require a `shared_model_name` at all, but rather that the model within the `model_artifact` will be utilized to identify the shared model it needs to point to. 
If the design cannot be implemented, then please explain the reasoning.

mlrun/serving/states.py
```diff
@@ -1298,9 +1400,16 @@ def add_model(
...
+            model_artifact,
+            ModelArtifact,
         ):
             outputs = [feature.name for feature in model_artifact.spec.outputs]
+        elif outputs is None and isinstance(
+            model_artifact,
+            LLMPromptArtifact,
+        ):
+            _model_artifact = model_artifact.model_artifact
+            outputs = [feature.name for feature in _model_artifact.spec.outputs]
```
theSaarco: This outputs parsing in only done if the user passed an artifact. What if a uri is used? Will the outputs not be calculated? What are the implications?

mlrun/serving/states.py
```diff
@@ -2131,6 +2254,129 @@ def __init__(
...
         self._track_models = False
+        self._shared_models: dict[str, tuple[str, dict]] = {}
+        self._shared_models_mechanism: dict[str, ParallelExecutionMechanisms] = {}
+        self._shared_max_processes = None
+        self._shared_max_threads = None
+        self._pool_factor = None
+
+    def add_shared_model(
+        self,
+        name: str,
```
theSaarco, line 2381: You see, here you use `name` and not `endpoint_name` 😄

tests/system/model_monitoring/test_app.py
```diff
@@ -320,7 +320,7 @@ def _test_api(cls, ep_id: str, apps_data: list[_AppData]) -> None:
 class TestMonitoringAppFlow(TestMLRunSystemModelMonitoring, _V3IORecordsChecker):
     project_name = "test-app-flow"
     # Set image to "<repo>/mlrun:<tag>" for local testing
-    image: typing.Optional[str] = None
+    image: typing.Optional[str] = "artifactory.iguazeng.com:10557/davids/mlrun:1.10.0"
```
theSaarco: Why are you pushing this?

theSaarco (changes requested):
Very nice, but there are places where you didn't follow the design, so need to understand why.

mlrun/serving/states.py
theSaarco: there's the add_shared_model function which adds the models, so users are not likely to confuse since their code will first add the shared models, then add the proxies as part of the graph construction.

mlrun/serving/states.py
davesh0812: We intentionally chose endpoint_name to clearly indicate that this parameter represents the model endpoint name. If I recall correctly, we discussed this in the HLD review and agreed on using this naming for clarity and consistency.

Also, just to clarify — even when model monitoring is disabled, the MEP (Model Endpoint) object will still be created. So the concept of an "endpoint name" still applies regardless of the monitoring state.

That said, if you feel strongly against this naming, we can reconsider — but personally, I think this name is more intuitive and aligns well with our existing conventions.

mlrun/serving/states.py
davesh0812: OK I will change it also on the `add_model` method

mlrun/serving/states.py
davesh0812: Because here it is not an endpoint :)

mlrun/serving/states.py
davesh0812: It will calculate on the pod, by  `MonitoringPreProcessor` using `_get_model_output_schema`... but I also thought to do it here and to get the artifact using it's uri... Do you think it's a good idea?

mlrun/serving/states.py
davesh0812: 1. It does follow the design—while the main use case is LLMs, users can also utilize other GPU-reliant models. To avoid loading the same model multiple times, the shared model mechanism provides a clean and efficient solution.

2. I agree that this part doesn't fully align with the original design, but I thought it would be more intuitive for users. That said, if you prefer, I can revert to the original design. In that case, I’ll set the shared model name to match the model artifact URI.

mlrun/serving/states.py
davesh0812: @GiladShapira94 please advice what do you think?

GiladShapira94: @theSaarco I spoke with @davesh0812, and I agree that we should indicate to the user that this parameter and the value the user provides affect the model_endpoint that has been created by deploying the model, in addition that creating a model endpoint for the deployed model.
Changing it to `name` will indicate something more global and less specific, and I think we should use the `endpoint_name`

mlrun/serving/states.py
```diff
@@ -530,6 +529,65 @@ def verify_model_runner_step(
             raise GraphError(
                 f"The graph already contains the model endpoints named - {common_endpoints_names}."
             )
+
+        # Check if shared models are defined in the graph
+        proxy_endpoint = []
```
royischoss: maybe it better to create a method verifying the shared models because it a large piece of code

mlrun/serving/states.py
```diff
@@ -530,6 +529,65 @@ def verify_model_runner_step(
...
+                step.class_args.get(schemas.ModelRunnerStepData.MODELS, {})
+                .get(name, ["", {}])[schemas.ModelsData.MODEL_PARAMETERS.value]
+                .get("artifact_uri")
+            )
+            prefix, _ = mlrun.datastore.parse_store_uri(model_artifact_uri)
+            if prefix == mlrun.utils.StorePrefix.LLMPrompt:
+                model_artifact, _ = mlrun.datastore.store_manager.get_store_artifact(
+                    model_artifact_uri
+                )
+                model_artifact_uri = model_artifact.model_uri
```
royischoss: add small comment explaining this part taking the model_uri

tests/serving/test_tracking.py
```diff
@@ -606,3 +628,138 @@ def test_sampling_model_runner(sampling_percentage: float):
...
+            ), "expected stream to get one message"
+            assert dummy_stream.event_list[0].get("resp", {}).get("outputs") == [2]
+            assert dummy_stream.event_list[0].get("request", {}).get("inputs") == [1]
+        else:
+            assert len(dummy_stream.event_list) == 0, "expected stream to be empty"
+
+        _test_graph_structure(server.graph, enable_tracking)
+
+
+def test_negative_for_shared_model():
```
royischoss: probably better use failure

tests/serving/test_tracking.py
```diff
@@ -327,127 +327,90 @@ def _test_graph_structure(graph: RootFlowStep, tracked: bool):
...
-        execution_mechanism="naive",
-        endpoint_name="dict_model",
-        input_path="inputs",
-        result_path="outputs",
-        outputs=["o1", "o2", "o3", "o4"],
-        raise_error=False,
-    )
-    graph.to(model_runner_step).respond()
+def test_tracked_model_runner(rundb_mock, enable_tracking: bool):
+    with patch("mlrun.get_run_db", return_value=rundb_mock):
```
royischoss: I think the with statement is redundant

royischoss (commented):
Hey added some comments but overall not bad 🥇

davesh0812:
pass test_app_flow[True, True]

mlrun/serving/states.py
```diff
@@ -1407,8 +1606,14 @@ def _calculate_monitoring_data(self) -> dict[str, dict[str, str]]:
         if isinstance(monitoring_data, dict):
             for model in monitoring_data:
                 monitoring_data[model][schemas.MonitoringData.OUTPUTS] = (
-                    monitoring_data[model][schemas.MonitoringData.OUTPUTS]
-                    or self._get_model_output_schema(model, monitoring_data)
+                    monitoring_data.get(model, {}).get(schemas.MonitoringData.OUTPUTS)
```
royischoss, line 1612: we might want to prevent this call if there is no MODEL_ENDPOINT_UID in monitoring data - this probably means that deploy was not called

mlrun/serving/system_steps.py
```diff
@@ -150,12 +150,16 @@ def _get_data_from_path(
     def do(self, event):
         monitoring_event_list = []
         model_runner_name = event._metadata.get("model_runner_name", "")
-        step = self.server.graph.steps[model_runner_name] if self.server else {}
+        step = self.server.graph.steps[model_runner_name] if self.server else None
+        if not step or not hasattr(step, "monitoring_data"):
```
royischoss, line 154: Nice

tests/serving/test_tracking.py
```diff
@@ -608,6 +605,138 @@ def test_sampling_model_runner(rundb_mock, sampling_percentage: float):
             )
```
royischoss, line 607: We might want to check here or in system test the usage of taking the output schema using the get_model_endpoint

royischoss (commented):
Hey small comments can be added in future PR

royischoss (approved):
Nice 🏆

mlrun/execution.py
```diff
@@ -961,6 +961,11 @@ def log_llm_prompt(
         :returns: The logged `LLMPromptArtifact` object.
         """
 
+        if not prompt_string and not prompt_path:
```
theSaarco, line 964: What if both are provided?

mlrun/serving/states.py
```diff
@@ -530,8 +529,77 @@ def verify_model_runner_step(
...
+        # Update model endpoints names in the root step
         root.update_model_endpoints_names(step_model_endpoints_names)
 
+    @staticmethod
+    def _verify_shared_models(
+        root: "RootFlowStep",
+        step: "ModelRunnerStep",
+        step_model_endpoints_names: list[str],
+    ) -> None:
+        proxy_endpoint = [
```
theSaarco:
```suggestion
        proxy_endpoints = [
```

mlrun/serving/states.py
```diff
@@ -530,8 +529,77 @@ def verify_model_runner_step(
...
+            else:
+                shared_models.append(actual_shared_name)
+
+        undefined_shared_models = list(
+            set(shared_models) - set(root.shared_models.keys())
+        )
+        if undefined_shared_models:
+            raise GraphError(
+                f"The following shared models are not defined in the graph: {undefined_shared_models}."
+            )
```
theSaarco, line 601: It might be useful to also do the reverse check, to see if there are shared models that are not accessed through any proxy. If there are, produce a warning (not error) since these are models that are initialized for nothing and will not be used.

theSaarco (approved):
Looks good. I have some very small comments which will be nice to change.

mlrun/execution.py
davesh0812: Already check inside the `LLMPromptArtifactSpec__init__`

mlrun/serving/states.py
davesh0812: In this stage I adding a single ModelRunnerStep so I don't think its the right place to do so.
If you want I can try to add it in the deployment stage, but I will do it in follow up PR.

royischoss (approved):

PR #6945, opened 2024-12-19
[Track] Updated mlflow version

ZeevRispler:
updated mlflow to 2.16 
https://iguazio.atlassian.net/browse/ML-8711

PR #8264, opened 2025-07-03
[Datastore] Bump pyarrow to 17

alxtkr77:
https://iguazio.atlassian.net/browse/ML-10075


mlrun/datastore/base.py
```diff
@@ -237,7 +237,7 @@ def reader(*args, **kwargs):
                     kwargs,
                 )
                 try:
-                    return df_module.read_parquet(*args, **kwargs)
+                    return df_module.read_parquet(*args, partitioning=None, **kwargs)
```
gtopper: Why disable reading partitioned data, and how does it relate to bumping pyarrow?

requirements.txt
```diff
@@ -13,7 +13,7 @@ pandas>=1.2, <2.2
 # used as a the engine for parquet files by pandas
 # >=10 to resolve https://issues.apache.org/jira/browse/ARROW-16838 bug that is triggered by ingest (ML-3299)
 # <18 to prevent bugs due to major upgrading
-pyarrow>=10.0, <17
+pyarrow>=10.0, <18
```
gtopper, line 16: @ZeevRispler, you [downgraded this](https://github.com/mlrun/mlrun/pull/6945/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R17) on purpose some months ago. What was the reason, and is there still an issue there?

ZeevRispler: There where some issues with 17, https://github.com/apache/arrow/issues/43574, @tomerm-iguazio and I discovered this when working on this ticket: https://iguazio.atlassian.net/browse/ML-8711?isEligibleForUserSurvey=true

gtopper: Thanks @ZeevRispler, the pyarrow issue you reference is still open.

@tomerm-iguazio, given your comment in https://github.com/mlrun/mlrun/pull/6945#issuecomment-2559766419, and given that the pyarrow ticket is still open, I'd think the issue is not resolved, but CI is passing. Do you think we can safely raise the pyarrow version to 17? If so, any idea what changed in the meantime?

assaf758: @gtopper the `Partition=None` is a workaround, mentioned in the py arrow issue [comment](https://github.com/apache/arrow/issues/43574#issuecomment-2332362113).
@alxtkr77 does `Partition=None` pass tests where we use partitioning? I think we have such in feature-store and in model-monitoring (test_app_flow).

gtopper: Seems like setting `partitioning=None` will result in [DirectoryPartitioning](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.DirectoryPartitioning.html) being chosen, at least according to [this page](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.partitioning.html). Since it's a different partitioning scheme than the one we write, I'd be very surprised if it doesn't cause failures.

alxtkr77: As far as I understand, `partitioning=None` only affects the **metadata layer** - specifically whether partition information gets added as additional columns to your result. The core parquet file reading and data processing remains identical.

**What stays the same:**
- All data from the actual parquet files is read correctly
- Data types, schemas, and values are processed identically
- File filtering, column selection, and other operations work normally
- Performance of reading the actual parquet data is the same

**What changes:**
- **With partitioning** (default): You get extra columns like `year=2023`, `month=01` extracted from directory names
- **With `partitioning=None`**: You only get the columns that physically exist in the parquet files

**Example to illustrate:**
```python
# Both of these read the exact same parquet data:
df1 = pq.read_table('data/year=2023/month=01/sales.parquet')
df2 = pq.read_table('data/year=2023/month=01/sales.parquet', partitioning=None)

# If sales.parquet contains columns: [product, revenue, date]
# df1 will have: [product, revenue, date, year, month]  # partition cols added
# df2 will have: [product, revenue, date]              # only file cols
```
It really depends on how we use the metadata. I assume we don't use it, but I can be mistaking

gtopper: We use these columns (year, month, etc.) for filtering, as per
```python
if start_time or end_time or additional_filters:
    partitions_time_attributes = find_partitions(url, file_system)
    set_filters(
        partitions_time_attributes,
        start_time,
        end_time,
        additional_filters,
        kwargs,
    )
    try:
        return df_module.read_parquet(*args, **kwargs)
```
(if you dig into `set_filters()`, you'll find [the relevant code](https://github.com/mlrun/storey/blob/64f76997b2d1538ccdac4be704c16cdeb7c107c1/storey/utils.py#L314))

So I would want to know whether that breaks as a result of setting `partitioning=None`.

gtopper: Btw, what's the error that we get if we don't set `partitioning=None`? I'm pretty sure we don't save the year, month, etc. fields into the parquet files, so I wouldn't expect https://github.com/apache/arrow/issues/43574#issuecomment-2332362113 to be an issue for us at all. Which is why I would like to understand the issue better.

alxtkr77: Good point! Will fix

gtopper: To answer my own question, the test that fails is `test_s3_fs_parquet_as_df`, and it fails with the following error:
```
pyarrow.lib.ArrowTypeError: Unable to merge: Field date has incompatible types: string vs dictionary<values=string, indices=int32, ordered=0>
```
and this still happens when upgrading pyarrow to version 17 without making any other changes.
Full output:
```
___________________________ test_s3_fs_parquet_as_df ___________________________
def test_s3_fs_parquet_as_df():
        data_item = mlrun.datastore.store_manager.object(
            "s3://aws-public-blockchain/v1.0/btc/blocks/date=2023-02-27/"
            "part-00000-7de4c87e-242f-4568-b5d7-aae4cc75e9ad-c000.snappy.parquet"
        )
>       data_item.as_df()

tests/datastore/test_base.py:62: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
mlrun/datastore/base.py:581: in as_df
    df = self._store.as_df(
mlrun/datastore/base.py:345: in as_df
    df = reader(url, **kwargs)
         ^^^^^^^^^^^^^^^^^^^^^
mlrun/datastore/base.py:243: in reader
    return df_module.read_parquet(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/usr/local/lib/python3.11/site-packages/pandas/io/parquet.py:670: in read_parquet
    return impl.read(
/usr/local/lib/python3.11/site-packages/pandas/io/parquet.py:272: in read
    pa_table = self.api.parquet.read_table(
/usr/local/lib/python3.11/site-packages/pyarrow/parquet/core.py:1793: in read_table
    dataset = ParquetDataset(
/usr/local/lib/python3.11/site-packages/pyarrow/parquet/core.py:1371: in __init__
    self._dataset = ds.dataset(path_or_paths, filesystem=filesystem,
/usr/local/lib/python3.11/site-packages/pyarrow/dataset.py:794: in dataset
    return _filesystem_dataset(source, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/usr/local/lib/python3.11/site-packages/pyarrow/dataset.py:486: in _filesystem_dataset
    return factory.finish(schema)
           ^^^^^^^^^^^^^^^^^^^^^^
pyarrow/_dataset.pyx:3089: in pyarrow._dataset.DatasetFactory.finish
    ???
pyarrow/error.pxi:155: in pyarrow.lib.pyarrow_internal_check_status
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   pyarrow.lib.ArrowTypeError: Unable to merge: Field date has incompatible types: string vs dictionary<values=string, indices=int32, ordered=0>
```

gtopper: Based on this, it seems to me that there is no problem whatsoever other than the test. We don't write the partition columns to the parquet files. The only issue is with a test that reads data from `s3://aws-public-blockchain`. Therefore, the obvious solution is to change or delete the test, and upgrade PyArrow with no further changes.

gtopper (changes requested):
Please see my analysis above. I reckon that the best course of action is to:
- fix or delete the offending test
- bump pyarrow
- make no other changes

requirements.txt
alxtkr77: As @gtopper  suggested, I reverted the changes and fixed the S3 test to use a public S3 file without conflicts. Assuming @gtopper is correct that we don't have partitioned data saved with conflicting data columns in our parquet files, this should work fine. I also added a test that covers parquet partitioning scenarios.

tests/datastore/test_base.py
```diff
@@ -264,3 +264,82 @@ def test_as_df_time_filters(start_time_tz, end_time_tz, df_tz):
...
+            assert (
+                row_date <= end_time
+            ), f"Date {row_date} should be before/equal {end_time}"
+
+        # Verify we got data from the right months
+        result_months = set(pd.to_datetime(result["timestamp"]).dt.month)
+        expected_months = {6, 7, 8}  # June, July, August
+        assert (
+            result_months == expected_months
+        ), f"Expected months {expected_months}, got {result_months}"
```
gtopper: Shorter and clearer to just compare the dataframes (+ improved comment a bit):
```suggestion
        # Should include: June 15 & 25 + July 5, 15 & 25 + August 5 & 15 = 7 records
        expected_df = df[
            (df[time_column] > start_time) & (df[time_column] <= end_time)
        ].reset_index(drop=True)

        pd.testing.assert_frame_equal(
            result,
            expected_df,
            check_like=True,
            check_dtype=False,
            check_categorical=False,
        )
```

gtopper (approved):
Minor suggestion. LGTM otherwise ✅

requirements.txt
gtopper, line 16: Btw, we should raise the upper limit to the latest version as soon as we can verify that it doesn't break anything.
(not in this PR of course)

tests/datastore/test_base.py
```diff
@@ -54,11 +54,37 @@ def test_http_fs_parquet_with_params_as_df():
...
+        error_msg = str(e)
+        print("Caught schema conflict with helpful message:")
+        print(error_msg)
+        # For testing, we can assert the message contains key information
+        assert "PYARROW SCHEMA CONFLICT DETECTED" in error_msg
+
+    except Exception as e:
+        # Other errors should still be raised normally
+        print(f"Other error (not schema conflict): {e}")
+        raise
```
gtopper: This is way too long and chatty...
```suggestion
    with pytest.raises(
        mlrun.errors.MLRunValueError,
        match="PYARROW SCHEMA CONFLICT DETECTED"
    ):
        data_item.as_df()
```

gtopper: This is way too long and chatty...
```suggestion
    with pytest.raises(
        mlrun.errors.MLRunValueError,
        match="PYARROW SCHEMA CONFLICT DETECTED"
    ):
        data_item.as_df()
```

mlrun/datastore/base.py
```diff
@@ -36,6 +39,141 @@
...
+
+    except pyarrow.lib.ArrowTypeError as e:
+        error_message = str(e)
+        conflict_info = _detect_partition_data_conflict(error_message)
+        if conflict_info:
+            helpful_message = _create_schema_conflict_message(conflict_info)
+
+            full_message = f"{helpful_message}\n\nOriginal error: {error_message}"
+
+            raise mlrun.errors.MLRunValueError(full_message)
```
gtopper: Probably better set the original error as the cause.
```suggestion
            message = _create_schema_conflict_message(conflict_info)
            raise mlrun.errors.MLRunValueError(message) from e
```

mlrun/datastore/base.py
```diff
@@ -36,6 +39,141 @@
...
+        else:
+            raise
+
+    except Exception:
+        raise
```
gtopper: This doesn't do anything.
```suggestion
```

mlrun/datastore/base.py
```diff
@@ -36,6 +39,141 @@
...
+    # Partition columns are typically string type
+    is_partition_conflict = (
+        (type1 == "string" and "dictionary" in type2)
+        or (type2 == "string" and "dictionary" in type1)
+        or (type1 == "string" and type2 != "string")
+        or (type2 == "string" and type1 != "string")
+    )
+
+    if not is_partition_conflict:
+        return None
```
gtopper: I don't really see what this has to do with dictionary types? If I were to make an assumption regarding the original type, it would be int (e.g. year, month), not dict.

mlrun/datastore/base.py
```diff
@@ -36,6 +39,141 @@
...
+    Returns:
+        dict: Contains conflict information or None if not a conflict
+    """
+    # Only check for conflicts in PyArrow 17+
+    if not is_pyarrow_17_or_later():
+        return None
+
+    # Pattern to match PyArrow schema conflict errors
+    conflict_pattern = (
+        r"Unable to merge: Field (\w+) has incompatible types: (.+?) vs (.+)$"
```
gtopper: This is probably too fragile. Any change to the error message in PyArrow will break it.

mlrun/datastore/base.py
```diff
@@ -36,6 +39,141 @@
...
+    if not is_partition_conflict:
+        return None
+
+    # Determine which is partition vs data type
+    if type1 == "string":
+        partition_type = type1
+        data_type = type2
+    else:
+        partition_type = type2
+        data_type = type1
```
gtopper: We don't know that though. It could easily be the case that one is string, the other is not string, and neither is a partition column.

mlrun/datastore/base.py
```diff
@@ -36,6 +39,141 @@
...
+
+Why this happens:
+Your file path contains partition information (e.g., '{field_name}=value/') which
+creates a partition column with type '{partition_type}'. However, the actual parquet
+ data also contains a '{field_name}' column with type '{data_type}'. PyArrow 17+
+cannot automatically merge these incompatible types.
+
+Solutions:
+
+1. Standardize column types:
```
gtopper: This is not really a solution. The issue is that the data stored in the partition name is inherently a string, but the original type might have been something else (usually an int). So the user can't 'standardize' them. The correct solution would be to remove the column from the parquet files (as explained in the upstream github issue).

mlrun/datastore/base.py
```diff
@@ -36,6 +39,141 @@
...
+
+2. Rename partition columns:
+   Modify your data pipeline to use different names for partition columns to avoid
+   name collisions. For example:
+   - Instead of: date=2023-02-27/file.parquet
+   - Use: partition_date=2023-02-27/file.parquet
+   - Or: dt=2023-02-27/file.parquet
+
+3. Downgrade to PyArrow 16:
+   pip install pyarrow==16.1.0
```
gtopper: No reason to pin it.
```suggestion
   pip install pyarrow~=16.0
```

mlrun/datastore/base.py
```diff
@@ -36,6 +39,141 @@
...
+
+   This restores the previous permissive behavior but may allow silent data corruption.
+
```
```suggestion
```

mlrun/datastore/base.py
```diff
@@ -36,6 +39,141 @@
...
+def _detect_partition_data_conflict(error_message):
+    """
+    Detect if the error is a PyArrow 17+ schema conflict between partition and data columns.
+
+    Args:
+        error_message (str): The error message from PyArrow
+
+    Returns:
+        dict: Contains conflict information or None if not a conflict
```
```suggestion
def _detect_partition_data_conflict(error_message: str) -> Optional[dict[str, str]]:
    """
    Detect if the error is a PyArrow 17+ schema conflict between partition and data columns.

    :param error_message: The error message from PyArrow

    :return: Conflict information dictionary, or None if not a conflict.
```

gtopper (changes requested):
I'm not convinced this approach can reliably distinguish between conflicts resulting from partitioning to other conflicts (due to files simply not having the same schema, unrelated to or even in the absence of partitioning).

tests/datastore/test_base.py
```diff
@@ -264,3 +264,62 @@ def test_as_df_time_filters(start_time_tz, end_time_tz, df_tz):
...
+        # Execute the filter
+        result = data_item.as_df(
+            start_time=start_time,
+            end_time=end_time,
+            time_column=time_column,
+            format="parquet",
+        )
+
+        # Calculate expected results manually
+        # Should include: June 15&25, July 5&15&25, August 5&15 = 7 records
```
gtopper: Minor cleanup part from my previous suggestion that didn't make it in:
```suggestion
        # Should include: June 15 & 25 + July 5, 15 & 25 + August 5 & 15 = 7 records
```

gtopper (approved):
Other than a very minor comment, LGTM ✅

gtopper (approved):

PR #8267, opened 2025-07-06
[Images] Fix ml-base deprecation check to respect image tag

Yacouby:
This PR fixes an issue in the `enrich_image_url` logic where the client version was used for the `mlrun/ml-base` deprecation check, even when the user explicitly provided a tag in the image URL.

Jira:
https://iguazio.atlassian.net/browse/ML-10336

### DIFF
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -907,8 +907,13 @@ def enrich_image_url(
 
     # it's an mlrun image if the repository is mlrun
     is_mlrun_image = image_url.startswith("mlrun/") or "/mlrun/" in image_url
-
+    if ":" in image_url:
+        _, image_tag = image_url.rsplit(":", 1)
+    else:
+        image_tag = None
     if is_mlrun_image and "mlrun/ml-base" in image_url:
+        # use the tag from image URL if available, else fallback to the given tag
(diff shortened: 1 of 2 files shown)


mlrun/utils/helpers.py
```diff
@@ -907,8 +907,13 @@ def enrich_image_url(
 
     # it's an mlrun image if the repository is mlrun
     is_mlrun_image = image_url.startswith("mlrun/") or "/mlrun/" in image_url
-
+    if ":" in image_url:
+        _, image_tag = image_url.rsplit(":", 1)
+    else:
+        image_tag = None
     if is_mlrun_image and "mlrun/ml-base" in image_url:
```
adamdelman, line 914: maybe startswith would be more resilient?

Yacouby: `rsplit` is safe here because it ensures we only split on the last colon — which is where the tag appears.

alonmr (approved):
lgtm

mlrun/utils/helpers.py
Yacouby: not exactly, since in some cases, the image url might include a registry (e.g., `quay.io/mlrun/ml-base`), so a simple startswith check might not be sufficient.

PR #8276, opened 2025-07-07
[Serving] Fix background task  model endpoint creation status check

royischoss:
Fix config file holding int value for `model_endpoint_creation_check_period`

### DIFF
--- a/mlrun/serving/system_steps.py
+++ b/mlrun/serving/system_steps.py
@@ -289,11 +289,6 @@ class BackgroundTaskStatus(storey.MapClass):
                 return None
         elif (
             self._background_task_state
-            == mlrun.common.schemas.BackgroundTaskState.failed
-        ):
-            return None
-        elif (
-            self._background_task_state
             == mlrun.common.schemas.BackgroundTaskState.succeeded
         ):
             return event
--- a/mlrun/serving/system_steps.py
+++ b/mlrun/serving/system_steps.py
@@ -263,17 +263,19 @@ class BackgroundTaskStatus(storey.MapClass):
     def do(self, event):
         if self.server is None:
             return None
+        timestamp = mlrun.utils.now_date()
         if (
             self._background_task_state
             == mlrun.common.schemas.BackgroundTaskState.running
             and (
                 self._background_task_check_timestamp is None
-                or mlrun.utils.now_date() - self._background_task_check_timestamp
+                or timestamp - self._background_task_check_timestamp
                 >= timedelta(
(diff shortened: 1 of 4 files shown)


mlrun/serving/system_steps.py
```diff
@@ -261,8 +261,8 @@ def __init__(self, context, **kwargs):
         super().__init__(**kwargs)
 
     def do(self, event):
-        if (self.context and self.context.is_mock) or self.context is None:
-            return event
+        # if (self.context and self.context.is_mock) or self.context is None:
+        #     return event
```
gtopper: ⚠️

tests/common_fixtures.py
```diff
@@ -762,6 +763,34 @@ def list_model_endpoints(
             endpoints=endpoints
         )
 
+    def get_project_background_task(self, project, task_name):
```
gtopper, line 766: This function doesn't seem to be used anywhere?

royischoss: it part of the rundb we use it inside the `BackgroundTaskStatus:do`

tests/serving/test_tracking.py
```diff
@@ -606,3 +606,33 @@ def test_sampling_model_runner(sampling_percentage: float):
                 f" some outputs with sampling_percentage"
                 f" = {sampling_percentage} "
             )
+
+
+def test_tracked_model_runner_bkg_check(rundb_mock):
```
gtopper: Maybe
```suggestion
def test_tracked_model_runner_background_task(rundb_mock):
```
?

tests/serving/test_tracking.py
```diff
@@ -606,3 +606,33 @@ def test_sampling_model_runner(sampling_percentage: float):
...
+    )
+    rundb_mock._get_background_task_calls = 0
+    graph.to(model_runner_step).respond()
+    function.set_tracking(stream_args={"mock": True})
+    function.set_tracking("dummy://", enable_tracking=True)
+    server = function.to_mock_server()
+    server.test("/", {"n": 1})
+    dummy_stream = server.context.stream.output_stream
+    assert len(dummy_stream.event_list) == 0, "expected stream to be empty"
+    sleep(20)
```
gtopper: Ah, 20 seconds is a very long time for a unit test... Is this really necesary?

gtopper: It works for me without any sleep. Including failing if I undo the fix.

gtopper (approved):

PR #8281, opened 2025-07-07
[Model Monitoring] Block writing out of order data in `evaluate`

jond01:
Implements [ML-10265](https://iguazio.atlassian.net/browse/ML-10265).

New parameter:

- `allow_unordered_data` - see the docstring for an explanation.

Changes:

- If `base_period` is specified, make sure the difference between `end` and `start` is a multiple of `base_period`.
- When writing to the databases, make sure the current `start` time is no earlier than the previous `end` time.
  - Store this data inside the application's schedules file in the `projects` folder in V3IO.

I ran the following system tests (which passed):

- `tests/system/model_monitoring/test_app.py::TestMonitoringAppFlow::test_app_flow[True-True]`
- `tests/system/model_monitoring/test_app.py::TestAppJobModelEndpointData`

[ML-10265]: https://iguazio.atlassian.net/browse/ML-10265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/model_monitoring/applications/base.py
```diff
@@ -616,17 +764,24 @@ def evaluate(
...
-                                  where :math:`m` is some positive integer.
+                                  (m - 1)\\cdot\\operatorname{base\\_period}, \\operatorname{end}]`,
+                                  where :math:`m` is a positive integer and :math:`\\operatorname{end} =
+                                  \\operatorname{start} + m\\cdot\\operatorname{base\\_period}`.
+                                  Please note that the difference between ``end`` and ``start`` must be a multiple of
+                                  ``base_period``.
         :param write_output:      Whether to write the results and metrics to the time-series DB. Can be ``True`` only
                                   if ``endpoints`` are passed.
                                   Note: the model monitoring infrastructure must be up for the writing to work.
+        :param allow_unordered_data: Relevant only when writing outputs to the database. When ``False``, and the
```
GiladShapira94, line 780:
```suggestion
        :param allow_unordered_data: Relevant only when write_output=True. When ``False``, and the
```
i think it will be clearer that you point to the `write_output` flag.

assaf758 (approved):
Very nice!
One comment for later - I think that using evaluate with write_output without ability to delete the database/schedules will be very cumbersome to the user.
Please make sure we have jira on this, and let's see if we can address this during 1.10.

jond01:
> One comment for later - I think that using evaluate with write_output without ability to delete the database/schedules will be very cumbersome to the user. Please make sure we have jira on this, and let's see if we can address this during 1.10.

I agree. [ML-10513](https://iguazio.atlassian.net/browse/ML-10513).

[ML-10513]: https://iguazio.atlassian.net/browse/ML-10513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mlrun/model_monitoring/applications/base.py
jond01: Addressed in #8344

PR #8282, opened 2025-07-07
[Model Monitoring] Fix uses of context in system steps

royischoss:
Fix providing remove context param from steps,  allowing storey to provide it to `storey.Flow::__init__`
`test_app_flow` fails on` _test_function_summaries` for v3io as spoke in data-flow meeting

### DIFF
--- a/mlrun/serving/system_steps.py
+++ b/mlrun/serving/system_steps.py
@@ -378,11 +378,7 @@ class MockStreamPusher(storey.MapClass):
     def __init__(self, output_stream=None, **kwargs):
         super().__init__(**kwargs)
         stream = self.context.stream if self.context else None
-        self.output_stream = output_stream or getattr(stream, "output_stream", None)
-        if self.output_stream is None:
-            raise mlrun.errors.MLRunInvalidArgumentError(
-                "output_stream is None meaning not provided and " "not part of contex"
-            )
+        self.output_stream = output_stream or getattr(stream, "output_stream")
 
     def do(self, event):
         self.output_stream.push(
--- a/mlrun/serving/system_steps.py
+++ b/mlrun/serving/system_steps.py
@@ -35,7 +35,9 @@ class MonitoringPreProcessor(storey.MapClass):
         **kwargs,
     ):
         super().__init__(**kwargs)
-        self.server: mlrun.serving.GraphServer = getattr(self.context, "server", None)
+        self.server: mlrun.serving.GraphServer = (
+            getattr(self.context, "server", None) if self.context else None
+        )
 
     def reconstruct_request_resp_fields(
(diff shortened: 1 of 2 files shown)


mlrun/serving/system_steps.py
```diff
@@ -380,9 +371,10 @@ def _pick_random_requests(num_of_reqs: int, percentage: float) -> list[int]:
 
 
 class MockStreamPusher(storey.MapClass):
-    def __init__(self, context, output_stream=None, **kwargs):
+    def __init__(self, output_stream=None, **kwargs):
         super().__init__(**kwargs)
-        self.output_stream = output_stream or context.stream.output_stream
+        stream = self.context.stream if self.context else None
+        self.output_stream = output_stream or getattr(stream, "output_stream", None)
```
gtopper: If `self.output_stream` does end up being `None`, all events will raise a `AttributeError: 'NoneType' object has no attribute 'push'` error.

royischoss: I will check but I think if it value is None it should work

mlrun/serving/system_steps.py
```diff
@@ -375,6 +379,10 @@ def __init__(self, output_stream=None, **kwargs):
         super().__init__(**kwargs)
         stream = self.context.stream if self.context else None
         self.output_stream = output_stream or getattr(stream, "output_stream", None)
+        if self.output_stream is None:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "output_stream is None meaning not provided and " "not part of contex"
```
gtopper, line 384: Something bad happened to this string. I think it would actually be better to just not use `getattr`, because we'll have a bit more information from the error message (like what type `stream` was).

mlrun/serving/system_steps.py
```diff
@@ -380,9 +375,14 @@ def _pick_random_requests(num_of_reqs: int, percentage: float) -> list[int]:
...
 class MockStreamPusher(storey.MapClass):
-    def __init__(self, context, output_stream=None, **kwargs):
+    def __init__(self, output_stream=None, **kwargs):
         super().__init__(**kwargs)
-        self.output_stream = output_stream or context.stream.output_stream
+        stream = self.context.stream if self.context else None
+        self.output_stream = output_stream or getattr(stream, "output_stream", None)
+        if self.output_stream is None:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "output_stream is None meaning not provided and " "not part of contex"
```
gtopper: Something bad happened to this string. I think it would actually be better to just not use getattr, because we'll have a bit more information from the error message (like what type stream was).

gtopper (approved):

PR #8285, opened 2025-07-08
[Tests] Disable problematic test

alxtkr77:


tests/datastore/test_base.py
```diff
@@ -54,10 +54,14 @@ def test_http_fs_parquet_with_params_as_df():
     data_item.as_df()
 
 
+# ML-10075
```
gtopper, line 57: I don't think this ticket is relevant enough to mention it here. It just happens to be the immediate motivation for raising the PyArrow version upper limit, but the issue would be the same if we had a different motivation (e.g. users want to be able to use a more recent version). Instead, I suggest a TODO.
```suggestion
TODO: find another dataset and re-enable this test
```

gtopper: Or just delete the test.

gtopper (approved):

PR #8291, opened 2025-07-08
[ModelProvider] Add async and custom invoke functionality, support LLModel

tomerm-iguazio:
1) Support async functionality - async_invoke.
2) Support custom invoke sync and async.
3) Added integration and system tests for async.
4) Moved setup functions and constants of model provider tests to shared location (to avoid code duplication).
5) Supported LLModel integration.

[ML-10255](https://iguazio.atlassian.net/browse/ML-10255)

[ML-10255]: https://iguazio.atlassian.net/browse/ML-10255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -103,15 +109,68 @@ def get_client_options(self):
         )
         return self._sanitize_options(res)
 
-    def customized_invoke(
+    def custom_invoke(
```
davesh0812, line 105: please add a test for this method

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -76,19 +78,23 @@ def load_client(self) -> None:
...
         """
         try:
-            from openai import OpenAI  # noqa
+            from openai import OpenAI, AsyncOpenAI  # noqa
 
             self._client = OpenAI(**self.options)
             self._default_operation = self.client.chat.completions.create
+
+            self._async_client = AsyncOpenAI(**self.options)
+            self._default_async_operation = self.async_client.chat.completions.create
```
davesh0812: Why do we need `_default_async_operation` and `_default_operation` if they're only called from a single place?
Unless there's a compelling reason I'm missing, I think having it could be confusing.

mlrun/config.py
```diff
@@ -193,7 +193,7 @@
     },
     "v3io_framesd": "http://framesd:8080",
     "model_providers": {
-        "openai_default_model": "gpt-4",
+        "openai_default_model": "gpt-4o",
```
davesh0812, line 196: I see you change the default... if user will use `gpt-4`... some of the method wont work?

davesh0812 (commented):
Nice!

Another thing I noticed — not all functions have type hints (especially return types, but also parameter types). It would be great to add them here or in a follow-up PR to keep things consistent.

mlrun/datastore/model_provider/openai_provider.py
tomerm-iguazio: we use it as default to all custom invokes (sync+ async)

mlrun/config.py
tomerm-iguazio: `gpt4` will work but they are some models for example that does not support the default operation (for example Whisper that transform speech to text, and does not generate text).

mlrun/datastore/model_provider/openai_provider.py
tomerm-iguazio: I added a test for it, please look at `test_custom_invoke`, There are 2 modes there , sync and async...

royischoss, line 105: I don't understand the need of both custom invoke and invoke in this class, or I expect one will call the other if they do similar things

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -103,15 +109,68 @@ def get_client_options(self):
...
+        :return:                    Same as ModelProvider.custom_invoke.
+
+        """
         invoke_kwargs = self.get_invoke_kwargs(invoke_kwargs)
         if operation:
             return operation(**invoke_kwargs, model=self.model)
         else:
             return self._default_operation(**invoke_kwargs, model=self.model)
 
+    async def async_custom_invoke(
```
royischoss, line 136: same comment as for the `custom_invoke`

mlrun/serving/states.py
```diff
@@ -1203,11 +1203,27 @@ def __init__(self, name: str, **kwargs):
     def predict(
         self, body: Any, messages: list[dict], model_configuration: dict
     ) -> Any:
+        if isinstance(
```
royischoss, line 1206: predict method should be overridden by user I don't think this is the place to add usage of the model_provider, you can create An example in the docs and create another model in tests or demo

mlrun/serving/states.py
```diff
@@ -1203,11 +1203,27 @@ def __init__(self, name: str, **kwargs):
...
+                messages=messages,
+                as_str=True,
+                **(model_configuration or {}),
+            )
         return body
 
     async def predict_async(
         self, body: Any, messages: list[dict], model_configuration: dict
     ) -> Any:
+        if isinstance(
```
royischoss, line 1219: same as above

royischoss (commented):
Hey nice clean work some comments

mlrun/datastore/model_provider/openai_provider.py
tomerm-iguazio: With custom invoke, the user should provide the callable function, it can be image generation for example, that are not supported in invoke, or different task that do not receive  `messages` argument like we do in regular invoke :)

royischoss: I think you can call the custom from the regular invoke, there is some duplication there

royischoss (approved):
Nice 😄

mlrun/datastore/model_provider/openai_provider.py
tomerm-iguazio: Now `custom_invoke` is been called from `invoke`, so no need the save `_default_operation`

So fixed:)

mlrun/datastore/model_provider/model_provider.py
```diff
@@ -122,39 +127,62 @@ def invoke(
...
+
+        The operation must be an async callable (e.g., a method from an async client) that accepts keyword arguments.
+
+        :param operation:       An async callable representing the model operation (e.g., an async_client method).
+        :param invoke_kwargs:   Keyword arguments to pass to the operation.
+        :return:                The full response returned by the awaited operation.
+        """
+        raise NotImplementedError("async_custom_invoke is not implemented")
 
     async def async_invoke(
```
assaf758, line 181: for readability, better have the async variant right after the sync variant.

tests/integration/model_providers/openai/test_openai.py
```diff
@@ -152,7 +137,10 @@ def check_basic_invoke(cls, model_url: str, secrets: dict, model_name: str):
         assert isinstance(response, openai.types.chat.ChatCompletion)
         assert token_count == 50
 
-    def test_basic_invoke(self):
+    @pytest.mark.parametrize("use_datastore_profile", [True, False])
```
assaf758, line 140: Change to   @pytest.mark.parametrize("sync_mode", ["async", "sync"]).
Similarly, cred_mode "datastore_profile" / "env_vars"

tests/integration/model_providers/openai/test_openai.py
```diff
@@ -103,46 +88,46 @@ def reset_env(cls):
...
-        messages = cls._get_messages(prompt)
+        messages = [formatted_messages[0]]
         model_provider = mlrun.get_model_provider(
-            url=model_url, secrets=secrets, default_invoke_kwargs={"max_tokens": 200}
+            url=model_url, secrets=secrets, default_invoke_kwargs={"max_tokens": 100}
         )
         model_provider = cast(OpenAIProvider, model_provider)
         assert model_provider.model == model_name
         result = model_provider.invoke(messages=messages, as_str=True)
-        assert "paris" in result.lower()
```
assaf758, line 141: add `assert isinstance(result, str)`

assaf758 (approved):
Looks good.
Some general comments (not in this pr):
Why do you need T in model-provider? It's not the type in the operation should be the same as in the result.
return type of ingest can be-> Union[str, object] (ie no optional), correct?
I'm approving but please resolve in follow-up pr

tests/integration/model_providers/openai/test_openai.py
tomerm-iguazio: will solved in hf_pr

tests/integration/model_providers/openai/test_openai.py
tomerm-iguazio: Similarly, cred_mode "datastore_profile" / "env_vars" will be fixed in hf PR....
about async and sync, in different PR.

PR #8292, opened 2025-07-09
[Model Monitoring] Add support for `V3IO` in the Function Summary API

Eyal-Danieli:
Following the intro of the function summary api (#8241 ), in this PR we add support for `V3IO` tsdb.
Please note that due to a known limitation with aggregation operations on V3IO, time values for both the latest metric and latest result will not be available.

https://iguazio.atlassian.net/browse/ML-10384

### DIFF
--- a/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
+++ b/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
@@ -804,25 +804,42 @@ class V3IOTSDBConnector(TSDBConnector):
     @staticmethod
     def _get_sql_query(
         *,
-        endpoint_id: str,
         table_path: str,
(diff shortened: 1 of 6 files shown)


mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -835,6 +852,22 @@ def _get_sql_query(
 
                 query.write(")")
 
+            if application_names:
+                if endpoint_id or metric_and_app_names:
```
assaf758: this logic is somewhat fragile, suggest you maintain a `where_added` boolean flag such that you know if to write an AND or a WHERE  (or simlpy check if there is a WHERE in the getvalue())

gtopper (approved):
🏅

PR #8293, opened 2025-07-09
[Model Monitoring] Fix processing ModelRunnerStep outputs

royischoss:
1. Change `MonitoringPreProcessor::transpose_by_key`
2. Added unit tests


mlrun/serving/system_steps.py
```diff
@@ -52,17 +53,8 @@ def reconstruct_request_resp_fields(
...
-            elif len(result) == 1:
-                outputs = (
-                    result[output_schema[0]]
-                    if output_schema
-                    else list(result.values())[0]
-                )
-            else:
-                outputs = []
+            # transpose by key the outputs:
+            outputs = self.transpose_by_key(result, output_schema)
```
davesh0812, line 54: :+1:

mlrun/serving/system_steps.py
```diff
@@ -104,28 +94,84 @@ def reconstruct_request_resp_fields(
...
+            keys = list(data.keys())
+        elif isinstance(schema, str):
+            keys = [schema]
+        else:
+            keys = schema
+
+        values = [data[key] for key in keys]
+
+        # Detect if all are scalars
+        all_scalars = all(not isinstance(v, (list, tuple, np.ndarray)) for v in values)
```
davesh0812, line 152: Can be also a strings right? 
The name is a bit confusing

mlrun/serving/system_steps.py
```diff
@@ -104,28 +94,84 @@ def reconstruct_request_resp_fields(
...
+            transposed = np.array([values])
+        elif all_lists and len(keys) > 1:
+            arrays = [np.array(v) for v in values]
+            mat = np.stack(arrays, axis=0)
+            transposed = mat.T
+        else:
+            return values[0]
+
+        if transposed.shape[1] == 1 and transposed.shape[0] == 1:
+            return transposed[:, 0].tolist()
```
davesh0812: please add a comment, something like : [[0]] -> [0]

davesh0812 (commented):
Added a few comments.
Maybe its worth to add documentation & typing to the predict method, and to demonstrate what we expected it to be.
Also lets make sure all the scenarios are working as expected also for strings.

royischoss:
I don't understand where we can add typing it is inside the body, I added a documentation in https://github.com/mlrun/mlrun/blob/addbceeb553ad634ffeb2abc3c34232dfe8e66cf/mlrun/serving/states.py#L1470

mlrun/serving/states.py
```diff
@@ -1555,6 +1555,38 @@ def add_model(
           :param result_path:         result path inside the user output event, expect scopes to be defined by dot
                                       notation (e.g "outputs.my_model_outputs") expects list or dictionary type object
                                       in path.
+
```
royischoss: @GiladShapira94 Tell me what you think

GiladShapira94: The `body` is the response of the step or the incoming request body?
Can you create an example based on the template you wrote, `outputs.my_model_outputs`.
it is important to understand what the step gets (if relevant) and what the step return and how it would look like

mlrun/serving/states.py
```diff
@@ -1567,11 +1567,27 @@ def add_model(
...
+          :param input_path:          Path inside the user request specify the monitored output. by default will
+                                      monitor the all request. expect scopes to be defined by dot notation
+                                      (e.g "inputs.my_model_inputs").
+                                      Inside the path expected type are list, list of lists, or dictionary-type object.
+                                      If a ``dict`` is provided, each key will represent a feature.
+                                      If a ``list`` or ``list of lists`` is provided, it must follow the order and
+                                      size defined by the input schema.
+                                      Example: If ``input_path = "my inputs"`` and  ``input_schema = ["f0", "f1"]`` and
+                                      ``input request = {"my inputs": [[1, 2], [3, 4]]}``,
+                                      the two monitored inputs will be: ``f1 = [1, 3]``, ``f2 = [2, 4]``.
```
GiladShapira94: 1. Replace `monitored output` with `monitored input`
2. Instead of writing what happens by default, you. could say what happens if the user uses this flag, for example: `selects the key/path in the request body to be used as the monitoring input, expect scopes to be defined by dot notation (e.g.,inputs.my_model_inputs")`.
3. I don't think you need to write information about how the request is built, and instead just give an example for each use case, for example:
```
  1. input_path="my inputs" and  ``input_schema = ["f0", "f1"]``,
   event body : {"my inputs": [[1, 2], [3, 4]]} , Input to be monitored={"f0: [1,3],"f1":[2,4]} 
   2. input_path="my inputs" and  ``input_schema = ["f0", "f1"]``,
   event body : {"my inputs": [1, 2]} , Input to be monitored={"f0: [1],"f1":[2]} 
```
4. here you wrote `input_schema = ["f0", "f1"]` but the example point to `f1` and `f2`
5. Please make sure to mention that when the event key is pointing to a list or a list of lists the user needs to make sure the order of the values matches the input schema
6. Maybe worth explaining that the input schema is the MEP input feature or point to a doc that explains that, for example, to get more details on MM see also - link

davesh0812 (approved):
@GiladShapira94 I will merge this one, and if you will have more comments regarding the docs please lets us know and we will open a follow up PR

PR #7150, opened 2025-01-22
[Model Monitoring] Add parallelism to model monitoring endpoints

davesh0812:
[ML-8960](https://iguazio.atlassian.net/browse/ML-8960), [ML-8290 ](https://iguazio.atlassian.net/browse/ML-8290 )

<html><head></head><body><h3>Summary of Updates:</h3>
<ol>
<li>
<p><strong>Run Model Endpoint Creation in Background Tasks (BGT)</strong>:</p>
<ul>
<li>The model endpoint creation process is now executed in the background to improve efficiency.</li>
</ul>
</li>
<li>
<p><strong>Parallelism for Endpoints</strong>:</p>
<ul>
<li>Enhanced parallelism has been implemented for <code inline="">get</code>, <code inline="">list</code>, <code inline="">delete</code>, and <code inline="">patch</code> operations on model endpoints.</li>
</ul>
</li>
<li>
<p><strong>Monitoring Delay Post-Deployment</strong>:</p>
<ul>
<li>At the end of the deployment process, there might be a brief period where the serving function is operational but not yet monitored.</li>
</ul>
</li>
<li>
<p><strong>Model Endpoint UID Retrieval</strong>:</p>
<ul>
<li>The model endpoint UID is now fetched during the first invocation rather than during initialization. Monitoring starts only after the BGT process completes.</li>
</ul>
</li>
</ol>
<hr>
<h3>BGT Performance Results (Serving with 4000 Model Endpoints):</h3>

Creation Strategy | Start Time | End Time | BGT Time (minutes) | Total Deployment Time (minutes)
-- | -- | -- | -- | --
First-Time (Inplace) | 13:22:05,086 | 13:25:28,051 | 3:20 | 5:09
Inplace | 13:28:11,623, 14:36:39,986 | 13:30:07,943, 14:38:34,344 | 1:56 | 3:50
Archive | 13:33:40,570, 14:51:38,124 | 13:36:57,675, 14:56:46,384 | 3:17, 5:10 | 5:00, 7:00
Overwrite | 13:38:19,588, 14:12:43,660 | 13:44:27,229, 14:18:35,430 | 6:00 | 7:30

<hr>

[ML-8960]: https://iguazio.atlassian.net/browse/ML-8960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8294, opened 2025-07-09
[Model Monitoring] Create MEPs before starting graph job

gtopper:
As part of [ML-10066](https://iguazio.atlassian.net/browse/ML-10066).

Passed (with 505a6b6):
* `TestModelMonitoringOverJob::test_job_from_serving_runtime_with_model_tracking`
* `TestKubejobRuntime::test_job_from_serving_runtime`
* `TestMonitoringAppFlow::test_app_flow (with_training_set=True, with_model_runner=*)`
* `TestNuclioRuntime::test_deploy_function_with_model_runner`

[ML-10066]: https://iguazio.atlassian.net/browse/ML-10066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/py/framework/api/utils.py
```diff
@@ -204,11 +206,87 @@ def _generate_function_and_task_from_submit_run_body(db_session: Session, data):
...
+                runtime=fn,
+                project=project,
+                name=function_name,
+            )
+            fn.spec.model_endpoint_creation_task_name = (
+                model_endpoint_creation_task_name
+            )
+
+            # TODO: there should be a better way to do this
+            serving_spec = getattr(fn.spec, "serving_spec")
```
royischoss, line 251: Can you explain why don't we add it to spec as you did with other parameters? or why do we loss it?

server/py/framework/api/utils.py
```diff
@@ -734,96 +816,79 @@ def submit_run_sync(
...
-            project = run.metadata.project
-            if run:
-                response = run.to_dict()
+            .secrets
+        )
+        param_file_secrets["V3IO_ACCESS_KEY"] = (
+            auth_info.data_session or auth_info.access_key
+        )
 
-    except HTTPException:
```
royischoss, line 816: Why did we loose the try except here? it all related to the rundb?

server/py/services/api/api/endpoints/functions.py
```diff
@@ -214,6 +215,7 @@ async def _filter_functions_by_permissions(_functions):
 @router.post("/build/function/")
 async def build_function(
     request: Request,
+    background_tasks: fastapi.BackgroundTasks,
```
royischoss: is there usage for this here?

mlrun/serving/server.py
```diff
@@ -87,12 +87,6 @@ def __init__(self, enabled: bool, parameters: dict, function_uri: str):
 
             stream_args = parameters.get("stream_args", {})
```
royischoss: I know it dead code but there is a bug regarding to stream while working with mock, I wouldn't delete it part of this PR

server/py/services/api/api/endpoints/nuclio.py
```diff
@@ -267,40 +267,17 @@ async def deploy_function(
...
-            model_endpoints_instructions=model_endpoints_instructions,
-        )
-        returned_background_tasks.background_tasks.append(returned_background_task)
-
-    model_endpoint_creation_task_name = (
-        returned_background_tasks.background_tasks[0].metadata.name
-        if returned_background_tasks.background_tasks
-        else None
+    (
+        function,
```
royischoss, line 271: 🏆

royischoss (commented):
Hey looks good, few comments

server/py/framework/api/utils.py
```diff
@@ -204,11 +206,87 @@ def _generate_function_and_task_from_submit_run_body(db_session: Session, data):
...
+    response = None
+
+    try:
+        fn, task = _generate_function_and_task_from_submit_run_body(db_session, data)
+        run_db = get_run_db_instance(db_session)
+        fn.set_db_connection(run_db)
+
+        track_models = getattr(fn.spec, "track_models", False)
+        logger.info(
+            "Starting model endpoint creation?",
```
theSaarco:
```suggestion
            "Starting model endpoint creation",
```

server/py/framework/api/utils.py
```diff
@@ -204,11 +206,87 @@ def _generate_function_and_task_from_submit_run_body(db_session: Session, data):
...
+                model_endpoint_creation_task_name
+            )
+
+            # TODO: there should be a better way to do this
+            serving_spec = getattr(fn.spec, "serving_spec")
+            if serving_spec:
+                serving_spec = json.loads(serving_spec)
+                serving_spec["model_endpoint_creation_task_name"] = (
+                    model_endpoint_creation_task_name
+                )
```
theSaarco, line 256: Why do we need the `model_endpoint_creation_task_name` in both the runtime and the serving spec?
It would be great if we could remove it from the general runtime spec, since it's not relevant for most runtimes.

server/py/framework/api/utils.py
```diff
@@ -204,11 +206,87 @@ def _generate_function_and_task_from_submit_run_body(db_session: Session, data):
...
+                background_tasks=background_tasks,
+                function=fn.to_dict(),
+                db_session=db_session,
+                is_batch=True,
+            )
+            fn = mlrun.new_function(
+                runtime=fn,
+                project=project,
+                name=function_name,
+            )
```
theSaarco, line 245: Maybe better to first create the function, then start the bg task, just in case the function creation fails for some reason (wrong parameters or whatever).

mlrun/db/httpdb.py
```diff
@@ -3408,6 +3408,7 @@ def _verify_background_task_in_terminal_state():
             logger,
             False,
             _verify_background_task_in_terminal_state,
+            fatal_exceptions=(mlrun.errors.MLRunAccessDeniedError,),
```
theSaarco, line 3411: I get this as a general enhancement to verify we don't retry unrecoverable situations, but what is the situation that you have in mind here? Specific case where this would get access denied due to the changes introduced?

mlrun/serving/server.py
```diff
@@ -547,14 +541,41 @@ async def async_execute_graph(
...
-            exec(f.read(), namespace)
+    code = os.getenv("MLRUN_EXEC_CODE")
+    if code:
+        code = base64.b64decode(code).decode("utf-8")
+        exec(code, namespace)
+    else:
+        source_filename = spec.get("filename", None)
+        if source_filename:
+            with open(source_filename) as f:
+                exec(f.read(), namespace)
```
theSaarco: Why is this needed? IIUC this is to be in line with existing options in a kubejob, but what is the situation where you want to run a serving graph as job and also provide code to execute? Why allow this?

theSaarco (changes requested):
Have some questions on some implementation decision.

server/py/framework/api/utils.py
gtopper: Because it's needed at run time as opposed to deploy time.

server/py/framework/api/utils.py
gtopper: We didn't lose it. It just moved to the caller, `async def submit_run`.

server/py/framework/api/utils.py
gtopper: Ah, that's actually a debug log I forgot to remove.

server/py/framework/api/utils.py
gtopper: The way it's currently implemented, the job needs to wait for the creation task to complete, and therefore needs to know its name. An alternative way to do this would be to have mlrun-api wait for the task to complete before running the job.

server/py/framework/api/utils.py
gtopper: We actually can't do that, because `start_model_endpoint_creation_background_task` modifies the function. There's some refactoring that could be done there, because we actually end up converting the function to a dict and back.

mlrun/db/httpdb.py
gtopper: Right. I added it because it actually happened to me during development, and the error was silently retried for a very long time. The reason I was getting it is my code at the time was trying to start it in the wrong project.

mlrun/serving/server.py
gtopper: There's a TODO elsewhere in the code that applies here, but should be repeated here:
```python
# TODO: find another way to pass this (needed for local run)
```
Currently, this is needed for local runs, because `MLRUN_EXEC_CODE` isn't set in that flow.

gtopper: Do note that we execute this user code either way. The only difference is whether we get it from the `MLRUN_EXEC_CODE` env var that the pod is deployed with, or from the original file (locally).

server/py/framework/api/utils.py
theSaarco: Question is not about that, I understand why we need it. My question is why do we need it in both the function spec and the serving spec. If you have this in the serving spec, then the function spec is redundant, isn't it?

mlrun/serving/server.py
theSaarco: So my question is - what are we running here? When will you have the `MLRUN_EXEC_CODE` set when deploying a serving graph? 
I mean, I get it when you run a generic kubejob which may have the code served this way. But, this is not that situation - this is a serving graph that we run as part of a kubejob, and we shouldn't have this exec code set in that use-case. Or am I missing something? Note that this means that some arbitrary user code will be executed prior to even running the serving graph.

server/py/services/api/api/endpoints/functions.py
gtopper: Will remove.

mlrun/serving/server.py
gtopper: > what are we running here?

The user's code as provided to `code_to_function` or `set_function`.

> When will you have the MLRUN_EXEC_CODE set when deploying a serving graph?

In remote pod (`local=False`), as opposed to a local run (`local=True`). Reading and evaluating `MLRUN_EXEC_CODE` is the same thing we do in a serving function (`v2_serving_init`). Indeed, we must evaluate the user code to be able to run their custom classes / functions, e.g. model implementations.

server/py/framework/api/utils.py
gtopper: It's added to the serving spec from the function spec, as are many other fields:
```python
"parameters": self.spec.parameters,
"graph": self.spec.graph.to_dict(strip=True) if self.spec.graph else {},
"load_mode": self.spec.load_mode,
"functions": function_name_uri_map,
"graph_initializer": self.spec.graph_initializer,
"error_stream": self.spec.error_stream,
"track_models": self.spec.track_models,
"default_content_type": self.spec.default_content_type,
"model_endpoint_creation_task_name": self.spec.model_endpoint_creation_task_name,
# TODO: find another way to pass this (needed for local run)
"filename": getattr(self.spec, "filename", None),
```
it might be possible to avoid, but need to consider the existing flow (deploy of a nuclio function) as well.

theSaarco: Not sure I follow. Why is `model_endpoint_creation_task_name` relevant to a "standard" nuclio function? It is only in use when deploying serving graphs. That's exactly why I don't understand why we need it in any other place other than the serving spec.

mlrun/serving/server.py
theSaarco: But this was never done in `v2_serving_init` - it doesn't have this code. Where is this part done in the nuclio flow?
For kubejob, reading this happens as part of the MLRun `run()` CLI command.
Let's discuss this offline to make sure we're on the same page.

gtopper: You're right. What happens in nuclio is that the user code is put into the nuclio function's user code (the one visible from the UI), via `function.spec.build.functionSourceCode`. So the user's code gets evaluated together with the `init_context` and `handler` functions.

gtopper: And yes, let's discuss when you're available 🙏

mlrun/serving/server.py
jond01: Why dead code?
It is used in the tests. For example:
https://github.com/mlrun/mlrun/blob/2d172ca51163fdec2ae58a2d8d97333a3cc83120/tests/serving/test_tracking.py#L66

And I assume @royischoss refers to [ML-10422](https://iguazio.atlassian.net/browse/ML-10422)

mlrun/serving/server.py
```diff
@@ -87,12 +87,6 @@ def __init__(self, enabled: bool, parameters: dict, function_uri: str):
 
             stream_args = parameters.get("stream_args", {})
 
-            if log_stream == DUMMY_STREAM:
-                # Dummy stream used for testing, see tests/serving/test_serving.py
-                stream_uri = DUMMY_STREAM
-            elif not stream_args.get("mock"):  # if not a mock: `context.is_mock = True`
-                stream_uri = mlrun.model_monitoring.get_stream_path(project=project)
-
```
jond01: Why did you remove this functionality?

jond01 (dismissed):

mlrun/serving/server.py
gtopper: I also thought it was dead code. I'll revert this change once I verify that the tests still pass.

mlrun/serving/server.py
gtopper: I also thought it was dead code. I'll revert this change once I verify that the tests still pass.

server/py/framework/api/utils.py
gtopper: Either way though, this isn't a change in this PR – it was introduced in #7150. Let's discuss it when you're back.

theSaarco (approved):
Approved per our offline discussion.

theSaarco (approved):
Approved per our offline discussion.

PR #8298, opened 2025-07-10
[LLMArtifact] Change prompt_template type, validate prompt legend, adding LLModel class 

royischoss:
Includes:
1. Changing `prompt_string` to `prompt_template` in `LLMPromptArifactSpec` include type change and format.
2. Adding validations to prompt_legend and prompt_template
3. Introducing LLModel as part of the support in new prompt_template type and prompt_legend
4. Fixing link between LLMPromptArtifact to shared model allowing shared model class to have the artifact_uri / ModelArtifact that points to it.
5. bumps storey to 1.10.8
https://iguazio.atlassian.net/browse/ML-10482
https://iguazio.atlassian.net/browse/ML-10363


mlrun/serving/states.py
```diff
@@ -1186,6 +1187,49 @@ def load(self):
         return None, None
 
 
+class LLModel(Model):
+    def __init__(self, name: str, **kwargs):
+        super().__init__(name, **kwargs)
+
+    def predict(self, body: Any, messages: list[dict]) -> Any:
```
davesh0812: lets add also `invoke_kwargs` and pass the `model_configuration` of the LLMPromptArtifact

mlrun/artifacts/llm_prompt.py
```diff
@@ -34,30 +35,41 @@ class LLMPromptArtifactSpec(ArtifactSpec):
...
+        if prompt_path:
+            self._verify_prompt_path(prompt_path)
+        if prompt_template:
+            if not (
+                isinstance(prompt_template, list)
+                and all(isinstance(item, dict) for item in prompt_template)
+            ):
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    "Expected prompt_template to be a list of dicts"
+                )
```
davesh0812: also need to check for role and content keys

mlrun/artifacts/llm_prompt.py
```diff
@@ -34,30 +35,36 @@ class LLMPromptArtifactSpec(ArtifactSpec):
...
             target_path=target_path,
             parent_uri=model_artifact.uri
             if isinstance(model_artifact, model_art.ModelArtifact)
             else model_artifact,
-            body=prompt_string,
             **kwargs,
         )
 
-        self.prompt_string = prompt_string
+        self._prompt_template = None
```
davesh0812: Do you need this?

mlrun/artifacts/llm_prompt.py
```diff
@@ -34,30 +35,36 @@ class LLMPromptArtifactSpec(ArtifactSpec):
...
+        prompt_template: Optional[list[dict]] = None,
         prompt_path: Optional[str] = None,
         prompt_legend: Optional[dict] = None,
         model_configuration: Optional[dict] = None,
         description: Optional[str] = None,
         target_path: Optional[str] = None,
         **kwargs,
     ):
-        if prompt_string and prompt_path:
+        self.PROMPT_TEMPLATE_KEYS = ("content", "role")
```
davesh0812: I think its better to put this before the `__init__` as cls param,
Or just to add schema?

mlrun/artifacts/llm_prompt.py
```diff
@@ -67,10 +74,69 @@ def __init__(
...
+    @staticmethod
+    def _verify_prompt_legend(prompt_legend: dict):
+        if prompt_legend is None:
+            return True
+        for place_holder, body_map in prompt_legend.items():
+            if isinstance(body_map, dict):
+                if body_map.get("field") is None:
+                    body_map["field"] = place_holder
+                body_map["description"] = body_map.get("description")
+                if diff := set(body_map.keys()) - {"field", "description"}:
```
davesh0812: You can add some schema for the legend,
Same for the prompt_template..

mlrun/artifacts/llm_prompt.py
```diff
@@ -137,33 +203,48 @@ def model_artifact(self) -> Optional[model_art.ModelArtifact]:
             return self.spec._model_artifact
         return None
 
-    def read_prompt(self) -> Optional[str]:
+    def read_prompt(self, as_str: bool = False) -> Optional[Union[str, list[dict]]]:
```
davesh0812: I think after all the changes we can avoid `as_str` flag, no?

mlrun/projects/project.py
```diff
@@ -1923,7 +1923,7 @@ def log_llm_prompt(
             )
 
         :param key: Unique key for the prompt artifact.
-        :param prompt_string: Raw prompt text. Mutually exclusive with `prompt_path`.
+        :param prompt_template: Raw prompt list of dict. Mutually exclusive with `prompt_path`.
```
davesh0812: Please add a word regarding OpenAI API (same for the other `log_llm_prompt`

mlrun/serving/states.py
```diff
@@ -1186,6 +1192,47 @@ def load(self):
...
+
+    def run(self, body: Any, origin_name: str, path: str) -> Any:
+        messages = self.enrich_prompt_with_legend(body, origin_name)
+        return self.predict(body, messages)
+
+    async def run_async(self, body: Any, origin_name: str, path: str) -> Any:
+        messages = self.enrich_prompt_with_legend(body, origin_name)
+        return await self.predict_async(body, messages)
+
+    def enrich_prompt_with_legend(self, body: dict, origin_name: str) -> list[dict]:
```
```suggestion
    def enrich_prompt(self, body: dict, origin_name: str) -> list[dict]:
```

davesh0812: please use this function to return also the `model_configuration`

mlrun/projects/project.py
```diff
@@ -1923,7 +1923,7 @@ def log_llm_prompt(
...
         :param prompt_path: Path to a file containing the prompt. Mutually exclusive with `prompt_string`.
         :param prompt_legend: A dictionary where each key is a placeholder in the prompt (e.g., ``{user_name}``)
```
davesh0812, line 1932: please update the docs, also in the other `log_llm_prompt`

mlrun/serving/states.py
```diff
@@ -1125,10 +1128,13 @@ def _load_artifacts(self) -> None:
...
-        if self.artifact_uri:
-            if mlrun.datastore.is_store_uri(self.artifact_uri):
-                artifact, _ = mlrun.store_manager.get_store_artifact(self.artifact_uri)
+    def _get_artifact_object(
+        self, proxy_uri: Optional[str] = None
+    ) -> Union[ModelArtifact, LLMPromptArtifact, None]:
+        uri = proxy_uri or self.artifact_uri
+        if uri:
+            if mlrun.datastore.is_store_uri(uri):
+                artifact, _ = mlrun.store_manager.get_store_artifact(uri)
```
davesh0812, line 1139: lets add some caching for the model artifacts.

mlrun/serving/states.py
```diff
@@ -1186,6 +1192,47 @@ def load(self):
...
+            llm_prompt_artifact_uri = self.shared_proxy_mapping[origin_name]
+            llm_prompt_artifact = self._get_artifact_object(llm_prompt_artifact_uri)
+        else:
+            llm_prompt_artifact = self._get_artifact_object()
+        if not (
+            llm_prompt_artifact and isinstance(llm_prompt_artifact, LLMPromptArtifact)
+        ):
+            raise MLRunInvalidArgumentError(
+                "LLMModel must be provided with LLMPromptArtifact"
+            )
```
davesh0812: I think we should only log a warn and to return None as the prompt

mlrun/serving/states.py
```diff
@@ -2494,7 +2558,21 @@ def init_object(self, context, namespace, mode="sync", reset=False, **extra_kwar
...
+                            monitored_step._shared_proxy_mapping.get(
+                                model_params.get("name"), {}
+                            )
+                        )
+                    else:
+                        model_params["shared_proxy_mapping"] = deepcopy(
+                            monitored_step._shared_proxy_mapping.get(
+                                model_params.get("name"), {}
+                            )
+                        )
```
davesh0812: Why the `deepcopy` is only on the else?

mlrun/serving/states.py
```diff
@@ -1233,12 +1280,13 @@ def select_runnables(self, event):
...
 class MonitoredStep(ABC, TaskStep, StepToDict):
     kind = "monitored"
-    _dict_fields = TaskStep._dict_fields + ["raise_exception"]
+    _dict_fields = TaskStep._dict_fields + ["raise_exception", "_shared_proxy_mapping"]
 
     def __init__(self, *args, name: str, raise_exception=True, **kwargs):
         super().__init__(*args, name=name, **kwargs)
         self.raise_exception = raise_exception
         self._monitoring_data = None
+        self._shared_proxy_mapping = {}
```
davesh0812: Why the monitored step should hold this attribute?
It should be a param of the `ModelRunnerStep`

tests/system/datastore/model_providers/test_openai.py
```diff
@@ -80,7 +80,7 @@ def test_basic_openai_model_runner(self):
         )
         llm_prompt_artifact = self.project.log_llm_prompt(
             "my_llm_prompt",
-            prompt_string=prompt_template,
+            prompt_template=prompt_template,
```
davesh0812, line 86: the `prompt_template` should be edit to the new format

tests/system/datastore/model_providers/assets/models.py
```diff
@@ -40,7 +40,7 @@ def predict(self, body):
     def enrich_prompt(self, body) -> str:
         # TODO: Update this once ML-8172 is completed
         if isinstance(self.invocation_artifact, mlrun.artifacts.LLMPromptArtifact):
-            prompt_template = self.invocation_artifact.spec.prompt_string
+            prompt_template = self.invocation_artifact.spec.prompt_template
             needed_params = ["question", "depth_level", "persona", "tone"]
             sub_dict = {k: body[k] for k in needed_params if k in body}
             return prompt_template.format(**sub_dict)
```
davesh0812: We should change this and to use `LLModel` class, and also to use the built in enrich you just wrote.

tests/artifacts/test_llm_prompts.py
```diff
@@ -88,8 +166,8 @@ def test_prompt_limitation():
 )
 def test_unauthorised_model(project_name_llm):
     project_name_model = "project-test"
-    artifact_path_llm = results_dir / (project_name_llm or "")
-    artifact_path_model = results_dir / project_name_model
+    artifact_path_llm = str(results_dir / (project_name_llm or ""))
+    artifact_path_model = str(results_dir / project_name_model)
```
davesh0812, line 170: Why you change those?

davesh0812 (commented):
Nice.
I added few comments :)

mlrun/artifacts/llm_prompt.py
royischoss: thought it is there thanks 😗

mlrun/serving/states.py
royischoss: because here I start the new dictionary for the update it already copying for a new location, it probably right for the first layer only so I do it too

mlrun/serving/states.py
royischoss: because my method is generic and returns monitored step, I can change it but than I might should change type hints

tests/system/datastore/model_providers/test_openai.py
royischoss: probably missed some of those on it

tests/artifacts/test_llm_prompts.py
royischoss: annoying type hints it returns Path object

mlrun/serving/states.py
davesh0812: I think we should change.
And to check if the monitoring step has `_shared_proxy_mapping` attribute, before using it

royischoss:
`test_app_flow[True, True]`, `test_model_runner_with_llm_and_shared_models` passed

davesh0812 (approved):

PR #8301, opened 2025-07-13
[Artifacts] tiny fix for llm prompt when model_artifact is a str

davesh0812:


moranbental (approved):
Please attach the Jira ticket

tests/serving/test_async_flow.py
```diff
@@ -736,7 +743,7 @@ def test_deploy_function_with_model_runner(
     with unittest.mock.patch(
         "mlrun.store_manager.get_store_artifact",
         side_effect=create_mocked_get_store_artifact(
-            model_artifact=llm_artifact or model_artifact
+            model_artifact=llm_artifact or model_artifact, origin_model=model_artifact
         ),
     ):
         if model_uri:
```
tomerm-iguazio, line 751: I think we should add a test for this fix, can you add a parameter for `test_deploy_function_with_model_runner`?

tomerm-iguazio (changes requested):

tomerm-iguazio (approved):
Nice!

PR #8304, opened 2025-07-14
[Artifacts] Reject deletion of parent artifact if it has children  

davesh0812:
https://iguazio.atlassian.net/browse/ML-10118


server/py/framework/db/sqldb/db.py
```diff
@@ -1267,6 +1262,12 @@ def validate_artifact_removal_preconditions(
                 f"{db_artifact.iteration}, producer_id {db_artifact.producer_id} and {db_artifact.uid} uid. "
                 f"The artifact is used by {dependent_endpoints_count} endpoints"
             )
+        if db_artifact.child_artifacts:
+            raise mlrun.errors.MLRunConflictError(
+                f"Failed deleting artifact {db_artifact.key} in project {db_artifact.project}, iteration "
+                f"{db_artifact.iteration}, producer_id {db_artifact.producer_id} and {db_artifact.uid} uid. "
+                f"The artifact has {len(db_artifact.child_artifacts)} child artifacts, please delete them first"
```
moranbental: We do not use please in error messages.
```suggestion
                f"The artifact has {len(db_artifact.child_artifacts)} child artifacts. Delete them before proceeding."
```

server/py/framework/db/sqldb/db.py
```diff
@@ -1159,10 +1155,9 @@ def del_artifacts(
 
         if failed_deletions_count or failed_deletions_count_integrity:
             if failed_deletions_count_integrity:
-                raise mlrun.errors.MLRunInternalServerError(
+                raise mlrun.errors.MLRunConflictError(
```
moranbental, line 1158: I guess this TODO should be updated as well?https://github.com/mlrun/mlrun/blob/364ada7fe025e86804c5f2975c7c07539649832d/server/py/services/api/crud/artifacts.py#L282-L285

server/py/framework/db/sqldb/db.py
```diff
@@ -5520,9 +5523,12 @@ def _delete_tagged_object(
 
             # get the object id from the object record
             object_id = object_record.id
-
             if cls == ArtifactV2:
                 self._update_artifact_latest_tag_on_deletion(session, object_record)
+                if object_record.child_artifacts:
```
moranbental, line 5527: Why is this check needed?
When deleting a single artifact, the validation is already handled by the `validate_artifact_removal_preconditions` method, and the deletion doesn’t proceed if the conditions aren't met.

In the case of deleting multiple artifacts, we might also need to validate them before starting the deletion process.

Currently, if the check happens after `_update_artifact_latest_tag_on_deletion`, the latest tag is moved to another artifact, and then an error is raised and the original artifact isn’t deleted. As a result, you end up with two artifacts tagged as latest.

This scenario should be covered by a test.

server/py/services/api/tests/unit/api/test_artifacts.py
```diff
@@ -1328,6 +1330,84 @@ def test_failed_to_delete_artifact_with_referenced_model_endpoint(
         "The artifact is used by" in response.text
     ), f"Expected conflict explanation in response, got: {response.text}"
 
+    # Attempt to delete the model artifact that is still referenced by the model endpoint
```
moranbental, line 1333: all your changes are in the DB, so the tests should be in `server/py/services/api/tests/unit/db/test_artifacts.py`

moranbental (changes requested):

moranbental (approved):
LGTM

PR #8305, opened 2025-07-14
[Jobs] Add retries attempts metadata to run status

Yacouby:
Added `retries` list to run status to track metadata about each retry attempt, including the attempt number, start time, end time and error message.

Jira:
https://iguazio.atlassian.net/browse/ML-10169

### DIFF
--- a/server/py/services/api/launcher.py
+++ b/server/py/services/api/launcher.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import base64
+import datetime
 import gzip
 from copy import deepcopy
 from typing import Optional, Union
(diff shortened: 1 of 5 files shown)


server/py/services/api/launcher.py
```diff
@@ -321,7 +322,15 @@ def _handle_retry(run: mlrun.run.RunObject):
...
-        # TODO: Maintain start time of each retry ML-10169
+        # record retry metadata
+        run.status.retries = run.status.retries or []
+        run.status.retries.append(
+            {
+                "attempt": run.status.retry_count,
+                "start_time": datetime.datetime.now(tz=datetime.timezone.utc),
+                "error": run.status.error,
+            }
+        )
```
alonmr: The `"attempt": run.status.retry_count,` needs to be the retry count before the increment at line 324
The start time needs to be the time of the previous attempt which is cleared at line 334
Perhaps we can add end time as well

alonmr (changes requested):

server/py/services/api/launcher.py
```diff
@@ -319,9 +319,21 @@ def _handle_retry(run: mlrun.run.RunObject):
...
+        retry_count = run.status.retry_count or 0
+        start_time = run.status.start_time
+
+        # record retry metadata
+        run.status.retries = run.status.retries or []
+        run.status.retries.append(
+            {
+                "attempt": retry_count,
+                "start_time": start_time,
+                "end_time": mlrun.utils.now_date().isoformat(),
```
alonmr:
```suggestion
                "end_time": run.status.end_time,
```

alonmr (commented):
Minor

alonmr (approved):

PR #8307, opened 2025-07-15
[Tests] Fix git fork extraction

alonmr:
Fixes `test_http_source` which worked locally with ssh remote but doesn't work with http remote url


Yacouby (approved):
LGTM. if this fix is a result of some failing system test, please mention that in the description so we have clear traceability and context for the change.

PR #8310, opened 2025-07-15
[Model Monitoring] Add committed offset and lag stats to `function-summaries` API [V3IO]

Eyal-Danieli:
Add stream statistics to the function summaries APIs. This PR includes the following changes:

- Implementing [Get Stream Shard Lags API](https://docs.nuclio.io/en/latest/reference/api/README.html#get-v3io-stream-shard-lags) in the async nuclio client.
- Adding stream statistics into both the function summaries and function summary APIs. In the function summary API, stats are presented per shard while in the function summaries API, stats are aggregated per function.
- Currently this PR supports only `v3io stream`. Support for `kafka` will be added in a future update.

https://iguazio.atlassian.net/browse/ML-9949


server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -948,6 +968,66 @@ def _get_function_summary_infra(
...
+    ):
+        """
+        Enrich the function with stream stats.
+        :param function_summaries: List of `FunctionSummary` objects to enrich with stream stats.
+        :param agg_stats: If True, aggregate the stream stats by function name.
+        """
+
+        if isinstance(
+            self._stream_profile, mlrun.datastore.datastore_profile.DatastoreProfileV3io
+        ):
```
rokatyy, line 986:
```suggestion
        if not isinstance(
            self._stream_profile, mlrun.datastore.datastore_profile.DatastoreProfileV3io
        ):
           return
```

rokatyy (approved):
LGTM, just a minor suggestion

server/py/services/api/crud/model_monitoring/deployment.py
Eyal-Danieli: It's correct that currently we only support `v3io` but in a few days we’ll add another PR that incldues `kafka` support as well. Therefore, it makes more sense to keep the current structure and later extend it with an `else`, something like:
```
        if isinstance(
            self._stream_profile, mlrun.datastore.datastore_profile.DatastoreProfileV3io
        ):
            # V3IO
            async with framework.utils.clients.async_nuclio.Client(
                self.auth_info
            ) as client: ...
       else:
           # kafka 
            import kafka...
```

rokatyy: @Eyal-Danieli got it 👍

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -948,6 +968,66 @@ def _get_function_summary_infra(
                 )
         return function_summaries_list, base_period
 
+    async def _enrich_with_stream_stats(
+        self,
+        function_summaries: typing.Optional[
+            list[mlrun.common.schemas.model_monitoring.FunctionSummary]
+        ],
+        agg_stats: bool = True,
+    ):
```
assaf758: return type

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -948,6 +968,66 @@ def _get_function_summary_infra(
...
+                        container_name=container,
+                    )
+
+                    stream_stats = stream_stats.get(
+                        f"{container}/{stream_path}", {}
+                    ).get("serving", {})
+                    if stream_stats and agg_stats:
+                        lag = 0
+                        committed = 0
+                        for shard, stats in stream_stats.items():
```
assaf758: shard is unused - can be removed

tests/system/model_monitoring/test_app.py
```diff
@@ -739,8 +739,21 @@ def _test_function_summaries(self) -> None:
...
+
+            evidently_stats = evidently_func_summary.stats
+
+            assert evidently_stats["potential_detection"] == 1
+            assert evidently_stats["detected"] == 0
+
+            # check the stream stats if stream is v3io
+            if isinstance(self.mm_stream_profile, DatastoreProfileV3io):
+                assert evidently_stats["stream_stats"]
+                assert evidently_stats["stream_stats"]["committed"] == 1
```
assaf758, line 751: do you have a test for when you don't aggregate across the shards/partitions?

Eyal-Danieli: yes, the second part of the test which is based on the `function summary API` (and not `function summaries`) - https://github.com/mlrun/mlrun/pull/8310/files/184cdd6f9979ea70e89ea98d71ee3c14250418e3#diff-ffeb62cf7cad099646f03479931c0c5b709ca6697d6fd05836cb18e7d26a3569R794-R809

assaf758 (approved):

PR #8319, opened 2025-07-16
[Workflows] Preventing simultaneous retries

yaelgen:
ticket: [ML-10491](https://iguazio.atlassian.net/browse/ML-10491) 

This PR enforces atomic, idempotent behavior on the retry_pipeline endpoint by:

- Locking the original `workflow‐runner` DB record and marking it `retrying` before scheduling a rerun
- On conflict (another retry attempt), fetching and returning the existing rerun’s `WorkflowResponse` instead of launching a new job
- Removing the “retrying” mark when the rerun completes

These changes prevent duplicate parallel retries and ensure callers receive a consistent response.

[ML-10491]: https://iguazio.atlassian.net/browse/ML-10491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/py/framework/db/sqldb/db.py
```diff
@@ -331,6 +331,42 @@ def update_run(self, session, updates: dict, uid: str, project: str, iter: int =
...
+
+        if retrying:
+            if mlrun_constants.MLRunInternalLabels.retrying in labels:
+                # flush and commit so the lock is released immediately
+                session.commit()
+                raise mlrun.errors.MLRunConflictError
+            else:
+                labels[mlrun_constants.MLRunInternalLabels.retrying] = "true"
+                # optionally bump counter here in follow-up
+                # labels["rerun_counter"] = str(int(labels.get("rerun_counter","0")) + 1)
```
quaark: Is this the plan then? If yes, write this as a TODO. Otherwise, remove the comments

server/py/framework/db/sqldb/db.py
```diff
@@ -331,6 +331,42 @@ def update_run(self, session, updates: dict, uid: str, project: str, iter: int =
...
+        else:
+            labels.pop("retrying", None)
+
+        update_in(struct, "metadata.labels", labels)
```
quaark: use `update_labels` instead as that does some validations and is the convention

server/py/framework/db/sqldb/db.py
```diff
@@ -331,6 +331,42 @@ def update_run(self, session, updates: dict, uid: str, project: str, iter: int =
...
+            if mlrun_constants.MLRunInternalLabels.retrying in labels:
+                # flush and commit so the lock is released immediately
+                session.commit()
+                raise mlrun.errors.MLRunConflictError
+            else:
+                labels[mlrun_constants.MLRunInternalLabels.retrying] = "true"
+                # optionally bump counter here in follow-up
+                # labels["rerun_counter"] = str(int(labels.get("rerun_counter","0")) + 1)
+        else:
+            labels.pop("retrying", None)
```
quaark: Not a completely necessary optimization, however, it does indeed make the decision tree shorter (exit fast) and makes the code nicer with less indentations:
```suggestion
        if not retrying:
            labels.pop("retrying", None)
        elif mlrun_constants.MLRunInternalLabels.retrying in labels:
            # flush and commit so the lock is released immediately
            session.commit()
            raise mlrun.errors.MLRunConflictError
        else:
            labels[mlrun_constants.MLRunInternalLabels.retrying] = "true"
```

server/py/services/api/api/endpoints/pipelines.py
```diff
@@ -209,6 +210,29 @@ async def retry_pipeline(
...
+            services.api.crud.Pipelines().get_running_rerun_runner,
+            db_session=db_session,
+            project=project.metadata.name,
+            original_workflow_id=original_workflow_id,
+        )
+        if workflow_response:
+            return workflow_response
+        raise mlrun.errors.MLRunConflictError(
+            "A retry is already in progress, but no existing rerun was found."
+        )
```
quaark:
```suggestion
    except mlrun.errors.MLRunConflictError as exc:
        workflow_response = await fastapi.concurrency.run_in_threadpool(
            services.api.crud.Pipelines().get_running_rerun_runner,
            db_session=db_session,
            project=project.metadata.name,
            original_workflow_id=original_workflow_id,
        )
        if workflow_response:
            return workflow_response
        raise mlrun.errors.MLRunConflictError(
            "A retry is already in progress, but no existing rerun was found."
        ) from exc
```

server/py/services/api/api/endpoints/workflows.py
```diff
@@ -425,3 +425,61 @@ async def get_workflow_id(
         engine=engine,
         db_session=db_session,
     )
+
+
+@router.post(
+    "/projects/{project}/workflows/{name}/runs/{uid}/set-retry-status",
+    status_code=HTTPStatus.NO_CONTENT.value,
+)
+async def toggle_run_retrying_flag(
```
quaark:
```suggestion
async def set_run_retrying_status(
```

server/py/services/api/crud/pipelines.py
```diff
@@ -293,32 +295,44 @@ def get_original_workflow_run(
         :raises:
             MLRunNotFoundError: If the run ID doesn't correspond to any remote workflow.
         """
-        runner_results = {
-            job_type: services.api.crud.Runs().list_runs(
+        job_type_label = mlrun_constants.MLRunInternalLabels.job_type
+        workflow_id_label = mlrun_constants.MLRunInternalLabels.workflow_id
+
+        def _list_runs(labels: list[str]) -> list[mlrun.model.RunObject]:
+            return services.api.crud.Runs().list_runs(
```
quaark, line 302: I'd make sure this has an order by created at timestamp, to make sure the `_first_or_none` works consistently as expected

server/py/services/api/crud/pipelines.py
```diff
@@ -293,32 +295,44 @@ def get_original_workflow_run(
...
+            runs = _list_runs(labels)
+            return runs.to_objects()[0] if runs else None
+
+        # direct workflow-runner
+        workflow_labels = [
+            f"{workflow_id_label}={run_id}",
+            f"{job_type_label}={mlrun_constants.JOB_TYPE_WORKFLOW_RUNNER}",
+        ]
+        if original := _first_or_none(workflow_labels):
+            return original, original.metadata.labels[workflow_id_label]
```
quaark: This can also be an inline method `_get_original_workflow` that receives a `workflow_id` so that you call it here and then call it again at line 330

server/py/services/api/crud/workflows.py
```diff
@@ -720,3 +723,11 @@ def _prepare_run_object(
         )
 
         return run_object
+
+    def set_run_retrying_state(
```
quaark: sometimes you call it a `retrying_status`, sometimes `retrying_flag` or `retrying_label` and now here `retrying_state`. I don't really mind any of these - but I suggest keeping it consistent. Personally prefer `retrying_status`

quaark (changes requested):
Nice! A few suggestions

server/py/framework/db/sqldb/db.py
yaelgen: yes it is :) 
but without str fun, will be included in follow-up PR

server/py/services/api/crud/pipelines.py
quaark: Think you missed this one

quaark (changes requested):

quaark (approved):

server/py/framework/db/sqldb/db.py
```diff
@@ -331,6 +331,53 @@ def update_run(self, session, updates: dict, uid: str, project: str, iter: int =
...
+            raise mlrun.errors.MLRunNotFoundError(f"Run {project}/{uid} not found")
+
+        struct = run.struct
+        labels = run_labels(struct)
+
+        if not retrying:
+            labels.pop("retrying", None)
+        elif mlrun_constants.MLRunInternalLabels.retrying in labels:
+            # flush and commit so the lock is released immediately
+            session.commit()
```
liranbg: I understand the idea of committing to release the lock but why I see you dont do it on line 359 above. perhaps a mistake? anyway, perhaps I think if the commit needs to happen on any exception being raised, Id wrap everything in try/except and handle the commit upon exception.

server/py/services/api/crud/pipelines.py
```diff
@@ -460,6 +479,45 @@ def rerun_pipeline_via_runner(
...
+        running_rerun_runners = services.api.crud.Runs().list_runs(
+            db_session=db_session,
+            project=project,
+            labels=[
+                f"{mlrun_constants.MLRunInternalLabels.job_type}={mlrun_constants.JOB_TYPE_RERUN_WORKFLOW_RUNNER}",
+                f"{mlrun_constants.MLRunInternalLabels.original_workflow_id}={original_workflow_id}",
+            ],
+            states=[mlrun.common.runtimes.constants.RunStates.running],
+        )
+        if running_rerun_runners:
```
liranbg, line 512: if not? raise not found?

liranbg (changes requested):
very well. minor suggestions/questions

liranbg (approved):

PR #8327, opened 2025-07-17
[DB] Ignore “nil/none” config inputs for DB modes

adamdelman:
## Changes
- Add `_EMPTY_DB_CONFIGURATIONS = {"nil", "none"}`.
- `set_configurations()` returns early if input is empty or contains sentinel.

## Test Coverage
-  Unit: `test_set_configurations_skips_when_nil_or_none`

https://iguazio.atlassian.net/browse/ML-10582


server/py/framework/utils/db/utils.py
```diff
@@ -171,13 +172,16 @@ def set_configurations(
         self,
         config_items: Optional[Union[list[str], dict[str, Any]]] = None,
     ) -> None:
-        config_items = config_items or self._DEFAULT_DB_CONFIGURATIONS
-        if not config_items:
+        items = config_items or self._DEFAULT_DB_CONFIGURATIONS
+        keys = _to_keyset(items)
+
+        if not keys or keys.intersection(self._EMPTY_DB_CONFIGURATIONS):
```
moranbental, line 178: In Provazio, for RDS, we set MLRUN_HTTPDB__DB__MYSQL__MODES to "nil". 
That means `_DEFAULT_DB_CONFIGURATIONS` is "nil" in this case. So I think I didn’t fully understand how your solution addresses the issue

adamdelman: We previously accepted both "nil" and "none" as empty values, so I used None for _DEFAULT_DB_CONFIGURATIONS to be more correct, and also allow using "null" which also deserializes to None.
This handling was removed when I refactored the DBUtil code, so I am adding it back.

liranbg (approved):

PR #8334, opened 2025-07-20
[Model Monitoring] MM API to get metrics for multiple MEPs has bad param name `events_format` and doesn't validate MEPs

danielperezz:
https://iguazio.atlassian.net/browse/ML-9758?linkSource=email

1. Fix `get_metrics_by_multiple_endpoints` to raise an error if any provided model-endpoint IDs do not exist (instead of silently ignoring non-existent IDs).
2. In case of `events_format` of type 'intersection', currently MEPs without metrics are ignored, fix to return empty result in such case


server/py/services/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -1207,6 +1207,16 @@ def get_model_endpoints_metrics(
                 df=df, type=type, project=project
             )
         elif metrics_format == mm_constants.GetEventsFormat.INTERSECTION:
```
Eyal-Danieli, line 1214: I know it's not related to your PR but in the docstring it says about `metrics_format`:
`:param metrics_format:  Determines the format of the result. Can be either 'list' or 'dict'.`

could you please fix to:
```:param metrics_format:  Determines the format of the result, which can be `single`, `separation`, or `intersection`.```

Eyal-Danieli (commented):
LGTM,
a single comment

Eyal-Danieli (approved):

PR #8336, opened 2025-07-20
[Serving] Fix `Model:predict` and `LLModel:predict`

royischoss:
Allow kwargs in the different model classes predict method


mlrun/serving/states.py
```diff
@@ -1200,15 +1200,15 @@ class LLModel(Model):
...
         super().__init__(name, **kwargs)
 
-    def predict(
-        self, body: Any, messages: list[dict], model_configuration: dict
-    ) -> Any:
-        return body
+    def predict(self, body: Any, **kwargs) -> Any:
+        messages: list[dict] = kwargs.get("messages", [])
+        model_configuration: dict = kwargs.get("model_configuration", {})
+        return body, messages, model_configuration
```
tomerm-iguazio: Predict should not return dict object?

mlrun/serving/states.py
```diff
@@ -1200,15 +1200,15 @@ class LLModel(Model):
...
 
-    async def predict_async(
-        self, body: Any, messages: list[dict], model_configuration: dict
-    ) -> Any:
-        return body
+    async def predict_async(self, body: Any, **kwargs) -> Any:
+        messages: list[dict] = kwargs.get("messages", [])
+        model_configuration: dict = kwargs.get("model_configuration", {})
+        return body, messages, model_configuration
```
tomerm-iguazio: Same comment as before

tomerm-iguazio (changes requested):
looks nice, 2 comments:)

tomerm-iguazio (approved):
Nice

PR #8338, opened 2025-07-20
[Model Monitoring] Fix project deletion

jond01:
Previously, we attempted to delete the user-space schedules folder from MLRun service, which doesn't have the
permissions to do it.
Instead, delete it with the SDK in user code.

This fixed the following error:
https://github.com/mlrun/mlrun/actions/runs/16383691421/job/46299860984#step:9:4254
```
FAILED tests/system/model_monitoring/test_app.py::TestAppJobModelEndpointData::test_count_app[True-True] - OSError: error: cannot get deploy status, details: MLRunNotFoundError('Failed to get function mm-job-mep-data-model-server deploy status'), caused by: 404 Client Error: Not Found for url: https://mlrun-api.default-tenant.app.mlrun-ci-test-3-6-0.iguazio-cd1.com/api/v1/projects/mm-job-mep-data/nuclio/model-server/deploy?name=model-server&project=mm-job-mep-data&tag=latest&last_log_timestamp=0&verbose=no
```
This error is caused by an improper deletion of the project.

To be included in the documentation.

### DIFF
--- a/mlrun/model_monitoring/__init__.py
+++ b/mlrun/model_monitoring/__init__.py
@@ -15,4 +15,5 @@
 from mlrun.common.schemas import ModelEndpoint, ModelEndpointList
 
 from .db import get_tsdb_connector
+from .db._schedules import delete_model_monitoring_schedules_user_folder
 from .helpers import get_stream_path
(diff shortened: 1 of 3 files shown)


assaf758:
So with the new approach, delete_project will leave the schedules as is?

jond01:
> So with the new approach, delete_project will leave the schedules as is?

Right. We can discuss the desired approach as a part of [ML-10164](https://iguazio.atlassian.net/browse/ML-10164)

[ML-10164]: https://iguazio.atlassian.net/browse/ML-10164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

assaf758 (approved):

PR #8339, opened 2025-07-21
[DB] Remove legacy artifacts table

alonmr:
Removing the legacy artifacts table means we don't support jumping from mlrun 1.6.x to 1.10
Users will have to move to any version in between to perform the required data migrations

https://iguazio.atlassian.net/browse/ML-10061


moranbental (changes requested):
You forgot to delete all the places that contain `# TODO: Remove once data migration v5 is obsolete` and all the places that call the `is_legacy_artifact` function

moranbental (changes requested):
What about adding a condition https://github.com/mlrun/mlrun/blob/1264f679fbd19251cf35a3eb96dbdfcd76086cbe/server/py/services/api/initial_data.py#L283-L286

server/py/services/api/migrations/versions/6e8e4df16a4e_remove_legacy_artifacts_table.py
```diff
@@ -0,0 +1,141 @@
...
+    op.drop_table("artifacts")
+    op.drop_index("_artifacts_labels_uc", table_name="artifacts_labels")
+    op.drop_index("idx_artifacts_labels_name_value", table_name="artifacts_labels")
+    op.drop_table("artifacts_labels")
+    op.drop_index("_artifacts_tags_uc", table_name="artifacts_tags")
+    op.drop_table("artifacts_tags")
+    # ### end Alembic commands ###
+
+
+def downgrade():
```
yaelgen, line 46: could be helpful to split logic into smaller helper methods per table to improve readability, not critical at all but could make it easier to follow.

yaelgen (approved):
LGTM

moranbental (approved):
LGTM

liranbg (approved):

PR #8344, opened 2025-07-21
[Model Monitoring] Add evaluate naming options and checks

jond01:
* Change `allow_unordered_data` --> NOT `fail_on_overlap`, as it better describes the behavior.
* Check MM apps do not end with `"-batch"` - resolves [ML-10513](https://iguazio.atlassian.net/browse/ML-10513).
* Support the function name as the `application_name` in evaluate, and add `"-batch"` suffix.

The system tests that I checked and passed:

- `tests/system/model_monitoring/test_app.py::TestAppJobModelEndpointData`
- `"tests/system/model_monitoring/test_app.py::TestMonitoringAppFlow::test_app_flow[True-True]"`

The BC change is due to the FQN regex change. The result name pattern was changed from `[a-zA-Z0-9_-]+` to the more accurate and constrained `[a-zA-Z_][a-zA-Z0-9_]*`.

[ML-10513]: https://iguazio.atlassian.net/browse/ML-10513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/model_monitoring/applications/base.py
```diff
@@ -271,7 +272,12 @@ def _handler(
                 "working with endpoints, without any custom data-frame input"
             )
 
-        application_name = self.__class__.__name__
+        if not application_name:
+            application_name = self.__class__.__name__
```
davesh0812: I think we should normalize it and to make sure the user want get any warning.
```suggestion
            application_name = mlrun.utils.normalize_name(self.__class__.__name__, False)
```

davesh0812: Also I think we can add some log to inform the user what is the name no?

mlrun/model_monitoring/applications/base.py
```diff
@@ -777,16 +785,18 @@ def evaluate(
...
+                                  wrote to the database - an error is raised.
+                                  If ``False``, when the previously described situation occurs, the relevant time
+                                  window is cut so that it starts at the earliest possible time after ``start``.
         :param stream_profile:    The stream datastore profile. It should be provided only when running locally and
                                   writing the outputs to the database (i.e., when both ``run_local`` and
                                   ``write_output`` are set to ``True``).
                                   For more details on configuring the stream profile, see
                                   :py:meth:`~mlrun.projects.MlrunProject.set_model_monitoring_credentials`.
+        :param application_name:  The application name to use for the results. If not set, the class name is used.
+                                  A ``"-batch"`` suffix is guaranteed to be added if not already there.
```
davesh0812: Why we don't use the function_name as we use in the realtime case?

davesh0812 (commented):
Nice!
Few comments and questions.

mlrun/model_monitoring/applications/base.py
jond01: Addressed - check via the regex pattern

jond01: Added info log in the `write_output=True` case. Otherwise, the `application_name` is not very relevant

mlrun/model_monitoring/applications/base.py
jond01: Done and removed `application_name`

davesh0812 (approved):

PR #8346, opened 2025-07-21
[DB] Fix partition bootstrapping for Postgres

adamdelman:
## Add cross-dialect support for partition key via triggers

This PR introduces explicit, trigger-based handling of partition key computation for `alert_activations`, supporting both MySQL and PostgreSQL in a consistent, portable way.

### Highlights

- Added a new `partition_key` column to `alert_activations`, required for range partitioning.
- The column is computed from `activation_time` using dialect-specific expressions:
  - **MySQL**: inline expression (e.g. `YEAR(...) * 100 + MONTH(...)`)
  - **PostgreSQL**: user-defined immutable SQL functions (e.g. `iso_yearweek_int(...)`)
- Partition key is updated via triggers:
  - MySQL: `BEFORE INSERT/UPDATE` triggers with inline expressions
  - PostgreSQL: `BEFORE INSERT/UPDATE` trigger calling the SQL function
- Partition key is included in the **primary key** to comply with both engines’ partitioning constraints.
- Created dialect-specific `RangePartitioner` bootstrappers for creating the initial partition.
- PostgreSQL setup creates required SQL functions and ensures `utf8_bin` collation exists.
- Updated Alembic migration to add the `partition_key` column.
- Improved dialect detection logic in backup utilities and general database logic.

#### Why triggers?

Generated columns are not portable:

- **PostgreSQL** requires expressions to be based on `IMMUTABLE` functions and does not allow calling user-defined functions in generated columns easily.
- **MySQL** allows only pure expressions (no user-defined functions) in generated columns used for partitioning, and requires `STORED`, not `VIRTUAL`.

Triggers are:

- Dialect-compatible
- Explicit and evolvable
- Able to apply logic on both insert and update

#### Why include `partition_key` in the primary key?

Partitioning columns must be part of the primary key or all unique constraints in PostgreSQL. 
Including `partition_key` ensures:

- Valid schema for partitioned tables
- Engine enforcement of uniqueness across all partitions

#### How `partition_key` is set in MySQL

Two triggers:

- `BEFORE INSERT`: computes `partition_key` from `NEW.activation_time`
- `BEFORE UPDATE`: re-computes it on update

Example:

```sql
SET NEW.partition_key = YEAR(NEW.activation_time) * 100 + MONTH(NEW.activation_time);
```
Merge after https://github.com/mlrun/mlrun/pull/8351 is merged

https://iguazio.atlassian.net/browse/ML-8545


server/py/framework/db/sqldb/partititioner.py
```diff
@@ -93,16 +93,13 @@ def bootstrap(
...
-            )
-        )
+
         session.execute(
             sqlalchemy.text(
                 f"""CREATE TABLE {quoted_partition_name}
-                     PARTITION OF {quoted_table_name}
-                     FOR VALUES FROM (MINVALUE) TO ({int(first_partition_upper_bound)})"""
+                       PARTITION OF {quoted_table_name}
+                       FOR VALUES FROM (MINVALUE) TO ({int(first_partition_upper_bound)})"""
```
rokatyy, line 98: I assume that this will only work for psql, not for mysql. Did you check it with real mysql test?

server/py/framework/db/sqldb/partititioner.py
```diff
@@ -82,9 +82,9 @@ def bootstrap(
         self,
         session: sqlalchemy.orm.Session,
         table_name: str,
-        partition_expression: str,
         first_partition_name: str,
         first_partition_upper_bound: str,
+        partition_expression: str,
```
rokatyy: why changing the order here?

server/py/framework/db/sqldb/models.py
```diff
@@ -1138,20 +1253,32 @@ def bootstrap_partitions(
...
+    target=Base.metadata,  # run before tables are created
+    identifier="before_create",
+    relevant_dialects=[mlrun.common.db.dialects.Dialects.POSTGRESQL],
+)
+def _pg_create_utf8_bin(_, connection, **kw):
+    connection.execute(
+        text(
+            "CREATE COLLATION IF NOT EXISTS utf8_bin "
+            "(provider = 'libc', locale = 'C', deterministic = true)"
         )
```
rokatyy: could you please add comment of why this is required?

server/py/framework/db/sqldb/models.py
```diff
@@ -1118,6 +1127,112 @@ def _sqlite_autoincrement(
...
+        mlrun.common.db.dialects.Dialects.POSTGRESQL,
+    ],
+)
+def create_postgres_partition_trigger(target: Table, connection, **_):
+    connection.exec_driver_sql(
+        f"""
+        CREATE OR REPLACE FUNCTION {target.name}_set_partition_key()
+        RETURNS trigger AS $$
+        BEGIN
+          NEW.partition_key := {AlertActivation._postgres_partition_expr}(NEW.activation_time);
```
rokatyy: One of the main alert activation feature requirements was good performance taking into account that we might have huge amount of alert activations happening at the same time. afaik triggers slow down performance and in this specific case might become a real bottleneck for inserts. In this case it can be calculated locally quite easily, so I don't see a real value of introducing them.

rokatyy (changes requested):
I'm generally not sure regarding triggers idea because of performance considerations. Regarding adding `partition_key` - I do understand that it's psql requirement to have it in db, but if adding it, let's make sure that we added partition_key filter to all the places where we get the data from the table to ensure the best performance. 


Also, I noticed that you added  the `partition_key` as `nullable=False` - what will happen with the user records from the previous version in this case? Does it require some additional migration?

server/py/framework/db/sqldb/models.py
adamdelman: I'm not sure what you mean by computed locally.

The problem is since we do partitioning not on the raw activation time, but on something extracted from it.
In MySQL, you have builtin expressions like `YEARWEEK`, which can be used for partitioning.
in Postgres, these do not exist ,and have to be substituted with function calls, which are not immutable as so cannot be used as partitioning keys, and there for have to be saved using a trigger into the partition_key column.
Also, we computed columns also cannot be used as partition keys in postgres.
Lastly, triggers in postgres are more performant than in mysql, since mysql does not support parallel execution of triggers.

rokatyy: @adamdelman why `NEW.partition_key := {AlertActivation._postgres_partition_expr}(NEW.activation_time);` this can't be computed on python side rather than doing this on db side? 

> In MySQL, you have builtin expressions like YEARWEEK, which can be used for partitioning.
in Postgres, these do not exist ,and have to be substituted with function calls, which are not immutable

but partition_key is the number that defines the partition, isn't it? Like a week number or year+day and etc. So we calculate this already manually for mysql, why we can't do the same for psql?

> Lastly, triggers in postgres are more performant than in mysql, since mysql does not support parallel execution of triggers.
So what is the point of having it in mysql then? If we already calculate partition_key locally in python?

PR #8347, opened 2025-07-21
[DB] Partitioning refactor: Add partition_key column, unify logic across MySQL and Postgres

adamdelman:
### 📝 Description

This PR refactors and enhances the database partitioning system for alert activations to improve cross-database compatibility, robustness, and maintainability. The changes establish a cleaner architecture for partition management while adding full PostgreSQL support alongside the existing MySQL implementation.

**Key improvements:**
- **Cross-database support**: Added native PostgreSQL partitioning alongside MySQL, with SQLite gracefully skipping partition operations
- **Explicit partition keys**: Introduced a dedicated `partition_key` integer column computed from `activation_time`, enabling consistent partitioning logic across databases
- **Partition metadata tracking**: New `table_partition_interval` table persists partition configuration, decoupling interval choice from table structure
- **Improved bootstrapping**: Separated partition creation logic into dedicated bootstrapper classes per database dialect
- **Enhanced testing**: Comprehensive integration tests for both MySQL and PostgreSQL partition operations

---

### 🛠️ Changes Made

**Database Schema:**
- Added `partition_key` column to `alert_activations` table (computed from `activation_time`)
- Created `table_partition_interval` metadata table to track partition configuration per table
- Updated primary key constraint to include `partition_key` for MySQL
- Added migration `c25e56faecce` to handle schema updates and backfill existing data

**Partitioning Logic:**
- Refactored `PartitionInterval` from `partition.py` → `partition_interval.py` with cleaner APIs
- Renamed `partititioner.py` → `partition_bootstrapper.py` with dialect-specific implementations
- Split partition creation logic into `PartitionBootstrapperMySQL`, `PartitionBootstrapperPostgres`, and `PartitionBootstrapperSqlite`
- Introduced `get_partition_key_value()` and `get_mysql_partition_key_sql()` methods for consistent key generation
- Removed database-specific partition expression queries in favor of stored metadata

**Database Layer:**
- Added `get_partition_interval_for_table()` and `set_partition_interval_for_table()` to base `DBInterface`
- Implemented dialect-specific `create_partitions()` and `drop_partitions()` methods
- Improved `SQLDB.__new__()` to properly route SQLite connections
- Added custom `now()` function with PostgreSQL-specific compilation for timestamp handling

**Testing & CI:**
- Reorganized test structure with shared fixtures in `tests/conftest.py`
- Added integration tests for partition operations on both MySQL and PostgreSQL
- Updated CI workflow to run integration tests against both database backends
- Improved coverage combination logic to handle multiple database test runs

**Code Quality:**
- Better separation of concerns between bootstrapping and ongoing partition management
- Reduced coupling between partition logic and database schema
- More robust error handling and validation
- Cleaner imports and reduced circular dependencies

---

### ✅ Checklist

---

### 🧪 Testing

**Integration Tests:**
- `test_partition_interval_yearweek_matches_mysql_yearweek_mode_1`: Make sure the Python YEARWEEK partition key calculation matches MySQL's calaulcation
- `test_insert_populates_partition_key`: Validates automatic partition key computation on insert
- `test_create_partitions_mysql/postgres`: Tests partition creation with different intervals (DAY, MONTH, YEARWEEK)
- `test_drop_partitions_mysql/postgres`: Validates retention-based partition cleanup
- Tests run against real MySQL 8.0 and PostgreSQL 17 containers via pytest-mock-resources

**Unit Tests:**
- Enhanced `test_partitioner.py` with parametrized tests for interval calculations
- Tests for partition name generation, boundary computation, and interval math
- Validation of environment variable parsing and metadata storage

**Manual Testing:**
- Verified migration runs cleanly on both MySQL and PostgreSQL
- Confirmed existing alert activations remain accessible after migration
- Tested partition creation/deletion during normal API operations

**CI Updates:**
- Integration tests now run against both `mysql` and `postgres` backends
- Coverage reports correctly combine results from multi-database test runs
- Migration tests validate against both database engines

---

### 🔗 References
- Ticket link: [ML-11553](https://iguazio.atlassian.net/browse/ML-11553)
- Related: Original partitioning implementation in `650f0ce2da6f_add_alert_activation.py`

---

### 🚨 Breaking Changes?

- [x] Yes (explain below)
- [ ] No

**Migration Required:**
- Existing deployments must run migration `c25e56faecce` which:
  - Adds `partition_key` column to `alert_activations`
  - Backfills partition keys for existing rows (MySQL only)
  - Updates primary key constraint
  - Creates `table_partition_interval` metadata table
- **PostgreSQL users**: This is the first release with PostgreSQL partitioning support. Initial deployment will create partitioned tables automatically.
- **Environment variable**: `PARTITION_INTERVAL` is now validated more strictly and must be one of: `DAY`, `MONTH`, `YEARWEEK`

---

### 🔍️ Additional Notes

**Architecture Decisions:**
- Chose integer `partition_key` over expression-based partitioning for PostgreSQL compatibility and performance
- Metadata table approach allows runtime interval changes in future (currently validated to prevent conflicts)
- SQLite support included for local development, gracefully no-ops partition operations

**Performance Considerations:**
- Partition key is computed once during insert via `before_insert` listener
- No performance impact on existing queries (partition key included in PK)
- Bulk backfill during migration uses single UPDATE statement

**Future Work:**
- Allow setting partition intervals for other tables other than `alert_activations`

[ML-11553]: https://iguazio.atlassian.net/browse/ML-11553?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/py/framework/db/sqldb/partition_bootstrapper.py
```diff
@@ -0,0 +1,197 @@
...
+            partition_interval=partition_interval,
+        )
+        if not partition_list:
+            mlrun.utils.logger.warning(
+                "No partitions to create for table",
+                table_name=table_name,
+            )
+            return
+
+        quoted_table = self._quote_table_name(session, table_name, partition_list[0][0])
```
rokatyy: this part seems completely the same as in Postgres below, can be moved in a common function in a base class

server/py/framework/db/base.py
```diff
@@ -956,7 +958,7 @@ def enrich_alert(
         self,
         session,
         alert: mlrun.common.schemas.AlertConfig,
-        state: Optional[framework.db.sqldb.models.AlertState] = None,
+        state: Optional["framework.db.sqldb.models.AlertState"] = None,
```
rokatyy, line 942: why in quotes?

server/py/framework/db/base.py
```diff
@@ -1442,3 +1437,18 @@ def cleanup_old_background_tasks(self, db_session: Session, max_age_seconds: int
...
         """
         pass
+
+    def get_partition_interval_for_table(
+        self,
+        session: sqlalchemy.orm.Session,
+        table_name: str,
+    ) -> Optional[mlrun.common.schemas.partition_interval.PartitionInterval]:
+        pass
+
```
rokatyy, line 1441: add docs

server/py/services/api/migrations/tests/mysql/test_mysql_partitions.py
rokatyy, line 1: no need for `mysql` in file name as it is in the folder `mysql` 

same in all other places

server/py/services/api/migrations/versions/c25e56faecce_add_partition_key_column_for_alerts_.py
```diff
@@ -0,0 +1,132 @@
...
+    for row_id, act_time in connection.execute(select_stmt):
+        connection.execute(
+            alert_table.update()
+            .where(
+                (alert_table.c.id == row_id)
+                & (alert_table.c.activation_time == act_time)
+            )
+            .values(partition_key=interval.get_partition_key_value(act_time))
+        )
+
```
rokatyy: If the table is huge, this might make the migration a very long process. 

Another question - if this is only mysql-related, why not extracting partition_key from partition name and update each partition in bulks?

server/py/services/api/utils/db/partitioner.py
```diff
@@ -11,135 +11,116 @@
...
+import mlrun.common.schemas.partition_interval
 
 import framework.db.sqldb.db
+import framework.db.sqldb.partition_bootstrapper
 import framework.utils.singletons.db
 
 
 class DBPartitioner:
-    def create_and_drop_partitions(
+    def __init__(self, buffer_multiplier_override: Optional[float] = None):
```
rokatyy: why is it float?

server/py/services/api/utils/db/partitioner.py
```diff
@@ -11,135 +11,116 @@
...
-        partition_number = partition_interval.get_number_of_partitions(
-            days=retention_days
-            + mlrun.mlconf.partitions_buffer_multiplier
-            * partition_interval.as_duration().days
-        )
-
-        # Create the calculated number of partitions.
-        self.create_partitions(
-            session, table_name, partition_number, partition_interval
+        partitions_count = max(
```
rokatyy: please add docs

server/py/services/api/utils/db/partitioner.py
```diff
@@ -11,135 +11,116 @@
...
+            1, math.ceil(partitions_to_create * (1 + self._buffer_multiplier))
```
rokatyy: why not using get_number_of_partitions?

server/py/services/api/utils/db/partitioner.py
```diff
@@ -11,135 +11,116 @@
...
-    def create_and_drop_partitions(
+    def __init__(self, buffer_multiplier_override: Optional[float] = None):
+        self._buffer_multiplier = (
+            mlrun.mlconf.partitions_buffer_multiplier
+            if buffer_multiplier_override is None
+            else float(buffer_multiplier_override)
+        )
+        self._db = framework.utils.singletons.db.get_db()
+
+    def create_new_partitions(
```
rokatyy:
```suggestion
    def create_partitions(
```

server/py/services/api/utils/db/partitioner.py
```diff
@@ -11,135 +11,116 @@
...
             table_name=table_name,
-            partitioning_information_list=partitioning_information_list,
+            partition_interval=partition_interval,
+            partitions_count=partitions_count,
         )
+        session.flush()
 
-    @staticmethod
-    def drop_partitions(
+    def create_and_drop_partitions(
```
rokatyy, line 49: shoudl be on top of the class and create/drop below it

server/py/services/api/utils/db/partitioner.py
```diff
@@ -11,135 +11,116 @@
...
     def get_partition_interval(
+        self, session: Session, table_name: str
+    ) -> mlrun.common.schemas.partition_interval.PartitionInterval:
+        """
+        Return the interval recorded for *table_name*.
+        Raises MLRunNotFoundError if the record is missing.
+        """
+        return self._db.get_partition_interval_for_table(session, table_name)
+
+    def drop_old_partitions(
```
rokatyy:
```suggestion
    def drop_partitions(
```

server/py/services/api/utils/db/partitioner.py
```diff
@@ -11,135 +11,116 @@
...
-            )
-
-        partition_interval = (
-            mlrun.common.schemas.partition.PartitionInterval.from_expression(
-                partition_expression
-            )
+        partition_interval: mlrun.common.schemas.partition_interval.PartitionInterval,
+        retention_days: int,
+    ) -> None:
+        db_client = framework.utils.singletons.db.get_db()
```
rokatyy: can use self.db

tests/alerts/test_alert_activations.py
```diff
@@ -14,6 +14,13 @@
 
 import datetime
 
+try:
+    from datetime import UTC  # UTC is only defined in Python 3.11+
+except ImportError:
+    from datetime import timezone
+
+    UTC = timezone.utc
```
rokatyy: is timezone.utc deprecated in 3.11?

rokatyy (commented):
Some comments to start with. 

And one more question - are these test-related changes a must for this PR or can be split into 2 PRs?

mlrun/common/schemas/partition_interval.py
```diff
@@ -114,39 +104,78 @@ def get_next_partition_time(self, current_datetime: datetime) -> datetime:
...
-            return int(days / 30.44)
+            return math.ceil(days / 30.44)
         elif self == PartitionInterval.YEARWEEK:
-            return int(days / 7)
+            return math.ceil(days / 7)
         else:
             raise ValueError(f"Unsupported PartitionInterval: {self}")
+
+    @classmethod
+    def get_partition_interval_from_env(cls) -> "PartitionInterval":
```
theSaarco, line 188: What if we want different partition intervals for different tables? Shouldn't we support possible env values per table name?

theSaarco, line 188: Also, why from env like this and not from MLRun config? Our standard method of operation would be to put this in config so you could override with env. Why do we need a different mechanism here?

server/py/framework/db/sqldb/models.py
```diff
@@ -1118,40 +1139,90 @@ def _sqlite_autoincrement(
...
+        mlrun.common.db.dialects.Dialects.MYSQL,
+        mlrun.common.db.dialects.Dialects.POSTGRESQL,
+    ],
+)
+def set_alert_activations_partition_key(
+    _: Mapper,
+    __: Connection,
+    target: AlertActivation,
+):
+    target.partition_key = target.compute_partition_key()
```
theSaarco: I would leave the option to the caller (i.e. the one doing the insert) to specify the partition key manually, so this should check if the `partition_key` is set, and only compute it if empty.
This can have advantages in the future when for example we want to specify that a specific object should not be subject to GC and put it in a special partition that would not be deleted automatically.

server/py/framework/db/sqldb/partition_bootstrapper.py
```diff
@@ -0,0 +1,197 @@
...
+                {join_str.join(partition_clauses)}
+            )
+        """
+        mlrun.utils.logger.info(
+            "Creating partitions",
+            partitions_count=len(partition_list),
+            table_name=table_name,
+        )
+        session.execute(sqlalchemy.text(ddl))
+        session.commit()
```
theSaarco, line 234: Nitpicking - DDLs do not require a `commit`.

server/py/services/api/migrations/versions/650f0ce2da6f_add_alert_activation.py
```diff
@@ -86,22 +85,24 @@ def upgrade():
     partition_interval = os.getenv("PARTITION_INTERVAL", "YEARWEEK").upper()
```
theSaarco: Use the new `get_partition_interval_from_env` method instead.

server/py/services/api/migrations/versions/c25e56faecce_add_partition_key_column_for_alerts_.py
```diff
@@ -0,0 +1,132 @@
...
+        autoload_with=connection,
+    )
+
+    select_stmt = sa.select(alert_table.c.id, alert_table.c.activation_time)
+    for row_id, act_time in connection.execute(select_stmt):
+        connection.execute(
+            alert_table.update()
+            .where(
+                (alert_table.c.id == row_id)
+                & (alert_table.c.activation_time == act_time)
```
theSaarco: Why do you need the 2nd filter? The row ID should be enough to uniquely identify the row.

server/py/services/api/migrations/versions/c25e56faecce_add_partition_key_column_for_alerts_.py
theSaarco: I second that and also add - you don't do commits anywhere in the process so any failure in the process will cause all updates done until that time to be lost.

server/py/services/api/migrations/versions/c25e56faecce_add_partition_key_column_for_alerts_.py
```diff
@@ -0,0 +1,132 @@
...
+                create_constraint=True,
+            ),
+            nullable=False,
+        ),
+        sa.PrimaryKeyConstraint("table_name"),
+    )
+    engine = op.get_bind()
+    interval = mlrun.common.schemas.partition_interval.PartitionInterval.get_partition_interval_from_env()
+
+    op.add_column(TABLE_NAME, sa.Column("partition_key", sa.Integer(), nullable=True))
```
theSaarco: This isn't idempotent, right? If the updated process failed (due to a failure in filling partition keys) then will this work the 2nd time?
Up until now we didn't really check for this in the upgrade process since it was only used for metadata, and if this failed then something was fundamentally broken, but since you now do the data migration at the same time, this makes it more important.
And, since we're on this topic - why do the `fill_partition_keys` here, and not in the data migration phase?

server/py/services/api/migrations/versions/c25e56faecce_add_partition_key_column_for_alerts_.py
```diff
@@ -0,0 +1,132 @@
...
+    _fill_partition_keys(engine, interval)
```
theSaarco: Will this make rows shift to other partitions? If the partition key is recalculated then the DB may need to move the rows physically across partitions, which usually is a delete/insert operation (and therefore longer).

theSaarco (changes requested):
Looks good in general. Have some questions on the actual implementation, for example the fact that you don't allow for different partitioning intervals per table, which is something we'll most likely want to have in the future.

server/py/framework/db/sqldb/partition_bootstrapper.py
adamdelman: They dont require a commit only on mysql, they do on postgres

mlrun/common/schemas/partition_interval.py
adamdelman: This is not part of the current supported configuration, and so can be added in a future PR.

server/py/framework/db/base.py
adamdelman: to prevent import loop

server/py/services/api/migrations/tests/mysql/test_mysql_partitions.py
adamdelman: using the same test_partitions file name triggers a failure in pytest that complains that multiple files have the same __file__ value

mlrun/common/schemas/partition_interval.py
adamdelman: it was previously set via env, this is not a new mechanism
I did add a table to set this in the future for other tables, but I am keeping the existing implementation for now

server/py/services/api/migrations/versions/c25e56faecce_add_partition_key_column_for_alerts_.py
adamdelman: No, since the partition expression is unchanged no rows are moved.

server/py/services/api/utils/db/partitioner.py
adamdelman: added docstring

PR #8355, opened 2025-07-22
[Model Monitoring] Fixes and additions to graph job monitoring

gtopper:
[ML-10613](https://iguazio.atlassian.net/browse/ML-10613)

* Add MEP list to serving spec and extract it at runtime
* Push batch completion event per MEP
* Allow the user to set timestamp column to set event time for monitoring purposes
* Disable monitoring when batch it too long (as measured from start time to end time)

Also:
* Fixes writes to MM stream
* Fixes failure to delete model monitoring resources when only some resources exist
* Improves warning to avoid completely silencing errors
* Improves debug log
* Expands relevant system test

Passed:
* `TestMonitoringAppFlow::test_app_flow[with_training_set=False, with_model_runner=*]`
* `TestModelMonitoringInitialize::test_model_monitoring_crud`
* `TestModelMonitoringOverJob::test_job_from_serving_runtime_with_model_tracking[*]`
* `TestKubejobRuntime::test_job_from_serving_runtime[*]`
* `TestNuclioRuntime::test_deploy_function_with_model_runner[raise_exception=*, with_object=True]`

[ML-10613]: https://iguazio.atlassian.net/browse/ML-10613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/serving/server.py
+++ b/mlrun/serving/server.py
@@ -42,6 +42,7 @@ from mlrun.secrets import SecretsStore
 
 from ..common.helpers import parse_versioned_object_uri
 from ..common.schemas.model_monitoring.constants import FileTargetKind
+from ..common.schemas.serving import MAX_BATCH_JOB_DURATION
 from ..datastore import DataItem, get_stream_pusher
 from ..datastore.store_resources import ResourceCache
 from ..errors import MLRunInvalidArgumentError
@@ -629,7 +630,7 @@ async def async_execute_graph(
             start_time = start_time.isoformat()
             end_time = end_time.isoformat()
             # TODO: tie this to the controller's base period
-            if time_range > pd.Timedelta("1w"):
+            if time_range > pd.Timedelta(MAX_BATCH_JOB_DURATION):
                 raise mlrun.errors.MLRunRuntimeError(
                     f"Dataframe time range is too long: {time_range}. "
                     "Please disable tracking or reduce the input dataset's time range to under one week."
--- a/mlrun/serving/server.py
+++ b/mlrun/serving/server.py
@@ -608,14 +608,15 @@ async def async_execute_graph(
                 f"(status='{task_state}')"
             )
 
-    mm_enabled = True
     df = data.as_df()
 
     if df.empty:
         context.logger.warn("Job terminated due to empty inputs (0 rows)")
         return []
 
-    if timestamp_column:
+    track_models = spec.get("track_models")
+
+    if track_models and timestamp_column:
         context.logger.info(f"Sorting dataframe by {timestamp_column}")
         df[timestamp_column] = pd.to_datetime(  # in case it's a string
             df[timestamp_column]
@@ -628,25 +629,24 @@ async def async_execute_graph(
             start_time = start_time.isoformat()
             end_time = end_time.isoformat()
             # TODO: tie this to the controller's base period
-            if time_range > pd.Timedelta("10000h"):
-                context.logger.warn(
-                    f"Dataframe time range is too long: {time_range}. Job will not be tracked!"
+            if time_range > pd.Timedelta("1w"):
+                raise mlrun.errors.MLRunRuntimeError(
+                    f"Dataframe time range is too long: {time_range}. "
+                    "Please disable tracking or reduce the input dataset's time range to under one week."
                 )
-                mm_enabled = False
         else:
             start_time = end_time = df["timestamp"].iloc[0].isoformat()
     else:
         start_time = datetime.now(tz=timezone.utc).isoformat()
 
-    if mm_enabled:
-        server.graph = add_system_steps_to_graph(
-            server.project,
-            copy.deepcopy(server.graph),
-            spec.get("track_models"),
-            context,
-            spec,
-            pause_until_background_task_completion=False,  # we've already awaited it
-        )
+    server.graph = add_system_steps_to_graph(
+        server.project,
+        copy.deepcopy(server.graph),
+        track_models,
+        context,
+        spec,
+        pause_until_background_task_completion=False,  # we've already awaited it
+    )
 
     if config.log_level.lower() == "debug":
         server.verbose = True
@@ -667,8 +667,6 @@ async def async_execute_graph(
     if server.verbose:
         context.logger.info(server.to_yaml())
 
-    responses = []
-
     async def run(body):
         event = storey.Event(id=index, body=body)
         if timestamp_column:
@@ -682,13 +680,13 @@ async def async_execute_graph(
                     f"Event body '{body}' did not contain timestamp column '{timestamp_column}'"
                 )
             event._original_timestamp = body[timestamp_column]
-        response = await server.run(event, context)
-        responses.append(response)
+        return await server.run(event, context)
 
     if batching and not batch_size:
         batch_size = len(df)
 
     batch = []
+    tasks = []
     for index, row in df.iterrows():
         data = row.to_list() if read_as_lists else row.to_dict()
         if nest_under_inputs:
@@ -696,13 +694,15 @@ async def async_execute_graph(
         if batching:
             batch.append(data)
             if len(batch) == batch_size:
-                await run(batch)
+                tasks.append(asyncio.create_task(run(batch)))
                 batch = []
         else:
-            await run(data)
+            tasks.append(asyncio.create_task(run(data)))
 
     if batch:
-        await run(batch)
+        tasks.append(asyncio.create_task(run(batch)))
+
+    responses = await asyncio.gather(*tasks)
 
     termination_result = server.wait_for_completion()
     if asyncio.iscoroutine(termination_result):
--- a/mlrun/serving/server.py
+++ b/mlrun/serving/server.py
@@ -670,7 +670,9 @@ async def async_execute_graph(
     async def run(body):
         event = storey.Event(id=index, body=body)
         if timestamp_column:
-            body = event.body
+            if batching:
+                # we use the first row in the batch to determine the timestamp for the whole batch
+                body = body[0]
             if not isinstance(body, dict):
                 raise mlrun.errors.MLRunRuntimeError(
                     f"When timestamp_column=True, event body must be a dict – got {type(body).__name__} instead"
--- a/mlrun/serving/server.py
+++ b/mlrun/serving/server.py
@@ -22,8 +22,10 @@ import os
 import socket
 import traceback
 import uuid
+from datetime import datetime, timezone
 from typing import Any, Optional, Union
 
+import pandas as pd
 import storey
 from nuclio import Context as NuclioContext
 from nuclio.request import Logger as NuclioLogger
@@ -561,6 +563,7 @@ def v2_serving_init(context, namespace=None):
 async def async_execute_graph(
     context: MLClientCtx,
     data: DataItem,
+    timestamp_column: Optional[str],
     batching: bool,
     batch_size: Optional[int],
     read_as_lists: bool,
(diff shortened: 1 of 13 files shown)


mlrun/serving/server.py
```diff
@@ -605,14 +608,45 @@ async def async_execute_graph(
...
+        )
+        df.sort_values(by=timestamp_column, inplace=True)
+        if len(df) >= 2:
+            start_time = df[timestamp_column].iloc[0]
+            end_time = df[timestamp_column].iloc[-1]
+            time_range = end_time - start_time
+            start_time = start_time.isoformat()
+            end_time = end_time.isoformat()
+            # TODO: tie this to the controller's base period
+            if time_range > pd.Timedelta("10000h"):
```
davesh0812: Lets make write it as a constant and maybe lets to it more meaningful and define it as a week, 7*24

mlrun/serving/server.py
```diff
@@ -605,14 +608,45 @@ async def async_execute_graph(
...
+                context.logger.warn(
+                    f"Dataframe time range is too long: {time_range}. Job will not be tracked!"
+                )
+                mm_enabled = False
```
davesh0812: I think that if `track_models=True` and the time range is to long we should raise an exception.

mlrun/serving/server.py
```diff
@@ -665,12 +708,42 @@ async def run(body):
     if asyncio.iscoroutine(termination_result):
```
davesh0812, line 713: The comment is on line 699 - Why we should await each run and not to execute it in parallel? and in the end to use `gather()`?

mlrun/serving/server.py
Eyal-Danieli: didn't we agree on 1 week? 10000h is about 2 months..

mlrun/serving/system_steps.py
```diff
@@ -168,6 +168,12 @@ def do(self, event):
                     request, resp = self.reconstruct_request_resp_fields(
                         event, model, monitoring_data[model]
                     )
+                    if hasattr(event, "_original_timestamp"):
+                        when = event._original_timestamp
+                    else:
+                        when = event._metadata.get(model, {}).get(
+                            mm_schemas.StreamProcessingEvent.WHEN
+                        )
```
davesh0812, line 204: So according to that each batch will have the same timestamp, right?
If it does we should write it somewhere Or better to fully cover this in the stream pod.

mlrun/serving/server.py
```diff
@@ -665,12 +708,42 @@ async def run(body):
...
+    batch_completion_time = datetime.now(tz=timezone.utc).isoformat()
+
+    if not timestamp_column:
+        end_time = batch_completion_time
+
+    mm_stream_record = dict(
+        kind="batch_complete",
+        project=context.project,
+        first_timestamp=start_time,
+        last_timestamp=end_time,
```
Eyal-Danieli, line 731: `end_time` is only valid when `timestamp_column` is provided.

server/py/framework/api/utils.py
```diff
@@ -249,16 +250,21 @@ async def submit_run(
...
                 serving_spec["model_endpoint_creation_task_name"] = (
                     model_endpoint_creation_task_name
                 )
+                serving_spec["model_endpoint_uids"] = model_endpoint_uids
                 fn.spec.serving_spec = json.dumps(serving_spec)
 
             logger.info(
                 "Started model endpoint creation task",
                 model_endpoint_creation_task_name=model_endpoint_creation_task_name,
+                serving_spec=json.dumps(serving_spec),
```
davesh0812: it can be a very large log... lets remove it from here

server/py/services/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -1087,12 +1087,16 @@ def delete_model_endpoint_monitoring_resources(
...
-        )
         stream_path = mlrun.model_monitoring.get_stream_path(
             project=project_name, profile=stream_profile
         )
+        logger.debug(
+            "Deleting model monitoring endpoints resources",
+            project_name=project_name,
+            stream_path=stream_path,
+            igz_version=mlrun.mlconf.igz_version,
+            v3io_api=mlrun.mlconf.v3io_api,
```
davesh0812: If we not running on igz/v3io it will return None?

server/py/services/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -1087,12 +1087,16 @@ def delete_model_endpoint_monitoring_resources(
...
-            "Deleting model monitoring endpoints resources", project_name=project_name
-        )
         stream_path = mlrun.model_monitoring.get_stream_path(
             project=project_name, profile=stream_profile
         )
+        logger.debug(
+            "Deleting model monitoring endpoints resources",
+            project_name=project_name,
+            stream_path=stream_path,
+            igz_version=mlrun.mlconf.igz_version,
```
Eyal-Danieli: don't think it's a problem but we might need to verify that this value is not removed from the CE during the installation

Eyal-Danieli (commented):
LGTM with few comments 
I didn’t go through the tests yet

server/py/services/api/utils/endpoints.py
```diff
@@ -74,4 +75,13 @@ async def start_model_endpoint_creation_background_task(
         else None
     )
 
-    return function, model_endpoint_creation_task_name, returned_background_tasks
+    model_endpoint_uids = [
+        model_endpoint.metadata.uid
+        for (model_endpoint, _) in model_endpoints_instructions
```
davesh0812, line 80: Why do you need it? I don't see any usage of it..

server/py/services/api/utils/endpoints.py
```diff
@@ -34,6 +34,7 @@ async def start_model_endpoint_creation_background_task(
     returned_background_tasks = mlrun.common.schemas.BackgroundTaskList(
         background_tasks=[]
     )
+    model_endpoints_instructions = []
```
davesh0812, line 37: I dont think we really need this

mlrun/serving/server.py
gtopper: Thanks, I forgot to update it. Isn't it over a year though? :)

server/py/framework/api/utils.py
```diff
@@ -249,16 +250,21 @@ async def submit_run(
 
             # TODO: there should be a better way to do this
             serving_spec = getattr(fn.spec, "serving_spec")
+            # update the graph from the function, because MEP IDs were added
             if serving_spec:
                 serving_spec = json.loads(serving_spec)
+                new_graph = fn.spec.graph.to_dict(strip=True) if fn.spec.graph else {}
```
davesh0812, line 256: Nice bug catch!!

server/py/services/api/api/endpoints/nuclio.py
```diff
@@ -271,6 +271,7 @@ async def deploy_function(
         function,
         model_endpoint_creation_task_name,
         returned_background_tasks,
+        _,
```
davesh0812, line 274: If you agree with the prev comment you can delete this also

davesh0812 (commented):
Nice!
Few comments and questions

mlrun/serving/server.py
gtopper: You mean only defined? Not so, although PyCharm's static analysis misses it:
```python
if not timestamp_column:
    end_time = batch_completion_time
```

server/py/services/api/crud/model_monitoring/model_endpoints.py
gtopper: It will just be `None`.

mlrun/serving/server.py
gtopper: Changed it to 1w. Where do you suggest placing this constant?

mlrun/serving/system_steps.py
gtopper: Not sure I understand why it would be the same timestamp. `_original_timestamp` comes from the timestamp column. `event._metadata[model][WHEN]` comes from the clock time the model ran.

Neither of these is per batch.

server/py/services/api/crud/model_monitoring/model_endpoints.py
gtopper: Right.

server/py/services/api/utils/endpoints.py
gtopper: What do you mean? `model_endpoint_uids` is returned immediately below.

server/py/services/api/utils/endpoints.py
gtopper: We need it to pass to `_create_model_endpoint_background_task`, and to extract and return `model_endpoint_uids`.

server/py/services/api/api/endpoints/nuclio.py
gtopper: We don't need the UID list in nuclio, but we need it for the job.

mlrun/serving/server.py
gtopper: Because we're waiting for the responses, this is actually a good point... Will definitely have to test it.

server/py/services/api/crud/model_monitoring/model_endpoints.py
gtopper: We probably don't really need these anyway though.

server/py/services/api/crud/model_monitoring/model_endpoints.py
gtopper: We probably don't really need these anyway though.

server/py/services/api/utils/endpoints.py
davesh0812: but we create those inside `_create_model_endpoints_instructions` and it comes before `_create_model_endpoint_background_task`

mlrun/serving/server.py
davesh0812: But why we need to wait?

mlrun/serving/server.py
davesh0812: Where you find it right

mlrun/serving/server.py
```diff
@@ -633,12 +667,21 @@ async def async_execute_graph(
...
+            body = event.body
+            if not isinstance(body, dict):
+                raise mlrun.errors.MLRunRuntimeError(
+                    f"When timestamp_column=True, event body must be a dict – got {type(body).__name__} instead"
+                )
+            if timestamp_column not in body:
+                raise mlrun.errors.MLRunRuntimeError(
+                    f"Event body '{body}' did not contain timestamp column '{timestamp_column}'"
+                )
+            event._original_timestamp = body[timestamp_column]
```
davesh0812, line 687: What will happen if you have a batch? In this case the `body[timestamp_column]` is a list.

mlrun/serving/system_steps.py
davesh0812: You have the ability to send a batch am I wrong?

```python
for index, row in df.iterrows():
        data = row.to_list() if read_as_lists else row.to_dict()
        if nest_under_inputs:
            data = {"inputs": data}
        if batching:
            batch.append(data)
            if len(batch) == batch_size:
                await run(batch)
                batch = []
        else:
            await run(data)
```

mlrun/serving/server.py
gtopper: I updated the code as per your suggestion, please have a look.

mlrun/serving/system_steps.py
gtopper: You're right, `batching` + `timestamp_column` isn't handled properly. Maybe the best way to handle this is by applying the first timestamp to the whole batch, but I think I need to check with @theSaarco.

server/py/services/api/utils/endpoints.py
gtopper: We need the empty list in case we didn't go into the conditional block.

mlrun/serving/system_steps.py
gtopper: I changed the code to use the first event in the batch for timestamp purposes for now.

davesh0812: OK @theSaarco, @assaf758  please have a look :)

mlrun/serving/server.py
```diff
@@ -633,44 +667,83 @@ async def async_execute_graph(
...
                 batch = []
         else:
-            await run(data)
+            tasks.append(asyncio.create_task(run(data)))
 
     if batch:
-        await run(batch)
+        tasks.append(asyncio.create_task(run(batch)))
+
+    responses = await asyncio.gather(*tasks)
```
davesh0812, line 710: NIce!

mlrun/serving/server.py
gtopper: Added a constant.

mlrun/serving/system_steps.py
theSaarco: After discussing this with @gtopper, I'm fine with the handling currently done, i.e. using the TS of the 1st event in the batch.

mlrun/serving/server.py
```diff
@@ -605,10 +609,41 @@ async def async_execute_graph(
...
+            start_time = df[timestamp_column].iloc[0]
+            end_time = df[timestamp_column].iloc[-1]
+            time_range = end_time - start_time
+            start_time = start_time.isoformat()
+            end_time = end_time.isoformat()
+            # TODO: tie this to the controller's base period
+            if time_range > pd.Timedelta(MAX_BATCH_JOB_DURATION):
+                raise mlrun.errors.MLRunRuntimeError(
+                    f"Dataframe time range is too long: {time_range}. "
+                    "Please disable tracking or reduce the input dataset's time range to under one week."
```
theSaarco:
```suggestion
                    "Please disable tracking or reduce the input dataset's time range below the limit defined - {MAX_BATCH_JOB_DURATION}."
```

theSaarco (approved):
LGTM. One small comment.

mlrun/serving/server.py
```diff
@@ -681,6 +754,7 @@ def execute_graph(
 
     :param context: The job's execution client context.
     :param data: The input data to the job, to be pushed into the graph row by row, or in batches.
+    :param timestamp_column: The name of the column that will be used as the timestamp for model monitoring purposes.
```
davesh0812, line 762: Lets add  this https://github.com/mlrun/mlrun/pull/8355#discussion_r2233830371, to the docs

Eyal-Danieli (approved):

davesh0812 (approved):

PR #8356, opened 2025-07-22
[Workflows] Add counter labels to track workflow retries

yaelgen:
ticket: https://iguazio.atlassian.net/browse/ML-10221

This PR adds two new labels to give better visibility into retry activity for remote workflows:

- `rerun-counter` - Applied to the original workflow‑runner run. Increments each time the pipeline is retried, so you can see how many retry attempts have been made.

- `rerun-index` - Applied to each RerunRunner invocation. Contains the 1‑based ordinal of this retry (e.g. 1, 2, …), making it easy to correlate a specific runner pod with its retry sequence.


liranbg (approved):

quaark (approved):

server/py/framework/db/sqldb/db.py
```diff
@@ -367,13 +367,12 @@ def set_run_retrying_status(
...
             elif mlrun_constants.MLRunInternalLabels.retrying in labels:
-                # flush and commit so the lock is released immediately
-                session.commit()
                 raise mlrun.errors.MLRunConflictError
             else:
-                # TODO: bump counter label here in follow-up
                 labels[mlrun_constants.MLRunInternalLabels.retrying] = "true"
-
+                labels[mlrun_constants.MLRunInternalLabels.rerun_counter] = (
+                    labels.get(mlrun_constants.MLRunInternalLabels.rerun_counter, 0) + 1
```
Yacouby: Doesn't labels.get(...) return a string? So won't labels.get(..., 0) + 1 result in a type error since we're adding an int to a str?

Yacouby (commented):
what about a unit test to cover this logic ? It would be better to add one, as it would help catch regressions if anything in this flow changes in the future.

Yacouby (approved):
Nice :) looks good

PR #8357, opened 2025-07-22
[DB] Fix iteration over None in DB configuration utility

yaelgen:
### [DBUtil] Fix crash on startup, improve config logic and tests

- Fixed crash when `config_items` was `None` (e.g. no custom DB config)
  - Previously caused `TypeError` on `",".join(None)`
  - Now uses `_EMPTY_DB_CONFIGURATIONS` as fallback
  - Ensures resolved items are passed to `_apply_configurations()`
- Fixed `is_valid()` port check (`or` → `and`)
- `_to_keyset(None)` now returns `set()` instead of `None`

**Tests:**
- Added integration tests for noop values (`"nil"`, `"none"`)
- Added live config apply/restore tests (`work_mem`, `PIPES_AS_CONCAT`)
- Added unit tests for skipped noop input and apply call args

https://iguazio.atlassian.net/browse/ML-10647

### DIFF
--- a/server/py/framework/utils/db/utils.py
+++ b/server/py/framework/utils/db/utils.py
@@ -102,7 +102,7 @@ class ParsedDsn:
             return False
         if not self.host or not self._HOST_REGEX.fullmatch(self.host):
             return False
-        if self.port is not None or not 1 <= self.port <= 65535:
+        if self.port is not None and not 1 <= self.port <= 65535:
             return False
 
         return True
@@ -175,7 +175,7 @@ class DBUtil:
         items = config_items or self._DEFAULT_DB_CONFIGURATIONS
         keys = _to_keyset(items)
 
-        if not keys or keys.issubset(self._EMPTY_DB_CONFIGURATIONS):
+        if not keys or keys.intersection(self._EMPTY_DB_CONFIGURATIONS):
             mlrun.utils.logger.debug(
                 "No configurations specified – skipping",
                 configs=config_items,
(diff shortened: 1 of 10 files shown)


liranbg (approved):

server/py/framework/utils/db/utils.py
```diff
@@ -172,7 +172,7 @@ def set_configurations(
         self,
         config_items: Optional[Union[list[str], dict[str, Any]]] = None,
     ) -> None:
-        items = config_items or self._DEFAULT_DB_CONFIGURATIONS
+        items = config_items or self._EMPTY_DB_CONFIGURATIONS
```
adamdelman: @yaelgen Can you please remove self._DEFAULT_DB_CONFIGURATIONS as it is not in use

adamdelman (approved):

server/py/framework/utils/db/utils.py
```diff
@@ -142,8 +142,8 @@ class DBUtil:
     _DIALECT = None
     _DSN_ENV = "MLRUN_HTTPDB__DSN"
     _DRIVER_CACHE: dict[str, Any] = {}
+    _EMPTY_DB_CONFIGURATIONS = {"nil", "none"}
```
moranbental, line 145: Why did you remove `set[str]`?

moranbental (changes requested):

moranbental (approved):

server/py/framework/utils/db/utils.py
adamdelman: because it it implicit in the way it is defined, no additional type hint required

PR #8358, opened 2025-07-22
[ModelProvider] Added Huggingface provider

tomerm-iguazio:
1) Added Hugging face provider.
2) Bugfix secrets override in datastore.py
3) Bugfix: Correct secrets handling in get_model_provider – disable caching
4) Added integration and system tests for hugginface, system test do not required any token.

[ML-10256)](https://iguazio.atlassian.net/browse/ML-10256)


mlrun/datastore/model_provider/huggingface_provider.py
```diff
@@ -0,0 +1,173 @@
...
+        Raises:
+            ImportError: If the `transformers` package is not installed.
+        """
+        try:
+            from transformers import pipeline, AutoModelForCausalLM  # noqa
+            from transformers import AutoTokenizer  # noqa
+
+            self._client = pipeline(model=self.model, **self.options)
+        except ImportError as exc:
+            raise ImportError("openai package is not installed") from exc
```
royischoss: is it openai ?

mlrun/datastore/model_provider/huggingface_provider.py
```diff
@@ -0,0 +1,173 @@
...
+            )
+        ```
+
+
+        :param operation:               A pipeline object
+        :param invoke_kwargs:           Keyword arguments to pass to the operation.
+        :return:                        The full response returned by the operation.
+
+        """
+        invoke_kwargs = self.get_invoke_kwargs(invoke_kwargs)
```
royischoss, line 141: self.get_invoke_kwargs looks redundant, if it merges two dicts

mlrun/datastore/model_provider/huggingface_provider.py
```diff
@@ -0,0 +1,173 @@
...
+    @staticmethod
+    def _extract_string_output(result) -> str:
+        """
+        Extracts the first generated string from Hugging Face pipeline output,
+        regardless of whether it's plain text-generation or chat-style output.
+        """
+        if not isinstance(result, list) or len(result) == 0:
+            raise ValueError("Empty or invalid pipeline output")
+
+        return result[0].get("generated_text")
```
royischoss, line 74: following today meeting , we might add the full text collector, maybe by adding a flag to this method

mlrun/datastore/model_provider/model_provider.py
```diff
@@ -148,7 +148,7 @@ def client(self) -> Any:
 
     @property
     def model(self) -> Optional[str]:
-        return None
+        return self.endpoint
```
royischoss, line 156: I don't know if it for this pr but there is no documentation of what we expect under this value in the init so maybe document this method

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -127,7 +123,9 @@ def custom_invoke(
         """
         invoke_kwargs = self.get_invoke_kwargs(invoke_kwargs)
         if operation:
-            return operation(**invoke_kwargs, model=self.model)
+            return operation(
```
royischoss: if we add a default kwarg model we might need to use try except or inspect the signature preventing error in the user code

mlrun/datastore/datastore.py
```diff
@@ -102,8 +103,7 @@ def schema_to_store(schema) -> DataStore.__subclasses__():
 def schema_to_model_provider(
     schema: str, raise_missing_schema_exception=True
 ) -> type[ModelProvider]:
-    #  TODO add hugging face and http
-    schema_dict = {"openai": OpenAIProvider}
+    schema_dict = {"openai": OpenAIProvider, "huggingface": HuggingFaceProvider}
```
royischoss, line 106: http will not be added?

mlrun/datastore/datastore_profile.py
```diff
@@ -486,6 +486,36 @@ def url(self, subpath):
...
+    task: typing.Optional[str] = None
+    token: typing.Optional[str] = None
+    device: typing.Optional[typing.Union[int, str]] = None
+    device_map: typing.Union[str, dict[str, typing.Union[int, str]], None] = None
+    trust_remote_code: bool = None
+    model_kwargs: typing.Optional[dict[str, typing.Any]] = None
+
+    def secrets(self) -> dict:
+        res = {}
+        if self.task:
```
royischoss: I think it can be cleaner as we said

royischoss (commented):
Hey nice work few comments

mlrun/datastore/model_provider/huggingface_provider.py
tomerm-iguazio: I use it in 3 different places, I think it will be better to create a function for that even if it is short

mlrun/datastore/model_provider/huggingface_provider.py
tomerm-iguazio: it will be in a future PR, if required by @theSaarco

mlrun/datastore/datastore.py
tomerm-iguazio: We decided not to add it for now.

mlrun/datastore/datastore_profile.py
tomerm-iguazio: yep, tnx:)

mlrun/datastore/model_provider/openai_provider.py
tomerm-iguazio: I think if there is an error we should still raise it, WDYT?

royischoss: maybe but you should give informative error msg cause you are forcing the user to give you an operation that allow kwargs or must support model param

tomerm-iguazio: As we discussed, I'll check operation type:)

tomerm-iguazio: done:)

royischoss (approved):
Nice 🏆

PR #8364, opened 2025-07-23
[Nuclio] Prevent conflict between functions code and source archive

Yacouby:
This PR adds a validation step in `pre_deploy_validation` for Nuclio functions to prevent a misconfiguration where both the function source code and a source archive are provided. 

Jira:
https://iguazio.atlassian.net/browse/ML-7678

### DIFF
--- a/mlrun/runtimes/nuclio/function.py
+++ b/mlrun/runtimes/nuclio/function.py
@@ -296,7 +296,8 @@ class RemoteRuntime(KubeResource):
         code = (
             self.spec.build.functionSourceCode if hasattr(self.spec, "build") else None
         )
-        if code and self.spec.build.source:
+        load_archive = self.spec.build.load_source_on_run and self.spec.build.source
+        if code and load_archive:
             raise mlrun.errors.MLRunInvalidArgumentError(
                 "cannot specify both code and source archive"
             )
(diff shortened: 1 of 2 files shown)


rokatyy (approved):
LGTM

moranbental (approved):

alonmr (changes requested):
I don't think this will work with application runtime because it might have both (code for the reverse proxy and source for the application)

Yacouby:
> I don't think this will work with application runtime because it might have both (code for the reverse proxy and source for the application)

@alonmr Added a check to ensure that for application runtimes, load_source_on_run is not set, since the source archive must be included in the image. This allows supporting both functionSourceCode (e.g., for reverse proxy) and build.source without conflict.

alonmr:
> > I don't think this will work with application runtime because it might have both (code for the reverse proxy and source for the application)
> 
> @alonmr Added a check to ensure that for application runtimes, load_source_on_run is not set, since the source archive must be included in the image. This allows supporting both functionSourceCode (e.g., for reverse proxy) and build.source without conflict.

The assumption that "source archive must be included in the image" is inaccurate because we have plans to support load on run as well. Furthermore, `load_source_on_run` can be False for nuclio runtime and it will not explode although it should.

PR #8363, opened 2025-07-23
[Model Monitoring] Forward 'batch complete' events from serving stream to controller

gtopper:
[ML-10615](https://iguazio.atlassian.net/browse/ML-10615)

Updated stream processing flow graph (left branch is new):
stream-processing-flow-graph

[ML-10615]: https://iguazio.atlassian.net/browse/ML-10615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8365, opened 2025-07-23
[Model Monitoring] Implement batch event handling in the controller

Eyal-Danieli:
As part of supporting monitoring for batch model endpoints generated through serving jobs (#8355 ), this PR implements batch event handling in the controller.

In more detail:

- When a `batch_complete` event is pushed to the controller stream (#8363 ), it is identified as a new type of event that already includes both the start and end times of a potentially large time range, along with relevant `endpoint_id`.
- The controller generates analysis windows starting from the maximum of `last_analyzed` and the start batch time for the processed model endpoint.
- The final window is always analyzed and closed, even if it’s only a partial window (`base_period` - `delta`), and `last_analyzed` is updated accordingly.

Additionally, this PR introduces a new mode filter in the `list_model_endpoints` API, which can be either `real_time` or `batch`.

https://iguazio.atlassian.net/browse/ML-10558

### DIFF
--- a/mlrun/model_monitoring/controller.py
+++ b/mlrun/model_monitoring/controller.py
@@ -572,7 +572,9 @@ class MonitoringApplicationController:
 
                 endpoint_mode = mm_constants.EndpointMode.REAL_TIME
 
-            logger.info("Starting to analyze", timestamp=last_stream_timestamp)
+            logger.info(
+                "Starting to analyze", timestamp=last_stream_timestamp.isoformat()
+            )
 
             with _BatchWindowGenerator(
                 project=project_name,
--- a/mlrun/model_monitoring/controller.py
+++ b/mlrun/model_monitoring/controller.py
@@ -883,7 +883,6 @@ class MonitoringApplicationController:
         :param endpoint_id: endpoint id string
         :param endpoint_name: the endpoint name string
         :param endpoint_type: Enum of the endpoint type
-
         """
         event = {
             ControllerEvent.KIND.value: kind,
--- a/mlrun/model_monitoring/controller.py
+++ b/mlrun/model_monitoring/controller.py
@@ -145,16 +145,32 @@ class _BatchWindow:
                 last_analyzed=last_analyzed,
             )
 
-        if last_analyzed and self._endpoint_mode == mm_constants.EndpointMode.BATCH:
+        if self._endpoint_mode == mm_constants.EndpointMode.BATCH:
             # If the endpoint is a batch endpoint, we need to update the last analyzed time
             # to the end of the batch time.
-
-            yield _Interval(
-                datetime.datetime.fromtimestamp(
-                    last_analyzed, tz=datetime.timezone.utc
-                ),
-                datetime.datetime.fromtimestamp(self._stop, tz=datetime.timezone.utc),
-            )
+            if last_analyzed:
+                if last_analyzed < self._stop:
+                    # If the last analyzed time is earlier than the stop time,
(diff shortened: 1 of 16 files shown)


mlrun/model_monitoring/controller.py
```diff
@@ -827,7 +881,7 @@ def push_to_controller_stream(
         :param endpoint_id: endpoint id string
         :param endpoint_name: the endpoint name string
         :param endpoint_type: Enum of the endpoint type
-        :param feature_set_uri: the feature set uri string
+
```
gtopper:
```suggestion
```

server/py/services/api/api/endpoints/model_endpoints.py
```diff
@@ -242,6 +242,7 @@ async def list_model_endpoints(
     start: Optional[datetime] = None,
     end: Optional[datetime] = None,
     top_level: bool = Query(False, alias="top-level"),
+    mode: mm_constants.EndpointMode = Query(None, alias="mode"),
```
gtopper: I don't think it needs an alias if the alias is the same as the parameter name.

mlrun/model_monitoring/controller.py
```diff
@@ -477,24 +515,65 @@ def model_endpoint_process(
...
+                    event[ControllerEvent.FIRST_REQUEST]
+                )
+
+                endpoint_updated = event[ControllerEvent.ENDPOINT_POLICY][
+                    ControllerEventEndpointPolicy.ENDPOINT_UPDATED
+                ]
+
+                endpoint_mode = mm_constants.EndpointMode.REAL_TIME
+
+            logger.info("Starting analyzing for", timestamp=last_stream_timestamp)
```
gtopper:
```suggestion
            logger.info("Starting analysis for", timestamp=last_stream_timestamp)
```
or
```suggestion
            logger.info("Starting to analyze for", timestamp=last_stream_timestamp)
```
And maybe drop the "for".

davesh0812 (approved):
Please add some tests for the new filter and lets wait for @gtopper pr and merge this after.

mlrun/model_monitoring/controller.py
```diff
@@ -477,24 +515,65 @@ def model_endpoint_process(
...
+                    event[ControllerEvent.FIRST_REQUEST]
+                )
+
+                endpoint_updated = event[ControllerEvent.ENDPOINT_POLICY][
+                    ControllerEventEndpointPolicy.ENDPOINT_UPDATED
+                ]
+
+                endpoint_mode = mm_constants.EndpointMode.REAL_TIME
+
+            logger.info("Starting to analyze", timestamp=last_stream_timestamp)
```
gtopper: Ah, I think you might need the string here, not the timestamp object, or you'll get an ugly `datetime.datetime(2025, 7, 27, 12, 17, 26, 153100)` type of thing in the logs.

Eyal-Danieli: https://github.com/mlrun/mlrun/pull/8365/commits/c3deb7da2ecb4befb068de61b89e1ce14da25d5a

mlrun/model_monitoring/controller.py
Eyal-Danieli: https://github.com/mlrun/mlrun/pull/8365/commits/7331d0c7e30a9193e5bbce171c2ea3c118107149

gtopper (approved):

davesh0812 (approved):

PR #8370, opened 2025-07-23
[Deprecations] Remove /user-secrets endpoint and the corresponding SDK method [feature/ig4-authentication]

moranbental:
The previous `/user-secrets` POST endpoint was never implemented and always returns HTTP 400. It will be removed and replaced with a scoped API under `/api/v1/user-secrets/tokens`, which is purpose-built for token management.

removing also the related SDK method `db.create_user_secrets(...)`
https://iguazio.atlassian.net/browse/ML-10452


moranbental:
Fails on backward compatibility because we removed an unused endpoint that always returned a Bad Request

TomerShor (approved):

PR #8372, opened 2025-07-24
[Model Monitoring] Fix avoid collecting default output schema from model endpoint

royischoss:
include:

- safer approach to key values in user output when processing dict.
- remove accessing model endpoint for output schema fetch - avoiding default schema [p0,...pn]

https://iguazio.atlassian.net/browse/ML-10656


gtopper:
For completeness, I believe this is following breakage in #8293.

davesh0812 (approved):

PR #8373, opened 2025-07-24
[RemoteModel] Remote model minor fixes

tomerm-iguazio:
According to @assaf758 comments in:
https://github.com/mlrun/mlrun/pull/8291#pullrequestreview-3036142980

+ Adding pillow to dev-requirements, due to hf image classification test.

Depends on https://github.com/mlrun/mlrun/pull/8358

[ML-10707](https://iguazio.atlassian.net/browse/ML-10707)

[ML-10707]: https://iguazio.atlassian.net/browse/ML-10707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/datastore/model_provider/huggingface_provider.py
```diff
@@ -0,0 +1,177 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import TYPE_CHECKING, Any, Optional, Union
+
+import mlrun
+from mlrun.datastore.model_provider.model_provider import ModelProvider
+
+if TYPE_CHECKING:
+    from transformers.pipelines import pipeline
```
tomerm-iguazio: change it to Pipeline

mlrun/datastore/model_provider/huggingface_provider.py
```diff
@@ -150,22 +148,24 @@ def custom_invoke(
 
     def invoke(
         self,
-        messages: Union[str, list[str], ChatType, list[ChatType]] = None,
+        messages: Union[str, list[str], "ChatType", list["ChatType"]] = None,
         as_str: bool = False,
         **invoke_kwargs,
-    ) -> Optional[Union[str, list, T]]:
+    ) -> Union[str, list]:
```
davesh0812, line 154: So it can be also a dictionary? 
According to the docs...

mlrun/datastore/model_provider/huggingface_provider.py
```diff
@@ -117,7 +115,7 @@ def get_client_options(self):
 
     def custom_invoke(
         self, operation: Optional["Pipeline"] = None, **invoke_kwargs
-    ) -> Optional[T]:
+    ) -> Union[list, Any]:
```
davesh0812: We don't have something more specific?

mlrun/datastore/model_provider/model_provider.py
```diff
@@ -127,67 +179,11 @@ def invoke(
...
-        """
-        raise NotImplementedError("async_custom_invoke is not implemented")
-
     async def async_invoke(
         self,
         messages: Optional[list[dict]] = None,
         as_str: bool = False,
         **invoke_kwargs,
-    ) -> Optional[str]:
+    ) -> str:
```
davesh0812: Can be Any no? like `def invoke() `

tests/integration/model_providers/openai/test_openai.py
```diff
@@ -115,31 +114,43 @@ def setup_datastore_profile(self):
...
+                messages=messages,
+                max_tokens=50,
+            )
+        else:
+            response = model_provider.invoke(
+                messages=messages,
+                max_tokens=50,
+            )
         assert isinstance(response, openai.types.chat.ChatCompletion)
+        token_count = len(encoding.encode(response.choices[0].message.content))
```
davesh0812: Why you are doing the count differently from L134?

davesh0812 (commented):
few minor comments

mlrun/datastore/model_provider/huggingface_provider.py
tomerm-iguazio: I do not think it's a good idea, it could be changed by adding tasks, and it can be different objects (it can be a picture for example).
I can add "dict" to return hints, because it's common too.

davesh0812 (approved):

PR #8375, opened 2025-07-27
[Model Monitoring] Fix MonitoringPreProcessor when the inputs/outputs contain str and int/float

davesh0812:
https://iguazio.atlassian.net/browse/ML-10667

### DIFF
--- a/tests/serving/test_tracking.py
+++ b/tests/serving/test_tracking.py
@@ -33,6 +33,7 @@ from mlrun.platforms.iguazio import KafkaOutputStream
 from mlrun.runtimes import ServingRuntime
 from mlrun.serving import Model, ModelRunnerStep, ModelSelector
 from mlrun.serving.states import RootFlowStep, RouterStep
+from mlrun.serving.system_steps import MonitoringPreProcessor
 from tests.serving.test_serving import _log_model
 
 assets_path = str(pathlib.Path(__file__).parent / "assets")
@@ -1056,9 +1057,6 @@ def test_tracked_model_runner_with_error_handler(
         }
 
     _test_graph_structure(server.graph, enable_tracking)
-
-
-from mlrun.serving.system_steps import MonitoringPreProcessor
 
 
 def test_transpose_by_key_with_str():
--- a/tests/serving/test_tracking.py
+++ b/tests/serving/test_tracking.py
@@ -283,20 +283,31 @@ def handle_error(event):
 
 
 class DictOutputModel(Model):
-    def predict(self, body):
+    def predict(self, body, **kwargs):
         body["outputs"] = {}
         for key, value in body["inputs"][self.name].items():
-            body["outputs"][key.replace("f", "o")] = (
-                value + 1 if not isinstance(value, list) else [v + 1 for v in value]
-            )
+            if not isinstance(value, list) and not isinstance(value, str):
+                body["outputs"][key.replace("f", "o")] = value + 1
+            elif not isinstance(value, list) and isinstance(value, str):
+                body["outputs"][key.replace("f", "o")] = value + "_output"
+            elif isinstance(value, list):
+                out_value = []
+                for v in value:
+                    print(f"Processing value: {v}")
+                    if isinstance(v, int):
+                        out_value.append(v + 1)
+                    elif isinstance(v, str):
+                        out_value.append(v + "_output")
+                print(f"Processed output values: {out_value}")
+                body["outputs"][key.replace("f", "o")] = out_value
         return body
 
-    async def predict_async(self, body):
+    async def predict_async(self, body, **kwargs):
         return self.predict(body)
 
 
 class StrDictOutputModel(Model):
-    def predict(self, body):
+    def predict(self, body, **kwargs):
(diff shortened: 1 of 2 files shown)


tests/serving/test_tracking.py
```diff
@@ -283,20 +283,31 @@ def handle_error(event):
...
-                value + 1 if not isinstance(value, list) else [v + 1 for v in value]
-            )
+            if not isinstance(value, list) and not isinstance(value, str):
+                body["outputs"][key.replace("f", "o")] = value + 1
+            elif not isinstance(value, list) and isinstance(value, str):
+                body["outputs"][key.replace("f", "o")] = value + "_output"
+            elif isinstance(value, list):
+                out_value = []
+                for v in value:
+                    print(f"Processing value: {v}")
```
gtopper: If we need it, let's make it a log rather than a print, although it's generally better to avoid chatty unit tests, since we can always run them in debug.

tests/serving/test_tracking.py
```diff
@@ -1037,3 +1056,35 @@ def test_tracked_model_runner_with_error_handler(
...
+        "extra": [123, 80],
+        "time": ["2020-01-01T01:00:00Z", "2020-01-01T02:00:00Z"],
+    }
+    result = MonitoringPreProcessor.transpose_by_key(data)
+
+    expected_result = [
+        [30.0, "Keyboard", 100, 123, "2020-01-01T01:00:00Z"],
+        [6.0, "Mouse", 200, 80, "2020-01-01T02:00:00Z"],
+    ]
+    assert expected_result == result, "Expected result to match the transposed data"
```
gtopper: Here too.

tests/serving/test_tracking.py
```diff
@@ -283,20 +283,31 @@ def handle_error(event):
...
+                        out_value.append(v + 1)
+                    elif isinstance(v, str):
+                        out_value.append(v + "_output")
+                print(f"Processed output values: {out_value}")
+                body["outputs"][key.replace("f", "o")] = out_value
         return body
 
-    async def predict_async(self, body):
+    async def predict_async(self, body, **kwargs):
         return self.predict(body)
```
gtopper:
```suggestion
        return self.predict(body, **kwargs)
```

tests/serving/test_tracking.py
```diff
@@ -283,20 +283,31 @@ def handle_error(event):
...
+                    if isinstance(v, int):
+                        out_value.append(v + 1)
+                    elif isinstance(v, str):
+                        out_value.append(v + "_output")
+                print(f"Processed output values: {out_value}")
+                body["outputs"][key.replace("f", "o")] = out_value
         return body
 
-    async def predict_async(self, body):
+    async def predict_async(self, body, **kwargs):
```
gtopper, line 304: Apparently the signature is wrong elsewhere in the test too (`MyModel`).

tests/serving/test_tracking.py
```diff
@@ -283,20 +283,31 @@ def handle_error(event):
...
+                body["outputs"][key.replace("f", "o")] = value + "_output"
+            elif isinstance(value, list):
+                out_value = []
+                for v in value:
+                    print(f"Processing value: {v}")
+                    if isinstance(v, int):
+                        out_value.append(v + 1)
+                    elif isinstance(v, str):
+                        out_value.append(v + "_output")
+                print(f"Processed output values: {out_value}")
```
gtopper: Here too.

tests/serving/test_tracking.py
```diff
@@ -1037,3 +1056,35 @@ def test_tracked_model_runner_with_error_handler(
...
+        "Price": 30.0,
+        "Product": "Keyboard",
+        "Stock": 100,
+        "extra": 123,
+        "time": "2020-01-01T01:00:00Z",
+    }
+    result = MonitoringPreProcessor.transpose_by_key(data)
+    expected_result = [[30.0, "Keyboard", 100, 123, "2020-01-01T01:00:00Z"]]
+
+    assert expected_result == result, "Expected result to match the transposed data"
```
gtopper: Expected result is actually supposed to be on the right side. Not sure whether this convention is set in stone in pytest (PyCharm thinks so), but it's certainly the convention in this test file.

tests/serving/test_tracking.py
gtopper: I think it would be clearer with a little refactoring:
```suggestion
            if isinstance(value, list):
                out_value = []
                for v in value:
                    print(f"Processing value: {v}")
                    if isinstance(v, int):
                        out_value.append(v + 1)
                    elif isinstance(v, str):
                        out_value.append(v + "_output")
                body["outputs"][key.replace("f", "o")] = out_value
            else:
                if isinstance(value, str):
                    body["outputs"][key.replace("f", "o")] = value + "_output"
                else:
                    body["outputs"][key.replace("f", "o")] = value + 1
```

davesh0812: I will stay with the original as we spoke.
Its a test code

gtopper (approved):

PR #8379, opened 2025-07-27
[Model Monitoring] Redirect MM function deployment to the new flow

danielperezz:
https://iguazio.atlassian.net/browse/ML-9675

replace the calls to deprecated `build_function` with calls to `deploy_function` in the model-monitoring functions deployment

### DIFF
--- a/server/py/services/api/crud/model_monitoring/deployment.py
+++ b/server/py/services/api/crud/model_monitoring/deployment.py
@@ -196,15 +196,19 @@ class MonitoringDeployment:
             fn = self._initial_model_monitoring_stream_processing_function(
                 stream_image=stream_image, parquet_target=parquet_target
             )
-            fn, ready = services.api.utils.functions.build_function(
+            fn = services.api.api.endpoints.nuclio._deploy_function(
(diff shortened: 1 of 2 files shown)


server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -253,15 +257,19 @@ def deploy_model_monitoring_controller(
...
-            fn, ready = services.api.utils.functions.build_function(
+            fn = services.api.api.endpoints.nuclio._deploy_function(
                 db_session=self.db_session,
                 auth_info=self.auth_info,
                 function=fn,
+                project=self.project,
+                name=fn.metadata.name,
+                builder_env=None,
+                client_version=None,
+                client_python_version=None,
```
davesh0812, line 268: Do we know what is the default behavior when sending those params as `None`?
Maybe its better to define it explicitly..

davesh0812 (commented):
Looks Good, I have one question :)

danielperezz:
passed the `test_app` system test

davesh0812 (approved):

PR #8386, opened 2025-07-28
[Artifacts] add has_children to minimal list format

davesh0812:
Add `has_children` to minimal artifact format.
This flag is needed for the UI in order to disable model deletion from the table view.


royischoss (approved):
LGTM maybe add small unit test preventing a change

PR #8388, opened 2025-07-28
[AuthVerifier] Implement IG4 Client Functionality [feature/ig4-authentication]

moranbental:
This PR implements the functionality for `iguazio.v4.Client`, including:

- Ensuring requests include either an `Authorization` header or an` _oauth2_proxy` cookie
- Sending an async request directly to Orca to retrieve user info
- Extracting the username and group IDs
- Creating an AuthInfo object to pass to OPA for verification
- Adding a CookieNames constant
- Updating the `iguazio_client` pytest fixture to support both IG3 and IG4 clients

Verification is covered by unit tests, as well as by testing on IG4 and IG3 systems
https://iguazio.atlassian.net/browse/ML-10294


mlrun/common/schemas/constants.py
```diff
@@ -125,6 +125,11 @@ class AuthorizationHeaderPrefixes:
     bearer = "Bearer "
 
 
+class CookieNames:
+    oauth2_proxy = "_oauth2_proxy"
+    iguazio = "session=j:"
```
rokatyy: please check my comment https://iguazio.atlassian.net/browse/NUC-542 - in some places this cookie is url-encoded and in some it is not, not sure whether it affects mlrun, but if you need to support both, then it won't work for encoded one

quaark: Also - technically `session` is the cookie name. The `j:` is already part of the content and denotes that the rest is JSON

server/py/framework/utils/clients/iguazio/v4.py
```diff
@@ -49,6 +77,40 @@ def _handle_error_response(
     ) -> None:
         raise NotImplementedError()
```
quaark: Implement just raising the error as discussed offline

quaark (changes requested):
Looks good!
Minor

mlrun/common/schemas/constants.py
moranbental: @rokatyy I also checked the flow in IG3 for creating a project, and it works fine

quaark (approved):
Nice!

rokatyy (approved):

TomerShor (approved):
LGTM

PR #8390, opened 2025-07-29
[DB] Harden DSN parser, add exhaustive tests

adamdelman:
**Refactored and relocated DSN parser logic**

- Moved `ParsedDsn` class from `framework/utils/db/utils.py` to `framework/utils/db/dsn.py`, renamed to `Dsn`.
- Refactored parser internals:
  - Improved percent-encoding for user-info components.
  - Stricter validation of dialect, driver, host, port, and database.
  - SQLite-specific path validation.
  - Uses `_encode_credentials()` and `urlsplit()` to avoid breaking on reserved characters.
- Enhanced unit tests:
  - Added coverage for encoded passwords, invalid port values, missing components, SQLite edge cases, and query strings.
https://iguazio.atlassian.net/browse/ML-10704

### DIFF
--- a/server/py/framework/utils/db/dsn.py
+++ b/server/py/framework/utils/db/dsn.py
@@ -1,0 +1,150 @@
+import re
+from typing import Any, Optional, Union
+from urllib.parse import parse_qs, quote, unquote, urlparse
+
+import sqlalchemy
+
+import mlrun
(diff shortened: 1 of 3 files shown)


moranbental (approved):
LGTM

PR #8391, opened 2025-07-29
[Retry] Fix retry logic for early pod deletion

Yacouby:
This PR fixes a bug in the retry flow where retry attempts were not triggered if the initial pod was deleted before the first retry attempt began.
Previously, the system failed to detect this as a retryable failure, causing the run to be marked as failed without retrying.
With this change, the deletion of the first pod is now correctly handled, allowing the retry mechanism to proceed according to the defined retry policy.

Jira:
https://iguazio.atlassian.net/browse/ML-10684

Also should resolve this ticket as it has the same problem:
https://iguazio.atlassian.net/browse/ML-10560

### DIFF
--- a/server/py/services/api/runtime_handlers/base.py
+++ b/server/py/services/api/runtime_handlers/base.py
@@ -642,7 +642,6 @@ class BaseRuntimeHandler(ABC):
                     "status.reason": reason,
                     "status.status_text": message,
                     "status.last_update": now.isoformat(),
-                    "status.end_time": now.isoformat(),
                 }
                 db.update_run(
                     db_session, updates=run_updates, uid=run_uid, project=project
--- a/server/py/services/api/runtime_handlers/base.py
+++ b/server/py/services/api/runtime_handlers/base.py
@@ -1769,6 +1769,8 @@ class BaseRuntimeHandler(ABC):
             elif run_state == RunStates.error:
                 # Try resolving the error reason
                 reason, message = self._resolve_container_error_status(runtime_resource)
+
+                # Check if the run should be retried, and update its status accordingly
                 run_state, message = self._evaluate_run_retry_state(
                     run, reason, message
                 )
--- a/server/py/services/api/runtime_handlers/base.py
+++ b/server/py/services/api/runtime_handlers/base.py
@@ -1929,9 +1929,7 @@ class BaseRuntimeHandler(ABC):
             message = f"Run failed attempt {attempt_number} of {max_retries + 1} with error: {message or reason}"
         elif 0 < max_retries <= retry_count:
             new_state = RunStates.error
-            message = (
-                f"Run failed after {attempt_number} attempts with error: {message or reason}"
-            )
+            message = f"Run failed after {attempt_number} attempts with error: {message or reason}"
         else:
             new_state = RunStates.error
         return new_state, message
--- a/server/py/services/api/runtime_handlers/base.py
+++ b/server/py/services/api/runtime_handlers/base.py
@@ -1919,14 +1919,14 @@ class BaseRuntimeHandler(ABC):
         max_retries = retry_spec.get("count", -1) if retry_spec else -1
         # Run status retry_count may be `None` if the run has never been retried
         retry_count = run.get("status", {}).get("retry_count") or 0
-        attempt_number = retry_count + 1
+        current_attempt = retry_count + 1
 
         if retry_count < max_retries:
             new_state = RunStates.pending_retry
-            message = f"Run failed attempt {attempt_number} of {max_retries + 1} with error: {message or reason}"
+            message = f"Run failed attempt {current_attempt} of {max_retries + 1} with error: {message or reason}"
         elif 0 < max_retries <= retry_count:
             new_state = RunStates.error
-            message = f"Run failed after {attempt_number} attempts with error: {message or reason}"
+            message = f"Run failed after {current_attempt} attempts with error: {message or reason}"
         else:
             new_state = RunStates.error
         return new_state, message
--- a/server/py/services/api/runtime_handlers/base.py
+++ b/server/py/services/api/runtime_handlers/base.py
@@ -633,15 +633,23 @@ class BaseRuntimeHandler(ABC):
                         "possibly it was preempted or evicted. "
                         "Additional details may be available from Kubernetes events."
                     )
-                logger.info(
-                    "Updating run state", run_uid=run_uid, run_state=RunStates.error
-                )
-                run.setdefault("status", {})["state"] = RunStates.error
 
-                run.setdefault("status", {})["reason"] = reason
-                run.setdefault("status", {})["last_update"] = now.isoformat()
-                run.setdefault("status", {})["end_time"] = now.isoformat()
-                db.store_run(db_session, run, run_uid, project)
+                # Check if the run should be retried, and update its status accordingly
+                message = ""
+                run_state, message = self._evaluate_run_retry_state(
+                    run, reason, message
+                )
+                logger.info("Updating run state", run_uid=run_uid, run_state=run_state)
+                run_updates = {
+                    "status.state": run_state,
+                    "status.reason": reason,
+                    "status.status_text": message,
+                    "status.last_update": now.isoformat(),
+                    "status.end_time": now.isoformat(),
+                }
+                db.update_run(
+                    db_session, updates=run_updates, uid=run_uid, project=project
+                )
 
     def _get_runtime_resources(self, label_selector: str, namespace: str):
         """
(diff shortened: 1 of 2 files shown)


server/py/services/api/runtime_handlers/base.py
```diff
@@ -1908,3 +1910,28 @@ def _get_meta(
...
+        run: dict, reason: str, message: str
+    ) -> tuple[str, str]:
+        """
+        Determine if the run should be retried or marked as failed, based on the retry policy and current attempt count.
+        """
+        retry_spec = run.get("spec", {}).get("retry")
+        max_retries = retry_spec.get("count") if retry_spec else -1
+        # Run status retry_count may be `None` if the run has never been retried
+        retry_count = run.get("status", {}).get("retry_count", 0) or 0
+        attempts = retry_count + 1
```
yaelgen:
```suggestion
        next_attempt_index = retry_count + 1
```

server/py/services/api/runtime_handlers/base.py
```diff
@@ -1908,3 +1910,28 @@ def _get_meta(
...
+    def _evaluate_run_retry_state(
+        run: dict, reason: str, message: str
+    ) -> tuple[str, str]:
+        """
+        Determine if the run should be retried or marked as failed, based on the retry policy and current attempt count.
+        """
+        retry_spec = run.get("spec", {}).get("retry")
+        max_retries = retry_spec.get("count") if retry_spec else -1
+        # Run status retry_count may be `None` if the run has never been retried
+        retry_count = run.get("status", {}).get("retry_count", 0) or 0
```
yaelgen:
```suggestion
        retry_spec = run.get("spec", {}).get("retry", {})
        max_retries = retry_spec.get("count", -1) 
        # Run status retry_count may be `None` if the run has never been retried
        retry_count = run.get("status", {}).get("retry_count") or 0
```

yaelgen (changes requested):
Looks good! minor suggestions

server/py/services/api/runtime_handlers/base.py
Yacouby: it is not really the next attempt index, it is the current attempt number, so im changing it to `attempt_number`

server/py/services/api/runtime_handlers/base.py
Yacouby: this :
max_retries = retry_spec.get("count", -1)
will not work, since if `retry_spec` is None, then it will explode

server/py/services/api/runtime_handlers/base.py
```diff
@@ -633,15 +633,23 @@ def _ensure_run_not_stuck_on_non_terminal_state(
...
-                    "Updating run state", run_uid=run_uid, run_state=RunStates.error
-                )
-                run.setdefault("status", {})["state"] = RunStates.error
 
-                run.setdefault("status", {})["reason"] = reason
-                run.setdefault("status", {})["last_update"] = now.isoformat()
-                run.setdefault("status", {})["end_time"] = now.isoformat()
-                db.store_run(db_session, run, run_uid, project)
+                # Check if the run should be retried, and update its status accordingly
+                message = ""
```
rokatyy: can be removed

server/py/services/api/runtime_handlers/base.py
```diff
@@ -633,15 +633,23 @@ def _ensure_run_not_stuck_on_non_terminal_state(
...
+                run_state, message = self._evaluate_run_retry_state(
+                    run, reason, message
+                )
+                logger.info("Updating run state", run_uid=run_uid, run_state=run_state)
+                run_updates = {
+                    "status.state": run_state,
+                    "status.reason": reason,
+                    "status.status_text": message,
+                    "status.last_update": now.isoformat(),
+                    "status.end_time": now.isoformat(),
```
rokatyy: specify timezone `datetime.utcnow()`

server/py/services/api/runtime_handlers/base.py
```diff
@@ -1908,3 +1910,26 @@ def _get_meta(
...
+
+    @staticmethod
+    def _evaluate_run_retry_state(
+        run: dict, reason: str, message: str
+    ) -> tuple[str, str]:
+        """
+        Determine if the run should be retried or marked as failed, based on the retry policy and current attempt count.
+        """
+        retry_spec = run.get("spec", {}).get("retry", {})
+        max_retries = retry_spec.get("count") if retry_spec else -1
```
rokatyy:
```suggestion
        max_retries = retry_spec.get("count", -1) if retry_spec else -1
```

rokatyy (commented):
Minor

server/py/services/api/runtime_handlers/base.py
rokatyy: just noticed this comment, that's right, but with the current implementation if retry spec is not empty but doesn't have `count` field then max_retries will remain `None`

server/py/services/api/runtime_handlers/base.py
Yacouby: why ? if we got to the last else block in `_evaluate_run_retry_state` then the message will not be defined

server/py/services/api/runtime_handlers/base.py
Yacouby: now is defined at the beginning of the function with `now_date()` which is timezone aware : https://github.com/mlrun/mlrun/blob/efc980a37d258341c4494374d918561cef544602/server/py/services/api/runtime_handlers/base.py#L545C9-L545C21

server/py/services/api/runtime_handlers/base.py
rokatyy: @Yacouby empty message could be defined in `_evaluate_run_retry_state` and do not be one of the function parameters

server/py/services/api/runtime_handlers/base.py
```diff
@@ -1908,3 +1907,26 @@ def _get_meta(
...
+        run: dict, reason: str, message: str = ""
+    ) -> tuple[str, str]:
+        """
+        Determine if the run should be retried or marked as failed, based on the retry policy and current attempt count.
+        """
+        retry_spec = run.get("spec", {}).get("retry", {})
+        max_retries = retry_spec.get("count", -1) if retry_spec else -1
+        # Run status retry_count may be `None` if the run has never been retried
+        retry_count = run.get("status", {}).get("retry_count") or 0
+        attempt_number = retry_count + 1
```
rokatyy:
```suggestion
    def _evaluate_run_retry_state(
        run: dict, reason: str,
    ) -> tuple[str, str]:
        """
        Determine if the run should be retried or marked as failed, based on the retry policy and current attempt count.
        """
        retry_spec = run.get("spec", {}).get("retry", {})
        max_retries = retry_spec.get("count", -1) if retry_spec else -1
        # Run status retry_count may be `None` if the run has never been retried
        retry_count = run.get("status", {}).get("retry_count") or 0
        attempt_number = retry_count + 1
        message = ""
```

Yacouby: this will not work, as there is another call for this function, where we pass the message to the function, so with your suggested way you are overriding it to an empty one.
my last fix where i pass empty message as a default one is working fine

rokatyy (approved):
LGTM

liranbg (approved):

server/py/services/api/runtime_handlers/base.py
```diff
@@ -633,15 +633,20 @@ def _ensure_run_not_stuck_on_non_terminal_state(
...
-                db.store_run(db_session, run, run_uid, project)
+                # Check if the run should be retried, and update its status accordingly
+                run_state, message = self._evaluate_run_retry_state(run, reason)
+                logger.info("Updating run state", run_uid=run_uid, run_state=run_state)
+                run_updates = {
+                    "status.state": run_state,
+                    "status.reason": reason,
+                    "status.status_text": message,
+                    "status.last_update": now.isoformat(),
+                    "status.end_time": now.isoformat(),
```
liranbg: this is the responsibility of _update_run_end_time function on sql to do, why do we do it here too?

liranbg (changes requested):

yaelgen (approved):
LGTM

server/py/services/api/runtime_handlers/base.py
```diff
@@ -633,15 +633,19 @@ def _ensure_run_not_stuck_on_non_terminal_state(
...
-                run.setdefault("status", {})["end_time"] = now.isoformat()
-                db.store_run(db_session, run, run_uid, project)
+                # Check if the run should be retried, and update its status accordingly
+                run_state, message = self._evaluate_run_retry_state(run, reason)
+                logger.info("Updating run state", run_uid=run_uid, run_state=run_state)
+                run_updates = {
+                    "status.state": run_state,
+                    "status.reason": reason,
+                    "status.status_text": message,
+                    "status.last_update": now.isoformat(),
```
liranbg: what about `_update_run_updated_time` and `last_update`?

liranbg (changes requested):

liranbg (approved):

PR #8392, opened 2025-07-29
[Application] Set multiple ports to runtime spec

rokatyy:
Jira - https://iguazio.atlassian.net/browse/ML-9752

Adds ability to set multiple ports to sidecar spec. Basically the idea is that internal port will remain with its purpose - being a port to which we redirect traffic through nuclio function. When `application_port` parameter will define all the ports that application should expose. Setters for these values contain special logic to make sure that application port always has internal port in it. 

In addition to that, added `port` to create_api_gateway in case `direct_access` is True, so that user be able to create api gateway on any of the ports of the deployed app.

Tested manually ✅


mlrun/runtimes/nuclio/application/application.py
```diff
@@ -447,6 +522,7 @@ def create_api_gateway(
         :param set_as_default:          Set the API gateway as the default for the application (`status.api_gateway`)
         :param gateway_timeout:         nginx ingress timeout in sec (request timeout, when will the gateway return an
                                         error)
+        :param port:                    The API gateway port
```
GiladShapira94: I will add that it has been used only when `direct_port_access=True`, except for that, it looks good

mlrun/utils/helpers.py
```diff
@@ -2427,3 +2427,8 @@ def get_data_from_path(
     if isinstance(output_data, (int, float)):
         output_data = [output_data]
     return output_data
+
+
+def is_valid_port(port):
+    if port < 0 or port > 65535:
+        raise ValueError("Port must be in the range 0-65535")
```
liranbg: looks like a dup of https://github.com/mlrun/mlrun/blob/0d2bc2ca1db5383c3c549d1fb966249f142b1751/mlrun/runtimes/nuclio/application/application.py#L139-L142
and https://github.com/mlrun/mlrun/blob/af7d53b2c30ea35970f0b71dfa4510a20aacd106/server/py/framework/utils/db/dsn.py#L127

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -431,6 +506,7 @@ def create_api_gateway(
         ssl_redirect: typing.Optional[bool] = None,
         set_as_default: bool = False,
         gateway_timeout: typing.Optional[int] = None,
+        port: typing.Optional[int] = None,
```
liranbg, line 508:
```suggestion
        listening_port: typing.Optional[int] = None,
```

perhaps more self explanatory name for sdk

liranbg (changes requested):
Minor comments and suggestions below

mlrun/runtimes/nuclio/application/application.py
rokatyy: not sure about this as it's more of a port to which we send requests when invoking. While it can be called "listening" in terms of that it's one of the listening ports of app runtime, it might confuse user a bit, so I'd stick with just `port`.

mlrun/utils/helpers.py
rokatyy: Unified these two

liranbg (approved):

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -126,22 +128,68 @@ def __init__(
         self.min_replicas = min_replicas or 1
         self.max_replicas = max_replicas or 1
 
+        # initializing internal application port and application ports
+        self._internal_application_port = None
+        self._application_ports = []
+
+        application_ports = application_ports or []
```
TomerShor, line 135:
```suggestion
        self._application_ports = application_ports or []
```

Then no need for line 148

rokatyy: @TomerShor there is a need for this. `application_ports` is an object with setter/getter, which should be run here. Init of private objects is a must as well as aplpication_ports and internal_application_port setters and getters are connected and we must ensure that both private objects exist before running these methods

TomerShor (approved):

PR #8394, opened 2025-07-30
[Model Monitoring] Fix resolve output/input schema when infer used dict type

royischoss:
Include:
1. resolve schema out of user inputs/outputs dict.
2. added input schema to model endpoints draft.
3. back to updating inputs/outputs schema out of model endpoint in ModelRunnerStep monitoring data

https://iguazio.atlassian.net/browse/ML-10657

### DIFF
--- a/mlrun/serving/states.py
+++ b/mlrun/serving/states.py
@@ -1653,6 +1653,7 @@ class ModelRunnerStep(MonitoredStep):
                 raise mlrun.errors.MLRunInvalidArgumentError("Artifact not found.")
 
         outputs = outputs or self._get_model_output_schema(model_artifact)
+        inputs = inputs or self._get_model_input_schema(model_artifact)
 
         model_artifact = (
             model_artifact.uri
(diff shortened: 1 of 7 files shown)


mlrun/serving/states.py
```diff
@@ -1735,12 +1736,29 @@ def _get_model_output_schema(
...
+        if isinstance(
+            model_artifact,
+            ModelArtifact,
+        ):
+            return [feature.name for feature in model_artifact.spec.inputs]
+        elif isinstance(
+            model_artifact,
+            LLMPromptArtifact,
+        ):
+            _model_artifact = model_artifact.model_artifact
```
davesh0812: This line can cause API call.. can we some how do it once for outputs and inputs?

mlrun/serving/system_steps.py
```diff
@@ -96,15 +98,19 @@ def reconstruct_request_resp_fields(
...
                     "'output_path' was specified if needed"
                 )
-        request = {"inputs": inputs, "id": getattr(event, "id", None)}
-        resp = {"outputs": outputs}
+        request = {
+            "inputs": inputs,
+            "id": getattr(event, "id", None),
+            "input_schema": input_schema or new_input_schema,
+        }
+        resp = {"outputs": outputs, "output_schema": output_schema or new_output_schema}
```
davesh0812, line 106: Not the opposite
 ```suggestion
        resp = {"outputs": outputs, "output_schema": new_output_schema or output_schema}
```

Same for inputs.

davesh0812 (commented):
Nice! few comments

mlrun/serving/system_steps.py
royischoss: I think it better this way allowing the user priority and allow us to fail him and catch bugs

royischoss:
`test_monitoring_with_model_runner_dict_infer` passed @davesh0812

davesh0812 (approved):

PR #8205, opened 2025-06-25
[DB] Properly create DB partitions when creating the DB from scratch

adamdelman:
- Introduced `RangePartitioner` with dialect-specific implementations for MySQL and PostgreSQL.
- Added `after_create` listener on `AlertActivation` table to configure partitioning at creation time.
- Partitioning is based on the `PARTITION_INTERVAL` env var (default: `YEARWEEK`).

https://iguazio.atlassian.net/browse/ML-10341
https://iguazio.atlassian.net/browse/ML-7954

PR #8397, opened 2025-07-30
[Migrations] Fix system id being populated before migrations run

adamdelman:
https://iguazio.atlassian.net/browse/ML-10728

### DIFF
--- a/server/py/services/api/initial_data.py
+++ b/server/py/services/api/initial_data.py
@@ -84,8 +84,7 @@ def init_data(
             engine,
             perform_migrations_if_needed,
         )
-    # initialize system id
-    framework.db.session.run_function_with_new_db_session(func=_init_system_id)
+
 
     mlrun.utils.logger.info("Initial data created")
 
@@ -118,6 +117,8 @@ def _initialize_db_from_scratch(
         version=str(latest_data_version),
     )
     mlrun.mlconf.httpdb.state = mlrun.common.schemas.APIStates.online
+    # initialize system id
+    framework.db.session.run_function_with_new_db_session(func=_init_system_id)
 
 
 def _migrate_existing_data(
@@ -289,7 +290,8 @@ def _perform_data_migrations(db_session: sqlalchemy.orm.Session):
 def _add_initial_data(db_session: sqlalchemy.orm.Session):
     db = framework.db.sqldb.db.SQLDB()
     _add_data_version(db, db_session)
-
+    # initialize system id
+    framework.db.session.run_function_with_new_db_session(func=_init_system_id)
 
 def _add_default_hub_source_if_needed(
     db: framework.db.sqldb.db.SQLDB, db_session: sqlalchemy.orm.Session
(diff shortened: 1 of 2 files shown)


liranbg (commented):
Can add integ tests to verify the flow?

Yacouby (commented):
I do not fully understand the solution.. before doing the refactoring to the db flow ( that were done in this PR: https://github.com/mlrun/mlrun/pull/8205 ) the initialization to the system id was done after `_perform_data_migrations` was called and it was working fine. this was the old code :

```
try:
        if is_migration_from_scratch or is_migration_needed:
            try:
                _perform_schema_migrations(alembic_util)
                framework.db.init_db.init_db()
                _add_initial_data(db_session)
                _perform_data_migrations(db_session)
            except Exception:
                state = mlrun.common.schemas.APIStates.migrations_failed
                logger.warning("Migrations failed, changing API state", state=state)
                config.httpdb.state = state
                raise

        # initialize system id
        _init_system_id(db_session)

```
maybe im missing something that was introduced in the mentioned PR above

adamdelman:
> I do not fully understand the solution.. before doing the refactoring to the db flow ( that were done in this PR: #8205 ) the initialization to the system id was done after `_perform_data_migrations` was called and it was working fine. this was the old code :
> 
> ```
> try:
>         if is_migration_from_scratch or is_migration_needed:
>             try:
>                 _perform_schema_migrations(alembic_util)
>                 framework.db.init_db.init_db()
>                 _add_initial_data(db_session)
>                 _perform_data_migrations(db_session)
>             except Exception:
>                 state = mlrun.common.schemas.APIStates.migrations_failed
>                 logger.warning("Migrations failed, changing API state", state=state)
>                 config.httpdb.state = state
>                 raise
> 
>         # initialize system id
>         _init_system_id(db_session)
> ```
> 
> maybe im missing something that was introduced in the mentioned PR above

Before, we would always run migrations.
Now we only run migrations if there was any existing data/tables, otherwise we create the schema from scratch using sqlalchemy, and this flow was missing setting the system id

Yacouby (commented):
if we got inside this block: https://github.com/mlrun/mlrun/blob/0a9349e3f08426fc4f7eb6f152a65efdf56f00e5/server/py/services/api/initial_data.py#L141
then we will not initialize the system id, will this be valid ?
so for example if we had an old system that didn’t yet have a system ID we’ll skip the system ID initialization in this case

adamdelman:
> if we got inside this block:
> 
> https://github.com/mlrun/mlrun/blob/0a9349e3f08426fc4f7eb6f152a65efdf56f00e5/server/py/services/api/initial_data.py#L141
> 
> 
> then we will not initialize the system id, will this be valid ?
> so for example if we had an old system that didn’t yet have a system ID we’ll skip the system ID initialization in this case

We only get inside this block (`perform_migrations_if_needed = False`) from `init_data()` which calls `framework.db.session.run_function_with_new_db_session(func=_init_system_id)` directly.

liranbg (approved):

PR #8401, opened 2025-07-31
[Project] Enable get_current_project in nuclio and job runtimes

rokatyy:
Enables get_current_project in nuclio and job runtimes, and basically in any runtime that doesn't have pipeline_context. 

tested by eval function:
image

Jira - https://iguazio.atlassian.net/browse/ML-10612

Added system test, run with custom mlrun image built with the latest changes:
image

### DIFF
--- a/mlrun/__init__.py
+++ b/mlrun/__init__.py
@@ -168,8 +168,23 @@ def set_environment(
 
 
 def get_current_project(silent: bool = False) -> Optional[MlrunProject]:
-    project_from_env = environ.get(MLRUN_ACTIVE_PROJECT, None)
-    project = pipeline_context.project or project_from_env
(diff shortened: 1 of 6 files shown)


liranbg (approved):

liranbg (commented):
While approving the logic - worth adding to an existing system test the usage of such functionality.
in addition, could we add UT coverage?

rokatyy:
@liranbg agree reg system tests. UT coverage is not really possible as it would simply test load_project and not the full flow

yaelgen (approved):
Looking good, agree with Liran on sys tests coverage

rokatyy:
@liranbg @yaelgen added system test + updated description

liranbg (approved):

PR #8405, opened 2025-08-03
[Launcher] Raise error when retry is set for dask job

Yacouby:
This PR enhances validation in `_validate_run()` to prevent configuring retries for unsupported runtimes.
For the case of dask job, there was a missing param passing to the run function ( the retry param )

So added a check that raises `MLRunInvalidArgumentError` if retry is configured, and the runtime kind is not supported, and the runtime is not local (to preserve backward-compatible behavior in local runs).

Jira:
https://iguazio.atlassian.net/browse/ML-10625

### DIFF
--- a/mlrun/launcher/base.py
+++ b/mlrun/launcher/base.py
@@ -159,7 +159,8 @@ class BaseLauncher(abc.ABC):
 
         # Raise an error if retry is configured for a runtime that doesn't support retries
         if (
-            run.spec.retry.count
+            not mlrun.runtimes.RuntimeKinds.is_local_runtime(runtime.kind)
+            and run.spec.retry.count
             and runtime.kind not in mlrun.runtimes.RuntimeKinds.retriable_runtimes()
         ):
             raise mlrun.errors.MLRunInvalidArgumentError(
--- a/mlrun/launcher/base.py
+++ b/mlrun/launcher/base.py
@@ -157,6 +157,16 @@ class BaseLauncher(abc.ABC):
         ]:
             mlrun.utils.helpers.warn_on_deprecated_image(image)
 
+        # Raise an error if retry is configured for a runtime that doesn't support retries
+        if (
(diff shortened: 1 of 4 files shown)


mlrun/launcher/base.py
```diff
@@ -157,6 +157,17 @@ def _validate_run(
         ]:
             mlrun.utils.helpers.warn_on_deprecated_image(image)
 
+        # Raise an error if retry is configured for a runtime that doesn't support retries
+        if (
+            not mlrun.runtimes.RuntimeKinds.is_local_runtime(runtime.kind)
```
liranbg, line 164: why do we need to first condition to happen? isnt the rest are enough?
```
run.spec.retry.count
            and runtime.kind not in mlrun.runtimes.RuntimeKinds.retriable_runtimes()
```

Yacouby: because in case it was a local run then we want to ignore the retry and continue. it was done in this PR : https://github.com/mlrun/mlrun/pull/8348

tests/launcher/test_local.py
```diff
@@ -300,3 +300,18 @@ def test_validate_run_retries_warning(logs_stream):
         in logs_stream.getvalue()
     )
     assert run.spec.retry.count == 0
+
+
+def test_validate_run_retries_invalid_runtime():
```
TomerShor, line 305: Add a test case where the local runtime does not raise

Yacouby: we already have one that does that ( `test_validate_run_retries_warning` )

mlrun/launcher/base.py
liranbg: Worth writing it explicitly

liranbg (commented):
Looking good. Worth updating the comment for clarity

TomerShor (approved):

alonmr (approved):

PR #8406, opened 2025-08-03
[Serving] Fix deploying `ModelRunnerStep` in child function

davesh0812:
In this we make sure the raiser and the MRS are deployed on the same function.

resolved https://iguazio.atlassian.net/browse/ML-10732

### DIFF
--- a/mlrun/datastore/__init__.py
+++ b/mlrun/datastore/__init__.py
@@ -39,6 +39,7 @@ __all__ = [
 from urllib.parse import urlparse
 
 import fsspec
+import storey
 
 import mlrun.datastore.wasbfs
 from mlrun.datastore.datastore_profile import (
@@ -168,11 +169,12 @@ def get_stream_pusher(stream_path: str, **kwargs):
             raise ValueError(f"unsupported stream path {stream_path}")
 
 
-class _DummyStream:
+class _DummyStream(storey.Flow):
     """stream emulator for tests and debug"""
 
     def __init__(self, event_list=None, **kwargs):
         self.event_list = event_list or []
+        super().__init__()
 
     def push(self, data, **kwargs):
         if not isinstance(data, list):
@@ -180,3 +182,13 @@ class _DummyStream:
         for item in data:
             logger.info(f"dummy stream got event: {item}, kwargs={kwargs}")
             self.event_list.append(item)
+
+    async def _do(self, event):
+        if event is storey.flow._termination_obj:
+            return await self._do_downstream(storey.flow._termination_obj)
+        if hasattr(event, "body"):
+            event = event.body
+            if not isinstance(event, list):
+                event = [event]
+            for item in event:
+                self.event_list.append(item)
--- a/mlrun/datastore/__init__.py
+++ b/mlrun/datastore/__init__.py
@@ -169,7 +169,7 @@ def get_stream_pusher(stream_path: str, **kwargs):
             raise ValueError(f"unsupported stream path {stream_path}")
 
 
-class _DummyStream(storey.Flow):
+class _DummyStream(storey.MapClass):
     """stream emulator for tests and debug"""
 
     def __init__(self, event_list=None, **kwargs):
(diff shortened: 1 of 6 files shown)


mlrun/datastore/__init__.py
```diff
@@ -168,15 +169,26 @@ def get_stream_pusher(stream_path: str, **kwargs):
...
         if not isinstance(data, list):
             data = [data]
         for item in data:
             logger.info(f"dummy stream got event: {item}, kwargs={kwargs}")
             self.event_list.append(item)
+
+    async def _do(self, event):
+        if event is storey.flow._termination_obj:
+            return await self._do_downstream(storey.flow._termination_obj)
+        if hasattr(event, "body"):
```
gtopper: Any reason we need to check this here? It's expected that the event will have a body field.

mlrun/datastore/__init__.py
```diff
@@ -168,15 +169,26 @@ def get_stream_pusher(stream_path: str, **kwargs):
             raise ValueError(f"unsupported stream path {stream_path}")
 
 
-class _DummyStream:
+class _DummyStream(storey.Flow):
     """stream emulator for tests and debug"""
 
     def __init__(self, event_list=None, **kwargs):
         self.event_list = event_list or []
```
gtopper, line 176: Not sure how serious we are about this `_DummyStream`, but state is normally set in `def _init`, which matters when the flow is reused (terminated and then run again).

davesh0812: Its only for simple  tests flow, not for users.

mlrun/datastore/__init__.py
davesh0812: I did it before the check on the `_termination_obj`... maybe it can be removed but it safer like that.
And as I mention on the next comment it only for simple test flows, not for users.

royischoss (approved):
NICE

mlrun/datastore/__init__.py
gtopper: So long as these tests don't involve restarting the flow it shouldn't matter.

mlrun/datastore/__init__.py
gtopper: I think in this case it should be easier to define a simple `def do`, not inherit `storey.Flow`, and not handle `storey.flow._termination_obj`. MLRun will wrap it in a `MapClass`. The low level functionality of `storey.Flow` isn't really relevant here.

davesh0812: Its not worked to me.. I have to use `storey.Flow` or to add a lot of attributes such as `_legal_first_step`

tests/serving/test_tracking.py
```diff
@@ -196,6 +196,41 @@ def test_child_function_tracking(
...
+        input_path="n",
+        result_path="n",
+        raise_error=False,
+        inc=1,
+    )
+    graph.to(">>", name="in", path="dummy://in").to(
+        model_runner_step, function="c1"
+    ).to(">>", name="out", path="dummy://out")
+    fn.set_tracking("dummy://", enable_tracking=True)
+    fn.add_child_function("c1", f"{assets_path}/child_function.py", r"mlrun\mlrun")
```
gtopper:
```suggestion
    fn.add_child_function("c1", f"{assets_path}/child_function.py", "mlrun/mlrun")
```

mlrun/datastore/__init__.py
```diff
@@ -168,15 +169,22 @@ def get_stream_pusher(stream_path: str, **kwargs):
             raise ValueError(f"unsupported stream path {stream_path}")
 
 
-class _DummyStream:
+class _DummyStream(storey.MapClass):
     """stream emulator for tests and debug"""
 
     def __init__(self, event_list=None, **kwargs):
         self.event_list = event_list or []
+        super().__init__()
```
gtopper:
```suggestion
        super().__init__(**kwargs)
```

gtopper (approved):
✅  Please see minor comments.

royischoss (approved):

PR #8408, opened 2025-08-03
[User secrets] Create API Endpoint to Store or Update Secret Tokens [feature/ig4-authentication]

moranbental:
This PR introduces a new API endpoint to store or update multiple offline tokens:
`PUT /api/v1/user-secrets/tokens`

For each token in the request, the following validations and operations are performed:
1. Token Name Validation – Ensure each token has a non-empty and unique name within the request.
2. JWT Decoding – Decode the offline token and verify it contains the required claims: sub (user ID) and exp (expiration).
3. User Ownership Validation – Confirm that the sub in the offline token matches the authenticated user ID 
4. Token Verification – Use the Iguazio SDK to validate all tokens via the `refresh_access_tokens` call. This will be enabled in a separate PR (https://github.com/mlrun/mlrun/pull/8443).
5. Secret Storage – (Not implemented in this PR) Creation and updating of Kubernetes secrets will be handled in a separate PR.

Response Structure – Return lists of token names categorized by action: created_tokens, updated_tokens, and skipped_tokens.

This functionality has been verified only via unit tests for now

Additional Changes
1. Renamed `api/endpoints/secrets.py` → `api/endpoints/project_secrets.py`.
2. Extracting the user_id from Orca’s session verification endpoint and saving it in the AuthInfo

https://iguazio.atlassian.net/browse/ML-10487


server/py/framework/utils/clients/iguazio/v4.py
```diff
@@ -26,6 +27,41 @@
...
+    def refresh_access_token(self, token_name: str, token: str) -> None:
+        """
+        Refresh the access token using the Iguazio client to validate the offline token.
+
+        :param token_name: Logical name of the token (used for error messages)
+        :param token: The offline token string
+        :raises mlrun.errors.MLRunUnauthorizedError: If the token is invalid or expired
+        """
+        try:
+            self._client.refresh_access_token(token)
```
quaark: is this the new iguazio SDK?
If so, we need to expose this method to the top level client as currently it isn't. Could you please open a ticket for this?

server/py/framework/utils/clients/iguazio/v4.py
```diff
@@ -26,6 +27,41 @@
...
+        self, secret_tokens: list[mlrun.common.schemas.SecretToken]
+    ) -> None:
+        """
+        Refresh all offline tokens using the Iguazio client to validate them.
+
+        :param secret_tokens: List of SecretToken objects
+        :raises mlrun.errors.MLRunUnauthorizedError: If any token is invalid or expired
+        """
+        try:
+            self._client.refresh_access_tokens(secret_tokens)
```
quaark: Same

server/py/framework/utils/singletons/k8s.py
```diff
@@ -1006,6 +1006,15 @@ def _init_k8s_config(self, log=True):
...
         return hashlib.sha224(access_key.encode()).hexdigest()
 
+    def create_or_update_user_token_secret(
+        self,
+        username: str,
+        token_name: str,
+        token: str,
+        expiration: int,
+    ) -> mlrun.common.schemas.SecretEventActions:
+        raise NotImplementedError()
```
quaark, line 1017: Future PR? Add a TODO then

server/py/services/api/crud/secrets.py
```diff
@@ -419,6 +422,112 @@ def get_project_secret(
...
+        if sub != expected_user_id:
+            raise mlrun.errors.MLRunAccessDeniedError(
+                f"Offline token '{token_name}' does not belong to the authenticated user"
+            )
+
+        return decoded
+
+    def _decode_offline_token(self, token_name: str, token: str) -> dict:
+        try:
+            return jwt.decode(token, options={"verify_signature": False})
```
quaark, line 507: Maybe add a comment here explaining that the token is already verified so there's no need to verify it here

quaark (changes requested):
Looks good!
A few minor comments

server/py/framework/utils/clients/iguazio/v4.py
rokatyy: @quaark https://github.com/iguazio/orca/pull/230 implement this function

server/py/services/api/crud/secrets.py
```diff
@@ -419,6 +422,112 @@ def get_project_secret(
...
+            elif action == mlrun.common.schemas.SecretEventActions.updated:
+                updated_tokens.append(token_name)
+            elif action == mlrun.common.schemas.SecretEventActions.skipped:
+                skipped_tokens.append(token_name)
+
+        return mlrun.common.schemas.StoreSecretTokensResponse(
+            created_tokens=created_tokens,
+            updated_tokens=updated_tokens,
+            skipped_tokens=skipped_tokens,
+        )
```
rokatyy: nitpicking - can be simplified a bit
```suggestion
  token_actions = defaultdict(list)
  
  for token_name, token_data in valid_tokens.items():
      action = self.secrets_provider.create_or_update_user_token_secret(
          username=authenticated_username,
          token_name=token_name,
          token=token_data["raw_token"],
          expiration=token_data["exp"],
      )
      token_actions[action].append(token_name)
  
  return StoreSecretTokensResponse(
      created_tokens=token_actions[SecretEventActions.created],
      updated_tokens=token_actions[SecretEventActions.updated],
      skipped_tokens=token_actions[SecretEventActions.skipped],
  )

```

server/py/services/api/crud/secrets.py
```diff
@@ -419,6 +422,112 @@ def get_project_secret(
...
+            )
+            decoded["raw_token"] = token.token  # Add original token string
+            result[token.name] = decoded
+
+        return result
+
+    def _validate_token_name(self, token_name: str, seen_names: set):
+        if not token_name or token_name in seen_names:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"Invalid or duplicate token name '{token_name}' found in request payload"
```
rokatyy, line 489: why failing on duplication, maybe we can just remove it from list or are there any side effects?

server/py/framework/utils/clients/iguazio/v4.py
rokatyy: now it will return the new token, so probably worth wrapping the returned object into the `SecretToken` and returning it. By the way, why passing `token_name` and `token` as a separate params insead of `SecretToken`?

rokatyy (commented):
Nice and clean!

Added some comments

server/py/services/api/crud/secrets.py
moranbental: @rokatyy but how do you know which one to remove? Also, this flow is called from the sync operation, so I think it's better to fail the request.
@quaark WDYT?

server/py/framework/utils/clients/iguazio/v4.py
moranbental: @rokatyy Why do we need to return the token? The token returned in the response is an access token, but in SecretToken it’s an offline token.

rokatyy: @moranbental right, thanks for clarifying that. So maybe we need to classes - for offline token and for access one. I think it will really solve the confusion between these two. wdyt?

server/py/services/api/crud/secrets.py
rokatyy: The last one may take precedence, but indeed it's more of a product-related decision

server/py/framework/utils/clients/iguazio/v4.py
moranbental: @rokatyy So for now, I'm not using the access token at all,  the SecretToken is actually only for the offline token. Maybe it would be better to rename it to SecretOfflineToken? But perhaps in the future it might not be limited to offline tokens? 
@quaark Your thoughts here, please?

server/py/services/api/crud/secrets.py
quaark: Personally I feel a failure is better and let the user decide which to remove from their local file before trying again

server/py/framework/utils/clients/iguazio/v4.py
quaark: This would be quite a large design change to add support for access tokens as well, and might not be relevant especially since we are considering making access token TTLs even shorter than they are now.
Regarding the name itself - I'm not against SecretOfflineToken that strongly, but I feel SecretToken is cleaner

server/py/services/api/crud/secrets.py
```diff
@@ -419,6 +423,117 @@ def get_project_secret(
...
+
+        valid_tokens = self._validate_and_decode_offline_tokens(
+            secret_tokens, authenticated_user_id
+        )
+
+        # TODO: move init iguazio_client
+        # iguazio_client = framework.utils.clients.iguazio.v4.Client()
+
+        # TODO: Enable this once it is implemented in the Iguazio client.
+        # iguazio_client.refresh_access_tokens(secret_tokens)
```
quaark, line 452: this should be before `_validate_and_decode_offline_tokens` since you use `verify_signature=False` there, so first try to use it and only if it works - then decode it.

quaark (commented):
Looks good!
Just a minor comment

server/py/services/api/crud/secrets.py
moranbental: @quaark The whole point was to perform the check on the mlrun side first and fail fast before sending the request to orca. Then, when decoding, we use `verify_signature=False` because the signature will be verified on the orca side later.

server/py/framework/utils/clients/iguazio/v4.py
```diff
@@ -26,6 +27,42 @@
...
+        """
+        Refreshes the access token by validating the provided offline token using the Iguazio client.
+
+        :param secret_token: SecretToken object containing the token name and the offline token string.
+        :raises mlrun.errors.MLRunUnauthorizedError: If the offline token is invalid or expired.
+        """
+        try:
+            self._client.refresh_access_token(secret_token.token)
+        except Exception as exc:
+            raise mlrun.errors.MLRunUnauthorizedError(
```
TomerShor: What if another exception occurred (e.g. timeout) - do we still return Unauthorized?

server/py/services/api/api/endpoints/user_secrets.py
```diff
@@ -0,0 +1,44 @@
...
+from fastapi.concurrency import run_in_threadpool
+from sqlalchemy.orm import Session
+
+import mlrun.common.schemas
+
+import framework.api.deps
+import framework.utils.auth.verifier
+import services.api.crud
+
+router = fastapi.APIRouter(prefix="/user-secrets/tokens")
```
TomerShor: The prefix here is valid for all endpoint handlers in this file.
if we will support other user secrets that are not tokens, this will not be true.
I suggest having `prefix="/user-secrets"` here, and in the handler add `tokens`

server/py/services/api/crud/secrets.py
```diff
@@ -419,6 +423,117 @@ def get_project_secret(
...
+
+        if not sub or not exp:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"Offline token '{token_name}' is missing required claims (exp/sub)"
+            )
+
+        if sub != expected_user_id:
+            raise mlrun.errors.MLRunAccessDeniedError(
+                f"Offline token '{token_name}' does not belong to the authenticated user"
+            )
```
TomerShor: As discussed offline - there is an exception to this case if you are a system admin. Should check permissions if we got here.

server/py/services/api/tests/unit/api/test_user_secrets.py
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+API_USER_SECRETS_PATH = "/user-secrets/tokens"
```
TomerShor: Why is this needed in this PR?

TomerShor (changes requested):

server/py/services/api/crud/secrets.py
moranbental: No, we are not enabling storing or updating tokens for the system admin, only listing and revoking are allowed.
This is the flow for storing tokens

server/py/services/api/tests/unit/api/test_user_secrets.py
moranbental: Not really needed, I just added it for future tests in the API

server/py/services/api/crud/secrets.py
moranbental: fixed

server/py/framework/utils/clients/iguazio/v4.py
```diff
@@ -26,6 +25,33 @@
...
+        self, secret_token: mlrun.common.schemas.SecretToken
+    ) -> None:
+        """
+        Refreshes the access token by validating the provided offline token using the Iguazio client.
+
+        :param secret_token: SecretToken object containing the token name and the offline token string.
+        :raises mlrun.errors.MLRunUnauthorizedError: If the offline token is invalid or expired.
+        """
+        # TODO: Implement this method once it is available in the Iguazio package
+        pass
```
quaark, line 41: It is now implemented

server/py/framework/utils/clients/iguazio/v4.py
```diff
@@ -26,6 +25,33 @@
...
+        self, secret_tokens: list[mlrun.common.schemas.SecretToken]
+    ) -> None:
+        """
+        Refresh all offline tokens using the Iguazio client to validate them.
+
+        :param secret_tokens: List of SecretToken objects
+        :raises mlrun.errors.MLRunUnauthorizedError: If any token is invalid or expired
+        """
+        # TODO: Implement this method once it is available in the Iguazio package
+        pass
```
quaark, line 53: same

server/py/services/api/crud/secrets.py
quaark: same - you can now add it in

quaark (changes requested):
Awesome!
I think you may now add the usage of the iguazio sdk - as the relevant methods are implemented.

server/py/framework/utils/clients/iguazio/v4.py
moranbental: I didn’t add it because no release rc has been published yet, so it will be handled in a different story/pr — https://github.com/mlrun/mlrun/pull/8443.
Also, there is an issue in the CI with importing Iguazio, which is why I added it in a separate pr.

server/py/framework/utils/clients/iguazio/v4.py
moranbental: no, refresh_access_tokens (for many tokens) was not implemented yet, only the refresh_access_token

quaark (approved):

PR #8409, opened 2025-08-03
[Model Monitoring] Make sure MEP will create only for the relevant function (child or parent)

davesh0812:
https://iguazio.atlassian.net/browse/ML-10731


royischoss (approved):
LGTM :)

PR #8413, opened 2025-08-04
[Model Monitoring] Fix change last_analyze resulotion

royischoss:
fix for https://iguazio.atlassian.net/browse/ML-10664

include:
1. Changing schedule file timestamp types to float including chief file.
2. Changing last_analyzed type to float in the controller 
3. Making interval to be of the kind of [start, end) not including the end point preventing data to be analyzed in two windows.


royischoss:
Passed `test_app_flow` , `test_serving_as_a_job` , `test_histogram_app`

mlrun/model_monitoring/controller.py
```diff
@@ -616,7 +620,10 @@ def model_endpoint_process(
                                 endpoint_id=endpoint_id,
                             )
                             self._push_to_applications(
-                                start_infer_time=start_infer_time,
+                                start_infer_time=start_infer_time
+                                - datetime.timedelta(
```
Eyal-Danieli, line 624: Just add here a small comment why we are doing it (to avoid data loss when the apps read the parquet files)

mlrun/model_monitoring/db/_schedules.py
```diff
@@ -208,15 +208,19 @@ def get_endpoint_last_request(self, endpoint_uid: str) -> Optional[int]:
             return None
```
Eyal-Danieli, line 208: See `get_endpoint_last_request`, I think that after your changes it returns float not int right? because `self._schedules` is not enriched with float timestamp. Please update the docstring accordingly

Eyal-Danieli (commented):
Nice!
2 small comments, not something about the logic :)

Eyal-Danieli (approved):
LGTM

davesh0812 (approved):

PR #8414, opened 2025-08-04
[Bugfix] Fix project deletion flow for projects with many artifacts

adamdelman:
https://iguazio.atlassian.net/browse/ML-10668

### DIFF
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -19,8 +19,10 @@ import hashlib
 import inspect
 import pathlib
 import re
+import time
 import typing
 import urllib.parse
+from contextlib import nullcontext
 from copy import deepcopy
 from datetime import datetime, timedelta, timezone
 from typing import Any, Optional, Union
@@ -148,6 +150,10 @@ conflict_messages = [
     "(pymysql.err.IntegrityError) (1062",
     "(pymysql.err.IntegrityError) (1586",
 ]
+
+
+def _next_sleep(prev: float, factor: float = 2.0, max_sleep: float = 5.0) -> float:
+    return min(prev * factor, max_sleep) if prev else 0.05
 
 
 def retry_on_conflict(function):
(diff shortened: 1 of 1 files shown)


server/py/framework/db/sqldb/db.py
```diff
@@ -150,6 +152,10 @@
 ]
 
 
+def _next_sleep(prev: float, factor: float = 2.0, max_sleep: float = 5.0) -> float:
```
liranbg, line 155: doesnt need to be on a function, could be inline

server/py/framework/db/sqldb/db.py
```diff
@@ -3194,52 +3200,86 @@ def skip_deletion():
...
+        Delete rows from *table* that satisfy *where_clause* in PK-ordered ranges.
+
+        * Retries the current range on MySQL 1205 (lock-wait timeout) or 1213 (deadlock),
+          backing off exponentially inside the function.
+        * Opens its own transaction only if the caller didn't start one, so it can be
+          used from inside existing `with session.begin():` blocks.
         """
-        last_id = 0
-        total_deleted = 0
+        pk = table.id
```
liranbg, line 3214: use meaninful names / why not uding `table.id` directly where needed?

server/py/framework/db/sqldb/db.py
```diff
@@ -3194,52 +3200,86 @@ def skip_deletion():
...
-            )
+        logger.info(
+            "starting_batched_range_delete",
+            table=str(table),
+            batch_size=batch_size,
+        )
 
-            if not ids_to_delete:
-                break
+        outer_cm = nullcontext() if session.in_transaction() else session.begin()
```
liranbg, line 3224: what's `cm`? use meaningful names

server/py/framework/db/sqldb/db.py
```diff
@@ -3194,52 +3200,86 @@ def skip_deletion():
...
 
-            id_values = [row.id for row in ids_to_delete]
+        with outer_cm:
+            while True:
+                start_id = session.scalar(
+                    select(pk).where(where_clause).order_by(pk).limit(1)
```
liranbg, line 3229: why's the order_by on the table id here?

server/py/framework/db/sqldb/db.py
```diff
@@ -3194,52 +3200,86 @@ def skip_deletion():
...
+                            start_id=start_id,
+                            end_id=end_id,
+                            mysql_code=mysql_code,
+                            sleep_seconds=sleep_secs,
+                            total_deleted=total_deleted,
+                        )
+                        time.sleep(sleep_secs)
+
+        logger.info(
+            "finished batched range delete",
```
liranbg, line 3279:
```suggestion
            "Finished batched range delete",
```

server/py/framework/db/sqldb/db.py
```diff
@@ -3194,52 +3200,86 @@ def skip_deletion():
...
+                    except sqlalchemy.exc.OperationalError as exc:
+                        # MySQL error codes: 1205 = lock-wait timeout, 1213 = deadlock
+                        mysql_code = getattr(exc.orig, "args", [None])[0]
+                        if mysql_code not in (1205, 1213):
+                            raise
+
+                        session.rollback()
+                        sleep_secs = _next_sleep(sleep_secs)
+                        logger.warning(
+                            "range delete lock contention",
```
liranbg, line 3269:
```suggestion
                            "Range delete lock contention, retrying...",
```

server/py/framework/db/sqldb/db.py
```diff
@@ -3194,52 +3200,86 @@ def skip_deletion():
...
+                            end_id=end_id,
+                            rows_deleted=rows,
+                            total_deleted=total_deleted,
+                        )
+                        break
+
+                    except sqlalchemy.exc.OperationalError as exc:
+                        # MySQL error codes: 1205 = lock-wait timeout, 1213 = deadlock
+                        mysql_code = getattr(exc.orig, "args", [None])[0]
+                        if mysql_code not in (1205, 1213):
```
liranbg, line 3263: This is a fragile code. in case `sqlalchemy` changes the structure of the returnd error, you will learn it only on production as the current test coverage wont check it.

what I suggest is to add a ut coverage that simulates few iterations of raising lock-wait or so and see the retrying kicks in.

server/py/framework/db/sqldb/db.py
```diff
@@ -3194,52 +3200,86 @@ def skip_deletion():
...
+                            raise
+
+                        session.rollback()
+                        sleep_secs = _next_sleep(sleep_secs)
```
liranbg, line 3267: I think the function under while True can be wrapped within `retry_until_successful` while instead of its `_timeout_exceeded` you can set a custom function for verification - wdyt?

liranbg (changes requested):

PR #8418, opened 2025-08-05
[Model Monitoring] batch inference fails with  'StatementError('(builtins.ValueError) badly formed hexadecimal UUID string')'

danielperezz:
https://iguazio.atlassian.net/browse/ML-10551

1. Update the documentation of `get_or_create_model_endpoint` end `record_results` to clarify that the `endpoint_id` passed must belong to an existing model endpoint—otherwise, a new UID will be generated anyway.

2. Add a validation on the client side to return a clearer, more indicative error earlier in case the provided `endpoint_id` is not a valid UUID (also in case the user hasn't provided sufficient arguments) and fix the tests that's been affected by this change.

3. Fix `test_model_endpoint_with_many_features` so it doesn’t pass the endpoint_id argument unnecessarily.

### DIFF
--- a/tests/system/alerts/test_alerts.py
+++ b/tests/system/alerts/test_alerts.py
@@ -190,7 +190,7 @@ class TestAlerts(TestMLRunSystem):
             project=self.project.metadata.name,
             model_endpoint_name="test-endpoint",
             context=mlrun.get_or_create_ctx("demo"),
-            endpoint_id=uuid.uuid4().hex # either endpoint-id or function_name and function_tag must be provided
+            endpoint_id=uuid.uuid4().hex,  # either endpoint-id or function_name and function_tag must be provided
         )
 
         # waits for the writer function to be deployed
--- a/tests/system/alerts/test_alerts.py
+++ b/tests/system/alerts/test_alerts.py
@@ -15,6 +15,7 @@
 import json
 import time
 import typing
+import uuid
 
 import deepdiff
 import pytest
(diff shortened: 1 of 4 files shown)


tests/system/alerts/test_alerts.py
```diff
@@ -184,10 +185,12 @@ def test_drift_detection_alert(self):
         nuclio_function_url = notification_helpers.deploy_notification_nuclio(
             self.project, self.image
         )
+        # generate a new model-endpoint
         model_endpoint = mlrun.model_monitoring.api.get_or_create_model_endpoint(
             project=self.project.metadata.name,
             model_endpoint_name="test-endpoint",
             context=mlrun.get_or_create_ctx("demo"),
+            endpoint_id=uuid.uuid4().hex,  # either endpoint-id or function_name and function_tag must be provided
```
davesh0812: It might be better to avoid attempting to retrieve the MEP when `endpoint_id`, `function_name`, and `function_tag` are all `None`. In such cases, we can automatically proceed to generate a new MEP?

tests/system/model_monitoring/test_model_monitoring.py
```diff
@@ -1743,6 +1743,7 @@ def test_inference_feature_set(self) -> None:
             model_path=self.project.get_artifact_uri(
                 key=self.model_name, category="model", tag="latest"
             ),
+            function_name=self.model_name,
```
davesh0812: Its a bit confusing to use model_name as function_name.
I think you can create `self.function_name` and to use it

davesh0812 (commented):
Nice. 
Few comments

davesh0812 (approved):

PR #8419, opened 2025-08-05
[Bugfix] Add missing schema for function retry

adamdelman:
https://iguazio.atlassian.net/browse/ML-10733


moranbental (approved):

Yacouby:
why there was the need to add the spec manually ? shouldn't the `extra = pydantic.v1.Extra.allow` cover it ?

PR #8420, opened 2025-08-05
[Model Monitoring] Resolve schema from dict input using ModelRunnerStep

royischoss:
Fixing https://iguazio.atlassian.net/browse/ML-10657
1. Normalize schema inputs and outputs inside system_steps.py
2. Updating mep features in deployment

### DIFF
--- a/mlrun/serving/system_steps.py
+++ b/mlrun/serving/system_steps.py
@@ -46,12 +46,14 @@ class MonitoringPreProcessor(storey.MapClass):
         input_path = model_monitoring_data.get(MonitoringData.INPUT_PATH)
 
         result = get_data_from_path(result_path, event.body.get(model, event.body))
+
+        new_output_schema, new_input_schema = None, None
         output_schema = model_monitoring_data.get(MonitoringData.OUTPUTS)
         input_schema = model_monitoring_data.get(MonitoringData.INPUTS)
         logger.debug("output schema retrieved", output_schema=output_schema)
         if isinstance(result, dict):
             # transpose by key the outputs:
-            outputs = self.transpose_by_key(result, output_schema)
+            outputs, new_output_schema = self.transpose_by_key(result, output_schema)
             if not output_schema:
                 logger.warn(
                     "Output schema was not provided using Project:log_model or by ModelRunnerStep:add_model order "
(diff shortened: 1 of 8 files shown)


royischoss:
Passed `test_app_flow with_training_set= [True,False] , with_model_runner= [True]`, passed `test_monitoring_with_model_runner_dict_infer`

mlrun/serving/system_steps.py
```diff
@@ -136,20 +146,27 @@ def transpose_by_key(
                          * If result is a matrix, returns a list of lists.
 
         :raises ValueError: If the values include a mix of scalars and lists, or if the list lengths do not match.
+                mlrun.MLRunInvalidArgumentError if the schema keys are not contained in the data keys.
         """
-
+        new_schema = None
+        # Normalize keys in data:
+        data = {
```
davesh0812: lets give it more meaningful name

mlrun/serving/system_steps.py
```diff
@@ -136,20 +146,27 @@ def transpose_by_key(
...
+        }
         # Normalize schema to list
         if not schema:
             keys = list(data.keys())
+            new_schema = keys
         elif isinstance(schema, str):
             keys = [schema]
         else:
-            keys = schema
+            keys = [mlrun.feature_store.api.norm_column_name(key) for key in schema]
```
davesh0812, line 164: Why do we need this? It already the MEP schema no?

server/py/services/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -537,6 +537,12 @@ def _enrich_features_from_model_obj(
                     for f in model_obj.spec.outputs
                 ]
                 model_endpoint.spec.label_names = model_label_names
+            elif model_endpoint.spec.label_names:
+                model_label_names = [
+                    mlrun.feature_store.api.norm_column_name(name)
+                    for name in model_endpoint.spec.label_names
+                ]
+                model_endpoint.spec.label_names = model_label_names
```
davesh0812, line 545: Need to do the same for `model_endpoint.spec.feature_names` no?

davesh0812 (commented):
Nice catch, few comments :)

mlrun/serving/system_steps.py
royischoss: We are not normalizing outside of this method, in the serving process, if it the user schema I will have to normalize it here for the first time

server/py/services/api/crud/model_monitoring/model_endpoints.py
royischoss: Nope 🙂

davesh0812 (approved):

PR #8423, opened 2025-08-05
[Monitoring] Add debounce mechanism to prevent client-server status conflicts

Yacouby:
This PR introduces a debouncing mechanism in the server-side monitoring logic to address a race condition between client-side status updates and server-side runtime monitoring.

**Background:**
In some scenarios - particularly when the runtime reaches a terminal state shortly after a client-side update (e.g., via user code or SDK) a race condition occurs:
The runtime has finished and returned a final (terminal) status.
The DB still holds a recent non-terminal update from the client.

The runs monitoring detects the mismatch and overwrites the newer DB state with the outdated runtime info, possibly losing important fields.
This issue is not limited to retries, but can happen in any flow where there's close timing between runtime termination and a client update.

**Fix:**
To mitigate this, a debounce check was added:
If the runtime is terminal, but the DB state is non-terminal, and the DB update is recent (within a configurable debounce interval), then monitoring will skip the status update to avoid overwriting a likely more up-to-date client-side update.

Jira:
https://iguazio.atlassian.net/browse/ML-10553

### DIFF
--- a/server/py/services/api/runtime_handlers/base.py
+++ b/server/py/services/api/runtime_handlers/base.py
@@ -1745,13 +1745,7 @@ class BaseRuntimeHandler(ABC):
                     run, reason, message
                 )
 
-        logger.info(
-            "Updating run state",
-            run_uid=uid,
-            run_state=run_state,
-            reason=reason,
-            my_message=message,
-        )
+        logger.info("Updating run state", run_uid=uid, run_state=run_state)
         run_updates = {
             "status.state": run_state,
             "status.reason": reason or "",
@@ -1833,12 +1827,16 @@ class BaseRuntimeHandler(ABC):
             last_update = datetime.fromisoformat(last_update_str)
             debounce_period = config.monitoring.runs.interval
             debounce_cutoff = now - timedelta(seconds=float(debounce_period))
-            db_terminal = db_run_state in RunStates.terminal_states()
-            runtime_terminal = run_state in RunStates.terminal_states()
+            is_db_terminal = db_run_state in RunStates.terminal_states()
+            is_runtime_terminal = run_state in RunStates.terminal_states()
 
             # If the runtime has reached a terminal state but the DB still shows a recent non-terminal state,
             # debounce the update to avoid prematurely overriding the newer DB state.
-            if not db_terminal and runtime_terminal and last_update > debounce_cutoff:
+            if (
+                not is_db_terminal
+                and is_runtime_terminal
+                and last_update > debounce_cutoff
+            ):
                 logger.warning(
                     "Monitoring found terminal state on runtime resource but DB record was recently updated and is "
                     "still non-terminal. Debouncing.",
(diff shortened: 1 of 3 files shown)


server/py/services/api/runtime_handlers/base.py
```diff
@@ -1770,7 +1745,13 @@ def _ensure_run_state(
                     run, reason, message
                 )
 
-        logger.info("Updating run state", run_uid=uid, run_state=run_state)
+        logger.info(
+            "Updating run state",
+            run_uid=uid,
+            run_state=run_state,
+            reason=reason,
+            my_message=message,
```
TomerShor: what's `my_message`?

server/py/services/api/runtime_handlers/base.py
```diff
@@ -1830,6 +1811,75 @@ def _ensure_run(
...
+        """
+
+        now = datetime.now(timezone.utc)
+        last_update_str = run.get("status", {}).get("last_update")
+
+        if last_update_str is not None:
+            last_update = datetime.fromisoformat(last_update_str)
+            debounce_period = config.monitoring.runs.interval
+            debounce_cutoff = now - timedelta(seconds=float(debounce_period))
+            db_terminal = db_run_state in RunStates.terminal_states()
```
TomerShor:
```suggestion
            is_db_terminal = db_run_state in RunStates.terminal_states()
```

server/py/services/api/runtime_handlers/base.py
```diff
@@ -1830,6 +1811,75 @@ def _ensure_run(
...
+            runtime_terminal = run_state in RunStates.terminal_states()
```
TomerShor:
```suggestion
            is_runtime_terminal = run_state in RunStates.terminal_states()
```

TomerShor (changes requested):
Minor comments

server/py/services/api/runtime_handlers/base.py
Yacouby: ops, added it for debugging, will remove

server/py/services/api/runtime_handlers/base.py
```diff
@@ -1727,39 +1727,14 @@ def _ensure_run_state(
...
-                    run_state=run_state,
-                )
+            if self._should_debounce_run_update(
+                run=run,
+                db_run_state=db_run_state,
+                run_state=run_state,
+                project=project,
+                uid=uid,
+            ):
+                return False, run_state, run
```
TomerShor, line 1759: Not a part of your changes, but do you know what is this boolean first return value is? looks like it is always ignored.
Worth adding a docstring to `_ensure_run_state` that says what is the return value

TomerShor (approved):
Approved with a general question

server/py/services/api/runtime_handlers/base.py
Yacouby: it indicates whether the run state was updated or not. for now it is ignored but we could use it in the future.

TomerShor (approved):

PR #8425, opened 2025-08-06
[Bugfix] Properly mask DB DSN in logs

adamdelman:
https://iguazio.atlassian.net/browse/ML-10758


server/py/framework/utils/singletons/db.py
```diff
@@ -53,21 +52,4 @@ def initialize_db(override_db=None):
 
 
 def _mask_dsn(dsn: str) -> str:
```
moranbental: Why not remove this function?

moranbental (changes requested):

server/go/proto/health.pb.go
liranbg, line 1: remove protofiles

liranbg (changes requested):

liranbg (approved):

PR #8441, opened 2025-08-11
[Serving] Work around issue running a graph job locally in Jupyter

gtopper:
By trying to use nest_asyncio, and raising a clear error if it's not installed.

New test to verify this functionality:
`TestKubejobRuntime::test_job_from_serving_runtime_from_inside_running_asyncio_loop`

[ML-10620](https://iguazio.atlassian.net/browse/ML-10620)

[ML-10620]: https://iguazio.atlassian.net/browse/ML-10620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/serving/server.py
```diff
@@ -746,6 +746,25 @@ async def run(body):
     return responses
 
 
+def _workaround_asyncio_nesting():
+    """Workaround for running with local=True in Jupyter (ML-10620)"""
```
royischoss: I think this name could be changed to _handle_asyncio_nesting and the ticket name should not be part of the docs string, maybe if we want to change it add a todo for a new ticket, or add a single line comment that mentions the current ticket

mlrun/serving/server.py
```diff
@@ -746,6 +746,25 @@ async def run(body):
...
+    try:
```
royischoss, line 750: suggestion: introduce a method called is_inside_running_loop handling this logic.

royischoss (commented):
Hey two small comments

mlrun/serving/server.py
gtopper: I thought of calling it that, but this file contains some _handlers_, and I didn't want to risk confusion. Why not include the ticket? This is basically an internal comment (I could change it to an actual comment), it's not API.

mlrun/serving/server.py
```diff
@@ -746,6 +746,26 @@ async def run(body):
...
 
+def _is_inside_asyncio_loop():
+    try:
+        asyncio.get_running_loop()
+        return True
+    except RuntimeError:
+        return False
+
+
+# Workaround for running with local=True in Jupyter (ML-10620)
```
royischoss, line 757: 👍

royischoss (approved):
LGTM

PR #8466, opened 2025-08-13
[HTTPDB] Disable retries for storing secret tokens [feature/ig4-authentication]

moranbental:
### 📝 Description

This PR updates the HTTP session retry logic to prevent retries for PUT requests to `/user-secrets/tokens`.

---

### 🛠️ Changes Made

1. Added `_is_retry_put_allowed` method to determine whether retries are allowed for PUT requests.
2. Defined `NON_RETRIABLE_PATHS` to include `/user-secrets/tokens`
3. Updated HTTP session initialization to respect `retry_on_put` based on the requested path.

---

### ✅ Checklist

---

### 🧪 Testing
  
  
Ran unit tests to verify retry logic.
Tested on an IG4 system by sending a request to `/user-secrets/tokens` and confirmed that it was sent only once (no retries).

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10847
---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -123,7 +123,8 @@ class HTTPRunDB(RunDBInterface):
     ]
 
     NON_RETRIABLE_PATHS = [
-        r"user-secrets/tokens",
+        # Storing user secret tokens is not idempotent — retrying the request may result inconsistent secret storage.
+        r"\/?user-secrets/tokens",
     ]
 
     def __init__(self, url):
@@ -287,10 +288,9 @@ class HTTPRunDB(RunDBInterface):
 
         retry_on_post = self._is_retry_on_post_allowed(method, path)
 
-        # PUT retries are allowed unless the path is in NON_RETRIABLE_PATHS
         retry_on_put = self._is_retry_put_allowed(method, path)
 
-        # if the method is POST or PUST, we need to update the session with the appropriate retry policy
+        # if the method is POST or PUT, we need to update the session with the appropriate retry policy
         if not self.session or method in ("POST", "PUT"):
             self.session = self._init_session(retry_on_post, retry_on_put)
 
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -122,6 +122,10 @@ class HTTPRunDB(RunDBInterface):
         r"\/?run\/.+\/.+",
     ]
 
+    NON_RETRIABLE_PATHS = [
+        r"user-secrets/tokens",
+    ]
+
     def __init__(self, url):
         self.server_version = ""
         self.session = None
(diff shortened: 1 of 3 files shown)


mlrun/db/httpdb.py
```diff
@@ -122,6 +122,10 @@ class HTTPRunDB(RunDBInterface):
         r"\/?run\/.+\/.+",
     ]
 
+    NON_RETRIABLE_PATHS = [
+        r"user-secrets/tokens",
+    ]
```
TomerShor: Explain WHY this is non retryable

mlrun/db/httpdb.py
```diff
@@ -281,10 +285,14 @@ def api_call(
...
-        # if the method is POST, we need to update the session with the appropriate retry policy
-        if not self.session or method == "POST":
-            retry_on_post = self._is_retry_on_post_allowed(method, path)
-            self.session = self._init_session(retry_on_post)
+        retry_on_post = self._is_retry_on_post_allowed(method, path)
+
+        # PUT retries are allowed unless the path is in NON_RETRIABLE_PATHS
+        retry_on_put = self._is_retry_put_allowed(method, path)
+
+        # if the method is POST or PUST, we need to update the session with the appropriate retry policy
```
```suggestion
        # if the method is POST or POST, we need to update the session with the appropriate retry policy
```

mlrun/db/httpdb.py
```diff
@@ -281,10 +285,14 @@ def api_call(
                     if isinstance(dict_[key], enum.Enum):
                         dict_[key] = dict_[key].value
 
-        # if the method is POST, we need to update the session with the appropriate retry policy
-        if not self.session or method == "POST":
-            retry_on_post = self._is_retry_on_post_allowed(method, path)
-            self.session = self._init_session(retry_on_post)
+        retry_on_post = self._is_retry_on_post_allowed(method, path)
+
+        # PUT retries are allowed unless the path is in NON_RETRIABLE_PATHS
```
TomerShor: This comment just tells how `_is_retry_put_allowed` is implemented, it doesn't matter here really.

mlrun/db/httpdb.py
```diff
@@ -281,10 +285,14 @@ def api_call(
...
+        retry_on_put = self._is_retry_put_allowed(method, path)
+
+        # if the method is POST or PUST, we need to update the session with the appropriate retry policy
+        if not self.session or method in ("POST", "PUT"):
```
TomerShor, line 294: what if `self.session is None` and the method is not POST or PUT? shouldn't it be initialized then?

mlrun/db/httpdb.py
```diff
@@ -122,6 +122,10 @@ class HTTPRunDB(RunDBInterface):
         r"\/?run\/.+\/.+",
     ]
 
+    NON_RETRIABLE_PATHS = [
+        r"user-secrets/tokens",
```
TomerShor: why not keep the same regex format (with the leading `/`) as the paths above?

TomerShor (changes requested):

mlrun/db/httpdb.py
moranbental: @TomerShor I’m not sure I understand. There’s an OR here, if the session is None and for example this is GET request, will it be initialized

TomerShor: Right, my bad.

TomerShor (approved):

PR #8471, opened 2025-08-14
[Secret Tokens] Enable Conditional Support Based on authentication.mode [feature/ig4-authentication]

moranbental:
### 📝 Description

Enable conditional support for secret tokens based on `mlrun.httpdb.authentication.mode`.
This PR restricts certain SDK methods and API endpoints so they can only be used when authentication.mode is set to `iguazio-v4`.

---

### 🛠️ Changes Made

1. Moved AuthenticationMode enum to a shared location.
2. Created a `iguazio_v4_only ` wrapper for SDK methods to control execution based on the authentication mode.
3. Added a FastAPI dependency for `user-secrets` endpoints to enforce mode-based access control.

---

### ✅ Checklist

---

### 🧪 Testing
  
  

1. Verified SDK methods fail when called outside iguazio-v4 mode.
2. Verified API user-secrets endpoints return 400 when the mode is not iguazio-v4.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10494
- Design docs links: https://iguazio.atlassian.net/wiki/spaces/MLRUN/pages/404521061/BE+Secret+Token+Support+HLD#2.1.9-Configurable-Authentication-Mode

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -80,14 +80,13 @@ def bool2str(val):
     return "yes" if val else "no"
 
 
-def require_iguazio_v4(function):
+def iguazio_v4_only(function):
     @functools.wraps(function)
     def wrapper(*args, **kwargs):
         authentication_mode = mlrun.mlconf.httpdb.authentication.mode
         if authentication_mode != AuthenticationMode.IGUAZIO_V4:
-            raise mlrun.errors.MLRunInvalidArgumentError(
-                f"This method is only supported when mlrun.authentication.mode='iguazio-v4', "
-                f"but current mode is '{authentication_mode}'."
+            raise mlrun.errors.MLRunRuntimeError(
+                "This method is only supported in an Iguazio V4 system."
             )
         return function(*args, **kwargs)
 
@@ -5035,7 +5034,7 @@ class HTTPRunDB(RunDBInterface):
         response = self.api_call("GET", endpoint_path, error_message)
         return mlrun.common.schemas.ProjectSummary(**response.json())
 
-    @require_iguazio_v4
+    @iguazio_v4_only
     def store_secret_token(
         self,
         secret_token: mlrun.common.schemas.SecretToken,
@@ -5070,7 +5069,7 @@ class HTTPRunDB(RunDBInterface):
             )
         return response
 
-    @require_iguazio_v4
+    @iguazio_v4_only
     def store_secret_tokens(
         self,
         secret_tokens: list[mlrun.common.schemas.SecretToken],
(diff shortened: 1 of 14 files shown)


mlrun/db/httpdb.py
```diff
@@ -78,6 +80,20 @@ def bool2str(val):
     return "yes" if val else "no"
 
 
+def require_iguazio_v4(function):
```
TomerShor: I would rename it to `iguazio_v4_only`, because this function `validates` and not `requires`.
Just sounds better to me

mlrun/db/httpdb.py
```diff
@@ -78,6 +80,20 @@ def bool2str(val):
...
+    @functools.wraps(function)
+    def wrapper(*args, **kwargs):
+        authentication_mode = mlrun.mlconf.httpdb.authentication.mode
+        if authentication_mode != AuthenticationMode.IGUAZIO_V4:
+            raise mlrun.errors.MLRunInvalidArgumentError(
```
TomerShor: This isn't an invalid argument, because the failure is not due to something the user passed to the function - it fails because the server isn't operating in the correct state.
`MLRunRuntimeError` suits better here, because the system is in an incompatible runtime state

mlrun/db/httpdb.py
```diff
@@ -78,6 +80,20 @@ def bool2str(val):
...
+                f"This method is only supported when mlrun.authentication.mode='iguazio-v4', "
+                f"but current mode is '{authentication_mode}'."
```
```suggestion
                "This method is only supported in an Iguazio V4 system."
```

Say you're running in OpenSource (CE) - setting this mode won't really enable this methods without the rest of the Iguazio services.

server/py/framework/api/deps.py
```diff
@@ -93,3 +94,11 @@ def expose_internal_endpoints(request: Request):
             raise mlrun.errors.MLRunPreconditionFailedError(
                 "Internal endpoints are not exposed"
             )
+
+
+def require_iguazio_v4(request: Request):
+    if mlrun.mlconf.httpdb.authentication.mode != AuthenticationMode.IGUAZIO_V4:
+        raise mlrun.errors.MLRunBadRequestError(
+            f"This endpoint is only supported when mlrun.mlconf.httpdb.authentication.mode='iguazio-v4', "
+            f"but current mode is '{mlrun.mlconf.httpdb.authentication.mode}'."
```
TomerShor: Same as the previous comments

TomerShor (changes requested):
Overall looking good, added some suggestion for improvement

TomerShor (approved):

PR #8473, opened 2025-08-14
[Framework] Ensure running _sync_system_metadata on both worker and chief

liranbg:
### 📝 Description

Issue was that chief starts with init_from_scrach generates the system id, then restarted and wont load the system id correctly. the worker has no concrete issue as the service initialize sets the system id once moved online.

---

### 🛠️ Changes Made

1. Ensure the system id is set on both chief and worker once moved online
2. Simplified the service setup flow. less conditions. easier to read
3. Added some docs and notes to init_data flow
4. Extended run_async_function_with_new_db_session functionality to allow running sync function within async context in a thread
5. Fixed couple of io-blocking calls within an async context [ 36130de40a778cc69e86c71655333e5138ff0317 / 26262a1f86d5815240bcddbd4656d733a7f4ffd3]

---

### ✅ Checklist

---

### 🧪 Testing
  
  

used `/patch_remote.py` before/after to ensure fix is applied.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10859
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/server/py/framework/db/session.py
+++ b/server/py/framework/db/session.py
@@ -47,8 +47,8 @@ def run_function_with_new_db_session(func, *args, **kwargs):
 async def run_async_function_with_new_db_session(func, *args, **kwargs):
     """
     Run an async function with a new db session.
-    If the func is async, use run_async_function_with_new_db_session below.
-    Otherwise, use the suggested logic.
+    If the func is a coroutine function (async def), use run_async_function_with_new_db_session below.
+    alternatively, given the async context, run the synchronous function in a thread pool.
 
     Any changes made by the new session will not be visible to old sessions until the old sessions commit
     due to isolation level.
--- a/server/py/framework/db/session.py
+++ b/server/py/framework/db/session.py
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import inspect
+
+import fastapi.concurrency
 from sqlalchemy.orm import Session
 
 import mlrun.utils.helpers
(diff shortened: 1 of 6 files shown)


server/py/framework/db/session.py
```diff
@@ -43,13 +46,28 @@ def run_function_with_new_db_session(func, *args, **kwargs):
 
 async def run_async_function_with_new_db_session(func, *args, **kwargs):
     """
-    Run an async function with a new db session, useful for concurrent requests where we can't share a single session.
-    However, any changes made by the new session will not be visible to old sessions until the old sessions commit
+    Run an async function with a new db session.
+    If the func is async, use run_async_function_with_new_db_session below.
+    Otherwise, use the suggested logic.
```
TomerShor: What do those 2 line even mean, and who are they addressing?

server/py/framework/service/__init__.py
```diff
@@ -407,6 +406,23 @@ def _resolve_paginated_methods(
...
+        """
+
+        db = framework.db.sqldb.db.SQLDB()
+        system_id = db.get_system_id(db_session)
+        if system_id is not None:
+            self._logger.debug(
+                "Existing system ID found in the database",
+                system_id=system_id,
+            )
+            mlrun.mlconf.system_id = system_id
```
TomerShor, line 424: Doing it here means that all services will do it (chief, worker & alerts) as opposed to only chief.
This is a bit of an anti pattern of how the worker and alerts are syncing the config from the chief.

server/py/services/api/initial_data.py
```diff
@@ -56,36 +57,56 @@
 def init_data(
     perform_migrations_if_needed: bool = False,
 ) -> None:
+    """
+    This function initializes the database with the necessary data.
+    It checks if the database exists and if it has any tables,
```
TomerShor, line 62: not really, that's what `_ensure_db_instance_initialized` checks

TomerShor (changes requested):

server/py/services/api/initial_data.py
```diff
@@ -56,36 +57,56 @@
...
+        mlrun.utils.logger.info("Migrating existing data")
+        _migrate_existing_data(
+            engine,
+            perform_migrations_if_needed,
+        )
+
+    mlrun.utils.logger.info("Initial data created")
+
+
+def _ensure_db_instance_initialized(engine: sqlalchemy.engine.Engine) -> bool:
```
rokatyy: name is a bit confusing. Maybe `initialize_db_if_needed` would be a bit more informative?

rokatyy (approved):
looks good to me, just one minor comment

server/py/services/api/initial_data.py
liranbg: Agree the name is not the best, will change it

server/py/services/api/initial_data.py
liranbg: well, yeah but that is the purpose of this function (the what), fact it uses some private functions is impl details (the how)

server/py/framework/service/__init__.py
liranbg: well, beforhand the chief config initialization was not stable and alerts never got this configuration
now they are all aligned. I would have expected that the clusterization distribution function responsible to set the system id but that was not the case. what I did is aligning the get/set across services. next part Id imagine is to consider moving config set logic via clusterization spec

server/py/framework/db/session.py
liranbg: updated comment

server/py/framework/service/__init__.py
TomerShor: > next part Id imagine is to consider moving config set logic via clusterization spec

That's what I thought should happen.
Can be done in a followup though.

TomerShor (approved):

PR #8476, opened 2025-08-17
[Runs] Prevent serialization issues from V1 Kubernetes objects in run objects

yaelgen:
### 📝 Description
This PR fixes an error introduced after the refactor that moved affinity and tolerations into the run object.
By storing raw Kubernetes V1 objects (with internal thread locks) in the run object, we exposed the API serialization path to non-serializable fields.

The issue occurred for two reasons:

- Deep copy of V1 objects pulled in _thread.RLock.
- Run object serialization tried to include those raw V1 objects, leading to errors when the API attempted to return run details.

As a result, we encountered internal server errors such as:
```
ValueError([TypeError("'_thread.RLock' object is not iterable"),
TypeError('vars() argument must have __dict__ attribute')])
```
When having tolerations ot affinity on the run object.

---

### 🛠️ Changes Made
- Sanitized affinity and tolerations on run objects to avoid storing raw V1 Kubernetes objects directly.
- Adjusted deep copy and serialization logic to strip non-serializable fields (e.g., _thread.RLock).
- Ensured affinity and tolerations can still be retrieved without breaking run detail retrieval.
---

### ✅ Checklist

---

### 🧪 Testing
- Unit tests: Adjusted the unit test and confirmed safe serialization.
- System tests: Added new coverage that executes runs with tolerations/affinity and validates that run details can be retrieved successfully.
- Manual verification: 
- Reproduced the failing scenario from the ticket on AKS and confirmed run details now return successfully without internal errors.
- Tested multiple types of functions (including spark and dask) to make sure it behaves as expected
---

### 🔗 References
Ticket link: [ML-10324](https://iguazio.atlassian.net/browse/ML-10324)
---

### 🚨 Breaking Changes?
- [ ] Yes (explain below)
- [x] No

[ML-10324]: https://iguazio.atlassian.net/browse/ML-10324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


automation/system_test/prepare.py
```diff
@@ -352,6 +353,21 @@ def _override_mlrun_api_env(self):
...
+                    [
+                        {
+                            "key": "app.iguazio.com/lifecycle",
+                            "operator": "Equal",
+                            "value": "preemptible",
+                            "effect": "NoSchedule",
+                        }
+                    ]
+                ).encode("utf-8")
+            ).decode("utf-8"),
```
quaark: Maybe move `base64.b64encode(json.dumps(<dict>).encode("utf-8")).decode("utf-8"),` to a helper method?

mlrun/k8s_utils.py
```diff
@@ -232,6 +232,53 @@ def handle_invalid(message):
...
+            api_client.sanitize_for_serialization(item)
+            if hasattr(item, "to_dict")
+            else item
+        )
+
+    return (
+        [_sanitize_item(item) for item in obj]
+        if isinstance(obj, list)
+        else _sanitize_item(obj)
+    )
```
quaark: I feel `item` and `obj` aren't great param names - I'd make them a bit more indicative of what you are trying to do here.
Also why is it `sanitize` and not `serialize`?

quaark (changes requested):

mlrun/k8s_utils.py
yaelgen: I used `sanitize` because that's what the underlying K8s method is called (`api_client.sanitize_for_serialization()`).
Also, we're not actually serializing here - we're just converting K8s objects to plain Python dicts. 
Happy to rename it if you have a preference though.

automation/system_test/prepare.py
```diff
@@ -313,6 +315,19 @@ def _run_command_remotely(
...
+        """
+        Convert a Python object to base64-encoded JSON string.
+
+        Args:
+            data: Any JSON-serializable Python object (dict, list, etc.)
+
+        Returns:
+            Base64-encoded JSON string
+        """
+        return base64.b64encode(orjson.dumps(data)).decode("utf-8")
```
adamdelman: no need for the "utf-8"

automation/system_test/prepare.py
```diff
@@ -313,6 +315,19 @@ def _run_command_remotely(
 
         return stdout, stderr, exit_status
 
+    @staticmethod
+    def _encode_json_to_base64(data) -> str:
```
adamdelman: type annotation?

mlrun/k8s_utils.py
```diff
@@ -232,6 +232,52 @@ def handle_invalid(message):
     return True
 
 
+def sanitize_k8s_objects(k8s_objects) -> typing.Union[dict, list[dict], None]:
```
adamdelman: type annotation

tests/system/projects/test_project.py
```diff
@@ -1332,10 +1332,26 @@ def _assert_workflow_status(workflow, status):
...
         self, project: mlrun.projects.MlrunProject
     ):
+        """
+        Create a function with node selector, tolerations and affinity, enable `prevent`
+        preemption mode, run it, and assert that:
+          1) The job's nodeSelector is pruned of preemptible labels (function is not mutated).
+          2) The job's preemptible toleration is removed; user tolerations remain.
+          3) The job's preemptible node-affinity rule is pruned (function is not mutated).
+        """
+        from kubernetes import client as k8s_client
```
adamdelman: why is the import here?

adamdelman (changes requested):
Looks good, minor changes

automation/system_test/prepare.py
yaelgen: oopsi :face_with_peeking_eye:

adamdelman (approved):

server/py/services/api/runtime_handlers/sparkjob/spark3job.py
```diff
@@ -798,6 +798,11 @@ def enrich_and_update(component: str, affinity, tolerations, preemption_mode):
                     tolerations=tolerations,
                 )
             )
+            enriched_tolerations, enriched_affinity = (
+                mlrun.k8s_utils.sanitize_scheduling_configuration(
+                    enriched_tolerations, enriched_affinity
+                )
```
liranbg, line 804: why not making the `enrich_preemption_mode` returns the sanitized?
also `sanitize_scheduling_configuration` is just a convenient function, why not keep calling sanitize_k8s_objects twice and having sanitize_scheduling_configuration removed all together?

mlrun/k8s_utils.py
```diff
@@ -26,6 +26,12 @@
 
 _running_inside_kubernetes_cluster = None
 
+V1Affinity = kubernetes.client.V1Affinity
+V1Toleration = kubernetes.client.V1Toleration
+K8sObj = typing.Union[V1Affinity, V1Toleration]
+K8sObjList = list[K8sObj]
+SanitizedK8s = dict[str, typing.Any]
```
liranbg: not sure why these are needed

mlrun/k8s_utils.py
```diff
@@ -232,6 +238,54 @@ def handle_invalid(message):
...
+            api_client.sanitize_for_serialization(k8s_obj)
+            if hasattr(k8s_obj, "to_dict")
+            else k8s_obj
+        )
+
+    return (
+        [_sanitize_k8s_object(k8s_obj) for k8s_obj in k8s_objects]
+        if isinstance(k8s_objects, list)
+        else _sanitize_k8s_object(k8s_objects)
+    )
```
liranbg, line 258:
```suggestion
    k8s_objects = mlrun.utils.as_list(k8s_objects)
    return list(map(_sanitize_k8s_object, k8s_objects))
```
more elegant I think.

liranbg (changes requested):
Nicely done. minor styling and suggestions.

server/py/services/api/runtime_handlers/sparkjob/spark3job.py
yaelgen: I’d keep `enrich_preemption_mode` returning K8s model objects, its role is enrichment only. Mixing in sanitization blurs concerns and breaks type stability, and also downstream code still expects V1 objects (changing that would also require broad test adaptations).

For `sanitize_scheduling_configuration`, it's a small helper we can reuse for both tolerations and affinity since we typically call it for both. Happy to remove it if you prefer, but it felt cleaner to me given the repeated usage.

mlrun/k8s_utils.py
yaelgen: I prefer keeping the current approach since it preserves the input/output type relationship as specified in the type hints. Using as_list() would require additional logic to track the original type, which doesn't add much value here. In your suggested approach, we would always return a list, even when the original input wasn't a list.

mlrun/k8s_utils.py
yaelgen: To have explicit type annotation, I reduced it a bit but I think it is better being explicit then implicit

server/py/services/api/runtime_handlers/sparkjob/spark3job.py
liranbg: it makes code more readable but the difference is not too big, may keep it. when it grows, perhaps we would need to relax it

mlrun/k8s_utils.py
liranbg: the type annotation can be used on the function params directly. also, what you did here is variable assignment not type declaration
to do type declaration you do

```
x: TypeAlias = int
```

because x = int means x is int variable x(5)

yaelgen: I removed the variable assignment, thats what I meant by "I reduced it a bit", look at the last commit.

mlrun/k8s_utils.py
```diff
@@ -26,6 +26,10 @@
 
 _running_inside_kubernetes_cluster = None
 
+K8sObj = typing.Union[kubernetes.client.V1Affinity, kubernetes.client.V1Toleration]
+K8sObjList = list[K8sObj]
+SanitizedK8s = dict[str, typing.Any]
```
liranbg:
```suggestion
SanitizedK8sObj = dict[str, typing.Any]
```

mlrun/k8s_utils.py
```diff
@@ -26,6 +26,10 @@
 
 _running_inside_kubernetes_cluster = None
 
+K8sObj = typing.Union[kubernetes.client.V1Affinity, kubernetes.client.V1Toleration]
+K8sObjList = list[K8sObj]
```
liranbg: how about

```suggestion
K8sObjList = typing.Union[list[K8sObj], list(SanitizedK8sObj)]
```

liranbg (commented):
some suggestions - I find it needless to declare type annotations being used only once
I understand the need for shortening the type hints on function params

moranbental (changes requested):
please fix docstring to our format

yaelgen:
> please fix docstring to our format
Per the committee’s decision, all new MLRun code need to use the new format

liranbg (approved):

PR #8477, opened 2025-08-17
[Model Monitoring] Fix function summary API including fixing app name 

Eyal-Danieli:
### 📝 Description

This PR includes fixes for the function summary API. 

---

### 🛠️ Changes Made
- Processing the API even if client version was not provided in the header
- Support providing only `start` or `end` time (used to support both or neither) 
- Fixing app name to be based on the original application name and not on the lower case name. Note that you must provide the exact original name to retrieve the stats data.

---

### ✅ Checklist

---

### 🧪 Testing
- `test_app`: Changed the app name of `DemoEvidentlyMonitoringApp` from `"evidently-app-test"` to `"Evidently-App-Test"`
- `test_app`: Providing just `start_time` to the function summary API 

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10746
- Design docs links: https://iguazio.atlassian.net/wiki/spaces/Dataflows/pages/352288854/ML-9609+Monitoring+Applications+APIs

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


mlrun/common/schemas/model_monitoring/functions.py
```diff
@@ -54,7 +54,11 @@ def from_function_dict(
 
         return cls(
             type=func_type,
-            name=func_dict["metadata"]["name"],
+            name=func_dict["metadata"]["name"]
+            if func_type != FunctionsType.APPLICATION
+            else func_dict["spec"]["graph"]["steps"]["PrepareMonitoringEvent"][
```
davesh0812: I am not sure it will work with batch application (evaluate).. because if I understand correctly those apps don't uses the graph at all. 

Note: same is relevant to [L64](https://github.com/mlrun/mlrun/pull/8477/files#diff-af4f90ed679a21644b7ed112185f69c0ba62ed0bfd4def6ce6c3dd5c54d670caR64)

mlrun/common/schemas/model_monitoring/functions.py
```diff
@@ -54,7 +54,11 @@ def from_function_dict(
...
+                "class_args"
+            ]["application_name"],
```
davesh0812: Maybe better to use get to prevent bug on the backend side
```suggestion
            else func_dict.get("spec",{}).get("graph",{}).get("steps", {}).get("PrepareMonitoringEvent", {}).get("class_args", {}).get("application_name", name)
```

davesh0812 (commented):
Nice.

Need to make sure how it will work with batch application.

davesh0812 (approved):

jillnogold (approved):

PR #8479, opened 2025-08-17
[KFP] Improve KFP run listing performance using experiments

adamdelman:
### 🎯 Description
Switches pipeline run listing and cleanup to **experiment ID filtering** instead of project name substring, reducing scanned data and improving accuracy.

---

### 🛠 Changes Made
- **Clean pipelines script**
  - Use `mlrun_pipelines.client.Client` instead of `kfp.Client`.
  - Resolve experiments via `get_candidate_experiments_for_projects`.
  - Build filters using `experiment_ids`, `start_date`, `end_date`.
  - Removed `dry_run` param from `_delete_runs`.

- **Filter construction**
  - `get_kfp_list_runs_filter` now accepts `experiment_ids` and drops `name` predicates.
  - Added `FilterFields` (`CREATED_AT`, `EXPERIMENT_ID`) and `FilterOperations.IN`.

- **KFP client**
  - Added `get_candidate_experiments_for_projects` to resolve experiments by project name.
  - Added `IN`, `IS_SUBSTRING`, `NOT_IN` to `FilterOperations`.

- **Server API**
  - Use experiment IDs when listing pipelines.
  - Added logs for filter generation and pagination.
  - Parallelized `_format_runs`.

- **Versions**
  - `mlrun-pipelines-kfp-common` → **0.5.9**  
  - `mlrun-pipelines-kfp-v1-8` → **0.5.8**

---

### ✅ Testing
- Verified experiment-based filtering returns correct results and scans less data.

---

### 🔗 References
- https://iguazio.atlassian.net/browse/ML-10987


mlrun/utils/helpers.py
```diff
@@ -1214,52 +1215,58 @@ def get_workflow_url(
...
+                "op": mlrun_pipelines.models.FilterOperations.GREATER_THAN_EQUALS.value,
                 "timestamp_value": start_date,
             }
         )
-    return json.dumps(filters)
+
+    if experiment_id:
+        new_predicates.append(
+            {
+                "key": "experiment_id",
```
yaelgen: Can be a const

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/client.py
```diff
@@ -664,6 +664,41 @@ def terminate_run(
...
+
+        experiments = (
+            self._experiment_api.list_experiment(filter=filter_json).experiments or []
+        )
+        for experiment in experiments:
+            normalized_name = experiment.name.rsplit("-", 1)[0]
+            if project_name in (normalized_name, experiment.name):
+                return experiment
+        else:
+            raise ValueError(f"No experiment found with for project '{project_name}'")
```
yaelgen:
```suggestion
            raise ValueError(f"No experiment found for project '{project_name}'")
```

server/py/services/api/crud/pipelines.py
```diff
@@ -72,42 +72,42 @@ def list_pipelines(
...
+                    for run in runs
+                    if self.resolve_project_from_pipeline(run) == project
+                ]
+            elif isinstance(project, list):
+                runs = [
+                    run
+                    for run in runs
+                    if self.resolve_project_from_pipeline(run) in project
+                ]
+
```
yaelgen, line 110:
```suggestion
if project:
    project_set = {project} if isinstance(project, str) else set(project)
    runs = [run for run in runs if self.resolve_project_from_pipeline(run) in project_set]
```

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/client.py
```diff
@@ -664,6 +664,41 @@ def terminate_run(
...
+            }
+        ).decode()
+
+        experiments = (
+            self._experiment_api.list_experiment(filter=filter_json).experiments or []
+        )
+        for experiment in experiments:
+            normalized_name = experiment.name.rsplit("-", 1)[0]
+            if project_name in (normalized_name, experiment.name):
+                return experiment
```
yaelgen: I think this is problematic and may cause us to lose results. The experiment name is:
`experiment = name or workflow_spec.name`
So it can either be name or the workflow name, and the name comes from:
`name = f"{self.metadata.name}-{name}" if name else self.metadata.name`
in` project.run()`. So depending on how users run their pipelines, a single project can have more than one experiment.
this function only finds ONE experiment per project, meaning users will lose visibility into runs from other naming patterns.

yaelgen (changes requested):

mlrun/utils/helpers.py
```diff
@@ -1214,52 +1215,58 @@ def get_workflow_url(
...
-    return json.dumps(filters)
+
+    final_filter_object = {"predicates": preserved_predicates + new_predicates}
+    if not final_filter_object["predicates"]:
+        return ""
+
+    logger.warning(
+        "Generated KFP runs filter",
+        filter_object_with_predicates=final_filter_object,
+    )
```
```suggestion
    logger.debug(
        "Generated KFP runs filter",
        filter_object_with_predicates=final_filter_object,
    )
```

pipeline-adapters/mlrun-pipelines-kfp-common/src/mlrun_pipelines/common/client.py
```diff
@@ -183,3 +183,18 @@ def upload_pipeline(
...
         raise NotImplementedError
+
+    @abstractmethod
+    def get_candidate_experiments_for_projects(
+        self,
+        project_name: str,
+    ):
+        """
+        Retrieve an experiment by project name.
+        This method searches for an experiment whose name matches the project name,
```
liranbg, line 194: you retrieve experiments (meaning project -> experiments)

server/py/services/api/crud/pipelines.py
```diff
@@ -72,42 +73,45 @@ def list_pipelines(
...
+            kfp_client, page_token, page_size, sort_by, filter_json
         )
-        if project_names:
-            runs = [
-                run
-                for run in runs
-                if self.resolve_project_from_pipeline(run) in project_names
-            ]
+
+        if project:
```
liranbg, line 97:
```suggestion
        if projects:
```
no?

server/py/services/api/crud/pipelines.py
```diff
@@ -664,6 +668,13 @@ def _paginate_runs(
         filter_: typing.Optional[str] = None,
     ) -> tuple[list[mlrun_pipelines.models.PipelineRun], typing.Optional[int]]:
         next_page_token = -1
+        mlrun.utils.logger.warning(
+            "Paginating runs from KFP",
+            page_token=page_token,
+            page_size=page_size,
+            sort_by=sort_by,
+            filter_=filter_,
+        )
```
liranbg: too verbose, remove

server/py/services/api/crud/pipelines.py
```diff
@@ -688,7 +699,14 @@ def _paginate_runs(
...
                 page_token = next_page_token
-
+        mlrun.utils.logger.warning(
+            "Finished paginating runs from KFP",
+            page_token=page_token,
+            page_size=page_size,
+            sort_by=sort_by,
+            filter_=filter_,
+            run_count=len(runs),
+        )
```
liranbg: too verbose, remove

server/py/services/api/crud/pipelines.py
```diff
@@ -700,18 +718,34 @@ def _list_runs_from_kfp(
         filter_: typing.Optional[str] = None,
     ) -> tuple[list[mlrun_pipelines.models.PipelineRun], typing.Optional[str]]:
         try:
+            mlrun.utils.logger.warning(
+                "Listing runs from KFP",
+                page_token=page_token,
+                page_size=page_size,
+                sort_by=sort_by,
+                filter_=filter_,
+            )
```
liranbg: too verbose, remove

server/py/services/api/crud/pipelines.py
```diff
@@ -700,18 +718,34 @@ def _list_runs_from_kfp(
...
                 sort_by=sort_by if not page_token else "",
                 filter=filter_ if not page_token else "",
             )
+            mlrun.utils.logger.warning(
+                "Recieved list runs response from KFP",
+                page_token=page_token,
+                page_size=page_size,
+                sort_by=sort_by,
+                filter_=filter_,
+            )
```
liranbg: too verbose, remove

server/py/services/api/crud/pipelines.py
```diff
@@ -727,8 +761,15 @@ def _format_runs(
...
-            formatted_runs.append(self._format_run(run, format_, kfp_client))
+
+        def worker(run):
+            return self._format_run(run, format_, kfp_client)
+
+        with ThreadPoolExecutor() as executor:
+            futures = [executor.submit(worker, run) for run in runs]
+            for future in as_completed(futures):
+                formatted_runs.append(future.result())
+
```
liranbg: I think that this needs to happen on a separated PR and have:
1. Make sure you perform the `get_error_from_pipeline` only for subset of the runs (specifically those in error or failed)
2. for those in error or failed -> format their error using the thread executor.

liranbg (changes requested):
Nicely done, few suggestions below

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/client.py
```diff
@@ -664,6 +664,43 @@ def terminate_run(
...
+                }
+            ).decode()
+
+            experiments = (
+                self._experiment_api.list_experiment(filter=filter_json).experiments
+                or []
+            )
+            for experiment in experiments:
+                normalized_name = experiment.name.rsplit("-", 1)[0]
+                if project_name in (normalized_name, experiment.name):
```
theSaarco: Not sure I understand why you use `rsplit` here. IIRC, the project-name is expected to be the prefix of the experiment name, so if the experiment name has dashes in it, this won't work. For example, an experiment called `exp-1` in project `proj` will be called `proj-exp-1` and the split will return `(proj-exp,1)`, not matching the query. Why not simply use `startswith`?

server/py/services/api/crud/pipelines.py
```diff
@@ -72,42 +73,45 @@ def list_pipelines(
...
-                project_name=project_names[0]
+                "Resolving experiments by project-based substring match",
+                project=project,
             )
+            experiments = kfp_client.get_candidate_experiments_for_projects(projects)
+            experiment_ids = [experiment.id for experiment in experiments]
+        filter_json = mlrun.utils.get_kfp_list_runs_filter(
+            filter_=filter_,
+            experiment_ids=experiment_ids,
+        )
```
theSaarco, line 91: You're ignoring the `name_contains` parameter here (if provided), and then performs post-filtering on the results later on. Seems to me it would be more efficient to send it to kfp as a query parameter, either here or even in the `get_candidate_experiments_for_projects` function.

theSaarco (commented):
Looks good, have some questions.

server/py/services/api/crud/pipelines.py
adamdelman: Ignoring the name contains is the whole point of this, as otherwise we get the `LIKE %project_name%` query which  causes the full table scan.

theSaarco: But you're doing it anyway in `get_candidate_experiments_for_projects`, so why not pass also the `name_contains` to it and perform this filtering already there?
My point is that you can already limit the list of experiment IDs based on the name, before even getting to retrieve runs.

pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/client.py
liranbg: fixed

server/py/services/api/crud/pipelines.py
adamdelman: I would have to change the router as well as well as other functions that use the URI path based argument, so I would rather not change it.

server/py/services/api/crud/pipelines.py
adamdelman: As we discussed offline:
Once this change is merged, we will only use substring matching in the experiments table, ignoring the name filtering in the filter_json string that is passed from the frontend.
Since the experiment name contains the project name, but not the rest of the name (e.g. the name of the workflow, or the creation date that gets appended to the run name), we will get incorrect results if we pass the name_contains value to the experiments filtering API call.
So we are only left with the option of filtering on the MLRun server side after fetching the data from KFP.

mlrun-ce-docker.env
liranbg, line 1: Remove file

server/py/services/api/tests/unit/api/db/mlrun.db
liranbg, line 1: Remove file

liranbg (commented):
Minor files to remove

liranbg (approved):

PR #8484, opened 2025-08-18
[User secrets] Implement store user token secrets in k8s [feature/ig4-authentication]

moranbental:
### 📝 Description

This PR implements storing user tokens in Kubernetes secrets by first attempting to read the secret (by name and labels).

- If the secret does not exist, it is created.
- If the secret exists, the token’s expiration is checked, and the secret is updated only if the new token has a later expiration.

---

### 🛠️ Changes Made

- Implemented store_user_token_secret to handle secret creation and conditional update based on expiration.
- Updated k8s.read_secret to find secrets by name and verify labels.
- Fixed SDK response methods to correctly report created, updated, and skipped tokens.
- Adjusted SDK warnings to only log when tokens are actually stored or updated.

---

### ✅ Checklist

---

### 🧪 Testing
  
  
Unit tests for secret creation, update, and skip scenarios.
Verified manually on IG4 system.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10492
- Design docs links: https://iguazio.atlassian.net/wiki/spaces/MLRUN/pages/404521061/BE+Secret+Token+Support+HLD#2.2.3-Token-Store-or-Update-Flow
- External links: https://iguazio.atlassian.net/wiki/spaces/ARC/pages/361103361/MLRun+Secret+Tokens+in+IG4

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

### DIFF
--- a/server/py/framework/utils/singletons/k8s.py
+++ b/server/py/framework/utils/singletons/k8s.py
@@ -601,18 +601,26 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
                 name=secret_name,
                 namespace=namespace,
             )
+
             if labels:
                 secret_labels = k8s_secret.metadata.labels or {}
-                for key, value in labels.items():
-                    if secret_labels.get(key) != value:
-                        if not silent:
-                            logger.debug(
-                                "Secret found but labels did not match",
-                                secret_name=secret_name,
-                                expected_labels=labels,
-                                actual_labels=secret_labels,
-                            )
-                        return None
+
+                # Find any label that does not match
+                mismatched = {
+                    k: (v, secret_labels.get(k))
+                    for k, v in labels.items()
+                    if secret_labels.get(k) != v
+                }
+                if mismatched:
+                    if not silent:
+                        logger.debug(
+                            "Secret found but labels did not match",
+                            secret_name=secret_name,
+                            expected_labels=labels,
+                            actual_labels=secret_labels,
+                            mismatched_labels=mismatched,
+                        )
+                    return None
         except k8s_client_rest.ApiException as exc:
             if silent:
                 return
@@ -1052,12 +1060,15 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
         :return: SecretEventActions.{created, updated, skipped}
         """
         namespace = self.resolve_namespace(namespace)
-        secret_name = f"mlrun-auth-{username}-{token_name}"
-        labels = {"mlrun/user": username}
+        secret_name = self._resolve_user_token_secret_name(username, token_name)
+        labels = {
+            mlrun_constants.MLRunInternalLabels.user_token_secret_label_key: username
+        }
 
         logger.debug(
             "Preparing to store user token secret",
             secret_name=secret_name,
+            namespace=namespace,
         )
 
         secret_data = self._encode_user_token(token_name, token, expiration)
@@ -1096,6 +1107,11 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
             expiration=expiration,
         )
         return mlrun.common.schemas.SecretEventActions.skipped
+
+    def _resolve_user_token_secret_name(self, username: str, token: str) -> str:
+        return mlrun.mlconf.secret_stores.kubernetes.user_token_secret_name.format(
+            username=username, token_name=token
+        )
 
     def _encode_user_token(
         self, token_name: str, token: str, expiration: int
--- a/server/py/framework/utils/singletons/k8s.py
+++ b/server/py/framework/utils/singletons/k8s.py
@@ -1053,6 +1053,11 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
         secret_name = f"mlrun-auth-{username}-{token_name}"
         labels = {"mlrun/user": username}
 
+        logger.debug(
+            "Preparing to store user token secret",
+            secret_name=secret_name,
+        )
+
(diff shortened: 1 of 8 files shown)


server/py/framework/utils/singletons/k8s.py
```diff
@@ -1006,15 +1033,102 @@ def _init_k8s_config(self, log=True):
...
+
+        :param username: The user who owns the token.
+        :param token_name: The logical name for the token.
+        :param token: The offline token string (JWT).
+        :param expiration: The token's expiration timestamp (int UNIX epoch).
+        :param namespace: Kubernetes namespace for the secret.
+        :return: SecretEventActions.{created, updated, skipped}
+        """
+        namespace = self.resolve_namespace(namespace)
+        secret_name = f"mlrun-auth-{username}-{token_name}"
```
rokatyy: template can be stored in https://github.com/mlrun/mlrun/blob/1eddf750b68e74a11947877b4770a76c425fb738/mlrun/config.py#L686

server/py/framework/utils/singletons/k8s.py
```diff
@@ -1006,15 +1033,102 @@ def _init_k8s_config(self, log=True):
...
+        labels = {"mlrun/user": username}
```
rokatyy: add the label to constants

server/py/framework/utils/singletons/k8s.py
```diff
@@ -1006,15 +1033,102 @@ def _init_k8s_config(self, log=True):
...
+
+        logger.debug(
+            "Preparing to store user token secret",
+            secret_name=secret_name,
+        )
```
rokatyy:
```suggestion
        logger.debug(
            "Preparing to store user token secret",
            secret_name=secret_name,
            namespace=namespace,
        )
```

server/py/framework/utils/singletons/k8s.py
```diff
@@ -1006,15 +1033,102 @@ def _init_k8s_config(self, log=True):
...
+            return mlrun.common.schemas.SecretEventActions.created
+
+        # Update only if expiration is newer
+        if self._should_update_token_secret(k8s_secret, expiration):
+            self._update_secret(
+                k8s_secret=k8s_secret,
+                namespace=namespace,
+                secret_name=secret_name,
+                secrets=secret_data,
+            )
```
rokatyy, line 1101: question - is it possible that user tries to store different `token` value but we the oldest expiration date? Won't this behaviour be a bit confusing?

server/py/framework/utils/singletons/k8s.py
```diff
@@ -580,22 +582,45 @@ def store_secrets(
...
+                for key, value in labels.items():
+                    if secret_labels.get(key) != value:
+                        if not silent:
+                            logger.debug(
+                                "Secret found but labels did not match",
+                                secret_name=secret_name,
+                                expected_labels=labels,
+                                actual_labels=secret_labels,
+                            )
+                        return None
```
rokatyy:
```suggestion
    # Find any label that does not match
        mismatched = {k: (v, secret_labels.get(k)) for k, v in labels.items() if secret_labels.get(k) != v}
        if mismatched:
            if not silent:
                logger.debug(
                    "Secret found but labels did not match",
                    secret_name=secret_name,
                    expected_labels=labels,
                    actual_labels=secret_labels,
                    mismatched_labels=mismatched,
                )
            return None
```
to show all mismatched in a one shot

rokatyy (commented):
Nice and clean! 

Couple of suggestions

server/py/framework/utils/singletons/k8s.py
moranbental: with different token name?

rokatyy: @moranbental no, same name, different value of the token

moranbental: @rokatyy Yes, it’s possible, a user may try to store a token with the same name, but if that token is expired or close to expiring while the existing secret already holds a valid token, we skip updating the secret

rokatyy: @moranbental but what if user intentionally moves from a token that expires in 3month to the one that expires in 1 month? Maybe company policy has changed, why not updating it in this case?

moranbental: Initially, we considered enabling a force option but this was discussed in the HLD and we concluded that it is not needed - https://iguazio.atlassian.net/wiki/spaces/ARC/pages/361103361/MLRun+Secret+Tokens+in+IG4#3.1.3.-Listing-Secret-Tokens:~:text=Compare%20token%20expiration%20with%20existing%20token%20expiration%20and%20don%E2%80%99t%20allow%20updating%20to%20an%20older%20Offline%20Token.%20(Provide%20option%20to%20force%20update%20to%20old%20token).
see also slack - https://iguazio.slack.com/archives/C092ZSBKL4R/p1751294834737509

quaark: @rokatyy I think in the case of a company policy change - the Admin may revoke the users' tokens and they can create new ones. But I don't see another use case for allowing this.

rokatyy: @quaark and some of the applications may stop working for some time in this case?

rokatyy (approved):
LGTM

PR #8486, opened 2025-08-18
[Runs] Amit coding exercise Part 1

elbamit:
### 📝 Description

Coding exercise Part 1:
Add a new spec 'foo' in FunctionSpec.

---

### 🛠️ Changes Made

- Add 'foo' as part of FunctionSpec schema and class
- Refactor the client spec (schema and class) to include a default value for 'foo'
- Add a 'with_foo()' method to set 'foo' after the object creation
- (BONUS) - Add 'foo' as part of the KubeResourceSpec to enable setting it from a runtime dict.
---

### ✅ Checklist
- [X ] I have tested the changes in this PR

---

### 🧪 Testing
  
  
Unit test
- Test that foo is getting created with default value "test_foo_created()"
- Test that foo is getting created with a custom value from a runtime dict "test_foo_from_runtime()
- Test the 'with_foo()' method "test_with_foo()"
- Test that foo is getting created after changing the config value "test_foo_from_env()"
---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ONB-446
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [ X] No

---


TomerShor (approved):
Nice :)

mlrun/runtimes/pod.py
```diff
@@ -107,6 +107,7 @@ class KubeResourceSpec(FunctionSpec):
         "track_models",
         "parameters",
         "graph",
+        "foo",
```
moranbental: Why do you need it here?

moranbental (changes requested):
Very nice!
What about adding foo to the ApplicationSpec constructor?

liranbg (approved):
Very well!

mlrun/runtimes/pod.py
elbamit: Right. Removed from there and put it under base class FunctionSpec instead

quaark (approved):
🚀

elbamit:
> Very nice! What about adding foo to the ApplicationSpec constructor?

Done.

moranbental (approved):

PR #8494, opened 2025-08-19
[Requirements] Fix protobuf version in mlrun-kfp

gtopper:
By requiring latest v3io-frames, which explicitly requires protobuf 3.x.

Fixes [ML-10877](https://iguazio.atlassian.net/browse/ML-10877).

[ML-10877]: https://iguazio.atlassian.net/browse/ML-10877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


liranbg (changes requested):
we need to have kfp 1.8.23 - so downgrading is not an option.

liranbg (approved):
approved for now. we will work on the 1.8.23 separately .

PR #8500, opened 2025-08-20
[API] Moved system id sync to online phase

liranbg:
### 📝 Description

before the sync was on `_setup_service` which may be finished while chief/worker is yet to be online as they are pending migration and not all tables are ready yet.
now the sync is part of moving to online phase, where the data/schemas are mgirated and db is ready to be used.

---

### 🛠️ Changes Made

---

### ✅ Checklist

---

### 🧪 Testing
  
  

patched vm, verified migration, from scratch and simple scale up/down to see system id is returned on client-spec requests.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10929
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/server/py/framework/service/__init__.py
+++ b/server/py/framework/service/__init__.py
@@ -66,6 +66,7 @@ class Service(ABC):
 
     async def move_service_to_online(self):
         self._logger.info("Moving service to online", service_name=self.service_name)
+        await run_async_function_with_new_db_session(self._sync_system_metadata)
         await self._move_service_to_online()
 
     # https://fastapi.tiangolo.com/advanced/events/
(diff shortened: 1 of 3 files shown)


server/py/framework/service/__init__.py
```diff
@@ -119,7 +119,7 @@ def is_forwarded_request(self, request: fastapi.Request) -> bool:
 
     @abstractmethod
     async def _move_service_to_online(self):
-        pass
+        await run_async_function_with_new_db_session(self._sync_system_metadata)
```
TomerShor: You're defeating the purpose of Abstract methods - they should either be completely empty (just `pass`) or raise `NotImplementedError`.
either remove the wrapper or call this in each implementation

TomerShor (changes requested):

server/py/framework/service/__init__.py
liranbg: oh I completely missed that decorator

TomerShor (approved):

PR #8504, opened 2025-08-20
[Serving] Fix usage of LLModel as shared model

davesh0812:
### 📝 Description
When using LLModel as shared model the param such as result_path and input_path were not provided to the shared class. 

---

### 🛠️ Changes Made
1. Change `add_shared_model` signature
2. change `add_shared_proxy_model`  signature
3. make sure the model endpoints creation have all the relevant fields that were move between the signatures
---

### ✅ Checklist

---

### 🧪 Testing
Added sys and unit tests

---

### 🔗 References
https://iguazio.atlassian.net/browse/ML-10933

---

### 🚨 Breaking Changes?
SDK change (within 1.10.0)


mlrun/serving/states.py
```diff
@@ -596,15 +598,40 @@ def _verify_shared_models(
                     step.class_args[schemas.ModelRunnerStepData.MODELS][name][
                         schemas.ModelsData.MODEL_PARAMETERS.value
                     ]["shared_runnable_name"] = actual_shared_name
-                    shared_models.append(actual_shared_name)
             elif actual_shared_name != shared_runnable_name:
                 raise GraphError(
                     f"Model endpoint {name} shared runnable name mismatch: "
                     f"expected {actual_shared_name}, got {shared_runnable_name}"
                 )
+            shared_models.append(actual_shared_name)
```
royischoss, line 606: I think we need to take model uri in here so the endpoint schema will be updated in deployment if there is no outputs or inputs here

mlrun/serving/states.py
```diff
@@ -1349,54 +1380,61 @@ async def predict_async(
...
-        if origin_name and self.shared_proxy_mapping:
-            llm_prompt_artifact = self.shared_proxy_mapping.get(origin_name)
-            if isinstance(llm_prompt_artifact, str):
-                llm_prompt_artifact = self._get_artifact_object(llm_prompt_artifact)
-                self.shared_proxy_mapping[origin_name] = llm_prompt_artifact
-        else:
-            llm_prompt_artifact = (
-                self.invocation_artifact or self._get_artifact_object()
-            )
+        llm_prompt_artifact = llm_prompt_artifact
```
royischoss: this line should be deleted

mlrun/serving/states.py
```diff
@@ -1448,6 +1486,26 @@ def enrich_prompt(
...
+        :return: LLMPromptArtifact object or None if not found.
+        """
+        if origin_name and self.shared_proxy_mapping:
+            llm_prompt_artifact = self.shared_proxy_mapping.get(origin_name)
+            if isinstance(llm_prompt_artifact, str):
+                llm_prompt_artifact = self._get_artifact_object(llm_prompt_artifact)
+                self.shared_proxy_mapping[origin_name] = llm_prompt_artifact
+        else:
+            llm_prompt_artifact = (
+                self.invocation_artifact or self._get_artifact_object()
```
royischoss: Not related to this function, but noticed that _load_artifacts doesn't make any caching maybe it better to add it avoiding multiple mlrun.store_manager.get_store_artifact(uri)

mlrun/serving/states.py
```diff
@@ -1669,11 +1712,21 @@ def add_shared_model_proxy(
                 "model_artifact must be a string, ModelArtifact or LLMPromptArtifact"
             )
         root = self._extract_root_step()
```
royischoss, line 1743: I don't know if it a common scenario but this method after deployment is irrelevant, we might also warn the user that after deployment he must build a new graph using set topology if he changes it

mlrun/serving/states.py
```diff
@@ -596,15 +598,40 @@ def _verify_shared_models(
...
+            step.class_args[schemas.ModelRunnerStepData.MONITORING_DATA][name][
+                schemas.MonitoringData.RESULT_PATH
+            ] = shared_model_params["result_path"]
+            step.class_args[schemas.ModelRunnerStepData.MONITORING_DATA][name][
+                schemas.MonitoringData.INPUTS
+            ] = shared_model_params["inputs"]
+            step.class_args[schemas.ModelRunnerStepData.MONITORING_DATA][name][
+                schemas.MonitoringData.OUTPUTS
+            ] = shared_model_params["outputs"]
+            step.class_args[schemas.ModelRunnerStepData.MONITORING_DATA][name][
```
royischoss: why do we add model class here and in the add proxy but all others only here? maybe adding the class only here will do the trick

mlrun/serving/states.py
```diff
@@ -2819,15 +2919,17 @@ def add_shared_model(
         self.shared_models[name] = (model_class, model_parameters)
         self.shared_models_mechanism[name] = execution_mechanism
```
royischoss, line 2949: maybe refactor name to get shared model parameters

tests/system/model_monitoring/test_model_monitoring.py
```diff
@@ -97,7 +97,7 @@ class TestModelEndpointsOperations(TestMLRunSystemModelMonitoring):
     """Applying basic model endpoint CRUD operations through MLRun API"""
 
     project_name = "mm-app-project"
-    image = "mlrun/mlrun"
+    image = "artifactory.iguazeng.com:10557/davids/mlrun-api:1.10.0"
```
royischoss: don't forget this, and 1.10 is old use 1.11

royischoss (changes requested):

mlrun/serving/states.py
royischoss: I commented on `get_shared_model_name_by_artifact_uri`

mlrun/serving/states.py
davesh0812: will open a ticket for it

mlrun/serving/states.py
davesh0812: I added to `loads_artifact`

royischoss (approved):
Nice LGTM

PR #8505, opened 2025-08-20
[Hub] HubSource should support multiple object types 

danielperezz:
### 📝 Description

The HubSource object previously included an `object_type` field in its spec. As part of the effort to support modules in the functions hub, we’ve removed this field to avoid tying a hub to a specific object type. Instead, `object_type` is now passed explicitly as a parameter where needed, with 'functions' set as the default for backward compatibility.
---

### 🛠️ Changes Made

- removed the object-type field
- passing it as a parameter wherever needed
- added 'modules' as a new possible hub item type
- modified a unit test accordingly
- removed from `get_relative_path` the check of the source value since it anyway can't take any invalid value
---

### ✅ Checklist

---

### 🧪 Testing
  
  
No need in new tests since we only removed a field and did not add any new logic.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-9797
- Design docs links: -
- External links: -

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


mlrun/common/schemas/hub.py
```diff
@@ -39,14 +39,14 @@ class Config:
 # Currently only functions are supported. Will add more in the future.
```
Eyal-Danieli:
```suggestion

```

Eyal-Danieli (approved):
nice! just remove 1 comment line and let's wait for an update about the failed test

Eyal-Danieli (approved):

PR #8507, opened 2025-08-20
[Nuclio] Add project role when creating project in Nuclio

rokatyy:
### 📝 Description
Add project role when creating project in Nuclio

---

### 🛠️ Changes Made
we add this header in igz client only, which led to incorrect determination of request origin in Nuclio.

---

### ✅ Checklist

---

### 🧪 Testing
after vs before:
image
---

### 🔗 References
- Ticket link:
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [ ] No


server/py/framework/utils/clients/nuclio.py
```diff
@@ -229,6 +229,14 @@ def _send_request_to_api(
         if kwargs.get("timeout") is None:
             kwargs["timeout"] = 20
 
+        if "projects" in path:
+            if mlrun.common.schemas.HeaderNames.projects_role not in kwargs.get(
+                "headers", {}
+            ):
+                kwargs.setdefault("headers", {})[
+                    mlrun.common.schemas.HeaderNames.projects_role
+                ] = "mlrun"
```
TomerShor: This should only be applied in the case where mlrun is the leader and nuclio is the follower - not always.
This can be determined by the `leader` and `followers` config fields: https://github.com/mlrun/mlrun/blob/1eddf750b68e74a11947877b4770a76c425fb738/mlrun/config.py#L497-L500

TomerShor (changes requested):

server/py/framework/utils/clients/nuclio.py
rokatyy: @TomerShor agree reg a check that nuclio is a follower, but is it really necessary to check that mlrun is a leader, which scenarios it may break?

TomerShor: I don't know of any scenario that mlrun is not the leader but has followers, but if it is allowed - it shouldn't be.

moranbental: @rokatyy you can use `is_request_from_leader` for that

server/py/framework/utils/clients/nuclio.py
```diff
@@ -229,6 +229,14 @@ def _send_request_to_api(
         if kwargs.get("timeout") is None:
             kwargs["timeout"] = 20
 
+        if "projects" in path:
```
moranbental: the same code exists in the Iguazio client as well, move it into shared code

moranbental (changes requested):

server/py/framework/utils/clients/nuclio.py
rokatyy: @moranbental `is_request_from_leader` is for determining where the request came from. Which is not relevant in this case, because that's mlrun who sends a request and who should populate the projects_role

server/py/framework/utils/clients/nuclio.py
```diff
@@ -240,6 +245,7 @@ def _send_request_to_api(
         auth = None
         if auth_info:
             auth = auth_info.to_nuclio_auth_info().to_requests_auth()
+        print("headers", kwargs)
```
moranbental: ?

moranbental (changes requested):

server/py/framework/utils/clients/nuclio.py
rokatyy: @moranbental was debugging test, removed

TomerShor (approved):

moranbental (approved):

yaelgen (approved):

server/py/framework/utils/clients/helpers.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import mlrun.common.schemas
+
+
+def add_project_role_headers_if_needed(path, kwargs):
```
yaelgen: add type annotations please :)

liranbg (approved):

PR #8412, opened 2025-08-04
[Model Monitoring] Fix support for processing of legacy offline model endpoints

Eyal-Danieli:
Fix backward compatibility issue with old batch model endpoints in which we need need to keep supporting legacy batch model endpoints, typically generated through the batch inference process.

As part of the fix, we’ve restored code that was previously removed in [#8365](https://github.com/mlrun/mlrun/pull/8365) and added deprecation notices for the `record_results` and `get_or_create_model_endpoint` functions, which are located in `mlrun/model_monitoring/api.py`.

We also introduced a new field in the model endpoint schema: `mode`, which currently supports two values:

- real-time (0)
- batch (1)

Note that Legacy batch model endpoints are still monitored via the real-time controller process. Therefore, although their `endpoint_type` is batch, their `mode` will be marked as real-time.

https://iguazio.atlassian.net/browse/ML-10693

PR #8512, opened 2025-08-21
[Model Monitoring] Add `batch_legacy` mode for old batch model endpoints

Eyal-Danieli:
### 📝 Description
Following #8412   we added a new mode for the model endpoint schema which can be either real-time or batch. However, for legacy batch model endpoints we still use real-time, since they’re processed in the same way as real-time endpoints. 

In this PR, we’re adding a new `batch_legacy` mode for these endpoints to avoid mixing them up with real-time ones.

---

### 🛠️ Changes Made
- Add new endpoint mode called `BATCH_LEGACY`
- Update `list_model_endpoints` API to support filtering by multiple modes (and not only a single mode)

---

### ✅ Checklist

---

### 🧪 Testing
- Test the new modes in `test_list_filters` unit test 

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10845
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


mlrun/common/schemas/model_monitoring/model_endpoints.py
```diff
@@ -131,6 +131,15 @@ def _uid_to_str(cls, v):  # noqa: N805
             return str(v)
         return v
 
+    @validator("mode", pre=True, always=True)
+    def _set_mode_based_on_endpoint_type(cls, v, values):  # noqa: N805
```
jond01, line 135: I suggest
https://github.com/mlrun/mlrun/blob/cc74f73a862905d1eb2adcbb80c8c7fc72550f00/mlrun/model_monitoring/applications/results.py#L82
Instead of `noqa: N805`

mlrun/db/base.py
```diff
@@ -722,7 +722,9 @@ def list_model_endpoints(
         tsdb_metrics: bool = False,
         metric_list: Optional[list[str]] = None,
         top_level: bool = False,
-        mode: Optional[mlrun.common.schemas.EndpointMode] = None,
+        modes: Optional[
+            Union[mm_constants.EndpointMode, list[mm_constants.EndpointMode]]
+        ] = None,
```
jond01: I would consider keeping only the list option for simplicity, especially since it's an internal API

mlrun/projects/project.py
```diff
@@ -3944,7 +3944,9 @@ def list_model_endpoints(
         start: Optional[datetime.datetime] = None,
         end: Optional[datetime.datetime] = None,
         top_level: bool = False,
-        mode: Optional[mlrun.common.schemas.EndpointMode] = None,
+        modes: Optional[
+            Union[mm_constants.EndpointMode, list[mm_constants.EndpointMode]]
+        ] = None,
```
jond01, line 3951: Here, it can be OK to keep both options to help users

mlrun/model_monitoring/controller.py
```diff
@@ -801,7 +801,11 @@ def push_regular_event_to_controller_stream(self) -> None:
         logger.info("Starting monitoring controller chief")
         applications_names = []
         endpoints = self.project_obj.list_model_endpoints(
-            tsdb_metrics=False, mode=mm_constants.EndpointMode.REAL_TIME
+            tsdb_metrics=False,
+            modes=[
+                mm_constants.EndpointMode.REAL_TIME,
+                mm_constants.EndpointMode.BATCH_LEGACY,
+            ],
```
jond01, line 808: What will happen after 1.9.x -> 1.10.0 upgrade before upgrading the controller?

jond01 (commented):
Looks OK to me. I left some questions and suggestions in the comments.

I have another question too:
Could you please point to the place where we created a "batch" model endpoints?

mlrun/model_monitoring/controller.py
Eyal-Danieli: as far as I know as part of the upgrade you have to update the mm infra (including the controller image) as well. Anyway, even if you didn't, I don’t think it really matters here, since the controller will recognize it’s a batch model endpoint from the `endpoint_type` and then just check if there’s new data in the parquet files. There is no difference in that case between `batch_legacy` and new `batch` ep when you are using a 1.9 controller.

Eyal-Danieli:
@jond01 as for https://github.com/mlrun/mlrun/pull/8512#pullrequestreview-3171605547
`BATCH` model endpoint - through serving as a job flow https://github.com/mlrun/mlrun/blob/168317623abc7410a370ef111ce0cfce81e30818/server/py/services/api/crud/model_monitoring/deployment.py#L2327
`BATCH_LEGACT` model endpoint - through get_or_create_model_endpoint flow https://github.com/mlrun/mlrun/pull/8512#pullrequestreview-3171605547

jond01 (approved):

mlrun/common/schemas/model_monitoring/model_endpoints.py
jond01: The example is irrelevant, but the comment is still relevant
https://docs.pydantic.dev/1.10/usage/validators/
> validators are "class methods", so the first argument value they receive is the UserModel class, not an instance of UserModel.

PR #8514, opened 2025-08-21
[User secrets] Implement revoke user token secret [feature/ig4-authentication]

moranbental:
### 📝 Description

This PR introduces support for revoking user offline tokens in the Iguazio client. Revoked tokens can no longer be used to obtain access tokens. The feature includes both SDK and API support for deleting tokens and automatically deleting the associated Kubernetes secret.

---

### 🛠️ Changes Made

1. Added SDK method: `revoke_secret_token` to revoke a user token.
2. Added API endpoint: `DELETE /user-secrets/tokens/{name}` for token revocation.
3. Integrated token revocation with Iguazio backend.
4. Retrieves and deletes the corresponding Kubernetes secret after revocation.
5. Updated `create_secret` and `update_secret` methods to accept encoded values.

---

### ✅ Checklist

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10499, https://iguazio.atlassian.net/browse/ML-10498
- Design docs links: https://iguazio.atlassian.net/wiki/spaces/MLRUN/pages/404521061/BE+Secret+Token+Support+HLD#2.2.6-Token-Revocation-Flow
- External links: https://iguazio.atlassian.net/wiki/spaces/ARC/pages/361103361/MLRun+Secret+Tokens+in+IG4

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


server/py/framework/utils/clients/iguazio/v4.py
```diff
@@ -19,7 +19,10 @@
 # iguazio package is only supported in Python >= 3.11
 if sys.version_info >= (3, 11):
     import iguazio
-    from iguazio.schemas.v1.resources.access_token import RefreshAccessTokenOptions
+    from iguazio.schemas.v1.resources.access_token import (
+        RefreshAccessTokenOptions,
+        RevokeOfflineTokenOptions,
+    )
```
quaark:
```suggestion
    from iguazio.schemas import (
        RefreshAccessTokenOptionsV1,
        RevokeOfflineTokenOptionsV1,
    )
```

quaark (changes requested):
LGTM!
one minor thing - with how we're supposed to use iguazio sdk

server/py/framework/utils/clients/iguazio/v4.py
moranbental: @quaark I got this from the SDK documentation - https://vigilant-adventure-prnre3k.pages.github.io/examples/managing-tokens.html#refreshing-an-access-token, so should we update the documentation?

quaark: yes we should then

server/py/framework/utils/clients/iguazio/v4.py
```diff
@@ -105,6 +108,49 @@ def refresh_access_tokens(
         # TODO: Implement this method once it is available in the Iguazio package
         pass
 
+    def revoke_offline_token(self, token: str) -> None:
```
rokatyy, line 111: why not passing options here?

server/py/framework/utils/clients/iguazio/v4.py
```diff
@@ -105,6 +108,49 @@ def refresh_access_tokens(
...
+        except httpx.HTTPStatusError as exc:
+            error_message, ctx = self._extract_response_error(exc.response)
+            self._logger.warning(
+                "Failed to revoke offline token from Iguazio",
+                status_code=exc.response.status_code,
+                error_message=error_message,
+                ctx=ctx,
+                exc=mlrun.errors.err_to_str(exc),
+            )
+            raise mlrun.errors.MLRunUnauthorizedError(
```
rokatyy, line 142: How about adding `raise_on_error` option here or do we want to raise always?

server/py/services/api/crud/secrets.py
```diff
@@ -502,6 +502,64 @@ def store_secret_tokens(
...
+            logger.warning(
+                "Token not found, nothing to revoke",
+                username=authenticated_username,
+                token_name=token_name,
+            )
+            return
+
+        # Revoke via Iguazio
+        iguazio_client = framework.utils.clients.iguazio.v4.Client()
+        iguazio_client.revoke_offline_token(token)
```
rokatyy, line 567: Secrets CRUD sounds more like something to manage secrets. Token logic is something that holds secrets under the hood and not the other way around, right? So, should Secrets class be aware of iguazio client existence and all this revoke logic? Same question to method above

rokatyy (changes requested):
Initial review, have a general question regarding the abstractions

server/py/framework/utils/clients/iguazio/v4.py
moranbental: https://github.com/iguazio/orca/pull/293

server/py/framework/utils/clients/iguazio/v4.py
moranbental: I think we need to raise an error here anyway, because if the token cannot be deleted in Orca, it might indicate another problem, and this is our way to catch it

server/py/framework/utils/clients/iguazio/v4.py
moranbental: Why options? The only input we get from the user is the token to delete, and then we build the options

server/py/services/api/crud/secrets.py
moranbental: So where do you think I should put it? Because the endpoint isn’t the right place

server/py/framework/utils/clients/iguazio/v4.py
rokatyy: @moranbental I agree that for user-side it might be just token. But if we pass options here we 1. align with approach in orca, 2. if we have some additional options in the future, we won't need to modify this function

server/py/services/api/crud/secrets.py
rokatyy: @moranbental how about CRUD for tokens? or some token manager?

moranbental: will be in different task - https://iguazio.atlassian.net/browse/ML-11077

server/py/framework/utils/clients/iguazio/v4.py
moranbental: I spoke with @rokatyy offline, and we decided to leave it as is.

rokatyy (approved):
🚀

PR #8522, opened 2025-08-24
[Serving] Add a dummy stream by default to a tracked mock server

jond01:
### 📝 Description

Previously, when one created a serving function `fn` and marked it tracked with:
```py
fn.set_tracking()
```

And then wanted its mock server:
```py
fn.to_mock_server()
```

It would fail with a not very friendly error.

Now, we internally create a dummy output stream when necessary, and also allow directly passing `stream_profile` to `to_mock_server`.

---

### 🛠️ Changes Made

- Create a dummy stream by default for a monitored model serving function.
- Support `stream_profile` in `to_mock_server`.
- Do not create Kafka/V3IO clients in the relevant stream classes when marked as mock.

---

### ✅ Checklist

---

### 🧪 Testing

- Added new unit tests.
- Modified existing unit tests.

---

### 🔗 References
- Ticket link: [ML-10422](https://iguazio.atlassian.net/browse/ML-10422)

---

### 🚨 Breaking Changes?
- [x] No

[ML-10422]: https://iguazio.atlassian.net/browse/ML-10422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/serving/server.py
```diff
@@ -92,12 +93,15 @@ def __init__(self, enabled: bool, parameters: dict, function_uri: str):
...
+            stream_profile = parameters.get("stream_profile")
 
             if log_stream == DUMMY_STREAM:
                 # Dummy stream used for testing, see tests/serving/test_serving.py
                 stream_uri = DUMMY_STREAM
             elif not stream_args.get("mock"):  # if not a mock: `context.is_mock = True`
-                stream_uri = mlrun.model_monitoring.get_stream_path(project=project)
+                stream_uri = mlrun.model_monitoring.get_stream_path(
+                    project=project, profile=stream_profile
+                )
```
davesh0812: 'stream_uri' value is not used, lets fix this logic and understand when we really need it.

mlrun/serving/server.py
```diff
@@ -203,6 +210,21 @@ def init_states(
...
+            and monitoring_mock
+            and not (
+                self.parameters.get(FileTargetKind.LOG_STREAM)
+                or mlrun.get_secret_or_env(
+                    mm_constants.ProjectSecretKeys.STREAM_PROFILE_NAME
+                )
+            )
+        ):
+            # Set a dummy log stream for mocking purposes
+            self.parameters[FileTargetKind.LOG_STREAM] = DUMMY_STREAM
```
davesh0812: If I understand correctly the `stream_profile` is relevant only for the mock flow?
If it does lets do all of that under :
`if is_mock:`

and please add some inline comment for that.

davesh0812 (commented):
Nice, few comments

mlrun/serving/server.py
jond01: ~~It is used in L109 below~~
Done

mlrun/serving/server.py
jond01: Done

davesh0812 (approved):
LGTM

PR #8527, opened 2025-08-26
[HubSource] Update during init not only when url has changed

rokatyy:
### 📝 Description

Check whether the default hub source has changed and updates it from the configuration if so instead of updating only on url change.
---

### ✅ Checklist

---

### 🧪 Testing
  
  

---

### 🔗 References
required for https://github.com/mlrun/mlrun/pull/8266


yaelgen (approved):

TomerShor (approved):
LGTM

server/py/services/api/initial_data.py
```diff
@@ -328,13 +328,14 @@ def _add_default_hub_source_if_needed(
...
-    # update the default hub if configured url has changed
-    hub_source_path = hub_source.source.spec.path if hub_source else None
-    if not hub_source_path or hub_source_path != default_hub_source.spec.path:
+    if not hub_source:
+        # if not exists, just create one
+        _update_default_hub_source(db, db_session, default_hub_source)
+        return
+    # update the default hub if configuration has changed
+    if default_hub_source.diff(hub_source.source):
         mlrun.utils.logger.debug(
```
Yacouby, line 337: wouldn't it be useful to also log the previous hub source (not just the new one) so we can see what actually changed ?

server/py/services/api/initial_data.py
```diff
@@ -328,13 +328,14 @@ def _add_default_hub_source_if_needed(
         index=mlrun.common.schemas.hub.last_source_index,
         raise_on_not_found=False,
     )
-
-    # update the default hub if configured url has changed
-    hub_source_path = hub_source.source.spec.path if hub_source else None
-    if not hub_source_path or hub_source_path != default_hub_source.spec.path:
+    if not hub_source:
+        # if not exists, just create one
```
Yacouby:
```suggestion
        # create the default hub source if it does not exist
```

requirements.txt
```diff
@@ -45,3 +45,4 @@ mlrun-pipelines-kfp-v1-8~=0.5.7
 docstring_parser~=0.16
 # TODO: Remove it after moving notificaions to the server side(ML-8069)
 aiosmtplib~=3.0
+deepdiff~=7.0
```
Yacouby, line 48: shouldn't you update the locked requirements too ?

Yacouby (commented):
some minor comments

server/py/services/api/initial_data.py
rokatyy: @Yacouby actually we can even log difference object to avoid logging too much

Yacouby (approved):

TomerShor (approved):

PR #8529, opened 2025-08-27
[Auth] Implement IGTokenProvider [feature/ig4-authentication]

rokatyy:
### 📝 Description
Implements IGTokenProvider, so that 

---

### 🛠️ Changes Made
1. added new values to mlrun config:
```
"auth_with_oauth_token": {
# To enable igz4 auth
    "enabled": False,
    "request_timeout": 5,
    "refresh_threshold": 0.75,
# file with tokens
    "auth_token_file": "~/.igz.yml",
# default token
    "auth_token_name": ""
},
"auth_token_endpoint": ""
```
2. Offline Token Resolution
The IGTokenProvider resolves an offline token through one of two mechanisms:
    a. **From Environment Variable**
    If `MLRUN_AUTH_OFFLINE_TOKEN` is set, it is used as the Offline Token.

    b. **From YAML Token File**
    If the environment variable is not set, the SDK attempts to load a file from the path specified `config.auth_with_oauth_token.auth_token_file` (default: `~/.igz.yml`).
2. Getting token from file
get either `auth_token_name` (dynamically enriched to `default` if empty), if not found, get the first one. 

---

### ✅ Checklist

---

### 🧪 Testing
Unit tests (all new functionality is covered):
```
  mlrun/db/auth_utils.py                                                  182     67    63%   34, 38, 43, 46, 49, 56-73, 76-77, 80, 83, 86-97, 100-137, 142-154, 185, 206, 250, 263, 266-267, 281, 320
  ```

---

### 🔗 References
- Ticket link:https://iguazio.atlassian.net/browse/ML-10609
- Design docs links:https://iguazio.atlassian.net/wiki/spaces/MLRUN/pages/411960071/Support+sdk-side+IG4+authentication+-+token+usage+and+management+HLD

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes
Open question: Should we fail if file has more than one tokens with the same name or should we get the first one?

### CODE REVIEW

**Review by moranbental — CHANGES_REQUESTED** [review]
Please change the target branch of the merge to `feature/ig4-authentication`

File: mlrun/db/auth_utils.py [hunk 1]
```diff
@@ -150,3 +157,236 @@ def _parse_response(self, data: dict):
...
         )
+
+
+class IGTokenProvider(TokenProvider):
+    def __init__(self, token_endpoint: str):
+        if not token_endpoint:
+            raise mlrun.errors.MLRunValueError(
+                "No token endpoint provided, cannot initialize token provider"
+            )
+        self.token_endpoint = token_endpoint
```
moranbental: Why not a private field?

mlrun/db/auth_utils.py
```diff
@@ -150,3 +157,236 @@ def _parse_response(self, data: dict):
...
+        now = datetime.now()
+        remaining_lifetime = (self._token_expiry_time - now).total_seconds()
+
+        return (
+            remaining_lifetime / self._token_total_lifetime
+            > config.auth_with_oauth_token.refresh_threshold
+        )
+
+    @staticmethod
+    def get_token_lifetime_and_expiry(
```
moranbental: Could be a private method and moved below the `fetch_access_token` function

mlrun/db/auth_utils.py
```diff
@@ -150,3 +157,236 @@ def _parse_response(self, data: dict):
...
+                "POST",
+                self.token_endpoint,
+                timeout=config.auth_with_oauth_token.request_timeout,
+                headers=headers,
+                json=request_body,
+            )
+            if not response.ok and raise_on_error:
+                response.raise_for_status()
+                return
+            response_data = response.json()
```
moranbental: move it to different function for sending refresh request

mlrun/db/auth_utils.py
```diff
@@ -150,3 +157,236 @@ def _parse_response(self, data: dict):
             expiry=str(self.token_expiry_time),
             refresh=str(self.token_refresh_time),
         )
+
+
+class IGTokenProvider(TokenProvider):
```
moranbental: Many methods here have the same purpose as in `OAuthClientIDTokenProvider` and could be shared or made more generic. It might be better to create a separate generic class that both could implement.

mlrun/db/auth_utils.py
```diff
@@ -150,3 +157,236 @@ def _parse_response(self, data: dict):
...
+        """
+        Calculate the total lifetime and expiration time of the token.
+
+        :param token: The access token to decode.
+        :return: A tuple containing the total lifetime of the token in seconds and its expiration time as a datetime.
+        """
+        if not token:
+            return 0, None
+        try:
+            decoded_token = jwt.decode(token, options={"verify_signature": False})
```
moranbental: Add a comment here - as we don't verify its signature here, because it has already been verified earlier during the refresh access token call.

mlrun/db/auth_utils.py
```diff
@@ -150,3 +157,236 @@ def _parse_response(self, data: dict):
...
+            exp_timestamp = decoded_token.get("exp")
+            iat_timestamp = decoded_token.get("iat")
+            if exp_timestamp and iat_timestamp:
```
moranbental: What if `exp` or `iat` do not exist?

moranbental (changes requested):

mlrun/db/auth_utils.py
rokatyy: because it's not private in OAuthClientIDTokenProvider

mlrun/db/auth_utils.py
rokatyy: @moranbental which once except http request?

mlrun/db/auth_utils.py
rokatyy: then return 0, None below to unpack correctly and we will re-send request every time

mlrun/db/auth_utils.py
moranbental: but it can be also private field in OAuthClientIDTokenProvider

mlrun/db/auth_utils.py
moranbental: These methods are the same - get_token, is_iguazio_session, parts of the initialization, sending requests, and parsing the response, but each token provider will implement them differently.
I think some of the higher level functionality could be shared.

moranbental (changes requested):

mlrun/db/auth_utils.py
```diff
@@ -150,3 +157,240 @@ def _parse_response(self, data: dict):
...
+            return None
+
+        try:
+            with open(token_file) as token_file_io:
+                data = yaml.safe_load(token_file_io)
+        except yaml.YAMLError as exc:
+            mlrun.utils.helpers.raise_or_log_error(
+                f"Failed to parse token file {token_file}: {exc}", raise_on_error
+            )
+            return None
```
moranbental: Move this code into shared code for reading the token file (including checking whether the secretTokens list is missing or empty), since we will reuse the same code in `sync_secret_tokens` - https://iguazio.atlassian.net/browse/ML-10502

moranbental (changes requested):

mlrun/utils/helpers.py
```diff
@@ -2297,3 +2297,15 @@ def encode_user_code(
             "Consider using `with_source_archive` to add user code as a remote source to the function."
         )
     return encoded
+
+
+def raise_or_log_error(message: str, raise_on_error: bool):
```
moranbental:
```suggestion
def raise_or_log_error(message: str, raise_on_error: bool = True):
```

moranbental (changes requested):

mlrun/db/auth_utils.py
rokatyy: right, agree!

Then I'll put get_token+sending request logic into the base class and define methods in abstract class for building/parsing the requests, so that these methods for building/parsing and etc are implemented in each implementation

requirements.txt
```diff
@@ -44,3 +44,4 @@ mlrun-pipelines-kfp-v1-8~=0.4.3
 docstring_parser~=0.16
 # TODO: Remove it after moving notificaions to the server side(ML-8069)
 aiosmtplib~=3.0
+pyjwt~=2.10
```
rokatyy, line 47: run lock requirements after review

mlrun/auth/providers.py
```diff
@@ -0,0 +1,372 @@
+# Copyright 2024 Iguazio
```
moranbental: 2025

mlrun/auth/providers.py
```diff
@@ -0,0 +1,372 @@
...
+        """
+        Fetch a new access token from the token endpoint.
+
+        This method builds the token request, sends it to the token endpoint, and parses the response.
+        If the request fails, it either raises an error or logs a warning based on the `raise_on_error` parameter.
+
+        :param raise_on_error: Whether to raise an error if token retrieval fails.
+        """
+        try:
+            request_body, headers = self._build_token_request(raise_on_error=True)
```
moranbental, line 99: why raise_on_error is true?

mlrun/auth/providers.py
```diff
@@ -0,0 +1,372 @@
...
+    """
+
+    def __init__(self, token_endpoint: str, timeout=5):
+        if not token_endpoint:
+            raise mlrun.errors.MLRunValueError(
+                "No token endpoint provided, cannot initialize token provider"
+            )
+        self._token = None
+        self._token_endpoint = token_endpoint
+        self.timeout = timeout
```
moranbental: can be private field too

mlrun/auth/providers.py
```diff
@@ -0,0 +1,372 @@
...
+        Refresh the access token if it is expired or about to expire.
+
+        :param raise_on_error: Whether to raise an error if token refresh fails.
+        :return: The refreshed access token.
+        """
+        # Check if there is an existing access token and if it is valid
+        if self._token and self._is_token_valid(cleanup_if_expired=True):
+            return self._token
+
+        # Use the offline token to fetch a new access token
```
moranbental: for OAuthClientIDTokenProvider it is also an offline token?

mlrun/auth/providers.py
```diff
@@ -0,0 +1,372 @@
...
+    def __init__(
+        self, token_endpoint: str, client_id: str, client_secret: str, timeout=5
+    ):
+        if not token_endpoint or not client_id or not client_secret:
+            raise mlrun.errors.MLRunValueError(
+                "Invalid client_id configuration for authentication. Must provide token endpoint, client-id and secret"
+            )
+        # should be set before calling the parent constructor
+        self.client_id = client_id
+        self.client_secret = client_secret
```
moranbental: could be private fields

mlrun/auth/providers.py
```diff
@@ -0,0 +1,372 @@
...
+            )
+            return
+
+        self._token = access_token
+        self._token_total_lifetime, self._token_expiry_time = (
+            self.get_token_lifetime_and_expiry(access_token)
+        )
+
+    @staticmethod
+    def get_token_lifetime_and_expiry(
```
moranbental: can be private method

mlrun/auth/utils.py
```diff
@@ -0,0 +1,142 @@
...
+    :param raise_on_error: If True, raises an error when the offline token cannot be resolved.
+                           If False, logs a warning instead.
+    :return: The offline token if found, otherwise None.
+    """
+    if token_env := get_offline_token_from_env():
+        return token_env
+    return get_offline_token_from_file(raise_on_error=raise_on_error)
+
+
+def get_offline_token_from_file(raise_on_error: bool = True) -> typing.Optional[str]:
```
moranbental, line 41: Separate this function to call a separate function that reads the file and validates the secretTokens are valid, and then call the parse_offline_token_data func.Then, we can reuse it in sync_secret_tokens

moranbental (changes requested):
Very nice! few comments below

mlrun/auth/providers.py
rokatyy: @moranbental we catch it below and do not raise if flag in the method is set to False, we just need to exit and return earlier and do not send request.

mlrun/auth/providers.py
rokatyy: No, will delete this comment, the logic of fetching depends on an implementation

mlrun/auth/providers.py
moranbental: still not private

mlrun/auth/utils.py
moranbental: now this can call `load_secret_tokens_from_file` and then `parse_offline_token_data` https://github.com/mlrun/mlrun/pull/8567

mlrun/auth/utils.py
```diff
@@ -0,0 +1,160 @@
...
+    if not data:
+        mlrun.utils.helpers.raise_or_log_error("Empty token data", raise_on_error)
+        return None
+    tokens = data.get("secretTokens")
+    if not isinstance(tokens, list) or not tokens:
+        mlrun.utils.helpers.raise_or_log_error(
+            "Invalid token file: 'secretTokens' must be a non-empty list",
+            raise_on_error,
+        )
+        return None
```
moranbental, line 126: no need that anymore as it will be handled in `load_secret_tokens_from_file`

server/py/services/api/tests/unit/api/test_auth_igz4.py
```diff
@@ -0,0 +1,328 @@
+# Copyright 2025 Iguazio
```
moranbental: These tests are in the wrong place. They belong on the SDK side, not the API side. The tests should be located under `tests/auth/test_auth_utils.py` and `tests/auth/test_auth_providers.py`

mlrun/auth/providers.py
moranbental: need to delete this line

mlrun/auth/providers.py
moranbental: still not private

moranbental (changes requested):

moranbental (approved):

PR #8535, opened 2025-08-28
[Retry] Fix runs with hyperparams

rokatyy:
### 📝 Description
This PR fixes an issue with runs that include hyperparameters. Previously, if a run failed and one or more of its tasks were set to `pendingRetry`, those tasks were incorrectly treated as `running`(=healthy). As a result, the entire run was marked as `completed`.

With this fix, if any task is in the `pendingRetry` state, the run itself will correctly be set to `pendingRetry`. And then be retried. 

---

### ✅ Checklist

---

### 🧪 Testing
Tested on vmdev:

image
First is the run which was executed on the image that contained the fix. We can see that 3/3 retries happened, they failed and function was set to error at the end of those. Between retries state was `pending`.

Second is the one on which I was reproducing the error. It is marked as `completed` and we can see that it is stuck on 1/3 retries, even though the run failed. 

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10816

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No


liranbg (changes requested):
well done. worth covering / extending existing UT to cover that pending retry runs indeed  change their pending retry counter

rokatyy:
@liranbg this wasn't a bug, retry counter works and a bit out of scope for this bug. The problem was that run was set to completed instead of pending retry. UT can be added specifically for this function to make sure that it treats statuses of nested tasks correctly. Or I can add a system test

mlrun/runtimes/utils.py
```diff
@@ -164,17 +166,25 @@ def results_to_iter(results, runspec, execution):
...
                 failed += 1
                 err = get_in(task, ["status", "error"], "")
-                logger.error(f"error in task  {execution.uid}:{id} - {err_to_str(err)}")
-            elif state != "completed":
+                logger.error(f"error in task {execution.uid}:{id} - {err_to_str(err)}")
+            elif state == mlrun.common.runtimes.constants.RunStates.pending_retry:
+                pending_retry += 1
+                err = get_in(task, ["status", "error"], "")
+                logger.warning(
+                    f"pending retry in task {execution.uid}:{id} - {err_to_str(err)}"
```
Yacouby: it might be useful here to log also the retry count ( which is part of the execution )

Yacouby (changes requested):
its good to add a unit test to cover the case where a task is in `pending_retry`, to make sure the run is not marked as completed

tests/runtimes/test_utils.py
```diff
@@ -143,3 +144,57 @@ def test_enrich_run_labels(labels, labels_to_enrich, expected_labels, env_vars_t
             )
             == {}
         )
+
+
+def test_results_to_iter(rundb_mock):
```
Yacouby: Add a comment here to explain what the test checks, since from its name it is not understandable.

Yacouby: and maybe change the test's name as it does not verify the `results_to_iter` full functionality

tests/runtimes/test_utils.py
```diff
@@ -143,3 +144,57 @@ def test_enrich_run_labels(labels, labels_to_enrich, expected_labels, env_vars_t
...
+        },
+        {
+            "spec": {"parameters": {"p1": 2, "p2": 1}},
+            "status": {"state": "completed", "results": {"multiplier": 2.0}},
+        },
+        {
+            "spec": {"parameters": {"p1": 2, "p2": 2}},
+            "status": {"state": "completed", "results": {"multiplier": 1.0}},
+        },
+    ]
```
Yacouby, line 174: the third result here is not used right ? it is kind of similar to the second one so it can be removed to shorten the test

tests/runtimes/test_utils.py
```diff
@@ -143,3 +144,57 @@ def test_enrich_run_labels(labels, labels_to_enrich, expected_labels, env_vars_t
...
+    run = mlrun.run.RunObject.from_dict(run)
+
+    execution = mlrun.execution.MLClientCtx.from_dict(
+        run.to_dict(),
+        rundb_mock,
+        autocommit=False,
+        is_api=True,
+        store_run=False,
+    )
+    # Replace execution.commit with a no-op
```
Yacouby:
```suggestion
    # Replace execution.commit with a no-op to avoid persisting changes during test
```

tests/runtimes/test_utils.py
```diff
@@ -143,3 +144,57 @@ def test_enrich_run_labels(labels, labels_to_enrich, expected_labels, env_vars_t
...
+    execution.commit = lambda: None
+
+    mlrun.runtimes.utils.results_to_iter(results, run, execution)
+    assert execution.state == mlrun.common.runtimes.constants.RunStates.pending_retry
+
+    # replace to make them all successful
+    results[0] = results[1]
```
Yacouby: This line is a bit misleading since it mutates the list in a way that hides the intent.
If the goal is to skip the first result, it would be clearer to explicitly take the second one instead:
`new_result = results[1]`
and then use new_result

or slice it from the results: `results = results[1:]`

Yacouby (changes requested):
some more comments for better test readability

tests/runtimes/test_utils.py
rokatyy: @Yacouby the goal is to make all 3 objects successfully executed, so can be just sliced, but then I'd leave the 3rd object, just to have more than 1

Yacouby (approved):

TomerShor (approved):

PR #8542, opened 2025-09-01
[Log Collector] Create file before finding pod

TomerShor:
### 📝 Description

When the log collector can't find the pod to stream logs from, it fails and doesn't create a file.
If the run is deleted before the api calls `start_logs` on it again, it falls back to the legacy log collection, [which does a k8s api call to get the logs](https://github.com/mlrun/mlrun/blob/3398b8ade56e6474cf195be28686fdc9804b7115/server/py/services/api/runtime_handlers/base.py#L1669-L1696).
In case we have many pods in this state, we are doing many k8s api call which can choke the API.

To alleviate this a bit, we now first create an empty file for the run in the log collector, before trying to find the pod.
The legacy log collection will not run, as it first checks if the file exists.

---

### 🛠️ Changes Made
- Ensure the run has a log file before starting log collection.

---

### ✅ Checklist

---

### 🧪 Testing
- All tests pass

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10765
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/server/go/services/logcollector/server.go
+++ b/server/go/services/logcollector/server.go
@@ -228,6 +228,21 @@ func (s *Server) StartLog(ctx context.Context,
 		return s.successfulBaseResponse(), nil
 	}
 
+	// create a log file for the run so the legacy log collector will not try collect logs for it on deletion
+	logFilePath := s.resolveRunLogFilePath(request.ProjectName, request.RunUID)
+	if err := common.EnsureFileExists(logFilePath); err != nil {
+		s.Logger.ErrorWithCtx(ctx,
+			"Failed to ensure log file for run",
+			"runUID", request.RunUID,
+			"projectName", request.ProjectName,
+			"logFilePath", logFilePath)
+		return &protologcollector.BaseResponse{
+			Success:      false,
+			ErrorCode:    common.ErrCodeNotFound,
+			ErrorMessage: err.Error(),
+		}, err
+	}
+
 	var pod v1.Pod
 
 	s.Logger.DebugWithCtx(ctx, "Getting run pod using label selector", "selector", request.Selector)
--- a/server/go/services/logcollector/server.go
+++ b/server/go/services/logcollector/server.go
@@ -238,7 +238,7 @@ func (s *Server) StartLog(ctx context.Context,
 			"logFilePath", logFilePath)
 		return &protologcollector.BaseResponse{
 			Success:      false,
-			ErrorCode:    common.ErrCodeNotFound,
+			ErrorCode:    common.ErrCodeInternal,
 			ErrorMessage: err.Error(),
 		}, err
 	}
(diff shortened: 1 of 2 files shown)


server/go/services/logcollector/server.go
```diff
@@ -228,6 +228,21 @@ func (s *Server) StartLog(ctx context.Context,
 		return s.successfulBaseResponse(), nil
 	}
 
+	// create a log file for the run so the legacy log collector will not try collect logs for it on deletion
+	logFilePath := s.resolveRunLogFilePath(request.ProjectName, request.RunUID)
```
liranbg, line 232: worth adding ut coverage (case - failing to start finishes with log file)

server/go/services/logcollector/server.go
```diff
@@ -228,6 +228,21 @@ func (s *Server) StartLog(ctx context.Context,
...
+	if err := common.EnsureFileExists(logFilePath); err != nil {
```
Yacouby, line 233: Since EnsureFileExists is called unconditionally, is there a chance of overwriting/truncating an existing log file (e.g., if StartLog is retried)?

server/go/services/logcollector/server.go
```diff
@@ -228,6 +228,21 @@ func (s *Server) StartLog(ctx context.Context,
...
+		s.Logger.ErrorWithCtx(ctx,
+			"Failed to ensure log file for run",
+			"runUID", request.RunUID,
+			"projectName", request.ProjectName,
+			"logFilePath", logFilePath)
+		return &protologcollector.BaseResponse{
+			Success:      false,
+			ErrorCode:    common.ErrCodeNotFound,
```
Yacouby: This error code might be misleading, since the failure here is not about "not found" but rather "failed to create/ensure a file.
maybe return `ErrCodeInternal` to be consistent with other failures in this method ?

Yacouby (commented):
Some questions + would be great to have a unit test (either a new one or extending an existing one) that verifies the behavior when pod lookup fails — specifically, ensuring that an empty log file is still created

server/go/services/logcollector/server.go
TomerShor: No, this function only creates the file if it doesn't exist.
if it retries, it won't do anything.
https://github.com/mlrun/mlrun/blob/3abb47b4b0b489a00dee47b8741573d1560677ab/server/go/framework/common/utils.go#L83-L101

server/go/services/logcollector/server.go
TomerShor: Good catch 👍

liranbg (commented):
nicely done. minor request

server/go/services/logcollector/server.go
TomerShor: added.

Yacouby (approved):
LGTM

PR #8553, opened 2025-09-02
[Model Monitoring] Run app code only if `sample_df` is non-empty

jond01:
### 📝 Description

Skip windows that do not include data for the MEP.

---

### 🛠️ Changes Made

First, check that an endpoint is configured and that the sample data frame is empty.
When it is the case, return `None` from the app call and do not include this in the results to be written in the DBs.

---

### ✅ Checklist

---

### 🧪 Testing

Manual testing.

---

### 🔗 References
- Ticket link: [ML-10681](https://iguazio.atlassian.net/browse/ML-10681)

---

### 🚨 Breaking Changes?

- [x] No

[ML-10681]: https://iguazio.atlassian.net/browse/ML-10681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/model_monitoring/applications/base.py
+++ b/mlrun/model_monitoring/applications/base.py
@@ -347,6 +347,21 @@ class ModelMonitoringApplicationBase(MonitoringApplicationToDict, ABC):
                         feature_stats=feature_stats,
                     )
                 )
+
+                if (
+                    monitoring_context.endpoint_id
+                    and monitoring_context.sample_df.empty
+                ):
+                    # The current sample is empty
+                    context.logger.warning(
+                        "No sample data available for tracking",
+                        application_name=application_name,
+                        endpoint_id=monitoring_context.endpoint_id,
+                        start_time=monitoring_context.start_infer_time,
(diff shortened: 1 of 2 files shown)


mlrun/model_monitoring/applications/base.py
```diff
@@ -347,6 +347,21 @@ def call_do_tracking(event: Optional[dict] = None):
                         feature_stats=feature_stats,
                     )
                 )
+
+                if (
+                    monitoring_context.endpoint_id
```
assaf758, line 352: why `monitoring_context.endpoint_id` is part of the if clause? I would think only sample_df.empty is important in the decision not to send to the app.

jond01: So I don't want to entirely block the option to work with empty data frames in evaluate.
It can be useful with external data frames for testing purposes.
Here, we block it only when working with model endpoint data.

assaf758 (approved):

PR #8554, opened 2025-09-02
[Serving] Recover in case of MEP creation background task deletion

royischoss:
### 📝 Description
As part of the decision of pushing data to model monitoring stream we should know that model endpoints were created, scale made us avoid listing the model endpoints so we checked the status of the background task that created those model endpoints. As part of future change where we want to delete those background task added a fallback for the case of background task that is not found by listing the model endpoints once and caching this result in memory.

---

### 🛠️ Changes Made

Allowing all model serving to use logic of fallback to list_model_endpoint if background task was not found

---

### ✅ Checklist

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10928

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No


royischoss:
passed `test_app_flow[True, with_model_runner=True]` and `test_app_flow[True, with_model_runner=False]`

server/py/services/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -1030,6 +1031,7 @@ async def list_model_endpoints(
                                     will be included.
         :param uids:                A list of unique ids of the model endpoints.
         :param latest_only:         When True, only the latest model endpoint will be returned.
+        :param as_dict:             When True, the result will be returned as a dictionary of str, ModelEndpoint schema.
```
davesh0812: Can you specify the dictionary structure?

server/py/services/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -1009,6 +1009,7 @@ async def list_model_endpoints(
         metric_list: Optional[list[str]] = None,
         uids: typing.Optional[list[str]] = None,
         latest_only: typing.Optional[bool] = None,
+        as_dict: typing.Optional[bool] = None,
     ) -> mlrun.common.schemas.ModelEndpointList:
```
davesh0812: please add `dict`

server/py/services/api/api/endpoints/model_endpoints.py
```diff
@@ -272,6 +273,7 @@ async def list_model_endpoints(
     :param latest_only:     Whether to return only the latest model endpoint for each name.
     :param auth_info:       The auth info of the request.
     :param db_session:      A session that manages the current dialog with the database.
+    :param as_dict:         When True, the result will be returned as a dictionary of str, ModelEndpoint schema.
```
davesh0812: same

mlrun/common/model_monitoring/helpers.py
```diff
@@ -87,3 +89,95 @@ def pad_features_hist(feature_stats: FeatureStats) -> None:
     for feature in feature_stats.values():
         if hist_key in feature:
             pad_hist(Histogram(feature[hist_key]))
+
+
+def get_model_endpoints_creation_task_status(
+    server,
+    logging_method: typing.Callable,
+) -> tuple[
+    mlrun.common.schemas.BackgroundTaskState, typing.Optional[mlrun.utils.now_date]
```
davesh0812:
```suggestion
    mlrun.common.schemas.BackgroundTaskState, typing.Optional[datetime]
```

mlrun/common/model_monitoring/helpers.py
```diff
@@ -87,3 +89,95 @@ def pad_features_hist(feature_stats: FeatureStats) -> None:
...
+        )
+    if background_task is None:
+        try:
+            model_endpoints = mlrun.get_run_db().list_model_endpoints(
+                project=server.project,
+                function_name=server.function_name,
+                function_tag=server.function_tag,
+                as_dict=True,
+            )
+            if model_endpoints:
```
davesh0812: we should check also that we have all the relevant endpoints no? maybe it is enough to check the expected length..

mlrun/common/model_monitoring/helpers.py
```diff
@@ -87,3 +89,95 @@ def pad_features_hist(feature_stats: FeatureStats) -> None:
...
+            else:
+                logger.warning(
+                    "Model endpoints not found after background task not found, model monitoring will not "
+                    "monitor events",
+                    project=server.project,
+                    function_name=server.function_name,
+                    function_tag=server.function_tag,
+                )
+                background_task_state = mlrun.common.schemas.BackgroundTaskState.failed
+        except mlrun.errors.MLRunNotFoundError:
```
davesh0812: When you think we will get this error? I think that in case there are no endpoints we will get empty dict/list no?

mlrun/common/model_monitoring/helpers.py
```diff
@@ -87,3 +89,95 @@ def pad_features_hist(feature_stats: FeatureStats) -> None:
     for feature in feature_stats.values():
         if hist_key in feature:
             pad_hist(Histogram(feature[hist_key]))
+
+
+def get_model_endpoints_creation_task_status(
+    server,
+    logging_method: typing.Callable,
```
davesh0812: do we have diffrent `logging_method`?

server/py/services/api/api/endpoints/model_endpoints.py
```diff
@@ -248,6 +248,7 @@ async def list_model_endpoints(
     latest_only: bool = Query(False, alias="latest-only"),
     auth_info: schemas.AuthInfo = Depends(framework.api.deps.authenticate_request),
     db_session: Session = Depends(deps.get_db_session),
+    as_dict: Optional[bool] = Query(None, alias="as_dict"),
 ) -> schemas.ModelEndpointList:
```
davesh0812: please add dict

mlrun/serving/v2_serving.py
```diff
@@ -474,22 +478,18 @@ def _initialize_model_logger(self):
         ) or getattr(self.context, "server", None)
         if not self.context.is_mock or self.context.monitoring_mock:
             if server.model_endpoint_creation_task_name:
-                background_task = mlrun.get_run_db().get_project_background_task(
-                    server.project, server.model_endpoint_creation_task_name
-                )
-                logger.debug(
-                    "Checking model endpoint creation task status",
-                    task_name=server.model_endpoint_creation_task_name,
+                background_task_state, _ = get_model_endpoints_creation_task_status(
```
davesh0812: Why here the time is not relevant?

davesh0812 (commented):
Nice, few comments

mlrun/serving/v2_serving.py
royischoss: We check it once

mlrun/common/model_monitoring/helpers.py
royischoss: Nope :) changed it

mlrun/common/model_monitoring/helpers.py
royischoss: Added as discussed added to the system steps

mlrun/common/model_monitoring/helpers.py
royischoss: removed it

mlrun/serving/system_steps.py
```diff
@@ -317,6 +333,11 @@ class BackgroundTaskStatus(storey.MapClass):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
+        self.matching_endpoints: MatchingEndpointsState = (
+            MatchingEndpointsState.not_yet_checked
+        )
+        self.graph_model_endpoint_uids = None
+        self.listed_model_endpoint_uids = []
```
davesh0812: lets define it as sets from first place.

davesh0812 (commented):
LOOKS GOOD. I have 1 minor.. and please run test_app if you didn't run it yet

davesh0812 (approved):

PR #8594, opened 2025-09-09
[Retry] Ignore outdated retry pods in monitor to prevent scheduler race

Yacouby:
### 📝 Description
This PR fixes a race condition in the runs monitor where pods from previous retry attempts were incorrectly considered when updating the run state.
The issue could cause the scheduler to start overlapping runs because the DB temporarily marked a retried run as error.

Now, the monitor ignores outdated pods and only tracks the latest retry attempt, ensuring correct run state and preventing duplicate runs.

---

### 🛠️ Changes Made

- Added `_is_pod_from_outdated_retry` to detect pods from outdated retry attempts.
- Updated monitor logic to skip outdated pods when determining run state.
- Improved debug logging for better visibility when updating a run's state.

---

### ✅ Checklist

---

### 🧪 Testing

- Added a unit test to test the new functionality added.
- Manual validation on patched vm showed that the scheduler no longer triggers overlapping runs for jobs with retries.
- Modified an existing unit test (`test_monitor_run_retry_exhausted`) to label pods with their retry attempt, since unlabeled pods would otherwise be skipped by the monitor logic and fail that test.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10803

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

### DIFF
--- a/server/py/services/api/runtime_handlers/base.py
+++ b/server/py/services/api/runtime_handlers/base.py
@@ -2039,11 +2039,8 @@ class BaseRuntimeHandler(ABC):
         )
         run_retry_count = run.get("status", {}).get("retry_count") or 0
 
-        # If no retry label is present, treat pod as initial attempt (normalized to -1).
-        pod_retry_attempt = int(pod_retry_label) if pod_retry_label is not None else -1
+        if pod_retry_label is None:
+            # pods without a retry label are outdated once retries have started
+            return run_retry_count > 0
 
-        # first run with unlabeled pod is not outdated
-        if run_retry_count == 0 and pod_retry_attempt == -1:
-            return False
-
-        return pod_retry_attempt < run_retry_count
+        return int(pod_retry_label) < run_retry_count
--- a/server/py/services/api/runtime_handlers/base.py
+++ b/server/py/services/api/runtime_handlers/base.py
@@ -636,7 +636,11 @@ class BaseRuntimeHandler(ABC):
 
                 # Check if the run should be retried, and update its status accordingly
                 run_state, message = self._evaluate_run_retry_state(run, reason)
-                logger.info("Updating run state", run_uid=run_uid, run_state=run_state)
+                logger.info(
+                    "Updating run state - non terminal recovery flow",
+                    run_uid=run_uid,
+                    run_state=run_state,
+                )
                 run_updates = {
                     "status.state": run_state,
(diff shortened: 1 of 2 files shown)


weilerN (approved):

server/py/services/api/runtime_handlers/base.py
```diff
@@ -1767,7 +1780,11 @@ def _ensure_run_state(
                     run, reason, message
                 )
 
-        logger.info("Updating run state", run_uid=uid, run_state=run_state)
+        logger.info(
+            "Updating run state - ensure_run_state flow",
```
TomerShor: What about:

```suggestion
            "Ensuring run state",
```

server/py/services/api/tests/unit/runtime_handlers/test_kubejob.py
```diff
@@ -972,6 +977,33 @@ async def test_retry_pod_deleted_before_first_attempt(
...
+            # first run, no retry label means pod is valid and not outdated
+            (None, 0, False),
+            # retry count > 0 and no retry label is present, pod is outdated
+            (None, 1, True),
+            # pod attempt is older than current run retry, pod is outdated
+            ("1", 2, True),
+            # pod attempt equals current run retry, pod is still valid
+            ("2", 2, False),
+        ],
+    )
```
TomerShor: Is an edge case possible where `pod_retry_label > run_retry_count`?
If so - what do we do in this case?

TomerShor (changes requested):
Looking good!
Minor comments

server/py/services/api/runtime_handlers/base.py
Yacouby: mmm
I think “updating run state” fits better, as we’re actually performing the update, not just ensuring it.

server/py/services/api/tests/unit/runtime_handlers/test_kubejob.py
Yacouby: In this edge case, the pod is treated as **not** outdated to avoid skipping a potentially valid, still-running attempt. Added a test case to cover this scenario.

liranbg (approved):

TomerShor (approved):

PR #8599, opened 2025-09-10
[Invoke] Fix invocation URL resolution

rokatyy:
### 📝 Description
Fixing the bug which led to function never being invoked by internal url if external url is empty.
The issue arises from the mismatch in how client and server handle function addresses.

Old behaviour:
**Server side (deployment)**
When a function is deployed, we set `function.spec.address` as `address = external_invocation_urls[0] if external_invocation_urls else ""`
This makes sense historically and for backward compatibility, since the address field has always represented the external URL.

**Client side (invocation)**
When invoking, the client runs through a fairly complex resolution flow:
First checks if `self.status.address` is set.
If not set, it validates trigger configuration.
If default http trigger is disabled and no custom http trigger exists → raises `MLRunPreconditionFailedError`.
Otherwise, it eventually still fails on the emptiness of self.status.address:
```
if not self.status.address:
    raise ValueError("no function address first run .deploy()")
```
Because of this, we never reach:
`path = self._resolve_invocation_url(path, force_external_address)`

This behaviour is incorrect. Even if address is empty (no external URL), the client should still be able to resolve the internal invocation URL — especially that there is a flag `force_external_address=False`.

---

### 🛠️ Changes Made
New logic:

* Always check function state first and warn if it’s not ready.
* Prefer internal URL when running inside Kubernetes and force_external_address=False.
* Otherwise, fall back to the external URL if available.
* If neither exists, use the legacy address for backward compatibility.
* If no address or valid HTTP trigger exists, raise a clear error.
* Log explicitly whether invocation is happening via internal URL for transparency.

---

### ✅ Checklist

---

### 🧪 Testing
unit tests

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10972
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No


mlrun/runtimes/nuclio/function.py
```diff
@@ -1228,19 +1210,47 @@ def _resolve_invocation_url(self, path, force_external_address):
...
             and self.status.internal_invocation_urls
             and mlrun.k8s_utils.is_running_inside_kubernetes_cluster()
         ):
-            return mlrun.utils.helpers.join_urls(
+            url = mlrun.utils.helpers.join_urls(
                 f"http://{self.status.internal_invocation_urls[0]}", path
             )
+            logger.info(
+                f"Using internal invocation url {url}. Make sure you have network access to the k8s cluster. "
+                f"Otherwise, set force_external_address to True"
```
TomerShor, line 1230: Why do we need to tell the user to make sure they have access, although we just checked that `mlrun.k8s_utils.is_running_inside_kubernetes_cluster() == True`?

mlrun/runtimes/nuclio/function.py
```diff
@@ -1228,19 +1210,47 @@ def _resolve_invocation_url(self, path, force_external_address):
...
+            )
+            return url
 
         if self.status.external_invocation_urls:
             return mlrun.utils.helpers.join_urls(
                 f"http://{self.status.external_invocation_urls[0]}", path
```
TomerShor, line 1236: not related to your changes, but why is it http and not https?...
we might want to at least have it configurable

TomerShor (commented):
Nice!
Minor comments

mlrun/runtimes/nuclio/function.py
rokatyy: @TomerShor what if it's another k8s cluster?

mlrun/runtimes/nuclio/function.py
rokatyy: I think it's better to be handled in a separate PR, because there are also other places with http:// hardcoded:

https://github.com/mlrun/mlrun/blob/d937925964752a6a218268779df2d8ce4be66b31/mlrun/runtimes/nuclio/function.py#L677-L693

mlrun/runtimes/nuclio/function.py
```diff
@@ -1228,19 +1210,47 @@ def _resolve_invocation_url(self, path, force_external_address):
...
         if (
             not force_external_address
             and self.status.internal_invocation_urls
             and mlrun.k8s_utils.is_running_inside_kubernetes_cluster()
         ):
-            return mlrun.utils.helpers.join_urls(
+            url = mlrun.utils.helpers.join_urls(
                 f"http://{self.status.internal_invocation_urls[0]}", path
             )
+            logger.info(
```
Yacouby: Wouldn't this be too verbose for common invocations ?
if yes then maybe better to be in debug

mlrun/runtimes/nuclio/function.py
```diff
@@ -968,24 +968,6 @@ def invoke(
...
-                    raise mlrun.errors.MLRunPreconditionFailedError(
-                        "Default http trigger creation is disabled and there is no any other custom http trigger, "
-                        "so function can not be invoked via http. Either enable default http trigger creation or "
-                        "create custom http trigger"
-                    )
-                state, _, _ = self._get_state()
-                if state not in ["ready", "scaledToZero"]:
-                    logger.warning(f"Function is in the {state} state")
-                if not self.status.address:
-                    raise ValueError("no function address first run .deploy()")
```
Yacouby, line 987: Removing the early `status.address` check in changes the error-handling flow. did you check that no other components or user code depend on the previous ValueError behavior ?

server/py/services/api/tests/unit/runtimes/test_nuclio.py
```diff
@@ -1857,6 +1860,116 @@ def test_masking_sensitive_fields(self):
...
+                None,
+                mlrun.errors.MLRunPreconditionFailedError,
+                True,
+            ),
+            # Error if no address and no triggers, default http trigger enabled
+            (True, False, [], [], "", None, ValueError, False),
+            (False, True, [], [], "", None, ValueError, False),
+        ],
+    )
+    def test_get_url(
```
Yacouby, line 1926: This test mainly checks `get_url`, but the bug fix affects invoke. wht about adding  a  test that directly calls invoke with only internal URLs (mocking the request) to verify the full invocation flow.

Yacouby (commented):
Minor comments

mlrun/runtimes/nuclio/function.py
rokatyy: @Yacouby this is exactly the bug, so it would be weird if users flow rely on it. I think in 99 percent of the cases users have external urls, so they really never caught it

server/py/services/api/tests/unit/runtimes/test_nuclio.py
rokatyy: @Yacouby all the logic now was moved to the resolving url method, which test_url uses under the hood, what is the point of testing the whole invoke? Invoke is used quite widely across all the tests..

Yacouby (approved):
LGTM

mlrun/runtimes/nuclio/function.py
TomerShor: agreed

mlrun/runtimes/nuclio/function.py
TomerShor: You already have a log line for every invocation: https://github.com/mlrun/mlrun/pull/8599/files#diff-91002e128e55ae0b16d8d62abbd2c9f54e83d5d584a287b43f0df436df7b2f88L1007
maybe worth adding it there.

PR #8602, opened 2025-09-11
[Datastore] Handle Azure container as part of URL

alxtkr77:
### 📝 Description

This PR introduces enhanced Azure Blob Storage support in `mlrun`.
It adds full handling for multiple Azure connection string formats, proper container extraction for `wasbs://` URLs, and improves error handling for malformed configurations.
It also includes extensive unit tests for Azure Blob behavior and updates integration/system test configurations to align with the new implementation.

---

### 🛠️ Changes Made

* **Azure Blob Store Enhancements**:

  * Improved container extraction logic:

    * Support for `wasbs://container@host` URL format.
    * Container can now be derived from endpoint, `storage_options`, or connection string.
  * Added robust error messages for missing container or account name.

* **Testing Improvements**:

  * Added a new, comprehensive unit test suite: `tests/datastore/test_azure_blob_store.py`.
  * Updated Azure integration test configuration with a valid SAS-based connection string.

---

### ✅ Checklist

* [x] I updated the documentation (docstring and comments added).
* [x] I have tested the changes in this PR:

  * Unit tests
  * Integration tests
  * System tests

---

### 🧪 Testing

* **Unit tests**:

  * Added extensive tests for container extraction, connection string handling, and Spark URL generation.
  * Verified error handling for missing container and malformed connection strings.
* **Integration tests**:

---

### 🔗 References

* Internal ticket: [ML-10748](https://iguazio.atlassian.net/browse/ML-10748), [ML-11003](https://iguazio.atlassian.net/browse/ML-11003)
* Related Azure documentation:
  [[Azure Storage Connection Strings](https://learn.microsoft.com/en-us/azure/storage/common/storage-configure-connection-string)](https://learn.microsoft.com/en-us/azure/storage/common/storage-configure-connection-string)

---

### 🚨 Breaking Changes?

* [ ] Yes
* [x] No

---

[ML-10748]: https://iguazio.atlassian.net/browse/ML-10748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/datastore/azure_blob.py
+++ b/mlrun/datastore/azure_blob.py
@@ -1,4 +1,4 @@
-# Copyright 2023 Iguazio
+# Copyright 2025 Iguazio
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import contextlib
 import time
 from pathlib import Path
 from typing import Optional
@@ -30,6 +31,40 @@ from .base import DataStore, FileStats, make_datastore_schema_sanitizer
 
 
 class AzureBlobStore(DataStore):
+    """
+    Azure Blob Storage datastore implementation.
+
+    Supports multiple URL schemas: az://, wasbs://, wasb://
+
+    Supported Connection String Formats:
+    ====================================
+
+    1. Account Key (Standard):
+       "DefaultEndpointsProtocol=https;AccountName=<account>;AccountKey=<key>;EndpointSuffix=core.windows.net"
+
+    2. SAS Token:
+       "BlobEndpoint=https://<account>.blob.core.windows.net/;SharedAccessSignature=<sas_token>"
+
+    3. Minimal BlobEndpoint:
+       "BlobEndpoint=https://<account>.blob.core.windows.net/;AccountName=<account>;AccountKey=<key>"
+
+    4. Custom Domain:
+       "BlobEndpoint=https://<account>.mydomain.com/;AccountName=<account>;AccountKey=<key>"
+
+    5. China/Government Cloud:
+       "DefaultEndpointsProtocol=https;AccountName=<account>;AccountKey=<key>;EndpointSuffix=core.chinacloudapi.cn"
+
+    6. Full Service Endpoints with SAS:
+       "BlobEndpoint=https://<account>.blob.core.windows.net/;QueueEndpoint=...;SharedAccessSignature=<sas>"
+
+    Authentication Methods:
+    ======================
+    - Account Key (connection_string or storage_options)
+    - SAS Token (connection_string or storage_options)
+    - OAuth/Azure AD (storage_options: client_id, client_secret, tenant_id)
+
+    """
+
     using_bucket = True
     max_concurrency = 100
     max_blocksize = 1024 * 1024 * 4
@@ -40,6 +75,12 @@ class AzureBlobStore(DataStore):
     def __init__(
         self, parent, schema, name, endpoint="", secrets: Optional[dict] = None
     ):
+        # Extract container from WASBS endpoint before calling super()
+        self._container_from_endpoint = None
+        if schema in ["wasbs", "wasb"] and endpoint and "@" in endpoint:
+            # Handle container@host format
+            self._container_from_endpoint, endpoint = endpoint.split("@", 1)
+
         super().__init__(parent, name, schema, endpoint, secrets=secrets)
         self._service_client = None
         self._storage_options = None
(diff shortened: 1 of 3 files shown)


mlrun/datastore/azure_blob.py
```diff
@@ -308,18 +384,23 @@ def spark_url(self):
...
-            primary_url, _, _ = parse_connection_str(
-                connection_string, credential=None, service=service
-            )
-            if primary_url.startswith("http://"):
-                primary_url = primary_url[len("http://") :]
-            if primary_url.startswith("https://"):
-                primary_url = primary_url[len("https://") :]
-            host = primary_url.rstrip("/")
-
+            with contextlib.suppress(Exception):
```
tomerm-iguazio, line 387: can we talk about that?
Why we skip errors?

mlrun/datastore/utils.py
```diff
@@ -331,8 +337,8 @@ def parse_url(url):
         endpoint = netloc[
             hostname_index_in_netloc : hostname_index_in_netloc + len(lower_hostname)
         ]
-    if parsed_url.port:
-        endpoint += f":{parsed_url.port}"
+        if parsed_url.port:
```
tomerm-iguazio, line 340: so if we use wasps/wasb we do not need a port?

tests/datastore/test_azure_blob_store.py
```diff
@@ -0,0 +1,614 @@
...
+    @patch("mlrun.datastore.azure_blob.parse_connection_str")
+    def test_get_spark_options_connection_string_parsing(self, mock_parse_conn):
+        """Test Spark options generation with connection string parsing"""
+        mock_parse_conn.return_value = (
+            "igztestdata.blob.core.windows.net",
+            None,
+            {
+                "account_name": "igztestdata",
+                "sas_token": (
+                    "sv=2019-12-12&ss=bfqt&srt=sco&sp=rwdlacupx"
```
tomerm-iguazio: what is this key example?

tests/datastore/test_azure_blob_store.py
```diff
@@ -0,0 +1,614 @@
...
+        store = self._create_store(schema="az", endpoint="endpoint-container")
+
+        mock_storage_options = {
+            "container": "storage-options-container",
+            "account_name": "teststorage",
+        }
+        with patch.object(store, "_storage_options", mock_storage_options):
+            result = store.spark_url
+            expected = (
+                "wasbs://storage-options-container@teststorage.blob.core.windows.net"
```
tomerm-iguazio, line 294: it should not stay "az" schema instead of wasps?

tests/datastore/test_azure_blob_store.py
```diff
@@ -0,0 +1,614 @@
...
+
+            mock_storage_options = {
+                "connection_string": "DefaultEndpointsProtocol=https;AccountName=test;",  # Missing AccountKey
+                "container": "mycontainer",
+            }
+            with patch.object(store, "_storage_options", mock_storage_options):
+                # Should fallback gracefully when connection string parsing fails
+                with pytest.raises(
+                    mlrun.errors.MLRunInvalidArgumentError,
+                    match="account_name is required",
```
tomerm-iguazio, line 416: account_name or account key?
because you miss account key

tests/datastore/test_azure_blob_store.py
```diff
@@ -0,0 +1,614 @@
...
+                # Should use container from storage_options and host from connection string
+                expected = (
+                    "wasbs://storage-options-container@mixedauth.blob.core.windows.net"
+                )
+                assert result == expected
+
+    def test_spark_url_wasbs_container_from_endpoint_url(self):
+        """Test that wasbs:// URLs can extract container from endpoint netloc"""
+        store = self._create_store(
+            schema="wasbs", endpoint="igztestcbs@igztestdata.blob.core.windows.net"
```
tomerm-iguazio: why you use real account name hardcoded?
I can see this issue in different places in the code.

tests/datastore/test_azure_blob_store.py
```diff
@@ -0,0 +1,614 @@
...
+        }
+        with patch.object(store, "_storage_options", mock_storage_options):
+            result = store.spark_url
+            expected = "wasbs://igztestcbs@igztestdata.blob.core.windows.net"
+            assert result == expected
+
+    def test_spark_url_wasbs_container_from_connection_string_blob_endpoint(self):
+        """Test that wasbs:// URLs can extract container from BlobEndpoint in connection string"""
+        secrets = {
+            "connection_string": "BlobEndpoint=https://igztestdata.blob.core.windows.net/igztestcbs/;SharedAccessSignature=sv=2019-12-12&ss=bfqt&srt=sco&sp=rwdlacupx&se=2022-01-01T00:00:00Z&st=2021-01-01T00:00:00Z&spr=https&sig=test",
```
tomerm-iguazio: where is this connection string from?

tomerm-iguazio (changes requested):
Hey, 
watch out from hardcoding real account names or connection strings.

mlrun/datastore/azure_blob.py
alxtkr77: The idea is that if we have both connection_string and account_name, and parse_connection_str() fails we still build the host as 'host = f"{account_name}.{service}.core.windows.net"'

mlrun/datastore/utils.py
alxtkr77: Because in the case of wasps/wasb , the endpoint is 'parsed_url.netloc' and it has port already embedded. Other cases use 'parsed_url.hostname' instead

tests/datastore/test_azure_blob_store.py
alxtkr77: Will replace with clearly dummy one

tests/datastore/test_azure_blob_store.py
alxtkr77: No, the expected string should remain wasbs://. The spark_url property always returns wasbs:// format regardless of the original schema because that's the format Spark requires for Azure Blob Storage integration.

tests/datastore/test_azure_blob_store.py
alxtkr77: account_name. Since we don't have the account_name in mock_storage_options, and we mock a malformed connection string that causes parse_connection_str() to raise (Missing AccountKey), the test looks for  'account_name ' exception

tests/datastore/test_azure_blob_store.py
alxtkr77: Will replace with a dummy one

tests/datastore/test_azure_blob_store.py
alxtkr77: Will replace with a clearly dummy one

tomerm-iguazio (approved):
Approved, pay attention that the account name and bucket at first were real, and it is a bad practice.
Keys: The keys were generated and not be connected to our system according to you.

assaf758:
> Approved, pay attention that the account name and bucket at first were real, and it is a bad practice. Keys: The keys were generated and not be connected to our system according to you.

Since I'm squashing the commits to one PR as part of the merge, early versions of the PR will have no trace - so I think it's safe.

PR #8612, opened 2025-09-12
[Authentication] Improve failure handling when token exchange fails [feature/ig4-authentication]

rokatyy:
### 📝 Description
Refresh Fallback Flow:

0. If the token refresh fails (i.e., `fetch_token()` raises an exception), retry the 1 and 2 steps up to 3 attempts:
1. Re-read the offline token from the configured file (to capture potential external updates such as rotation).
2. Retry exchanging the offline token at the token_endpoint.
3. If all refresh attempts fail, delegate handling to _post_fetch_hook:
        3.1 If a cached access token is still valid (not fully expired), it will continue to be used.
        3.2 If the token is expired, raise an error.
---

### ✅ Checklist

---

### 🧪 Testing
UT

---

### 🔗 References
- Ticket link:  https://iguazio.atlassian.net/browse/ML-10780
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

### DIFF
--- a/mlrun/auth/providers.py
+++ b/mlrun/auth/providers.py
@@ -87,25 +87,19 @@ class DynamicTokenProvider(TokenProvider):
         self._refresh_token_if_needed()
         return self._token
 
-    def fetch_token_with_retries(self, max_retries=None):
-        if not max_retries:
-            max_retries = self._max_retries
-        for attempt in range(max_retries):
-            try:
-                self.fetch_token()
-                # Successfully fetched the token, exit the method
-                return
-            except Exception as exc:
-                logger.warning(
-                    "Token fetch attempt failed",
-                    attempt=attempt + 1,
-                    max_retries=max_retries,
-                    error=str(exc),
-                )
-                if attempt == max_retries - 1:
-                    logger.warning("Max retries reached, failed to fetch token")
-
     def fetch_token(self):
+        try:
+            mlrun.utils.helpers.run_with_retry(
+                retry_count=self._max_retries,
+                func=self._fetch_token,
+            )
+        except Exception as exc:
+            logger.warning(
+                "Max retries reached, failed to fetch token",
+                error=str(exc),
+            )
+
+    def _fetch_token(self):
         """
         Fetch a new access token from the token endpoint.
 
@@ -162,7 +156,7 @@ class DynamicTokenProvider(TokenProvider):
         if self._token and self._is_token_valid(cleanup_if_expired=True):
             return self._token
 
-        self.fetch_token_with_retries()
+        self.fetch_token()
         self._post_fetch_hook()
         return self._token
 
@@ -378,13 +372,15 @@ class IGTokenProvider(DynamicTokenProvider):
 
     def _post_fetch_hook(self):
         # After fetching a new token, sync the local token file with the backend to ensure the latest token is stored
-        mlrun.secrets.sync_secret_tokens()
-
-        # if we get there and have a token which is not valid, but not empty,
-        # it means that refresh threshold is reached and it will soon expire
+        mlrun.utils.helpers.run_with_retry(
+            retry_count=self._max_retries,
+            func=mlrun.secrets.sync_secret_tokens,
+        )
+        # if we reach this point and the token is non-empty but invalid,
+        # it means the refresh threshold has been reached and the token will expire soon.
         if self._token and self._is_token_valid(cleanup_if_expired=True):
             logger.warning(
-                "Fetching a new token failed and the existing token is still valid, but close to expiry. "
+                "Failed to fetch a new token. Using the existing token, which remains valid but is close to expiring."
             )
 
         if not self._token:
--- a/mlrun/auth/providers.py
+++ b/mlrun/auth/providers.py
@@ -67,7 +67,7 @@ class DynamicTokenProvider(TokenProvider):
         self._token = None
         self._token_endpoint = token_endpoint
         self._timeout = timeout
-        self._max_retries = 1
+        self._max_retries = 0
 
         # Since we're only issuing POST requests, which are actually a disguised GET, then it's ok to allow retries
         # on them.
@@ -299,7 +299,7 @@ class IGTokenProvider(DynamicTokenProvider):
 
     def __init__(self, token_endpoint: str, timeout=5):
         super().__init__(token_endpoint=token_endpoint, timeout=timeout)
-        self._max_retries = 3
+        self._max_retries = 2
 
     def _cleanup(self):
         self._token = None
--- a/mlrun/auth/providers.py
+++ b/mlrun/auth/providers.py
@@ -371,12 +371,6 @@ class IGTokenProvider(DynamicTokenProvider):
         )
 
     def _post_fetch_hook(self):
-        # After fetching a new token, sync the local token file with the backend to ensure the latest token is stored
-        # and all other tokens are up-to-date.
-        mlrun.utils.helpers.run_with_retry(
-            retry_count=self._max_retries,
-            func=mlrun.secrets.sync_secret_tokens,
-        )
         # if we reach this point and the token is non-empty but invalid,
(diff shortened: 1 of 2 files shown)


mlrun/auth/providers.py
```diff
@@ -86,24 +87,34 @@ def get_token(self):
         self._refresh_token_if_needed()
         return self._token
 
-    def fetch_token(self, raise_on_error=False):
+    def fetch_token_with_retries(self, max_retries=None):
```
liranbg: I dont think it makes sense to expose they "how" on the function method. perhaps it would be better to unify both fetch token and always retry.

another thing, this entire functionality exists on `helpers.run_with_retry`

mlrun/auth/providers.py
```diff
@@ -345,17 +366,32 @@ def _parse_response(self, response_data: dict, raise_on_error=False):
...
 
+    def _post_fetch_hook(self):
+        # After fetching a new token, sync the local token file with the backend to ensure the latest token is stored
+        mlrun.secrets.sync_secret_tokens()
+
+        # if we get there and have a token which is not valid, but not empty,
+        # it means that refresh threshold is reached and it will soon expire
+        if self._token and self._is_token_valid(cleanup_if_expired=True):
+            logger.warning(
+                "Fetching a new token failed and the existing token is still valid, but close to expiry. "
```
liranbg:
```suggestion
                "Failed to fetch a new token. Using the existing token, which remains valid but is close to expiring. "
```

mlrun/auth/providers.py
```diff
@@ -345,17 +366,32 @@ def _parse_response(self, response_data: dict, raise_on_error=False):
...
         self._token_total_lifetime, self._token_expiry_time = (
             self._get_token_lifetime_and_expiry(access_token)
         )
 
+    def _post_fetch_hook(self):
+        # After fetching a new token, sync the local token file with the backend to ensure the latest token is stored
+        mlrun.secrets.sync_secret_tokens()
+
+        # if we get there and have a token which is not valid, but not empty,
+        # it means that refresh threshold is reached and it will soon expire
```
liranbg:
```suggestion
        # if we reach this point and the token is non-empty but invalid,
       # it means the refresh threshold has been reached and the token will expire soon.
```

mlrun/auth/providers.py
```diff
@@ -345,17 +366,32 @@ def _parse_response(self, response_data: dict, raise_on_error=False):
...
 
         self._token = access_token
+
         self._token_total_lifetime, self._token_expiry_time = (
             self._get_token_lifetime_and_expiry(access_token)
         )
 
+    def _post_fetch_hook(self):
+        # After fetching a new token, sync the local token file with the backend to ensure the latest token is stored
+        mlrun.secrets.sync_secret_tokens()
```
liranbg: could this one failed on api calls? should it be retried as well?

liranbg (commented):
Few comments and suggestions below

mlrun/auth/providers.py
rokatyy: @liranbg added retry 👍

liranbg (approved):

mlrun/auth/providers.py
```diff
@@ -345,17 +360,34 @@ def _parse_response(self, response_data: dict, raise_on_error=False):
...
 
+    def _post_fetch_hook(self):
+        # After fetching a new token, sync the local token file with the backend to ensure the latest token is stored
+        mlrun.utils.helpers.run_with_retry(
+            retry_count=self._max_retries,
+            func=mlrun.secrets.sync_secret_tokens,
+        )
+        # if we reach this point and the token is non-empty but invalid,
+        # it means the refresh threshold has been reached and the token will expire soon.
+        if self._token and self._is_token_valid(cleanup_if_expired=True):
```
moranbental: The comment says “token is non-empty but invalid,” but the code actually checks that the token is valid. What we really need is to check that the token is non-empty and close to expiring.

mlrun/auth/providers.py
```diff
@@ -345,17 +360,34 @@ def _parse_response(self, response_data: dict, raise_on_error=False):
...
         self._token_total_lifetime, self._token_expiry_time = (
             self._get_token_lifetime_and_expiry(access_token)
         )
 
+    def _post_fetch_hook(self):
+        # After fetching a new token, sync the local token file with the backend to ensure the latest token is stored
+        mlrun.utils.helpers.run_with_retry(
+            retry_count=self._max_retries,
+            func=mlrun.secrets.sync_secret_tokens,
+        )
```
moranbental: sync tokens should be called last. If self._token is None, it will raise an error, meaning you end up syncing something unnecessary.

moranbental (changes requested):
Looks good, added some comments

mlrun/auth/providers.py
rokatyy: @moranbental not sure I agree here. In provider we work with a single `default` token, but this function syncs all of the tokens. So I think what user expects is that all the tokens are synced no matter what's wrong with the default one.

mlrun/auth/providers.py
rokatyy: nice catch! fixed

mlrun/auth/providers.py
moranbental: ok, so it’s worth adding a comment on that here

rokatyy: @moranbental the question is do we want to raise an exception if sync didn't succeed or not?

mlrun/auth/providers.py
```diff
@@ -345,17 +360,34 @@ def _parse_response(self, response_data: dict, raise_on_error=False):
...
+        # if we reach this point and the token is non-empty but invalid,
+        # it means the refresh threshold has been reached and the token will expire soon.
+        if self._token and not self._is_token_valid(cleanup_if_expired=True):
```
moranbental, line 376: I don’t think this is enough, since _is_token_valid will return False when remaining_lifetime < 0. Maybe add a unit test here to cover this case

moranbental (changes requested):

mlrun/auth/providers.py
rokatyy: @moranbental what do you mean that's not enough? it will clean up the token if it's expired. There is already a test for it `test_token_cleanup_when_expired`

moranbental: @rokatyy I mean adding a test where the token exists but is close to expiring, and we still use this token, do you have a test that covers that case?

mlrun/auth/providers.py
moranbental: @rokatyy so I’m not sure we want to raise an exception here, since it could be spammy for the user.
@theSaarco / @quaark correct me if I'm wrong.

mlrun/auth/providers.py
rokatyy: @moranbental `test_post_fetch_hook_warns_near_expiry` already exists.

moranbental (approved):

mlrun/auth/providers.py
rokatyy: @moranbental also not sure, because how else user will know that something is wrong with the tokens sync? is there supposed to be any other flow where it will be synced?

moranbental: user can call `mlrun.sync_secret_tokens` directly.

quaark: Spoke with @moranbental offline. I think there's been some confusion here. The sync tokens should not happen on every api call. Only when importing MLRun. Or if the user ran the method specifically

rokatyy: Make sense, removing it from Token provider

moranbental (commented):
The only thing missing is calling sync tokens when importing MLRun, meaning in the token provider initialization, after refreshing the token if needed. But, this can be done in a different pr

PR #8634, opened 2025-09-17
[Nuclio] Warn and remove code when using source archive without repo

Yacouby:
### 📝 Description
This PR ensures that when a source archive is provided without with_repo ( `with_repo=False` ), any inline code is removed so Nuclio will correctly use the provided archive. 
A warning is also added to inform users about the behavior, aligning with existing practices for job runtimes.

---

### 🛠️ Changes Made
- Added a warning when both inline code and a source archive are detected without `with_repo`.
- Removed inline code in such cases to ensure the archive source is applied.

---

### ✅ Checklist

---

### 🧪 Testing
- Added a unit test capturing the warning log and verifying the inline code is removed when `with_source_archive` is called.
- Manually verified by running a notebook that builds a Nuclio function with both inline code and a source archive to ensure the archive source is used.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-7678

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


mlrun/runtimes/nuclio/function.py
```diff
@@ -423,6 +423,18 @@ def with_source_archive(
...
         """
         self.spec.build.source = source
+
+        code = (
+            self.spec.build.functionSourceCode if hasattr(self.spec, "build") else None
+        )
+        if code:
+            # Warn and clear any inline code so the archive is actually used
+            logger.warning(
+                "Function already contains inline code. Removing it so the provided "
```
GiladShapira94: Look good, minor suggestion is to keep the same warning we have for jobs 
```
cannot specify both code and source archive, removing it so the provided source archive will be used instead
```

GiladShapira94 (approved):

TomerShor (approved):

mlrun/runtimes/nuclio/function.py
```diff
@@ -423,6 +423,18 @@ def with_source_archive(
                 )
         """
         self.spec.build.source = source
+
+        code = (
+            self.spec.build.functionSourceCode if hasattr(self.spec, "build") else None
+        )
+        if code:
+            # Warn and clear any inline code so the archive is actually used
+            logger.warning(
```
adamdelman, line 432: Is it enough to raise a warning? 
Should we maybe consider raising an error in this case?

adamdelman (commented):
One small question

mlrun/runtimes/nuclio/function.py
Yacouby: this isn’t a fatal error, we automatically resolve the conflict by preferring the source archive. raising `logger.error ` would imply that the process failed, which isn’t the case.

PR #8646, opened 2025-09-21
[Hub] Support import python module from user/default hub

danielperezz:
### 📝 Description
Add import_module SDK to support working with this new hub asset type in mlrun.

---

### 🛠️ Changes Made
- added the `import_module()` and `get_hub_module()` with their  utilities to a new SDK directory ('mlrun/hub')
- modified the `extend_hub_uri_if_needed()` function so it could serve to construct urls of any object from the hub, not just the `function.yaml`
- added the `item_type` parameter to `list_hub_sources` to clarify the item's type in case of filtering on specific item
---

### ✅ Checklist

---

### 🧪 Testing
  
two new integration tests (one for `hub_module`, one for `import_module`)
  
- hub url doesn't exist
- provided `local_path` doesn't exist
- the files were not downloaded before `module()`
---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-9319
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


mlrun/hub/module.py
```diff
@@ -0,0 +1,137 @@
...
+import mlrun.common.types
+import mlrun.utils
+from mlrun.common.schemas.hub import HubSourceType
+from mlrun.run import function_to_module, get_object
+
+from ..model import ModelObj
+from ..utils import extend_hub_uri_if_needed
+
+
+class ModuleType(mlrun.common.types.StrEnum):
```
Eyal-Danieli: I suggest to move it to `mlrun.common.schemas.hub`

mlrun/hub/module.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def module(self):
+        try:
+            return function_to_module(self.filename, self.local_path)
+        except FileNotFoundError:
+            searched_path = self.local_path or "./"
+            mlrun.utils.logger.warning(
+                f"Module file {self.filename} not found in {searched_path}, try calling download_module_files() first"
+            )
+            return None
```
Eyal-Danieli: I don't think we should return None because it's indeed file not found error. Let's raise it (while keeping the important warn you added).

mlrun/hub/module.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def install_requirements(self) -> None:
+        """
+        Install pip-style requirements (e.g., ["pandas>=2.0.0", "requests==2.31.0"]).
+        """
+        for req in self.requirements:
+            print(f"[INFO] Installing {req} ...")
```
Eyal-Danieli: I guess you added it for debug :)
If you meant to log something to the client, use `logger.info` instead
```suggestion
```

mlrun/hub/module.py
```diff
@@ -0,0 +1,137 @@
...
+            try:
+                subprocess.run(
+                    [sys.executable, "-m", "pip", "install", req], check=True, text=True
+                )
+                print(f"[SUCCESS] Installed {req}")
```
Eyal-Danieli: same

mlrun/hub/module.py
```diff
@@ -0,0 +1,137 @@
...
+            except subprocess.CalledProcessError as e:
+                print(f"[ERROR] Failed to install {req} (exit code {e.returncode})")
```
Eyal-Danieli: same (and here you should use `logger.error`)

mlrun/hub/module.py
```diff
@@ -0,0 +1,137 @@
...
+        self.categories: list = categories or []
+        self.requirements: list = requirements or []
+        self.local_path: str = local_path or ""
+        self.filename: str = filename or name + ".py"
+        self.example: str = example or ""
+        self.url: str = url or ""
+
+    def module(self):
+        try:
+            return function_to_module(self.filename, self.local_path)
```
Eyal-Danieli: just a recommendation to use keywords argument, it's making it more clear for someone who is not familiar with these apis. 
```suggestion
            return function_to_module(code=self.filename, workdir=self.local_path)
```

mlrun/hub/module.py
```diff
@@ -0,0 +1,137 @@
...
+            print(f"[INFO] Installing {req} ...")
+            try:
+                subprocess.run(
+                    [sys.executable, "-m", "pip", "install", req], check=True, text=True
+                )
+                print(f"[SUCCESS] Installed {req}")
+            except subprocess.CalledProcessError as e:
+                print(f"[ERROR] Failed to install {req} (exit code {e.returncode})")
+
+    def download_module_files(self, local_path=None, secrets=None):
```
Eyal-Danieli, line 83: please add a short desc, e.g. what is the default local path? what we are going to download? why it's important?

mlrun/hub/module.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def _download_object(self, obj_url, target_name, secrets=None):
+        data = get_object(obj_url, secrets=secrets)
+        target_dir = self.local_path if self.local_path is not None else os.getcwd()
+        target_filepath = os.path.join(target_dir, target_name)
+        with open(target_filepath, "wb") as f:
+            f.write(data)
+
+    @staticmethod
+    def verify_directory(path) -> Path:
```
Eyal-Danieli:
```suggestion
    def verify_directory(path: str) -> Path:
```

mlrun/hub/module.py
```diff
@@ -0,0 +1,137 @@
...
+        if path:
+            path = Path(path)
+            if not path.exists():
+                raise ValueError(f"Path does not exist: {path}")
+            if not path.is_dir():
+                raise ValueError(f"Path is not a directory: {path}")
+        return path
+
+
+def get_hub_module(url="", download_files=True, secrets=None, local_path=None):
```
Eyal-Danieli:
```suggestion
def get_hub_module(url="", download_files=True, secrets=None, local_path=None) -> HubModule:
```

Eyal-Danieli (commented):
really good job!

few comments, mostly about improving clarity rather than the overall logic. Also, please make sure the title includes `[Hub]` :)

Eyal-Danieli (approved):
LGTM

PR #8650, opened 2025-09-21
[Model Monitoring] Add parquet target to write the stats & Adding writer version to mlrun config

davesh0812:
1. Added parquet target for stats.
2. New config that define which writer to deploy

https://iguazio.atlassian.net/browse/ML-11101


royischoss (approved):
LGTM

PR #8652, opened 2025-09-25
[Model Monitoring] Adding async model monitoring writer

davesh0812:
### 📝 Description

This PR introduces support for deploying a new **writer function**.

Key changes include:

* Leveraging **async execution** and **Storey** for improved performance.
* Added a **new graph** structure:

  * **2 TSDB targets** for application results and metrics.
  * **1 Parquet target** for histogram statistics.
  * A new step, **`MlRunAPIRemoteStep`**, implemented as an inherited step to handle alert generation. This step runs asynchronously and sends alerts to the MLRun API.

### 🛠️ Changes Made

* Added support for deploying a new writer function.
* Introduced **configurable writer versioning** via MLRun configuration (users can define which writer version to use).
* Created a **new graph** with multiple targets and alerting via `MlRunAPIRemoteStep`.

---

### ✅ Checklist

---

### 🧪 Testing
Passed with both writer versions
1.  `TestAlerts::test_drift_detection_alert` 
2. `TestMonitoringAppFlow::test_app_flow`
3. `TestNuclioRuntime::test_list_mep_through_api_step`

---

### 🔗 References
[ML-10331](https://iguazio.atlassian.net/browse/ML-10331)

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [X] No

---

### 🔍️ Additional Notes

V3IO Graph : 

image

TDEngine Graph : 

image

[ML-10331]: https://iguazio.atlassian.net/browse/ML-10331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/model_monitoring/writer.py
```diff
@@ -226,3 +230,199 @@ def do(self, event: _RawEvent) -> None:
...
+                event[WriterEvent.END_INFER_TIME]
+            )
+        if kind == WriterEventKind.STATS:
+            result_event[StatsData.STATS] = json.dumps(result_event[StatsData.STATS])
+        return result_event
+
+
+class KindChoice(storey.Choice):
+    def select_outlets(self, event):
+        logger.info("Selecting the outlet for the event", kind=event.get("kind"))
```
gtopper: `event.get("kind")` is repeated 5 times in this method, so I'd extract it
```python
kind = event.get("kind")
```

mlrun/model_monitoring/writer.py
```diff
@@ -226,3 +230,199 @@ def do(self, event: _RawEvent) -> None:
...
+            raise _WriterEventValueError(
+                f"Unknown event kind: {event.get('kind')}, expected one of: {WriterEventKind.list()}"
+            )
+        return outlets
+
+
+class AlertGenerator(storey.MapClass):
+    def __init__(self, project: str):
+        self.project = project
+        super().__init__()
```
gtopper: I'm not sure how this works without kwargs, when the step name needs to be passed in.
```suggestion
    def __init__(self, project: str, **kwargs):
        self.project = project
        super().__init__(**kwargs)
```

tests/model_monitoring/test_writer_graph.py
```diff
@@ -0,0 +1,56 @@
...
+@pytest.mark.parametrize(
+    "tsdb_profile",
+    [
+        DatastoreProfileV3io(name="v3io-tsdb-test"),
+        DatastoreProfileTDEngine(
+            name="tdengine-test", user="root", host="localhost", port=6041
+        ),
+    ],
+)
+def test_plot_writer_graph(tsdb_profile: DatastoreProfile) -> None:
```
gtopper, line 36: Nice.

mlrun/serving/remote.py
```diff
@@ -454,3 +459,68 @@ def _get_data(self, data, headers):
...
+
+
+class MLRunAPIRemoteStep(RemoteStep):
+    """Graph step implementation for calling MLRun API endpoints"""
+
+    def __init__(
+        self, method: str, path: str, fill_placeholders: bool = False, **kwargs
+    ):
+        """
+        Graph step implementation for calling MLRun API endpoints
```
gtopper, line 471: I don't think we need this repeated.

mlrun/serving/remote.py
```diff
@@ -454,3 +459,68 @@ def _get_data(self, data, headers):
         ) and isinstance(data, (str, bytes)):
             data = json.loads(data)
         return data
+
+
+class MLRunAPIRemoteStep(RemoteStep):
+    """Graph step implementation for calling MLRun API endpoints"""
+
+    def __init__(
+        self, method: str, path: str, fill_placeholders: bool = False, **kwargs
```
gtopper: Optional parameters should default to `None` to avoid them appearing when serialized.

mlrun/serving/remote.py
```diff
@@ -454,3 +459,68 @@ def _get_data(self, data, headers):
...
+        :param fill_placeholders: if True, fill placeholders in the path using event fields
+        :param kwargs:  other arguments passed to RemoteStep
+        """
+        super().__init__(url="", method=method, **kwargs)
+        self.rundb = None
+        self.path = path
+        self.fill_placeholders = fill_placeholders
+
+    def _generate_request(self, event, body):
+        method = self.method or event.method or "POST"
```
gtopper, line 486: This is a bit risky. If we get an event from a GET call, `event.method` will be GET, but that could have nothing to do with the mlrun-api call. Now, the user can always change this, but at the very least we should be clear about it in the docstring.

mlrun/serving/remote.py
```diff
@@ -454,3 +459,68 @@ def _get_data(self, data, headers):
...
+class MLRunAPIRemoteStep(RemoteStep):
+    """Graph step implementation for calling MLRun API endpoints"""
+
+    def __init__(
+        self, method: str, path: str, fill_placeholders: bool = False, **kwargs
+    ):
+        """
+        Graph step implementation for calling MLRun API endpoints
+
+        :param method:  HTTP method (GET, POST, ...)
```
gtopper: Should mention default.

tests/model_monitoring/test_writer_graph.py
```diff
@@ -0,0 +1,56 @@
+# Copyright 2024 Iguazio
```
gtopper:
```suggestion
# Copyright 2025 Iguazio
```

mlrun/model_monitoring/db/tsdb/tdengine/writer_graph_steps.py
```diff
@@ -0,0 +1,88 @@
+# Copyright 2024 Iguazio
```
gtopper:
```suggestion
# Copyright 2025 Iguazio
```

mlrun/model_monitoring/writer.py
```diff
@@ -226,3 +230,199 @@ def do(self, event: _RawEvent) -> None:
...
+        else:
+            raise _WriterEventValueError(
+                f"Unknown event kind: {kind}, expected one of: {WriterEventKind.list()}"
+            )
+        missing_keys = [key for key in expected_keys if key not in result_event]
+        if missing_keys:
+            raise _WriterEventValueError(
+                f"The received event misses some keys compared to the expected "
+                f"monitoring application event schema: {missing_keys} for event kind {kind}"
+            )
```
gtopper, line 341: I know v1 will eventually be deleted, but maybe we can still extract this duplicated bit out?

mlrun/model_monitoring/writer.py
```diff
@@ -226,3 +230,199 @@ def do(self, event: _RawEvent) -> None:
...
+        return result_event
+
+
+class KindChoice(storey.Choice):
+    def select_outlets(self, event):
+        logger.info("Selecting the outlet for the event", kind=event.get("kind"))
+        if event.get("kind") == WriterEventKind.METRIC:
+            outlets = ["tsdb_metrics"]
+        elif event.get("kind") == WriterEventKind.RESULT:
+            outlets = ["tsdb_app_results", "AlertGenerator"]
```
gtopper: Seems like we're missing some logic to not generate an alert for every result.

server/py/services/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -1465,6 +1503,59 @@ def _add_feature_stats(
...
+                kind="parquet_stats",
+            )
+        )
+
+        target = ParquetTarget(
+            path=parquet_target,
+            partition_cols=[WriterEvent.ENDPOINT_ID, mm_constants.StatsData.STATS_NAME],
+        )
+        try:
+            df = target.as_df(
```
gtopper: This is actually be a blocking I/O call, but we run it inside the asyncio event loop without an executor.

server/py/services/api/crud/model_monitoring/model_endpoints.py
```diff
@@ -1465,6 +1503,59 @@ def _add_feature_stats(
...
+            )
+        except Exception as exc:
+            logger.warning(
+                "Failed to read stats from parquet, you may need to check the parquet file",
+                project=project,
+                endpoint_id=uid,
+                kind=kind,
+                error=mlrun.errors.err_to_str(exc),
+            )
+            return {}, None
```
gtopper, line 1534: Needs a bit vertical spacing for readability.
```suggestion
            return {}, None
            
```

tests/system/alerts/test_alerts.py
```diff
@@ -212,8 +212,9 @@ def test_drift_detection_alert(self):
                 model_endpoint.metadata.uid, result_name, model_endpoint.metadata.name
             )
         )
-
-        time.sleep(5)
+        # wait for the event to be processed by the writer, changed as part of adding tsdb target with flush
+        # configuration of 30 seconds
+        time.sleep(35)
```
gtopper: Elsewhere in the test, this is dealt with using the config value directly, i.e.:
```suggestion
        time.sleep(5 + mlconf.model_endpoint_monitoring.writer_graph.flush_after_seconds)
```

mlrun/model_monitoring/writer.py
```diff
@@ -226,3 +230,199 @@ def do(self, event: _RawEvent) -> None:
...
 
         logger.info("Model monitoring writer finished handling event")
+
+
+class WriterGraphFactory:
+    def __init__(
+        self,
+        parquet_path: str,
+        parquet_batching_max_events: int = 1000,
+        parquet_batching_timeout_secs: int = 10,
```
gtopper: Maybe the defaults should be the config values? Otherwise, one might be lead to believe that these values (1000 and 10) have some effect.

mlrun/model_monitoring/writer.py
```diff
@@ -226,3 +230,199 @@ def do(self, event: _RawEvent) -> None:
...
+            else self.parquet_path + "/",
+            max_events=self.parquet_batching_max_events,
+            flush_after_seconds=self.parquet_batching_timeout_secs,
+            columns=[
+                StatsData.TIMESTAMP,
+                StatsData.STATS,
+                WriterEvent.ENDPOINT_ID,
+                StatsData.STATS_NAME,
+            ],
+            key_bucketing_number=0,
```
gtopper: This parameter actually doesn't exist in the storey target. It exists in the feature store target, which isn't relevant here.

gtopper (commented):
Great stuff! Some comments.

mlrun/model_monitoring/writer.py
davesh0812: Its not relevant even if we are using S3? 
I copied it from the parquet target of the stream pod...

mlrun/model_monitoring/writer.py
davesh0812: I don't want to mixup between both of them because as you said we will remove the old version soon.

server/py/services/api/crud/model_monitoring/model_endpoints.py
davesh0812: I’ll add `run_in_threadpool` one layer earlier, to avoid issues with target serialization.

mlrun/model_monitoring/writer.py
gtopper: It's not a parameter of storey's `ParquetTarget`, so it's no more relevant than any other nonexistent parameter...

mlrun/model_monitoring/writer.py
```diff
@@ -226,3 +231,219 @@ def do(self, event: _RawEvent) -> None:
...
+        logger.info("Selecting the outlet for the event", kind=kind)
+        if kind == WriterEventKind.METRIC:
+            outlets = ["tsdb_metrics"]
+        elif kind == WriterEventKind.RESULT:
+            outlets = ["tsdb_app_results", "alert_generator"]
+        elif kind == WriterEventKind.STATS:
+            outlets = ["stats_writer"]
+        else:
+            raise _WriterEventValueError(
+                f"Unknown event kind: {event.get('kind')}, expected one of: {WriterEventKind.list()}"
```
gtopper:
```suggestion
                f"Unknown event kind: {kind}, expected one of: {WriterEventKind.list()}"
```

mlrun/serving/remote.py
```diff
@@ -454,3 +459,67 @@ def _get_data(self, data, headers):
...
+
+class MLRunAPIRemoteStep(RemoteStep):
+    def __init__(
+        self, method: str, path: str, fill_placeholders: Optional[bool] = None, **kwargs
+    ):
+        """
+        Graph step implementation for calling MLRun API endpoints
+
+        :param method:  HTTP method (GET, POST, ...) as default defines the method of the request, if empty string or
+                        None the method will be taken from the event (default to POST)
```
gtopper: There are 3 different defaults mentioned here. I don't think a user would be able to make sense of it, so it needs a bit of a rewrite.

mlrun/serving/remote.py
```diff
@@ -454,3 +459,67 @@ def _get_data(self, data, headers):
...
+        :param path:    API path (e.g. /api/projects)
+        :param fill_placeholders: if True, fill placeholders in the path using event fields
```
gtopper: False by default.

mlrun/serving/remote.py
```diff
@@ -454,3 +459,67 @@ def _get_data(self, data, headers):
...
+        :param kwargs:  other arguments passed to RemoteStep
+        """
+        super().__init__(url="", method=method, **kwargs)
+        self.rundb = None
+        self.path = path
+        self.fill_placeholders = fill_placeholders or False
```
gtopper: The default actually shouldn't be set in the constructor, because it will cause `fill_placeholders` to be written as `False` when the class is serialized, rather than being omitted (when the user didn't specify it).

mlrun/model_monitoring/db/tsdb/tdengine/writer_graph_steps.py
```diff
@@ -0,0 +1,88 @@
...
+            event[mm_schemas.EventFieldType.TABLE_COLUMN] = (
+                f"{table_name}_{event[mm_schemas.MetricData.METRIC_NAME]}"
+            ).replace("-", "_")
+            event[mm_schemas.WriterEvent.START_INFER_TIME] = datetime.fromisoformat(
+                event[mm_schemas.WriterEvent.START_INFER_TIME]
+            )
+        return event
+
+
+class ErrorExtractor(mlrun.feature_store.steps.MapClass):
```
gtopper: If it's an exact copy of the TDEngine one, we could probably avoid duplicating it.

mlrun/model_monitoring/db/tsdb/tdengine/writer_graph_steps.py
```diff
@@ -0,0 +1,88 @@
...
+        Prepare the event for insertion into the TDEngine error table
+        """
+        super().__init__(**kwargs)
+
+    def do(self, event):
+        error = str(event.get("error"))
+        if len(error) > 1000:
+            error = error[-1000:]
+            logger.warning(
+                f"Error message exceeds 1000 chars: The error message writen to TSDB will be it last "
```
gtopper: I know this is copied from TDEngine, but `1000` should be extracted to a constant or configuration.

mlrun/model_monitoring/writer.py
```diff
@@ -226,3 +231,219 @@ def do(self, event: _RawEvent) -> None:
...
+            "mlrun.serving.remote.MLRunAPIRemoteStep",
+            name="alert_generator_api_call",
+            after="filter_none",
+            method="POST",
+            path=f"projects/{fn.metadata.project}/events/{{kind}}",
+            fill_placeholders=True,
+        )
+
+        graph.add_step(
+            "mlrun.datastore.storeytargets.ParquetStoreyTarget",
```
gtopper, line 290: Shouldn't these steps be added inside `apply_writer_steps()`?

gtopper (commented):
Some follow ups. Please look over all unresolved threads and resolve conflicts with development.

mlrun/model_monitoring/db/tsdb/tdengine/writer_graph_steps.py
davesh0812: I deleted this class so its not relevant to this PR

mlrun/model_monitoring/writer.py
davesh0812: Its part of the graph creation so I put every think under `apply_writer_graph`.. not sure what is the problem?

mlrun/model_monitoring/db/tsdb/tdengine/writer_graph_steps.py
```diff
@@ -0,0 +1,54 @@
...
+            f"{event[mm_schemas.WriterEvent.APPLICATION_NAME]}"
+        )
+        if kind == mm_schemas.WriterEventKind.RESULT:
+            # Write a new result
+            event[mm_schemas.EventFieldType.TABLE_COLUMN] = (
+                f"{table_name}_{event[mm_schemas.ResultData.RESULT_NAME]}"
+            ).replace("-", "_")
+            event[mm_schemas.WriterEvent.START_INFER_TIME] = datetime.fromisoformat(
+                event[mm_schemas.WriterEvent.START_INFER_TIME]
+            )
```
Eyal-Danieli: you can move this part outside the if clause (assuming that the kind is either result or metric). The only difference is the table column.

gtopper (approved):
LGTM

mlrun/model_monitoring/writer.py
```diff
@@ -226,3 +231,218 @@ def do(self, event: _RawEvent) -> None:
...
+        )
+
+    def _generate_event_data(
+        self,
+        entity_id: str,
+        result_status: int,
+        event_value: dict,
+        project_name: str,
+        result_kind: int,
+    ) -> Event:
```
Eyal-Danieli: remove `from mlrun.common.schemas import Event` and set:

```suggestion
    ) -> mlrun.common.schemas.Event:
```

davesh0812: lama?

Eyal-Danieli (approved):
Nice!

assaf758 (approved):

PR #8656, opened 2025-09-25
[Workflow] Fix scheduled workflows to use custom source in workflow request

Yacouby:
### 📝 Description
This PR fixes an issue where scheduled workflows always used `project.spec.source`, even when a custom source was provided in the workflow request.

---

### 🛠️ Changes Made

- Removed the explicit `url=project.spec.source` argument in the schedule method so now `_prepare_run_object` will consistently set the workflow URL based on `_validate_source`, ensuring that a custom source from the workflow request is used when provided, otherwise falling back to the project source.
- Removed the now unused url parameter from `_prepare_run_object` and related calls, as the source resolution is fully handled by `_validate_source` and no need for the url param now.

---

### ✅ Checklist

---

### 🧪 Testing
Added a new unit test `test_schedule_workflow_with_custom_source` that ensures:
When a custom source is provided in the workflow request, it is used in the scheduled workflow.
The scheduled workflow no longer defaults to `project.spec.source` when an override is given.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10293
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/server/py/services/api/tests/unit/crud/test_workflows.py
+++ b/server/py/services/api/tests/unit/crud/test_workflows.py
@@ -211,3 +211,55 @@ class TestWorkflows(services.api.tests.unit.conftest.MockedK8sHelper):
             )
         run = runner_class().run(**params)
         assert run.spec.parameters["save"] == expected_save
+
+    def test_schedule_workflow_with_custom_source(
+        self,
+        db: sqlalchemy.orm.Session,
+        k8s_secrets_mock,
+    ):
+        # This test verifies that if a custom source is provided in the workflow request,
+        # it is used for the scheduled workflow instead of the project's source.
+
+        project = mlrun.common.schemas.ProjectOut(
+            metadata=mlrun.common.schemas.ProjectMetadata(name="project-name"),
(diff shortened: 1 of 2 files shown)


server/py/services/api/tests/unit/crud/test_workflows.py
```diff
@@ -211,3 +211,55 @@ def test_run_workflow_save_project(
...
+                        engine="remote",
+                        image="mlrun/mlrun",
+                    ),
+                    source=custom_source,
+                ),
+                auth_info=mlrun.common.schemas.AuthInfo(),
+            )
+
+            # Extract the stored scheduled task
+            scheduled_task = scheduler_mock.return_value.store_schedule.call_args[1][
```
adamdelman: Make using `assert_called_once_with` could be a bit nicer?

adamdelman (changes requested):
minor issue

TomerShor (approved):

yaelgen (approved):
🚀

PR #8657, opened 2025-09-25
[Projects] Fix race condition during project logs deletion

rokatyy:
### 📝 Description
When deleting project logs, the primary flow calls the log collector service.
If that fails (e.g. connectivity or gRPC errors), we fall back to `delete_project_logs_legacy`, which removes the logs directory with `shutil.rmtree`.

On NFS-backed storage, transient `.nfs*` files are created when a file is deleted but still open by some process. And. then they are automatically removed once the file handle is released. This creates a race condition:

1. `logs_path.exists()` (in https://github.com/mlrun/mlrun/blob/57ce59eea2f798176466d9ac1874ae77a62582e4/server/py/services/api/crud/logs.py#L415) succeeds.
2. `shutil.rmtree` starts walking the tree.
3. `.nfs*` file disappears mid-deletion.
4. `FileNotFoundError is raised`, breaking cleanup.

---

### 🛠️ Changes Made
Wrapped `shutil.rmtree` with an onerror handler that:

* Ignores only `FileNotFoundError` (`ENOENT`) 
* Surfaces all other errors (e.g. permissions and etc).

---

### ✅ Checklist

---

### 🧪 Testing
Not tested, very hard to reproduce, bug was identified using logs and code reading
---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10980
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

### DIFF
--- a/server/py/services/api/crud/logs.py
+++ b/server/py/services/api/crud/logs.py
@@ -16,6 +16,7 @@ import os
 import pathlib
 import shutil
 import typing
+import errno
 from http import HTTPStatus
 
 from fastapi.concurrency import run_in_threadpool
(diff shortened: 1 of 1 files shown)


TomerShor (approved):

server/py/services/api/crud/logs.py
```diff
@@ -77,12 +78,20 @@ async def delete_runs_logs(self, project: str, run_uids: list[str]):
         await self._delete_logs(project, run_uids)
 
     @staticmethod
-    def delete_project_logs_legacy(
-        project: str,
-    ):
+    def delete_project_logs_legacy(project: str):
+        def _ignore_missing_files(func, path, exc_info):
+            # covers a race condition where some of the files were deleted by log-collector
```
Yacouby:
```suggestion
            # handles race conditions where files disappear during deletion (e.g., removed by log-collector or NFS cleanup)
```

Yacouby (changes requested):
minor comment improvement

Yacouby (approved):

PR #8660, opened 2025-09-28
[Labels] Change 'owner' label in workflow-runner pod to the actual username of the triggering user

elbamit:
### 📝 Description 
In remote/schedule workflow, change the 'owner' label of the workflow-runner pod from 'pipelines' to the actual username of the running user.

**Cause of the issue:**
'pipelines' originated from using V3IO_USERNAME env var from the mlrun-api when resolving the owner label.

**Cases to handle**
1. A run/workflow is executed directly by some user (i.e. `project.run()`) - in this case the owner of the workflow is the user that executed it. The person who ran it is the one owning the workflow.
2. A schedule/retry is triggered - in this case the owner of the workflow is the user who created the schedule/original workflow since it contains the details on what to run as well as how to run it - parameters to use, when to run it etc.. The user (or system for scheduled WF) who triggers it is just an executor

**How is it handled?**
1. Uses the auth info to extract the correct username of the user and set it as the owner label.
2. During the retry process, the original workflow is extracted. Use his 'owner' label to set the rerun-workflow-runner 'owner' label accordingly.

---

### 🛠️ Changes Made

- Refactor `enrich_run_labels()` function to get an owner value to enrich
- Add a `resolve_owner()` function
- Refactor the `_enrich_run_labels_and_env()` function to accept `auth_username` and `original_runner_owner` params and enrich accordingly.
- Propagate the `auth_info.username` and `original_runner_owner` to label enrichment function

---

### ✅ Checklist

---

### 🧪 Testing
Manual + Unit testing + Added case in an existing system test

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10622 
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


mlrun/runtimes/utils.py
```diff
@@ -460,8 +461,11 @@ def enrich_run_labels(
     """
     # Merge the labels with the labels enrichment extension
     labels_enrichment = {
-        mlrun_constants.MLRunInternalLabels.owner: os.environ.get("V3IO_USERNAME")
-        or getpass.getuser(),
+        mlrun_constants.MLRunInternalLabels.owner: (
+            auth_username
+            if labels.get("job-type") == "workflow-runner"
```
liranbg: use constants.JOB_TYPE_WORKFLOW_RUNNER

also, note that, for CE (non enterprise deployments) that owner will be what? as opposed to "getpass.getuser()" before.

I think worth moving this resolvement to a function that will check if auth_username is empty. function would be self explanatory and can be unit-tested easily.

liranbg (changes requested):

server/py/services/api/crud/workflows.py
```diff
@@ -105,7 +105,7 @@ def prepare_and_run(
         :param rerun_request:    Workflow request containing the rerun spec.
         :return: RunObject with run metadata, results, and status.
         """
-        self._enrich_run_labels_and_env(labels, runner)
+        self._enrich_run_labels_and_env(labels, runner, auth_info.username)
```
moranbental: Will the user who triggers this endpoint always be the owner of the workflow? Or could it also be an admin or someone else?

moranbental (approved):
Great job!

mlrun/runtimes/utils.py
elbamit: Changed.
If there's no auth info then it will fallback to the default values of V3IO_ACCESS_KEY or getpass.getuser()

yaelgen (approved):
Looks Great 🚀

liranbg (approved):

PR #8662, opened 2025-09-29
[KFP] Format KFP runs concurrently instead of one by one

adamdelman:
### 📝 Description
Introduced concurrent formatting for runs and aligned tests/mocks with the new flow.

---

### 🛠️ Changes Made
- **Concurrent formatting pipeline**
  - Added `Pipelines._format_runs_concurrently(...)` using `ThreadPoolExecutor` (with optional bounded `Semaphore`) to format runs in parallel while preserving discovery order and skipping per-run failures with a warning.
  - `list_pipelines(...)` now invokes the concurrent formatter for each streamed page.
---

### ✅ Checklist

---

### 🧪 Testing
- Verified parallel formatting behavior returns correct data.
- Confirmed error enrichment happens only for unsuccessful statuses.
- Asserted filter behavior when `experiment_ids` are present vs. absent.
- Updated/ran unit tests that exercise the new concurrent formatter and low-level KFP list runs mocks.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11228
---

### 🚨 Breaking Changes?
- [X] No


server/py/services/api/crud/pipelines.py
```diff
@@ -692,21 +693,50 @@ def _format_run(
...
+                if semaphore:
+                    future.add_done_callback(lambda _f: semaphore.release())
+                futures_by_index[run_index] = future
+
+            formatted_runs = []
+            for future in futures_by_index:
+                try:
+                    formatted_runs.append(future.result())
+                except Exception:
+                    mlrun.utils.logger.warning(
```
Yacouby: shouldn't this be error and not warning ?

server/py/services/api/crud/pipelines.py
```diff
@@ -692,21 +693,50 @@ def _format_run(
...
+        queue_size: typing.Optional[int] = None,
     ) -> list[dict]:
-        formatted_runs = []
-        for run in runs:
-            formatted_runs.append(
-                self._format_run(
-                    run=run,
+        """
+        Submit formatting tasks concurrently and emit results in discovery order.
+        """
```
Yacouby, line 725: worth to add a check at the beginning if runs is empty then return

server/py/services/api/crud/pipelines.py
```diff
@@ -692,21 +693,50 @@ def _format_run(
...
+        max_workers: int = 32,
+        queue_size: typing.Optional[int] = None,
     ) -> list[dict]:
-        formatted_runs = []
-        for run in runs:
-            formatted_runs.append(
-                self._format_run(
-                    run=run,
+        """
+        Submit formatting tasks concurrently and emit results in discovery order.
```
Yacouby, line 706: add more doc explaining the use of the semaphore here and the max_workers

server/py/services/api/crud/pipelines.py
```diff
@@ -692,21 +693,50 @@ def _format_run(
...
-    def _format_runs(
+    def _format_runs_concurrently(
         self,
-        runs: Iterable[PipelineRun],
+        kfp_client: mlrun_pipelines.client.Client,
+        runs: list[mlrun_pipelines.models.PipelineRun],
         format_: mlrun.common.formatters.PipelineFormat = mlrun.common.formatters.PipelineFormat.metadata_only,
-        kfp_client: mlrun_pipelines.client.Client = None,
+        *,
+        max_workers: int = 32,
```
Yacouby, line 702: can this be too high on small pods ? can be reduced or derived from available CPUs ?

Yacouby (changes requested):
some suggestions

server/py/services/api/crud/pipelines.py
adamdelman: since this mostly is used to solve io problems, and the cpu usage here is pretty low, I don't think it makes sense to bound it on available cpus

Yacouby (approved):

liranbg (approved):

PR #8667, opened 2025-09-29
[User Secrets] Revoke token by sending the request header [feature/ig4-authentication]

moranbental:
### 📝 Description

When attempting to call `revoke_offline_token` in Iguazio, it fails with the error:
`No authentication is available, login required when trying to revoke token.`
This happens because the request requires passing the access token from the request headers to Iguazio.

---

### 🛠️ Changes Made

1. Call `set_override_auth_headers` before revoking the token in Iguazio using the request headers.
2. Bump the Iguazio package to 0.0.1a12.

---

### ✅ Checklist

---

### 🧪 Testing
  
  
Tested on my ig4 system 

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11075
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


quaark (changes requested):
Looks great! Just add a UT for the case where we do provide the request headers

quaark (approved):

PR #8670, opened 2025-09-30
[Serving] Add `ChoiceByField` Step

daniels290813:
### 📝 Description
The default `Choice` step by default sends the event to all the outlets (children) of the step. Any other logic requires subclassing and adding a custom `select_outlets` flow. 

A more user-friendly approach would be for the `Choice` step to receive a name of field from which to extract the route path. This way the user would create a custom logic step that adds a field into the event with the route name, and then just use the default choice step logic with that field name.

---

### 🛠️ Changes Made
Adding ChoiceByField class that subclass `storey.Choice`.

---

### ✅ Checklist

---

### 🧪 Testing
- Unitest
---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10097
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [] Yes (explain below)
- [X] No

---


mlrun/serving/steps.py
```diff
@@ -0,0 +1,44 @@
+# Copyright 2023 Iguazio
```
gtopper:
```suggestion
# Copyright 2025 Iguazio
```

docs/serving/available-steps.md
```diff
@@ -63,4 +63,5 @@ See also [Data transformations](../feature-store/transformations.md#data-transfo
 | Class name                                                 | Description                                                                                                   |   
 |------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------|
 | {py:class}`~mlrun.feature_store.steps.FeaturesetValidator` | Validate feature values according to the feature set validation policy. Supported also by the Pandas engines. | 
+| {py:class}`~mlrun.serving.steps.ChoiceByField`             | Adding logic to `Choice`. receive a name of field from which to extract the route path.                       |
```
gtopper:
```suggestion
| {py:class}`~mlrun.serving.steps.ChoiceByField`             | Routes events to downstream steps based on an event field that contains the step name or names. | 
```

mlrun/serving/steps.py
```diff
@@ -0,0 +1,44 @@
...
+        outlets = (
+            [event[self.field_name]]
+            if isinstance(event[self.field_name], str)
+            else event[self.field_name]
+        )
+        if not outlets:
+            raise mlrun.MLRunNotFoundError(
+                f"Steps not found for given field name {self.field_name}."
+            )
+        return outlets
```
gtopper, line 62:
```suggestion
from typing import Union

import storey

import mlrun.errors


class ChoiceByField(storey.Choice):
    """
    Selects downstream outlets to route each event based on a predetermined field.
    :param field_name: event field name that contains the step name or names of the desired outlet or outlets
    """

    def __init__(self, field_name: Union[str, list[str]], **kwargs):
        self.field_name = field_name
        super().__init__(**kwargs)

    def select_outlets(self, event):
        outlet = event.get(self.field_name)
        if outlet is None:
            raise mlrun.errors.MLRunRuntimeError(
                f"Field name '{self.field_name}' is not contained in the event keys {list(event.keys())}"
            )
        if not isinstance(outlet, (str, list, tuple)):
            raise mlrun.errors.MLRunInvalidArgumentTypeError(
                f"Field '{self.field_name}' must be a string or list of strings, but is instead of type '{type(outlet).__name__}'"
            )
        outlets = [outlet] if isinstance(outlet, str) else outlet
        return outlets
```

gtopper: Note that this makes `[]` a legal result. If you want to block it, please raise a specific error to that effect.

tests/serving/test_serving.py
```diff
@@ -910,3 +911,51 @@ def test_serialize():
...
+    try:
+        result = fn_server.test(
+            body={"field_name": field_name, "field_targets": field_targets}
+        )
+        if isinstance(field_targets, list):
+            assert result["field_visited"] in field_targets
+        else:
+            assert result["field_visited"] == field_targets
+    except Exception as e:
+        assert isinstance(e, (RuntimeError, mlrun.MLRunNotFoundError))
```
gtopper: Use `with pytest.raises(...):`

tests/serving/test_serving.py
```diff
@@ -910,3 +911,51 @@ def test_serialize():
...
+    ).respond()
+    fn_server = fn.to_mock_server()
+    try:
+        result = fn_server.test(
+            body={"field_name": field_name, "field_targets": field_targets}
+        )
+        if isinstance(field_targets, list):
+            assert result["field_visited"] in field_targets
+        else:
+            assert result["field_visited"] == field_targets
```
gtopper: Should probably make this check more strict.

tests/serving/test_serving.py
```diff
@@ -910,3 +911,51 @@ def test_serialize():
...
+
+@pytest.mark.parametrize("field_name", ["fieldA", "fieldB"])
+@pytest.mark.parametrize(
+    "field_targets",
+    ["outlet1", "outlet2", ["outlet1", "outlet2"], [], ["outlet3"], "outlet3"],
+)
+def test_choice_by_field(field_name, field_targets):
+    choice_step = ChoiceByField(field_name=field_name)
+    fn = mlrun.new_function(
+        name="choice-by-field-example", kind="serving", image="mlrun/mlrun"
```
gtopper:
```suggestion
        name="choice-by-field-example", kind="serving",
```

mlrun/serving/steps.py
daniels290813: When testing with [] (no outlets returned from select outlets) - the test stuck waiting for response.
I'll raise a specific error

mlrun/serving/steps.py
```diff
@@ -0,0 +1,48 @@
...
+            )
+        if not isinstance(outlet, (str, list, tuple)):
+            raise mlrun.errors.MLRunInvalidArgumentTypeError(
+                f"Field '{self.field_name}' must be a string or list of strings"
+                f"but is instead of type '{type(outlet).__name__}'"
+            )
+        outlets = [outlet] if isinstance(outlet, str) else outlet
+        if not outlets:
+            raise mlrun.errors.MLRunRuntimeError(
+                f"Predetermined field value '{self.field_name}' cannot be empty"
```
gtopper: Predetermined? We get this when the field contains an empty list, don't we? Need a better message to clarify that.

mlrun/serving/steps.py
gtopper: Right, if the user throws away the event, the response will never happen.

tests/serving/test_serving.py
```diff
@@ -910,3 +911,70 @@ def test_serialize():
...
+    try:
+        with pytest.raises(
+            (RuntimeError, mlrun.MLRunNotFoundError, ValueError)
+        ) as exc_info:
+            result = fn_server.test(
+                body={"field_name": field_name, "field_targets": field_targets}
+            )
+            assert isinstance(
+                exc_info.value, (RuntimeError, mlrun.MLRunNotFoundError, ValueError)
+            )
```
gtopper:
```suggestion
```

tests/serving/test_serving.py
```diff
@@ -910,3 +911,70 @@ def test_serialize():
...
+        _fn='({"field_visited": "outlet2"})',
+        after="pick_outlets",
+    )
+    graph.add_step(
+        "storey.Extend", name="end", _fn="({})", after=["outlet1", "outlet2"]
+    ).respond()
+    fn_server = fn.to_mock_server()
+    try:
+        with pytest.raises(
+            (RuntimeError, mlrun.MLRunNotFoundError, ValueError)
```
gtopper: Pretty sure there is only one type of exception we should be expecting here. Also, need to add match="..." parameter for the error message.

tests/serving/test_serving.py
```diff
@@ -910,3 +911,70 @@ def test_serialize():
...
+        ) as exc_info:
+            result = fn_server.test(
+                body={"field_name": field_name, "field_targets": field_targets}
+            )
+            assert isinstance(
+                exc_info.value, (RuntimeError, mlrun.MLRunNotFoundError, ValueError)
+            )
+    except AssertionError:
+        # means no exception was raised → handle as success case
```
gtopper: We should know whether we expect an error or not. Please either use the parameters for this or break it up into two tests.

gtopper (commented):
Cool, a couple more.

mlrun/serving/steps.py
daniels290813: This should catch cases like empty lists or string.
added this edge cases to the test.
I will update the error for something like
"The value of the given field name '{field_name}' in the event cannot be an empty string/list"
wdyt ?

gtopper: It won't catch an empty string though.

daniels290813:
@gtopper I've decided to move it to a new file I hope thats cool
The case of an empty string field name is actually valid here, wdyt ?

tests/serving/test_serving.py
```diff
@@ -910,3 +911,70 @@ def test_serialize():
...
+        if isinstance(field_targets, (list, tuple)):
+            if len(field_targets) < 2:
+                assert sorted(result["field_visited"]) == sorted(field_targets)
+            else:
+                assert result["field_visited"] in field_targets
+        else:
+            assert result["field_visited"] == field_targets
```
gtopper, line 980: The test doesn't terminate because the mock server isn't terminated. Missing a call to `fn_server.wait_for_completion()` at the end (in a finally block, or a fixture.)

tests/serving/test_serving.py
```diff
@@ -910,3 +911,70 @@ def test_serialize():
...
+            assert isinstance(
+                exc_info.value, (RuntimeError, mlrun.MLRunNotFoundError, ValueError)
+            )
+    except AssertionError:
+        # means no exception was raised → handle as success case
+        if isinstance(field_targets, (list, tuple)):
+            if len(field_targets) < 2:
+                assert sorted(result["field_visited"]) == sorted(field_targets)
+            else:
+                assert result["field_visited"] in field_targets
```
gtopper, line 978: Need to be more specific, or it could pass even with the wrong result.

gtopper (approved):
LGTM ✅

PR #8463, opened 2025-08-13
[KFP] Refactor KFP image naming and upgrade dependencies for Python 3.11

adamdelman:
### 🎯 Description
Unify Python/KFP handling across builds. Default `mlrun-kfp` to Py 3.11, add explicit Py 3.9 variants, align dependency pins (Dask, grpcio, scikit-learn, kfp), and use per-Python constraints + versioned lockfiles. Always append Python suffix in image tags.

---

### 🛠 Changes Made
- **Image tag enrichment**
  - Always add suffix from `resolve_image_tag_suffix`; remove `mlrun-kfp` exception.
  - Tests: add `mlrun/mlrun-kfp` cases (`3.9 → -py39`, `3.11 → no suffix`).
  - Added Python 3.11 to the matrix for the `kfp-v1-8` pipeline adapter. 
- **Dockerfiles**
  - `mlrun-kfp`: default `MLRUN_PYTHON_VERSION=3.11`; use `locked-requirements_${MLRUN_PYTHON_VERSION}.txt`; pass `--python-version` to `uv`.
  - `test`: use `locked-requirements_${MLRUN_PYTHON_VERSION}.txt`.

- **Makefile**
  - Remove `MLRUN_PIPELINES_KFP_VERSION`.
  - `mlrun-kfp` target uses `common-image`, passes `MLRUN_PYTHON_VERSION` and `--constraint dockerfiles/constraints-py$(MLRUN_PYTHON_VERSION).txt`.
  - Add `upgrade-mlrun-kfp-deps-lock-all` / `upgrade-mlrun-test-deps-lock-all` + per-Python subtasks (3.11/3.9).
  - All `uv pip compile` include `--python-version`; test lock compile uses per-Python constraint; add ignore of TF-Keras test on 3.11 unit runs.

- **Patch tooling**
  - `automation/patch_igz/patch_remote.py`: define `python_39_suffix="-py39"` and `python_39_targets=[mlrun, mlrun_kfp]`; loop build py39 images and map names to images.

- **Constraints & CI**
  - Add `dockerfiles/constraints-py3.11.txt` (`kfp>=1.8.23`, `urllib3>=2.5.0`, `v3io-frames==0.13.10`).
  - Add `dockerfiles/constraints-py3.9.txt` (`kfp==1.8.22`, `urllib3==1.26.20`, `v3io-frames==0.10.15`).
  - Workflow matrix: build `mlrun-kfp` for 3.11 (default) and 3.9 (suffix variant).

[...]

PR #8676, opened 2025-09-30
[Dask] Pin dask verions to prevent client-scheduler verions mismatch

Yacouby:
### 📝 Description
This PR fixes a version mismatch issue between the Dask client (e.g., Jupyter) and the Dask scheduler/worker.

In a recently added PR ( https://github.com/mlrun/mlrun/pull/8463 ), the dependency specifiers for dask in `extras-requirements.txt` used >=, which allowed the client to install newer/latest Dask versions. 

This caused incompatibilities between the client and the scheduler/worker, leading to runtime failures since both the client and the scheduler/worker must always use the same versions. ( reference: https://blog.dask.org/2023/04/14/scheduler-environment-requirements )

---

### 🛠️ Changes Made
Updated `extras-requirements.txt` to replace >= with == or ~= for Dask/Distributed packages.

---

### ✅ Checklist
- x ] I have tested the changes in this PR

---

### 🧪 Testing
Validated that the Dask version in the Jupyter requirements is now aligned with the scheduler/worker by running make, ensuring consistent versions.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11080
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


tests/runtimes/test-dask-local/0/chart
```diff
@@ -0,0 +1 @@
+abc
```
rokatyy: I think this should be removed, same with pkl file and mlrun .env

mlrun.env
```diff
@@ -0,0 +1,4 @@
+MLRUN_DBPATH=https://mlrun-api.default-tenant.app.vmdev26.lab.iguazeng.com
+V3IO_USERNAME=iguazio
+V3IO_ACCESS_KEY=d498c232-e818-471c-911f-1b4b33c5cadf
+MLRUN_HTTPDB__HTTP__VERIFY=false
```
yaelgen, line 4: Is this meant to be here?

yaelgen: This and the hack file meant to be here?

Yacouby: by mistake, thanks :)

tests/runtimes/test-dask-local/0/chart
Yacouby: Thanks, were added by mistake :)

dependencies.py
```diff
@@ -67,10 +67,10 @@ def extra_requirements() -> dict[str, list[str]]:
         "dask": [
             # dask 2023 does not work on python 3.11
             # dask 2024 requires dependencies that current mlrun with 3.9 cannot support
-            'dask>=2023.12.1; python_version < "3.11"',
-            'dask>=2024.8; python_version >= "3.11"',
-            'distributed>=2023.12.1; python_version < "3.11"',
-            'distributed>=2024.8; python_version >= "3.11"',
+            'dask~=2023.12.1; python_version < "3.11"',
```
rokatyy, line 74: Add the explanation from the pr description to a comment here as well for a future self

rokatyy (approved):
Minor thingy, otherwise approved

mlrun.env
rokatyy: remove this v3io key from the system

dependencies.py
Yacouby: Done 👍

liranbg (approved):

TomerShor (approved):

PR #8685, opened 2025-10-01
[Function] Do not allow running of functions that use secrets from the wrong project

rokatyy:
### 📝 Description
Do not allow running of functions that use secrets from the wrong project, 
Prevent functions from using secrets that belong to a different project.
This applies to the following cases:

* Volumes mounting secrets from another project
* Environment variables referencing secrets from another project
* Tasks including secrets from another project in their secrets list

---

### ✅ Checklist

---

### 🧪 Testing
UT + vmdev + system test

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11219, https://iguazio.atlassian.net/browse/ML-11214, https://iguazio.atlassian.net/browse/ML-11220, https://iguazio.atlassian.net/browse/ML-11222
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [x] Yes (if users were relying on this bug)
- [ ] No

---

### 🔍️ Additional Notes
image

### DIFF
--- a/server/py/framework/api/utils.py
+++ b/server/py/framework/api/utils.py
@@ -313,12 +313,6 @@ def validate_function_secret_sources(function):
     secrets = mlrun.secrets.SecretsStore.from_list(secrets_list)
 
     project_name = get_in(function, ["metadata", "project"])
-    if k8s_secrets := secrets.get_k8s_secrets():
-        for secret_name in k8s_secrets.keys():
-            validate_secret_allowed(
-                project_name=project_name,
-                secret_name=secret_name,
-            )
 
     if azure_k8s_secrets := secrets.get_azure_vault_k8s_secret():
         validate_secret_allowed(
--- a/server/py/framework/api/utils.py
+++ b/server/py/framework/api/utils.py
@@ -336,6 +336,7 @@ def apply_enrichment_and_validation_on_function(
     validate_service_account: bool = True,
     mask_sensitive_data: bool = True,
     ensure_security_context: bool = True,
+        allow_empty_access_key: bool = False,
 ):
     """
     This function should be used only on server side.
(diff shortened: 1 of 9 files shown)


tests/system/projects/test_project.py
```diff
@@ -1967,6 +1963,78 @@ def test_notifications_on_workflow_start_and_completion(self):
             len(notifications) == 2
         ), f"Expected 2 notifications, got {len(notifications)}"
 
+    @pytest.mark.parametrize("kind", ["nuclio", "job", "mpijob"])
+    def test_deploy_function_with_mounted_project_secret(self, kind):
```
liranbg, line 1967: I wonder, how "slow" this test is? covering all parameterized. if it fails fast, then should be good, right?

server/py/framework/api/utils.py
```diff
@@ -288,11 +294,37 @@ async def submit_run(
...
+    secrets_list = get_in(function, ["spec", RunKeys.secrets], default=[])
+    secrets = mlrun.secrets.SecretsStore.from_list(secrets_list)
+
+    project_name = get_in(function, ["metadata", "project"])
+    if k8s_secrets := secrets.get_k8s_secrets():
+        for secret_name in k8s_secrets.keys():
+            validate_secret_allowed(
+                project_name=project_name,
+                secret_name=secret_name,
+            )
```
theSaarco: The `kubernetes` secret source type does not allow you to specify the secret name - it only lets you specify a list of keys from this project's secrets. Quoting the docstring from `with_secrets`:

> If using with k8s secrets, the k8s secret is managed by MLRun, through the project-secrets
> mechanism. The secrets will be attached to the running pod as environment variables.
> task.with_secrets('kubernetes', ['secret1', 'secret2'])

The list of keys you iterate over are not secret names but rather keys within that secret. Therefore this check is not needed (even if a user puts a key that has the name of a secret in it, it will not mount that secret).

liranbg (commented):
nice and neat! small question regarding system tests

theSaarco (changes requested):
Looks very good, with a single comment that you perform extra validation where not needed (on project-secret secret sources).

tests/system/projects/test_project.py
rokatyy: @liranbg yes, it's fast, cause it doesn't really deploy function and fails almost immediately after api gets the request

server/py/framework/api/utils.py
rokatyy: @theSaarco I see, naming confused me and made me think that those are secrets, not keys. Will remove this validation of k8s type entirely then

theSaarco (approved):
Looks good! 🚀

PR #8693, opened 2025-10-05
[Model Monitoring] Update infra labels during the data migration process

Eyal-Danieli:
### 📝 Description
As part of exposing the new functions summaries API (#7959), we also added a label to help identify the model monitoring infra functions (stream, controller, and writer). To keep things backward compatible, this PR also updates existing infra functions from older versions (<1.10) by adding the new labels during the migration process.

---

### 🛠️ Changes Made
- add model monitoring infra label to the infra funcs within `functions` table 
- add a corresponding label to the `functions_labels` table
- adjust `_migrate_data` method to support committing multiple records
- adjust `_migrate_data` method to support break condition by max iterations

---

### ✅ Checklist

---

### 🧪 Testing
Deployed model monitoring infrastructure in 1.9, then upgrade to 1.10 and verify that the labels were added as expected during the migration. 

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11142

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/server/py/services/api/initial_data.py
+++ b/server/py/services/api/initial_data.py
@@ -210,7 +210,7 @@ def _migrate_existing_data(
 data_version_prior_to_table_addition = 1
 
 # NOTE: Bump this number when adding a new data migration
-latest_data_version = 9
+latest_data_version = 10
 
 
 def update_default_configuration_data():
@@ -303,6 +303,8 @@ def _perform_data_migrations(db_session: sqlalchemy.orm.Session):
                 _perform_version_8_data_migrations(db, db_session)
             if current_data_version < 9:
                 _perform_version_9_data_migrations(db, db_session)
+            if current_data_version < 10:
+                _perform_version_10_data_migrations(db, db_session)
 
             db.create_data_version(db_session, str(latest_data_version))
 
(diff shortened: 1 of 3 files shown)


server/py/services/api/initial_data.py
```diff
@@ -605,6 +607,49 @@ def _migrate_artifact_labels(
     return labels
 
 
+def _migrate_monitoring_functions_labels(db: framework.db.sqldb.db.SQLDB, db_session):
```
moranbental, line 610: Add a default `chunk_size` so you can test this function with `chunk_size=1` and pass it to the `_migrate_data function`.

server/py/services/api/initial_data.py
```diff
@@ -605,6 +607,49 @@ def _migrate_artifact_labels(
...
+
+        return record, new_label
+
+    return _migrate_data(
+        db=db,
+        db_session=db_session,
+        model=framework.db.sqldb.models.Function,
+        filter_func=filter_infra_func,
+        handle_field_record_func=add_infra_label,
+        max_iterations=1,
```
moranbental, line 649: Why is max_iterations set to 1? I'm not sure I understand the purpose of max_iterations

moranbental (changes requested):
Missing some unit tests to cover this in `test_initial_data`:
add a test for `_migrate_monitoring_functions_labels`, and update the existing test `test_perform_data_migrations_from_initial_supported_version` to include `_perform_version_10_data_migrations`

moranbental (changes requested):
you also need to run `make create-migration`

server/py/services/api/initial_data.py
moranbental, line 649: Why is max_iterations set to 1? I'm not sure I understand the purpose of max_iterations

moranbental (changes requested):
One more thing -

assaf758:
Due to Eyal's vacation, closing this PR - issue will be addressed by Daniel in #8726

PR #8694, opened 2025-10-05
[Security] Normalize and verify user controlled path in import_function_to_dict

elbamit:
### 📝 Description

Fix security vulnerability by normalizing and checking a file path in `import_function_to_dict` that is user-controlled.

---

### 🛠️ Changes Made

-  Refactor`import_function_to_dict` 
- Create test `test_import_function_to_dict_value_errors`
---

### ✅ Checklist

---

### 🧪 Testing
  
  
Unit testing

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11151
- Design docs links:
- External links: https://github.com/mlrun/mlrun/security/code-scanning/115

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/mlrun/run.py
+++ b/mlrun/run.py
@@ -431,7 +431,7 @@ def import_function_to_dict(
             if not path.isfile(candidate_path):
                 raise ValueError(f"no file in exec path (spec.command={code_file})")
 
-            # Optionally enforce absolute path requirement
+            # Check that the path is absolute
             if not os.path.isabs(code_file):
                 raise ValueError(
                     f"exec file spec.command={code_file} is relative, it must be absolute. Change working dir"
--- a/mlrun/run.py
+++ b/mlrun/run.py
@@ -413,21 +413,29 @@ def import_function_to_dict(
             with open(code_file, "wb") as fp:
                 fp.write(code)
         elif cmd:
-            if not path.isfile(code_file):
-                slash_index = url.rfind("/")
-                if slash_index < 0:
-                    raise ValueError(f"no file in exec path (spec.command={code_file})")
-                base_dir = os.path.normpath(url[: slash_index + 1])
-                candidate_path = _ensure_path_confined_to_base_dir(
-                    base_directory=base_dir,
-                    relative_path=code_file,
-                    error_message_on_escape=f"exec file spec.command={code_file} is outside of allowed directory",
-                )
-                if path.isfile(candidate_path):
-                    raise ValueError(
-                        f"exec file spec.command={code_file} is relative, change working dir"
-                    )
+            slash_index = url.rfind("/")
+            if slash_index < 0:
                 raise ValueError(f"no file in exec path (spec.command={code_file})")
+            base_dir = os.path.normpath(url[: slash_index + 1])
+            candidate_path = _ensure_path_confined_to_base_dir(
+                base_directory=base_dir,
+                relative_path=code_file,
+                error_message_on_escape=f"exec file spec.command={code_file} is outside of allowed directory",
+            )
+            # Resolve to absolute path and verify it’s still inside the base directory
+            abs_candidate = os.path.abspath(candidate_path)
+            if not abs_candidate.startswith(base_dir):
+                raise ValueError(
+                    f"exec file spec.command={code_file} resolves outside allowed directory"
+                )
+            # Check if the target file actually exists
+            if not os.path.isfile(abs_candidate):
+                raise ValueError(f"no file in exec path (spec.command={code_file})")
+            # Prevent relative (non-absolute) commands explicitly
+            if not os.path.isabs(code_file):
+                raise ValueError(
+                    f"exec file spec.command={code_file} is relative, it must be absolute. Change working dir"
+                )
         else:
             raise ValueError("command or code not specified in function spec")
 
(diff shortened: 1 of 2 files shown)


weilerN (approved):

mlrun/run.py
```diff
@@ -413,21 +413,29 @@ def import_function_to_dict(
...
+                    f"exec file spec.command={code_file} is outside of allowed directory"
+                ),
+            )
+
+            # Only now it's safe to check file existence
+            if not path.isfile(candidate_path):
                 raise ValueError(f"no file in exec path (spec.command={code_file})")
+
+            # Optionally enforce absolute path requirement
+            if not os.path.isabs(code_file):
```
Yacouby, line 435: You should not use `code_file` here, for consistency and safety you should use `candidate_path`

tests/run/test_run.py
```diff
@@ -443,3 +447,60 @@ def test_run_status_retry_updates(rundb_mock):
         result["status"]["state"]
         == mlrun.common.runtimes.constants.RunStates.pending_retry
     ), "Expected run state to be pending_retry"
+
+
+def test_import_function_to_dict_value_errors():
+    """
+    Test import_function_to_dict for various error cases:
+    missing command/code, path traversal, relative path, and nonexistent file.
+    """
```
Yacouby: All the tests are for errors, what about adding a positive test to ensure the function works in the normal case ?

mlrun/run.py
```diff
@@ -413,21 +413,29 @@ def import_function_to_dict(
...
+                error_message_on_escape=(
+                    f"exec file spec.command={code_file} is outside of allowed directory"
+                ),
+            )
+
+            # Only now it's safe to check file existence
+            if not path.isfile(candidate_path):
                 raise ValueError(f"no file in exec path (spec.command={code_file})")
+
+            # Optionally enforce absolute path requirement
```
Yacouby: This comment is misleading, since the absolute path check is actually enforced, not optional.

tests/run/test_run.py
```diff
@@ -443,3 +447,60 @@ def test_run_status_retry_updates(rundb_mock):
...
+
+    # 4. File does not exist
+    with tempfile.NamedTemporaryFile(
+        suffix=".yaml", mode="w", delete=False
+    ) as temp_file:
+        yaml.dump({"kind": "local", "spec": {"command": "nonexistent.py"}}, temp_file)
+        temp_file_path = temp_file.name
+    with pytest.raises(ValueError, match="no file in exec path"):
+        import_function_to_dict(temp_file_path)
+    os.remove(temp_file_path)
```
Yacouby: Another improvement would be to reduce repetitive `os.remove(...)` calls by using `tempfile.TemporaryDirectory` instead of `NamedTemporaryFile(delete=False)`. 
This way, all temporary files can be created inside a single temporary folder and will be automatically cleaned up when the block exits, making the test cleaner and shorter and more robust against exceptions.

Yacouby (changes requested):
Couple of suggestions and an important one for the unit test, to ensure proper cleanup of temporary files

mlrun/run.py
elbamit: Using `candidate_path` won't check if the user path file is absolute since `candidate_path` is a normalized path

tests/run/test_run.py
```diff
@@ -443,3 +447,62 @@ def test_run_status_retry_updates(rundb_mock):
...
+            match="exec file spec.command=../escape.py is outside of allowed directory",
+        ):
+            import_function_to_dict(temp_file_path)
+
+        # 3. Absolute path required but relative given
+        temp_file_path = os.path.join(temp_dir, "relative.yaml")
+        relative_py_path = os.path.join(temp_dir, "relative.py")
+        with open(temp_file_path, "w") as temp_file:
+            yaml.dump({"kind": "local", "spec": {"command": "relative.py"}}, temp_file)
+        open(relative_py_path, "w").close()
```
Yacouby: Same effect but more readable
The with statement automatically handles closing the file

```suggestion
        with open(relative_py_path, "w"):
                pass
```

tests/run/test_run.py
```diff
@@ -443,3 +447,62 @@ def test_run_status_retry_updates(rundb_mock):
...
+
+
+def test_import_function_to_dict_value_errors():
+    """
+    Test import_function_to_dict for various error cases:
+    missing command/code, path traversal, relative path, and nonexistent file.
+    """
+    with tempfile.TemporaryDirectory() as temp_dir:
+        # 1. Missing command and code
+        temp_file_path = os.path.join(temp_dir, "missing.yaml")
```
Yacouby: Consider using `pathlib.Path` ( you can look how it is begin used in other tests ) instead of `os.path.join`, it makes path handling cleaner, and less error prone.

tests/run/test_run.py
```diff
@@ -443,3 +447,62 @@ def test_run_status_retry_updates(rundb_mock):
         result["status"]["state"]
         == mlrun.common.runtimes.constants.RunStates.pending_retry
     ), "Expected run state to be pending_retry"
+
+
+def test_import_function_to_dict_value_errors():
```
Yacouby: need to rename the test now since it does not check only errors..

Yacouby (changes requested):
Additional improvement suggestions, 
and another one for the unit test, is to use `pytest.mark.parametrize` for the test cases instead of running all scenarios sequentially in a single function.
This will make each case run independently..so with parametrization, pytest will report each scenario separately, making it easier to identify which specific case failed without stopping the rest of the tests.

Yacouby (approved):
Looks good :)

PR #8695, opened 2025-10-05
[Model Monitoring] Set model monitoring app directly from the HUB

danielperezz:
### 📝 Description
`set_model_monitoring_function` now accepts `hub://<func_name>` or` hub://<source>/<func-name>`, which uses the `import_module` SDK to retrieve the monitoring app code from the hub and generate a model-monitoring application that can be deployed to a project.

---

### 🛠️ Changes Made
- adding support in `func` argument which is a hub URL 
- removing the `str` option for the `requirements` parameter since it raises an error (should be either list or None)
- adding a `merge_requirements` helper function which takes two lists of requirements and merge them while giving priority to the first one's specifiers (used for prioritizing user requirements over the module's ones to allow additions and overriding versions)
- changed the value of `HubModuleType.monitoring_app` to match the type of the only module in the hub
---

### ✅ Checklist

---

### 🧪 Testing
- utilize the `test_serving_as_a_job` test to test deployment of monitoring-app from the hub (count-events)
- new unit test for the new helper function (`merge_requirements`)

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11159
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


mlrun/model_monitoring/api.py
```diff
@@ -565,6 +567,17 @@ def _create_model_monitoring_function_base(
         raise mlrun.errors.MLRunValueError(
             "Model monitoring application names cannot end with `-batch`"
         )
+    if check_if_hub_uri(func):
```
assaf758, line 572: Should move after the check that func is not None

mlrun/model_monitoring/api.py
```diff
@@ -565,6 +567,17 @@ def _create_model_monitoring_function_base(
...
+        hub_module = mlrun.get_hub_module(url=func, local_path=local_path)
+        if hub_module.kind != HubModuleType.monitoring_app:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "The provided module is not a monitoring application"
+            )
+        requirements = mlrun.model.ImageBuilder._resolve_requirements(
```
assaf758: I know it was before your pr, but please add typing in the functions being called (in _resolve_requirements, get_hub_module)

tests/system/model_monitoring/test_app.py
```diff
@@ -1122,15 +1125,18 @@ def _log_model(self) -> None:
 
     def _deploy_monitoring_app(self) -> None:
         self.project = typing.cast(mlrun.projects.MlrunProject, self.project)
+        # create temp dir in cwd
+        Path.cwd().joinpath("temp").mkdir(exist_ok=True)
```
assaf758: Use `with` construct and the tempfile package (instead of shutil) to make sure the temp-dir gets deleted also on exceptions.
Something like
```
        with tempfile.TemporaryDirectory(dir=Path.cwd()) as temp_dir:
            fn = self.project.set_model_monitoring_function(
                func=self.app_data["url"],
                application_class=self.app_data["class_name"],
                name=self.app_data["app_name"],
                image="mlrun/mlrun" if self.image is None else self.image,
                local_path=temp_dir,
            )
            self.project.deploy_function(fn)
            return fn
```

mlrun/utils/helpers.py
```diff
@@ -2423,6 +2429,44 @@ def set_data_by_path(
...
+    if reqs is None:
+        return []
+    if isinstance(reqs, str):
+        s = reqs.strip()
+        return [s] if s else []
+    return [s.strip() for s in reqs if s and s.strip()]
+
+
+def merge_requirements(
+    reqs1: typing.Union[str, list[str], None], reqs2: typing.Union[str, list[str], None]
```
assaf758: Since the two reqs are not interchangeable, I think it's better to make this explicit in their naming - for example `reqs_priority` and `reqs_secondary`.
Then when calling the function, `merge_requirements(reqs_priority=requirements, reqs_secondary=hub_module.requirements)` is self-explanatory.

tests/system/model_monitoring/test_app.py
```diff
@@ -1195,11 +1201,16 @@ def _test_batch_ep_results(
         monitoring_results = run_db.get_model_endpoint_monitoring_metrics(
             project=self.project_name,
             endpoint_id=model_endpoint.metadata.uid,
-            type="results",
+            type="metrics",
         )
 
         assert len(monitoring_results) == 2
-        result_names = [result.full_name for result in monitoring_results]
+        result_name = self.app_data["result_name"]
```
assaf758: since `count` is actually a metric (and not result), it makes sense to change result->metric here (result_name -> metric_name etc)

assaf758 (changes requested):
Looks good in general, but have some comments.

mlrun/hub/module.py
```diff
@@ -117,11 +118,22 @@ def verify_directory(path: str) -> Path:
                 raise ValueError(f"Path does not exist: {path}")
             if not path.is_dir():
                 raise ValueError(f"Path is not a directory: {path}")
-        return path
+            return path
+        return Path(os.getcwd())
```
theSaarco, line 125: With this new behavior, I'd change the signature of this function to be:
`def verify_directory(path: Optional[str] = None) -> Path:`
And note this behavior (returning CWD if no path provided) in the docstring.

mlrun/model_monitoring/api.py
```diff
@@ -567,6 +569,19 @@ def _create_model_monitoring_function_base(
         )
     if func is None:
         func = ""
+    if check_if_hub_uri(func):
+        hub_module = mlrun.get_hub_module(url=func, local_path=local_path)
+        if hub_module.kind != HubModuleType.monitoring_app:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "The provided module is not a monitoring application"
+            )
+        requirements = mlrun.model.ImageBuilder._resolve_requirements(
```
theSaarco: Also, given that now it's called from outside the class, change the name of `_resolve_requirements` to `resolve_requirements` to indicate it's being used externally.

mlrun/utils/helpers.py
```diff
@@ -2423,6 +2429,45 @@ def set_data_by_path(
...
+    return [s.strip() for s in reqs if s and s.strip()]
+
+
+def merge_requirements(
+    reqs_priority: typing.Union[str, list[str], None],
+    reqs_secondary: typing.Union[str, list[str], None],
+) -> list[str]:
+    """
+    Merge two requirement collections into a union. If the same package
+    appears in both, the specifier from reqs_secondary wins.
```
theSaarco:
```suggestion
    appears in both, the specifier from reqs_priority wins.
```

mlrun/utils/helpers.py
```diff
@@ -2423,6 +2429,45 @@ def set_data_by_path(
...
+    Args:
+        reqs_priority: str | list[str] | None  (priority input)
+        reqs_secondary: str | list[str] | None
+
+    Returns:
+        list[str]: pip-style requirements.
+    """
+    merged: dict[str, Requirement] = {}
+
+    for r in _normalize_requirements(reqs_priority):
```
theSaarco:
```suggestion
    for r in _normalize_requirements(reqs_secondary) + _normalize_requirements(reqs_primary):
```
If you do this, you can just keep this loop, and get rid of the other one. Since the primary will get processed after the secondary, then overrides will happen automatically once you set it to the same name in the `merged` dict.

mlrun/projects/project.py
```diff
@@ -2417,6 +2418,8 @@ def set_model_monitoring_function(
         :param application_class:       Name or an Instance of a class that implements the monitoring application.
         :param application_kwargs:      Additional keyword arguments to be passed to the
                                         monitoring application's constructor.
+        :param local_path:              Path to a local directory to save the downloaded monitoring-app code files in,
+                                        in case func is a hub URL.
```
theSaarco: Add to the docstring of the `func` param that it can begin with `hub://` and explain what it does.

tests/utils/test_helpers.py
```diff
@@ -1891,3 +1892,20 @@ def test_set_data_by_path_invalid_path(path, value, exc_type, exc_msg):
     data = {}
     with pytest.raises(exc_type, match=exc_msg):
         set_data_by_path(path, data, value)
+
+
+@pytest.mark.parametrize(
+    "priority_reqs, reqs, expected_result",
+    [
+        (None, None, []),
+        ([], ["requests"], ["requests"]),
```
theSaarco, line 1901: Should add `(["requests"],[],["requests"])` for completeness.

theSaarco (changes requested):
Looks good. Have some (mostly styling) comments and suggestions.

assaf758 (approved):

theSaarco (approved):
👍 Looks good.

PR #8696, opened 2025-10-05
[Model Monitoring] Fix `test_list_endpoints_mode` by providing int instead of `EndpointMode`

Eyal-Danieli:
### 📝 Description
Fixing a failure in system tests (https://github.com/mlrun/mlrun/actions/runs/18252292125/job/51968589294#step:9:5610) when the client provides the modes parameter (`EndpointMode` IntEnum object). The API fails because it expects an explicit `int` value:
 `mlrun.errors.MLRunHTTPError: details: [{'loc': ['query', 'mode', 0], 'msg': 'value is not a valid integer', 'type': 'type_error.integer'}]`. Note that this issue occurs when the client runs on python 3.9 (as in the CI env). 
With python 3.11, the API works as expected even if the client provided `EndpointMode`. 

To fix this issue and to ensure compatibility with both versions, we’ll convert the `EndpointMode` value to an `int` before sending the request to the server. No changes are needed on the server side, as `FastAPI` automatically converts incoming `int` query values (e.g. `mode=0`) into the corresponding `EndpointMode` enum instance.

---

### 🛠️ Changes Made

- `list_model_endpoints` (client): Convert the `EndpointMode` `IntEnum` object to an int before making the API call.
- `test_list_endpoints_mode` : Extend the test with an additional assertion when modes provided as a list of `EndpointMode` objects.

---

### ✅ Checklist

---

### 🧪 Testing
- `test_list_endpoints_mode` is passing now with both python 3.9 and 3.11

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11213

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No


mlrun/db/httpdb.py
```diff
@@ -3802,8 +3804,13 @@ def list_model_endpoints(
         labels = self._parse_labels(labels)
         if names and isinstance(names, str):
             names = [names]
-        if isinstance(modes, mm_constants.EndpointMode):
-            modes = [modes]
+        # convert modes into int representation for the API
```
assaf758: Please reference the bug or describe the behavior causing us to do this.

mlrun/projects/project.py
```diff
@@ -3959,9 +3959,7 @@ def list_model_endpoints(
         start: Optional[datetime.datetime] = None,
         end: Optional[datetime.datetime] = None,
         top_level: bool = False,
-        modes: Optional[
-            Union[mm_constants.EndpointMode, list[mm_constants.EndpointMode]]
-        ] = None,
+        modes: Union[mm_constants.EndpointMode, list[mm_constants.EndpointMode]] = None,
```
assaf758: Makes sense to add support for non-list argument, but this is a change in the api.. I prefer we don't do it as part of a bug fix - unless we have to?

assaf758 (commented):
Looks good.
Some minor comment and question

assaf758 (approved):

PR #8712, opened 2025-10-09
[Model Monitoring] Do not pass null results when there's no data in `evaluate`

jond01:
### 📝 Description

Previously, in #8553, we adapted the behavior in MM app evaluation to not run the app code when the sample data is empty (when working with model endpoints).
This, however, still returned a `None` result from the evaluate job, and updated the "last analyzed" time for the application on the given model endpoints.

In this PR, this behavior is fixed:
When there is no data in a given window, the application is still not triggered, AND

1. there is no returned result associated with this window.
2. the last analyzed time is not updated.

Above (2) allows for rerunning the app on the same window, in case it will hold data in the future (and no later window has completed its run).

---

### ✅ Checklist

---

### 🧪 Testing

Manual tests, adaptation of existing tests.

---

### 🔗 References

- Ticket link: [ML-11188](https://iguazio.atlassian.net/browse/ML-11188), [ML-11189](https://iguazio.atlassian.net/browse/ML-11189)
- Previous PR: #8553

---

### 🚨 Breaking Changes?

- [x] No

[ML-11188]: https://iguazio.atlassian.net/browse/ML-11188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ML-11189]: https://iguazio.atlassian.net/browse/ML-11189?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/model_monitoring/applications/base.py
+++ b/mlrun/model_monitoring/applications/base.py
@@ -26,7 +26,6 @@ import pandas as pd
 import mlrun
 import mlrun.common.constants as mlrun_constants
 import mlrun.common.helpers
-import mlrun.common.schemas
 import mlrun.common.schemas.model_monitoring.constants as mm_constants
 import mlrun.common.types
 import mlrun.datastore.datastore_profile as ds_profile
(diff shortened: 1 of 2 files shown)


assaf758 (approved):
Nice!
One comment - if I understand correctly there is no coverage for the content of the schedules file, if correct we need to consider adding.

PR #8720, opened 2025-10-13
[Runtimes] Service-as-a-job: Add parameter validation

alxtkr77:
### 📝 Description

  Adds runtime validation to prevent users from mistakenly passing DataItem parameters via `params` instead of `inputs` when running MLRun functions. This addresses a common user confusion that resulted in cryptic `AttributeError: 'int' object has no attribute 'as_df'` errors.

  **Problem**: Users were encountering confusing errors when passing data via the wrong parameter:
  ```python
  # ❌ Wrong - causes AttributeError
  resp = project.run_function(job, params={"data": 123}, local=True)
```
  Solution: Added validation in get_func_arg() that checks function parameter type hints and raises a clear, actionable error message when DataItem parameters are passed via params.

  ---
  🧪 Testing

  - Unit test: test_dataitem_parameter_passed_via_params_raises_error - Directly tests get_func_arg() validation

  Before (error message):
  AttributeError: 'int' object has no attribute 'as_df'

  After (error message):
  MLRunInvalidArgumentError: Parameter 'data' has type hint '<class 'mlrun.datastore.base.DataItem'>' but was passed via 'params'.
  Data files and artifacts must be passed via the 'inputs' parameter, not 'params'.
  The 'params' parameter is for simple configuration values (strings, numbers, booleans),
  while 'inputs' is for data files that need to be loaded.
  Example: run_function(..., inputs={'data': 'path/to/data.csv'}, params={other_config: value})

  ---
  🔗 References

  - Ticket link: ML-11203
  - Related issue: Users passing serving graph data via params instead of inputs

  ---
  🚨 Breaking Changes?

  - No

  This change only adds validation that catches incorrect usage - it does not affect any correct usage patterns.

### CODE REVIEW

File: tests/serving/test_serving.py [hunk 1]
```diff
@@ -910,3 +910,23 @@ def test_serialize():
     # simulate mlrun/__main__.py
     eval_fn_result = py_eval(str(fn.to_dict()))
     mlrun.utils.helpers.as_dict(eval_fn_result)
+
+
+def test_execute_graph_dataitem_parameter_validation():
+    """Test that passing a DataItem parameter via params instead of inputs raises a clear error"""
+    from mlrun.execution import MLClientCtx
+    from mlrun.serving.server import execute_graph
```
jond01, line 918: Why not import these in the top level imports?

jond01 (approved):
LGTM

tests/serving/test_serving.py
assaf758: @alxtkr77 please consider for follow-up pr.

PR #8621, opened 2025-09-16
[Model Monitoring] Check model monitoring status before enable MM

davesh0812:
### 📝 Description

Check the model monitoring status before enabling MM. The status is determined by the deployment state of the MM infrastructure functions. If any of them are deployed, MM is considered **enabled**; otherwise, it is **disabled**.

---

### 🛠️ Changes Made

When attempting to enable MM, if any of the infrastructure functions are already deployed, a conflict error is raised. To re-enable MM, the user must first disable it.

### ✅ Checklist
- [X] I updated the documentation (if applicable)
- [X] I have tested the changes in this PR

---

### 🧪 Testing
Update `test_model_monitoring_crud` accordingly.
---

### 🔗 References
[ML-10815](https://iguazio.atlassian.net/browse/ML-10815)

---

### 🚨 Breaking Changes?

- [X] No

[ML-10815]: https://iguazio.atlassian.net/browse/ML-10815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8723, opened 2025-10-14
[Tests] Fix model-monitoring test_enable_model_monitoring_after_failure

assaf758:
### 📝 Description

A recent change in the behavior of enable_model_monitoring raises an exception if model-monitoring is already enabled (checked by checking if any of the MM infra functions is deployed).
This change caused test_enable_model_monitoring_after_failure to fail, as the exception was not anticipated

---

### 🛠️ Changes Made

This PR adopts test_enable_model_monitoring_after_failure to expect the exception mentioned.
Also: 
- Updated the wording of said exception
- Fixed some minor import issue

---

### ✅ Checklist

---

### 🧪 Testing
Tested test_enable_model_monitoring_after_failure to pass

---

### 🔗 References
- Ticket link: ML-11211
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [X] No

---


server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -155,9 +155,9 @@ def deploy_monitoring_functions(
             self.set_credentials()
         if deployed_functions := self.get_deployed_model_monitoring_functions():
             raise mlrun.errors.MLRunConflictError(
-                f"Model monitoring functions are already deployed: {deployed_functions}"
-                f". If you want to redeploy them, please use disable_model_monitoring first "
-                f"and enable it again."
+                f"The following model-montioring infra functions are already deployed, aborting: {deployed_functions}\n"
```
jond01:
```suggestion
                f"The following model-montioring infrastructure functions are already deployed, aborting: {deployed_functions}\n"
```

And the same below

tests/system/model_monitoring/test_app.py
```diff
@@ -1814,9 +1814,14 @@ def test_enable_model_monitoring_after_failure(self) -> None:
...
-            image=self.image or "mlrun/mlrun",
-        )
+        with pytest.raises(
+            mlrun.errors.MLRunConflictError,
+            match="The following model-montioring infra functions are already deployed, aborting",
+        ):
+            self.project.enable_model_monitoring(
+                image=self.image or "mlrun/mlrun",
+                wait_for_deployment=True,
+            )
```
jond01, line 1824: I understand that this should fix the failure due to #8621.
If I understand correctly, the change in the behavior from #8621 is backward-incompatible, in the sense that two consecutive `project. enable_model_monitoring()` will fail.
Is this OK? Don't we want to at least document it somewhere?

mlrun/runtimes/__init__.py
```diff
@@ -50,6 +48,7 @@
 from .nuclio.serving import serving_subkind
 from .remotesparkjob import RemoteSparkRuntime
 from .sparkjob import Spark3Runtime
+from .utils import resolve_spark_operator_version
```
jond01: Why is this needed?
Please check item 13 in our Python coding conventions -https://github.com/mlrun/mlrun/blob/development/CONTRIBUTING.md#python-code-conventions

I realize that it was `from x import y` before, maybe we can fix it now

jond01 (approved):
Looks OK, please address the comments

tests/system/model_monitoring/test_app.py
assaf758: Right, I've set External-Release-Note='candidate' on https://iguazio.atlassian.net/browse/ML-10815

mlrun/runtimes/__init__.py
assaf758: There was an LSP diagnostic warning in test_enable_model_monitoring_after_failure when accessing RunError: "utils" is not a known attribute of module ".runtimes"
Will change to import x

server/py/services/api/crud/model_monitoring/deployment.py
jond01: Minor but the line below is still with "infra"

mlrun/runtimes/__init__.py
jond01: Re-looking at it, this import is not strictly internal - users could have previously used `mlrun.runtimes.resolve_spark_operator_version` in their code.
So I suggest reverting the changes in this file.

jond01 (approved):
Looks good!

PR #8728, opened 2025-10-16
[ServingRuntime] Resolve naming conflict between functions and jobs

alxtkr77:
### 📝 Description

Fixes ML-11201: Prevents database collision when converting serving/local functions to jobs via `.to_job()`.

**Problem**: When deploying a serving function and then running it as a job via `.to_job()`, the job overwrites the serving function in the database (they share the same name). This causes:
- ServingRuntime: `.invoke()` fails because deployed Nuclio function state is destroyed
- LocalRuntime: `project.get_function()` returns job instead of original function (version collision where job steals "latest" tag)

**Solution**: Auto-append "-batch" suffix to job names created by `.to_job()`, ensuring separate database entries for functions and their corresponding batch jobs.

---

### 🛠️ Changes Made

**Core Changes:**
- **`mlrun/runtimes/nuclio/serving.py`**: Updated `ServingRuntime.to_job()` to auto-append "-batch" suffix with deep copy of metadata
- **`mlrun/runtimes/local.py`**: Updated `LocalRuntime.to_job()` to auto-append "-batch" suffix with deep copy of metadata
- **`mlrun/run.py`**: Added validation in `code_to_function()` and `new_function()` to prevent users from creating functions with reserved "-batch" suffix
- **`mlrun/launcher/local.py`**: Updated local launcher to strip/restore "-batch" suffix for temporary execution wrappers to pass validation
- **`tests/system/runtimes/test_kubejob.py`**: Updated test assertions to expect "-batch" suffix and added `func_name` parameter tests

**Unified Suffix:**
- Uses existing `_RESERVED_EVALUATE_FUNCTION_SUFFIX = "-batch"` constant from model monitoring for consistency
- Model monitoring applications already use this pattern via `_determine_job_name()`

---

### ✅ Checklist

---

### 🧪 Testing

**System Tests:**
- Updated `test_job_from_serving_runtime` to verify auto-rename behavior
- Added `test_to_job_with_custom_func_name` to test optional `func_name` parameter override
- All existing tests passing with updated naming expectations

---

### 🔗 References
- Ticket link: ML-11201
- Related issue: ML-10794 (model monitoring workflow regression)
- Design discussions: Team consensus on in-place change vs `.to_job_v2()` (Assaf, Saar, Jond, Gilad)

---

### 🚨 Breaking Changes?

- [x] Yes (explain below)
- [ ] No

**Nature of Change**: The `.to_job()` method now automatically appends "-batch" suffix to function names.

**Impact:**
- Batch jobs created via `.to_job()` will have "-batch" appended to their names
- Workflows, KFP pipelines, or scripts that reference job names by hardcoded strings will fail
- Example: Function "my-func" now creates job "my-func-batch" instead of "my-func"

**Justification:**
- **ServingRuntime**: Prevents severe bug where job execution destroys serving function deployment state, making `.invoke()` fail
- **LocalRuntime**: Prevents version collision where job steals "latest" tag from original function
- `.to_job()` is a relatively new feature with **limited field usage** (confirmed by @Gilad Shapira, Oct 2025)
- Team decision: Change behavior in-place rather than introduce `.to_job_v2()` (per @Assaf Ben Amitai, Oct 2025)

**Mitigation:**
- Use `func_name` parameter to specify custom job name: `job = fn.to_job(func_name="custom-name")`
- Update workflow definitions to use new naming convention: `"my-func-batch"`
- Function references in code should use variables, not hardcoded strings

---

### 🔍️ Additional Notes

**Affected Areas:**
- Serving runtimes (Nuclio-based)
- Local runtimes
- Model monitoring applications (already using "-batch", now unified)
- Function creation via `code_to_function()` and `new_function()` (new validation)
- Local launcher (strips/restores suffix for temporary execution wrappers)

**Implementation Highlights:**
- Helper function `ensure_batch_job_suffix()` returns tuple `(modified_name, was_renamed, suffix)` for flexible usage
- Deep copy of metadata prevents reference sharing between original function and job
- Informative logging when auto-rename occurs
- Validation prevents users from manually creating functions with reserved "-batch" suffix
- Optional `func_name` parameter allows explicit override if needed

**Field Usage:**
- `.to_job()` is relatively new feature with **limited production usage** (confirmed by @Gilad Shapira, Oct 2025)
- Team decision: Change behavior in-place rather than introduce `.to_job_v2()` (per @Assaf Ben Amitai, Oct 2025)
- Breaking change acceptable due to: new feature status, prevents silent corruption, mitigation available via `func_name`

**Known Issues:**
- ML-10794: Model monitoring workflows may break if they hardcode function names (workaround: use `func_name` parameter or update workflow definitions)


mlrun/launcher/local.py
```diff
@@ -243,7 +243,35 @@ def _create_local_function_for_execution(
 
         # if the handler has module prefix force "local" (vs "handler") runtime
         kind = "local" if isinstance(handler, str) and "." in handler else ""
-        fn = mlrun.new_function(meta.name, command=command, args=args, kind=kind)
+
```
assaf758, line 246: i'm not sure why we need to strip the prefix in the local launcher?

mlrun/runtimes/local.py
```diff
@@ -201,9 +201,51 @@ class LocalRuntime(BaseRuntime, ParallelRunner):
     kind = "local"
     _is_remote = False
 
-    def to_job(self, image=""):
+    def to_job(self, image="", func_name: Optional[str] = None):
```
assaf758, line 204: also in the `local` runtime, why do we need to change to_job?

jond01:
@alxtkr77 

> 🔗 References
>
> - Ticket link: ML-10940

This ticket is closed. Are you sure that is the correct ticket?

> This is backward compatible. Existing code will continue to work, with the only visible change being that batch jobs created via to_job() will now have -batch appended to their names automatically.

I'm quite sure it is a breaking change. Changing the names of jobs is not backwards compatible.
For example, see this issue which was opened just for a specific private case: [ML-10794](https://iguazio.atlassian.net/browse/ML-10794)

This point alone is backwards incompatible, and you have more:

> - Validation prevents users from manually creating functions with the reserved suffix

[ML-10794]: https://iguazio.atlassian.net/browse/ML-10794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

server/py/services/api/api/endpoints/nuclio.py
```diff
@@ -331,7 +331,7 @@ async def deploy_status(
     if fn.get("kind") not in mlrun.runtimes.RuntimeKinds.nuclio_runtimes():
         framework.api.utils.log_and_raise(
             HTTPStatus.BAD_REQUEST.value,
-            reason=f"Runtime kind {fn.kind} is not a nuclio runtime",
+            reason=f"Runtime kind `{fn.get('kind')}` is not a nuclio runtime",
```
jond01, line 334: Nice, it's a bug fix, right? (because `fn` a a dictionary)

mlrun/model_monitoring/api.py
```diff
@@ -561,9 +561,10 @@ def _create_model_monitoring_function_base(
             "An application cannot have the following names: "
             f"{mm_constants._RESERVED_FUNCTION_NAMES}"
         )
-    if name and name.endswith(mm_constants._RESERVED_EVALUATE_FUNCTION_SUFFIX):
+    _, was_changed, suffix = mlrun.utils.helpers.ensure_batch_job_suffix(name)
+    if name and not was_changed:
```
jond01: I find the new condition harder to understand. How about defining `ends_with_batch = not was_changed`?

mlrun/launcher/local.py
alxtkr77: The reason we strip the "-batch" suffix in the local launcher is because mlrun.new_function() has validation that prevents local function names from ending with the reserved batch suffix.

mlrun/runtimes/local.py
alxtkr77: LocalRuntime has the exact same problem:
```
  # Step 1: Create and store local function
  local_fn = project.set_function("script.py", name="my-func", kind="local")
  # DB: {name: "my-func", tag: "latest", kind: "local", version: 1}

  # Step 2: Convert to job and run (creates NEW VERSION)
  job = local_fn.to_job()  # Without suffix: name="my-func"
  job.run()
  # DB: {name: "my-func", tag: "latest", kind: "job", version: 2}  ← Steals "latest"!
  # DB: {name: "my-func", tag: "",       kind: "local", version: 1}  ← Lost "latest" tag

  # Step 3: Try to retrieve the local function
  fn = project.get_function("my-func")  # Defaults to tag="latest"
  # → Returns: {kind: "job", version: 2}  ← Wrong! User expected local function

  # Step 4: User confusion
  print(fn.kind)  # "job" (expected "local")
  fn.run()  # Runs as job, not as local function
```

mlrun/utils/helpers.py
```diff
@@ -474,6 +474,71 @@ def normalize_name(name: str):
...
+
+    """
+    suffix = mlrun_constants.RESERVED_BATCH_JOB_SUFFIX
+
+    # Handle None or empty string
+    if not function_name:
+        return function_name, False, suffix
+
+    if function_name.endswith(suffix):
+        return function_name[: -len(suffix)], True, suffix
```
jond01: I suggest using the more direct [`str.removesuffix`](https://docs.python.org/3/library/stdtypes.html#str.removesuffix) method:
```suggestion
        return function_name.removesuffix(suffix), True, suffix
```

mlrun/launcher/local.py
jond01: +1

And what would be the behavior when running a remote job with `local=False` and then `True`?
I think that depending on the job name, we may get 2 functions in the DB if the original ends with `"-batch"`, or a single one otherwise.

alxtkr77:
@jond01 , somehow cannot reply to your comment
1. Fixed the Ticket link: ML-10940 to ML-11201
2. Update the description of the PR to address your concerns

mlrun/run.py
```diff
@@ -576,6 +576,16 @@ def new_function(
         )
 
     runner.prepare_image_for_deploy()
+
+    # Validate serving and local function names don't end with reserved batch suffix
+    if kind in ["serving", "serving_v2", "local"]:
```
jond01: I suggest using the kinds defined in `RuntimeKinds`.
Please note that `"serving_v2"` is not there - can the kind have this value?

If not, please update also below `function_type` -> `kind`.

mlrun/run.py
```diff
@@ -576,6 +576,16 @@ def new_function(
...
+        _, was_changed, suffix = mlrun.utils.helpers.ensure_batch_job_suffix(name)
+        if name and not was_changed:
```
jond01: Same here, it's not that straightforward to understand the condition from the `was_changed` variable name and `not`

mlrun/run.py
jond01, line 1: Is it possible to unify the checks from `new_function` and `code_to_function`?

mlrun/runtimes/local.py
```diff
@@ -201,9 +201,51 @@ class LocalRuntime(BaseRuntime, ParallelRunner):
...
+            The original local function remains unchanged.
+        """
+        from copy import deepcopy
+
+        # Deep copy to prevent reference sharing
         struct = self.to_dict()
         obj = KubejobRuntime.from_dict(struct)
+
+        # Deep copy metadata to prevent reference sharing
+        obj.metadata = deepcopy(obj.metadata)
```
jond01: 1. Why importing here?
2. I think `deepcopy` is redundant here, `struct = self.to_dict()` is not shared.

mlrun/runtimes/local.py
```diff
@@ -201,9 +201,51 @@ class LocalRuntime(BaseRuntime, ParallelRunner):
...
+        original_name = obj.metadata.name
+
+        if func_name:
+            # User provided explicit job name
+            obj.metadata.name = func_name
+            logger.info(
+                f"Creating job '{func_name}' from local function '{original_name}'"
+            )
```
jond01: Why is this log helpful? I think it should be debug level if it stays here

mlrun/runtimes/nuclio/serving.py
```diff
@@ -886,8 +898,32 @@ def to_job(self) -> KubejobRuntime:
...
+
+        job_metadata = deepcopy(self.metadata)
+        original_name = job_metadata.name
+
+        if func_name:
+            # User provided explicit job name
+            job_metadata.name = func_name
+            logger.info(
+                f"Creating job '{func_name}' from serving function '{original_name}'"
+            )
```
jond01: Same

tests/system/runtimes/test_kubejob.py
```diff
@@ -814,6 +837,56 @@ async def test_job_from_serving_runtime_from_inside_running_asyncio_loop(self):
         finally:
             v3io_client.close()
 
+    def test_to_job_with_custom_func_name(self):
```
jond01: I think this test can be a unit test, it barely uses the system

jond01 (commented):
Looks good!
I had some comments.

server/py/services/api/api/endpoints/nuclio.py
alxtkr77: yes

mlrun/model_monitoring/api.py
alxtkr77: will fix

mlrun/utils/helpers.py
alxtkr77: will fix

mlrun/run.py
alxtkr77: Regarding fn = mlrun.new_function(meta.name, command=command, args=args, kind=kind): I immediately restore the name of the function :        
``` fn = mlrun.new_function(
            local_function_name, command=command, args=args, kind=kind
        )
        fn.metadata = meta
```
So no problem should be here. We just need to bypass the new_function() validator. Another option would be to add a parameter to the new_function(skip_suffix_verification=False). I think they both ugly :)

as for the second remark -will fix

mlrun/run.py
alxtkr77: will fix

mlrun/run.py
alxtkr77: will refactor

mlrun/runtimes/local.py
alxtkr77: will fix

mlrun/runtimes/local.py
alxtkr77: will fix

mlrun/runtimes/nuclio/serving.py
alxtkr77: will fix

tests/system/runtimes/test_kubejob.py
alxtkr77: will move

mlrun/launcher/local.py
alxtkr77: local=FalsRegarding fn = mlrun.new_function(meta.name, command=command, args=args, kind=kind): I immediately restore the name of the function :

            local_function_name, command=command, args=args, kind=kind
        )
        fn.metadata = meta
So no problem should be here. We just need to bypass the new_function() validator. Another option would be to add a parameter to the new_function(skip_suffix_verification=False)

mlrun/utils/helpers.py
```diff
@@ -474,6 +474,98 @@ def normalize_name(name: str):
     return name.lower()
 
 
+def ensure_batch_job_suffix(
+    function_name: typing.Optional[str],
+) -> tuple[str, bool, str]:
```
yaelgen:
```suggestion
) -> tuple[typing.Optional[str], bool, str]:
```

the docstring says “or empty string if input was empty,” but code may return None. Align docstring and typing with behavior, or change behavior

mlrun/utils/helpers.py
```diff
@@ -474,6 +474,98 @@ def normalize_name(name: str):
...
+            f"{function_name}{suffix}",
+            True,
+            suffix,
+        )
+    return function_name, False, suffix
+
+
+def strip_batch_job_suffix(
+    function_name: typing.Optional[str],
+) -> tuple[str, bool, str]:
```
yaelgen: same

mlrun/runtimes/nuclio/serving.py
```diff
@@ -886,8 +898,32 @@ def to_job(self) -> KubejobRuntime:
...
+            if was_renamed:
+                logger.info(
+                    f"Creating job '{job_metadata.name}' from serving function '{original_name}' "
+                    f"(auto-appended '{suffix}' to prevent database collision)"
+                )
+            else:
+                logger.info(
+                    f"Creating job '{job_metadata.name}' from serving function '{original_name}'"
+                )
+
```
yaelgen, line 949: The` to_job()` flows in Serving and Local are ~identical (deep-copy/derive metadata, optional func_name, ensure/append -batch, log). Consider extracting a single helper and use it from both places.

mlrun/utils/helpers.py
```diff
@@ -474,6 +474,98 @@ def normalize_name(name: str):
...
+
+    :param name: Function name to validate
+    :param kind: Function kind ("serving", "serving_v2", "local", etc.)
+
+    :raises MLRunValueError: If the function name ends with the reserved batch suffix
+    """
+    import mlrun.errors
+
+    # Only validate for kinds that have to_job() methods that add the suffix
+    if kind not in ["serving", "serving_v2", "local"]:
```
yaelgen: I Suggest centralizing the list of kinds that support to_job() instead of hardcoding it (constant or utility)

mlrun/runtimes/local.py
```diff
@@ -201,9 +201,46 @@ class LocalRuntime(BaseRuntime, ParallelRunner):
...
+            if was_renamed:
+                logger.info(
+                    f"Creating job '{obj.metadata.name}' from local function '{original_name}' "
+                    f"(auto-appended '{suffix}' to prevent database collision)"
+                )
+            else:
+                logger.info(
+                    f"Creating job '{obj.metadata.name}' from local function '{original_name}'"
+                )
```
yaelgen: valuable but might be noisy, consider logger.debug for the “no rename” path and keep logger.info only when a rename actually occurred (also I would flip both to structured logs with fields {original_name, new_name, suffix, was_renamed} for easier filtering).

mlrun/utils/helpers.py
```diff
@@ -474,6 +474,98 @@ def normalize_name(name: str):
...
+    # Handle None or empty string
+    if not function_name:
+        return function_name, False, suffix
+
+    if not function_name.endswith(suffix):
+        return (
+            f"{function_name}{suffix}",
+            True,
+            suffix,
+        )
```
yaelgen, line 543: Appending `-batch` increases the function name length, which might push it over Kubernetes’ 63-character limit for resource names, we may want to enforce or validate max name length before appending the suffix?

mlrun/utils/helpers.py
```diff
@@ -474,6 +474,98 @@ def normalize_name(name: str):
     return name.lower()
```
yaelgen, line 512: Add ut coverage for both please

yaelgen (dismissed):

jond01:
> @jond01 , somehow cannot reply to your comment
> 
> 1. Fixed the Ticket link: ML-10940 to ML-11201
> 2. Update the description of the PR to address your concerns

@alxtkr77 thanks, there is a small typo in the description - ML-11021 instead of ML-11201


will fix

mlrun/utils/helpers.py
jond01: https://github.com/mlrun/mlrun/pull/8728#discussion_r2454528436

I suggest using the kinds defined in `RuntimeKinds`.
Please note that `"serving_v2"` is not there - can the kind have this value?

If not, please update also below `function_type` -> `kind`.

tests/system/runtimes/test_kubejob.py
jond01: I don't see these tests anymore 😮 

I moved it to test_to_job.py

mlrun/launcher/local.py
```diff
@@ -243,7 +243,35 @@ def _create_local_function_for_execution(
...
+        # Strip batch job suffix if present - local execution doesn't need collision prevention
+        # since it's a temporary runtime wrapper, not a stored function
+        #   ServingRuntime.to_job("my-function")
```
jond01: Please change to `LocalRuntime`

mlrun/launcher/local.py
jond01: Ok by me - I see that we get the following error:
```
MLRunValueError: Local function names cannot end with `-batch`
```

jond01 (changes requested):
I encountered the following unexpected behavior:

```py
import mlrun

project = mlrun.get_or_create_project("ml-11201", allow_cross_project=True)
# `func.py` is a simple Python file with a function named `func`
fn = project.set_function(func="func.py", name="myfunc", kind="local", image="quay.io/mlrun/mlrun:v1.10.0-rc37", handler="func")
fn.save()

fn_job = fn.to_job()
fn_job.save()
assert fn_job.kind == "job"  # <-- fails
```

The `to_job` function kind is still "local".
I also tried to run it and encountered an issue.



Nice catch - an old bug! Will fix and add tests

tests/system/runtimes/test_kubejob.py
alxtkr77: Added a new module - tests/runtimes/test_to_job.py

mlrun/utils/helpers.py
alxtkr77: will fix

mlrun/utils/helpers.py
alxtkr77: will fix

mlrun/runtimes/nuclio/serving.py
alxtkr77: I would leave it as is. It does not fit the ModelMonitoringApplicationBase::to_job(), and it's only the Serving and Local runtimes. Also, the piece of code to refactor is mostly logging - the logic is already wrapped into ensure_batch_job_suffix().

mlrun/utils/helpers.py
alxtkr77: will fix

mlrun/utils/helpers.py
alxtkr77: Actually , you reveal a bigger architectural gap - we have scattered, inconsistent validation for input parameters:

  Current State

  ✅ What IS Validated:

  | Parameter         | Where               | When         | Pattern/Limit            |
  |-------------------|---------------------|--------------|--------------------------|
  | Project names     | projects/project.py | Creation     | DNS-1123 label, 63 chars |
  | Label keys/values | k8s_utils.py        | When setting | DNS-1123, 63 chars       |
  | Run names         | runtimes/base.py    | Before run   | Label value, 63 chars    |
  | Tag names         | utils/helpers.py    | When setting | Label value, 63 chars    |
  | Node selectors    | k8s_utils.py        | When setting | Label key/value          |

  ❌ What is NOT Validated:

  | Parameter       | Should Be                | Current State   | Impact             |
  |-----------------|--------------------------|-----------------|--------------------|
  | Function names  | DNS-1123 (63 max)        | Only normalized | Fails at k8s time  |
  | Container names | DNS-1123 (63 max)        | Not validated   | Pod creation fails |
  | Service names   | DNS-1123 (63 max)        | Not validated   | Service fails      |
  | Env var names   | ^[A-Za-z_][A-Za-z0-9_]*$ | Not validated   | Pod fails          |
  | Volume names    | DNS-1123 (63 max)        | Not validated   | Pod fails          |
  | Secret/CM keys  | 253 char max             | Not validated   | Secret fails       |

While this analysis maybe incomplete, I am tagging @assaf758 and @theSaarco 
Meanwhile I will add specific verification you suggest, but broader approach is required.

mlrun/utils/helpers.py
alxtkr77: will do

mlrun/utils/helpers.py
alxtkr77: serving_v2 is a subkind, you are correct. Will remove
Also will refactor

mlrun/launcher/local.py
alxtkr77: will fix

alxtkr77:
> I encountered the following unexpected behavior:
> 
> ```python
> import mlrun
> 
> project = mlrun.get_or_create_project("ml-11201", allow_cross_project=True)
> # `func.py` is a simple Python file with a function named `func`
> fn = project.set_function(func="func.py", name="myfunc", kind="local", image="quay.io/mlrun/mlrun:v1.10.0-rc37", handler="func")
> fn.save()
> 
> fn_job = fn.to_job()
> fn_job.save()
> assert fn_job.kind == "job"  # <-- fails
> ```
> 
> The `to_job` function kind is still "local". I also tried to run it and encountered an issue.

Nice catch - an old bug! Will fix and add tests

mlrun/runtimes/local.py
alxtkr77: will fix

mlrun/runtimes/__init__.py
```diff
@@ -221,6 +233,24 @@ def is_local_runtime(kind):
             return True
         return False
 
+    @staticmethod
+    def requires_k8s_name_validation(kind):
```
jond01:
```suggestion
    def requires_k8s_name_validation(kind: str) -> bool:
```

mlrun/runtimes/__init__.py
```diff
@@ -186,6 +186,18 @@ def local_runtimes():
             RuntimeKinds.handler,
         ]
 
+    @staticmethod
+    def runtimes_with_to_job():
```
jond01:
```suggestion
    def runtimes_with_to_job() -> list[str]:
```

mlrun/runtimes/local.py
```diff
@@ -201,9 +201,58 @@ class LocalRuntime(BaseRuntime, ParallelRunner):
     kind = "local"
     _is_remote = False
 
-    def to_job(self, image=""):
+    def to_job(self, image="", func_name: Optional[str] = None):
```
jond01, line 204:
```suggestion
    def to_job(self, image: str = "", func_name: Optional[str] = None) -> KubejobRuntime:
```

mlrun/utils/helpers.py
```diff
@@ -476,6 +476,114 @@ def normalize_name(name: str):
...
+    :param name: Function name to validate
+    :param kind: Function kind ("serving", "serving_v2", "local", etc.)
+
+    :raises MLRunValueError: If the function name ends with the reserved batch suffix or would exceed
+                            the Kubernetes name length limit after appending the suffix
+    """
+    # Use dynamic import to avoid circular dependency detected by import linter
+    import importlib
+
+    runtimes_module = importlib.import_module("mlrun.runtimes")
```
jond01: I think it's better to move this function instead of performing this dirty import, which the linter does not catch, but still contradicts its rule:
https://github.com/mlrun/mlrun/blob/db8a1f9777dfb14e4b3b7eb33a051d70032feaa0/pyproject.toml#L106

jond01 (approved):
Great work! 🚀 
I left some minor comments, and please check the failing tests.

mlrun/k8s_utils.py
```diff
@@ -236,6 +236,43 @@ def handle_invalid(message):
...
+    Refer to https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names
+
+    :param name: The function name to validate (after normalization)
+    :raises MLRunInvalidArgumentError: If the function name is invalid for Kubernetes
+    """
+    # Import here to avoid circular dependency (helpers imports k8s_utils)
+    import mlrun.utils.helpers
+
+    if not name:
+        raise mlrun.errors.MLRunInvalidArgumentError("Function name cannot be empty.")
```
jond01: Looks like this condition causes many errors:
https://github.com/mlrun/mlrun/actions/runs/18844587037/job/53765192200?pr=8728#step:9:14649

mlrun/runtimes/__init__.py
alxtkr77: will fix

mlrun/runtimes/__init__.py
alxtkr77: will fix

mlrun/runtimes/local.py
alxtkr77: will fix

mlrun/utils/helpers.py
alxtkr77: will fix

mlrun/k8s_utils.py
alxtkr77: will fix

mlrun/k8s_utils.py
```diff
@@ -236,6 +236,92 @@ def handle_invalid(message):
     return True
 
 
+def validate_function_name(name: str) -> None:
```
theSaarco: Why not put this inside `mlrun.utils.helpers` (which will help you get rid of the internal import as well)? While the regex is driven by the k8s limitation, this looks like a general validation that may be useful as a general helper, not just in server-side context.

mlrun/run.py
```diff
@@ -576,6 +579,10 @@ def new_function(
         )
 
     runner.prepare_image_for_deploy()
+
+    # Validate serving and local function names don't end with reserved batch suffix
+    mlrun.k8s_utils.validate_function_name_for_batch_suffix(name, kind)
```
theSaarco: This is problematic in terms of BC, since it effectively mean that allowed length for ANY serving function is not reduced to 57 chars, instead of previously allowed 63. This is true even if the user never intends to perform `to_job()` on the serving function.

mlrun/runtimes/local.py
```diff
@@ -201,9 +201,60 @@ class LocalRuntime(BaseRuntime, ParallelRunner):
...
-    def to_job(self, image=""):
+    def to_job(
+        self, image: str = "", func_name: Optional[str] = None
+    ) -> KubejobRuntime:
+        """Convert this LocalRuntime to a KubejobRuntime.
+
+        Args:
+            image: Optional Docker image to use for the job.
+            func_name: Optional custom name for the job function. If not provided, automatically
+                      appends '-batch' suffix to the local function name to prevent database collision.
```
theSaarco: Why is this needed for local runtimes? We are trying to solve an issue with serving runtimes, who get converted from a nuclio function to an MLRun runtime when using `to_job()`. These issues are not necessarily shared with the local runtime and therefore this is out of scope for this fix.
While this may be helpful, it also breaks BC for existing functionality, and therefore we shouldn't do it here (definitely not if the current situation doesn't generate issues).

mlrun/runtimes/__init__.py
```diff
@@ -186,6 +186,18 @@ def local_runtimes():
...
+    @staticmethod
+    def runtimes_with_to_job() -> list[str]:
+        """Runtimes that support conversion to batch jobs via to_job() method.
+
+        These runtimes auto-append -batch suffix to prevent database collision
+        when converted to jobs.
+        """
+        return [
+            RuntimeKinds.serving,
+            RuntimeKinds.local,
```
theSaarco: See my comment in the local runtime code - just because the local runtime has a `to_job()` method, it doesn't mean we need to add a suffix to it.

theSaarco (changes requested):
I started reviewing, but at some point stopped since I think we need to consider a different approach here. My main concern is that this PR turned out to be potentially disruptive to various current user flows. For example, it now validates all serving function names must conform to a stricter length, even if the user never intends to run them as batch. This may break existing functionality for users.
In addition, you added the `-batch` suffix to local runtimes in their `to_job()` which again differs from existing functionality and is not BC (while arguably probably not used too much).

I think we need to go with a different approach:
* Validate function name full length as usual (i.e. keep the original 63 chars length). This should be done for all runtimes, since it does the same validation already applied in various places. Note that this also applies for serving functions.
* When the user runs `to_job()` in a serving function **and only then**, try to add the `-batch` suffix. If the resulting length is too long - fail the action, generate a meaningful error to the users and instruct them to provide a different name (through the `func_name` parameter of `to_job()`).

This will result in BC behavior, only introducing a change to existing functionality for the isolated use-case that we are interested in.

mlrun/k8s_utils.py
alxtkr77: My reasoning is that all the k8s validation functions are here: verify_label_key(), verify_label_value() etc. 
The 'import' and the comment message is wrong though - will remove

mlrun/run.py
alxtkr77: will fix following your suggestion above

mlrun/runtimes/local.py
alxtkr77: Looks like the LocalRuntime has the same problem:
```
 local_fn = project.set_function("script.py", name="my-func", kind="local")
 # DB: {name: "my-func", tag: "latest", kind: "local", version: 1}

  # Step 2: Convert to job and run (creates NEW VERSION)
  job = local_fn.to_job()  # Without suffix: name="my-func"
  job.run()
  # DB: {name: "my-func", tag: "latest", kind: "job", version: 2}  ← Steals "latest"!
  # DB: {name: "my-func", tag: "",       kind: "local", version: 1}  ← Lost "latest" tag

  # Step 3: Try to retrieve the local function
  fn = project.get_function("my-func")  # Defaults to tag="latest"
  # → Returns: {kind: "job", version: 2}  ← Wrong! User expected local function

  # Step 4: User confusion
  print(fn.kind)  # "job" (expected "local")
  fn.run()  # Runs as job, not as local function
```

mlrun/runtimes/__init__.py
alxtkr77: See above

mlrun/runtimes/local.py
assaf758: Since allows different run-times to share the same function name, I agree with Saar that there is no justification in creating a change in behavior for Local.to_job() (which exist in the field).

mlrun/run.py
```diff
@@ -538,6 +538,9 @@ def new_function(
     # make sure function name is valid
     name = mlrun.utils.helpers.normalize_name(name)
 
+    if kind not in RuntimeKinds.runtimes_with_to_job():
```
theSaarco: Since you're checking the full length in `validate_function_name`, then why do you need this check? Why not do the check also on serving?

mlrun/runtimes/nuclio/serving.py
```diff
@@ -890,8 +905,59 @@ def to_job(self) -> KubejobRuntime:
...
+            ):
+                max_allowed_length = (
+                    mlrun.common.constants.K8S_DNS_1123_LABEL_MAX_LENGTH - len(suffix)
+                )
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    f"Cannot convert serving function '{original_name}' to batch job: "
+                    f"the resulting name '{job_metadata.name}' ({len(job_metadata.name)} characters) "
+                    f"exceeds Kubernetes limit of {mlrun.common.constants.K8S_DNS_1123_LABEL_MAX_LENGTH} characters. "
+                    f"Please provide a custom name via the func_name parameter, "
+                    f"with at most {max_allowed_length} characters."
```
theSaarco:
```suggestion
                    f"with at most {mlrun.common.constants.K8S_DNS_1123_LABEL_MAX_LENGTH} characters."
```
Since we don't add `-batch` when a custom name is given.

mlrun/runtimes/nuclio/serving.py
```diff
@@ -890,8 +905,59 @@ def to_job(self) -> KubejobRuntime:
...
+                )
+
+            if was_renamed:
+                logger.info(
+                    "Creating job from serving function (auto-appended suffix to prevent collision)",
+                    original_name=original_name,
+                    new_name=job_metadata.name,
+                    suffix=suffix,
+                    was_renamed=True,
```
theSaarco: Half of these params in the log are redundant. The `new_name` and `suffix` are enough - the rest can be easily deduced.

mlrun/runtimes/nuclio/serving.py
```diff
@@ -890,8 +905,59 @@ def to_job(self) -> KubejobRuntime:
...
+                )
+            else:
+                logger.debug(
+                    "Creating job from serving function (name already has suffix)",
+                    original_name=original_name,
+                    new_name=job_metadata.name,
+                    suffix=suffix,
+                    was_renamed=False,
```
theSaarco: Same here - the `original_name` (and perhaps the `suffix`) are enough.

mlrun/k8s_utils.py
theSaarco: Conceptually, if at some point there will be other limiting factors on function name driven from another constraint, they will be added in this function, making it wrongly placed in k8s_utils. In that case we'd want to move the function to the utils section instead.

theSaarco (changes requested):
Looks good, with the name validation being an important addition to the MLRun logic.
Repeating my claim - local runtime names should not be altered. Let's keep this change specific to serving runtimes.

mlrun/k8s_utils.py
alxtkr77: will fix

mlrun/runtimes/local.py
alxtkr77: will remove

mlrun/runtimes/__init__.py
alxtkr77: will fix

mlrun/runtimes/nuclio/serving.py
alxtkr77: will fix

mlrun/runtimes/nuclio/serving.py
alxtkr77: will fix

mlrun/runtimes/nuclio/serving.py
alxtkr77: will fix

theSaarco (approved):
LGTM. 
Very nice UT coverage.

PR #8732, opened 2025-10-18
[Test] Normalize artifacts datetime fields prior to deepdiff comparison

elbamit:
### 📝 Description

Test was failing transiently due to a difference in datetime object formatting. In `datetime.isoformat()` if the microsecond value is 0, Python omits it by default (documented [here](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat). SQLAlchemy stores timestamps with microsecond precision, even if they are `.0000`. This can sometimes cause a formatting mismatch, resulting in the assertion error described in the ticket.
This PR normalize datetime fields prior to the deepdiff comparison

---

### 🛠️ Changes Made

- Added a function `_normalize_datetime_fields` inside `_validate_run_artifacts` to handle the formatting mismatch

---

### ✅ Checklist

---

### 🧪 Testing

To test this I have altered the `_update_artifact_record_from_dict` function which is responsible to update the `created` and `updated` fields when storing an artifact. I made this function save the artifact with hardcoded values for the datetime fields, making sure that the changes in the PR solve them.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11231
- Design docs links:
- External links: https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [ ] No

---

### DIFF
--- a/server/py/services/api/tests/unit/crud/test_runs.py
+++ b/server/py/services/api/tests/unit/crud/test_runs.py
@@ -14,6 +14,7 @@
 
 import unittest.mock
 import uuid
+from datetime import datetime
 
 import deepdiff
 import pytest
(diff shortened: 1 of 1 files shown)


server/py/services/api/tests/unit/crud/test_runs.py
```diff
@@ -935,6 +936,20 @@ def _validate_run_artifacts(
         iter=0,
         run_format: mlrun.common.formatters.RunFormat = None,
     ):
+        def _normalize_datetime_fields(artifact):
```
liranbg:
```suggestion
        def normalize_datetime_fields(artifact):
```

no need to underscore a nested function. also better of adding a one-liner comment atop function of what is the function actually doing. e.g "Ensures timestamp has milliseconds" or so

PR #8741, opened 2025-10-20
[Serving] Add  `RemoteFunctionStep` Graph Step

daniels290813:
### 📝 Description
To call a remote URL as part of graph invocation, users today should use the `RemoteStep`, which has various options for invoking remote endpoints.

Suggestion is to make this easier when wanting to call a remote function. This will be done by creating a new step class `RemoteFunctionStep` that will subclass the RemoteStep and instead of getting a URL will be created with the remote function object or the function’s URI.

---

### 🛠️ Changes Made
RemoteFunctionStep added to `serving/remote` 

---

### ✅ Checklist

---

### 🧪 Testing
unittest

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10099
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [X] No

---


mlrun/serving/remote.py
```diff
@@ -527,3 +527,59 @@ def post_init(self, mode="sync", **kwargs):
...
+        super().__init__(url="", method=method, **kwargs)
+        self.rundb = None
+        self.fn = fn
+
+    def post_init(self, mode="sync", **kwargs):
+        self.rundb = mlrun.get_run_db()
+        if not self.fn:
+            raise mlrun.errors.MLRunRuntimeError(
+                f"Cannot find function '{self.fn}' in the MLRun DB. "
+                "Please verify that the function URI is correct and that it is stored properly."
```
jillnogold: "Verify that the function URI is correct and that it is stored properly."

mlrun/serving/remote.py
```diff
@@ -527,3 +527,59 @@ def post_init(self, mode="sync", **kwargs):
...
+            )
+
+        if isinstance(self.fn, str):
+            self.fn = self.rundb.get_function(
+                self.fn.split("/")[0], self.fn.split("/")[1]
+            )
+            if not self.fn:
+                raise mlrun.errors.MLRunRuntimeError(
+                    f"Cannot find function '{self.fn}' in the MLRun DB. "
+                    "Please verify that the function URI is correct and that the function is stored properly."
```
jillnogold: "Verify that the function URI is correct and that the function is stored properly."

mlrun/serving/remote.py
```diff
@@ -527,3 +527,59 @@ def post_init(self, mode="sync", **kwargs):
...
+                )
+        # Derive Nuclio function URL
+        url = (
+            getattr(self.fn.status, "address", None)
```
GiladShapira94: I think it's worth changing the order and using the `internal` URLs as the first check

mlrun/serving/remote.py
```diff
@@ -527,3 +527,59 @@ def post_init(self, mode="sync", **kwargs):
...
+            raise mlrun.errors.MLRunRuntimeError(
+                f"Cannot find function '{self.fn}' in the MLRun DB. "
+                "Please verify that the function URI is correct and that it is stored properly."
+            )
+        if not isinstance(self.fn, (mlrun.runtimes.RemoteRuntime, str)):
+            raise mlrun.errors.MLRunInvalidArgumentTypeError(
+                "Parameter 'fn' must be of type mlrun.runtimes.RemoteRuntime or str."
+            )
+
+        if isinstance(self.fn, str):
```
GiladShapira94: If that's possible, we need to ask the user to provide only the function name and inherit the project name from the client or the serving function spec

mlrun/serving/remote.py
```diff
@@ -527,3 +527,59 @@ def post_init(self, mode="sync", **kwargs):
         self.fill_placeholders = self.fill_placeholders or False
         self.rundb = mlrun.get_run_db()
         self.url = self.rundb.get_base_api_url(self.path)
+
+
+class NuclioRemoteStep(RemoteStep):
```
GiladShapira94: As Nuclio is the base for all remote functions, it may be worth change the Class name to `RemoteFunctionStep`

mlrun/serving/remote.py
daniels290813: @theSaarco wdyt ? 
Currently nuclio function uri is in the format `project name/function name` 
Gilad suggested to use config.active_project or even get it as an additional class parameter.

daniels290813:
@GiladShapira94 Changed step name to RemoveFunctionStep and class can get project_name parameter or it will derive it using active_project

mlrun/serving/remote.py
```diff
@@ -527,3 +527,86 @@ def post_init(self, mode="sync", **kwargs):
...
         self.rundb = mlrun.get_run_db()
         self.url = self.rundb.get_base_api_url(self.path)
+
+
+class RemoteFunctionStep(RemoteStep):
+    """
+    Graph step implementation for invoking functions remotely.
+    :param fn: Either an mlrun.runtimes.RemoteRuntime object or
+    a string URI in the form 'function_name' or 'project_name/function_name'
+    :param project_name: Project containing the function.
```
assaf758: I'm not sure if it's a good idea to have two options to specify the project (in the `fn` or the `project` params) - which one wins if they are both specified?

mlrun/serving/remote.py
```diff
@@ -527,3 +527,86 @@ def post_init(self, mode="sync", **kwargs):
...
+    """
+
+    def __init__(
+        self,
+        fn: Union[mlrun.runtimes.RemoteRuntime, str, None] = None,
+        project_name: str = "",
+        method: str = "POST",
```
assaf758: why do you specify method param here?

assaf758 (commented):
Some initial comments inline.
Please create a storey jira for the implementation.
Since we are long past feature-freeze, flagging this as 1.11.0 feature

mlrun/serving/remote.py
daniels290813: There are three options for getting the project name (by this order) - 
1. Using the fn string (i.e passing <project_name/function_name>
2. Using the project_param
3. derive project_name from mlconf
this is to enable flexibility using this step - the user can specify which function exactly or he can just pass a function name from current project (requested by @GiladShapira94 )

mlrun/serving/remote.py
```diff
@@ -527,3 +527,86 @@ def post_init(self, mode="sync", **kwargs):
...
+        **kwargs,
+    ):
+        super().__init__(url="", method=method, **kwargs)
+        self.rundb = None
+        self.fn = fn
+        self.project_name = project_name
```
theSaarco, line 564: My assumption is that the project name is only used as reference to the function to call, not related to the actual step being generated. Therefore I would verify that `project_name` is only provided if `fn` is a string and doesn't contain a project name in it (i.e. not a URI but just the function name).

mlrun/serving/remote.py
```diff
@@ -527,3 +527,86 @@ def post_init(self, mode="sync", **kwargs):
...
+
+        if not self.fn:
+            raise mlrun.errors.MLRunRuntimeError(
+                "Parameter 'fn' can not be an empty string.\n"
+            )
+
+        self.project_name = self.project_name or mlrun.mlconf.active_project
+        if isinstance(self.fn, str):
+            try:
+                fn_split = self.fn.split("/")
```
theSaarco: This is a simplistic usage and doesn't account for tags or hash keys which may appear if the user is passing a full function uri (retrieved by using `fn.uri`). You should use the `parse_versioned_object_uri` method that knows how to fully parse these parameters, then you can pass the extracted parts to the `get_function` method.
Of course, if the user just passed a function name then the returned project will be empty, and you can use `self.project_name` in that case.

mlrun/serving/remote.py
```diff
@@ -527,3 +527,86 @@ def post_init(self, mode="sync", **kwargs):
...
+            or getattr(self.fn.status, "address", None)
+        )
+        if not url:
+            raise mlrun.errors.MLRunRuntimeError(
+                f"Could not determine the function URL for '{self.fn.metadata.name}'. \n"
+                "Make sure the function is deployed and reachable."
+            )
+
+        if not url.startswith("http"):
+            url = f"http://{url}"
```
theSaarco: The `RemoteRuntime` class has the `_resolve_invocation_url` method which does this calculation (and it also checks the status and some other things). No need to re-invent the wheel.

theSaarco (changes requested):
There are some changes needed in how you parse the function URL, and also you can use already existing method for invocation URL calculation.

mlrun/serving/remote.py
daniels290813: The thing is that `mlrun.get_run_db().get_function()` returns a dict and not RemoteRuntime - current implementation addresses both objects.
I can create a new_function using the dict and then use the `_resolve_invocation_url` - wdyt ?

mlrun/serving/remote.py
assaf758: Ok, so I guess we just need to improve the docstring to capture the above

mlrun/serving/remote.py
daniels290813: Added `RemoteRuntime.get_url()` to derive function url (uses `_resolve_invocation_url`)

mlrun/serving/remote.py
daniels290813: added the usage of `parse_versioned_object_uri(self.fn)`

mlrun/serving/remote.py
daniels290813: I've updated the usage of `project_name`

mlrun/serving/remote.py
daniels290813: Removed. thanks !

mlrun/serving/remote.py
daniels290813: I've updated the function DocString to show the correct order of using the project_name param

daniels290813:
@theSaarco @davesh0812 Can you please review

mlrun/serving/remote.py
```diff
@@ -527,3 +528,79 @@ def post_init(self, mode="sync", **kwargs):
...
+
+        if not self.fn:
+            raise mlrun.errors.MLRunRuntimeError(
+                "Parameter 'fn' can not be an empty string.\n"
+            )
+
+        if isinstance(self.fn, str):
+            project, uri, tag, hash_key = parse_versioned_object_uri(self.fn)
+
+            if self.project_name and project:
```
davesh0812, line 581: if `self.project_name == project` I don't think we should raise an error

mlrun/serving/remote.py
```diff
@@ -527,3 +528,79 @@ def post_init(self, mode="sync", **kwargs):
...
+        fn: Union[mlrun.runtimes.RemoteRuntime, str, None] = None,
+        project_name: str = "",
+        **kwargs,
+    ):
+        super().__init__(url="", **kwargs)
+        self.rundb = None
+        self.fn = fn
+        self.project_name = project_name
+
+    def post_init(self, mode="sync", **kwargs):
```
davesh0812:
```suggestion
    def post_init(self, mode="sync", **kwargs) -> None:
```

mlrun/serving/remote.py
```diff
@@ -527,3 +528,79 @@ def post_init(self, mode="sync", **kwargs):
...
+        self.rundb = mlrun.get_run_db()
+        if not isinstance(self.fn, (mlrun.runtimes.RemoteRuntime, str)):
+            raise mlrun.errors.MLRunInvalidArgumentTypeError(
+                "Parameter 'fn' must be of type mlrun.runtimes.RemoteRuntime or str."
+            )
+
+        if not self.fn:
+            raise mlrun.errors.MLRunRuntimeError(
+                "Parameter 'fn' can not be an empty string.\n"
+            )
```
davesh0812, line 576:
```suggestion
            raise mlrun.errors.MLRunRuntimeError(
                "Parameter 'fn' have to be initialized." 
            )
```

mlrun/serving/remote.py
```diff
@@ -527,3 +528,79 @@ def post_init(self, mode="sync", **kwargs):
...
+                    name=uri, project=project, tag=tag, hash_key=hash_key
+                )
+                if isinstance(self.fn, dict):
+                    self.fn = mlrun.runtimes.RemoteRuntime.from_dict(self.fn)
+
+            except mlrun.MLRunNotFoundError:
+                raise mlrun.MLRunNotFoundError(
+                    f"Cannot find function '{self.fn}' in the MLRun DB.\n"
+                    "Verify that the function URI is correct and that the function is stored properly."
+                )
```
davesh0812: I think it good enough to raise the origin `MLRunNotFoundError`.

davesh0812 (approved):
Nice.
Few comments

assaf758 (approved):

assaf758 (approved):

PR #8747, opened 2025-10-21
[TDengine]  Convert database from a class attribute to an instance attribute that's evaluated in __init__

alxtkr77:
### 📝 Description

Fixes an import-time initialization bug in `TDEngineConnector` where the `database` attribute was being set at class definition time instead of instance creation time. This caused the `system_id` to be captured as an empty string before `mlrun.mlconf` was loaded, resulting in database names like `mlrun_model_monitoring_` (with trailing underscore) instead of `mlrun_model_monitoring_<system_id>`.

**Problem**: When `TDEngineConnector` was imported before the MLRun config was loaded/reloaded, the database name would not include the system ID, even after calling `mlrun.mlconf.reload()`.

**Solution**: Move the `database` attribute initialization from class-level to instance-level in the `__init__` method, ensuring it captures the current `system_id` at the time of object creation.

---

### 🧪 Testing

**Unit Test**: Added `test_database_name_with_system_id()` which:
- Creates `TDEngineConnector` instances with a dummy DSN (no actual connection needed)
- Verifies default database name includes `mlrun.mlconf.system_id`
- Verifies explicit `database` parameter via kwargs overrides the default
- Test passes in all environments without `@pytest.mark.skipif`

**Manual Verification**: Reproduced the original issue and confirmed the fix:
```python
# Before fix: database = "mlrun_model_monitoring_"
# After fix: database = "mlrun_model_monitoring_2qhv51"
```

---

### 🔗 References
- Ticket link: ML-11285
- Related issue: Import order dependency where `system_id` was not populated in database name

---

### 🚨 Breaking Changes?

- [x] No

### DIFF
--- a/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
+++ b/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
@@ -55,7 +55,6 @@ class TDEngineConnector(TSDBConnector):
     """
 
     type: str = mm_schemas.TSDBTarget.TDEngine
-    database = f"{tdengine_schemas._MODEL_MONITORING_DATABASE}_{mlrun.mlconf.system_id}"
 
     def __init__(
         self,
@@ -70,6 +69,11 @@ class TDEngineConnector(TSDBConnector):
 
         self._timestamp_precision: Final = (  # cannot be changed after initialization
             timestamp_precision
+        )
+
+        self.database = kwargs.get(
+            "database",
+            f"{tdengine_schemas._MODEL_MONITORING_DATABASE}_{mlrun.mlconf.system_id}",
         )
 
         self._init_super_tables()
--- a/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
+++ b/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
@@ -61,7 +61,6 @@ class TDEngineConnector(TSDBConnector):
         project: str,
         profile: DatastoreProfile,
         timestamp_precision: TDEngineTimestampPrecision = TDEngineTimestampPrecision.MICROSECOND,
-        **kwargs,
     ):
         super().__init__(project=project)
 
@@ -71,10 +70,15 @@ class TDEngineConnector(TSDBConnector):
             timestamp_precision
         )
 
+        if not mlrun.mlconf.system_id:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "system_id is not set in mlrun.mlconf. "
+                "TDEngineConnector requires system_id to be configured for database name construction. "
+                "Please ensure MLRun configuration is properly loaded before creating TDEngineConnector."
+            )
         self.database = (
             f"{tdengine_schemas._MODEL_MONITORING_DATABASE}_{mlrun.mlconf.system_id}"
         )
-
         self._init_super_tables()
 
     @property
--- a/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
+++ b/mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
@@ -71,9 +71,8 @@ class TDEngineConnector(TSDBConnector):
             timestamp_precision
         )
 
-        self.database = kwargs.get(
-            "database",
-            f"{tdengine_schemas._MODEL_MONITORING_DATABASE}_{mlrun.mlconf.system_id}",
+        self.database = (
+            f"{tdengine_schemas._MODEL_MONITORING_DATABASE}_{mlrun.mlconf.system_id}"
         )
 
         self._init_super_tables()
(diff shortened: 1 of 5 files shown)


mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
```diff
@@ -72,6 +71,11 @@ def __init__(
             timestamp_precision
         )
 
+        self.database = kwargs.get(
```
assaf758: why do we need the capability to provide custom database?
I saw this is used in test-2, but i'm not sure what does test-2 tests that is relevant to this bug?

tests/model_monitoring/db/tsdb/tdengine/test_tdengine_connector.py
```diff
@@ -204,3 +205,37 @@ def test_write_application_event(
...
+    """
+    # Use a dummy DSN - we're not actually connecting to TDEngine
+    dummy_dsn = "taosws://testuser:testpass@localhost:888"
+    profile = DatastoreProfileTDEngine.from_dsn(
+        profile_name="test-profile", dsn=dummy_dsn
+    )
+
+    # Test 1: Default database name should use system_id from mlrun.mlconf (at instance creation time)
+    conn = TDEngineConnector(project, profile=profile)
+    expected_db = f"{schemas._MODEL_MONITORING_DATABASE}_{mlrun.mlconf.system_id}"
```
assaf758: Please make sure `system_id` is not an empty string

mlrun/model_monitoring/db/tsdb/tdengine/tdengine_connector.py
alxtkr77: will remove

tests/model_monitoring/db/tsdb/tdengine/test_tdengine_connector.py
alxtkr77: will do

assaf758 (approved):

tests/model_monitoring/db/tsdb/tdengine/test_tdengine_connector.py
```diff
@@ -204,3 +208,34 @@ def test_write_application_event(
...
+
+        # Attempt to create TDEngineConnector without system_id should raise error
+        with pytest.raises(
+            mlrun.errors.MLRunInvalidArgumentError, match="system_id.*not set"
+        ):
+            TDEngineConnector(project, profile=profile)
+
+    finally:
+        # Restore original system_id
+        mlrun.mlconf.system_id = original_system_id
```
jond01, line 240: I suggest monkeypatching instead of the manual try-catch and cleanup.
https://docs.pytest.org/en/stable/reference/reference.html#pytest.MonkeyPatch.setattr

For example:

```py
import mlrun
import pytest

def test_sys_id(monkeypatch: pytest.MonkeyPatch) -> None:
    monkeypatch.setattr(mlrun.mlconf, "system_id", "say")
    assert mlrun.mlconf.system_id == "say"
```

tests/model_monitoring/test_tdengine.py
```diff
@@ -551,10 +553,18 @@ def test_get_records_with_interval_query(
...
+        mlrun.mlconf.system_id = uuid.uuid4().hex
+
+        try:
+            profile = DatastoreProfileTDEngine(
+                name="mm-profile", host="localhost", port=6041, user="root"
+            )
+            yield TDEngineConnector(project="test-project", profile=profile)
+        finally:
+            # Prevent test cross-contamination
+            mlrun.mlconf.system_id = prev_mlrun_mlconf_system_id
```
jond01: Same here

tests/model_monitoring/db/tsdb/tdengine/test_tdengine_connector.py
```diff
@@ -43,16 +43,20 @@ def is_tdengine_defined() -> bool:
...
+    mlrun.mlconf.system_id = uuid.uuid4().hex
+    conn = TDEngineConnector(project, profile=profile)
+    connection = taosws.connect(connection_string)
+    drop_database(connection, conn.database)
     try:
         yield conn
     finally:
-        drop_database(connection)
+        # Prevent test cross-contamination
+        mlrun.mlconf.system_id = prev_mlrun_mlconf_system_id
```
jond01: And here

tests/model_monitoring/db/tsdb/tdengine/test_tdengine_connector.py
alxtkr77: will fix

tests/model_monitoring/test_tdengine.py
alxtkr77: will fix

tests/model_monitoring/db/tsdb/tdengine/test_tdengine_connector.py
alxtkr77: will fix

assaf758 (approved):

PR #8758, opened 2025-10-22
[Pipelines] Experiment name normalization

elbamit:
### 📝 Description

When the experiment name of a pipeline doesn't start with the project name as a prefix `{project}-{experiment}`, then `list_pipelines` doesn't show the workflow (also in UI). This is handled in the SDK side when doing `project.run()`. 
However a user-provided `experiment name` is accepted in:
- `httpdb.submit_pipeline()`
- `mlrun._run_pipeline` which is widely used in kfp tests

To keep BC, this PR moves that code logic from `project.run()` to `_create_pipeline` which is used in the endpoint.

Changing this from `project.run()` to the API shouldn't have any effect because the `name` parameter (the experiment name) is not used anywhere along the flow after being prefixed with the project name

---

### 🛠️ Changes Made

- Remove the experiment name normalization from `project.run()`
- Add an experiment name normalization in `_create_pipeline()`

---

### ✅ Checklist

---

### 🧪 Testing
  
  

Manual + added assertion in existing unit test

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11260
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -3644,6 +3644,7 @@ class MlrunProject(ModelObj):
         workflow_spec.cleanup_ttl = cleanup_ttl or workflow_spec.cleanup_ttl
         workflow_spec.run_local = local
 
+        name = f"{self.metadata.name}-{name}" if name else self.metadata.name
         artifact_path = artifact_path or self._enrich_artifact_path_with_workflow_uid()
 
         if not schedule:
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -3644,7 +3644,6 @@ class MlrunProject(ModelObj):
         workflow_spec.cleanup_ttl = cleanup_ttl or workflow_spec.cleanup_ttl
         workflow_spec.run_local = local
 
-        name = f"{self.metadata.name}-{name}" if name else self.metadata.name
         artifact_path = artifact_path or self._enrich_artifact_path_with_workflow_uid()
 
         if not schedule:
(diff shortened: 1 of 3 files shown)


server/py/services/api/tests/unit/api/test_pipelines.py
```diff
@@ -364,6 +364,7 @@ def test_create_pipeline(
     )
     response_body = response.json()
     assert response_body["id"] == "some-run-id"
+    assert response_body["name"].startswith("getting-started-tutorial-iguazio")
```
yaelgen, line 367: Add a costume experiment to the post request, that does not start with the project name, will make the test more accurate

yaelgen (changes requested):

mlrun/projects/project.py
```diff
@@ -3644,7 +3644,6 @@ def run(
         workflow_spec.cleanup_ttl = cleanup_ttl or workflow_spec.cleanup_ttl
         workflow_spec.run_local = local
 
-        name = f"{self.metadata.name}-{name}" if name else self.metadata.name
```
liranbg, line 3647: This will make the log line `Started run workflow..` down there to be wrong. why do we delete it from here?

elbamit: Just to not repeat itself here and also in the server.. But you're right that it can be wrong in case the given name is empty. I'll add it back

PR #8759, opened 2025-10-22
[Bugfix] client-side block on mounting auth/project secrets; validate env/volumes in KubeResourceSpec

adamdelman:
### 📝 Description
Block client-side use of **MLRun internal auth secrets** when:
- Setting env vars from secrets in pod runtimes (`set_env_from_secret`, `set_env(value_from=...)`).
- Configuring Azure Key Vault via `with_secrets(kind="azure_vault", source={"k8s_secret": ...})` on **Projects**, **RunTemplates**, and **Serving** runtimes.

This update **centralizes** the pattern in `mlrun.common.secrets.AUTH_SECRET_PATTERN` (built from `mlconf.secret_stores.kubernetes.auth_secret_name`), and **reuses** it across all call sites—no duplicate regex construction.

---

### 🛠️ Changes Made
- **mlrun/common/secrets.py**
  - Add config-aware, compiled regex: `AUTH_SECRET_PATTERN` (derived from `mlconf.secret_stores.kubernetes.auth_secret_name.format(hashed_access_key="")`).
  - New lightweight import of `mlrun.config` and `re`.

- **mlrun/runtimes/pod.py**
  - Guardrails:
    - `set_env_from_secret(...)` → rejects secrets matching `AUTH_SECRET_PATTERN`.
    - `set_env(..., value_from=...)` → extracts secret name from `V1EnvVarSource` **or** dict payloads (supports `valueFrom.secretKeyRef` and `value_from.secret_key_ref`, plus top-level `secretKeyRef`) and rejects if matched.
  - Helpers:
    - `_validate_no_auth_secret(secret_name: Optional[str])`
    - `_extract_secret_name_from_value_from(value_from: typing.Any) -> Optional[str>`
  - Added type hints; no behavioral change except validation.

- **mlrun/projects/project.py**
  - Docstring: add `azure_vault` to supported kinds.
  - `with_secrets(...)`: when `kind="azure_vault"`, block `source["k8s_secret"]` if it matches `AUTH_SECRET_PATTERN`.

- **mlrun/model.py (RunTemplate)**
  - `with_secrets(...)`: same `azure_vault` guard as above.

- **mlrun/runtimes/nuclio/serving.py**
  - `with_secrets(...)`: same `azure_vault` guard as above.

- **tests/**
  - **tests/projects/test_project.py**
    - Block/allow cases for `Project.with_secrets("azure_vault", ...)` using a stubbed `AzureVaultStore`.
  - **tests/serving/test_serving.py**
    - Same block/allow coverage for serving runtime; stubbed `AzureVaultStore`.
  - **tests/runtimes/test_pod.py**
    - Blocks:
      - `set_env_from_secret(...)` with forbidden secret.
      - `set_env(..., value_from=...)` with:
        - `V1EnvVarSource(secret_key_ref=...)`
        - dict `valueFrom.secretKeyRef` (camelCase)
        - dict `value_from.secret_key_ref` (snake_case)
        - top-level `secretKeyRef` dict
    - Allows:
      - Literal env values and non-secret `value_from` (e.g., ConfigMap).

---

### ✅ Checklist
- [x] Unit tests added and passing  
- [x] No server/API surface changes (client-side validation only)  
- [x] Pattern is config-driven, not hard-coded  
- [x] Clear error messaging (`MLRunInvalidArgumentError`)  
- [ ] (Docs) Note that `mlrun-auth-secrets.*` (config-derived) is reserved

---

### 🧪 Testing
**Blocked (expect `MLRunInvalidArgumentError`):**
- Any secret name matching the configured template (e.g., `mlrun-auth-secrets.*`) via:
  - `Project.with_secrets("azure_vault", {"k8s_secret": ...})`
  - `RunTemplate.with_secrets("azure_vault", {"k8s_secret": ...})`
  - `ServingRuntime.with_secrets("azure_vault", {"k8s_secret": ...})`
  - `set_env_from_secret(...)`
  - `set_env(value_from=...)` (dict & `V1EnvVarSource`)

**Allowed:**
- Regular project/K8s secrets
- ConfigMap-derived envs
- Non-secret literal env values

---

### 🔗 References
- Jira: https://iguazio.atlassian.net/browse/ML-11226

---

### 🚨 Breaking Changes?
- [ ] Yes  
- [x] No

---

### 🔍 Additional Notes
- Validation is **client-side only**, preventing unsafe exposure before submitting specs to the API.
- `with_secrets` and runtime env helpers now share a **single compiled regex** (`AUTH_SECRET_PATTERN`) to ensure consistent behavior.


mlrun/runtimes/pod.py
```diff
@@ -87,6 +87,42 @@
 }
 
 
+class _ValidatingList(list):
```
Yacouby: Some doc to explain its purpose would be useful

mlrun/runtimes/pod.py
```diff
@@ -619,6 +684,100 @@ def _prune_node_selector_requirements_from_node_selector_terms(
...
+        if spec_object:
+            nested_metadata = getattr(spec_object, "metadata", None)
+            if nested_metadata and getattr(nested_metadata, "project", None):
+                return nested_metadata.project
+        return None
+
+    @classmethod
+    def _assert_secret_allowed(
+        cls, owner: typing.Any, secret_name: str, context: str
+    ) -> None:
```
Yacouby: Add a doc summarizing its purpose and the rules for blocked secrets

mlrun/runtimes/pod.py
```diff
@@ -244,6 +283,18 @@ def __init__(
         self._termination_grace_period_seconds = None
         self.__fields_pending_discard = {}
 
+    @property
+    def env(self) -> list:
+        return self._env
+
+    @env.setter
+    def env(self, environment_variables: typing.Iterable):
```
Yacouby: doc also here

mlrun/runtimes/pod.py
```diff
@@ -255,6 +306,20 @@ def volumes(self, volumes):
             for vol in volumes:
                 set_named_item(self._volumes, vol)
 
+    @property
+    def volumes(self) -> list:
+        return list(self._volumes.values())
+
+    @volumes.setter
+    def volumes(self, volumes: typing.Iterable):
```
Yacouby: same as above

mlrun/runtimes/pod.py
```diff
@@ -619,6 +684,100 @@ def _prune_node_selector_requirements_from_node_selector_terms(
                 )
         return new_node_selector_terms
 
+    @staticmethod
+    def _get_current_project_name(owner: typing.Any) -> typing.Optional[str]:
```
Yacouby: Add a test case where the project is nested differently, e.g., missing metadata or spec.metadata entirely, to ensure it handles all possible object structures safely

tests/runtimes/test_pod_secret_blocking.py
```diff
@@ -0,0 +1,199 @@
...
+# limitations under the License.
+import types
+from typing import Optional
+
+import kubernetes.client as k8s_client
+import pytest
+
+import mlrun.errors
+from mlrun.runtimes.pod import KubeResource, KubeResourceSpec
+
```
Yacouby: Most tests assign one item at a time. what about adding a test with multiple items (mixed allowed and blocked) to verify that the first blocked secret stops the assignment correctly.

Yacouby (changes requested):
Suggested improvements + 
it is not clear from each test's name what it does so a docstring ( even one line ) can help future maintainers understand the intent quickly.

mlrun/runtimes/pod.py
```diff
@@ -619,6 +684,100 @@ def _prune_node_selector_requirements_from_node_selector_terms(
...
+                or value_from.get("secret_key_ref")
+                or value_from.get("valueFrom", {}).get("secretKeyRef")
+                or value_from.get("valueFrom", {}).get("secret_key_ref")
+            )
+            if isinstance(possible_ref, dict):
+                return possible_ref.get("name")
+        return None
+
+    @staticmethod
+    def _get_secret_name_from_volume(
```
rokatyy: all these volumes/envs and other checks are already implemented in https://github.com/mlrun/mlrun/blob/7b1424a9f14da2ef7f9ec56cab40d3d7147a9825/server/py/framework/api/utils.py#L369-L371 and may be reused if passing validating function to them or just returning the list first and validating later.

adamdelman: These funcs are used in the server side and contain assumptions that secret can be fetched from kuberenetes, whereas the code I am changing is only on the sdk/client side, and as such are not suitable for this use case

rokatyy: @adamdelman what do you mean by `secret can be fetched from kuberenetes`? None of these function fetches secrets from k8s.

mlrun/runtimes/pod.py
```diff
@@ -155,6 +234,9 @@ class KubeResourceSpec(FunctionSpec):
         "tolerations",
         "node_selector",
     ]
+    _FORBIDDEN_AUTH_SECRET_PATTERN = re.compile(r"^mlrun-auth-secrets.*$")
+    _PROJECT_SECRET_PATTERN = re.compile(r"^mlrun-project-secrets-(?P<project>.+)$")
+    _SMTP_CONFIG_SECRET = "mlrun-smtp-config"
```
rokatyy: why compiling on every object if these regexps are the same always?

mlrun/runtimes/pod.py
```diff
@@ -244,16 +328,56 @@ def __init__(
...
+
+        Wraps the provided environment variables in a `_ValidatingList` so that
+        every addition or mutation is validated through `_validate_single_env_item`.
+        Uses the spec as the owner, allowing validation logic to access contextual
+        attributes (e.g., project name) to enforce secret access rules.
+        """
+        validated_items = []
+        for single_env in list(environment_variables or []):
+            self._validate_single_env_item(single_env)
+            validated_items.append(single_env)
```
rokatyy: no need for that, already done in __init__

mlrun/runtimes/pod.py
```diff
@@ -619,6 +743,144 @@ def _prune_node_selector_requirements_from_node_selector_terms(
...
+                f"mounting auth-secret is not allowed on the client side"
+            )
+
+        project_match = self._PROJECT_SECRET_PATTERN.match(secret_name)
+        if not project_match:
+            return
+
+        target_project_name = project_match.group("project")
+
+        if not current_project_name or target_project_name != current_project_name:
```
rokatyy: if you got here, `current_project_name` is not empty

mlrun/runtimes/pod.py
```diff
@@ -619,6 +743,144 @@ def _prune_node_selector_requirements_from_node_selector_terms(
...
+            if isinstance(secret_dict, dict):
+                return secret_dict.get("secretName") or secret_dict.get("secret_name")
+        return None
+
+    def _validate_single_env_item(self, env_item: typing.Any) -> None:
+        try:
+            value_from = (
+                getattr(env_item, "value_from", None)
+                if not isinstance(env_item, dict)
+                else env_item.get("valueFrom") or env_item.get("value_from")
```
rokatyy: you can use `get_in` for part of this check (though case difference will require double check anyway, eg for `valueFrom` and `value_from`)

mlrun/runtimes/pod.py
```diff
@@ -619,6 +743,144 @@ def _prune_node_selector_requirements_from_node_selector_terms(
                 )
         return new_node_selector_terms
 
+    def _assert_secret_allowed(
+        self,
+        owner: typing.Any,
+        secret_name: str,
+        context: str,
```
rokatyy: I didn't find a place where you would pass owner. For instance, above you do `self._assert_secret_allowed(self, secret_name, "spec.volumes")`. I suppose secret name is not owner, right? And one more moment - it's better to pass named args

rokatyy (changes requested):
Questions and suggestions below

mlrun/runtimes/pod.py
liranbg: move regex to `utils/regex.py` and these keys must be taken from config.py. you can take the value as a prefix where template is empty. e.g
`X-Y-{zz}` regex would be `X-Y-`

mlrun/runtimes/pod.py
```diff
@@ -774,6 +1036,7 @@ def set_envs(
         if (not env_vars and not file_path) or (env_vars and file_path):
             raise mlrun.errors.MLRunInvalidArgumentError(
                 "must specify env_vars OR file_path"
+                "must specify env_vars OR file_path"
```
liranbg: remove duplicate

mlrun/runtimes/pod.py
```diff
@@ -244,16 +328,56 @@ def __init__(
...
+        Set environment variables with per-item validation.
+
+        Wraps the provided environment variables in a `_ValidatingList` so that
+        every addition or mutation is validated through `_validate_single_env_item`.
+        Uses the spec as the owner, allowing validation logic to access contextual
+        attributes (e.g., project name) to enforce secret access rules.
+        """
+        validated_items = []
+        for single_env in list(environment_variables or []):
+            self._validate_single_env_item(single_env)
```
liranbg: why validation single env *and* validating the entire list?

liranbg (commented):
intermediate review

mlrun/runtimes/pod.py
```diff
@@ -619,6 +743,144 @@ def _prune_node_selector_requirements_from_node_selector_terms(
...
+            or secret_name == self._SMTP_CONFIG_SECRET
+        ):
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"Blocking secret '{secret_name}' in {context}: "
+                f"mounting auth-secret is not allowed on the client side"
+            )
+
+        project_match = self._PROJECT_SECRET_PATTERN.match(secret_name)
+        if not project_match:
+            return
```
liranbg: why do we handle project secret validation here? lets focus on auth secret validations only
adding project to spec causes huge side effect on resources

mlrun/runtimes/pod.py
```diff
@@ -708,19 +716,54 @@ def spec(self) -> KubeResourceSpec:
...
+                    or value_from
+                )
+                sk = (vf or {}).get("secretKeyRef") or (vf or {}).get("secret_key_ref")
+                if isinstance(sk, dict):
+                    secret_name = sk.get("name")
+
+            if secret_name and _AUTH_SECRET_PATTERN.match(secret_name):
+                raise mlrun.errors.MLRunInvalidArgumentError(
+                    f"Forbidden secret '{secret_name}' matches MLRun auth-secret pattern."
+                )
```
rokatyy: this is the same code as above, could be a separate function

rokatyy (commented):
Looks good! One comment regarding code duplication and there is also one case that is not covered from the client side. Namely, it is also possible to mount a secret via
```
function.apply(mlrun.runtimes.mounts.mount_secret("mlrun-project-secrets-secrets-test-new", "./secrets/"))
```

tests/runtimes/test_pod_env_auth_secret_block.py
liranbg, line 1: name the file according to the file it tests, not the functionality it tests

tests/projects/test_project_with_secrets_auth_block.py
liranbg, line 1: name the file according to the file it tests, not the functionality it tests

mlrun/runtimes/pod.py
```diff
@@ -1366,6 +1401,35 @@ def print_log(_text):
...
+        if isinstance(value_from, k8s_client.V1EnvVarSource):
+            if value_from.secret_key_ref:
+                return value_from.secret_key_ref.name
+        elif isinstance(value_from, dict):
+            vf = (
+                value_from.get("valueFrom")
+                or value_from.get("value_from")
+                or value_from
+            )
+            sk = (vf or {}).get("secretKeyRef") or (vf or {}).get("secret_key_ref")
```
liranbg: use full variable names

liranbg (commented):
nicely KISSed. minor comments

liranbg (approved):

mlrun/runtimes/pod.py
```diff
@@ -86,6 +87,15 @@
     "driver_security_context": sanitized_types["security_context"],
 }
 
+_AUTH_SECRET_PATTERN = re.compile(
+    re.escape(
```
theSaarco: Would be best to start this with `^` to prevent situations where it's not the full name of the secret. For example, this will match `not-an-mlrun-auth-secrets.blablabla`.

theSaarco (changes requested):
This is not complete - note that there are implementations of `with_secrets` also for `RunTemplate` and `ServingRuntime` (this may be a good opportunity to unify their implementation as well - since they were very simple this was not done before).
I would be also very happy if we could block project-secrets while we're at it, though it's not part of the MVP here.

liranbg (approved):

rokatyy (changes requested):
`mlrun.runtimes.mounts.mount_secret`comment from above is still relevant

adamdelman:
@rokatyy This PR is very far from fool proof.
Since there are very many ways to bypass the rather feeble protections we are adding here, and `mlrun.runtimes.mounts.mount_secret` was not defined as part of the required changes for this ticket, I don't want to start chasing the many ways it can be bypassed.

theSaarco (approved):
LGTM 👍

rokatyy (approved):

PR #8763, opened 2025-10-23
[Model Monitoring] Fix test_batch_drift

tomerm-iguazio:
### 📝 Description

The `test_batch_drift` test was broken because the timestamp in `infer_df` was older than first_request.
 As a result, none of the events in the batch were processed since they didn’t pass the timestamp filters (they did not fit the timestamp window).
 
That's happen because the first_request time determined in this case in `_generate_model_endpoint`, which set it as default to datetime.now.

---

### 🛠️ Changes Made
Test updated to use additional 20 seconds to include these events in the controller window.

---

### ✅ Checklist

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11276
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/tests/system/model_monitoring/test_model_monitoring.py
+++ b/tests/system/model_monitoring/test_model_monitoring.py
@@ -1551,10 +1551,11 @@ class TestBatchDrift(TestMLRunSystemModelMonitoring):
                 "p0": [0, 0],
             }
         )
+        # Add 20 seconds because the batch window is determined using datetime.now later in _generate_model_endpoint
+        # ML-11276
         infer_results_df[mlrun.common.schemas.EventFieldType.TIMESTAMP] = (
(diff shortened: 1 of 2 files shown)


Eyal-Danieli (approved):

tests/system/model_monitoring/test_model_monitoring.py
```diff
@@ -1599,7 +1601,7 @@ def test_batch_drift(self):
         )
 
         # Wait for the controller, app and writer to complete
-        sleep(180)
+        sleep(300)
```
assaf758, line 1604: why do you need to increase the sleep?

PR #8774, opened 2025-10-26
[BackGroundTask] Prevent duplicate bg labels by updating unique constraint

Yacouby:
### 📝 Description
This PR Fixes a race condition in `BackgroundTaskLabel` creation caused by an incorrect unique constraint definition.

Previously, the unique constraint used (task_id, name), allowing duplicate labels with the same name and value when multiple identical background tasks (e.g., terminate pipeline requests) were created concurrently.

This PR updates the unique constraint to ensure label uniqueness across (project, name, value) combinations.

---

### 🛠️ Changes Made

- Updated the `BackgroundTaskLabel` table unique constraint to use (project, name, value)
- Added a database migration to apply the new constraint
- Refactored background task creation logic:
Labels are now appended directly to `background_task_record.labels `before adding the task to the session, instead of being collected separately and upserted together.
This improves clarity and leverages proper SQLAlchemy relationship handling.

---

### ✅ Checklist

---

### 🧪 Testing

- Added a new DB unit test to verify that duplicate labels (same project/name/value) are not created
- Verified that adding labels with different values or in different projects works as expected
- Triggered migration and validated it applied without errors.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11299
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


server/py/services/api/tests/unit/db/test_background_tasks.py
```diff
@@ -327,3 +328,87 @@ def test_cleanup_old_background_tasks(self):
...
+    def _get_labels(self, project, name, value):
+        """
+        Helper function to query BackgroundTaskLabel by project, name, and value.
+        Returns a list of matching labels.
+        """
+        return (
+            self._db_session.query(framework.db.sqldb.models.BackgroundTaskLabel)
+            .filter_by(project=project, name=name, value=value)
+            .all()
+        )
```
TomerShor: Move to the bottom, as it is only used in `_store_task_and_assert_label`

server/py/services/api/tests/unit/db/test_background_tasks.py
```diff
@@ -327,3 +328,87 @@ def test_cleanup_old_background_tasks(self):
...
         assert len(remaining_names) == 1
+
+    def test_background_task_label_unique_constraint(self):
+        """
+        Test the unique constraint behavior for BackgroundTaskLabel.
+
+        This test ensures:
+        1. A label can be added to a task.
+        2. Adding the same label again does not create a duplicate row.
+        3. A different label for the same task can be added.
```
TomerShor: this case is missing from the test

TomerShor (changes requested):

server/py/services/api/tests/unit/db/test_background_tasks.py
Yacouby: old doc. updated it

TomerShor: I would actually want to see this case being tested too

Yacouby: Actually this case isn’t possible given the current logic. once the background task exists, labels aren’t re-added or updated in subsequent calls. The label creation happens only during the initial insert (when no record exists yet).

TomerShor: Interesting.. we might need to address this but not in this PR for sure.

TomerShor (approved):

server/py/framework/db/sqldb/models.py
```diff
@@ -530,14 +529,14 @@ class BackgroundTaskLabel(Base):
             nullable=False,
             index=True,
         )
+        project = Column(framework.db.sqldb.sql_types.Utf8BinText, nullable=False)
```
liranbg, line 534: why not making it a foreign key? that way, when deleting a project - it will delete those label tasks automatically

Yacouby: @liranbg Aren’t these labels already deleted when the parent task is deleted (through the task_id foreign key cascade)?

liranbg: make sense (tho not optimal for deleting project and its sub resources - for that we will have a bigger-scoped ticket)

liranbg (approved):

PR #8796, opened 2025-10-29
[Secrets] Remove labels param when reading secret in store_secrets() [feature/ig4-authentication]

elbamit:
### 📝 Description

As part of https://github.com/mlrun/mlrun/pull/8484, `read_secret()` was refactored to (optionally) check for a label in the k8s secret. This is needed for IG4's secret tokens.
However, the labels parameter is not needed in `store_secrets` function (k8s.py). 

When running a workflow, the secret is initially created in k8s with the label `mlrun/username:admin` (for the admin user). However, when resolving KFP secret environment variables to k8s secrets, `store_secrets` is called with the label `mlrun/username:MLRUN_AUTH_SESSION`. Using this label with `read_secret` results in None being returned instead of the actual secret.

---

### 🛠️ Changes Made

- Remove `label` param when calling `read_secret` in `store_secrets`

---

### ✅ Checklist

---

### 🧪 Testing
  
  

Regression test that was failing now works when run manually

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11330
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/server/py/services/api/tests/unit/utils/singletons/test_k8s_utils.py
+++ b/server/py/services/api/tests/unit/utils/singletons/test_k8s_utils.py
@@ -279,6 +279,23 @@ def test_store_secret(
         assert data == secrets_to_store
 
 
+# Labels param in read_secret should be used only with IG4 secrets
+def test_store_secrets_does_not_pass_labels_to_read_secret(k8s_helper):
+    k8s_helper.read_secret = mock.MagicMock(
+        side_effect=k8s_dynamic_exceptions.NotFoundError(
+            k8s_client_rest.ApiException(status=404)
+        )
+    )
+    k8s_helper.store_secrets(
+        secret_name="my-secret",
+        secrets={"key1": "value1"},
+        namespace="default",
+    )
+    # Ensure 'labels' is not in the call arguments
+    for call in k8s_helper.read_secret.call_args_list:
+        assert "labels" not in call.kwargs
+
+
 @pytest.mark.parametrize(
     "k8s_secret_data, secrets_data, expected_action, expected_secret_data",
     [
--- a/server/py/services/api/tests/unit/utils/singletons/test_k8s_utils.py
+++ b/server/py/services/api/tests/unit/utils/singletons/test_k8s_utils.py
@@ -279,8 +279,12 @@ def test_store_secret(
         assert data == secrets_to_store
 
 
-# Labels param in read_secret should be used only with IG4 secrets
+
 def test_store_secrets_does_not_pass_labels_to_read_secret(k8s_helper):
+    """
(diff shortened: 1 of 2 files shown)


moranbental (approved):

liranbg (approved):

server/py/services/api/tests/unit/utils/singletons/test_k8s_utils.py
```diff
@@ -279,6 +279,23 @@ def test_store_secret(
         assert data == secrets_to_store
 
 
+# Labels param in read_secret should be used only with IG4 secrets
+def test_store_secrets_does_not_pass_labels_to_read_secret(k8s_helper):
```
liranbg: move comment to test func doc so we can see it nicely on test output when it fails
"Test ensures that ..."

server/py/services/api/tests/unit/utils/singletons/test_k8s_utils.py
```diff
@@ -279,6 +279,27 @@ def test_store_secret(
...
+        )
+    )
+    k8s_helper.store_secrets(
+        secret_name="my-secret",
+        secrets={"key1": "value1"},
+        namespace="default",
+    )
+    # Ensure 'labels' is not in the call arguments
+    for call in k8s_helper.read_secret.call_args_list:
+        assert "labels" not in call.kwargs
```
Yacouby: can be simplified with a direct call argument assertion:
```suggestion
 k8s_helper.read_secret.assert_called_once_with(secret_name="my-secret", namespace="default")
```

server/py/services/api/tests/unit/utils/singletons/test_k8s_utils.py
```diff
@@ -279,6 +279,27 @@ def test_store_secret(
         assert data == secrets_to_store
 
 
+
+def test_store_secrets_does_not_pass_labels_to_read_secret(k8s_helper):
```
Yacouby: test name too long. if you write in the comment of the test what it does, then IMO it can be shortened, something like:
```suggestion
def test_no_labels_in_read_secret(k8s_helper):
```

Yacouby (changes requested):
Few suggestions to improve readability.

server/py/services/api/tests/unit/utils/singletons/test_k8s_utils.py
Yacouby: or `test_store_secrets_ignores_labels`

liranbg (approved):

moranbental (approved):

Yacouby (approved):

PR #8802, opened 2025-10-30
[Tools] Improve setup_local_ce.py script RC versions support and image tag selection

adamdelman:
### Description

Refactors and hardens the local CE installer to improve version discovery, default image selection, and operational robustness. The script now uses `semver` for GitHub tag ordering (with optional RC support), prefers chart-pinned image tags by default, exposes explicit CLI switches for version policy, and strengthens Helm/Kubernetes flows and debug output.

---

#### Version discovery and policy

- Added `semver`-based parsing and ordering for tags, including normalization of prerelease labels (`rc38 → rc.38`) to ensure correct comparisons.
- Introduced `VersionSource` and `GitHubApiPath` enums and centralized GitHub access in `_github_get_json(...)` with optional token auth.
- New resolvers:
  - `resolve_from_releases(...)` – highest semantic release (optionally includes prereleases).
  - `resolve_from_tags(...)` – tag-based fallback.
  - `choose_first_matching_version(...)` – stable-only vs. stable+RC enforcement.
- Extended `get_latest_valid_version(...)` with:
  - `allow_dev_versions: bool` – enables RC selection.
  - `version_source: "auto" | "release" | "tag"` – controls and orders source preference.

#### Chart defaults vs. “latest”

- Added `_read_chart_image_versions(...)` to read MLRun/Nuclio tags from chart values (cleaning leading `v` and architecture suffices).
- Chart-pinned tags are now preferred by default; GitHub lookup is optional via flag.
- `setup_ce(...)` / `install_ce(...)` now auto-resolve missing CE/MLRun/Nuclio versions using chart values or GitHub, respecting `--dev-versions`.

#### CLI and UX

- New flags:
  - `--dev-versions` – allow RC tags.
  - `--use-latest-versions` – ignore chart tags and query GitHub.
- Migrated all paths to `pathlib.Path` and strengthened typing across helpers.
- Improved debug logging and docstrings for `echo_color(...)`, `run_command(...)`, and related helpers.

#### Command execution, Helm, and Kubernetes

- `run_command(...)` now:
  - Auto-adds `--debug` to Helm commands when debug mode is on.
  - Prints combined stdout+stderr under `[COMMAND OUTPUT]` on failures before raising/warning.
- Improved Helm robustness:
  - Defensive repo handling in `ensure_helm_repo(...)` and `update_helm_repos(...)`.
  - Debug-aware `helm status` during ingress setup.
- Namespace and `/etc/hosts` safety:
  - `clear_namespaces(...)` skips missing namespaces instead of erroring.
  - `/etc/hosts` updates now back up the file and apply changes via a temp file swap.

#### Image upgrade flow

- `upgrade_images(...)`:
  - Reads chart defaults first when `use_chart_versions=True`.
  - Falls back to GitHub when chart values are missing or `--use-latest-versions` is set.
  - Removes leading `v` from MLRun/Jupyter tags before applying chart values.
  - Only sets Helm image keys when values exist (avoids overriding chart defaults).
  - Rebuilds dependencies before applying upgrades.

- `install_ce(...)` orchestrates the entire workflow with the new version-policy logic and improved logging.

---

### Testing

- Verified namespace clearing skips non-existent namespaces safely.
- Confirmed prerelease ordering (`1.10.0-rc38` > `1.9.2`) when `--dev-versions` is used.
- Default install uses chart-pinned MLRun/Nuclio tags.
- `--use-latest-versions` correctly pulls versions from GitHub with fallback logic.
- Failure paths surface combined command output via `[COMMAND OUTPUT]`.

---

### Breaking Changes

- None intended; version selection may behave more strictly or more predictably depending on tag formats.

https://iguazio.atlassian.net/browse/ML-11621


automation/deployment/setup_local_ce.py
```diff
@@ -511,9 +558,11 @@ def upgrade_images(
         debug=debug,
     )
 
-    # nuclio maps x86_64 to amd64
-    if arch == "x86_64":
-        arch = "amd64"
+    # Always use arm64 for Nuclio images regardless of provided arch
+    nuclio_arch = "arm64"
```
liranbg: why

automation/deployment/setup_local_ce.py
```diff
@@ -511,9 +558,11 @@ def upgrade_images(
...
+
+    # For jupyter and log-collector images, drop a leading 'v' if provided
```
liranbg:
```suggestion
    # For mlrun images, drop a leading 'v' if provided
```

automation/deployment/setup_local_ce.py
```diff
@@ -324,17 +330,58 @@ def get_all_tags(repository: str):
...
+        patch = int(m.group("patch"))
+        rc_str = m.group("rc")
+        is_final = 1 if rc_str is None else 0
+        rc_num = float("inf") if rc_str is None else int(rc_str)
+        cleaned = f"{maj}.{min_}.{patch}" + (
+            f"-rc{rc_str}" if rc_str is not None else ""
+        )
+        return (maj, min_, patch, is_final, rc_num), cleaned
+
+    candidates: list[tuple[tuple[int, int, int, int, int], str]] = []
```
liranbg: use namedtuple, it would be better

automation/deployment/setup_local_ce.py
```diff
@@ -324,17 +330,58 @@ def get_all_tags(repository: str):
...
-    raise ValueError(f"No semver tag found for {repo_name}")
+        name = tag.get("name", "")
+        parsed = parse_tag(name)
+        if parsed:
+            candidates.append(parsed)
+
+    if not candidates:
+        raise ValueError(f"No semver tag found for {repo_name}")
+
+    best_key, best_cleaned = max(candidates, key=lambda t: t[0])
```
liranbg: what do you max on? the major? if all versions are [(1,1,0), (1,2,0)] how will it take the second item?

automation/deployment/setup_local_ce.py
```diff
@@ -324,17 +330,58 @@ def get_all_tags(repository: str):
 
 
 def get_latest_valid_version(repo_name: str) -> str:
```
liranbg: I dont understand the definition of "latest valid"). I think we need to condition these 2 flows into
1. by default, get the latest GA versions and use them
2. if provided with `--devel` or so flag, take latest-greatest RC available

automation/deployment/setup_local_ce.py
```diff
@@ -271,89 +336,363 @@ def setup_registry_secret(
...
+
+
+def get_latest_release_tag(
+    repository: str,
+    include_prereleases: bool,
+) -> str | None:
+    """
+    Return the *greatest semantic* release tag, optionally excluding prereleases.
+    """
+    releases = get_all_releases(repository)
```
liranbg, line 563: why do you need to get all_releases and not get the release marked as latest?

liranbg (commented):
still relevant - https://github.com/mlrun/mlrun/pull/8802/files#r2476835951

automation/deployment/setup_local_ce.py
adamdelman: /releases/latest only returns the most recent stable release and skips RCs or newer tags, so we fetch all releases to pick the true latest by semver.

liranbg (approved):

PR #8808, opened 2025-10-30
[DB] Add MySQL indexes to optimize artifact queries

adamdelman:
### 📝 Description
Add new MySQL indexes to improve performance of artifact-related queries:
- **Artifact listing path:** `(project, best_iteration, kind, updated DESC, id DESC)`  
  Optimizes `list_artifacts()` by reducing sort cost and leveraging index order.
- **Latest tag lookup:** `(name, obj_id)`  
  Speeds up lookups of `"latest"` artifacts by object reference.

---

### 🛠️ Changes Made
- **alembic/versions/a123b456c789_add_artifact_indexes.py**
  - Added:
    - `idx_project_bi_updated` → `artifacts_v2 (project, best_iteration, kind, updated DESC, id DESC)`
    - `idx_artifacts_tags_name_obj` → `artifacts_v2_tags (name, obj_id)`
  - MySQL-safe migration (no `IF NOT EXISTS`; assumes clean state)
  - Included downgrade that drops both indexes

---

### ✅ Checklist
- [] I updated the documentation (if applicable)

---

### 🧪 Testing
- Verified improved query plans for:
  - `list_artifacts()` with project/kind filters + ORDER BY updated DESC
  - Tag lookups by `(name='latest', obj_id=...)`

---

### 🔗 References
- Ticket link: [ML-11045](https://iguazio.atlassian.net/browse/ML-11045)

---

### 🚨 Breaking Changes?
- [ ] Yes  
- [x] No

---

### 🧭 Additional Notes
Indexes align with MLRun’s most frequent artifact query patterns, reducing full-table scans and filesorts on large datasets.

[ML-11045]: https://iguazio.atlassian.net/browse/ML-11045


server/py/framework/db/sqldb/models.py
```diff
@@ -366,7 +366,14 @@ class ArtifactV2(
...
             # Used explicitly in list_artifacts, as most of the queries request best_iteration, and all always sort by
             # updated. See https://iguazio.atlassian.net/browse/ML-9189
-            Index("idx_project_bi_updated", "project", "best_iteration", "updated"),
+            Index(
+                "idx_project_bi_updated",
+                "project",
+                "best_iteration",
+                "kind",
+                text("updated DESC"),
+                text("id DESC"),
```
Yacouby: The Alembic migration currently creates this index without the DESC part. 
So you either drop the DESC here or need to update the migration.

Also, can you clarify what adding DESC improves in practice? From sql side, it can usually scan indexes in reverse, so I’m wondering if this gives any real benefit

Yacouby (changes requested):

liranbg (approved):

Yacouby (approved):

theSaarco (changes requested):
I've discussed this with @liranbg - I agree the artifacts index is critical to improving the performance, and was missing the `kind`. OTOH, I don't think the artifacts-tag idx is very helpful in this scenario, and IMO the correct way to handle this is add another filter to the query on `artifacts_v2_tags.project` (in addition to the project filter on the artifacts table). This will utilize the existing index on the artifact-tags table, and will actually improve the performance more since it will enable the join to happen on smaller row-sets (since currently it extract all the tags with `latest` from all projects, which is not optimal).
Another concern would be that creating the new index will make the upgrade flow longer - in large DBs this can take a long time. We should test this on a large DB to understand the expected time to complete (hopefully no longer than 30min), and inform support about this.

PR #8823, opened 2025-11-02
[User secret] Validate stored token belongs to authenticated user [feature/ig4-authentication]

elbamit:
### 📝 Description

During `store_secret_tokens` CRUD function, validate that each token's `sub` is equal to the authenticated user's id. This will prevent from user A to store user B's secret token.

---

### 🛠️ Changes Made

- In `store_secret_tokens` crud function, `_validate_token_name` has been refactored to `_validate_token_name_and_user` that checks if the token's sub = authenticated user id 

---

### ✅ Checklist

---

### 🧪 Testing
  
  
Unit test for `_validate_secret_token_name`
Added a new unit test for `store_secret_tokens` for when authenticated user is not equal to token's sub
Refactored existing unit tests for `store_secret_tokens` to accept auth_info instead of just username

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11318
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


server/py/services/api/crud/secrets.py
```diff
@@ -596,7 +598,21 @@ def revoke_secret_token(
...
+    ):
+        token_name = secret_token.name
+        token_sub = Secrets._decode_offline_token(token_name, secret_token.token).get(
+            "sub"
+        )
+
+        if token_sub != authenticated_id:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"Offline token '{token_name}' subject '{token_sub}' does not match the authenticated user ID. "
+                "Stored tokens can only belong to the authenticated user."
```
liranbg: I dont think that revealing the subject makes sense from client POV. you may log it with "warning" and raise the exception

liranbg (commented):
minor comment

server/py/services/api/crud/secrets.py
elbamit: When you say log with "warning" and then raise the exception, do you mean like this:

```suggestion
            if token_sub != authenticated_id:
               mlrun.utils.logger.warning(f"Offline token '{token_name}' subject {token_sub} does not match the authenticated user ID: {authenticated_id}.")
                raise mlrun.errors.MLRunInvalidArgumentError(
                    f"Offline token '{token_name}' does not match the authenticated user ID. "
                    "Stored tokens can only belong to the authenticated user."
            )
```

elbamit: @liranbg Please do not commit this suggestion as it will require changes in some tests

server/py/services/api/crud/secrets.py
```diff
@@ -596,7 +598,21 @@ def revoke_secret_token(
...
 
     @staticmethod
-    def _validate_token_name(token_name: str, seen_names: set):
+    def _validate_token_name_and_user(
+        secret_token: mlrun.common.schemas.SecretToken,
+        seen_names: set,
+        authenticated_id: str,
+    ):
+        token_name = secret_token.name
+        token_sub = Secrets._decode_offline_token(token_name, secret_token.token).get(
```
moranbental: If you decode the token here, there’s no need to decode it again in `_extract_and_validate_expiration`. You can decode the token once at the beginning, extract both the sub and exp fields, and fail early before sending the request to Iguazio.

moranbental (changes requested):

server/py/services/api/crud/secrets.py
elbamit: @liranbg I've committed this change for now. Let me know if you think it needs to be different

server/py/services/api/crud/secrets.py
elbamit: I've changed the function flow to perform a single jwt decoding, then all the required validations

server/py/services/api/crud/secrets.py
```diff
@@ -595,22 +593,55 @@ def revoke_secret_token(
...
+                        f" {authenticated_id}."
+                    )
+                    raise mlrun.errors.MLRunInvalidArgumentError(
+                        f"Offline token '{token_name}' does not match the authenticated user ID. "
+                        "Stored tokens can only belong to the authenticated user."
+                    )
+
+                # Store token info
+                token_values[secret_token.name] = {
+                    "token_sub": decoded_token.get("sub"),
```
moranbental: Why do you need to return `token_sub`? I don’t see it being used anywhere

server/py/services/api/crud/secrets.py
```diff
@@ -595,22 +593,55 @@ def revoke_secret_token(
...
+                        f"Offline token '{token_name}' is missing the 'sub' (subject) claim"
+                    )
+
+                # Validate token belongs to the authenticated user
+                token_sub = decoded_token.get("sub")
+                if token_sub != authenticated_id:
+                    mlrun.utils.logger.warning(
+                        f"Offline token '{token_name}' subject {token_sub} does not match the authenticated user ID:"
+                        f" {authenticated_id}."
+                    )
```
moranbental, line 630: use logger.warning args
```suggestion
                    logger.warning(
                        "Offline token subject does not match the authenticated user", token_name=token_name, token_sub=token_sub, user_id=authenticated_id
                    )
```

server/py/services/api/tests/unit/crud/test_secrets.py
```diff
@@ -744,16 +776,18 @@ def test_store_secret_tokens_invalid_offline_token_jwt_decode(mock_iguazio_clien
...
 
 @pytest.mark.parametrize(
     "payload",
     [
-        {"sub": "user-123"},  # missing exp
-        {"sub": "user-123", "exp": None},  # exp is None
-        {"sub": "user-123", "exp": ""},  # exp is empty
+        {"sub": "user-id-123"},  # missing exp
+        {"sub": "user-id-123", "exp": None},  # exp is None
+        {"sub": "user-id-123", "exp": ""},  # exp is empty
```
moranbental: What about checking here also if sub is missing or None or empty?

moranbental (changes requested):

server/py/services/api/tests/unit/crud/test_secrets.py
```diff
@@ -819,11 +874,142 @@ def test_store_secret_tokens_refresh_access_tokens_failure(mock_iguazio_client):
...
+        #     {
+        #         "token_name": "token2",
+        #         "token_payload": {"sub": "user-123", "exp": 9999999999},
+        #     },
+        #     True,
+        #     "Offline token 'token1' is missing the 'sub' (subject) claim",
+        #     None,
+        #     None,
+        #     "user-123",
+        # ),
```
moranbental: ?

moranbental (changes requested):

server/py/services/api/tests/unit/crud/test_secrets.py
elbamit: forgot to uncomment

moranbental (approved):
Great job!

liranbg (approved):

PR #8824, opened 2025-11-02
[IGTokenProvider] Change error message on invalid offline token fetching [feature/ig4-authentication]

elbamit:
### 📝 Description

Change the error message flow when the process of fetching an offline token fails. 
Currently the flow contains the following issues:
- Warnings with retry attempts {0/0}:

> 2025-11-05 14:43:17,801 [warning] Attempt {0/ 0} failed with exception: 401 Client Error: Unauthorized for url: https://igz-api.amite.vmdev44ig4.lab.iguazeng.com/api/v1/refresh-access-token
> 2025-11-05 14:43:17,801 [warning] Max retries reached, failed to fetch token: {"error":"401 Client Error: Unauthorized for url: https://igz-api.amite.vmdev44ig4.lab.iguazeng.com/api/v1/refresh-access-token"}

This happens because the `max_retries` is initialized during the IGTokenProvider however the `refresh_token_if_needed` function is called during it's parent class init - `DynamicTokenProvider`

- Single traceback with the following error message:
> mlrun.errors.MLRunRuntimeError: Failed to fetch access token, no token available after fetch

The error message is not informative enough, and we have to rely on the error message displayed in the warning.

The PR fixes the number of retry attempts (3 attempts total) + changes the displayed error message and chain errors so they are displayed as:

>....
>mlrun.errors.MLRunUnauthorizedError: 401 Client Error: Unauthorized for url: https://igz-api.amite.vmdev44ig4.lab.iguazeng.com/api/v1/refresh-access-token
>The above exception was the direct cause of the following exception:
....
>mlrun.errors.MLRunRuntimeError: Failed to fetch a valid access token. Authentication procedure stopped.

---

### 🛠️ Changes Made

- Change how we set `DynamicTokenProvider` max_retries value
- Refactor `_refresh_token_if_needed` to display chained error messages

---

### ✅ Checklist

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11266
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [x] Yes (explain below)
- [] No

`OAuthClientIDTokenProvider` which is used in Platform MCK relied on the error getting raised by `_post_fetch_hook`, which wasn't implemented for `OAuthClientIDTokenProvider`. 
So an invalid token sent in Platform MCK will get errored in the client side before reaching the API.

---


moranbental (dismissed):

mlrun/auth/providers.py
```diff
@@ -370,18 +373,17 @@ def _parse_response(self, response_data):
...
         if self._token and not self._is_token_valid(cleanup_if_expired=True):
             logger.warning(
                 "Failed to fetch a new token. Using the existing token, which remains valid but is close to expiring."
             )
 
-        if not self._token:
-            raise mlrun.errors.MLRunRuntimeError(
-                "Failed to fetch access token, no token available after fetch"
-            )
-
```
rokatyy, line 390: I'd still leave it here, just to make sure token is not empty after fetching

tests/rundb/test_httpdb.py
```diff
@@ -442,17 +442,12 @@ def test_client_id_auth(requests_mock: requests_mock_package.Mocker, monkeypatch
...
     time.sleep(2)
     requests_mock.reset_mock()
 
-    db.trigger_migrations()
-    assert len(requests_mock.request_history) == 2
-    assert (
-        mlrun.common.schemas.HeaderNames.authorization
-        not in requests_mock.last_request.headers
-    )
-    assert db.token_provider.get_token() is None
```
rokatyy, line 455: Looks like a flow change in `trigger_migrations`. Are you sure that it won't affect any existing flow?

rokatyy (commented):
Some questions/suggestions

tests/rundb/test_httpdb.py
elbamit: @theSaarco Can you give your input on that?

mlrun/auth/providers.py
elbamit: Got it.
After offline discussion we agreed on leaving it there as a secondary validation in case token fetch fails but doesn't raise an error

tests/auth/test_auth_providers.py
```diff
@@ -170,16 +170,6 @@ def test_post_fetch_hook_warns_near_expiry(monkeypatch, encoded_jwt_token):
     assert provider._token == token
 
 
-def test_post_fetch_hook_raises_if_no_token(monkeypatch):
```
moranbental, line 173: If you remove this test, you need to create another test to cover the case where fetching the token fails, the token is None, and an error is raised

tests/rundb/test_httpdb.py
moranbental: @rokatyy From what I understand, it’s not really related to `trigger_migrations`. It can occur in any api call, which will now raise an error on the sdk side if the token is expired, instead of on the api side as before

moranbental (changes requested):

tests/rundb/test_httpdb.py
theSaarco: This test is indeed not related to a specific API call (there's a comment in the test explaining we use the trigger-migrations flow since it only does a single API call and is easy to mock).
What I'm not sure I understand is why do the changes added really change the error flow, and why do we now fail on client-side rather than go without authn like previously done. If this is a behavior that is related to the IG flow, then why is this test (which uses client-id auth) affected?

theSaarco (commented):
Added my note on the discussion around the test - I'm not sure why this test is affected.
Also, note that you're looking to merge this into `development` rather than the feature branch - is this intended?

tests/rundb/test_httpdb.py
elbamit: @theSaarco Because the introduced change is in the parent class of both `IGTokenProvider` and `OAuthClientIDTokenProvider`.
Is it required for client-id auth to be able to send requests without authentication? I mean the request will eventually fail either way

theSaarco: There is no requirement as such. This flow is a fallback which relies on the BE to fail the call if needed. In our case it's not really happening, but there are situations where not all endpoints need authentication, so in terms of logic the thinking was to not fail immediately if no auth exists. 
Since we require auth for everything then I guess there is no harm in failing on client side, but it is a side-effect that needs to be tested once the change is introduced (for example in PMCK).

elbamit: Yes that's why I've wrote about it in breaking change of the PR description

tests/auth/test_auth_providers.py
elbamit: Test has been restored

moranbental (approved):
LGTM

rokatyy (approved):

PR #8840, opened 2025-11-04
[Runs] Enable authentication for jobs [feature/ig4-authentication]

moranbental:
### 📝 Description

This PR handles cases where a job that runs in MLRun imports the MLRun SDK, meaning the job itself needs to authenticate with the MLRun API.

Since we no longer use the V3IO access key for authentication, we must provide an offline token so the SDK can authenticate via Iguazio. To enable this, we configure environment variables to support authentication using the offline token.

Specifically, the offline token is passed as the `MLRUN_AUTH_OFFLINE_TOKEN` environment variable.
Although this is not the recommended approach (the preferred method is using the igz.yml file), it provides a quick and functional solution until the proper implementation is completed — see [ML-11292](https://iguazio.atlassian.net/browse/ML-11292).

In this implementation, we read the offline token from the secret that belongs to the authenticated user (extracted from the AuthInfo), using the "default" token name.

Also fixing the bug of `sync_secret_tokens` fails when offline token is provided via env var and igz.yml is missing (https://iguazio.atlassian.net/browse/ML-11389)

---

### 🛠️ Changes Made

---

### ✅ Checklist

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link: fixing also https://iguazio.atlassian.net/browse/ML-11389
- Design docs links: https://iguazio.atlassian.net/wiki/spaces/ARC/pages/361103361/MLRun+Secret+Tokens+in+IG4
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No
Should be only supported in IG4. 

---

### 🔍️ Additional Notes

[ML-11292]: https://iguazio.atlassian.net/browse/ML-11292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/db/httpdb.py
```diff
@@ -639,7 +639,11 @@ def connect(self, secrets=None):
                 traceback=traceback.format_exc(),
             )
 
-        if config.is_iguazio_v4_mode() and config.auth_with_oauth_token.enabled:
+        if (
+            config.is_iguazio_v4_mode()
+            and config.auth_with_oauth_token.enabled
+            and not os.getenv("MLRUN_AUTH_OFFLINE_TOKEN")
+        ):
```
liranbg: comment why/what's the purpose here?

mlrun/db/httpdb.py
```diff
@@ -5354,6 +5358,12 @@ def revoke_secret_token(self, token_name: str) -> None:
             "delete user secret token",
         )
 
+    @mlrun.utils.iguazio_v4_only
+    def get_secret_token(
+        self, authenticated_username: str, token_name: str
+    ) -> mlrun.common.schemas.SecretToken:
+        raise NotImplementedError
```
liranbg: why it is not implemented? what's the reason?

mlrun/runtimes/base.py
```diff
@@ -443,19 +443,42 @@ def _get_db_run(
...
         :return: Dictionary with all the variables that could be parsed
         """
         active_project = self.metadata.project or config.active_project
+
+        offline_token = None
+        if auth_info and auth_info.username:
+            try:
+                secret = self._get_db().get_secret_token(
+                    authenticated_username=auth_info.username,
+                    token_name="default",
```
liranbg: constant?

mlrun/runtimes/base.py
```diff
@@ -443,19 +443,42 @@ def _get_db_run(
...
+                )
+                offline_token = secret.token if secret else None
+            except Exception as exc:
+                mlrun.utils.logger.warning(
+                    "Failed to retrieve offline token for user",
+                    username=auth_info.username,
+                    error=err_to_str(exc),
+                )
```
liranbg: I think we should fail the run

mlrun/runtimes/base.py
```diff
@@ -443,19 +443,42 @@ def _get_db_run(
...
+
         runtime_env = {
             mlrun_constants.MLRUN_ACTIVE_PROJECT: active_project,
             # TODO: Remove this in 1.12.0 as MLRUN_DEFAULT_PROJECT is deprecated and should not be injected anymore
             "MLRUN_DEFAULT_PROJECT": active_project,
+            "MLRUN_AUTH_OFFLINE_TOKEN": offline_token,
+            "MLRUN_AUTH_WITH_OAUTH_TOKEN__ENABLED": "true",
+            "MLRUN_AUTH_TOKEN_ENDPOINT": "https://igz-api.moran.vmdev44ig4.lab.iguazeng.com/api/v1/refresh-access-token",
```
liranbg: wrong endpoint. use inCluster address from mlrun config.py

mlrun/runtimes/base.py
```diff
@@ -443,19 +443,42 @@ def _get_db_run(
...
+            "MLRUN_HTTPDB__HTTP__VERIFY": "false",
```
liranbg: should not be the default

mlrun/runtimes/base.py
```diff
@@ -443,19 +443,42 @@ def _get_db_run(
...
+                    username=auth_info.username,
+                    error=err_to_str(exc),
+                )
+
         runtime_env = {
             mlrun_constants.MLRUN_ACTIVE_PROJECT: active_project,
             # TODO: Remove this in 1.12.0 as MLRUN_DEFAULT_PROJECT is deprecated and should not be injected anymore
             "MLRUN_DEFAULT_PROJECT": active_project,
+            "MLRUN_AUTH_OFFLINE_TOKEN": offline_token,
+            "MLRUN_AUTH_WITH_OAUTH_TOKEN__ENABLED": "true",
```
liranbg: only if iguazio v4

server/py/framework/utils/clients/iguazio/base.py
```diff
@@ -240,7 +240,11 @@ async def _send_request_to_api_async(
         response = None
         try:
             response = await self._async_session.request(
-                method, url, verify_ssl=False, retry_options=retry_options, **kwargs
+                method,
+                url,
+                verify_ssl=mlrun.mlconf.httpdb.http.verify,
```
liranbg: change to iguazio's config

server/py/framework/utils/clients/iguazio/v4.py
```diff
@@ -43,7 +43,12 @@ def __init__(self, *args, **kwargs) -> None:
             raise mlrun.errors.MLRunRuntimeError(
                 "The 'iguazio' client is only supported in Python >= 3.11"
             )
-        self._client = iguazio.Client(api_url=self._api_url, auto_login=False)
+        self._client = iguazio.Client(
+            api_url=self._api_url,
+            auto_login=False,
+            load_token_file=False,
+            verify_ssl=mlrun.mlconf.httpdb.http.verify,
```
liranbg: change to iguazio config

server/py/services/api/crud/secrets.py
```diff
@@ -595,6 +595,42 @@ def revoke_secret_token(
...
+        :param token_name: Name of the token to retrieve.
+        :return: SecretToken object containing the token name and token value.
+        :raises mlrun.errors.MLRunNotFoundError: If the token does not exist for the user.
+        :raises mlrun.errors.MLRunRuntimeError: If reading or decoding the token fails.
+        """
+        logger.debug(
+            "Getting secret token for user",
+            username=authenticated_username,
+            token_name=token_name,
+        )
```
liranbg: too verbose, remove

server/py/services/api/crud/secrets.py
```diff
@@ -595,6 +595,42 @@ def revoke_secret_token(
...
+        token_value = self.secrets_provider.get_user_token_secret_value(
+            username=authenticated_username,
+            token_name=token_name,
+        )
+
+        logger.debug(
+            "Successfully retrieved secret token for user",
+            username=authenticated_username,
+            token_name=token_name,
+        )
```
liranbg: remove

liranbg (changes requested):

mlrun/secrets.py
```diff
@@ -264,6 +265,11 @@ def sync_secret_tokens() -> None:
     # TODO: Runtime Context Check - Avoid sending a backend request when running inside a runtime, where secrets
     #  are already injected via Kubernetes and syncing is unnecessary
 
+    # Skip syncing secrets when running inside a job,
+    # since the offline token is already provided via the environment variable
+    if os.getenv("MLRUN_AUTH_OFFLINE_TOKEN"):
```
liranbg, line 272: I think this heuristic of assuming running as mlrun job is wrong, need better logic

liranbg (changes requested):

mlrun/auth/providers.py
```diff
@@ -23,6 +23,7 @@
 import mlrun.errors
 import mlrun.secrets
 import mlrun.utils.helpers
+from mlrun.config import config
```
Yacouby, line 26: do not use from X import Y when importing from local code ( see bullet 14 here: https://iguazio.atlassian.net/wiki/spaces/TLE/pages/2457650/MLRun+Python+-+Coding+Conventions )

mlrun/runtimes/base.py
```diff
@@ -154,6 +154,7 @@ class BaseRuntime(ModelObj):
     _default_fields_to_strip = ModelObj._default_fields_to_strip + [
         "status",  # Function status describes the state rather than configuration
     ]
+    _default_token_name = "default"
```
elbamit, line 158: What about defining this as part of the config instead?

mlrun/utils/helpers.py
```diff
@@ -2557,7 +2557,7 @@ def get_relative_module_name_from_path(
 def iguazio_v4_only(function):
     @functools.wraps(function)
     def wrapper(*args, **kwargs):
-        if not mlrun.mlconf.is_iguazio_v4_mode():
+        if not config.is_iguazio_v4_mode():
```
Yacouby, line 2560: what is the problem with using mlrun.mlconf... ?

server/py/services/api/crud/secrets.py
```diff
@@ -593,6 +593,31 @@ def revoke_secret_token(
...
+        :raises mlrun.errors.MLRunNotFoundError: If the token does not exist for the user.
+        :raises mlrun.errors.MLRunRuntimeError: If reading or decoding the token fails.
+        """
+
+        token_value = self.secrets_provider.get_user_token_secret_value(
+            username=authenticated_username,
+            token_name=token_name,
+        )
+
+        return mlrun.common.schemas.SecretToken(
```
Yacouby, line 616: Should we catch token retrieval errors and raise a user-friendly error here?

mlrun/utils/helpers.py
moranbental: We are getting a circular import when importing mlrun inside a job

mlrun/auth/providers.py
moranbental: We are getting a circular import, this is why we are using `mlrun.config`

server/py/services/api/crud/secrets.py
moranbental: `get_user_token_secret_value` will raise an error of `"Token '{token_name}' not found for user '{username}'`

mlrun/auth/providers.py
moranbental: I also tried using `import mlrun.config` instead of `from mlrun.config import config`, and I got the following error:
```
partially initialized module 'mlrun' has no attribute 'config' (most likely due to a circular import)
```

mlrun/secrets.py
moranbental: I updated the comment, as the check is actually related to sync failures when using the offline token from the env var - https://iguazio.atlassian.net/browse/ML-11389

mlrun/runtimes/base.py
moranbental: So I don’t think it should be in the config, as this is server only and will actually be removed when we implement https://iguazio.atlassian.net/browse/ML-11292

liranbg (approved):

Yacouby (approved):

elbamit (approved):

quaark (changes requested):
Everything looks great except for the `get_secret_token` endpoint and `httpdb` method. That's a massive security risk and we should never return stored tokens as http responses.
Since only the internal `sqldb` Implementation is the one that's needed, lets remove the rest.
Everything else LGTM 🚀

quaark (approved):

PR #8849, opened 2025-11-04
[Kaniko] Add S3/MinIO Build Context Support and Improve Builder Environment Handling

adamdelman:
### 📝 Description
Extend Kaniko build support to handle S3 build-context sources and improve how builder environment variables are generated.  
Also standardize logging calls in builder flows and ensure Kubernetes pod environments are always initialized as typed lists.

---

#### **Kubernetes Helper**
- Initialize `Pod.env` as a typed list (`list[V1EnvVar]`) instead of `None` to ensure safe and consistent mutation.

#### **Builder**
- Introduce S3 context detection:
  - Add `is_s3_source` flag.
  - Refine source-scheme logic to treat `s3://` similarly to local/v3io contexts.
- Integrate new `_enrich_kaniko_env_for_s3_context()`:
  - When the build context is `s3://`, set `AWS_REGION` using:
    - existing env vars → or process env → or fallback `"default-region"`.
  - Only if no real region is found (fallback case), also add `S3_FORCE_PATH_STYLE=true` for MinIO compatibility.
- Replace all `mlrun.utils.logger.*` calls with `mlrun.utils.helpers.logger.*` for consistent structured logging.
- Update `_generate_builder_env()`:
  - Accept list-style builder env (`[{name, value}]`) instead of a dict.
  - Convert entries into `V1EnvVar` objects.
  - Merge with project secrets, avoiding duplicates based on the env name.

---

### 🧪 Testing
- Verified build flows with:
  - `s3://` → correct region inference and MinIO path-style injection.
  - `v3io://`, local paths, and `http://` → unchanged behavior.
- Confirmed builder env generation correctly merges user env and project secrets.
- Ensured updated logging source prints structured log fields as expected.

---

### 🔗 References
- [CEML-520](https://iguazio.atlassian.net/browse/CEML-520)

---

### 🚨 Breaking Changes?
- [x] No

---

### 🔍 Additional Notes
- Improves robustness for S3/MinIO user workflows without requiring extra configuration.
- Eliminates inconsistencies around builder environment mutation and logging.


server/py/framework/utils/singletons/k8s.py
```diff
@@ -1677,6 +1678,9 @@ def add_label(self, key, value):
     def add_annotation(self, key, value):
         self._annotations[key] = str(value)
 
+    def add_env_var(self, env_var: client.V1EnvVar):
+        self.env.append(env_var)
```
liranbg: the operation is not idempotent and prone to duplicate envvars

server/py/framework/utils/singletons/k8s.py
```diff
@@ -1724,6 +1728,26 @@ def mount_secret(self, name, path="/secret", items=None, sub_path=None):
...
+            if env_var.name in env_var_names:
+                if env_var.value is None:
+                    logger.warning("Skipping empty secret value")
+                    continue
+                secret_env_var = self._create_secret_for_env_var(
+                    name=env_var.name,
+                    value=env_var.value,
+                    k8s_helper=k8s_helper,
+                )
+                self.env[index] = secret_env_var
```
liranbg: what component is responsible to remove those secret later on? if I have 100 envvars, will it create 100 secrets? what's the reasoning behind it?

server/py/framework/utils/singletons/k8s.py
```diff
@@ -1768,6 +1792,30 @@ def _get_spec(self, template=False):
...
+    def _create_secret_for_env_var(
+        self,
+        name: str,
+        value: str,
+        k8s_helper: K8sHelper,
+    ) -> V1EnvVar:
+        secret_name, _ = k8s_helper.store_auth_secret(
+            username=name,
+            access_key=value,
+        )
```
liranbg: does that aligned with igz4/oauth/igz3 expectations?

server/py/services/api/utils/builder.py
```diff
@@ -1176,3 +1203,49 @@ def _resolve_function_image_secret(
...
+        return
+    if not env_vars:
+        return
+
+    region_defined = any(env.name == "AWS_REGION" and env.value for env in env_vars)
+    if not region_defined:
+        env_vars.append(
+            client.V1EnvVar(
+                name="AWS_REGION",
+                value="us-east-1",
```
liranbg: why defaulting to this region?

liranbg (changes requested):
The PR incorporate few changes at once
1. Fix for s3 region/force path
2. Env to secrets
3. Access key handling

This PR needs to be broken into smaller prs, each has its own risks and need to be carefully taken care of. for the (2) I am not sure I udnerstand the reason behind it. for (3) it seems to be incomplete and also, why is it needed? and for (1) - I think the safest way to proceed is to check if we can already can overcome the issue by providing explicit kaniko build args.

server/py/framework/utils/singletons/k8s.py
adamdelman: This exactly the same as we do in `mlrun_pipelines.common.ops._replace_env_vars_with_secrets`, [which was added in this PR you approved](https://github.com/mlrun/mlrun/pull/6370), where we create obfuscate secret env vars for each pipeline run.
The reasoning is that we do not want AWS/V3IO credentials to be visible on pod YAML's.

server/py/framework/utils/singletons/k8s.py
adamdelman: Again, this is the same as `mlrun_pipelines.common.ops._replace_env_vars_with_secrets`

server/py/services/api/utils/builder.py
adamdelman: Because kaniko expects there to be a region set, even for minio, and this is the default AWS region.

server/py/framework/utils/singletons/k8s.py
adamdelman: Neither is `add_volume`, why do we need it to be idempotent?

adamdelman:
> The PR incorporate few changes at once
> 
> 1. Fix for s3 region/force path
> 2. Env to secrets
> 3. Access key handling
> 
> This PR needs to be broken into smaller prs, each has its own risks and need to be carefully taken care of. for the (2) I am not sure I udnerstand the reason behind it. for (3) it seems to be incomplete and also, why is it needed? and for (1) - I think the safest way to proceed is to check if we can already can overcome the issue by providing explicit kaniko build args.

@liranbg 
The reason for adding item number 2 is that we are adding sensitive credentials to the kaniko pod, and I don't want the to be visible in the pod yaml.
Item number 3 was not really added, if you look carefully it was already there , in this code block:
```    access_key = builder_env.get(
        "V3IO_ACCESS_KEY", auth_info.data_session or auth_info.access_key
    )
    username = builder_env.get("V3IO_USERNAME", auth_info.username)```
 The only difference is that now I also obfuscate these if they exist.

File: server/py/services/api/utils/builder.py [hunk 5]
```diff
@@ -425,14 +425,14 @@ def build_image(
...
+        mlrun.utils.helpers.logger.info("Forcefully building image")
     elif not inline_code and not source and not commands and not requirements:
-        mlrun.utils.logger.info("Skipping build, nothing to add")
+        mlrun.utils.helpers.logger.info("Skipping build, nothing to add")
         return "skipped"
 
     context = "/context"
     to_mount = False
-    is_v3io_source, is_http_source = False, False
+    is_v3io_source, is_s3_source, is_http_source = False, False, False
```
liranbg, line 435: is_s3_source never changes, needs to be initialized with a different value != False

server/py/services/api/utils/builder.py
liranbg: > Because kaniko expects there to be a region set, even for minio, and this is the default AWS region.

not 100%, on aws - admin can set its default region. I suggest to not default ourselves by either get it using aws (perhpas using aws command on init container / mlrun configuration) but def not hardcoded

server/py/services/api/utils/builder.py
```diff
@@ -1176,3 +1185,40 @@ def _resolve_function_image_secret(
...
+        env_vars.append(
+            client.V1EnvVar(
+                name="AWS_REGION",
+                value="us-east-1",
+            )
+        )
+
+        # MinIO requires path-style URLs
+        already_has_path_style = any(
+            env.name == "S3_FORCE_PATH_STYLE" and env.value for env in env_vars
```
liranbg: It is not 100% safe for "new" aws regions not supporting this ENV/configuration as they require virtual-host path. maybe we can force this configuration when we deploy against minio (mlrun ce-like configuration) - but not when working natively against s3

liranbg (changes requested):

liranbg (changes requested):
looking better, worth adding some ut use cases coverage to cover conventional flows

liranbg (approved):

PR #8854, opened 2025-11-05
[IGTokenProvider] Fix token validation calculation logic [feature/ig4-authentication]

elbamit:
### 📝 Description

The current logic of `_is_token_valid()` wasn't correctly calculating if the remaining time is still within the refresh threshold. See Jira for a detailed explanation of the percentage miscalculation.

This PR fixes the percentage calculation and refactors the function name since `valid` can be confusing. A token is valid until the expiration time expires. However it needs refresh if it's past the refresh threshold.

---

### 🛠️ Changes Made

- Refactor `_is_token_valid()` name into `_is_token_within_refresh_threshold()`
- Fix the percentage calculation
---

### ✅ Checklist

---

### 🧪 Testing
  
  

The test logic wasn't correct and it got fixed.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11391
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/mlrun/auth/providers.py
+++ b/mlrun/auth/providers.py
@@ -152,7 +152,7 @@ class DynamicTokenProvider(TokenProvider):
         :return: The refreshed access token.
         """
         # Check if there is an existing access token and if it is valid
-        if self._token and self._is_token_valid(cleanup_if_expired=True):
+        if self._token and self._is_token_within_refresh_threshold(cleanup_if_expired=True):
             return self._token
 
         self.fetch_token()
@@ -168,7 +168,7 @@ class DynamicTokenProvider(TokenProvider):
         pass
 
     @abstractmethod
-    def _is_token_valid(self, cleanup_if_expired=True) -> bool:
+    def _is_token_within_refresh_threshold(self, cleanup_if_expired=True) -> bool:
         """
         Check if the current access token is valid.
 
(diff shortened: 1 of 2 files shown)


liranbg (approved):

mlrun/auth/providers.py
```diff
@@ -152,7 +152,9 @@ def _refresh_token_if_needed(self):
         :return: The refreshed access token.
         """
         # Check if there is an existing access token and if it is valid
-        if self._token and self._is_token_valid(cleanup_if_expired=True):
+        if self._token and self._is_token_within_refresh_threshold(
```
moranbental, line 155: So actually, this code is in the `DynamicTokenProvider`, which means it’s also used by the `OAuthClientIDTokenProvider` class. The function name is fine for `IGTokenProvider`, but I’m not sure if it’s fine for `OAuthClientIDTokenProvider` as we are not using the refresh threshold there

moranbental (changes requested):

mlrun/auth/providers.py
moranbental: Oh, my mistake. There is a refresh threshold validation

moranbental (approved):

PR #8858, opened 2025-11-05
[Dockerfiles] Add align_mlrun.sh to jupyter image

adamdelman:
### 📝 Description
Add `align_mlrun.sh` to the Jupyter image used in MLRun CE.  
The script aligns the **mlrun client** version inside the notebook pod with the **server (API) version**.
---

### 🛠️ Changes Made
- **Dockerfile (jupyter):**
  - Copy `dockerfiles/jupyter/align_mlrun.sh` into the image (owned by `$NB_UID:$NB_GID`), available in `$HOME`.
- **New script (`align_mlrun.sh`):**

---

### ✅ Checklist

---

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/CEML-421
- Design docs links: —
- External links: —
  
---

### 🚨 Breaking Changes?
- [ ] Yes (explain below)
- [x] No

---


dockerfiles/jupyter/Dockerfile
```diff
@@ -124,4 +125,4 @@ RUN mkdir data && tar -cvf /tmp/basehome.tar $HOME
 # and avoid zombie processes
 ENTRYPOINT ["tini", "--"]
 
-CMD echo "1" > "${HOME}/.intdata" && /usr/local/bin/mlce-start.sh
```
liranbg, line 127: why deleting the `echo "1" > "${HOME}/.intdata"`? this argument is useful *in case you run it as a docker container* so you skip extraction..
on k8s - w doe extract because of line 120/121

dockerfiles/jupyter/align_mlrun.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+for env in base jupyter; do
```
liranbg: I am not sure it is relevant. on mlrun's jupyter we ship `mlrun` and `base` env.
the reason. in addition, I dont think the persistent comment here is relevant at all. need to double verify which env is persisnt and which not on our default mlrun's ce deployment and take an action accordingly

liranbg (changes requested):

dockerfiles/jupyter/Dockerfile
adamdelman: On line 121 we don't extract it, was just back it up

liranbg: I know, but the line 120 tells why we back it up (on k8s pvc gets deleting removes the home entirely)
in here, the CMD command puts "1" - it runs and executed *only* when runninng this image as a docker container, and the writing "1" is to ensure, when container is up, it wont *extract* the homedir needlessly *again*

liranbg (approved):

PR #8789, opened 2025-10-28
[AuthZ] Support IG4 Authorization

quaark:
### 📝 Description
Enable project authorization features in IG4 MLRun. Set up with MLRun as the project leader and IG4 BE as an additional follower to nuclio. (Currently a naive forwarding approach, a proper new project sync mechanism for IG4 TBD).

To enable IG4 project authorization, set the following configs in mlrun api:
```
# IG4 authentication
MLRUN_HTTPDB__AUTHENTICATION__MODE: iguazio-v4
MLRUN_HTTPDB__AUTHENTICATION__IGUAZIO__SESSION_VERIFICATION_ENDPOINT: v1/identity/self
MLRUN_IGUAZIO_API_URL: http://igz-api:8000

# IG4 Authorization
MLRUN_HTTPDB__PROJECTS__FOLLOWERS=nuclio,igz
MLRUN_HTTPDB__AUTHORIZATION__NAMESPACES__RESOURCES=resources
MLRUN_HTTPDB__AUTHORIZATION__NAMESPACES__MGMT=mgmt

# disable project sync
MLRUN_HTTPDB__PROJECTS__PERIODIC_SYNC_INTERVAL=0s
```

---

### 🛠️ Changes Made
- Extended iguazio v4 client to support project crud
- Added iguazio v4 client as a project follower
- passed auth info from project member class to all method signatures
- added optional namespaces to OPA resource paths, the namespace can be selected when calling the query function and defaults to `resources` (can be `mgmt`). If the config values are unset, the namespaces are ignored and the behaviour remains BC.
- added a ig4 only check on create projects to check if the user has a Developer policy (irrelevant and won't work in IG3).

---

### ✅ Checklist
- [x] I have tested the changes in this PR

---

### 🧪 Testing
- Added UTs
- Manual testing with multiple users and project cruds in an IG4 system to ensure the behaviour is expected

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10396
- Design docs links: https://iguazio.atlassian.net/wiki/spaces/PLAT/pages/98271458/Authorization+HLD

PR #8859, opened 2025-11-06
[AuthZ] Support IG4 Authorization

quaark:
### 📝 Description

**REOPEN OF**: https://github.com/mlrun/mlrun/pull/8789 after merge into `development` branch. Had issues with rebasing, sorry for the long commit history. Will squash upon merging this PR.

Enable project authorization features in IG4 MLRun. Set up with MLRun as the project leader and IG4 BE as an additional follower to nuclio. (Currently a naive forwarding approach, a proper new project sync mechanism for IG4 TBD).

To enable IG4 project authorization, set the following configs in mlrun api:
```
# IG4 authentication
MLRUN_HTTPDB__AUTHENTICATION__MODE: iguazio-v4
MLRUN_HTTPDB__AUTHENTICATION__IGUAZIO__SESSION_VERIFICATION_ENDPOINT: v1/identity/self
MLRUN_IGUAZIO_API_URL: http://igz-api:8000

# IG4 Authorization
MLRUN_HTTPDB__PROJECTS__FOLLOWERS=nuclio,igz
MLRUN_HTTPDB__AUTHORIZATION__NAMESPACES__RESOURCES=resources
MLRUN_HTTPDB__AUTHORIZATION__NAMESPACES__MGMT=mgmt

# disable project sync
MLRUN_HTTPDB__PROJECTS__PERIODIC_SYNC_INTERVAL=0s
```

---

### 🛠️ Changes Made
- Extended iguazio v4 client to support project crud
- Added iguazio v4 client as a project follower
- passed auth info from project member class to all method signatures
- added optional namespaces to OPA resource paths, the namespace can be selected when calling the query function and defaults to `resources` (can be `mgmt`). If the config values are unset, the namespaces are ignored and the behaviour remains BC.
- added a ig4 only check on create projects to check if the user has a Developer policy (irrelevant and won't work in IG3).

---

### ✅ Checklist

---

### 🧪 Testing
- Added UTs
- Manual testing with multiple users and project cruds in an IG4 system to ensure the behaviour is expected
- Also manually tested in an IG3 system all the project cruds with different users to ensure no regressions

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10396
- Design docs links: https://iguazio.atlassian.net/wiki/spaces/PLAT/pages/98271458/Authorization+HLD


liranbg (approved):

rokatyy (approved):
LGTM

mlrun/common/schemas/auth.py
```diff
@@ -143,6 +148,8 @@ def get_member_ids(self) -> list[str]:
         member_ids = []
         if self.user_id:
             member_ids.append(self.user_id)
+        if self.username:
+            member_ids.append(self.username)
```
moranbental, line 152: Why do you need to add the username if we’re already getting the user id from orca via the `identity/self` request? In that case, you’re adding both the user id and the username for the same user to member_ids.

server/py/framework/utils/clients/iguazio/v4.py
```diff
@@ -73,37 +77,20 @@ def refresh_access_token(
...
-                exc=exc_str,
-            )
-            raise mlrun.errors.MLRunUnauthorizedError(
-                f"Failed to refresh token '{secret_token.name}' from Iguazio: {exc_str}"
-            ) from exc
+
+        return self._try_callback_with_httpx_exceptions(
+            _refresh_access_token,
+            mlrun.errors.MLRunUnauthorizedError,
+            "Failed to refresh access token from Iguazio",
```
moranbental: As it was before
```suggestion
            f"Failed to refresh access token '{secret_token.name}' from Iguazio",
```

server/py/framework/utils/clients/iguazio/v4.py
```diff
@@ -73,37 +77,20 @@ def refresh_access_token(
...
+        )
 
     def refresh_access_tokens(
```
moranbental, line 95: What about changing this function as well to use `_try_callback_with_httpx_exceptions`?

server/py/framework/utils/clients/iguazio/v4.py
```diff
@@ -185,33 +172,180 @@ def revoke_offline_token(
...
+            "Failed to revoke offline token from Iguazio",
+        )
+
+    def create_project(
+        self,
+        session: sqlalchemy.orm.Session,
+        project: mlrun.common.schemas.Project,
+        auth_info: mlrun.common.schemas.AuthInfo = mlrun.common.schemas.AuthInfo(),
+    ):
+        self._logger.debug("Creating default project policies in Iguazio V4")
```
moranbental: I’m not sure why we need to include “Iguazio V4” in the logs every time, we already know we’re running in an Iguazio V4 system.

server/py/framework/utils/projects/follower.py
```diff
@@ -106,28 +106,27 @@ def create_project(
...
         if framework.utils.helpers.is_request_from_leader(
-            projects_role, leader_name=self._leader_name
+            auth_info.projects_role, leader_name=self._leader_name
         ):
             services.api.crud.Projects().create_project(db_session, project)
             return project, False
         else:
             is_running_in_background = self._leader_client.create_project(
-                leader_session, project, wait_for_completion
+                auth_info.session, project, auth_info, wait_for_completion
```
moranbental, line 121: I don’t think we initialize `auth_info.session` in Iguazio v4

moranbental (changes requested):

mlrun/common/schemas/auth.py
quaark: Because in IG4 we decided to use the username instead of the user id.
And it doesn't cost a lot to have both in for both IG3 use case and IG4 use case

server/py/framework/utils/projects/follower.py
quaark: That's perfectly fine here. This is interfaced and the session is only really used when MLRun is CRUDing projects in its own DB through the mechanism.

moranbental (approved):
LGTM

TomerShor (approved):

PR #8866, opened 2025-11-06
[GHA] Add a monthly workflow to update Ruff

jond01:
### 📝 Description

Add a GitHub workflow that runs monthly and on demand to update Ruff version.
This workflow updates the version in the relevant files, formats the code, and creates a PR with the changes.

One point to consider in the future is to notify when `make fmt-ruff && make lint-ruff` doesn't finish successfully.

---

### 🛠️ Changes Made

* Separate the Ruff formatting commands from `make fmt` into `make fmt-ruff`.
* Rename the `lint-check` to `lint-ruff`.
* Add the new workflow.

---

### 🧪 Testing

The workflow passed successfully on my fork. For example, this is a PR opened by it:
https://github.com/jond01/mlrun/pull/11


Makefile
```diff
@@ -858,11 +858,14 @@ html-docs-dockerized: build-test ## Build html docs dockerized
...
 
-.PHONY: fmt
-fmt: ## Format the code using Ruff and blacken-docs
+.PHONY: fmt-ruff
+fmt-ruff: ## Format the code using Ruff
 	@echo "Running ruff checks and fixes..."
-	python -m ruff check --fix-only
-	python -m ruff format
+	ruff check --fix-only
+	ruff format
```
liranbg: not using `python -m` means you might be using your "global" `ruff` tool - if that wasnt installed using your current activated `python` venv

jond01: I expect devs to work in their virtual environment when formatting and linting Python.
In this case, invoking the command with `python -m` is limiting since it doesn't allow global ruff as used in the new workflow.
In short, ruff could be used without Python.

.github/workflows/update-ruff.yaml
```diff
@@ -0,0 +1,103 @@
...
+          - Update dev-requirements.txt with Ruff ${{ steps.latest-version.outputs.ruff-version }}
+          - Update pyproject.toml minimum Ruff version requirement
+          - Apply automatic formatting with make fmt
+          - Verify linting passes with make lint
+        branch: "automated/ruff-update-${{ github.ref_name }}"
+        delete-branch: true
+        # See:
+        # https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#token
+        # Fine-grained Personal Access Token (PAT) with contents: write and pull-requests: write scopes.
+        token: ${{ secrets.MLRUN_PR_PAT }}
```
liranbg: we no longer use mlrun_pr_pat but rather apps token. use @tal-haim to update your PR

liranbg (changes requested):

.github/workflows/update-ruff.yaml
jond01: This PAT is still used here:
https://github.com/mlrun/mlrun/blob/0f548c6c21c01cf4e180ea392bfbb44daf70e6c1/.github/workflows/upgrade-lock.yaml#L60
Where is the apps token used?

tal-haim: @jond01 
see https://github.com/mlrun/mlrun/commit/92852f8f21f3ae01745bb441377e3dc16b9feac7

liranbg (approved):

PR #8867, opened 2025-11-06
[Nuclio] Masking creds in rabbitmq url

rokatyy:
### 📝 Description

Parses url and password from rabbitmq url if it contains such. We leave url in the mlrun config without creds and put username/password in separate fields.  Then procedure is the same as with other sensitive fields in nuclio, meaning that once we send them, mlrun doesn't have them and we keep everything in a nuclio secret. 

---

### 🛠️ Changes Made

---

### ✅ Checklist

---

### 🧪 Testing
unit + vmdev (screenshots are in additional notes section)

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11382
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes
MLRun spec:
image
Successful deploy/redeploy and etc:

image

### DIFF
--- a/mlrun/runtimes/nuclio/function.py
+++ b/mlrun/runtimes/nuclio/function.py
@@ -19,6 +19,7 @@ import typing
 import warnings
 from datetime import datetime
 from time import sleep
+from urllib.parse import urlparse, urlunparse
 
 import inflection
 import nuclio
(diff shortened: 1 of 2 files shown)


server/py/services/api/tests/unit/runtimes/test_nuclio.py
```diff
@@ -1862,6 +1862,75 @@ def test_masking_sensitive_fields(self):
...
+    )
+    def test_masking_rabbitmq_url(self, nuclio_version):
+        """Test that RabbitMQ URL credentials are extracted and masked properly."""
+        password = "rabbit123"
+        url = (
+            f"amqp://user:{password}@my-rabbitmq.default-tenant.svc.cluster.local:5672"
+        )
+
+        def _validate_masked_trigger(masked_trigger):
+            if nuclio_version >= "1.14.15":
```
liranbg: This is not really semver valid way to perform version comparison

mlrun/runtimes/nuclio/function.py
```diff
@@ -1080,6 +1068,79 @@ def _set_sidecar(self, name: str) -> dict:
...
+        This ensures credentials are not exposed in the URL.
+        """
+
+        # supported only for nuclio higher than 1.14.15
+        if not validate_nuclio_version_compatibility("1.14.15"):
+            return
+        if not isinstance(trigger, dict):
+            return
+
+        if trigger.get("kind") != "rabbit-mq":
```
liranbg, line 1105: Is it the only way to set rabbit kind ?

server/py/services/api/tests/unit/runtimes/test_nuclio.py
rokatyy: @liranbg fair, will change to `==` sign not to overcomplicate

mlrun/runtimes/nuclio/function.py
rokatyy: @liranbg yes, we removed support of `rabbitMQ` recently, because it was still tech preview and supporting both isn't really reasonable. `rabbit-mq` is the only documented and supported way.

liranbg (approved):

TomerShor (approved):

PR #8878, opened 2025-11-09
[Bugfix] Fix usage of index hint for list artifacts

adamdelman:
### 📝 Description
Improved the handling of default artifact listing detection and corrected the MySQL index hint syntax to ensure the optimizer uses the intended index.  
This restores consistent query performance for default UI listing requests and avoids invalid SQL hint syntax on MySQL.

---

### 🛠️ Changes Made
- Updated `_is_default_list_artifacts_query` to treat empty containers (`[]`, `{}`, `()`) as equivalent to `None` for default comparison.
- Added debug logging for non-default parameters to help identify when queries deviate from the default path.
- Fixed MySQL index hint syntax by using `USE INDEX (idx_project_bi_updated)` with `dialect_name="mysql"`.
- Left previous hint form commented for traceability.

---

### ✅ Checklist
- [x] Code updated and verified locally  
- [ ] Documentation update (not required)  
- [x] Verified improved query hint behavior on MySQL  
- [x] Verified consistent detection of default list parameters  

---

### 🧪 Testing
- Confirmed `_is_default_list_artifacts_query` returns `True` when parameters are empty lists or dicts.  
- Verified MySQL queries render with `USE INDEX (idx_project_bi_updated)` in `EXPLAIN` output.  
- Checked PostgreSQL and other dialects remain unaffected (no hint applied).  
- Observed correct debug logs when non-default parameters are present.

---

### 🔗 References
- Internal query optimization for artifact listing performance  
- Ensures compatibility with MySQL 8.x hint syntax  

---

### 🚨 Breaking Changes?
- [ ] Yes  
- [x] No  

---

### 🔍️ Additional Notes
This fix restores proper optimizer behavior and improves traceability when query parameters deviate from the UI default list flow.


server/py/framework/db/sqldb/db.py
```diff
@@ -1858,7 +1858,9 @@ def _find_artifacts(
             partition_order,
             parent_uri,
         ):
-            query = query.with_hint(ArtifactV2, "USE INDEX idx_project_bi_updated")
+            query = query.with_hint(
+                ArtifactV2, "USE INDEX idx_proj_best_kind_updated_id"
+            )
```
liranbg, line 1865: this is a wrong index name - see https://github.com/mlrun/mlrun/blob/9da681e4252da70f5bd64e5abb7d0f4c668edf6a/server/py/framework/db/sqldb/models.py#L370

liranbg (changes requested):
This is a non existing

server/py/framework/db/sqldb/db.py
```diff
@@ -1858,7 +1858,11 @@ def _find_artifacts(
             partition_order,
             parent_uri,
         ):
-            query = query.with_hint(ArtifactV2, "USE INDEX idx_project_bi_updated")
+            query = query.with_hint(
+                ArtifactV2,
+                "USE INDEX (idx_project_bi_updated)",
+                dialect_name="mysql",
+            )
```
liranbg, line 1865: lets try capturing it with a unit test to ensure that "default list (tag=latest, best-iteration=true) will get to here

also, dialect_name=mysql = what's the impact on postgres on that case?

server/py/framework/db/sqldb/db.py
```diff
@@ -2077,7 +2081,9 @@ def _is_default_list_artifacts_query(
 
         # Check if all current parameters match their default values
         return all(
-            default_list_params[key] == value for key, value in current_params.items()
+            default_list_params[key] == value
+            or (default_list_params[key] is None and value in (None, [], {}, ()))
```
liranbg, line 2084: worth changing the attach_tags as well to its "default" = True

server/py/framework/db/sqldb/db.py
adamdelman: It simply wont emit the hint in postgres, which only uses hints if you use the `pg_hint_plan` extension

server/py/framework/db/sqldb/db.py
```diff
@@ -2070,7 +2070,6 @@ def _is_default_list_artifacts_query(
             "producer_uri": producer_uri,
             "best_iteration": best_iteration,
             "most_recent": most_recent,
-            "attach_tags": attach_tags,
```
liranbg, line 2069: shouldnt we add `"attach_tags": True` to `default_list_params.update(`?

server/py/services/api/tests/unit/db/test_artifacts.py
```diff
@@ -2793,6 +2794,110 @@ def test_delete_parent_artifacts(self):
...
+            most_recent=False,
+            attach_tags=attach_tags_value,
+            offset=None,
+            limit=1001,
+            with_entities=with_entities_value,
+            partition_by=None,
+            partition_sort_by=mlrun.common.schemas.SortField.updated,
+            partition_order=mlrun.common.schemas.OrderType.desc,
+            parent_uri=None,
+        )
```
liranbg: this function call is tightly coupled with the default sets within the `_is_default_list_artifacts_query` and I think that changes to the function parameters wont trigger a failure on this UT - which means it is needless?

server/py/services/api/tests/unit/db/test_artifacts.py
```diff
@@ -2793,6 +2794,110 @@ def test_delete_parent_artifacts(self):
...
+                if dialect_name == "mysql" and "USE INDEX" in str(text):
+                    hint_called_flag["value"] = True
+                return original_with_hint(self_query_obj, selectable, text, dialect_name=dialect_name)
+
+            monkeypatch.setattr(Query, "with_hint", with_hint_spy, raising=True)
+
+            _ = self._db._find_artifacts(
+                self._db_session,
+                project=self.project,
+                ids=None,
```
liranbg: why setting all function parameters and not *just* the parameters we _consider_ as defaults for UI purposes? that way, changing on the function parameters will trigger a failure when calling find_artifacts
the purpose is that if we change ui defaults, e.g now we find artifacts with parent_uri - meaning we change the default function signature - we should see this UT test failing. is that the case here?

server/py/framework/db/sqldb/db.py
adamdelman: I think it should use the index regardless of if attach_Tags is used, so we can just drop it from the comparison

server/py/framework/db/sqldb/db.py
adamdelman: I think it should use the index regardless of if attach_Tags is used, so we can just drop it from the comparison

liranbg (approved):

PR #8879, opened 2025-11-10
[IGTokenProvider] Fix circular Import when setting MLRUN_DBPATH before Importing MLRun

moranbental:
### 📝 Description

This pr is fixing the problem occurs when defining `MLRUN_DBPATH` as env var before importing mlrun, which causes circular import and prevents auth env vars from being applied

With the fix, we can now use:
```
import os

os.environ["MLRUN_DBPATH"] = "https://mlrun-api.test-qa.vmdev44ig4.lab.iguazeng.com"
os.environ["MLRUN_AUTH_WITH_OAUTH_TOKEN__ENABLED"] = "true"
os.environ["MLRUN_AUTH_TOKEN_ENDPOINT"] = "https://igz-api.moran.vmdev44ig4.lab.iguazeng.com/api/v1/refresh-access-token"

import mlrun
```

instead of
```
import os
os.environ["MLRUN_AUTH_WITH_OAUTH_TOKEN__ENABLED"] = "true"
os.environ["MLRUN_AUTH_TOKEN_ENDPOINT"] = "https://igz-api.moran.vmdev44ig4.lab.iguazeng.com/api/v1/refresh-access-token"

import mlrun
api = "https://mlrun-api.moran.vmdev44ig4.lab.iguazeng.com"
mlrun.set_environment(api)
```

---

### 🛠️ Changes Made

---

### ✅ Checklist

---

### 🧪 Testing
  
  
Tested on IG4 system with a running job remotely 

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11110
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [X] No

---


mlrun/auth/providers.py
```diff
@@ -23,7 +23,7 @@
 import mlrun.errors
 import mlrun.secrets
 import mlrun.utils.helpers
-from mlrun.config import config
+from mlrun.config import config as mlconf
```
Yacouby, line 26: what is the difference between this and :
`from mlrun import mlconf `
?

Yacouby (commented):
how about adding a simple unit test that verifies mlrun can be safely imported after setting the db path and to make sure this behavior doesn't regress in the future ?

mlrun/auth/providers.py
moranbental: It causes a circular import:
```
ImportError: cannot import name 'mlconf' from partially initialized module 'mlrun'
```
I think the problem occurs because the providers code is being called inside httpdb.init

rokatyy, line 26: it sounds more like initialisation order issue rather then circular imports one. We use the same import logic in many places across the code. Could you please add more info about reasoning why it fixed the problem?

moranbental: So, it’s not really true. Sometimes we use `from mlrun import mlconf`, and other times we use `from mlrun.config import config as mlconf`.

The problem was that we were using `mlrun.mlconf` instead of config, but since the code is called inside httpbd.init, we need to use config. Additionally, sync uses the run_db object, which is called in the connect method, causing a circular import.

rokatyy: @moranbental I think the reason why the error happens is that config wasn't populated when we import config object from config file. Because we need to import the config file itself (not a specific) object. So that `populate` function is called before we start doing smth with config. This can be tested by adding print to populate func:

```
def _populate(skip_errors=False):
    """Populate configuration from config file (if exists in environment) and
    from environment variables.

    populate will run only once, after first call it does nothing.
    """
    print("POPULATING")
    global _loaded

    with _load_lock:
        _do_populate(skip_errors=skip_errors)
```
Then import mlrun locally:
```
>>> import mlrun.config
POPULATING

>>> 
>>> from  mlrun.config import config
>>> 
```
As you can see second object didn't run `_populate`

moranbental: @rokatyy The problem occurs because, at the beginning, the first call to `_populate` triggers `_do_populate`, which then calls `get_run_db`. That call initializes httpdb, which initializes the token provider. The token provider then calls `load_secret_tokens_from_file`, which tries to use the file name from the configuration. At that point, we cannot use mlrun.mlconf yet because it is only partially initialized.

rokatyy: @moranbental so the problem is not circular imports but initialisation order. Don't you think we need to restructure the code then? Because now it's very hacky in my opinion.

liranbg: @rokatyy I agree there's initialization order issue. apart from that, and since changing `from x import y` to `from x import y as xy` should not be impacted from initialization order - I think this can be tackled by following PR as the initialization order should be handled delicately while fixing the bug (circular loop) must take place first, right?

to @Yacouby the `from mlrun import mlconf` - `mlconf` is basically an import alias to mlrun.config.config and not declaration. (declaration happens on mlrun.__init__` and every import of mlrun can use it). that's why using `from mlrun import mlconf` is wrong anywhere - because code is subjected to circular loop issue.

This PR - aligns its behavior with most use-cases - what I do see, is a pattern that needs to be remediate across the board by following https://docs.python.org/3/faq/programming.html#what-are-the-best-practices-for-using-import-in-a-module

liranbg: @rokatyy me and @moranbental discussed offline today

The approach we see fit is:
1. Merge this pr to unblock the circular import loop
2. Take action into ensuring the dependency between mlrun config and other files mentioned here (providers, secrets) would be using dependency injection rather than mutual dependency (e.g when calling secrets - provide the configuration as a parameter so that secrets wont need to find the default parameters from mlrun config).

moranbental: will be handled as part of https://iguazio.atlassian.net/browse/ML-11448

PR #8884, opened 2025-11-10
[Auth utils] Remove suffix validation when reading the token file for iguazio v4

moranbental:
### 📝 Description

This PR removes the support for enforcing `.yaml` or `.yml` suffixes in MLRun token files. The `read_secret_tokens_file` function now reads the configured file as-is, regardless of its extension, because only the file content format matters, not the suffix.

Additionally, the default token file in Iguazio has been changed from a `.yaml` file to `igz.yml`, which required upgrading the Iguazio package.

---

### 🛠️ Changes Made

- Removed suffix validation and fallback logic for .yaml / .yml.
- Updated read_secret_tokens_file to read the configured file directly.
- Bumped Iguazio package version to 0.0.1a16.

---

### ✅ Checklist

---

### 🧪 Testing
  
  
UT + checking on ig4 system

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/IG4-630
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


adamdelman (commented):
Overall LGTM, but maybe it could be worth it to validate that the file contents are actually yaml?

moranbental:
> Overall LGTM, but maybe it could be worth it to validate that the file contents are actually yaml?

This is actually already handled when we call `data = yaml.safe_load(token_file_io)`, which will raise a `yaml.YAMLError` that we catch.
I also added a unit test to cover this case.

liranbg (approved):

tests/auth/test_auth_utils.py
elbamit, line 313: I won't call this function "edge_cases" as it also checks for a valid case.

tests/auth/test_auth_utils.py
```diff
@@ -357,6 +332,22 @@ def test_read_secret_tokens_file_edge_cases(
...
+    # Point config to the invalid file
+    monkeypatch.setattr(config.auth_with_oauth_token, "token_file", str(file_path))
+
+    result = mlrun.auth.utils.read_secret_tokens_file(raise_on_error=False)
+    assert result is None
+
+    with pytest.raises(mlrun.errors.MLRunRuntimeError):
+        mlrun.auth.utils.read_secret_tokens_file(raise_on_error=True)
+
+
```
elbamit, line 350: This test is kind of a duplicate of `test_read_secret_tokens_file_edge_cases` case 2 `Non-dict YAML`.
If you specifically want to test this string type yaml content, it can be added as a parameter case in the above function

moranbental:
fixing in https://github.com/mlrun/mlrun/pull/8889

PR #8901, opened 2025-11-12
[Nuclio] Support Iguazio 4 authentication

TomerShor:
### 📝 Description

This PR adds support for Iguazio 4 authentication to Nuclio runtimes.
MLRun communicates with Nuclio in in multiple ways:
- Synchronous client
- Asynchronous client
- [nuclio-jupyter](https://github.com/nuclio/nuclio-jupyter) package

In Iguazio 3, using the `auth_info` session or basic auth with the session as the password was enough.
In Iguazio 4, we need instead to pass the request headers, that contain the `Authorization` header with the user's access token for the authentication info to be passed to Nuclio.

---

### 🛠️ Changes Made

- Implemented the `NuclioAuthInfoWithHeaders` class that injects a request object with the request headers from `auth_info`.
  - This applies both to the sync client and nuclio-jupyter usage, as they both call the `to_request_headers` method.
- For the async client, we inject the headers manually as it is implemented differently since it doesn't use the `requests` package, but `aiohttp`.
- To support nuclio function handlers that use MLRun, moved the Offline Token env var enrichment to a common location `mlrun.auth.utils.enrich_auth_env` , and called in both BaseRuntime and Nuclio function.

---

### ✅ Checklist

---

### 🧪 Testing

This was tested manually against a running IG4 system.
Successfully managed projects and functions.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/NUC-503
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

### DIFF
--- a/mlrun/auth/utils.py
+++ b/mlrun/auth/utils.py
@@ -21,6 +21,9 @@ import mlrun.common.schemas
 import mlrun.utils.helpers
 from mlrun.config import config as mlconf
 
+if typing.TYPE_CHECKING:
+    import mlrun.db
+
 
 def load_offline_token(raise_on_error=True) -> typing.Optional[str]:
     """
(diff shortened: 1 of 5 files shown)


mlrun/common/schemas/auth.py
```diff
@@ -160,3 +162,37 @@ def get_session(self) -> str:
...
+    def __init__(self, base_auth, headers):
+        self.base_auth = base_auth
+        self.headers = headers
+
+    def __call__(self, r):
+        if self.base_auth:
+            r = self.base_auth(r)
+        for k, v in self.headers.items():
+            r.headers[k] = v
+        return r
```
liranbg: what's r

liranbg (commented):
nice. worth using better variable names tho

liranbg (approved):

mlrun/common/schemas/auth.py
```diff
@@ -160,3 +162,37 @@ def get_session(self) -> str:
...
+        super().__init__(**kwargs)
+        self._headers = headers or {}
+
+    def to_requests_auth(self):
+        # Return custom auth handler that applies headers
+        base_auth = super().to_requests_auth()
+        return CustomAuthWithHeaders(base_auth, self._headers)
+
+
+class CustomAuthWithHeaders:
```
moranbental: could be private class

mlrun/auth/utils.py
```diff
@@ -21,6 +21,9 @@
 import mlrun.utils.helpers
 from mlrun.config import config as mlconf
 
+if typing.TYPE_CHECKING:
```
quaark, line 24: Not sure how popular this feature is. Add a comment saying this enables type hints without actually importing the module during runtime.

quaark (changes requested):
Nice! Very minor

mlrun/auth/utils.py
liranbg: it is actually pretty much known (especially for project that enforces type hinting ;)) 

https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

moranbental (changes requested):

elbamit (approved):

moranbental (approved):
Nice!

quaark (approved):

PR #8905, opened 2025-11-13
[Automation] Fix AttributeError when overwrite_registry is None

alxtkr77:
## Problem
The `patch_remote.py` script crashes with AttributeError when `overwrite_registry` is None.

**Reference**: ML-11462

## Solution
Add None check before calling `.endswith()` on `overwrite_registry` parameter.

## Changes
- `automation/patch_igz/patch_remote.py:636` - Added None check: `if overwrite_registry and overwrite_registry.endswith("/")`

### DIFF
--- a/automation/patch_igz/patch_remote.py
+++ b/automation/patch_igz/patch_remote.py
@@ -631,10 +631,10 @@ class MLRunPatcher:
     def _resolve_overwrite_registry(self):
         docker_registry = self._config["DOCKER_REGISTRY"]
         overwrite_registry = self._config["OVERWRITE_IMAGE_REGISTRY"]
-        if docker_registry.endswith("/"):
-            docker_registry = docker_registry[:-1]
-        if overwrite_registry and overwrite_registry.endswith("/"):
-            overwrite_registry = overwrite_registry[:-1]
+        if docker_registry:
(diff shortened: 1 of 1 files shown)


automation/patch_igz/patch_remote.py
```diff
@@ -633,7 +633,7 @@ def _resolve_overwrite_registry(self):
         overwrite_registry = self._config["OVERWRITE_IMAGE_REGISTRY"]
         if docker_registry.endswith("/"):
             docker_registry = docker_registry[:-1]
-        if overwrite_registry.endswith("/"):
+        if overwrite_registry and overwrite_registry.endswith("/"):
             overwrite_registry = overwrite_registry[:-1]
```
liranbg:
```suggestion
        if docker_registry:
            docker_registry = docker_registry.rstrip("/")
        if overwrite_registry:
            overwrite_registry = overwrite_registry.rstrip("/")
```

alxtkr77: ✅ Applied! Changed to use `rstrip("/")` for both registries. Much cleaner - thanks for the suggestion! 

Also added None check for `docker_registry` for consistency.

liranbg (approved):

PR #8916, opened 2025-11-16
[Model Monitoring] Fix performance of long run parquet reading

royischoss:
### 📝 Description

Added specific paths formatting using the filter given as start and end time for better performance while reading parquet shard by time

---

### 🛠️ Changes Made

Changes in parquet reading default behavior

---

### ✅ Checklist

---

### 🧪 Testing

unit tests + added system test `test_partitioned_parquet_time_filter`, passed `test_app_flow`, checked performance on the long run project using v3io

---

### 🔗 References

- Ticket link: https://iguazio.atlassian.net/browse/ML-11277

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No


mlrun/datastore/base.py
```diff
@@ -157,6 +158,121 @@ def upload(self, key, src_path):
...
+                    if key == "year":
+                        parts.append(f"year={current.year}")
+                    elif key == "month":
+                        parts.append(f"month={current.month:02d}")
+                    elif key == "day":
+                        parts.append(f"day={current.day:02d}")
+                    elif key == "hour":
+                        parts.append(f"hour={current.hour:02d}")
+
+                urls.append(f"{base_path.rstrip('/')}/" + "/".join(parts))
```
davesh0812: My concern is what happens if the parquet file is partitioned by another field before the time partition?

mlrun/datastore/base.py
```diff
@@ -157,6 +158,121 @@ def upload(self, key, src_path):
...
+            try:
+                kwargs["filters"] = clean_filters_for_partitions(
+                    kwargs["filters"], partition_keys
+                )
+                df = df_module.read_parquet(url, **kwargs)
+                logger.info("Reading DataFrame", url=url, columns=df.columns)
+                dfs.append(df)
+            except Exception as e:
+                # Skip partitions that don't exist or have no data
+                logger.info("Failed to read DataFrame", url=url, exception=e)
```
gtopper: Need to catch the specific exception, not all exceptions.

mlrun/datastore/base.py
```diff
@@ -157,6 +158,121 @@ def upload(self, key, src_path):
...
+                    elif key == "month":
+                        parts.append(f"month={current.month:02d}")
+                    elif key == "day":
+                        parts.append(f"day={current.day:02d}")
+                    elif key == "hour":
+                        parts.append(f"hour={current.hour:02d}")
+
+                urls.append(f"{base_path.rstrip('/')}/" + "/".join(parts))
+
+                # Advance to next period
```
gtopper: This approach has the advantage of not having to list any directories ourselves. It has the disadvantage of generating a lot of unnecessary requests when the data is sparse. In the worst case, a very large time range is provided by the user, causing a very large number of requests to be made, even if very little or no data exists.

mlrun/datastore/base.py
gtopper: Another reason we should probably implement this using directory tree traversal.

royischoss: Added a verification, by extracting all partitions from `find_partitions` in storey utils @gtopper @davesh0812

mlrun/datastore/base.py
royischoss: added an input of time_partitioning_granularity allowing the user to control the depth we will create the urls

gtopper: It doesn't seem to address the concern of a large range with sparse data at all though? And wasn't the partition granularity is already discovered by `find_partitions`?

royischoss: The time_partitioning_granularity the user provides decide the depth of the base url, even if the target defined different time_partitioning_granularity. More over we take it out of the base period in the application context

mlrun/datastore/base.py
```diff
@@ -157,6 +158,193 @@ def upload(self, key, src_path):
...
+
+            while current <= end_time:
+                # Build directory structure from partition keys
+                parts = []
+                exists = True
+                for key in fmt_keys:
+                    if key == "year":
+                        parts.append(f"year={current.year}")
+                    elif key == "month":
+                        parts.append(f"month={current.month:02d}")
```
gtopper: This is actually not guaranteed. It could be `month=2`, which is what you'd normally get if you partition by an integer column.

mlrun/datastore/base.py
```diff
@@ -157,6 +158,193 @@ def upload(self, key, src_path):
...
+                        parts.append(f"hour={current.hour:02d}")
+
+                    if parts and not filesystem.isdir(
+                        f"{base_path.rstrip('/')}/" + "/".join(parts)
+                    ):
+                        exists = False
+                        break
+
+                if exists:
+                    urls.append(f"{base_path.rstrip('/')}/" + "/".join(parts))
```
gtopper: URL or path?

mlrun/datastore/base.py
```diff
@@ -262,6 +485,10 @@ def as_df(
         file_url = self._sanitize_url(url)
         is_csv, is_json, drop_time_column = False, False, False
         file_system = self.filesystem
+        create_partition_path = kwargs.pop("create_partition_path", True)
+        time_partitioning_granularity = kwargs.pop(
```
gtopper: These new APIs seem to be undocumented? I don't think the solution should force users to tweak these (even if only in some cases).

mlrun/datastore/base.py
```diff
@@ -157,6 +158,193 @@ def upload(self, key, src_path):
...
+
+                if exists:
+                    urls.append(f"{base_path.rstrip('/')}/" + "/".join(parts))
+
+                # Advance to next period
+                if exists or key == time_partitioning_granularity:
+                    if (
+                        "hour" in actual_partition_keys
+                        or "day" in actual_partition_keys
+                    ):
```
```suggestion
                    if step:
````
and set `step=None` at the start. Makes it more obviously correct to humans as well as to static analysis tools like PyCharm. And saves a few cycles in the process.

File: mlrun/datastore/base.py (hunk 4 above)
Comment (royischoss) [reply mlrun/datastore/base.py]: Fixed by casting to integer and working on existing directories as discussed

File: mlrun/datastore/base.py (hunk 6 above)
Comment (royischoss) [reply mlrun/datastore/base.py]: removed time_partitioning_granularity, create_partition_path is currently default behavior with some verification if we even can travel the tree

File: tests/datastore/test_base.py [hunk 8]
```diff
@@ -315,6 +315,7 @@ def test_partition_filtering_year_month():
             end_time=end_time,
             time_column=time_column,
             format="parquet",
+            create_partition_path=False,
```
gtopper: Shouldn't this new parameter be documented somewhere? I'm not sure what it's for myself.

mlrun/datastore/base.py
```diff
@@ -157,6 +158,176 @@ def upload(self, key, src_path):
...
+        filesystem,
+    ):
+        paths = []
+
+        def list_partition_paths_helper(
+            paths: list[str],
+            start_time,
+            end_time,
+            current_path: str,
+            partion_level: str,
```
```suggestion
            partition_level: str,
```

mlrun/datastore/base.py
```diff
@@ -157,6 +158,176 @@ def upload(self, key, src_path):
...
+    def _list_partitioned_paths(
+        base_path: str,
+        start_time: datetime.datetime,
+        end_time: datetime.datetime,
+        partition_level: str,
+        filesystem,
+    ):
+        paths = []
+
+        def list_partition_paths_helper(
```
gtopper: Consider moving it to an outer scope to avoid variable shadowing.

mlrun/datastore/base.py
```diff
@@ -157,6 +158,176 @@ def upload(self, key, src_path):
...
+            paths: list[str],
+            start_time,
+            end_time,
+            current_path: str,
+            partion_level: str,
+        ):
+            timing = current_path.rsplit("/")[-1].split("=")[0]
```
gtopper: Can use `split(..., 1)` here and elsewhere, as well as multiple assignment notation. We may also want to guard against a situation where `=` isn't in the directory name at all.

mlrun/datastore/base.py
```diff
@@ -157,6 +158,176 @@ def upload(self, key, src_path):
...
+            if filesystem.isfile(current_path):
```
gtopper: As discussed offline, this shouldn't happen normally, so checking `isfile` is too expensive for a just-in-case. Better just check if it ends in `.parquet` (and maybe `.pq`).

mlrun/datastore/base.py
```diff
@@ -157,6 +158,176 @@ def upload(self, key, src_path):
...
+        for i in range(len(paths)):
+            paths[i] = DataStore._reconstruct_path_from_base_path(base_path, paths[i])
+        return paths
+
+    @staticmethod
+    def _reconstruct_path_from_base_path(base_path: str, returned_path: str) -> str:
+        base_path = base_path.rstrip("/")
+        clean_return = returned_path.strip("/")
+
+        prefix, _, base_suffix = base_path.partition(":///")
```
gtopper: `base_suffix` is never used. Isn't this just

```suggestion
        prefix = base_path.split(":///", 1)[0]
```
?

mlrun/datastore/base.py
```diff
@@ -157,6 +158,176 @@ def upload(self, key, src_path):
...
+        )
+        partition_end = (
+            partition_start
+            + relativedelta(
+                years=1 if month is None else 0,
+                months=1 if day is None and month is not None else 0,
+                days=1 if hour is None and day is not None else 0,
+                hours=1 if hour is not None else 0,
+            )
+            - datetime.timedelta(microseconds=1)
```
gtopper, line 190: Not really necessary given that we're comparing with `>` below.
```suggestion
```

mlrun/datastore/base.py
```diff
@@ -157,6 +158,176 @@ def upload(self, key, src_path):
...
+        prefix, _, base_suffix = base_path.partition(":///")
+        schema = prefix + ":///"
+
+        return schema + clean_return
+
+    @staticmethod
+    def _clean_filters_for_partitions(
+        filters,
+        partition_keys,
+    ):
```
gtopper, line 270: Could be clearer with type annotations.

mlrun/datastore/base.py
```diff
@@ -157,6 +158,176 @@ def upload(self, key, src_path):
...
+        """
+        Remove partition keys from filters.
+
+        :param filters: (list of lists of tuples) pandas-style filters
+                Example: [[('year','=',2025),('month','=',11),('timestamp','>',ts1)]]
+        :param partition_keys: (list of str) partition columns handled via directory
+
+        :return list of list of tuples: cleaned filters
```
gtopper: Could be clearer on what it means to "clean" them.

gtopper: Also on the fence regarding whether we should do this at all or just leave the filters as-is.

mlrun/datastore/base.py
```diff
@@ -157,6 +158,176 @@ def upload(self, key, src_path):
...
+    def _read_partitioned_parquet(
+        base_path,
+        start_time,
+        end_time,
+        partition_keys,
+        df_module,
+        filesystem,
+        **kwargs,
+    ):
+        """Read only the relevant partitions using pandas filters and concat."""
```
```suggestion
        """Reads only the relevant partitions and concatenates the results."""
```

mlrun/datastore/base.py
```diff
@@ -157,6 +158,176 @@ def upload(self, key, src_path):
...
+        logger.debug("starting paths partition process")
```
```suggestion
        logger.debug(f"Starting partition discovery process for {base_path}")
```

mlrun/datastore/base.py
```diff
@@ -157,6 +158,176 @@ def upload(self, key, src_path):
...
+        if not partition_keys:
+            raise ValueError(
```
gtopper: Need to raise an error from `mlrun.errors`. But really, this check isn't necessary. We just need to check this before this function is called (which we do).

mlrun/datastore/base.py
```diff
@@ -157,6 +158,176 @@ def upload(self, key, src_path):
...
+    @staticmethod
+    def _read_partitioned_parquet(
+        base_path,
+        start_time,
+        end_time,
+        partition_keys,
+        df_module,
+        filesystem,
+        **kwargs,
+    ):
```
gtopper, line 296: Could benefit from type annotations.

mlrun/datastore/base.py
```diff
@@ -157,6 +158,176 @@ def upload(self, key, src_path):
...
+
+        dfs = []
+        errors = []
+        for current_path in paths:
+            try:
+                kwargs["filters"] = DataStore._clean_filters_for_partitions(
+                    kwargs["filters"], partition_keys
+                )
+                df = df_module.read_parquet(current_path, **kwargs)
+                logger.debug("Reading DataFrame", url=current_path, columns=df.columns)
```
gtopper: It's already done at this point (read, past tense, not reading). And we probably don't want to repeat the potentially many columns potentially many times – even in debug.

mlrun/datastore/base.py
```diff
@@ -157,6 +158,176 @@ def upload(self, key, src_path):
...
+                dfs.append(df)
+            except (FileNotFoundError, pyarrow.lib.ArrowInvalid) as e:
+                # Skip partitions that don't exist or have no data
+                logger.warning(
+                    "Failed to read DataFrame", url=current_path, exception=e
+                )
+                errors.append(e)
+                continue
```
gtopper: Don't exist yeah, but why are we suppressing real errors with `ArrowInvalid` and then raising the first error in the end anyway? Why not just raise it immediately? (by not catching it)

mlrun/datastore/base.py
```diff
@@ -157,6 +158,176 @@ def upload(self, key, src_path):
...
+            pyarrow_errors = [
+                err for err in errors if isinstance(err, pyarrow.lib.ArrowInvalid)
+            ]
+            if pyarrow_errors:
+                raise pyarrow_errors[0]
+            else:
+                return pd.DataFrame()
+
+        final_df = pd.concat(dfs)
+        logger.debug("Finished reading DataFrame columns", columns=final_df.columns)
```
```suggestion
        logger.debug("Finished reading partitioned parquet files", url=base_path,  columns=final_df.columns)
```

tests/datastore/test_base.py
royischoss: I wonder where to put it, because it relevant only for the parquet and more over `DataStore.as_df` is not documented

mlrun/datastore/base.py
royischoss: The way I worked before file could not exist, I will change it

mlrun/datastore/base.py
```diff
@@ -157,6 +158,178 @@ def upload(self, key, src_path):
...
+            return
+        for directory in directories:
+            current_path = directory["name"]
+            parts = [p for p in current_path.split("/") if "=" in p]
+            kwargs = {}
+            for part in parts:
+                part = part.split("=")
+                key = part[0]
+                value = int(part[1]) if part[1].isdigit() else part[1]
+                kwargs[key] = value
```
gtopper: Simpler:
```suggestion
                key, value = part.split("=", 1)
                if value.isdigit():
                    value = int(value)
                kwargs[key] = value
```

mlrun/datastore/base.py
```diff
@@ -262,6 +471,7 @@ def as_df(
         file_url = self._sanitize_url(url)
         is_csv, is_json, drop_time_column = False, False, False
         file_system = self.filesystem
+        create_partition_path = kwargs.pop("create_partition_path", True)
```
gtopper: Parameter name could be much clearer (`optimize_partition_discovery`?). A comment might be helpful as well.

mlrun/datastore/base.py
```diff
@@ -157,6 +158,178 @@ def upload(self, key, src_path):
...
+                cleaned_filters.append(new_group)
+        return cleaned_filters
+
+    @staticmethod
+    def _read_partitioned_parquet(
+        base_path: str,
+        start_time: datetime.datetime,
+        end_time: datetime.datetime,
+        partition_keys: list[str],
+        df_module: Union[pd, object],
```
gtopper: `Union[pd, object]`? Pretty sure everything is an `object` in Python, so this is equivalent to just `object`.

mlrun/datastore/base.py
gtopper: Moved, but not changed, so still outstanding.

tests/system/feature_store/test_feature_store.py
```diff
@@ -968,6 +968,210 @@ def test_filtering_parquet_by_time(self, with_tz, local):
...
+        if partition_keys == ["year"]:
+            granularity = "year"
+        elif partition_keys == ["year", "month"]:
+            granularity = "month"
+        elif partition_keys == ["year", "month", "day"]:
+            granularity = "day"
+        elif partition_keys == ["year", "month", "day", "hour"]:
+            granularity = "hour"
+        else:
+            raise ValueError("Unexpected partition_keys")
```
gtopper: This list comes from us, so it should suffice to just make it

```suggestion
        granularity = partition_keys[-1]
```

mlrun/datastore/base.py
royischoss: If we passed year in the path and gave year it will search it in the DataFrame

tests/datastore/test_base.py
gtopper: We've decided that a better name plus probably a comment would be the best approach.

mlrun/datastore/base.py
```diff
@@ -157,6 +158,174 @@ def upload(self, key, src_path):
...
+            paths, start_time, end_time, base_path, partition_level, filesystem
+        )
+        for i in range(len(paths)):
+            paths[i] = DataStore._reconstruct_path_from_base_path(base_path, paths[i])
+        return paths
+
+    @staticmethod
+    def _reconstruct_path_from_base_path(base_path: str, returned_path: str) -> str:
+        base_path = base_path.rstrip("/")
+        clean_return = returned_path.strip("/")
```
gtopper: Here too, it's a bit confusing what "clean return" means.

mlrun/datastore/base.py
```diff
@@ -157,6 +158,174 @@ def upload(self, key, src_path):
...
+        )
+
+        dfs = []
+        for current_path in paths:
+            kwargs["filters"] = DataStore._clean_filters_for_partitions(
+                kwargs["filters"], partition_keys
+            )
+            df = df_module.read_parquet(current_path, **kwargs)
+            logger.debug(
+                "Finish reading DataFrame from specific url",
```
```suggestion
                "Finished reading DataFrame from subpath",
```

mlrun/datastore/base.py
```diff
@@ -157,6 +158,174 @@ def upload(self, key, src_path):
...
+                url=current_path,
+                columns=df.columns,
```
gtopper: Again, do we really want to spam with potentially many columns, repeated per partition? This could be a lot of repetitive output. Even for debug.

mlrun/datastore/base.py
gtopper: We probably do want to tolerate a missing file, though. If only to avoid failure on concurrent deletions.

gtopper (commented):
Almost there...

tests/system/feature_store/test_feature_store.py
```diff
@@ -968,6 +968,192 @@ def test_filtering_parquet_by_time(self, with_tz, local):
...
+                {
+                    key: i + 1,
+                    "timestamp": base_time + timedelta(hours=i),
+                    "value": i * 10,
+                }
+                for i in range(4)
+            ]
+        )
+
+        granularity = partition_keys[-1]
```
gtopper: I guess I'd just inline it at this point.

tests/system/feature_store/test_feature_store.py
```diff
@@ -968,6 +968,192 @@ def test_filtering_parquet_by_time(self, with_tz, local):
...
+                {
+                    key: str(uuid.uuid4()),
+                    "timestamp": base_time + timedelta(hours=i),
+                    "value": i * 10,
+                }
+                for i in range(4)
+            ]
+        )
+
+        granularity = partition_keys[-1]
```
gtopper: I guess I'd just inline it at this point.

tests/system/feature_store/test_feature_store.py
```diff
@@ -968,6 +968,192 @@ def test_filtering_parquet_by_time(self, with_tz, local):
...
+        large_base_period_end = base_time + timedelta(days=1)
+        start = datetime.now(tz=pytz.UTC if with_tz else None)
+        result_df = target.as_df(
+            start_time=large_base_period_start,
+            end_time=large_base_period_end,
+            time_column="timestamp",
+        )
+        end = datetime.now(tz=pytz.UTC if with_tz else None)
+        assert end - start < timedelta(
+            seconds=120
```
gtopper: Should be a lot less than that, no?

tests/system/feature_store/test_feature_store.py
```diff
@@ -968,6 +968,192 @@ def test_filtering_parquet_by_time(self, with_tz, local):
...
+
+        large_base_period_start = base_time - timedelta(days=365)
+        large_base_period_end = base_time + timedelta(days=1)
+        start = datetime.now(tz=pytz.UTC if with_tz else None)
+        result_df = target.as_df(
+            start_time=large_base_period_start,
+            end_time=large_base_period_end,
+            time_column="timestamp",
+        )
+        end = datetime.now(tz=pytz.UTC if with_tz else None)
```
gtopper: To measure actual duration, the simplest and most reliable way to do it is with `time.monotonic()`.

tests/datastore/test_base.py
```diff
@@ -315,6 +315,7 @@ def test_partition_filtering_year_month():
             end_time=end_time,
             time_column=time_column,
             format="parquet",
+            apply_discovery_urls=False,
```
gtopper: I don't think this is a good name either. What this feature flag does is potentially disable a directory traversal (or 'discovery') optimization. URLs might be saved in the process, but that's not really the point of this feature flag. I'd call it `optimize_discovery`, `disable_discovery_optimization`, `skip_out_of_range_partitions`, or something like that.

gtopper: And we probably want to mention that it's a feature flag in a comment.

mlrun/datastore/base.py
```diff
@@ -388,6 +615,28 @@ def _is_dd(df_module):
         except ImportError:
             return False
 
+    @staticmethod
+    def verify_path_partition_level(path: str, partitions: list[str]) -> bool:
```
gtopper: `path` parameter shadows imported `path` – `from os import path, remove`. Better change the name or the import.

mlrun/datastore/base.py
```diff
@@ -388,6 +615,28 @@ def _is_dd(df_module):
...
+        if not partitions:
+            return False
+
+        path_parts = urlparse(path).path.strip("/").split("/")
+        path_parts = [part.split("=")[0] for part in path_parts if "=" in part]
```
gtopper: There's a risk that the equal sign could appear in the base path.

mlrun/datastore/base.py
```diff
@@ -388,6 +615,28 @@ def _is_dd(df_module):
...
+            return False
+        for i, part in enumerate(partitions):
+            if (
+                part in path_parts
+                or part in ["year", "month", "day", "hour"]
+                or i > hour_index
+            ):
+                continue
+            else:
+                return False
```
gtopper: Can be simplified a bit:
```suggestion
            if not (
                part in path_parts
                or part in ["year", "month", "day", "hour"]
                or i > hour_index
            ):
                return False
```

tests/system/feature_store/test_feature_store.py
```diff
@@ -968,6 +968,192 @@ def test_filtering_parquet_by_time(self, with_tz, local):
...
+        "partition_keys",
+        [
+            ["year"],
+            ["year", "month"],
+            ["year", "month", "day"],
+            ["year", "month", "day", "hour"],
+        ],
+    )
+    @pytest.mark.parametrize("with_tz", [True, False])
+    def test_partitioned_parquet_as_df(self, partition_keys, with_tz):
```
gtopper:
```suggestion
    def test_partitioned_parquet_as_df_time_filtering_optimization(self, partition_keys, with_tz):
```
We're not just testing `as_df` in general here.

mlrun/datastore/base.py
gtopper: This one is still outstanding – I think it should tolerate missing partitions (with a warning for each seems reasonable), but not tolerate other failures (ArrowInvalid). Then, the warning should be specific to the directory being missing.
And we shouldn't be accumulating errors to raise one of them later. Fail early where we fail, and tolerate what we've decided to tolerate.

mlrun/datastore/base.py
```diff
@@ -157,6 +158,192 @@ def upload(self, key, src_path):
...
+                cleaned_filters.append(new_group)
+        return cleaned_filters
+
+    @staticmethod
+    def _read_partitioned_parquet(
+        base_path: str,
+        start_time: datetime.datetime,
+        end_time: datetime.datetime,
+        partition_keys: list[str],
+        df_module: Union[pd, Any],
```
gtopper: Again, `Union[pd, Any]` is the same as `Any`. Also, I don't think `pd` is even a type... It's a module object. The correct type is `ModuleType` – i.e. `from types import ModuleType`.

```suggestion
        df_module: ModuleType,
```

gtopper (commented):
Mostly minor comments, but still a few significant points that need to be ironed out before we merge.

mlrun/datastore/base.py
royischoss: I work here only on the base path, this method checks that if there is another layer that should be filter before the timing it is provided in the url. Also allows a field after the timing.

mlrun/datastore/base.py
royischoss: returned it because I don't want to miss datetime that are smaller than end_time

mlrun/datastore/base.py
gtopper: What if there's an equal sign in the base path itself? A bit weird, but legal.

mlrun/datastore/base.py
```diff
@@ -388,6 +609,26 @@ def _is_dd(df_module):
         except ImportError:
             return False
 
+    @staticmethod
+    def verify_path_partition_level(base_path: str, partitions: list[str]) -> bool:
```
gtopper: I guess it should be made private?

gtopper (approved):
LGTM!

PR #8958, opened 2025-11-23
[Docs] Add links to hub page, fix multi-port API gateway example

jillnogold:
### 📝 Description
Add links to import model monitoring apps from a few pages
Fix the code example for exposing multiple ports in API gateway


docs/model-monitoring/applications.md
```diff
@@ -3,6 +3,10 @@
 
 Learn how to create your own model monitoring applications for LLMs, gen AI, deep-learning models, etc., based on the `ModelMonitoringApplicationBase` class.
 
+```{admonition} Note
```
GiladShapira94: I think we should add this note only in one place. we can direct for example this page to the one that it is most relevant to this topic

docs/runtimes/application.ipynb
```diff
@@ -195,10 +196,9 @@
     "app.with_sidecar(ports=[8010,8020])\n",
     "# Set the internal (main) port\n",
     "app.set_internal_application_port(8010)\n",
-    "# Disable the API gateway\n",
-    "app.deploy(create_default_api_gateway=False)\n",
-    "# Add API gateways for both ports\n",
-    "url_direct = app.create_api_gateway(port=8010,name=\"port-8010\",direct_port_access=True,authentication_mode=mlrun.common.schemas.api_gateway.APIGatewayAuthenticationMode.none)\n",
+    "# Create an API gateway. By default, this API Geatway forwards events to the intiernal port (8010)\n",
```
GiladShapira94: Maybe change to. - By default for each application MLRun created an API gateway that forwards the events to the internal port, in this example 8010

docs/runtimes/application.ipynb
liranbg, line 1: @rokatyy - perhaps you can help out here - it doesnt seem to be a conventinoal flow or am I missing something here?


```
# Set the internal (main) port
app.set_internal_application_port(8010)
# Create another API gateway. 
app.deploy(create_default_api_gateway=True)
```


what's the another? @jillnogold app.deploy - deploys the application - `create_default_api_gateway` is just a flag telling it to create APIGW. there no another here

> \# Creates another APIGateway to the application on port 8020 using direct access (add link to what's direct access is). I think that should be better
url_direct = app.create_api_gateway(port=8020, name="port-8020",direct_port_access=True)

liranbg (changes requested):

docs/runtimes/application.ipynb
```diff
@@ -183,22 +184,22 @@
...
     "direct_port_access=True\n",
     "# Set multiple ports\n",
     "app.with_sidecar(ports=[8010,8020])\n",
     "# Set the internal (main) port\n",
     "app.set_internal_application_port(8010)\n",
-    "# Disable the API gateway\n",
-    "app.deploy(create_default_api_gateway=False)\n",
-    "# Add API gateways for both ports\n",
-    "url_direct = app.create_api_gateway(port=8010,name=\"port-8010\",direct_port_access=True,authentication_mode=mlrun.common.schemas.api_gateway.APIGatewayAuthenticationMode.none)\n",
+    "# Create another API gateway. \n",
```
rokatyy: `deploy` doesn’t only create an API Gateway — it does much more. Specifically, it deploys the Nuclio function together with the custom user image. I’d suggest adding something like:

“If you want to deploy an application with the default API Gateway, simply run app.deploy().
If you don’t need the default API Gateway, or if you prefer to create your own custom one, set `create_default_api_gateway=False` when calling `deploy()` and then create a custom API Gateway manually.
Note that if you specify a custom port, you must set direct_port_access=True; otherwise, the value will be ignored and the internal application port will be used instead.”

jillnogold: @rokatyy Is this code correct to define a second API GW?
```
# Enable direct_port_access
direct_port_access=True
# Set multiple ports
app.with_sidecar(ports=[8010,8020])
# Set the internal (main) port
app.set_internal_application_port(8010)
# Add API gateway for second port
url_direct = app.create_api_gateway(port=8020,name="port-8020",direct_port_access=True,authentication_mode=mlrun.common.schemas.api_gateway.APIGatewayAuthenticationMode.none)
```

rokatyy: yes

rokatyy (approved):

rokatyy (approved):

liranbg (approved):

PR #8961, opened 2025-11-23
[Pagination] Fix potential race condition

liranbg:
### 📝 Description

When two users are trying to create same token it might be that they will fail on db conflict since record was just created by another user.

---

### 🛠️ Changes Made

- Used `for_update` notion to ensure GET/UPDATE operation locked for its critical path for ensuring pagination token.
- Reduce DB calls from 2 to 1 when updating existing tokens
- Improved UT coverage
- 

---

### ✅ Checklist

---

### 🧪 Testing

UT coverage and sanity for pagination

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11468
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -7760,7 +7760,7 @@ class SQLDB(DBInterface):
         ).hexdigest()
 
         # it could be that the record was passed, but belongs to a different user
-        # therefor, we need to re-fetch for teh correct user.
+        # therefore, we need to re-fetch for the correct user.
         # the scenario where pagination_cache_record.user != user is when
         # user A (user) using token belongs to user B (pagination_cache_record.user).
         # in that case, ensure we retrieve the correct record for user A.
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -7758,16 +7758,10 @@ class SQLDB(DBInterface):
         key = hashlib.sha256(
             f"{user}/{function}/{page_size}/{kwargs}".encode()
         ).hexdigest()
-
-        # it could be that the record was passed, but belongs to a different user
-        # therefore, we need to re-fetch for the correct user.
-        # the scenario where pagination_cache_record.user != user is when
-        # user A (user) using token belongs to user B (pagination_cache_record.user).
-        # in that case, ensure we retrieve the correct record for user A.
-        # use case is covered by UT - "test_paginate_no_auth"
-        if not pagination_cache_record or pagination_cache_record.user != user:
+        if not pagination_cache_record:
+            # in this case, we just lock for update to make sure no one else is writing to it
             pagination_cache_record = self.get_paginated_query_cache_record(
-                session, key, for_update=True
+                session, key=key, for_update=True
             )
         if pagination_cache_record:
             pagination_cache_record.current_page = current_page
--- a/server/py/framework/db/sqldb/db.py
+++ b/server/py/framework/db/sqldb/db.py
@@ -7743,6 +7743,9 @@ class SQLDB(DBInterface):
         current_page: int,
         page_size: int,
         kwargs: dict,
+        pagination_cache_record: typing.Optional[
+            framework.db.sqldb.models.PaginationCache
+        ] = None,
     ):
         self._validate_integer_max_value(
             PaginationCache.__table__.c.current_page, current_page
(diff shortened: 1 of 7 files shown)


server/py/framework/db/sqldb/db.py
```diff
@@ -7757,11 +7764,21 @@ def store_paginated_query_cache_record(
...
             f"{user}/{function}/{page_size}/{kwargs}".encode()
         ).hexdigest()
-        existing_record = self.get_paginated_query_cache_record(session, key)
-        if existing_record:
-            existing_record.current_page = current_page
-            existing_record.last_accessed = datetime.now(timezone.utc)
-            param_record = existing_record
+
+        # it could be that the record was passed, but belongs to a different user
+        # therefor, we need to re-fetch for teh correct user.
```
```suggestion
        # therefore, we need to re-fetch for the correct user.
```

server/py/framework/db/sqldb/db.py
rokatyy: if exception happens after locking and before _upsert, then we don't commit and don't rollback or did I miss something?

rokatyy (commented):
Have just one question regarding locking

server/py/framework/utils/pagination.py
theSaarco, line 263: In this case there is no commit/rollback after using `for_update`. However, in this case no row was returned so just to verify that in this case no rows are being locked?

server/py/framework/db/sqldb/db.py
```diff
@@ -7757,11 +7764,21 @@ def store_paginated_query_cache_record(
...
+        # the scenario where pagination_cache_record.user != user is when
+        # user A (user) using token belongs to user B (pagination_cache_record.user).
+        # in that case, ensure we retrieve the correct record for user A.
+        # use case is covered by UT - "test_paginate_no_auth"
+        if not pagination_cache_record or pagination_cache_record.user != user:
+            pagination_cache_record = self.get_paginated_query_cache_record(
+                session, key, for_update=True
+            )
```
theSaarco: Wouldn't it be cleaner if you pass the `user` to `get_paginated_query_cache_record` instead, in which case you won't get conflicting cache entries to begin with?

server/py/framework/utils/pagination.py
```diff
@@ -232,12 +232,30 @@ def _create_or_update_pagination_cache_record(
...
+        """
+        pagination_cache_record = None
         if token:
             self._logger.debug(
                 "Token provided, updating pagination cache record", token=token
             )
             pagination_cache_record = (
-                self._pagination_cache.get_pagination_cache_record(session, key=token)
+                self._pagination_cache.get_pagination_cache_record(
+                    session, key=token, for_update=True
```
theSaarco, line 257: For example, you could pass `auth_info.user_id` (or `None` if no auth info) here, to get only the correct cache record.

theSaarco (approved):
Looks good, I have some optional suggestions and a question but nothing critical (and I believe I know the answer to the question anyway).

server/py/framework/db/sqldb/db.py
liranbg: we close the session which releases the per-row locks

server/py/framework/utils/pagination.py
liranbg: related to https://github.com/mlrun/mlrun/pull/8961#discussion_r2556286621 in a way.
*but* the thing you raise here is - we do lock on a row that never existed but since we use "repeatable read" (mysql defaults) - it does lead to the row being locked for inserts (unless we will commit). 

the thing is that - it actually the intention here. we are currently "getting" and since it does not exists - we will "lock" the row and currently flow is that we use the get in the "get or create" so until we create - the row is locked. once created, the other session will *get*.

server/py/framework/db/sqldb/db.py
liranbg: looking at it again - I am not sure the use-case here is legit because `if user and (not auth_info or auth_info.user_id != user):` (appears few lines later) should enforce no 2 users sharing the token.

the use-case came from a irrelvant use case where token was saved for user without auth and requested by user with auth. I have removed this check as it is not needed and updated the test.

@quaark please review it as well

server/py/framework/db/sqldb/db.py
rokatyy: and we close using `finally` block, which means that even if exception happens, it's safe https://github.com/mlrun/mlrun/blob/aca13500b05b960b629ef00a848c4286b08ceaec/server/py/framework/api/deps.py#L38-L40

rokatyy (approved):
LGTM

server/py/framework/utils/pagination.py
liranbg: answered @ https://github.com/mlrun/mlrun/pull/8961#discussion_r2569985146

theSaarco (approved):
LGTM

PR #8964, opened 2025-11-24
[Model Monitoring] Add timestamp_column parameter to TimescaleDB _get_records()

alxtkr77:
## Summary

Add `timestamp_column` parameter to TimescaleDB `_get_records()` method to allow flexible time filtering. The parameter specifies which timestamp column to use in the WHERE clause for time range filtering.

## Changes

- Add `timestamp_column` parameter throughout query chain (connector -> queries -> builder -> schema)
- Default to `None` to use each table's default time column
- Update WHERE clause logic in `_get_records_query()` to use custom or default timestamp column
- Add comprehensive test to verify functionality with custom timestamp columns

## Use Case

Enables filtering by one timestamp while sorting by another:

```python
df = connector._get_records(
    table="app_results",
    start=datetime.min,
    end=datetime.now(),
    timestamp_column="start_infer_time"  # Filter using this column
).sort_values(by="end_infer_time")       # Sort using different column
```

## Reference

ML-11516

### DIFF
--- a/mlrun/model_monitoring/db/tsdb/timescaledb/timescaledb_schema.py
+++ b/mlrun/model_monitoring/db/tsdb/timescaledb/timescaledb_schema.py
@@ -283,12 +283,17 @@ class TimescaleDBSchema:
         desc: Optional[bool] = None,
         use_pre_aggregates: bool = True,
         group_by: Optional[list[str]] = None,
+        timestamp_column: Optional[str] = None,
     ) -> str:
-        """Build query to get records from the table or its pre-aggregates."""
+        """Build query to get records from the table or its pre-aggregates.
+
+        :param timestamp_column: Optional timestamp column to use for time filtering.
+                                If not provided, uses the table's default time_column.
+        """
 
         # Determine table to query
         table_name = self.table_name
-        time_col = self.time_column
+        time_col = timestamp_column or self.time_column
 
         if interval and agg_funcs and use_pre_aggregates:
             # Use continuous aggregate if available
@@ -302,9 +307,10 @@ class TimescaleDBSchema:
                 if interval and agg_funcs and use_pre_aggregates:
                     # For pre-aggregates, use column names as-is since they should already be
                     # the correct names from the continuous aggregate view
+                    source_time_col = timestamp_column or self.time_column
                     modified_columns = []
                     for col in columns_to_filter:
-                        if col == self.time_column:
+                        if col == source_time_col:
                             modified_columns.append(TIME_BUCKET_COLUMN)
                         else:
                             # Use column name as-is - caller should provide correct pre-agg column names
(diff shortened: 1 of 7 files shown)


mlrun/model_monitoring/db/tsdb/timescaledb/timescaledb_schema.py
```diff
@@ -283,12 +283,17 @@ def _get_records_query(
         desc: Optional[bool] = None,
         use_pre_aggregates: bool = True,
         group_by: Optional[list[str]] = None,
+        timestamp_column: Optional[str] = None,
     ) -> str:
-        """Build query to get records from the table or its pre-aggregates."""
+        """Build query to get records from the table or its pre-aggregates.
+
+        :param timestamp_column: Optional timestamp column to use for time filtering.
+                                If not provided, uses the table's default time_column.
```
Eyal-Danieli: Just to keep consistency, if we provide docstring about params we should provide for all of the params and not just one. However, in this case I would remove that docstring and maybe add a comment before L296, saying 
`# If timestamp_column not provided, use the table's default time_column` 
```suggestion

```

mlrun/model_monitoring/db/tsdb/timescaledb/timescaledb_schema.py
```diff
@@ -302,9 +307,10 @@ def _get_records_query(
                 if interval and agg_funcs and use_pre_aggregates:
                     # For pre-aggregates, use column names as-is since they should already be
                     # the correct names from the continuous aggregate view
+                    source_time_col = timestamp_column or self.time_column
```
Eyal-Danieli:
```suggestion
                    source_time_col = time_col
```

mlrun/model_monitoring/db/tsdb/timescaledb/timescaledb_schema.py
```diff
@@ -302,9 +307,10 @@ def _get_records_query(
...
                     modified_columns = []
                     for col in columns_to_filter:
-                        if col == self.time_column:
+                        if col == source_time_col:
```
Eyal-Danieli: I think you don't need that `source_time_col` at all and you can just use the `time_col`
```suggestion
                        if col == time_col:
```

Eyal-Danieli (commented):
nice!
few comments

mlrun/model_monitoring/db/tsdb/timescaledb/timescaledb_schema.py
alxtkr77: Fixed

mlrun/model_monitoring/db/tsdb/timescaledb/timescaledb_schema.py
alxtkr77: Fixed

mlrun/model_monitoring/db/tsdb/timescaledb/timescaledb_schema.py
alxtkr77: Fixed

Eyal-Danieli (approved):

PR #8968, opened 2025-11-24
[Formatting] Upgrade Ruff target version to py3.11

Yacouby:
### 📝 Description
Upgrade Ruff's target version from Python 3.9 to 3.11 and modernize the Python codebase to adopt Python 3.10/3.11 type hinting improvements.

---

### 🛠️ Changes Made
- Updated `pyproject.toml` to set Ruff target version to Python 3.11.

- Ignored `UP007` in Ruff linter:
  While PEP 604 encourages replacing `Union[X, Y]` with the` X | Y` syntax, Ruff classifies this rule as unsafe.
   Many third-party types used across mlrun - e.g., TensorFlow, TensorSpec, PyTorch, and scikit-learn classes - do not support the | operator at runtime.
  Converting these annotations would cause runtime TypeErrors.
  reference in the Fix safety section: 
  https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/

- Updated datetime handling to use `UTC` alias instead of `timezone.utc`
   reference: 
   https://docs.astral.sh/ruff/rules/datetime-timezone-utc/

- Standardized imports of type related classes (e.g., Callable, Iterable, Mapping) to use `collections.abc` instead of `typing`, as applied automatically by Ruff formatter.
  reference:
  https://peps.python.org/pep-0585/

- Cleaned up unnecessary import typing where modern syntax made it redundant.

---

### ✅ Checklist

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11554
- Design docs links:
- External links: 

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ requires = [
 ]
 
 [tool.ruff]
-target-version = "py39"
+target-version = "py311"
 required-version = ">=0.8.0"
 
 [tool.ruff.lint]
@@ -19,6 +19,7 @@ extend-select = [
     "B006", # mutable default arguments
     "RUF013",  # forbid implicit `Optional` type hints
 ]
+ignore = ["UP007"]
 explicit-preview-rules = true
 
 [tool.ruff.lint.isort]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,8 @@ extend-select = [
     "B006", # mutable default arguments
     "RUF013",  # forbid implicit `Optional` type hints
 ]
+# Ignoring UP007 for now. See ticket ML-11573 for investigating runtime-safety of converting certain type
+# annotations to PEP 604 (`X | Y`) syntax.
 ignore = ["UP007"]
 explicit-preview-rules = true
 
--- a/automation/scripts/get_demos.py
+++ b/automation/scripts/get_demos.py
@@ -20,7 +20,7 @@ import shutil
 import sys
 import tarfile
 from concurrent.futures import ThreadPoolExecutor, as_completed
-from datetime import datetime, timezone
+from datetime import UTC, datetime
 from time import sleep
 
 import requests
@@ -256,7 +256,7 @@ def create_manifest(mlrun_version, demo_versions):
     """Create a manifest file with version information for all downloaded demos."""
     manifest = {
         "mlrun_version": mlrun_version,
-        "download_date": datetime.now(timezone.utc).isoformat(),
+        "download_date": datetime.now(UTC).isoformat(),
         "github_org": GITHUB_ORG,
         "demos": demo_versions,
     }
--- a/mlrun/artifacts/dataset.py
+++ b/mlrun/artifacts/dataset.py
@@ -366,9 +366,9 @@ def get_df_stats(df):
     for col, values in df.describe(include="all").items():
         stats_dict = {}
         for stat, val in values.dropna().items():
-            if isinstance(val, (float, np.floating, np.float64)):
+            if isinstance(val, float | np.floating | np.float64):
                 stats_dict[stat] = float(val)
-            elif isinstance(val, (int, np.integer, np.int64)):
+            elif isinstance(val, int | np.integer | np.int64):
                 stats_dict[stat] = int(val)
             else:
                 stats_dict[stat] = str(val)
--- a/mlrun/artifacts/plots.py
+++ b/mlrun/artifacts/plots.py
@@ -42,7 +42,7 @@ class PlotArtifact(Artifact):
         import matplotlib
 
         if not self.spec.get_body() or not isinstance(
-            self.spec.get_body(), (bytes, matplotlib.figure.Figure)
+            self.spec.get_body(), bytes | matplotlib.figure.Figure
         ):
             raise ValueError(
                 "matplotlib fig or png bytes must be provided as artifact body"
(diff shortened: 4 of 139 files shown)


pyproject.toml
```diff
@@ -19,6 +19,7 @@ extend-select = [
     "B006", # mutable default arguments
     "RUF013",  # forbid implicit `Optional` type hints
 ]
+ignore = ["UP007"]
```
jond01, line 24: I reviewed all the current changes - they look fine.

I don't understand the reason for ignoring this rule, and I'm quite sure we don't want to ignore it.
This is one of the key typing improvements in Python 3.10 and 3.11.

rokatyy (commented):
Nice! 

Minor comments

rokatyy (commented):
Nice!

Minor comments

pyproject.toml
```diff
@@ -4,7 +4,7 @@ requires = [
 ]
 
 [tool.ruff]
-target-version = "py39"
+target-version = "py311"
 required-version = ">=0.8.0"
```
rokatyy, line 8: there is already 0.14 available https://github.com/astral-sh/ruff/releases/tag/0.14.6, so while we are here, let's bump this dependency as well

pyproject.toml
rokatyy, line 24: add comment explaining what it does (the same as in pr description)

rokatyy (commented):
Nice!

Minor comments

jond01:
> * Ignored `UP007` in Ruff linter:
>   While PEP 604 encourages replacing `Union[X, Y]` with the` X | Y` syntax, Ruff classifies this rule as unsafe.
>   Many third-party types used across mlrun - e.g., TensorFlow, TensorSpec, PyTorch, and scikit-learn classes - do not support the | operator at runtime.
>   Converting these annotations would cause runtime TypeErrors.
>   reference in the Fix safety section:
>   https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/

The rule is not "unsafe" - only the fix could be unsafe.
Please provide a concrete example from our codebase for such a problem.

pyproject.toml
jond01: https://github.com/mlrun/mlrun/pull/8866 should handle this on a monthly basis :)

Yacouby:
> > * Ignored `UP007` in Ruff linter:
> >   While PEP 604 encourages replacing `Union[X, Y]` with the` X | Y` syntax, Ruff classifies this rule as unsafe.
> >   Many third-party types used across mlrun - e.g., TensorFlow, TensorSpec, PyTorch, and scikit-learn classes - do not support the | operator at runtime.
> >   Converting these annotations would cause runtime TypeErrors.
> >   reference in the Fix safety section:
> >   https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/
> 
> The rule is not "unsafe" - only the fix could be unsafe. Please provide a concrete example from our codebase for such a problem.

@jond01 
This is one example : https://github.com/mlrun/mlrun/blob/ad30cfd730b60385f2ce903d8b2356fe8f39cf5a/mlrun/frameworks/tf_keras/model_handler.py#L44
and we have a hundreds like this. ( in the framework dir and more other )

In this example both Tensor and TensorSpec are not valid operands for | , and Python 3.11 will raise:

`TypeError: unsupported operand type(s) for |: 'type' and 'type'`

This is exactly the kind of unsafe rewrite Ruff warns about in the UP007 rule's fix safety section.

For more examples you can also look at the failures in a previous PR that I had where this rule was not ignored : https://github.com/mlrun/mlrun/actions/runs/19649251640/job/56272037188

jond01:
> > > * Ignored `UP007` in Ruff linter:
> > >   While PEP 604 encourages replacing `Union[X, Y]` with the` X | Y` syntax, Ruff classifies this rule as unsafe.
> > >   Many third-party types used across mlrun - e.g., TensorFlow, TensorSpec, PyTorch, and scikit-learn classes - do not support the | operator at runtime.
> > >   Converting these annotations would cause runtime TypeErrors.
> > >   reference in the Fix safety section:
> > >   https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/
> > 
> > 
> > The rule is not "unsafe" - only the fix could be unsafe. Please provide a concrete example from our codebase for such a problem.
> 
> @jond01 This is one example :
> 
> https://github.com/mlrun/mlrun/blob/ad30cfd730b60385f2ce903d8b2356fe8f39cf5a/mlrun/frameworks/tf_keras/model_handler.py#L44
> 
> 
> and we have a hundreds like this. ( in the framework dir and more other )
> In this example both Tensor and TensorSpec are not valid operands for | , and Python 3.11 will raise:
> 
> `TypeError: unsupported operand type(s) for |: 'type' and 'type'`
> 
> This is exactly the kind of unsafe rewrite Ruff warns about in the UP007 rule's fix safety section.
> 
> For more examples you can also look at the failures in a previous PR that I had where this rule was not ignored : https://github.com/mlrun/mlrun/actions/runs/19649251640/job/56272037188

I tried the following in the same `mlrun` Python 3.11 version I have:

```py
import numpy as np
import tensorflow as tf
import keras

IOSample = tf.Tensor | tf.TensorSpec | keras.KerasTensor | np.ndarray
```

It works just fine.
What issue do you encounter?

Yacouby:
After offline discussion with @jond01 , we decided to open a ticket (https://iguazio.atlassian.net/browse/ML-11573) to investigate this further, as it is out of scope for the current Python 3.9 deprecation feature.

jond01 (approved):

rokatyy (approved):
LGTM

PR #8969, opened 2025-11-24
[RemoteModels] Support MRS Hf thread pool with 2 models

tomerm-iguazio:
### 📝 Description
Issue: When loading two different pipelines in a thread pool, we encountered a race condition that caused problems during client creation.
Fix: Add a thread-level lock to prevent concurrent pipeline initialization.

---

### 🛠️ Changes Made
Add a thread lock  to address the race in client creation.
Add system + integration tests.

---

### ✅ Checklist
  - [x] I updated the relevant Jira ticket for documentation

---

### 🧪 Testing
System and integration test of 2 models in MRS, with all execution mechanisms.

---

### 🔗 References
- Ticket link: [ML-11517](https://iguazio.atlassian.net/browse/ML-11517)
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes

[ML-11517]: https://iguazio.atlassian.net/browse/ML-11517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/integration/model_providers/huggingface/test_huggingface.py
```diff
@@ -270,7 +272,8 @@ def test_custom_invoke(self, use_datastore_profile):
 class TestHuggingFaceAIModel(TestBasicHuggingFaceProvider):
     @pytest.mark.parametrize(
         "execution_mechanism",
-        ["naive", "process_pool", "dedicated_process", "thread_pool"],
+        # ["naive", "process_pool", "dedicated_process", "thread_pool"],
+        ["thread_pool"],
```
davesh0812: please change back

tests/system/datastore/model_providers/test_huggingface.py
```diff
@@ -81,12 +84,18 @@ def test_basic_huggingface_model_runner(self, execution_mechanism):
...
+            "limits": {"cpu": "7", "memory": "20Gi"},
+            "requests": {"cpu": "25m", "memory": "1Mi"},
         }
         function.spec.max_replicas = (
             1  # to avoid allocating extended resources to multiple pods
         )
+
+        # removed worker_timeout=500
+        # now removed gateway_timeout=600
+        function.with_http(worker_timeout=1000, workers=None)
```
davesh0812: please explain why you are using `workers=None` and create a new bug for this issue. (add it's number here as well) .

I saw you are using it in few please explain it in all the relevant places.

mlrun/datastore/model_provider/huggingface_provider.py
```diff
@@ -224,7 +227,8 @@ def load_client(self) -> None:
 
             self.options["model_kwargs"] = self.options.get("model_kwargs", {})
             self.options["model_kwargs"]["local_files_only"] = True
-            self._client = pipeline(model=self.model, **self.options)
+            with self._client_lock:
+                self._client = pipeline(model=self.model, **self.options)
```
gtopper, line 231: What's the purpose of this lock?

davesh0812: The reason is that the pipeline initialization is not thread safe.
But I think this might not cover all the cases as you (@tomerm-iguazio) already saw... so maybe we should raise an error in case the user wants to use the same HF model from diffrent threads.. and to advise to use the process pool mechanism instead.

royischoss (approved):
LGTM

mlrun/datastore/model_provider/huggingface_provider.py
assaf758: Can you please add a comment on the jira what is still failing with this fix?
Also, if we end up thinking that we need to advise the user tha he should use the process mechanism when reusing the same HF model, maybe we should not skip this locking?

davesh0812: it was a miss understanding between us. This fix should do the trick.

PR #8981, opened 2025-11-27
[Setup] Python 3.11 alignment and cleanups

Yacouby:
### 📝 Description
This PR removes all remaining Python 3.9 support from the our codebase (following the PR: https://github.com/mlrun/mlrun/pull/8948 ).

---

### 🛠️ Changes Made
- Dropped Python 3.9 support by requiring Python 3.11+ in `setup.py` and removing the Python 3.9 classifier.
- Patch remote script: remove building and pushing of python 3.9 images.
- Pipelines-adapters: updated the python requires constraint from `>=3.9, <3.12` to `>=3.11, <3.12`
- Replaced the custom `StrEnum` with an alias to Python 3.11’s built-in `enum.StrEnum`
- Use Enum instead of StrEnum in `_TDEngineColumn` because the enum members hold _TDEngineColumnType objects not strings which StrEnum in Python 3.11 no longer accepts.

---

### ✅ Checklist

---

### 🧪 Testing
Verified locally that installing MLRun under Python 3.9 fails as expected:

image

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11421
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/mlrun/common/types.py
+++ b/mlrun/common/types.py
@@ -14,18 +14,10 @@
 
 import enum
 
-
-# TODO: From python 3.11 StrEnum is built-in and this will not be needed
-class StrEnum(str, enum.Enum):
-    def __str__(self):
-        return self.value
-
-    def __repr__(self):
-        return self.value
+# Alias to Python's built-in StrEnum (Python 3.11+)
+StrEnum = enum.StrEnum
 
 
-# Partial backport from Python 3.11
-# https://docs.python.org/3/library/http.html#http.HTTPMethod
 class HTTPMethod(StrEnum):
     GET = "GET"
     POST = "POST"
--- a/pipeline-adapters/mlrun-pipelines-kfp-v1-8/pyproject.toml
+++ b/pipeline-adapters/mlrun-pipelines-kfp-v1-8/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "mlrun-pipelines-kfp-v1-8"
-version = "0.5.9"
+version = "0.6.0"
 description = "MLRun Pipelines adapter package for providing KFP 1.8 compatibility"
 readme = "README.md"
 requires-python = ">=3.11, <3.12"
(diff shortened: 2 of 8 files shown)


pipeline-adapters/mlrun-pipelines-kfp-v1-8/pyproject.toml
```diff
@@ -17,8 +17,7 @@ dependencies = [
 
 [project.optional-dependencies]
 kfp = [
-  "kfp~=1.8.22; python_version < '3.11'",
-  "kfp~=1.8.23; python_version >= '3.11'",
+  "kfp~=1.8.23",
 ]
```
liranbg, line 21: you bump the toml, need to bump the version as well

pipeline-adapters/mlrun-pipelines-kfp-v1-8/pyproject.toml
```diff
@@ -1,9 +1,9 @@
 [project]
 name = "mlrun-pipelines-kfp-v1-8"
-version = "0.5.8"
+version = "0.5.9"
```
liranbg: bump to 0.6

pipeline-adapters/mlrun-pipelines-kfp-v2/pyproject.toml
```diff
@@ -1,9 +1,9 @@
 [project]
 name = "mlrun-pipelines-kfp-v2"
-version = "0.5.3"
+version = "0.5.4"
```
liranbg:
```suggestion
version = "0.6.0"
```

liranbg: the change you are doing here is not a "patch" modification but rather a minor since we now dropping backwards compatibility support of python 3.9 - meaning, at least updating minor here.

liranbg (changes requested):

mlrun/common/types.py
```diff
@@ -14,18 +14,10 @@
 
 import enum
 
+# Alias to Python's built-in StrEnum (Python 3.11+)
+StrEnum = enum.StrEnum
```
adamdelman, line 18: Why keep the alias, wouldn't it be to refer to the correct type directly?

adamdelman (changes requested):
Overall LGTM, minor comment.

mlrun/common/types.py
liranbg: it might be still in use be other components using mlrun as it is on `common`

liranbg (approved):

Eyal-Danieli (approved):

adamdelman (approved):

PR #8983, opened 2025-11-27
[Nuclio] Add batching support to with_http

rokatyy:
### 📝 Description
Adds batching support to `with_http`. By default, the `NuclioSpec` still work with no batching, and does not serialize this object at all if batching is not enabled, to support older versions of nuclio.

The `RemoteRuntime.with_http` method accept a new parameter `batching_spec: BatchingSpec` which allows setting batching mode on the HTTP trigger (nuclio only supports a single HTTP trigger per function).

Disabling this mode can be done by either calling `with_http` and setting `batching_spec.enabled=False`, or setting `batching_spec=None` which will nullify any configuration of the async spec in the function’s spec and will revert to the default configuration (i.e. no batching).

---

### ✅ Checklist
- [x] I updated the documentation (with_http docs)

---

### 🧪 Testing
system(tested on my system and made sure config in nuclio is correct - screenshot attached below), unit tests

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11434
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes
image

### DIFF
--- a/mlrun/common/schemas/function.py
+++ b/mlrun/common/schemas/function.py
@@ -140,3 +140,26 @@ class Function(pydantic.v1.BaseModel):
 
     class Config:
         extra = pydantic.v1.Extra.allow
+
+
+class BatchingSpec(pydantic.v1.BaseModel):
+    # Set to True to enable batching
+    enabled: bool
+    # Maximal events to batch together
+    batch_size: typing.Optional[int]
+    # How long to wait before sending the batch
+    timeout: typing.Optional[str]
+
+    def get_nuclio_batch_config(self):
+        if not self.enabled:
+            return None
+
+        config = {"mode": "enable"}
+
+        if self.batch_size:
+            config["batchSize"] = self.batch_size
+
+        if self.timeout:
+            config["timeout"] = self.timeout
+
+        return config
--- a/mlrun/common/schemas/function.py
+++ b/mlrun/common/schemas/function.py
@@ -145,9 +145,9 @@ class Function(pydantic.v1.BaseModel):
 class BatchingSpec(pydantic.v1.BaseModel):
     # Set to True to enable batching
     enabled: bool
-    # Maximal events to batch together
+    # Maximal events to batch together. Default size is 10.
     batch_size: typing.Optional[int]
-    # The maximum amount of time to wait before processing the batch.
+    # The maximum amount of time to wait before processing the batch. Default timeout is 1s.
     # Once this time passes, the batch is processed even if it hasn’t reached the full batch size.
     timeout: typing.Optional[str]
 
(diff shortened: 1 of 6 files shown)


mlrun/common/schemas/function.py
```diff
@@ -140,3 +140,26 @@ class Function(pydantic.v1.BaseModel):
...
     class Config:
         extra = pydantic.v1.Extra.allow
+
+
+class BatchingSpec(pydantic.v1.BaseModel):
+    # Set to True to enable batching
+    enabled: bool
+    # Maximal events to batch together
+    batch_size: typing.Optional[int]
+    # How long to wait before sending the batch
```
liranbg: I think from mlrun pov that's not sending it's about processing - right? Because the http would be sent all together but processed in batches

rokatyy: @liranbg good point, rephrased ✅

mlrun/runtimes/nuclio/function.py
```diff
@@ -484,6 +485,8 @@ def with_http(
         :param trigger_name:    alternative nuclio trigger name
         :param annotations:     key/value dict of ingress annotations
         :param extra_attributes: key/value dict of extra nuclio trigger attributes
+        :param batching_spec: BatchingSpec objects that defines batching configuration.
```
TomerShor:
```suggestion
        :param batching_spec: BatchingSpec object that defines batching configuration.
```

mlrun/common/schemas/function.py
```diff
@@ -140,3 +140,27 @@ class Function(pydantic.v1.BaseModel):
 
     class Config:
         extra = pydantic.v1.Extra.allow
+
+
+class BatchingSpec(pydantic.v1.BaseModel):
+    # Set to True to enable batching
+    enabled: bool
+    # Maximal events to batch together
+    batch_size: typing.Optional[int]
```
TomerShor: Nuclio has a default batch size. Documentation-wise, I think it's more convenient to show it here too

TomerShor (commented):
Nice one!

mlrun/common/schemas/function.py
rokatyy: added

mlrun/common/schemas/function.py
```diff
@@ -140,3 +140,27 @@ class Function(pydantic.v1.BaseModel):
 
     class Config:
         extra = pydantic.v1.Extra.allow
+
+
+class BatchingSpec(pydantic.v1.BaseModel):
```
liranbg, line 145: Lastly this can be used only by a specific nuclio version. Should be reflected with a decorator as a feature gate, right?

rokatyy: 💯

liranbg (approved):

TomerShor (approved):

PR #8986, opened 2025-11-30
[Auth] Change user token auth secret naming convention

liranbg:
### 📝 Description

The current implementation assumes username and token name part of the secret name - which can lead to k8s validation issues when username contains @ or token name has some `_` characters which is not k8s valid for secret names.

---

### 🛠️ Changes Made

- Changed secret name to be hashed-base rather than username/token name
- To allow retrieving user tokens or user secrets, we leverage secret labels and getting those secret by using username/token name labels. this change is opaque to end users.
- Relaxed some spammy logs

---

### ✅ Checklist

---

### 🧪 Testing

- Unit testings
- Manual testings

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11576 https://iguazio.atlassian.net/browse/ML-11510
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [n] No

---

### DIFF
--- a/server/py/framework/utils/singletons/k8s.py
+++ b/server/py/framework/utils/singletons/k8s.py
@@ -1205,7 +1205,6 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
         }
 
         create = False
-        secret_data = self._encode_user_token(token_name, token, expiration)
         k8s_secret = self._get_user_token_secret(username, token_name, namespace)
         if not k8s_secret:
             create = True
@@ -1216,7 +1215,7 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
                 labels=labels,
                 namespace=namespace,
                 secret_name=self._resolve_auth_secret_name(username, token_name),
-                secrets=secret_data,
+                secrets=self._encode_user_token(token_name, token, expiration),
                 encoded=True,
             )
             return mlrun.common.schemas.SecretEventActions.created
@@ -1227,7 +1226,7 @@ class K8sHelper(mlsecrets.SecretProviderInterface):
                 k8s_secret=k8s_secret,
                 namespace=namespace,
                 secret_name=self._resolve_auth_secret_name(username, token_name),
-                secrets=secret_data,
+                secrets=self._encode_user_token(token_name, token, expiration),
                 encoded=True,
             )
             return mlrun.common.schemas.SecretEventActions.updated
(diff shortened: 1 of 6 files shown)


server/py/framework/utils/singletons/k8s.py
```diff
@@ -1199,34 +1199,23 @@ def store_user_token_secret(
...
         }
 
-        logger.debug(
-            "Preparing to store user token secret",
-            secret_name=secret_name,
-            namespace=namespace,
-        )
-
+        create = False
         secret_data = self._encode_user_token(token_name, token, expiration)
```
elbamit: Encoding here becomes useless if the code doesn't access any of the if conditions below. You can encode only if you're going to create/update

elbamit (approved):

server/py/framework/utils/singletons/k8s.py
```diff
@@ -1199,34 +1199,23 @@ def store_user_token_secret(
...
-            namespace=namespace,
-            labels=labels,
-            silent=True,
-        )
-
+        k8s_secret = self._get_user_token_secret(username, token_name, namespace)
         if not k8s_secret:
+            create = True
+
+        if create:
```
TomerShor, line 1213: why not keep it in the `if not k8s_secret:` clause?

server/py/framework/utils/singletons/k8s.py
```diff
@@ -1444,71 +1419,33 @@ def get_user_token_secret_value(
...
         :return: The offline token string.
         :raises mlrun.errors.MLRunNotFoundError: If the token is not found in the secret.
         :raises mlrun.errors.MLRunRuntimeError: If decoding/parsing fails.
         """
+        username = k8s_secret.metadata.labels[
+            mlrun_constants.MLRunInternalLabels.auth_username
+        ]
+        token_name = k8s_secret.metadata.labels[
+            mlrun_constants.MLRunInternalLabels.auth_token_name
+        ]
```
TomerShor: If the labels are missing (from whatever reason) you'll get an exception here.
Move it into the `try` as well

server/py/framework/utils/singletons/k8s.py
liranbg: just for explicitness - thought it make it a bit more readable

server/py/framework/utils/singletons/k8s.py
liranbg: that's exactly the point - if the secrets are missing those *important* / *mandatory* labels - we must explode. the contract here is secret must have these 2 labels in order for them to be reachable. if mlrun-api created them malformed, we wont be able to find them anyway.

TomerShor: I completely agree, I just thing that moving it into the `try` will make the `except Exception as exc:` catch it and return a more descriptive error

quaark (approved):
🚀

PR #9013, opened 2025-12-04
[KubejobRuntime] Secret token mounting for KubejobRuntime

elbamit:
### 📝 Description

This PR updates the MLRun API to mount the Kubernetes secret corresponding to IG4’s offline token `default` (for the running user) as a file inside the run pod at a predefined folder owned by MLRun. The folder path is exposed to the pod via an environment variable so that the SDK running inside the pod can access the token file.

The `ServerSideLauncher._enrich_runtime` method saves the token name in `run.spec.auth`. The `KubejobRuntimeHandler` then uses this token name to determine the Kubernetes secret name and mounts it on the runtime before pod creation.

As a result, `volumes` and `volume_mounts` are added to both the runtime and pod specifications.

The `MLRUN_AUTH_OFFLINE_TOKEN` env var that was used in https://github.com/mlrun/mlrun/pull/8840 is removed and not needed anymore.

---

### 🛠️ Changes Made

- Removed `MLRUN_AUTH_OFFLINE_TOKEN` and set `MLRUN_AUTH_WITH_OAUTH_TOKEN__TOKEN_FILE` during auth env enrichment for pods
- Added an `auth` field to `RunSpec` and to `MLClientCtx`(needed for it to be saved in DB)
- Find secret's name and mount it as a file to the runtime during `KubejobRuntimeHandler.run()`

---

### ✅ Checklist

---

### 🧪 Testing
  
  
- Unit test for mount secret token to runtime and for ensuring IG4 auth envs are set to runtime
- Manual tests that job pod is deployed and authenticates with mlrun using the file

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11583
- Design docs links: https://iguazio.atlassian.net/wiki/spaces/MLRUN/pages/416121541
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes

In future PR's, token other than `default` will be allowed and resolved


server/py/services/api/launcher.py
```diff
@@ -695,6 +696,15 @@ def _validate_retry(runtime_kind: str, retry: Optional["mlrun.model.Retry"]):
...
                 )
 
+    # TODO implement tests during ML-11600
+    def _resolve_and_validate_token_name_from_secret(self, run: mlrun.run.RunObject):
+        if run.spec.auth.get("token_name"):
+            # TODO perform token name validation ML-11600
+            pass
+        else:
+            # TODO perform token name resolution and validation ML-11600
+            run.spec.auth["token_name"] = "default"
```
liranbg: duplicate TODO, simplify it. you can also write a function that "validate" the token name and put the todo there.

mlrun/common/constants.py
```diff
@@ -38,6 +38,9 @@
 JOB_TYPE_RERUN_WORKFLOW_RUNNER = "rerun-workflow-runner"
 MLRUN_ACTIVE_PROJECT = "MLRUN_ACTIVE_PROJECT"
 
+MLRUN_AUTH_SECRET_PATH = "/var/mlrun-secrets/auth"
```
liranbg:
```suggestion
MLRUN_JOB_AUTH_SECRET_PATH = "/var/mlrun-secrets/auth"
```


job/functions/everything that mlrun spins of curse but mainly for mlrun run pods *as opposed* to mlrun IDE or so.

mlrun/auth/utils.py
```diff
@@ -330,21 +331,15 @@ def enrich_auth_env(
...
-                token_name=_default_token_name,
-                username=auth_info.username,
-            )
-            env["MLRUN_AUTH_OFFLINE_TOKEN"] = secret.token
-
         env["MLRUN_AUTH_WITH_OAUTH_TOKEN__ENABLED"] = "true"
-        env["MLRUN_AUTH_TOKEN_ENDPOINT"] = (
-            mlrun.mlconf.iguazio_api_url + "/api/v1/refresh-access-token"
+        env["MLRUN_AUTH_TOKEN_ENDPOINT"] = urljoin(
+            mlrun.mlconf.iguazio_api_url, "/api/v1/refresh-access-token"
```
rokatyy: there is a reason why we should be careful with `urljoin`:
```
>>> urljoin("no-http-prefix-link.com/", "/mypath")
'/mypath'
```

Do we always specify it with http/https prefix?

mlrun/auth/utils.py
```diff
@@ -330,21 +331,15 @@ def enrich_auth_env(
...
         )
         env["MLRUN_HTTPDB__HTTP__VERIFY"] = str(
             mlrun.mlconf.iguazio_api_ssl_verify
         ).lower()
+        env["MLRUN_AUTH_WITH_OAUTH_TOKEN__TOKEN_FILE"] = str(
+            PurePosixPath(mlrun.common.constants.MLRUN_AUTH_SECRET_PATH)
+            / mlrun.common.constants.MLRUN_AUTH_SECRET_FILE
+        )
```
rokatyy:
```suggestion
       env["MLRUN_AUTH_WITH_OAUTH_TOKEN__TOKEN_FILE"] = os.path.join(
    mlrun.common.constants.MLRUN_AUTH_SECRET_PATH,
    mlrun.common.constants.MLRUN_AUTH_SECRET_FILE
)
```

server/py/services/api/launcher.py
```diff
@@ -695,6 +696,15 @@ def _validate_retry(runtime_kind: str, retry: Optional["mlrun.model.Retry"]):
                     f"must be less than {staleness_threshold_seconds} seconds, got {max_delay} seconds"
                 )
 
+    # TODO implement tests during ML-11600
+    def _resolve_and_validate_token_name_from_secret(self, run: mlrun.run.RunObject):
+        if run.spec.auth.get("token_name"):
```
rokatyy:
```suggestion
       if not run.spec.auth.get("token_name"):
           # TODO perform token name resolution
           run.spec.auth["token_name"] = "default"
       # TODO perform token name validation
        
```

mlrun/model.py
```diff
@@ -1008,6 +1008,7 @@ def __init__(
         tolerations=None,
         affinity=None,
         retry=None,
+        auth=None,
```
rokatyy, line 1011: why is this needed?

rokatyy (commented):
Looking good!

Just couple of nitpickings and questions

mlrun/model.py
elbamit: To save the auth as part of the run object spec

mlrun/auth/utils.py
elbamit: I think so, however to be on the safe side I'll change it to `os.path.join`

server/py/services/api/launcher.py
elbamit: Not quite, because during token resolution we also want to validate each token before choosing it. I'll change the logic to something else that takes care of this

server/py/services/api/launcher.py
```diff
@@ -695,6 +696,32 @@ def _validate_retry(runtime_kind: str, retry: Optional["mlrun.model.Retry"]):
...
+        auth = run.spec.auth or {}
+
+        if auth.get("token_name"):
+            self._validate_token_name(
+                auth["token_name"],
+                explicit=True,
+            )
+            return
+
+        run.spec.auth["token_name"] = "default"
```
Yacouby, line 710: could this break runs if we fall back to "default" even when that token is missing or invalid?

mlrun/auth/utils.py
```diff
@@ -330,21 +329,15 @@ def enrich_auth_env(
     :param auth_info: The AuthInfo object containing authentication details.
```
Yacouby: you have to update the docstring after removing those 2 params

server/py/services/api/launcher.py
Yacouby: + you are using hardcoded value , what about creating a constant for it ?

mlrun/auth/utils.py
```diff
@@ -330,21 +329,15 @@ def enrich_auth_env(
...
-                token_name=_default_token_name,
-                username=auth_info.username,
-            )
-            env["MLRUN_AUTH_OFFLINE_TOKEN"] = secret.token
-
         env["MLRUN_AUTH_WITH_OAUTH_TOKEN__ENABLED"] = "true"
-        env["MLRUN_AUTH_TOKEN_ENDPOINT"] = (
-            mlrun.mlconf.iguazio_api_url + "/api/v1/refresh-access-token"
+        env["MLRUN_AUTH_TOKEN_ENDPOINT"] = os.path.join(
+            mlrun.mlconf.iguazio_api_url, "/api/v1/refresh-access-token"
```
Yacouby: I think this is wrong since the second argument starts with a `/`, it ignores the first part and will always return "/api/v1/refresh-access-token", which is not intended.

server/py/services/api/runtime_handlers/kubejob.py
```diff
@@ -60,6 +60,13 @@ def run(
             runtime.store_run(run)
         new_meta = self._get_meta(runtime, run)
 
+        if mlrun.mlconf.is_iguazio_v4_mode():
+            self._mount_secret_token_to_runtime(
+                runtime,
+                token_name=run.spec.auth.get("token_name"),
+                username=auth_info.username,
```
Yacouby, line 68: what happens if the username is `None` ?

server/py/services/api/tests/unit/runtime_handlers/test_kubejob.py
```diff
@@ -1008,3 +1008,43 @@ def _mock_list_resources_pods(self, pod=None):
         pod = pod or self.completed_job_pod
         mocked_responses = self._mock_list_namespaced_pods([[pod]])
         return mocked_responses[0].items
+
+    def test_mount_secret_token_to_runtime(self):
```
Yacouby: need to add more tests here as you check only the happy path where the secret exists. you should also cover cases like missing secrets, volume with the same name already exists
+
verifying the generate env variable is set in the runtime env

Yacouby (changes requested):
Suggestions

mlrun/auth/utils.py
elbamit: Will remove the leading `/`

server/py/services/api/runtime_handlers/kubejob.py
elbamit: Why would username be None?

server/py/services/api/launcher.py
elbamit: Currently we're always injecting default.
During ML-11600 we'll make the token resolution logic as described in the HLD https://iguazio.atlassian.net/wiki/spaces/MLRUN/pages/416121541/Secret+Token+Job+Function+Mounts+HLD#Token-name-resolution
So this is just temporary and will change.

server/py/services/api/runtime_handlers/kubejob.py
Yacouby: `auth_info` itself can be None

elbamit: When? In CE?

Yacouby: I'm not sure, I just asked that because I see that the passed `auth_info` is optional and may be None

elbamit: I wasn't able to find any other usage of `KubeJobRuntimeHandler.run()` other than in the server side launcher

server/py/services/api/launcher.py
```diff
@@ -310,6 +310,7 @@ def _enrich_run(
 
         self._handle_retry(run)
         run = self._pre_run_image_pull_secret_enrichment(run)
+        self._resolve_and_validate_token_name_from_secret(run)
```
```suggestion
        self._enrich_and_validate_auth_token_name(run)
```

server/py/services/api/launcher.py
```diff
@@ -695,6 +696,32 @@ def _validate_retry(runtime_kind: str, retry: Optional["mlrun.model.Retry"]):
...
+        run.spec.auth["token_name"] = "default"
+        # for token_name in services.api.crud.Secrets().list_secret_tokens(self._auth_info.username):
+        #     if self._validate_token_name(token_name):
+        #         run.spec.auth["token_name"] = token_name
+        #         return
+
+        # raise ValueError(
+        #     "No valid authentication token found. "
+        #     "Please create a valid offline token or provide one explicitly."
+        # )
```
liranbg: why is this commented?

liranbg (commented):
minor comments

server/py/services/api/launcher.py
elbamit: removed it

server/py/services/api/launcher.py
```diff
@@ -695,6 +696,22 @@ def _validate_retry(runtime_kind: str, retry: Optional["mlrun.model.Retry"]):
...
+            self._validate_token_name(
+                auth["token_name"],
+                explicit=True,
+            )
+            return
+
+        run.spec.auth["token_name"] = "default"
+
+    def _validate_token_name(self, token_name: str, explicit: bool = False):
+        pass
```
liranbg, line 714: leave the TODO here, above this function

liranbg (approved):
very well. minor suggestion

liranbg (approved):

server/py/services/api/tests/unit/crud/test_secrets.py
```diff
@@ -1165,5 +1166,101 @@ def test_get_secret_token_not_found():
...
+            [{"mountPath": "/var/mlrun-secrets/auth", "name": "secret"}],
+            [
+                {
+                    "name": "secret",
+                    "secret": {"items": [], "secretName": "old-secret"},
+                }
+            ],
+            1,
+        ),
+        # Volume with a different name already exists (should add new one)
```
Yacouby, line 1185: both volumes has the the same name "other-volume", so this comment is wrong

server/py/services/api/tests/unit/crud/test_secrets.py
```diff
@@ -1165,5 +1166,101 @@ def test_get_secret_token_not_found():
...
+        (
+            [{"mountPath": "/some/other/path", "name": "other-volume"}],
+            [{"name": "other-volume", "other-volume": {"items": []}}],
```
Yacouby: is this valid Kubernetes schema ?

server/py/services/api/tests/unit/crud/test_secrets.py
```diff
@@ -1165,5 +1166,101 @@ def test_get_secret_token_not_found():
...
+
+
+def test_mount_secret_token_to_runtime_non_existing_secret():
+    token_name = "test-token"
+    username = "test-user"
+
+    runtime = mlrun.runtimes.kubejob.KubejobRuntime()
+    # # Runtime that inherit from BaseRuntime already have existing volumes/mounts
+    # runtime.spec.volume_mounts = []
+    # runtime.spec.volumes = []
```
Yacouby: this comment is a bit misleading, you say that runtime already have existing volumes, but then expect them to be empty ?

server/py/services/api/tests/unit/crud/test_secrets.py
```diff
@@ -1165,5 +1166,101 @@ def test_get_secret_token_not_found():
...
+
+    assert len(runtime.spec.volumes) == expected_secret_count
+
+
+def test_mount_secret_token_to_runtime_non_existing_secret():
+    token_name = "test-token"
+    username = "test-user"
+
+    runtime = mlrun.runtimes.kubejob.KubejobRuntime()
+    # # Runtime that inherit from BaseRuntime already have existing volumes/mounts
```
Yacouby:
```suggestion
    # Runtime that inherit from BaseRuntime already have existing volumes/mounts
```

Yacouby (changes requested):

server/py/services/api/tests/unit/crud/test_secrets.py
elbamit: One is the volumes, the other is the volume mounts.
The volume that the function add is called `secret`

Yacouby (approved):

rokatyy (approved):

PR #9050, opened 2025-12-14
[RemoteRuntime] Secret token mounting for RemoteRuntime

elbamit:
### 📝 Description

Add secret token mounting for RemoteRuntime functions.
Save `auth` as part of the function's spec. Use it to mount the secret into the runtime during function config compilation.

---

### 🛠️ Changes Made

- Added `auth` as part of `NuclioSpec`, `ApplicationSpec` and `ServingSpec`
- Enrich auth token name on the function during deployment, then mount secret to runtime during function config compilation.
- Refactor `mount_secret_token_to_runtime` to remove existing auth secret volumes/mounts
- Refactor `_enrich_and_validate_auth_token_name` to `enrich_and_validate_auth_token_name_on_object` to support both `runObject` and `remoteRuntime`

---

### ✅ Checklist

---

### 🧪 Testing
  
  
Unit tests + manual tests of deployment of nuclio/application/serving

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11584
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes

As part of https://iguazio.atlassian.net/browse/ML-11599, need to handle redeployment of a function with a different auth token name.


server/py/services/api/launcher.py
```diff
@@ -697,20 +697,31 @@ def _validate_retry(runtime_kind: str, retry: Optional["mlrun.model.Retry"]):
                 )
 
     # TODO In ML-11600, implement token name resolution and validation + tests
-    def _enrich_and_validate_auth_token_name(self, run: mlrun.run.RunObject):
-        auth = run.spec.auth or {}
+    def enrich_and_validate_auth_token_name_on_object(
```
```suggestion
    def enrich_and_validate_auth_token_name(
```
why renaming? also, function name should reflect what it does (enriches/validates the auth token) the added information is subjected to the argument - which can be omitted from the name for redundancy or (TMI)

server/py/services/api/tests/unit/runtimes/test_nuclio.py
```diff
@@ -2069,6 +2070,82 @@ def test_get_url(
                             else url.startswith("http")
                         )
 
+    def test_compile_function_config_iguazio_v4_mounts_auth_secret(self):
```
liranbg:
```suggestion
    def test_compile_function_config_with_auth_secret(self):
```

server/py/services/api/tests/unit/crud/test_secrets.py
```diff
@@ -1264,5 +1283,99 @@ def test_mount_secret_token_to_runtime_non_existing_secret():
...
+)
+def test_remove_auth_secret_volumes(
+    initial_volumes,
+    initial_mounts,
+    expected_volumes,
+    expected_mounts,
+):
+    runtime = mlrun.runtimes.kubejob.KubejobRuntime()
+    runtime.spec.volumes = initial_volumes.copy()
+    runtime.spec.volume_mounts = initial_mounts.copy()
```
liranbg: why do you need to copy them?

server/py/services/api/launcher.py
```diff
@@ -697,20 +697,31 @@ def _validate_retry(runtime_kind: str, retry: Optional["mlrun.model.Retry"]):
...
+        self, object: Union[mlrun.run.RunObject, mlrun.runtimes.RemoteRuntime]
+    ):
+        # Ensure the auth dictionary exists
```
```suggestion
```
code is self explanatory

server/py/services/api/launcher.py
```diff
@@ -697,20 +697,31 @@ def _validate_retry(runtime_kind: str, retry: Optional["mlrun.model.Retry"]):
...
+        provided_token_name = object.spec.auth.get("token_name")
 
-        run.spec.auth["token_name"] = "default"
+        # Resolve token name
+        if provided_token_name:
+            token_name = provided_token_name
+            self._validate_token_name(token_name, raise_error_on_failure=True)
+        else:
+            token_name = "default"
+            self._validate_token_name(token_name, raise_error_on_failure=False)
```
liranbg: is default token name a constant? I think it should be
also, comment on why you raise an exception?



```suggestion
        # explain me
        raise_error_on_failure = bool(provided_token_name)
        token_name = provided_token_name or "default"
        self._validate_token_name(token_name, raise_error_on_failure=False)
```

server/py/services/api/crud/secrets.py
```diff
@@ -839,6 +842,33 @@ def mount_secret_token_to_runtime(
...
+
+        # Identify volumes to remove
+        for vol in volumes:
+            if "secret" in vol:
+                secret = vol["secret"]
+                secret_name = secret.get("secretName", "")
+
+                # Pattern of auth secret volumes
+                if secret_name.startswith("mlrun-auth-secrets"):
+                    volumes_to_remove.add(vol["name"])
```
liranbg:
```suggestion
        for volume in volumes:
          if secret := volume.get("secret"):
            secret_name = secret.get("secretName", "")
            if secret_name.startswith("mlrun-auth-secrets"):
              volumes_to_remove.add(volume["name"])
```

liranbg: shorter, with full name

server/py/services/api/crud/secrets.py
```diff
@@ -839,6 +842,33 @@ def mount_secret_token_to_runtime(
...
+
+        # Filter out only the matched volumes
+        runtime.spec.volumes = [
+            v for v in volumes if v["name"] not in volumes_to_remove
+        ]
```
liranbg:
```suggestion
        runtime.spec.volumes = [
            volume
            for volume in volumes 
            if volume["name"] not in volumes_to_remove
        ]
```

use full name, makes it more readable

server/py/services/api/crud/secrets.py
```diff
@@ -839,6 +842,33 @@ def mount_secret_token_to_runtime(
...
+
+        # Filter out matching mounts
+        runtime.spec.volume_mounts = [
+            m for m in mounts if m["name"] not in volumes_to_remove
```
liranbg: make it readable

server/py/services/api/crud/secrets.py
```diff
@@ -823,6 +823,9 @@ def mount_secret_token_to_runtime(
             username=username, token_name=token_name
         )
 
+        # Remove any existing auth secret volumes/mounts
+        services.api.crud.secrets.Secrets._remove_auth_secret_volumes(runtime)
```
liranbg: seeing these calls (static method) from within same class static method) shows me that the decision to put the logic under `Secrets` class that it was a bit wrong.
I think we need to change it and move logic under runtime/etc. then we could do

`runtime.enrich_and_validate()`
`function.remove_auth_secret_volumes()`

using self, etc

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -249,7 +249,12 @@ def _compile_function_config(
 
     :return: function name, project name, nuclio function config
     """
-
+    if mlrun.mlconf.is_iguazio_v4_mode():
+        function = services.api.crud.secrets.Secrets.mount_secret_token_to_runtime(
+            function,
+            token_name=function.spec.auth.get("token_name"),
+            username=auth_info.username,
+        )
```
liranbg: e.g for https://github.com/mlrun/mlrun/pull/9050/changes#r2617421254

`function.mount_secret_token_to_runtime(token_name, auth_info)`

liranbg (changes requested):
Some suggestions and renamings, overall very well. I suggest to run some nuclio system tests to see how it plays and no regressions were made.

server/py/services/api/crud/secrets.py
```diff
@@ -839,6 +842,33 @@ def mount_secret_token_to_runtime(
...
+        volumes_to_remove = set()
+
+        # Identify volumes to remove
+        for vol in volumes:
+            if "secret" in vol:
+                secret = vol["secret"]
+                secret_name = secret.get("secretName", "")
+
+                # Pattern of auth secret volumes
+                if secret_name.startswith("mlrun-auth-secrets"):
```
rokatyy: use constant https://github.com/mlrun/mlrun/blob/feb28bdc39338bd93c6f750039b936cad19a7e94/mlrun/config.py#L705 and just `format(hashed_access_key="")` to get this prefix

server/py/services/api/crud/secrets.py
```diff
@@ -839,6 +842,33 @@ def mount_secret_token_to_runtime(
...
+        volumes = runtime.spec.volumes or []
+        mounts = runtime.spec.volume_mounts or []
+
+        volumes_to_remove = set()
+
+        # Identify volumes to remove
+        for vol in volumes:
+            if "secret" in vol:
+                secret = vol["secret"]
+                secret_name = secret.get("secretName", "")
```
rokatyy: using https://github.com/mlrun/mlrun/blob/4d6dfa3254490a4f1e747a0c2d518c80b4ec8f89/mlrun/utils/helpers.py#L568 might make it a bit cleaner

rokatyy (commented):
minor suggestions

server/py/services/api/crud/secrets.py
elbamit: @rokatyy's suggestion is more robust. Using her's instead

server/py/services/api/launcher.py
elbamit: @quaark I can use `mlconf.auth_with_oauth_token.token_name`, however if user sets this env var prior to importing mlrun, then runs/deploy a function, the token resolution logic will use that instead. 
I assume we don't want that behavior right?

elbamit: Ok so we'll make it a constant not a config

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -362,15 +362,20 @@ def _apply_escaped_config(config, parent_key, items: dict):
...
-def _resolve_env_vars(function, auth_info=None):
+def _enrich_config_spec(function, username):
     # Add secret configurations to function's pod spec, if secret sources were added.
     # Needs to be here, since it adds env params, which are handled in the next lines.
     # This only needs to run if we're running within k8s context. If running in Docker, for example, skip.
     if framework.utils.singletons.k8s.get_k8s_helper(
         silent=True
     ).is_running_inside_kubernetes_cluster():
-        _add_secrets_config_to_function_spec(function)
+        token_name = mlrun.utils.get_in(function.spec, "auth.token_name", None)
```
rokatyy, line 375: do we want to skip if token is empty? Or will it be handled later in the flow?

rokatyy (approved):
LGTM

server/py/services/api/crud/runtimes/nuclio/function.py
elbamit: Later in the flow it will try to list the secrets matching that token, but there won't be such, so we won't perform the mounting.

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -362,15 +362,20 @@ def _apply_escaped_config(config, parent_key, items: dict):
         mlrun.utils.update_in(config, f"{parent_key}.\\{key}\\", value)
 
 
-def _resolve_env_vars(function, auth_info=None):
+def _enrich_config_spec(function, username):
```
liranbg: why not passing auth_info entirely? a better practice as you might late on use its other fields

server/py/services/api/runtime_handlers/base.py
```diff
@@ -299,6 +299,8 @@ def add_secrets_to_spec_before_running(
         self,
         runtime: mlrun.runtimes.pod.KubeResource,
         project_name: Optional[str] = None,
+        token_name: Optional[str] = None,
+        username: Optional[str] = None,
```
liranbg: pass auth_info

server/py/services/api/runtime_handlers/base.py
```diff
@@ -402,7 +412,12 @@ def add_k8s_secrets_to_spec(
         runtime: mlrun.runtimes.pod.KubeResource,
         project_name: Optional[str] = None,
         encode_key_names: bool = True,
+        token_name: Optional[str] = None,
+        username: Optional[str] = None,
```
liranbg: pass auth_info

server/py/services/api/runtime_handlers/base.py
```diff
@@ -465,6 +480,35 @@ def add_k8s_secrets_to_spec(
         if not encode_key_names and secrets.keys():
             runtime.set_env("MLRUN_PROJECT_SECRETS_LIST", ",".join(secrets.keys()))
 
+    @staticmethod
+    def _mount_secret_token_to_runtime(
+        runtime: mlrun.runtimes.base.BaseRuntime, token_name: str, username: str
```
liranbg: pass auth_info

server/py/services/api/runtime_handlers/kubejob.py
```diff
@@ -60,16 +59,11 @@ def run(
...
-                token_name=run.spec.auth.get("token_name"),
-                username=auth_info.username,
-            )
-
         self.add_secrets_to_spec_before_running(
-            runtime, project_name=run.metadata.project
+            runtime,
+            project_name=run.metadata.project,
+            token_name=(run.spec.auth or {}).get("token_name"),
+            username=auth_info.username,
```
liranbg: simply pass the auto info - it allows the function to extend and use the class capabilities later on when we want to identify user not just by username but even perhaps its group

liranbg (changes requested):
Looks much much better. minor suggestion to keep the coding better

liranbg (approved):

PR #9054, opened 2025-12-14
[CI] Fix patch_remote script

liranbg:
### 📝 Description

`OVERWRITE_IMAGE_REGISTRY` is optional. not always needed. when not provided, use docker_registry to pull kfp image from.

---

### 🛠️ Changes Made

- use `.get` instead of direct access to avoid key error

---

### ✅ Checklist

---

### 🧪 Testing

manual patching

---

### 🔗 References
- Ticket link:
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


automation/patch_igz/patch_remote.py
```diff
@@ -616,6 +617,16 @@ def _resolve_namespace(self, namespace: str = "") -> str:
             return namespace
         return self._config.get("NAMESPACE", Constants.default_namespace)
 
+    def _resolve_docker_registry(self):
+        mlrun_docker_registry = self._config["DOCKER_REGISTRY"].rstrip("/")
```
Yacouby: why not using .get also here ?

automation/patch_igz/patch_remote.py
```diff
@@ -295,29 +289,36 @@ def _make_targets(
...
-                )
-
-                mlrun_api_container = self._deploy_patch["mlrun_api"]["spec"][
-                    "template"
-                ]["spec"]["containers"][0]
-                env_vars = mlrun_api_container.setdefault("env", [])
-                existing_var = next(
-                    (var for var in env_vars if var.get("name") == "MLRUN_KFP_IMAGE"),
-                    None,
+                docker_registry = self._resolve_docker_registry()
```
Yacouby: isn't this a duplicate call ? you already have it in `mlrun_docker_registry`

automation/patch_igz/patch_remote.py
```diff
@@ -295,29 +289,36 @@ def _make_targets(
...
+                kfp_registry = (
+                    overwrite_registry if overwrite_registry else docker_registry
```
Yacouby:
```suggestion
                kfp_registry = overwrite_registry or docker_registry
```

automation/patch_igz/patch_remote.py
```diff
@@ -295,29 +289,36 @@ def _make_targets(
...
-
-                if existing_var:
-                    existing_var["value"] = kfp_image_uri
-                else:
-                    env_vars.append(
-                        {
-                            "name": "MLRUN_KFP_IMAGE",
-                            "value": kfp_image_uri,
-                        }
+                if kfp_registry:
```
Yacouby, line 293: If both `OVERWRITE_IMAGE_REGISTRY` and `DOCKER_REGISTRY` are missing, `MLRUN_KFP_IMAGE` is not set. Is this a valid scenario, or should we fail fast or at least log a warning?

automation/patch_igz/patch_remote.py
```diff
@@ -295,29 +289,36 @@ def _make_targets(
...
 
+                    if existing_var:
+                        existing_var["value"] = kfp_image_uri
+                    else:
+                        env_vars.append(
+                            {
+                                "name": "MLRUN_KFP_IMAGE",
+                                "value": kfp_image_uri,
+                            }
+                        )
```
Yacouby: this block is difficult to read—a lot of indentation and nested logic which makes it hard to quickly see that it’s just updating an existing env var or appending a new one. my suggestion :
```suggestion
                    for var in env_vars:
                           if var.get("name") == "MLRUN_KFP_IMAGE":
                                  var["value"] = kfp_image_uri
                                  break
                    else:
                           env_vars.append({"name": "MLRUN_KFP_IMAGE", "value": kfp_image_uri})
```

Yacouby (changes requested):
Suggestions for better readability

automation/patch_igz/patch_remote.py
liranbg: you will fail much earlier

Yacouby (approved):
much better 👍

PR #9066, opened 2025-12-16
[Hub] Implement get_hub_step new SDK and new HubStep class

danielperezz:
### 📝 Description

Implement `get_hub_step` new SDK and new `HubStep` class as part of supporting this new hub asset in MLRun.

---

### 🛠️ Changes Made
- New `HubAsset` class to implement all the common hub assets functionalities
- Make `HubModule` inherit from `HubAsset`
- New `HubStep` class which also inherits from `HubAsset`
- new `get_hub_step` SDK 

---

### ✅ Checklist
  - [x] I updated the relevant Jira ticket for documentation

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11704
- Design docs links: https://iguazio.atlassian.net/wiki/spaces/Dataflows/pages/588742659/ML-10357+-+Support+Importing+Steps+From+the+Hub
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


mlrun/hub/base.py
```diff
@@ -0,0 +1,141 @@
...
+        if path:
+            path = Path(path)
+            if not path.exists():
+                raise ValueError(f"Path does not exist: {path}")
+            if not path.is_dir():
+                raise ValueError(f"Path is not a directory: {path}")
+            return path
+        return Path(os.getcwd())
+
+    def get_src_file_path(self):
```
Eyal-Danieli:
```suggestion
    def get_src_file_path(self) -> str:
```

mlrun/hub/base.py
```diff
@@ -0,0 +1,141 @@
...
+        self.categories: list = categories or []
+        self.requirements: list = requirements or []
+        self.local_path: Optional[Path] = local_path
+        self.filename: str = filename or name
+        self.example: str = example or ""
+        self.url: str = url or ""
+
+    def module(self):
+        """Import the code of the asset as a module."""
+        src_path = self.get_src_file_path()
```
Eyal-Danieli: If it’s only used in the error exception, please move it there. No need to init this variable if it’s not used

mlrun/hub/base.py
```diff
@@ -0,0 +1,141 @@
...
+        )
+        self._download_object(obj_url=source_url, target_name=self.filename)
+        if download_example and self.example:
+            example_url, _ = extend_hub_uri_if_needed(
+                uri=self.url, asset_type=self.ASSET_TYPE, file=self.example
+            )
+            self._download_object(obj_url=example_url, target_name=self.example)
+
+    def _download_object(self, obj_url, target_name, secrets=None):
+        data = get_object(url=obj_url, secrets=secrets)
```
Eyal-Danieli, line 104: maybe we should add here some error handling in case the file is not found?

mlrun/hub/module.py
```diff
@@ -12,129 +12,106 @@
...
+        self.kind = kind
+
+    # TODO: Remove this in 1.13.0
+    @deprecated(
+        version="1.11.0",
+        reason="This function is deprecated and will be removed in 1.13. You can download module files by calling "
+        "download_files() instead.",
+        category=FutureWarning,
+    )
+    def download_module_files(
```
Eyal-Danieli, line 68: Instead of duplicating the download logic, a better practice would be calling `download_files` from `download_module_fules()`.

mlrun/hub/module.py
```diff
@@ -12,129 +12,106 @@
...
-            if not path.is_dir():
-                raise ValueError(f"Path is not a directory: {path}")
-            return path
-        return Path(os.getcwd())
+        super().download_files(
+            local_path=local_path,
+            download_example=download_example,
+        )
 
     def get_module_file_path(self):
```
Eyal-Danieli, line 114: should we keep this method on the module level? expect for the name, it's similar to the parent method. Maybe we need just use `get_src_file_path` as is and update the one place where we use it (`hub_module.get_module_file_path()` -> `hub_module.get_src_file_path()` within `mlrun.model_monitoring.api`)

Eyal-Danieli (commented):
Nice!
few comments

mlrun/hub/module.py
danielperezz: I didn't use `download_files` since it doesn't accept `secrets` while `download_module_files()` still need to accept it

mlrun/hub/base.py
danielperezz: and what would we wanna do in the case of such error?

mlrun/hub/module.py
danielperezz: Shouldn’t we keep the original `get_module_file_path()`  function to follow the deprecation protocol?

Eyal-Danieli: alright, but add a deprecation message

Eyal-Danieli (approved):

PR #9074, opened 2025-12-17
[DataStore] Fix timezone retrieving when checking partition directory [1.10.x]  

davesh0812:
### 📝 Description
1. When calling _is_directory_in_range, either the start or end must exist, so now we retrieve the timezone from whichever one is available.
2. Adding check if `start_time or end_time` where given before calling to the new optimized flow
---

### 🛠️ Changes Made

---

### ✅ Checklist

---

### 🧪 Testing
Added unit test

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11782
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [ ] No

---

### 🔍️ Additional Notes

[ML-11782]: https://iguazio.atlassian.net/browse/ML-11782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/datastore/base.py
+++ b/mlrun/datastore/base.py
@@ -160,8 +160,8 @@ class DataStore(BaseRemoteClient):
 
     @staticmethod
     def _is_directory_in_range(
-        start_time: datetime.datetime,
-        end_time: datetime.datetime,
+        start_time: Union[datetime.datetime, None],
+        end_time: Union[datetime.datetime, None],
         year: int,
         month: Optional[int] = None,
         day: Optional[int] = None,
@@ -198,8 +198,8 @@ class DataStore(BaseRemoteClient):
     @staticmethod
     def _list_partition_paths_helper(
         paths: list[str],
-        start_time,
-        end_time,
+        start_time: Union[datetime.datetime, None],
+        end_time: Union[datetime.datetime, None],
         current_path: str,
         partition_level: str,
         filesystem,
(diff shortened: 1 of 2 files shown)


mlrun/datastore/base.py
```diff
@@ -176,7 +176,7 @@ def _is_directory_in_range(
             month=month or 1,
             day=day or 1,
             hour=hour or 0,
-            tzinfo=start_time.tzinfo,
+            tzinfo=start_time.tzinfo if start_time else end_time.tzinfo,
```
gtopper, line 179: According to the type annotation, `start_time` isn't supposed to be optional.

davesh0812: It's not an optional, we have to pass it but it can be None as well.
Same as all the flow.. so I changing to `Union[datetime.datetime, None]`,

mlrun/datastore/base.py
```diff
@@ -160,8 +160,8 @@ def get_spark_options(self, path=None):
 
     @staticmethod
     def _is_directory_in_range(
-        start_time: datetime.datetime,
-        end_time: datetime.datetime,
+        start_time: Union[datetime.datetime, None],
```
gtopper: Not sure why it doesn't let me comment on both lines at the same time, but the conventional way of doing this is
```suggestion
        start_time: Optional[datetime.datetime],
```

mlrun/datastore/base.py
gtopper: Can be None is the definition of optional in python, though :)

gtopper (approved):
LGTM

PR #9082, opened 2025-12-18
[CI] Add concurrency settings to system tests workflow and add release branches trigger

pioardi:
### 📝 Description

---

### 🛠️ Changes Made

---

### ✅ Checklist

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link:
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [ ] No

---

### DIFF
--- a/.github/workflows/system-tests-opensource.yml
+++ b/.github/workflows/system-tests-opensource.yml
@@ -18,10 +18,16 @@ permissions:
   contents: read
   packages: write
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+  
 on:
    pull_request:
     branches:
       - development
+      - release/*
+
    workflow_dispatch:
     inputs:
       pr_number:
--- a/.github/workflows/system-tests-opensource.yml
+++ b/.github/workflows/system-tests-opensource.yml
@@ -28,7 +28,6 @@ on:
       - development
       - '[0-9]+.[0-9]+.x'
 
-
    workflow_dispatch:
     inputs:
       pr_number:
--- a/.github/workflows/system-tests-opensource.yml
+++ b/.github/workflows/system-tests-opensource.yml
@@ -26,7 +26,7 @@ on:
    pull_request:
     branches:
       - development
-      - release/*
+      - '[0-9]+.[0-9]+.x'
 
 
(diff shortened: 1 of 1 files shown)


pioardi:
@assaf758 concurrency test https://github.com/mlrun/mlrun/actions/runs/20331146171 , this was canceled because I pushed another commit ( really good hint thanks mate ) 

I added a trigger for release branches ( PR )

.github/workflows/system-tests-opensource.yml
```diff
@@ -18,10 +18,17 @@ permissions:
...
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+  
 on:
    pull_request:
     branches:
       - development
+      - release/*
```
assaf758: our release branches are 1.10.x, 1.9.x etc.

.github/workflows/system-tests-opensource.yml
```diff
@@ -18,10 +18,17 @@ permissions:
   contents: read
   packages: write
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+  
 on:
    pull_request:
```
assaf758, line 26: As i understand, we need this to be on push. 
Otherwise, I don't understand how we expect dev to iterate and fix the failure since the test will run only on the first push (correct me if i'm wrong?)

pioardi: They need to open a PR ( even a draft ) and this will run. We don' t want to run this on every push in early stages to avoid high load on runners. When they need, they open a PR, if they don' t need  to run it they can push without loading the system

.github/workflows/system-tests-opensource.yml
pioardi: I will use - '[0-9]+.[0-9]+.x' as suggested by Liran

.github/workflows/system-tests-opensource.yml
pioardi: Me and Assaf are in synch now

liranbg (approved):

assaf758 (approved):

PR #9087, opened 2025-12-18
[CI] private release setup and change in the delivery process to first del…

pioardi:
This PR implements the new release process that allow us to first test releases in a private registry ( for both container images and python packages ). For that goal we have implemented two new workflows : 

1. Private release, which is building as usual but pushing only on our private artifactory. In order to do that, we need to build and push on a self-hosted runner so we made the runner group configurable in the build-internal workflow while keeping the previous behavior as default
2. We implemented a promotion pipeline that without rebuilding, take the artifacts ( container images and py packages ) and push them on the public channel
3. Last but not least, we added an approval gate so that those workflows will need an approval by one of the mlrun leaders to run ( I have set for now Liran,Assaf,Raya,me and Or, but we can change as needed and we can add up to 6 approvers ) 

### 📝 Description

---

### 🛠️ Changes Made

---

### ✅ Checklist

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link:
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [ ] No

---


.github/workflows/release-promotion.yaml
```diff
@@ -0,0 +1,348 @@
...
+        with:
+          registry: registry.hub.docker.com
+          username: ${{ secrets.DOCKER_HUB_DOCKER_REGISTRY_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_DOCKER_REGISTRY_PASSWORD }}
+
+      - name: Docker login (Source Artifactory)
+        if: steps.check-skip.outputs.skip == 'false'
+        uses: docker/login-action@5e57cd118135c172c3672efd75eb46360885c0ef # 3.6.0
+        with:
+          registry: artifactory.iguazeng.com:6555
```
liranbg: better of get it from a secret as well

.github/workflows/release-promotion.yaml
```diff
@@ -0,0 +1,348 @@
...
+          PREVIOUS_VERSION_INPUT: ${{ github.event.inputs.previous_version }}
+
+  promote-docker-images:
+    name: Promote Docker images from source to target Artifactory
+    runs-on: self-hosted
+    needs: prepare-inputs
+    strategy:
+      fail-fast: false
+      matrix:
+        image: [mlrun, mlrun-gpu, mlrun-kfp, mlrun-api, jupyter, log-collector, mlrun-ui]
```
liranbg: can't that be resolved from image_matrix? keep in mind we have versions where we support multi python versions, aka
`mlrun/mlrun:release` and `mlrun/mlrun:release-py39`

.github/workflows/release-promotion.yaml
```diff
@@ -0,0 +1,348 @@
...
+          rm -rf docs/welcome.ipynb docs/README.md
+      - name: Add tutorials tar to release
+        uses: ncipollo/release-action@b7eabc95ff50cbeeedec83973935c8f306dfcd0b # 1.20.0
+        with:
+          allowUpdates: true
+          tag: v${{ needs.prepare-inputs.outputs.version }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+          artifacts: mlrun-tutorials.tar
+
+  update-demos:
```
liranbg, line 330: duplicaetes with release.yaml - isnt it?

liranbg (changes requested):
nice work on refactoring the release process, left some comments to proceed with

.github/workflows/release-promotion.yaml
pioardi: release.yaml is not part of the new release process, it is kept only for backup reason and I would delete it after a bit ( 1/2 weeks )

.github/workflows/release-promotion.yaml
pioardi: Good catch solved

.github/workflows/release-promotion.yaml
pioardi: Good catch will use the matrix action as it is better

.github/workflows/build-internal.yaml
```diff
@@ -122,17 +150,20 @@ jobs:
...
         echo "mlrun_docker_repo=$( \
-          input_docker_repo=$INPUT_DOCKER_VERSION && \
+          input_docker_repo=$INPUT_DOCKER_REPO && \
           default_docker_repo=$(echo ${{ github.repository_owner }} | tr '[:upper:]' '[:lower:]') && \
           echo ${input_docker_repo:-`echo $default_docker_repo`})" >> $GITHUB_OUTPUT
+
         echo "mlrun_docker_registries=$( \
-          input_docker_registries=$INPUT_DOCKER_REGISTRIES && \
-          echo ${input_docker_registries:-ghcr.io/})" >> $GITHUB_OUTPUT
+          if [ \"${{ inputs.is_private_release }}\" = \"true\" ]; then \
```
assaf758, line 158: I suggest two things:
1. In case of is_private_release we are silently ignoring `inputs.docker_registries`. Therefore, please fail the job early (or prevent the job submit) if `is_private_release` and `inputs.docker_registries` is not empty (i.e. not default)
2. You might consider introducing a new var `docker_registires` and set it early-on: if `is_private_release` set it to `secrets.ARTIFACTORY_URL`, otherwise to `inputs.docker_registries` with default `ghchr.io/`. This way you can safely use docker_registries in the rest of the job, without checking flags.

.github/workflows/build-internal.yaml
```diff
@@ -57,6 +57,18 @@ on:
...
         default: ''
         type: string
+      runner:
+        description: 'The GitHub runner to use for building images (e.g., ubuntu-latest, self-hosted)'
+        default: 'ubuntu-latest'
+        type: string
+      is_private_release:
+        description: 'Whether this is a private release build'
+        default: 'false'
+        type: string
```
assaf758: why not boolean?

.github/workflows/build-internal.yaml
```diff
@@ -57,6 +57,18 @@ on:
...
+      build_to_cache:
```
assaf758, line 68: Suggest to rename to `is_build_to_cache` and also here use type boolean

.github/workflows/build-internal.yaml
```diff
@@ -159,6 +190,14 @@ jobs:
         username: ${{ secrets.DOCKER_HUB_DOCKER_REGISTRY_USERNAME }}
         password: ${{ secrets.DOCKER_HUB_DOCKER_REGISTRY_PASSWORD }}
 
+    - name: Docker login (Artifactory)
+      if: contains(inputs.docker_registries, 'artifactory.iguazeng.com')
```
assaf758: don't we want to login if we are actually using `secrets.ARTIFACTORY_URL`?
my suggestion above (to introduce new var `docker_registires`) might help here

.github/workflows/release.yaml
```diff
@@ -60,9 +59,17 @@ on:
         default: ''
 
 jobs:
+  approval-gate:
```
assaf758, line 62: 1. Also here - please add the approval gate to release only (and not to bump-rc).
2. Please remove 'private' from name and echo down below

.github/workflows/private-release.yaml
```diff
@@ -0,0 +1,177 @@
...
+        required: false
+        default: 'private-mlrun-py'
+        type: string
+      ui_ref:
+        description: 'The UI reference (branch / tag name) to use for the UI image, (development, 1.3.3, etc)'
+        required: false
+        default: ''
+
+jobs:
+  approval-gate:
```
assaf758, line 51: In my opinion, we don't need approval for rc. Please add the approval gate to release only (and not to bump-rc).

assaf758 (commented):
Looks good overall :-) 
Some comments inline.

.github/workflows/build-internal.yaml
pioardi: Was using the same convention as the other pipeline types ( see build_from_cache ), but if it works I am glad to change it doing a quick test now

.github/workflows/build-internal.yaml
pioardi: I would like to keep the name consistent with build_from_cache that we have already, since the PR is already big I would not to refactor names at this stage

.github/workflows/release.yaml
pioardi: We should instead use the private name as this is a private release. This is not intended to be used for rc but mainly for stable releases - Although we are supporting rc just in case and IMHO there is no need to remove it from the approval gate and make complexity. I want this pipeline to be run with an approval.

.github/workflows/build-internal.yaml
pioardi: We are doing the login only if private release is set to true now, good catch

.github/workflows/private-release.yaml
pioardi: This is not intended to be used for rc, but we are keeping it for future usage eventually.
Let just keep it as it is and if we run into issues in future we can adjust, I want this pipeline to only run after an approval for now

.github/workflows/build-internal.yaml
pioardi: solved

.github/workflows/build-internal.yaml
pioardi: The build internal pipeline is used in different places, the less we touch it the better ( as you know I want to simplify this down the road ).

If private release is specified it only uses artifactory and skip other, what I did is to add this to the is_private_release input parameter

liranbg (approved):

PR #9102, opened 2025-12-22
[DEPRECATIONS] Support mlflow version 3.x

omermaim:
…n new version of mlflow

### 📝 Description
upgraded mlflow version to 3.0.0

---

### 🛠️ Changes Made
changed logic in model retrievel due to new mlflow objects. the models are no longer in list_artifacts function

---

### ✅ Checklist

---

### 🧪 Testing
ran alll tracker unit tests that is the only package that has this library.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11537?actionerId=712020%3A4a32fc43-a2c3-4d13-b5d2-7e4c9f30172e&sourceType=assign
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [x] Yes (explain below)
- [ ] No

---

### 🔍️ Additional Notes

removes mlflow from the mlrun/mlrun image.


dependencies.py
```diff
@@ -61,7 +61,7 @@ def extra_requirements() -> dict[str, list[str]]:
             "avro~=1.11",
         ],
         "redis": ["redis~=4.3"],
-        "mlflow": ["mlflow~=2.22"],
+        "mlflow": ["mlflow~=3.0.0"],
```
theSaarco: Having discussed this in the release-tracking meeting, we want to remove the `mlflow` requirement from the `complete` extra. This means that this line should be removed from here, and added below (see L84 and further) so that when users ask for `mlrun["mlflow"]` they get just mlflow, and to install complete + mlflow they'll need to do install `mlrun["complete, mlflow"]`.

dependencies.py
theSaarco, line 98: Should add `mlflow` to this location.

extras-requirements.txt
```diff
@@ -38,7 +38,7 @@ kafka-python~=2.1.0
 avro~=1.11
 redis~=4.3
 graphviz~=0.20.0
-mlflow~=2.22
+mlflow~=3.0.0
```
theSaarco: As mentioned above, this also needs to be removed so we don't install this package by default in our images.

theSaarco (changes requested):
Our general approach should be to *not install* the `mlflow` package unless specifically asked for. This means that we want to remove this package from the `complete` extra, and install it only if the user asks for the `mlflow` extra. This will reduce our exposure to the package and its possible vulns and dependencies.
I've added some comments explaining the exact changes that are needed to facilitate this.
@liranbg, @guy1992l  - FYI.

jond01:
> ### 🚨 Breaking Changes?
> * [ ]  Yes (explain below)
> * [ x] No

@omermaim - please mark this as a breaking change :) Also in Jira

dependencies.py
```diff
@@ -85,6 +84,7 @@ def extra_requirements() -> dict[str, list[str]]:
         {
             "dev-postgres": ["pytest-mock-resources[postgres]~=2.12"],
             "kfp18": ["mlrun_pipelines_kfp_v1_8[kfp]~=0.6.0"],
+            "mlflow": ["mlflow~=3.0.0"],
```
gilad-shaham: If we went all the way up to 3, why ~=3.0.0 and not 3.8.0? Does something break for us between 3.0 and 3.8?

theSaarco (approved):
LGTM

dependencies.py
guy1992l: in the ticket it said to move to version 3.0.0. Anyway this change should support later versions as well.

guy1992l: @gilad-shaham @theSaarco 

`mlflow~=3.0.0` is very strict. Question is should we strict it as security violations do not exist for it or should we move to `mlflow~=3.0`. The changes in their SDK and API should be fine.

dockerfiles/test/requirements.txt
```diff
@@ -1 +1,2 @@
 mlrun-pipelines-kfp-v1-8[kfp]~=0.6.0
+mlflow~=3.0.0
```
guy1992l: Same here...

tests/test_requirements.py
```diff
@@ -138,6 +140,7 @@ def test_requirement_specifiers_convention():
         "aioresponses": {"~=0.7"},
         "testcontainers[k3s]": {"~=4.10.0"},
         "scikit-learn": {"~=1.5.2"},
+        "mlflow": {"~=3.0.0"},
```
guy1992l: same here

mlrun/track/trackers/mlflow_tracker.py
```diff
@@ -440,19 +439,17 @@ def _log_model(
...
-        archive_path = pathlib.Path(tmp_path) / f"{model_uri.stem}.zip"
-        if not os.path.exists(model_uri):
-            local_path = mlflow.artifacts.download_artifacts(
-                artifact_uri=str(model_uri)
-            )
-            model_uri = pathlib.Path(local_path)
+        local_model_path = mlflow.artifacts.download_artifacts(
+            artifact_uri=str(model_uri)
+        )
+        model_path = pathlib.Path(local_model_path)
```
guy1992l, line 446: Please keep the headline comments.

```suggestion
        # Download the model and set the path to local path:
        local_model_path = mlflow.artifacts.download_artifacts(
            artifact_uri=str(model_uri)
        )
        model_path = pathlib.Path(local_model_path)
```

tests/track/platform_trackers/test_mlflow_tracker.py
```diff
@@ -476,7 +476,12 @@ def test_import_model(rundb_mock, handler):
 
         # Access model's uri through mlflow's last run
         mlflow_run = mlflow.last_active_run()
-        model_uri = f"{mlflow_run.info.artifact_uri}/model"
+
+        logged_models = mlflow.search_logged_models(
+            filter_string=f"source_run_id = '{mlflow_run.info.run_id}'"
+        )
+
+        model_uri = logged_models.iloc[0].artifact_location
```
guy1992l: add `output_format="list"` to `search_logged_models` to get a list instead of DataFrame.

guy1992l (changes requested):
VERY GOOD JOB :)

guy1992l (approved):

PR #9104, opened 2025-12-22
[Serving] Avoid creating an output artifact in serving job without `respond()` [1.10.x]

gtopper:
[ML-11400](https://iguazio.atlassian.net/browse/ML-11400)

Also:
* Avoid creating per-model artifacts when the output schema doesn't fit
* Avoid failing the job if per-model artifact creation fails
* Expand tests to cover output artifacts
* Fix minor static code analysis warning

[ML-11400]: https://iguazio.atlassian.net/browse/ML-11400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/serving/server.py
+++ b/mlrun/serving/server.py
@@ -649,7 +649,7 @@ async def async_execute_graph(
 
     if df.empty:
         context.logger.warn("Job terminated due to empty inputs (0 rows)")
-        return []
+        return
 
     track_models = spec.get("track_models")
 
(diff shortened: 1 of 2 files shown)


mlrun/serving/server.py
```diff
@@ -649,7 +649,7 @@ async def async_execute_graph(
 
     if df.empty:
         context.logger.warn("Job terminated due to empty inputs (0 rows)")
```
danielperezz, line 651: Perhaps we should consider adding after this line:
context.log_result("num_rows", run_call_count)  # run_call_count=0

danielperezz (approved):
Looks great! had just one comment, not directly related to your changes

mlrun/serving/server.py
gtopper: But the input is empty, and `run_call_count` will necessarily always be zero? It doesn't get incremented until later in the function.

PR #9114, opened 2025-12-24
[Serving] Add ModelRunnerSelector to route models and outlets

davesh0812:
### 📝 Description
Add user ability to select the outlets after the model execution - relevant to agents (tool selection).

---

### 🛠️ Changes Made
1. `ModelRunnerSelector` - new user interface that have 2 method one for models selection and one for outlets selection.
2. `ModelRunnerErrorRaiser` is a separate branch that comes after the MRS step and not pull in to the graph.
3. Deprecated `ModelSelector` class and `model_selector` and `model_selector_parameters` params of `MRS.__init__`

---

### ✅ Checklist

---

### 🧪 Testing
Add unit tests include mm tests

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11804
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


mlrun/serving/states.py
```diff
@@ -1672,7 +1731,30 @@ def preprocess_event(self, event):
...
+                user_outlets if isinstance(user_outlets, list) else [user_outlets]
+            ) + sys_outlets
+        return None
+
+    def _is_error(self, event: dict) -> bool:
+        if len(self.runnables) == 1:
+            if isinstance(event, dict):
+                return event.get("error") is not None
+        else:
+            for model in event:
```
alxtkr77, line 1766: Bug: The `return` statement is inside the loop, so it returns on the first model checked regardless of result.

Example: If first model has no error but second model has error → returns `False` incorrectly.

Suggested fix:
```python
else:
    for model in event:
        body_by_model = event.get(model)
        if isinstance(body_by_model, dict) and "error" in body_by_model:
            return True
    return False
```

mlrun/serving/states.py
```diff
@@ -1619,6 +1621,57 @@ def _get_invocation_artifact(
         return llm_prompt_artifact
 
 
+class ModelRunnerSelector(ModelObj):
```
alxtkr77, line 1624: This class is exported in `__init__.py` as a public API for users to subclass, but the docstring is minimal.

Consider adding:
- What `None` return value means for each method
- Example usage with agent/tool workflows
- How outlets map to graph step names
- Integration with cyclic graphs

Example:
```python
class ModelRunnerSelector(ModelObj):
    """
    Strategy for controlling model selection and output routing in ModelRunnerStep.
    
    Subclass this to implement custom logic for agent workflows:
    - `select_models()`: Called BEFORE execution to choose which models run
    - `select_outlets()`: Called AFTER execution to route output to downstream steps
    
    Return `None` from either method to use default behavior (all models / all outlets).
    
    Example::
    
        class ToolSelector(ModelRunnerSelector):
            def select_outlets(self, event):
                tool = event.get("tool_call")
                return [tool] if tool else ["final"]
    """
```

mlrun/serving/states.py
```diff
@@ -2238,9 +2384,37 @@ def init_object(self, context, namespace, mode="sync", reset=False, **extra_kwar
             max_processes=self.max_processes,
             max_threads=self.max_threads,
             pool_factor=self.pool_factor,
+            raise_exception=self.raise_exception,
             **extra_kwargs,
         )
 
+    def convert_model_selector_to_model_runner_selctor(
+        self,
+        model_selector,
```
alxtkr77, line 2407: Typo: `selctor` → `selector`

```python
def convert_model_selector_to_model_runner_selector(  # Fixed
```

tests/serving/test_tracking.py
```diff
@@ -41,6 +41,8 @@
 from mlrun.serving.system_steps import MonitoringPreProcessor
 from tests.serving.test_serving import _log_model
 
+from .demo_states import *  # noqa
```
alxtkr77: Avoid wildcard imports. Prefer explicit imports:

```python
from .demo_states import LLModelWithTools, MySelector, Tool
```

tests/serving/test_async_flow.py
```diff
@@ -680,10 +707,16 @@ def test_model_runner_with_selector(execution_mechanism: str):
...
     graph = function.set_topology("flow", engine="async")
-    model_runner_step = ModelRunnerStep(
-        name="my_model_runner",
-        model_selector=MyModelSelector(models=["m1", "m2"]),
-    )
+    if selector == "new":
+        model_runner_step = ModelRunnerStep(
+            name="my_model_runner",
+            model_runner_selector=MyModelRunnerSelector(models=["m1", "m2"]),
+        )
```
alxtkr77, line 714: Consider adding a test that verifies the `FutureWarning` is emitted when using the deprecated `model_selector` parameter:

```python
with pytest.warns(FutureWarning, match="model_selector.*deprecated"):
    ModelRunnerStep(
        name="my_model_runner",
        model_selector=MyModelSelector(models=["m1", "m2"]),
    )
```

mlrun/serving/states.py
```diff
@@ -1619,6 +1621,57 @@ def _get_invocation_artifact(
...
+        cls._dict_fields = list(
+            set(cls._dict_fields)
+            | set(inspect.signature(cls.__init__).parameters.keys())
+        )
+        cls._dict_fields.remove("self")
+
+    def select_models(
+        self,
+        event: Any,
+        available_models: list[Model],
```
alxtkr77, line 1656: Minor: Return type is `Optional[Union[list[str], list[Model]]]` here, but the Adapter class (line 2405) uses `Union[Sequence[str], Sequence[Model]]`.

Consider using consistent types - either both `list` or both `Sequence`.

alxtkr77:
## Test Coverage Gaps

The happy path is well covered, but error handling and edge cases are missing:

| Missing Test | Risk |
|--------------|------|
| `_is_error()` with multiple models | Would catch the early-return bug |
| Error routing to `{name}_error_raise` outlet | Error flow untested |
| `select_models()` with new `ModelRunnerSelector` | Only `select_outlets()` is tested via `MySelector` |
| `select_outlets()` returning `None` | Default routing behavior untested |
| `model_runner_selector_parameters` dict | Only string class name is tested |
| Deprecation warning assertion | Warning emitted but not verified with `pytest.warns()` |

Suggested additions:

```python
def test_model_runner_error_routing():
    """Test that errors route to error_raise outlet"""
    ...

def test_model_runner_selector_select_models():
    """Test select_models() method of new ModelRunnerSelector"""
    class MyModelsSelector(ModelRunnerSelector):
        def select_models(self, event, available_models):
            return [m for m in available_models if m.name in event.get("models", [])]
    ...

def test_is_error_multiple_models():
    """Test _is_error detects error in any model, not just first"""
    ...
```

mlrun/serving/states.py
```diff
@@ -2238,9 +2384,37 @@ def init_object(self, context, namespace, mode="sync", reset=False, **extra_kwar
...
 
+    def convert_model_selector_to_model_runner_selctor(
+        self,
+        model_selector,
+    ) -> "ModelRunnerSelector":
+        """
+        Wrap a ModelSelector into a ModelRunnerSelector for backward compatibility.
+        """
+
+        class Adapter(ModelRunnerSelector):
```
royischoss, line 2413: nice, better use a private _convert_model_selector_to_model_runner_selctor signature

mlrun/serving/states.py
davesh0812: It is the way I need it to be :)

royischoss (commented):
LGTM small comment and what we discussed

mlrun/serving/states.py
davesh0812: All the mention use case are covered by new or old tests please check again

mlrun/serving/states.py
alxtkr77: Maybe then something like this? 
```
def _is_error(self, event: dict) -> bool:
    if len(self.runnables) == 1:
        return isinstance(event, dict) and event.get("error") is not None
    
    if not event:
        return False
    body_by_model = next(iter(event.values()))
    return isinstance(body_by_model, dict) and "error" in body_by_model
```
The 'for' loop is really confusing

royischoss: @alxtkr77 the error can come from a single model in the end of the list, we want the error raiser step will handle it and 1 is enough

tests/system/runtimes/test_nuclio.py
```diff
@@ -596,6 +597,49 @@ def test_serving_with_cyclic_graph(self):
         ):
             function.invoke(path="/", body={"counter": -5})
 
+    @pytest.mark.parametrize("with_object", [True, False])
```
royischoss, line 598: 🔥

mlrun/serving/states.py
alxtkr77: But you never get to the end of the list - you unconditionally return on the first iteration

royischoss (approved):
LGTM 👍

mlrun/serving/states.py
royischoss: "error" field is in the body only if error was raised. else the body is the user model answer, there might be an issue if a user adding "error" to his body first layer

davesh0812: @alxtkr77 sorry missed this one.. will fix in follow up pr

PR #9118, opened 2025-12-24
[Model Monitoring] Fix TSDB table creation when same credentials are set (ML-11807)

alxtkr77:
## Summary
- Fix bug where `set_credentials()` returns early without creating TSDB tables when the same credentials are already set
- This caused `UndefinedTable` errors if tables were deleted or didn't exist yet
- The fix calls `_create_tsdb_tables()` before the early return, which is safe because all TSDB connectors use idempotent table creation

### DIFF
--- a/server/py/services/api/crud/model_monitoring/deployment.py
+++ b/server/py/services/api/crud/model_monitoring/deployment.py
@@ -1829,7 +1829,7 @@ class MonitoringDeployment:
                     )
                     # Even if credentials match, ensure TSDB tables exist (ML-11807).
                     # This handles cases where tables were deleted or don't exist yet.
-                    # The create_tables() call is idempotent (uses CREATE TABLE IF NOT EXISTS).
+                    # The create_tables() call is idempotent for all TSDB connectors.
                     if tsdb_profile_name:
                         tsdb_profile = self._validate_and_get_tsdb_profile(
                             tsdb_profile_name
--- a/server/py/services/api/crud/model_monitoring/deployment.py
+++ b/server/py/services/api/crud/model_monitoring/deployment.py
@@ -1824,9 +1824,17 @@ class MonitoringDeployment:
                 self.check_if_credentials_are_set()
                 if self._is_the_same_cred(stream_profile_name, tsdb_profile_name):
                     logger.debug(
-                        "The same credentials are already set for the project - aborting with no error",
+                        "The same credentials are already set for the project - ensuring TSDB tables exist",
                         project=self.project,
                     )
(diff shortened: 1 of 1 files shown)


server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -1824,9 +1824,17 @@ def set_credentials(
                 self.check_if_credentials_are_set()
                 if self._is_the_same_cred(stream_profile_name, tsdb_profile_name):
                     logger.debug(
-                        "The same credentials are already set for the project - aborting with no error",
+                        "The same credentials are already set for the project - ensuring TSDB tables exist",
                         project=self.project,
                     )
+                    # Even if credentials match, ensure TSDB tables exist (ML-11807).
+                    # This handles cases where tables were deleted or don't exist yet.
+                    # The create_tables() call is idempotent (uses CREATE TABLE IF NOT EXISTS).
```
gtopper: This depends on the connector though. Are they all idempotent? They don't all use `CREATE TABLE IF NOT EXISTS`, that's certain.

gtopper: It seems like the answer is that they are idempotent, so just need to clear up the comment.

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -1824,9 +1824,17 @@ def set_credentials(
...
+                    if tsdb_profile_name:
+                        tsdb_profile = self._validate_and_get_tsdb_profile(
+                            tsdb_profile_name
+                        )
```
gtopper, line 1836: Do we also need to
```python
            secrets_dict[
                mlrun.common.schemas.model_monitoring.ProjectSecretKeys.TSDB_PROFILE_NAME
            ] = tsdb_profile_name
```
?

alxtkr77: No, we don't need to set `secrets_dict` here. 

This is the early return path where credentials are **already stored** in project secrets:
1. `check_if_credentials_are_set()` (line 1824) confirmed credentials exist
2. `_is_the_same_cred()` (line 1825) confirmed they match the provided values
3. We return at line 1838, never reaching `store_project_secrets()` (line 1880)

The `secrets_dict` variable is only used later to store secrets, but since the credentials are already stored correctly, we just need to ensure the TSDB tables exist.

server/py/services/api/crud/model_monitoring/deployment.py
alxtkr77: fixed

gtopper (approved):

PR #9121, opened 2025-12-27
[Runtime] Implement runtime context detection in SDK sync method

elbamit:
### 📝 Description

During `import mlrun`, authentication tokens stored locally in the igz.yml file are synced to Kubernetes and persisted as secrets. This PR prevents that synchronization from occurring when the code is executed from within a runtime.

To detect runtime execution, an environment variable `MLRUN_RUNTIME_KIND` is injected into the pods and used as the indicator. Its value is derived from the `mlrun/class` label.

---

### 🛠️ Changes Made

- Refactor the `_generate_runtime_env` to include both the logic of nuclio's env generation function (`_get_runtime_env`) and the mlrun one used for other runtimes
- Created a `_generate_external_source_runtime_envs` for non-static env vars such as `MLRUN_RUNTIME_KIND`
- Created a `_generate_k8s_runtime_env` for k8s format
- Update all callees of above functions accordingly
- Created an `is_running_in_runtime` helper function
- Updated the `sync_secret_tokens` to not run if inside a runtime

---

### ✅ Checklist

---

### 🧪 Testing
  
  
- Refactor existing tests for `kubejob`, `nuclio` and `dask`
- Refactor `_get_pod_env_vars_as_dict` used in test_submit because it assumed all external sources env vars are secret kind

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10519
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


mlrun/runtimes/base.py
```diff
@@ -444,14 +432,14 @@ def _get_db_run(
             return task.to_dict()
 
     def _generate_runtime_env(
-        self, runobj: RunObject = None, auth_info: mlrun.common.schemas.AuthInfo = None
-    ) -> dict:
+        self, runobj: RunObject = None, return_k8s_format: bool = False
```
liranbg: ideally, you want to have helper function to convern between formats - but, regardless this is a bit problematic because `_generate_runtime_env` may include `valueFrom` in which you wont be able to flatten. thus, I think we need to calibrate why we have some consumers using one way and other another way (perhaps diff between remote runtime and local runs). for that purpose, we need to move it to launchers or so.

liranbg (changes requested):
preliminary review

mlrun/runtimes/base.py
```diff
@@ -460,6 +446,11 @@ def _generate_runtime_env(
             "MLRUN_DEFAULT_PROJECT": active_project,
         }
 
+        if self.metadata.credentials.access_key:
+            runtime_env[
+                mlrun.common.runtimes.constants.FunctionEnvironmentVariables.auth_session
+            ] = self.metadata.credentials.access_key
+
```
liranbg: what's that for?

liranbg (commented):
minor question/comment - lgtm

mlrun/runtimes/base.py
elbamit: This comes from nuclio's `_get_runtime_env` function 
https://github.com/mlrun/mlrun/pull/9121/changes#diff-91002e128e55ae0b16d8d62abbd2c9f54e83d5d584a287b43f0df436df7b2f88L884-L887

mlrun/runtimes/base.py
```diff
@@ -446,7 +446,14 @@ def _generate_runtime_env(self, runobj: RunObject = None):
...
 
-        if self.metadata.credentials.access_key:
+        # Import here to avoid circular import
+        import mlrun.runtimes
+
+        # Set auth session only for nuclio runtimes that have an access key
+        if (
+            self.kind in mlrun.runtimes.RuntimeKinds.nuclio_runtimes()
+            and self.metadata.credentials.access_key
+        ):
```
liranbg, line 456: Would be fix on a following PR

liranbg (approved):

PR #9135, opened 2025-12-29
[Model Monitoring] Optimize `drift-over-time` API for V3IO TSDB

Eyal-Danieli:
Optimize the `drift-over-time` API for V3IO TSDB. The main issue was that we were iterating over the data multiple times after retrieving it, even though this can be done in a single pass. We also added some optimization in the code logic (see more details below). 

---

### 🛠️ Changes Made
- Use integers representing nanoseconds instead of `datetime` objects for bucketing. This also means working in nanoseconds rather than microseconds for better performance during this phase.
- Filter out invalid data early, before any processing. Previously, we processed everything first and filtered later. With the new design, we avoid doing unnecessary work.
- Reduce multiple separate iterations over the data into a single pass that aggregates directly while iterating.
- Perform `datetime` conversions only at the very end, after all filtering and processing is done.
- Create fewer objects, reducing memory allocations.

---

### ✅ Checklist

---

### 🧪 Testing
- `TestMonitoringAppFlow` passed (includes testing for drift over time values). 

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11820
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
+++ b/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
@@ -1659,7 +1659,6 @@ class V3IOTSDBConnector(TSDBConnector):
                 timestamp_counts[bucket_start_ns] = (count_suspected, count_detected)
 
         # Convert to final format (sorted by timestamp)
-        convert_start = mlrun.utils.datetime_now()
         values = [
             (
                 pd.Timestamp(bucket_ns, unit="ns", tz="UTC").to_pydatetime(),
--- a/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
+++ b/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import math
-from datetime import UTC, datetime, timedelta
+from datetime import datetime, timedelta
 from io import StringIO
 from typing import Literal, Optional, Union
 
@@ -1550,6 +1550,7 @@ class V3IOTSDBConnector(TSDBConnector):
     ) -> mm_schemas.ModelEndpointDriftValues:
         table = mm_schemas.V3IOTSDBTables.APP_RESULTS
         start, end, interval = self._prepare_aligned_start_end(start, end)
+
         raw_frames: list[v3io_frames.client.RawFrame] = self._get_records(
             table=table,
             start=start,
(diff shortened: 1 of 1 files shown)


mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -1561,91 +1562,115 @@ def get_drift_data(
...
+        suspected_val = mm_schemas.constants.ResultStatusApp.potential_detection.value
+        detected_val = mm_schemas.constants.ResultStatusApp.detected.value
+
+        # Single dictionary to track: bucket_start_ns -> {endpoint_id -> max_status}
+        # This allows us to calculate max per endpoint per bucket in one pass
+        bucket_endpoint_status = {}
+
+        # Single pass through all frames and data points
+        total_data_points = 0
+        filtered_data_points = 0
```
assaf758: both vars are only assigned-to but never used - please check if can be removed?

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -1561,91 +1562,115 @@ def get_drift_data(
...
+
+                # Initialize bucket if needed
+                if bucket_start_ns not in bucket_endpoint_status:
+                    bucket_endpoint_status[bucket_start_ns] = {}
+
+                # Update max status for this endpoint in this bucket
+                if endpoint_id not in bucket_endpoint_status[bucket_start_ns]:
+                    bucket_endpoint_status[bucket_start_ns][endpoint_id] = status
+                elif status > bucket_endpoint_status[bucket_start_ns][endpoint_id]:
+                    bucket_endpoint_status[bucket_start_ns][endpoint_id] = status
```
danielperezz: This can be simplified using `defaultdict`:

``` suggestion
                bucket_endpoint_status = defaultdict(dict)
                
                bucket = bucket_endpoint_status[bucket_start_ns]
                bucket[endpoint_id] = max(bucket.get(endpoint_id, status), status)
```

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
```diff
@@ -1561,91 +1562,115 @@ def get_drift_data(
...
+            for bucket_ns, (count_suspected, count_detected) in sorted(
+                timestamp_counts.items()
+            )
         ]
 
-        return result
+        return mm_schemas.ModelEndpointDriftValues(values=values)
 
     @staticmethod
     def _convert_drift_data_to_values(
```
danielperezz: I think this function can be removed now

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
danielperezz: This can be simplified using `defaultdict` and `get`:

``` suggestion
                bucket_endpoint_status = defaultdict(dict) # place above instead of the current initializtion
                
                bucket = bucket_endpoint_status[bucket_start_ns]
                bucket[endpoint_id] = max(bucket.get(endpoint_id, status), status)
```

danielperezz (commented):
Looks great :)
Had two comments

mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py
Eyal-Danieli: we do use them in https://github.com/mlrun/mlrun/blob/fbe6dd9ef3bff7941e2968cc3406bd69063a9092/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py#L1615 and https://github.com/mlrun/mlrun/blob/fbe6dd9ef3bff7941e2968cc3406bd69063a9092/mlrun/model_monitoring/db/tsdb/v3io/v3io_connector.py#L1624

assaf758: These are also just assignments - it's not a case where these vars are **used**.

danielperezz (approved):

assaf758 (approved):

PR #9147, opened 2026-01-01
[IG4] Autoset runtime igz4 oauth configuration

liranbg:
### 📝 Description

Remove runtime-level auth env var injection (`enrich_auth_env`) and centralize auth configuration through the client-spec mechanism upon import mlrun. Auth token file path is *currently* set directly during httpdb config sync when running inside k8s cluster.

---

### 🛠️ Changes Made

- Removed `enrich_auth_env` function from `mlrun/auth/utils.py`
- Set `token_file` path directly in `httpdb.py` during config sync when running inside k8s
- Removed `enrich_auth_env` calls from `base.py` and `nuclio/function.py`
- Removed related test that validated the old env injection behavior
- Fixed patch_remote.py to handle tags with `+` in its mlrun version.

---

### ✅ Checklist

---

### 🧪 Testing

- Unit tests: `test_client_spec.py`, `test_kubejob.py`, `test_httpdb.py` OAuth tests
- Manual testing required on IG4 environment

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11898
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes

Part of the IG4 AuthN initiative. OAuth token config (enabled flag, endpoint) continues to be set via client-spec sync. Token file path is now set directly when running inside k8s cluster instead of being injected per-runtime.

### DIFF
--- a/mlrun/auth/utils.py
+++ b/mlrun/auth/utils.py
@@ -17,7 +17,6 @@ import typing
 
 import yaml
 
-import mlrun.common.constants
 import mlrun.common.schemas
 import mlrun.utils.helpers
 from mlrun.config import config as mlconf
@@ -313,30 +312,3 @@ def translate_secret_tokens(
                 raise_on_error,
             )
     return tokens
-
-
-def enrich_auth_env(
-    env: dict,
-):
-    """
-    Enrich the given environment dictionary with authentication information.
-
-    This function adds authentication-related environment variables to the provided
-    environment dictionary based on the given AuthInfo object.
-
-    :param env: The environment dictionary to enrich.
-    """
-
-    if mlrun.mlconf.is_iguazio_v4_mode():
-        env["MLRUN_AUTH_WITH_OAUTH_TOKEN__ENABLED"] = "true"
-        env["MLRUN_AUTH_TOKEN_ENDPOINT"] = os.path.join(
-            mlrun.mlconf.iguazio_api_url,
-            mlrun.mlconf.httpdb.authentication.iguazio.authentication_endpoint,
-        )
-        env["MLRUN_HTTPDB__HTTP__VERIFY"] = str(
-            mlrun.mlconf.iguazio_api_ssl_verify
-        ).lower()
-        env["MLRUN_AUTH_WITH_OAUTH_TOKEN__TOKEN_FILE"] = os.path.join(
-            mlrun.common.constants.MLRUN_JOB_AUTH_SECRET_PATH,
-            mlrun.common.constants.MLRUN_JOB_AUTH_SECRET_FILE,
-        )
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import enum
 import http
+import os
 import re
 import time
 import traceback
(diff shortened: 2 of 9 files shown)


mlrun/db/httpdb.py
```diff
@@ -658,6 +658,14 @@ def connect(self, secrets=None) -> typing.Self:
 
                 config.auth_with_oauth_token.enabled = True
 
+                # TODO: change to os.getenv("MLRUN_RUNTIME_KIND") when https://github.com/mlrun/mlrun/pull/9121
+                # is merged. The reason we can't do it for all k8s pods is dev-envs like jupyter.
+                if mlrun.k8s_utils.is_running_inside_kubernetes_cluster():
+                    config.auth_with_oauth_token.token_file = os.path.join(
+                        mlrun.common.constants.MLRUN_JOB_AUTH_SECRET_PATH,
+                        mlrun.common.constants.MLRUN_JOB_AUTH_SECRET_FILE,
```
weilerN, line 666: We could combine these two constants into a single one that represents the full path, since there’s no clear need to call `os.path.join` on every invocation

tests/rundb/test_httpdb.py
```diff
@@ -544,6 +544,18 @@ def test_iguazio_v4_oauth_config_uses_internal_endpoint_in_cluster(
...
+        "read_secret_tokens_file",
+        lambda raise_on_error: {
+            "secretTokens": [
+                {
+                    "name": "default",
+                    "token": "offline",
+                }
+            ]
+        },
+    )
```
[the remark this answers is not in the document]
liranbg: done, below lines 574-580

weilerN (approved):
1 suggestion inline, beside that LGTM

mlrun/auth/utils.py
```diff
@@ -313,30 +312,3 @@ def translate_secret_tokens(
...
-
-    if mlrun.mlconf.is_iguazio_v4_mode():
-        env["MLRUN_AUTH_WITH_OAUTH_TOKEN__ENABLED"] = "true"
-        env["MLRUN_AUTH_TOKEN_ENDPOINT"] = os.path.join(
-            mlrun.mlconf.iguazio_api_url,
-            mlrun.mlconf.httpdb.authentication.iguazio.authentication_endpoint,
-        )
-        env["MLRUN_HTTPDB__HTTP__VERIFY"] = str(
-            mlrun.mlconf.iguazio_api_ssl_verify
-        ).lower()
```
TomerShor, line 338: How / where are these env vars being set then?
Or are they not needed anymore? unclear...

liranbg: This is the intention of this pr.

no need to inject all these envvars because:
- upon import mlrun, you do client-spec request which tells you "you are under iguazio 4" so all those enrichment? needless
- this pr - handle the runtimes - so basically, all it needed is to get the mount point (which it has now)

mlrun/db/httpdb.py
liranbg: these consts are being used together and separately. joining them is upon demand is noop (painless)

TomerShor (approved):

PR #9150, opened 2026-01-04
[SDK] Fix setting token file name

liranbg:
### 📝 Description

To ensure token file names can be explicitly set and (correctly) auto set for runtimes / local development it was required to remove the ~/.igz.yml from config.py and set it according to running env (k8s, jupyter, local).

This fixes Iguazio v4 OAuth token file auto-initialization to correctly distinguish between Kubernetes runtime environments and Jupyter environments. Previously, the token file was always overwritten to the k8s secret path when running inside Kubernetes, which broke authentication for Jupyter environments.

---

### 🛠️ Changes Made

- Modified the token file initialization logic to check for Jupyter environment (`JPY_SESSION_NAME`) before overriding to k8s secret path
- Changed default `token_file` config from `~/.igz.yml` to empty string, allowing dynamic initialization based on runtime context
- Updated docstring in `sync_secret_tokens` to reflect the new behavior

---

### ✅ Checklist

---

### 🧪 Testing

- UT
- igz4 envs

---

### 🔗 References
- Ticket link:
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/mlrun/db/httpdb.py
+++ b/mlrun/db/httpdb.py
@@ -648,6 +648,23 @@ class HTTPRunDB(RunDBInterface):
                 config.httpdb.authentication.mode
                 == mlrun.common.types.AuthenticationMode.IGUAZIO_V4.value
             ):
+                if not config.auth_with_oauth_token.token_file:
+                    user_token_file = os.path.expanduser("~/.igz.yml")
+
+                    # runtimes
+                    # TODO: change to os.getenv("MLRUN_RUNTIME_KIND")
+                    # when https://github.com/mlrun/mlrun/pull/9121 is done.
+                    if (
+                        mlrun.k8s_utils.is_running_inside_kubernetes_cluster()
+                        and not os.environ.get("JPY_SESSION_NAME")
+                    ):
+                        user_token_file = os.path.join(
+                            mlrun.common.constants.MLRUN_JOB_AUTH_SECRET_PATH,
+                            mlrun.common.constants.MLRUN_JOB_AUTH_SECRET_FILE,
+                        )
+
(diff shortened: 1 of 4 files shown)


mlrun/db/httpdb.py
```diff
@@ -648,6 +648,23 @@ def connect(self, secrets=None) -> typing.Self:
...
+                        mlrun.k8s_utils.is_running_inside_kubernetes_cluster()
+                        and not os.environ.get("JPY_SESSION_NAME")
+                    ):
+                        user_token_file = os.path.join(
+                            mlrun.common.constants.MLRUN_JOB_AUTH_SECRET_PATH,
+                            mlrun.common.constants.MLRUN_JOB_AUTH_SECRET_FILE,
+                        )
+
+                    config.auth_with_oauth_token.token_file = user_token_file
+
```
elbamit, line 667: If a user sets `MLRUN_AUTH_WITH_OAUTH_TOKEN__TOKEN_FILE` as an env var on the runtime, then he will override the `/var/mlrun-secrets/auth/` that is supposed to be used for runtimes.
It's fine to leave it like this for now but once the `os.getenv("MLRUN_RUNTIME_KIND")` is introduced we'll need to block the user from setting this env var on a runtime

elbamit (approved):

mlrun/db/httpdb.py
liranbg: indeed can be harmful if setting this envvar mistakenly - but that is the purpose of this PR - to allow overriding

PR #9152, opened 2026-01-04
[Requirements] Bump nuclio-jupyter

royischoss:
Bump nuclio-jupyter to version 0.12.0 as part of https://iguazio.atlassian.net/browse/NUC-704
No breaking changes
Tested by CI

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -20,3 +20,4 @@ sphinx-image-inverter~=1.1
 # snowballstemmer==3.0.0, which was released with breaking changes and removed the 'porter' algorithm, leading to
 # the error 'Stemming algorithm 'porter' not found'
 snowballstemmer!=3.0.0
+-r ../extras-requirements.txt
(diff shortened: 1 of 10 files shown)


liranbg (changes requested):
did you update deps lock on all targets?

royischoss:
@liranbg Running make upgrade-mlrun-deps-lock seems to get boto3 back to the locked requirements but still fails in sphinx building maybe missing installation of extras-requirements

liranbg:
> @liranbg Running make upgrade-mlrun-deps-lock seems to get boto3 back to the locked requirements but still fails in sphinx building maybe missing installation of extras-requirements

discussed offline and agreed to add extras to docs as now nuclio jupyter comes without boto3

docs/requirements.txt
```diff
@@ -20,3 +20,4 @@ sphinx-image-inverter~=1.1
 # snowballstemmer==3.0.0, which was released with breaking changes and removed the 'porter' algorithm, leading to
 # the error 'Stemming algorithm 'porter' not found'
 snowballstemmer!=3.0.0
+-r ../extras-requirements.txt
```
rokatyy, line 23: why this is needed?

royischoss: Because docs import datastore that depends on requirements from extras-requirements.txt, specific for boto3 that was removed from nuclio-jupyter deps

liranbg: as mentioned @ https://github.com/mlrun/mlrun/pull/9152#issuecomment-3708362875

rokatyy (approved):

liranbg (approved):

PR #9117, opened 2025-12-24
[Serving] Support model monitoring with list + tests

tomerm-iguazio:
### 📝 Description
1) Support list and dictionary outputs for batch (as list) inputs.
2) Add model monitoring to the batching tests.
3) Add functionality to helper functions and their tests.

---

### 🛠️ Changes Made

---

### ✅ Checklist
- [x] I updated the documentation (if applicable)
- [x] I have tested the changes in this PR
- [x] I confirmed whether my changes are covered by system tests
  - [x] If yes, I ran all relevant system tests and ensured they passed before submitting this PR
  - [x] I updated existing system tests and/or added new ones if needed to cover my changes
- [x] If I introduced a deprecation:
  - [x] I followed the [Deprecation Guidelines](./DEPRECATION.md)
  - [x] I updated the relevant Jira ticket for documentation

---

### 🧪 Testing

Unit test for model monitoring + helpers functions.
System test for main functionality.
---

### 🔗 References
- Ticket link: [ML-11681](https://iguazio.atlassian.net/browse/ML-11681)
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes

[ML-11681]: https://iguazio.atlassian.net/browse/ML-11681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9156, opened 2026-01-04
[RemoteModels] OpenAI batch Support

tomerm-iguazio:
### 📝 Description
Added support for OpenAI batch processing, including asynchronous batch execution.
Added integration and unit tests.

Depend on https://github.com/mlrun/mlrun/pull/9117
---

### 🛠️ Changes Made
Added batch invocation support (sync via `ThreadPoolExecutor`, async via `asyncio.gather`) with two-level concurrency control: (1) instance-level thread pool (_executor) for sync batches with per-batch semaphore limiting concurrent threads, and (2) class-level global async semaphore (_global_async_semaphore) shared across all instances to enforce total async request limits while per-batch semaphores ensure fair distribution-preventing API rate limit violations and resource monopolization.

---

### ✅ Checklist

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link: [ML-11682](https://iguazio.atlassian.net/browse/ML-11682)
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [ ] No

---

### 🔍️ Additional Notes

[ML-11682]: https://iguazio.atlassian.net/browse/ML-11682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -283,25 +310,261 @@ def _response_handler(
...
+
+            for future in as_completed(futures):
+                idx = futures[future]
+                if first_exception is not None:
+                    # We've already seen a failure – best-effort cancel remaining work
+                    future.cancel()
+                    continue
+                try:
+                    results[idx] = future.result()
+                except BaseException as exc:  # noqa: B902
```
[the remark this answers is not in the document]
tomerm-iguazio: @copilot irrelevant comment, I need to hold `exc` variable, not possible without mentioning the exception class.

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -330,35 +600,50 @@ def invoke(
             Additional keyword arguments passed to the OpenAI client.
 
         :return:
-            A string, dictionary, or `ChatCompletion` object, depending on `invoke_response_format`.
+            Single invocation: A string, dictionary, or `ChatCompletion` object.
+            Batch invocation: A list of responses in the same order as input messages.
+            Response format depends on `invoke_response_format`.
         """
+        # Detect if this is a batch invocation (list of lists)
+        if messages and isinstance(messages[0], list):
```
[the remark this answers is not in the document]
tomerm-iguazio: changed to raise an error if there is a mixed input.

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -68,6 +84,17 @@ def __init__(
...
+        self._max_workers_per_batch = (
+            self._get_secret_or_env("OPENAI_BATCH_MAX_WORKERS")
+            or mlrun.mlconf.model_providers.openai_batch_max_workers
+        )
+        # Async concurrency limit per batch
+        self._max_concurrent_per_batch = (
+            self._get_secret_or_env("OPENAI_BATCH_MAX_CONCURRENT")
+            or mlrun.mlconf.model_providers.openai_batch_max_concurrent
+        )
+
```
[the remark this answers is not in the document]
tomerm-iguazio: added int, type check seems to me as an overkill.

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -387,11 +679,21 @@ async def async_invoke(
...
         :return:
-            A string, dictionary, or `ChatCompletion` object, depending on `invoke_response_format`.
+            Single invocation: A string, dictionary, or `ChatCompletion` object.
+            Batch invocation: A list of responses in the same order as input messages.
+            Response format depends on `invoke_response_format`.
         """
-        response = await self.async_custom_invoke(messages=messages, **invoke_kwargs)
-        return self._response_handler(
+        # Detect if this is a batch invocation (list of lists)
+        if messages and isinstance(messages[0], list):
```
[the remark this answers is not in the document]
tomerm-iguazio: solved

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -283,25 +310,261 @@ def _response_handler(
...
+        :param invoke_response_format:
+            Specifies the format of the returned response for all invocations.
+
+        :param invoke_kwargs:
+            Additional keyword arguments passed to each invoke call.
+
+        :return:
+            List of responses in the same order as messages_list.
+            Each response format depends on `invoke_response_format`.
+        """
```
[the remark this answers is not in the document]
tomerm-iguazio: I want to return the original answer of the client, which will raise an exception if I got a empty list as an input.

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -68,6 +84,17 @@ def __init__(
...
+            or mlrun.mlconf.model_providers.openai_batch_max_workers
+        )
+        # Async concurrency limit per batch
+        self._max_concurrent_per_batch = int(
+            self._get_secret_or_env("OPENAI_BATCH_MAX_CONCURRENT")
+            or mlrun.mlconf.model_providers.openai_batch_max_concurrent
+        )
+
     @classmethod
     def _import_response_class(cls) -> None:
```
gtopper: This one isn't called anywhere.

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -27,6 +30,10 @@
     from openai._models import BaseModel  # noqa
     from openai.types.chat.chat_completion import ChatCompletion
 
+# Type aliases for response types
+InvokeResponse = Union["ChatCompletion", str, dict[str, Any]]
+BatchInvokeResponse = list[InvokeResponse]
```
royischoss: I think it better working with list[InvokeResponse] other than BatchInvokeResponse

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -283,25 +310,285 @@ def _response_handler(
...
+    async def _get_or_create_global_async_semaphore(cls) -> asyncio.Semaphore:
+        """
+        Get or create the class-level global async semaphore.
+        This semaphore is shared across ALL OpenAIProvider instances to enforce
+        a global limit on concurrent async requests.
+
+        :return: Shared global asyncio.Semaphore instance.
+        """
+        async with cls._global_async_semaphore_lock:
+            if cls._global_async_semaphore is None:
```
gtopper: Use double-checked locking to avoid this lock in all but startup time (as an optimization).

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -283,25 +310,285 @@ def _response_handler(
...
+                    global_semaphore,
+                    messages,
+                    invoke_response_format,
+                    **invoke_kwargs,
+                )
+                futures[future] = idx
+
+            first_exception: Optional[BaseException] = None
+
+            for future in as_completed(futures):
```
gtopper: If we remove `as_completed`, we won't have to track the index.

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -283,25 +310,285 @@ def _response_handler(
...
+        """
+
+        global_semaphore = self._get_or_create_global_thread_semaphore()
+        results: BatchInvokeResponse = [None] * len(messages_list)  # type: ignore
+
+        max_workers_for_batch = min(len(messages_list), self._max_workers_per_batch)
+        with ThreadPoolExecutor(max_workers=max_workers_for_batch) as executor:
+            futures: dict[Any, int] = {}
+            for idx, messages in enumerate(messages_list):
+                future = executor.submit(
```
gtopper: This code looks like an exercise we're all familiar with 😉
I think you could use `executor.map()` to improve and simplify this code quite a bit.

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -283,25 +310,285 @@ def _response_handler(
...
+        try:
+            # gather() stops on first exception - fast fail
+            return await asyncio.gather(*tasks)
+        except:
+            # Cancel all remaining tasks
+            for task in tasks:
+                task.cancel()
+
+            # Wait for all tasks to acknowledge cancellation
+            await asyncio.gather(*tasks, return_exceptions=True)
```
royischoss, line 340: if gather stops on the first exception you will never await the cancellations

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -283,25 +310,285 @@ def _response_handler(
...
+        Validate messages format and detect if this is a batch invocation.
+
+        :param messages: Either a list of message dicts (single) or list of message lists (batch)
+        :return: True if batch invocation, False if single invocation
+        :raises MLRunInvalidArgumentError: If messages format is invalid (mixed types)
+        """
+        if not messages or not isinstance(messages, list):
+            return False
+
+        has_list = any(isinstance(item, list) for item in messages)
```
royischoss, line 401: do we validate that user didn't send strings?

royischoss (commented):
Hey looks good some comments

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -283,25 +310,285 @@ def _response_handler(
...
+        :param invoke_kwargs:
+            Additional keyword arguments passed to each invoke call.
+
+        :return:
+            List of responses in the same order as messages_list.
+            Each response format depends on `invoke_response_format`.
+        """
+
+        global_semaphore = self._get_or_create_global_thread_semaphore()
+        results: BatchInvokeResponse = [None] * len(messages_list)  # type: ignore
```
royischoss: I think it better loose the # type: ignore cause it only an IDE comment

mlrun/datastore/model_provider/openai_provider.py
tomerm-iguazio: same comment as below

mlrun/datastore/model_provider/openai_provider.py
tomerm-iguazio: Using `executor.map()` here would simplify the code slightly, but it does not support fast-fail behavior. With executor.map(), all tasks are submitted immediately, and exceptions are only raised when iterating over the results, meaning other tasks in the queue will still run.

The current `submit() `+ `as_completed()` approach allows us to process tasks as they finish and cancel remaining tasks immediately if one fails, which is exactly what we want for a fast-fail mechanism.

mlrun/datastore/model_provider/openai_provider.py
tomerm-iguazio: Actually, in my code the except block does await all tasks with `gather(..., return_exceptions=True)`, so even if the first task fails, all remaining tasks - including cancelled ones - are properly awaited.

mlrun/datastore/model_provider/openai_provider.py
tomerm-iguazio: Added, tnx roy!

mlrun/datastore/model_provider/openai_provider.py
tomerm-iguazio: I think it's better to have it, so we'll know it's on purpose

mlrun/datastore/model_provider/openai_provider.py
gtopper: It doesn't create all tasks though if you give it an iterator – only N tasks at a time, where N is the number of threads in the thread pool.

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -44,6 +51,15 @@ class OpenAIProvider(ModelProvider):
     support_async = True
     response_class = None
 
+    # Class-level shared async semaphore for global concurrency control
+    _global_async_semaphore: Optional[asyncio.Semaphore] = None
+    _global_async_semaphore_lock = asyncio.Lock()
```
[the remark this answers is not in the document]
gtopper: Copilot is correct about this. I think a more obvious solution than the one it suggests is to make the lock thread local.

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -283,25 +299,297 @@ def _response_handler(
...
+                            other_future.cancel()
+
+            if first_exception is not None:
+                raise first_exception
+
+        return results
+
+    def _invoke_with_global_semaphore(
+        self,
+        global_semaphore: threading.Semaphore,
```
gtopper: I don't think we need to pass it as a parameter. It's global :)

mlrun/datastore/model_provider/openai_provider.py
gtopper, line 267: Not new, but can be cleaned up a bit
```suggestion
        model_kwargs = {}
        if accepts_param(operation, "model"):
            model_kwargs["model"] = invoke_kwargs.pop("model", self.model)
        if operation:
            if not inspect.iscoroutinefunction(operation):
                raise mlrun.errors.MLRunInvalidArgumentError(
                    "OpenAI async_custom_invoke operation must be a coroutine function"
                )
            return await operation(**invoke_kwargs, **model_kwargs)
        else:
            return await self.async_client.chat.completions.create(
                **invoke_kwargs, **model_kwargs
            )
```

tomerm-iguazio: this implementation is problematic because if operation is not callable in , accepts_param will raise an error (in sync mode)
Thats why I prefer not to touch this part now.

mlrun/datastore/datastore_profile.py
```diff
@@ -526,6 +526,8 @@ class OpenAIProfile(DatastoreProfile):
     base_url: typing.Optional[str] = None
     timeout: typing.Optional[float] = None
     max_retries: typing.Optional[int] = None
+    batch_max_workers: typing.Optional[int] = None
```
davesh0812: Is it still relevant?

mlrun/datastore/datastore_profile.py
```diff
@@ -541,6 +543,11 @@ def secrets(self) -> dict:
             res["OPENAI_TIMEOUT"] = self.timeout
         if self.max_retries:
             res["OPENAI_MAX_RETRIES"] = self.max_retries
+        #  per batch
+        if self.batch_max_workers:
+            res["OPENAI_BATCH_MAX_WORKERS"] = self.batch_max_workers
```
davesh0812: relevant?

mlrun/config.py
```diff
@@ -200,6 +200,7 @@
     "v3io_framesd": "http://framesd:8080",
     "model_providers": {
         "openai_default_model": "gpt-4o",
+        "openai_batch_max_concurrent": 10,
```
davesh0812, line 203: I think the default can be much higher lets go on 64, unless you think diffrent

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -283,25 +283,156 @@ def _response_handler(
...
+    ) -> bool:
+        """
+        Validate messages format and detect if this is a batch invocation.
+
+        :param messages: Either a list of message dicts (single) or list of message lists (batch)
+        :return: True if batch invocation, False if single invocation
+        :raises MLRunInvalidArgumentError: If messages format is invalid (mixed types or strings)
+        """
+        if not messages or not isinstance(messages, list):
+            return False
```
davesh0812: is it valid option?
we not suppose to raise MLRunInvalidArgumentError as you mention on the docs?

davesh0812 (commented):
Looks good so far, just need to go over the tests.
 One thing we should still think about is how the naive execution mechanism interacts with the asyncio event loop.

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -387,11 +551,23 @@ async def async_invoke(
             Additional keyword arguments passed to the OpenAI client.
 
         :return:
-            A string, dictionary, or `ChatCompletion` object, depending on `invoke_response_format`.
+            Single invocation: A string, dictionary, or `ChatCompletion` object.
+            Batch invocation: A list of responses in the same order as input messages.
+            Response format depends on `invoke_response_format`.
```
davesh0812, line 576: lets add also `:raise:` and explain what we discussed today

mlrun/config.py
tomerm-iguazio: ok

tests/datastore/remote_model/test_openai_unit.py
```diff
@@ -0,0 +1,296 @@
...
+            results = provider.invoke(messages=messages_list)
+            duration = time.perf_counter() - start
+
+        state = mock_async_single_invoke.state
+        assert len(results) == total_messages
+        assert state["call_count"] == total_messages
+        assert state["max_concurrent_observed"] <= per_batch_limit
+
+        expected_duration = (total_messages / per_batch_limit) * latency
+        upper_bound = expected_duration + 0.2
```
tomerm-iguazio: change to 0.1

tests/datastore/remote_model/test_openai_unit.py
tomerm-iguazio, line 1: Add test, If I inside a async context, to check that regular batch is working

tests/integration/model_providers/openai/test_openai.py
```diff
@@ -313,7 +369,7 @@ def test_model_runner_with_openai(self, execution_mechanism):
             assert len(encoding.encode(answer)) == 100
```
tomerm-iguazio: add a range here too

mlrun/datastore/model_provider/openai_provider.py
```diff
@@ -283,25 +284,158 @@ def _response_handler(
...
+                )
+
+        tasks = [
+            asyncio.create_task(_bounded_invoke(messages)) for messages in messages_list
+        ]
+
+        try:
+            # gather() stops on first exception - fast fail
+            return await asyncio.gather(*tasks)
+        except:
```
davesh0812, line 334: please add relevant list of exceptions here

davesh0812 (approved):
LGTM,

one comment for followup pr

PR #9159, opened 2026-01-05
[Serving] Enable async http mode to serving

royischoss:
### 📝 Description

Enable http async mode using serving graph in mlrun.
Bump nuclio-jupyter to 0.13.0

---

### 🛠️ Changes Made
1. Make sync mode be the default mode of http trigger as before, by using with_http without AsyncSpec or as default http trigger created by mlrun
2. Change the number of workers to 1 if user use with_http with multiple workers
3. Test serving with new handler as async def signature as added in nuclio-jupyter v0.13.0
4. Bump nuclio-jupyter to 0.13.0

---

### ✅ Checklist

---

### 🧪 Testing

Added system test `test_async_http_mode_serving_graph`

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11897
- Design docs links: https://iguazio.atlassian.net/wiki/pages/resumedraft.action?draftId=627015683
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes

Both of this changes require a nuclio version of 1.15.3 and above

[NUC-696]: https://iguazio.atlassian.net/browse/NUC-696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/runtimes/nuclio/function.py
```diff
@@ -510,6 +511,13 @@ def with_http(
                 "Adding HTTP trigger despite the default HTTP trigger creation being disabled"
             )
 
+        if async_spec and async_spec.enabled:
+            if workers is not None and workers > 1:
+                logger.warning(
+                    "When async mode is enabled, the number of workers is set to 1"
+                )
```
rokatyy: actually, having many workers in async mode is supported by nuclio. While it's unnecessary it most of the cases, users might still want to use this capability. So not sure this strict enforcement is really needed

royischoss: Yesterday in the HLD @TomerShor describe it as undefined behavior because all workers will be free in async mode

rokatyy: @royischoss and this is fine, we have a round robin mechanism there

tests/system/runtimes/assets/async_serving_func.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2023 Iguazio
```
davesh0812: :)
```suggestion
# Copyright 2025 Iguazio
```

davesh0812 (approved):
LGTM

mlrun/runtimes/nuclio/function.py
royischoss: @theSaarco Hey can you please provide your input, because it was mentioned in the requirement that we do want to make a single worker as default, but do we want to prevent the user from providing a higher number of workers

royischoss: @rokatyy I change it- if the user provided number of workers we will take his value and the default value while using async will be 1 and for sync we will keep it with 8 workers.

theSaarco: @royischoss - sounds good - that's what I was going to propose 👍. Thanks.

rokatyy (approved):
LGTM

PR #9173, opened 2026-01-07
[Runtime] Remove function level import and fix circular import

elbamit:
### 📝 Description
This is a following PR of https://github.com/mlrun/mlrun/pull/9121 that fixes a circular import in the runtime module and removes the need for function-level import.

`RuntimeKinds` was moved into a dedicated module (`mlrun/runtimes/constants.py`) so it can be imported without pulling in nuclio runtime modules. The nuclio resolver helpers (`resolve_nuclio_runtime` / `resolve_nuclio_sub_kind`) cannot live in `constants.py` because they depend on nuclio runtime classes (e.g. `RemoteRuntime`), which import `base.py` and create an import cycle.

To preserve backwards compatibility, the resolvers remain available as:
`RuntimeKinds.resolve_nuclio_runtime` and `RuntimeKinds.resolve_nuclio_sub_kind`,
implemented in `mlrun/runtimes/__init__.py` and attached as `staticmethod`s.

---

### 🛠️ Changes Made
- Moved `RuntimeKinds` to `mlrun/runtimes/constants.py`
- Removed nuclio-related imports from `constants.py` to avoid import cycles
- Implemented nuclio resolver helpers in `mlrun/runtimes/__init__.py`
- Kept the existing `RuntimeKinds.resolve_nuclio_*` API working via `staticmethod` assignment

---

### ✅ Checklist

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10519
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [ ] No

---


mlrun/runtimes/__init__.py
```diff
@@ -55,6 +56,39 @@
...
+    if kind == RuntimeKinds.serving:
+        is_nuclio = True
+        sub_kind = serving_subkind
+    elif kind == RuntimeKinds.application:
+        is_nuclio = True
+    return is_nuclio, sub_kind
+
+
+RuntimeKinds.resolve_nuclio_runtime = staticmethod(resolve_nuclio_runtime)
+RuntimeKinds.resolve_nuclio_sub_kind = staticmethod(resolve_nuclio_sub_kind)
```
liranbg: This is not a good practice. move it to constants.py . I dont see a reason why not to put these functions under runtime class. if you need, use them as helper functions but make sure callees call these functions directly and not via the class overload functions.

liranbg (changes requested):

mlrun/runtimes/__init__.py
elbamit: The nuclio resolver helpers (resolve_nuclio_runtime / resolve_nuclio_sub_kind) cannot live in constants.py because they depend on nuclio runtime classes (e.g. RemoteRuntime), which import base.py and create an import cycle.

liranbg (approved):

PR #9187, opened 2026-01-10
[Auth tokens] Support listing and revoking auth tokens with System Admin

elbamit:
### 📝 Description

Allow System Admin users to list/revoke tokens for different users. Only system admins can make operations on tokens that belong to different users.

The `httpdb.list_secret_tokens` and `httpdb.revoke_secret_token` now get a `username` parameter.  When listing tokens, "*" can be used to get tokens of all users (for system-admins)

---

### 🛠️ Changes Made

- Updated auth and secret schemas. Created a new `class RevokeSecretTokenResponse` 
- Refactored `httpdb` functions to handle the `username` parameter and the `RevokeSecretTokenResponse`. Also updated other db classes (base/nopdb/sqldb)
- Added helper functions in endpoint `_is_system_admin`, `_resolve_target_username_for_list_secret_tokens` and `_resolve_target_username_for_revoke_secret_tokens`+ updated list/revoke endpoints accordingly
- Crud update of `list`/`delete`/`get` secret tokens functions (`get` needed for deletion)
- In `k8s.py`:
  - Updated timestamps to datetimes
  - `list_user_token_secrets` now adds an empty label key `mlrun/token` to only filter auth tokens secrets (instead of listing all secrets then filtering) + this is required for admins to list tokens of multiple users.
  - Added validation for k8s labels value

---

### ✅ Checklist

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-10775
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [ ] No

---


mlrun/common/schemas/secret.py
```diff
@@ -61,8 +62,13 @@ class StoreSecretTokensResponse(BaseModel):
...
+    expiration: datetime
+    username: str
 
 
 class ListSecretTokensResponse(BaseModel):
     secret_tokens: list[SecretTokenInfo]
+
+
+class RevokeSecretTokenResponse(BaseModel):
+    revoked: bool = True  # False if token revocation fails
```
Yacouby: move comment to be above the code line

mlrun/db/httpdb.py
```diff
@@ -5390,27 +5390,86 @@ def store_secret_tokens(
...
+            params=params,
         )
 
         return mlrun.common.schemas.ListSecretTokensResponse(**response.json())
 
     @mlrun.utils.iguazio_v4_only
-    def revoke_secret_token(self, token_name: str) -> None:
+    def revoke_secret_token(
+        self, token_name: str, username: Optional[str] = None
+    ) -> None:
```
Yacouby:
```suggestion
    ) -> Optional[mlrun.common.schemas.RevokeSecretTokenResponse]:
```

server/py/services/api/api/endpoints/user_secrets.py
```diff
@@ -28,6 +30,70 @@
...
+            resource_namespace=mlrun.common.schemas.AuthorizationResourceNamespace.mgmt,
+        )
+    )
+
+
+async def _resolve_target_username(
+    auth_info: mlrun.common.schemas.AuthInfo,
+    username: Optional[str],
+    action: mlrun.common.schemas.AuthorizationAction,
+) -> Optional[str]:
```
Yacouby: This function behaves differently depending on the action: admin users without a username list all users for READ, but act on themselves for DELETE. This is not obvious and could be broken easily. 
I recommend documenting it clearly since it controls access and is security sensitive.

server/py/services/api/api/endpoints/user_secrets.py
```diff
@@ -51,32 +117,65 @@ async def store_secret_tokens(
...
         services.api.crud.Secrets().list_secret_tokens,
-        auth_info.username,
+        username=target_username,
     )
 
 
-@router.delete("/tokens/{name}", status_code=HTTPStatus.OK.value)
+@router.delete(
+    "/tokens/{name}",
+    status_code=HTTPStatus.OK.value,
```
Yacouby, line 88: this returns 200 even with `revoked=False` when the token is not found, is it intended (and documented) ?

Yacouby (changes requested):
Missing auth edge case tests: regular users with username, admins revoking others' tokens, and admins deleting without username affecting only themselves

mlrun/db/httpdb.py
elbamit: Function is not returning it, just using it to log a warning to the user.
Do you think it should be also returned?

Yacouby: Yes, it lets callers and tests know if revocation succeeded, instead of only logging it.

server/py/services/api/api/endpoints/user_secrets.py
elbamit: I didn't touch the status code, just added the response model.
How do you think this should behave?

If `revoked=False` we log a message on the client side that token revocation failed.

mlrun/db/httpdb.py
```diff
@@ -5390,27 +5390,91 @@ def store_secret_tokens(
...
+                )
+
+            # As a system admin, list tokens for a specific user
+            user_tokens = db.list_secret_tokens(username="john_doe")
+
+            # As a system admin, list tokens for all users
+            all_tokens = db.list_secret_tokens()
         """
+        if username == "":
+            raise MLRunInvalidArgumentError("Username cannot be an empty string.")
```
liranbg: can be removed

mlrun/db/httpdb.py
```diff
@@ -5390,27 +5390,91 @@ def store_secret_tokens(
...
         endpoint_path = "user-secrets/tokens"
+        params = None
+        if username is not None:
+            params = {"username": username}
```
liranbg, line 5424:
```suggestion
        params = {"username": username or None}
```

mlrun/db/httpdb.py
```diff
@@ -5390,27 +5390,91 @@ def store_secret_tokens(
...
             endpoint_path,
             "delete user secret token",
+            params=params,
         )
+        result = mlrun.common.schemas.RevokeSecretTokenResponse(**response.json())
+        user_info = f" for user '{username}'" if username else ""
+        if result.revoked:
+            logger.info(f"Token '{token_name}'{user_info} was successfully revoked.")
+        else:
+            logger.warning(f"Token '{token_name}'{user_info} could not be revoked.")
```
liranbg: could we tell _why_?

server/py/framework/utils/singletons/k8s.py
```diff
@@ -1299,26 +1307,41 @@ def list_user_token_secrets(
             if token_info:
                 secret_tokens.append(token_info)
 
+        # Sort by username (A-Z), then by token name (A-Z)
+        secret_tokens.sort(key=lambda t: (t.username, t.name))
```
liranbg: comment why?

liranbg (commented):
preliminary review

mlrun/db/httpdb.py
elbamit: Why remove the client-side check? Why not the API side check? Isn't that an unnecessary API call?

mlrun/db/httpdb.py
elbamit: This change + removing the `if username == ""` check makes param be:
`params = {"username":None}` when `username=""`. This is the behavior that we don't want, because it returns all users.

mlrun/db/httpdb.py
elbamit: In think that for security reasons we don't want to tell why. See https://iguazio.atlassian.net/browse/ML-10775?focusedCommentId=192704

server/py/framework/utils/singletons/k8s.py
elbamit: There wasn't real justification for that sorting.. It only looked better. Perhaps for UI this will be helpful..?

mlrun/db/httpdb.py
liranbg: usually, for two reasons:
1. making the client "smarter" (adding validatinos) making it hard to overcome it on future versions of the API (it is easier to upgrade api rather than clients)
2. the validations should occur on api side, for the reason the business  logic is on server side (this allows both UI and SDK and native API to enjoy it) and there's no reason to duplicate the logic

mlrun/db/httpdb.py
```diff
@@ -5390,27 +5390,91 @@ def store_secret_tokens(
...
             endpoint_path,
             "list user secret tokens",
+            params=params,
         )
 
         return mlrun.common.schemas.ListSecretTokensResponse(**response.json())
 
     @mlrun.utils.iguazio_v4_only
-    def revoke_secret_token(self, token_name: str) -> None:
+    def revoke_secret_token(
```
rokatyy: should this endpoint be added to https://github.com/mlrun/mlrun/blob/57b5f00c7ba0681aa67ff96bbfe8984d1403adb1/server/py/framework/api/deps.py#L71-L84?

mlrun/common/schemas/auth.py
```diff
@@ -110,6 +111,7 @@ def to_resource_string(
             # workflow define how to run a pipeline and can be considered as the specification of a pipeline.
             AuthorizationResourceTypes.workflow: "/projects/{project_name}/workflows/{resource_name}",
             AuthorizationResourceTypes.api_gateway: "/projects/{project_name}/api-gateways/{resource_name}",
+            AuthorizationResourceTypes.tokens: "/tokens",
```
rokatyy: maybe the same as in api -  `user_secrets/tokens`? A bit more clear what is the dedicated place in the code for this

mlrun/db/httpdb.py
```diff
@@ -5390,27 +5390,91 @@ def store_secret_tokens(
...
+                    f"Username: {token.username}, Token name: {token.token_name}, Expiration: {token.expiration}"
+                )
+
+            # As a system admin, list tokens for a specific user
+            user_tokens = db.list_secret_tokens(username="john_doe")
+
+            # As a system admin, list tokens for all users
+            all_tokens = db.list_secret_tokens()
         """
+        if username == "":
```
rokatyy: below in endpoints dir, description of api suggests that it's not mandatory `"Optional username to filter tokens. Only system admins can use this parameter.",`

server/py/services/api/api/endpoints/user_secrets.py
```diff
@@ -51,32 +53,159 @@ async def store_secret_tokens(
...
+    response_model=mlrun.common.schemas.RevokeSecretTokenResponse,
+)
 async def revoke_secret_token(
     name: str,
+    username: Optional[str] = None,
```
rokatyy: There is also some inconsistency in terms of required fields. Typing suggests username is optional, while below we raise an error if it is such. If this is a normal user, should not we have all the required information in auth_info?

rokatyy (commented):
A couple of questions below

mlrun/db/httpdb.py
elbamit: Why?

rokatyy: @elbamit because it doens't depend on migration, does it?

liranbg: I think the question is - do we want this endpoint to be operated when mlrun is not "ready to serve requests" (due to migrations)? - and the answer is - 
1. Depending on the request - does it consume db session? if so - can't
2. Do we consider it as critical path? if no - not necessarily

liranbg (changes requested):

mlrun/db/httpdb.py
elbamit: 1. Doesn't consume db session, but requires an httpdb object which I'm not sure will be available.
2. Not a critical path - I don't see any real reason why we would need to allow token revocation when mlrun is not ready to server requests

server/py/services/api/api/endpoints/user_secrets.py
elbamit: We raise an error only if it's an empty string (user inputs `""`). `_resolve_target_username_for_revoke_secret_tokens` resolves the username to use, taking into consideration whether it's None or not, and whether it's a system admin who made the request.

mlrun/db/httpdb.py
elbamit: After looking at it with @rokatyy - token revocation is already under the `user-secrets/token` path, therefore no need to make any changes here.

server/py/framework/utils/singletons/k8s.py
```diff
@@ -1283,12 +1290,19 @@ def list_user_token_secrets(
...
         namespace = self.resolve_namespace(namespace)
-        labels = {mlrun_constants.MLRunInternalLabels.auth_username: username}
+        # Always filter by auth token label to only get auth token secrets
+        # Use None as value to perform "label exists" check (more efficient than fetching all secrets)
+        labels = {mlrun_constants.MLRunInternalLabels.auth_token_name: None}
+        # "*" means list all users' tokens, so skip the username filter
+        if username != "*":
+            # Validate username is a valid K8s label value
+            verify_field_regex("username", username, regex.label_value)
+            labels[mlrun_constants.MLRunInternalLabels.auth_username] = username
```
liranbg: seeing this, I think we need to use `sanitize_label_value` before we put the label as well as before we use the label value to list

server/py/framework/utils/singletons/k8s.py
liranbg: I think this is needless

server/py/framework/utils/singletons/k8s.py
```diff
@@ -1538,6 +1575,10 @@ def _get_user_token_secret(
         token_name: str,
         namespace: typing.Optional[str] = None,
     ):
+        # Validate username and token_name are valid K8s label values
```
liranbg: use `sanitize_label_value` here and before storing the secret

server/py/services/api/tests/unit/api/test_user_secrets.py
```diff
@@ -12,23 +12,183 @@
...
 
     assert response.status_code == HTTPStatus.BAD_REQUEST.value
+    mlrun.mlconf.httpdb.authentication.mode = orig_mode
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "action, allowed",
+    [
+        (mlrun.common.schemas.AuthorizationAction.read, True), # System admin
```
Yacouby: move all comments to be above the code ( coding convention )

server/py/services/api/tests/unit/api/test_user_secrets.py
```diff
@@ -12,23 +12,183 @@
...
+    "is_admin, action, username_param, expected_result, expected_error_message",
+    [
+        # === read ===
+        (
+            False,
+            mlrun.common.schemas.AuthorizationAction.read,
+            None,
+            "auth-user",
+            None,
+        ),  # regular user, list without username -> self
```
Yacouby: is this comment for this use case or the one after it ? 
put the comment before the use case so it is clear
( valid for all other cases below )

tests/rundb/test_unit_httpdb.py
```diff
@@ -362,3 +362,28 @@ def test_restricted_methods_in_wrong_mode(monkeypatch, method_name):
...
+    method = getattr(db, method_name)
+
+    with pytest.raises(
+        mlrun.errors.MLRunInvalidArgumentError,
+        match="Username cannot be an empty string.",
+    ):
+        if token_name is not None:
+            method(token_name, username="")
+        else:
+            method(username="")
```
Yacouby: new line missing

server/py/services/api/tests/unit/api/test_user_secrets.py
```diff
@@ -12,23 +12,183 @@
...
+)
+async def test_resolve_target_username(
+    monkeypatch,
+    is_admin: bool,
+    action: mlrun.common.schemas.AuthorizationAction,
+    username_param: str | None,
+    expected_result: str | None,
+    expected_error_message: str | None,
+):
+    async def _fake_is_system_admin(
```
Yacouby: could be moved outside the function if more tests like this are added in the future

Yacouby (changes requested):
minor

server/py/framework/utils/singletons/k8s.py
```diff
@@ -1205,32 +1207,28 @@ def store_user_token_secret(
...
         labels = {
-            mlrun_constants.MLRunInternalLabels.auth_userid: auth_info.user_id,
+            mlrun_constants.MLRunInternalLabels.auth_userid: user_id,
             mlrun_constants.MLRunInternalLabels.auth_token_name: token_name,
         }
-        annotations = {
-            mlrun_constants.InternalAnnotations.auth_username: mlrun.k8s_utils.sanitize_label_value(
-                auth_info.username
-            ),
-        }
```
liranbg, line 1216: why removing?

server/py/services/api/api/endpoints/user_secrets.py
```diff
@@ -51,31 +53,170 @@ async def store_secret_tokens(
...
         name,
+        target_username,
         auth_info,
     )
+
+
+async def _is_system_admin(
+    auth_info: mlrun.common.schemas.AuthInfo,
+    action: mlrun.common.schemas.AuthorizationAction,
+) -> bool:
```
liranbg: the function name and how it does it is not aligned. perhaps "can_list_tokens" is the function?
also, I suggest moving it to `framework/utils/auth/verifier` and provide a wrapper function like

"can_list(resource=tokens, action="GET|LIST", auth_info, mgmt)" or so.

server/py/services/api/crud/secrets.py
```diff
@@ -528,30 +538,45 @@ def revoke_secret_token(
...
+        # Resolve the target user_id from the username
+        target_user_id = self._resolve_target_user_id(auth_info, username)
+
         logger.debug(
             "Revoking secret token for user",
-            user_id=auth_info.user_id,
-            username=auth_info.username,
+            target_user_id=target_user_id,
+            target_username=username,
+            requesting_user=auth_info.username,
```
liranbg, line 559: when are those different?

server/py/services/api/tests/unit/api/test_user_secrets.py
```diff
@@ -12,23 +12,202 @@
...
 
 API_USER_SECRETS_PATH = "/user-secrets"
 API_USER_SECRETS_TOKENS_PATH = API_USER_SECRETS_PATH + "/tokens"
 
+_AUTH_USERNAME = "auth-user"
+_AUTH_USER_ID = "auth-user-id"
+
+
+def _auth_info() -> mlrun.common.schemas.AuthInfo:
+    return mlrun.common.schemas.AuthInfo(username=_AUTH_USERNAME, user_id=_AUTH_USER_ID)
```
liranbg: use fixture decorator

server/py/services/api/tests/unit/api/test_user_secrets.py
```diff
@@ -12,23 +12,202 @@
...
+
+
+def _mock_is_system_admin(
```
liranbg: use fixture

liranbg (changes requested):

server/py/framework/utils/singletons/k8s.py
elbamit: Merge conflict mistake. Fixing it.

server/py/services/api/api/endpoints/user_secrets.py
elbamit: Well there is already `query_global_resource_permissions`. I'll just use it without the `_is_system_admin` wrapper

server/py/services/api/crud/secrets.py
elbamit: When a system admin revokes the token of a different user.
`username="user1"`
`auth_info.username="admin"`

mlrun/db/httpdb.py
```diff
@@ -5392,27 +5392,72 @@ def store_secret_tokens(
...
             endpoint_path,
             "list user secret tokens",
+            params=params,
         )
 
         return mlrun.common.schemas.ListSecretTokensResponse(**response.json())
 
     @mlrun.utils.iguazio_v4_only
-    def revoke_secret_token(self, token_name: str) -> None:
+    def revoke_secret_token(
```
liranbg:
```suggestion
    def delete_secret_token(
```

the revoke is derived from the action. lets be consistent with CRUD operations the HTTPDB is performing.

mlrun/db/httpdb.py
```diff
@@ -5392,27 +5392,72 @@ def store_secret_tokens(
...
+        self, token_name: str, username: Optional[str] = None
+    ) -> mlrun.common.schemas.RevokeSecretTokenResponse:
+        """
+        Revoke a secret token. Only system-administrators can revoke tokens for other users.
+
```
liranbg:
```suggestion
        Delete a secret token. Only system-administrators can revoke tokens for other users.

```

server/py/framework/utils/singletons/k8s.py
```diff
@@ -1205,19 +1209,26 @@ def store_user_token_secret(
         :param namespace: Kubernetes namespace for the secret.
         :return: SecretEventActions.{created, updated, skipped}
         """
+        user_id = auth_info.user_id
+
+        # Sanitize user_id and token_name to be valid K8s label values
+        sanitized_user_id = mlrun.k8s_utils.sanitize_label_value(user_id or "")
+        sanitized_token_name = mlrun.k8s_utils.sanitize_label_value(token_name or "")
```
liranbg: sanitizing the value here is wrong. if the input, given by function callee is bad, then k8s should raise the error. we dont want to get mutated value for callee input as the values here are not "for presentational" reasons but for 1:1 mapping. so mutating the value could lead to unexpected results

server/py/framework/utils/singletons/k8s.py
```diff
@@ -1301,12 +1312,19 @@ def list_user_token_secrets(
...
         """
         namespace = self.resolve_namespace(namespace)
-        labels = {mlrun_constants.MLRunInternalLabels.auth_userid: user_id}
+        # Always filter by auth token label to only get auth token secrets
+        # Use None as value to perform "label exists" check (more efficient than fetching all secrets)
+        labels = {mlrun_constants.MLRunInternalLabels.auth_token_name: None}
+        # "*" means list all users' tokens, so skip the username filter
+        if user_id != "*":
+            # Sanitize user_id to be a valid K8s label value
+            sanitized_user_id = mlrun.k8s_utils.sanitize_label_value(user_id)
```
liranbg: user_id *must* be k8s label value as user_id is uuid. see comment above

server/py/services/api/api/endpoints/user_secrets.py
```diff
@@ -51,31 +53,152 @@ async def store_secret_tokens(
...
     )
 
 
-@router.delete("/tokens/{name}", status_code=HTTPStatus.OK.value)
+@router.delete(
+    "/tokens/{name}",
+    status_code=HTTPStatus.OK.value,
+    response_model=mlrun.common.schemas.RevokeSecretTokenResponse,
+)
 async def revoke_secret_token(
```
liranbg:
```suggestion
async def delete_secret_token(
```

server/py/services/api/api/endpoints/user_secrets.py
```diff
@@ -51,31 +53,152 @@ async def store_secret_tokens(
...
     name: str,
+    username: Optional[str] = None,
     auth_info: mlrun.common.schemas.AuthInfo = fastapi.Depends(
         framework.api.deps.authenticate_request
     ),
-    db_session: Session = fastapi.Depends(framework.api.deps.get_db_session),
 ):
-    # TODO: Support revoking user token with System Admin (ML-10775)
+    """
+    Revoke a secret token.
```
liranbg:
```suggestion
    Delete a secret token.
```

server/py/services/api/crud/secrets.py
```diff
@@ -528,30 +539,43 @@ def revoke_secret_token(
...
+            For regular users, this must be their own username.
+            For system admins, this can be any user's username.
         :param auth_info:
-            Authentication information of the user who owns the token.
+            Authentication information of the requesting user.
+        :return: RevokeSecretTokenResponse with revoked=True if token was revoked,
+                 or revoked=False if token was not found.
         """
+        # Sanitize token_name to be a valid K8s label value
+        sanitized_token_name = mlrun.k8s_utils.sanitize_label_value(token_name or "")
```
liranbg: no need to sanitize token_name. the token_name should be k8s label compatible from first place. if that is not the case we need to discuss it. perhaps encoding it / decoding it to see its original value

liranbg (changes requested):

mlrun/db/httpdb.py
```diff
@@ -5392,27 +5392,72 @@ def store_secret_tokens(
...
+            params=params,
         )
 
         return mlrun.common.schemas.ListSecretTokensResponse(**response.json())
 
     @mlrun.utils.iguazio_v4_only
-    def revoke_secret_token(self, token_name: str) -> None:
+    def delete_secret_token(
+        self, token_name: str, username: Optional[str] = None
+    ) -> mlrun.common.schemas.DeleteSecretTokenResponse:
```
Yacouby, line 5437: the return type should be optional ? to match the declaration in `base.py` ?

mlrun/db/httpdb.py
```diff
@@ -5392,27 +5392,72 @@ def store_secret_tokens(
...
+        :return: A ``ListSecretTokensResponse`` object containing a list of
+                 ``SecretTokenInfo`` objects.
+
+        Example::
+
+            # As a regular user, list your own tokens
+            tokens_response = db.list_secret_tokens()
+            for token in tokens_response.secret_tokens:
+                print(
+                    f"User ID: {token.user_id}, Token name: {token.token_name}, "
```
Yacouby:
```suggestion
                    f"User ID: {token.user_id}, Token name: {token.name}, "
```
SecretTokenInfo name ?

mlrun/db/httpdb.py
```diff
@@ -5392,27 +5392,72 @@ def store_secret_tokens(
...
+        response = self.api_call(
             mlrun.common.types.HTTPMethod.DELETE,
             endpoint_path,
             "delete user secret token",
+            params=params,
         )
+        result = mlrun.common.schemas.DeleteSecretTokenResponse(**response.json())
+        user_info = f" for user '{username}'" if username else ""
+        if result.deleted:
+            logger.info(f"Token '{token_name}'{user_info} was successfully deleted.")
```
Yacouby:
```suggestion
            logger.info("Token was successfully deleted", token_name=token_name, username=username)
```
also for the next warning log

server/py/services/api/crud/secrets.py
```diff
@@ -608,6 +630,60 @@ def get_secret_token(
...
+        :return: The user_id corresponding to the username.
+        :raises mlrun.errors.MLRunNotFoundError: If the user is not found.
+        :raises mlrun.errors.MLRunUnauthorizedError: If the request is unauthorized.
+        """
+        # TODO: move init iguazio_client (ML-11077)
+        iguazio_client = framework.utils.clients.iguazio.v4.Client()
+        try:
+            return iguazio_client.get_user_id_by_username(username, request_headers)
+        except mlrun.errors.MLRunUnauthorizedError as exc:
+            raise mlrun.errors.MLRunNotFoundError(
```
Yacouby: this will incorrectly report "user not found" instead of an authorization issue

Yacouby (changes requested):
Looks great !
last minor suggestions

server/py/services/api/crud/secrets.py
elbamit: `db = mlrun.get_run_db()`
`tokens = db.list_secret_tokens(username="non-existing-user")`

This returns:
`mlrun.errors.MLRunNotFoundError: list user secret tokens details: MLRunNotFoundError("User 'non-existing-user' not found")`

Why do you this error is incorrect?

Yacouby: I meant in case of `MLRunUnauthorizedError` error, you are catching it but raising `MLRunNotFoundError`

elbamit: I remove the `_get_user_id_by_username` function and changed the error type in `get_user_id_by_username` (from v4.py)

Yacouby (approved):

liranbg (approved):

PR #9188, opened 2026-01-10
[Nuclio] Add RabbitMQ trigger support (ML-7879)

alxtkr77:
## Summary

Add support for RabbitMQ triggers in Nuclio functions, enabling MLRun users to consume messages from RabbitMQ queues or topic-based routing.

## Changes Made

- **New `RabbitMQTrigger` class** (`mlrun/runtimes/nuclio/triggers.py`)
  - Supports queue-based and topic-based (routing keys) consumption
  - Handles credential extraction from URL or explicit parameters
  - Integrates with datastore profiles for credential management
  - Explicit parameters override profile defaults (including falsy values like 0)

- **New `add_rabbitmq_trigger()` method** on `RemoteRuntime` (`mlrun/runtimes/nuclio/function.py`)
  - Convenience method for adding RabbitMQ triggers
  - Full parameter support matching Nuclio's RabbitMQ trigger spec

- **New `DatastoreProfileRabbitMQ`** (`mlrun/datastore/datastore_profile.py`)
  - Secure credential management via datastore profiles
  - Supports all RabbitMQ trigger configuration options

- **Dependencies**
  - Added `pika~=1.3` for RabbitMQ client support

## Testing

- 13 unit tests covering:
  - Queue-based triggers
  - Topic-based triggers (routing keys)
  - Credential extraction from URL
  - Error handling configuration
  - Validation (queue_name vs topics mutual exclusivity)
  - Datastore profile integration
  - Explicit falsy value override

- 3 system tests covering:
  - End-to-end message flow (RabbitMQ → Nuclio → MLRun graph → Parquet output)
  - Topic-based routing with wildcards
  - Nuclio spec configuration verification

## Example Usage

```python
# Direct configuration
function.add_rabbitmq_trigger(
    url="amqp://rabbitmq-host:5672",
    exchange_name="my-exchange",
    queue_name="my-queue",
    username="user",
    password="pass",
)

# With topics (routing keys)
function.add_rabbitmq_trigger(
    url="amqp://rabbitmq-host:5672",
    exchange_name="my-exchange",
    topics=["events.#", "notifications.*"],
)

# Using datastore profile
function.add_rabbitmq_trigger(url="ds://my-rabbitmq-profile")
```

## Reference

- Jira: [ML-7879](https://iguazio.atlassian.net/browse/ML-7879)
- Design: [Confluence - RabbitMQ Trigger Support](https://iguazio.atlassian.net/wiki/spaces/Dataflows/pages/621805571/ML-7879+-+RabbitMQ+Trigger+Support+for+MLRun)
- Nuclio docs: https://docs.nuclio.io/en/latest/reference/triggers/rabbitmq.html

## Breaking Changes

None - this is a new feature addition.

[ML-7879]: https://iguazio.atlassian.net/browse/ML-7879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/runtimes/nuclio/triggers.py
+++ b/mlrun/runtimes/nuclio/triggers.py
@@ -86,13 +86,12 @@ class RabbitMQTrigger(NuclioTrigger):
         :param url:                       RabbitMQ connection URL in AMQP format
                                           (e.g., 'amqp://host:port' or 'amqp://user:pass@host:port')
                                           or a datastore profile URL (e.g., 'ds://profile-name')
-        :param exchange_name:             The exchange that contains the queue (required unless
-                                          using a datastore profile)
-        :param queue_name:                Specific queue to consume from. Either queue_name or
-                                          topics must be specified, but not both.
+        :param exchange_name:             The exchange that contains the queue
+        :param queue_name:                Specific queue to consume from. Mutually exclusive
+                                          with topics.
         :param topics:                    List of topics (routing keys) to subscribe to. Creates
-                                          a unique queue and binds it to these routing keys. Either
-                                          queue_name or topics must be specified, but not both.
+                                          a unique queue and binds it to these routing keys.
+                                          Mutually exclusive with queue_name.
         :param username:                  RabbitMQ username (can also be embedded in URL)
         :param password:                  RabbitMQ password (can also be embedded in URL)
         :param prefetch_count:            Broker channel prefetch limit (0 = unlimited)
@@ -152,38 +151,12 @@ class RabbitMQTrigger(NuclioTrigger):
             if worker_termination_timeout is None:
                 worker_termination_timeout = attrs.get("worker_termination_timeout")
 
-        # Apply defaults for parameters still None after profile merge
-        if prefetch_count is None:
-            prefetch_count = 0
-        if durable_exchange is None:
-            durable_exchange = False
-        if durable_queue is None:
-            durable_queue = False
-        if on_error is None:
-            on_error = "nack"
-        if requeue_on_error is None:
-            requeue_on_error = False
-        if reconnect_duration is None:
-            reconnect_duration = "5m"
-        if reconnect_interval is None:
-            reconnect_interval = "15s"
-        if num_workers is None:
-            num_workers = 1
-        if worker_termination_timeout is None:
-            worker_termination_timeout = "10s"
-
-        # Validate exchange_name is provided
-        if not exchange_name:
-            raise ValueError("exchange_name is required")
-
-        # Validate that exactly one of queue_name or topics is specified
+        # Validate that queue_name and topics are mutually exclusive
         if queue_name and topics:
             raise ValueError("Cannot specify both queue_name and topics. Choose one.")
-        if not queue_name and not topics:
-            raise ValueError("Must specify either queue_name or topics.")
 
-        # Validate on_error value
-        if on_error not in ("ack", "nack"):
+        # Validate on_error value if provided
+        if on_error is not None and on_error not in ("ack", "nack"):
             raise ValueError(f"on_error must be 'ack' or 'nack', got '{on_error}'")
 
         # Extract credentials from URL if not provided explicitly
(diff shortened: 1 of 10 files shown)


mlrun/runtimes/nuclio/triggers.py
```diff
@@ -0,0 +1,240 @@
...
+        if requeue_on_error is None:
+            requeue_on_error = False
+        if reconnect_duration is None:
+            reconnect_duration = "5m"
+        if reconnect_interval is None:
+            reconnect_interval = "15s"
+        if num_workers is None:
+            num_workers = 1
+        if worker_termination_timeout is None:
+            worker_termination_timeout = "10s"
```
rokatyy: it will be enriched on nuclio side as well. Are we sure that we want it to be set to defaults in mlrun? As far as I remember we normally rely on default values in nuclio in this case..

mlrun/runtimes/nuclio/triggers.py
```diff
@@ -0,0 +1,240 @@
...
+
+        # Validate exchange_name is provided
+        if not exchange_name:
```
rokatyy: it's not mandatory. The only mandatory field is queue. Because nuclio is a consumer. it is not a mandatory step for nuclio to create resources. While nuclio can prepare infra if user haven't done it in advance, we should not enforce that. Same regarding topics.

rokatyy (changes requested):
Looking good!

Left some comments regarding trigger validations.

mlrun/runtimes/nuclio/triggers.py
alxtkr77: done

mlrun/runtimes/nuclio/triggers.py
alxtkr77: done

mlrun/runtimes/nuclio/triggers.py
```diff
@@ -0,0 +1,221 @@
...
+
+        # Build clean URL without credentials if they were embedded
+        if parsed_url.username or parsed_url.password:
+            # Reconstruct URL without credentials
+            clean_url = f"{parsed_url.scheme}://{parsed_url.hostname}"
+            if parsed_url.port:
+                clean_url += f":{parsed_url.port}"
+            if parsed_url.path:
+                clean_url += parsed_url.path
+            url = clean_url
```
rokatyy: FYI - this logic is performed on nuclio side as well, but i think it's nice to have on mlrun side as well. I'd only suggest having this logic not in rabbitmq trigger, but in some helper function outside of rabbitmq class, so that it can be reused in the future.

mlrun/runtimes/nuclio/triggers.py
```diff
@@ -0,0 +1,221 @@
...
+            if requeue_on_error is None:
+                requeue_on_error = attrs.get("requeue_on_error")
+            if reconnect_duration is None:
+                reconnect_duration = attrs.get("reconnect_duration")
+            if reconnect_interval is None:
+                reconnect_interval = attrs.get("reconnect_interval")
+            if num_workers is None:
+                num_workers = attrs.get("num_workers")
+            if worker_termination_timeout is None:
+                worker_termination_timeout = attrs.get("worker_termination_timeout")
```
rokatyy: can be a separate func `def _resolve_datastore_profile`

mlrun/runtimes/nuclio/triggers.py
```diff
@@ -0,0 +1,221 @@
...
+        if prefetch_count is not None:
+            struct["attributes"]["prefetchCount"] = prefetch_count
+        if durable_exchange is not None:
+            struct["attributes"]["durableExchange"] = durable_exchange
+        if durable_queue is not None:
+            struct["attributes"]["durableQueue"] = durable_queue
+        if on_error is not None:
+            struct["attributes"]["onError"] = on_error
+        if requeue_on_error is not None:
+            struct["attributes"]["requeueOnError"] = requeue_on_error
```
rokatyy: more readable:
```
attribute_mapping = {
    "exchangeName": exchange_name,
    "queueName": queue_name,
    ...
}

for key, value in attribute_mapping.items():
    if value is not None:
        struct["attributes"][key] = value
```

rokatyy (commented):
Some suggestions inline

mlrun/runtimes/nuclio/triggers.py
alxtkr77: fixed

mlrun/runtimes/nuclio/triggers.py
alxtkr77: fixed

mlrun/runtimes/nuclio/triggers.py
alxtkr77: fixed

mlrun/runtimes/nuclio/triggers.py
```diff
@@ -12,11 +12,57 @@
...
+    # Reconstruct URL without credentials
+    clean_url = f"{parsed.scheme}://{parsed.hostname}"
+    if parsed.port:
+        clean_url += f":{parsed.port}"
+    if parsed.path:
+        clean_url += parsed.path
+
+    # Decode URL-encoded characters (e.g., %40 -> @, %20 -> space)
+    username = unquote(parsed.username) if parsed.username else None
+    password = unquote(parsed.password) if parsed.password else None
```
rokatyy: copy paste of https://github.com/mlrun/mlrun/blob/82428637340a4819ba9529900d2b4f4a609a2287/mlrun/runtimes/nuclio/function.py#L1166-L1186

rokatyy (commented):
Looking good, 

just 1 comments below

mlrun/runtimes/nuclio/triggers.py
alxtkr77: I see the similarity, but looking closer the actual overlap is just 3-4 lines (urlparse, credential check, username/password extraction). The implementations differ in important ways:                 
                                                                                                                                                                                                          
  - URL reconstruction: triggers.py uses simple string concatenation, function.py uses urlunparse() to preserve query params/fragments                                                                    
  - URL decoding: triggers.py uses unquote() to decode special characters (e.g., %40 → @), function.py doesn't                                                                                            
  - Return values: One returns a NamedTuple, the other modifies a dict in place                                                                                                                           
                                                                                                                                                                                                         
Extracting a shared helper would either need to accommodate both use cases (adding complexity) or we'd still need wrapper logic in each place. I think the current approach is clearer given the different requirements at each layer.

rokatyy: @alxtkr77 it can be just a common helper function that parses a url and returns `cleanURL, username, password`. It's fine to leave the wrappers, but it just nice to have at least part of the logic in a single place

rokatyy (approved):

PR #9190, opened 2026-01-10
[Package] Support Collection as Input

guy1992l:
### 📝 Description

Allowing users to pass a `dict` or a `list` as elements in the `inputs` dictionary of an MLRun function's `run` method - basically passing an input of inputs. 

https://iguazio.atlassian.net/browse/ML-11892

---

### 🛠️ Changes Made

- [x] Updated `mlrun.package` type hints for Python 3.11.
- [x] `inputs` for an MLRun function `run` method can now be passed as collections (lists and dictionaries) of inputs.
- [x] Added new configurations to `mlconf`:
  - [x] `auto_unpack_inputs=False`: Whether to automatically unpack inputs with no type hints instead of leaving them as 
    `mlrun.DataItem` objects. If True, all inputs without type hints that were originally logged via `mlrun.package` 
    will be unpacked automatically. If False, the inputs will remain `DataItem`s. Default is False.
  - [x] `auto_pack_outputs=False` and `auto_pack_key="artifact"`: Whether to automatically pack outputs, even if no log hints were provided by the user running the function. If True, returned objects will be packed with their default 
    packager and their artifact key will be equal to the following name template: 
    `"<function_name>-<handler>-<auto_pack_key>-<i>"` where `"i"` is enumerated. If False, the returned objects will 
    simply be ignored. Default is False.
- [x] Bundle and unbundle feature:
  - [x] Made packagers support unpacking a collection of inputs via inner arguments of type hints (or default behavior of the new available configurations). 
  - [x] Any bundle-able type can be returned as a collection now and its inner items will be packaged separately. 
- [x] Type hinting changed for Python 3.10+ format. 

**IMPORTANT**: Change docs for `mlrun_op` in `mlrun_pipelines` to reflect the new functionality. In addition in line 278 in `ops.py` change:
```python
for input_param, val in inputs.items():
   cmd += ["-i", f"{input_param}={val}"]
```
to:
```python
for input_param, val in inputs.items():
   cmd += ["-i", f"{input_param}={json.dumps(val) if isinstance(val, dict | list) else val}"]
```

---

### ✅ Checklist

---

### 🧪 Testing
  
  
Multiple tessts were added to check any new feature / function / method.
---

### 🔗 References
- Ticket link:
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [ ] No

---

### DIFF
--- a/mlrun/config.py
+++ b/mlrun/config.py
@@ -856,13 +856,23 @@ default_config = {
         # Whether to enable packagers. True will wrap each run in the `mlrun.package.handler` decorator to log and parse
         # using packagers.
         "enabled": True,
+        # Whether to automatically unpack inputs with no type hints instead of leaving them as `mlrun.DataItem` objects.
+        # If True, all inputs without type hints that were originally logged via `mlrun.package` will be unpacked
+        # automatically. Default is False.
+        "auto_unpack_inputs": False,
+        # Whether to automatically pack outputs, even if not log hints were provided by the user running the function.
+        # If True, returned objects will be packed with their default packager and their artifact key will be equal to
+        # the following name template: "<context_name>-<auto_pack_key>-<i>" where "i" is enumerated. If
+        # False, the returned objects will simply be ignored. Default is False.
+        "auto_pack_outputs": False,
+        "auto_pack_key": "artifact",
         # Whether to treat returned tuples from functions as a tuple and not as multiple returned items. If True, all
         # returned values will be packaged together as the tuple they are returned in. Default is False to enable
         # logging multiple returned items.
         "pack_tuples": False,
         # In multi-workers run, only the logging worker will pack the outputs and log the results and artifacts.
         # Otherwise, the workers will log the results and artifacts using the same keys, overriding them. It is common
-        # that only the main worker (usualy rank 0) will log, so this is the default value.
+        # that only the main worker (usually rank 0) will log, so this is the default value.
         "logging_worker": 0,
         # TODO: Consider adding support for logging from all workers (ignoring the `logging_worker`) and add the worker
         #       number to the artifact / result key (like "<key>-rank<#>". Results can have reduce operation in the
--- a/mlrun/execution.py
+++ b/mlrun/execution.py
@@ -585,9 +585,10 @@ class MLClientCtx:
 
         def recursive_get_input(input_key: str, input_url: str | dict | list):
             if isinstance(input_url, dict):
+                inputs_dict = {}
                 for k, v in input_url.items():
-                    input_url[k] = recursive_get_input(k, v)
-                return input_url
+                    inputs_dict[k] = recursive_get_input(k, v)
+                return inputs_dict
             if isinstance(input_url, list):
                 return [
                     recursive_get_input(f"{input_key}_{i}", v)
--- a/mlrun/execution.py
+++ b/mlrun/execution.py
@@ -562,7 +562,7 @@ class MLClientCtx:
         """
         return mlrun.get_secret_or_env(key, secret_provider=self._secrets_manager)
 
-    def get_input(self, key: str, url: str = ""):
+    def get_input(self, key: str, url: str | dict | list = ""):
         """
         Get an input :py:class:`~mlrun.DataItem` object,
         data objects have methods such as .get(), .download(), .url, .. to access the actual data.
@@ -574,7 +574,7 @@ class MLClientCtx:
 
         :param key:  The key name for the input url entry.
         :param url:  The url of the input data (file, stream, ..) - optional, saved in the inputs dictionary
-                     if the key is not already present.
+                     if the key is not already present. Can be passed as a list or dictionary of urls as well.
 
         :return:     :py:class:`~mlrun.datastore.base.DataItem` object
         """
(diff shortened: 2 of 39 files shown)


mlrun/execution.py
```diff
@@ -574,20 +574,34 @@ def get_input(self, key: str, url: str = ""):
...
-            key,
-            project=self._project,
-            allow_empty_resources=self._allow_empty_resources,
-        )
+
+        def recursive_get_input(input_key: str, input_url: str | dict | list):
+            if isinstance(input_url, dict):
+                for k, v in input_url.items():
+                    input_url[k] = recursive_get_input(k, v)
+                return input_url
```
theSaarco: Any specific reason for doing in-place replacement here, rather than generate a new dict like you do in the list case?

mlrun/package/utils/type_hint_utils.py
```diff
@@ -290,6 +288,20 @@ def _reduce_type_hint(type_hint: type) -> list[type]:
...
+        #       * `List[int | str | dict[str, int | float]]` will be reduced to `List[int | str | dict[str, int]]` and
+        #         `List[int | str | dict[str, float]]`, which then both will yield `List[int | str | dict]`, and finally
+        #         `list`.
+        #       The algorithm should find the deepest typing type possible (like union) and reduce it. Once there are no
+        #       more typing types to reduce, the deepest args should be reduced until we reach the core origin we now
+        #       return by default. If there are two or more args to reduce, all combinations (permutations) should be
+        #       returned.
+        #       For now, we reduce only to the origin type itself regardless of its args, as it seems to be sufficient
+        #       for our users.
+        return [origin]
```
theSaarco, line 408: This is a change in logic - before this PR it would return `[]` and now you do return the origin, does that mean a change in what the user eventually sees?

mlrun/config.py
```diff
@@ -836,13 +836,23 @@
...
         # using packagers.
         "enabled": True,
+        # Whether to automatically unpack inputs with no type hints instead of leaving them as `mlrun.DataItem` objects.
+        # If True, all inputs without type hints that were originally logged via `mlrun.package` will be unpacked
+        # automatically. Default is False.
+        "auto_unpack_inputs": False,
+        # Whether to automatically pack outputs, even if not log hints were provided by the user running the function.
+        # If True, returned objects will be packed with their default packager and their artifact key will be equal to
+        # the following name template: "<context_name>-<auto_pack_key>-<i>" where "i" is enumerated. If
+        # False, the returned objects will simply be ignored. Default is False.
```
theSaarco, line 866: What is the meaning of ignored here? Do we drop them from the list of returned objects? Or simply not packaging them?

theSaarco (approved):
I have some comments, but I don't think they're critical.
The main thing I'd like to have is a documentation page (not just docstrings, which we have plenty) that explains the entire concept and usage of packagers, including this special use-case. I guess that's going to happen in another PR?

mlrun/execution.py
guy1992l: no particular reason, I will create a new dict.

mlrun/package/utils/type_hint_utils.py
guy1992l: no, when it was return `[]` it still returns `[]`. I changed the order of the logic a little.

mlrun/config.py
guy1992l: they will not be packaged - same as today.

mlrun/package/packagers/default_packager.py
```diff
@@ -284,6 +301,12 @@ def unpack_x(self, data_item: mlrun.DataItem, ...) -> Any:
     #: The default artifact type to unpack from.
     DEFAULT_UNPACKING_ARTIFACT_TYPE = ArtifactType.OBJECT
 
+    #: Whether the `PACKABLE_OBJECT_TYPE` can be constructed from a list to be used as a collection.
```
guy1992l: need to align this comment

mlrun/package/packagers/python_standard_library_packagers.py
```diff
@@ -465,6 +488,9 @@ class BytesPackager(ListPackager):
     """
 
     PACKABLE_OBJECT_TYPE = bytes
+    # Disable bundling since bytes items must be integers (0-255), and IntPackager doesn't support
+    # file unpacking, making bundling from files impractical.
```
guy1992l: old unrelated comment

mlrun/package/utils/type_hint_utils.py
```diff
@@ -38,15 +41,46 @@ def is_typing_type(type_hint: type) -> bool:
...
+            return False
+
+        # Check for generic aliases (like `list[int]` or `List[int]`), which are not pure type hints
+        # (`typing._GenericAlias` is tested via `type(typing.List[int])`):
+        if (
+            type(type_hint)
+            is types.GenericAlias  # Python objects as type hints: list[int], dict[str, int], etc.
+            or type(type_hint)
+            is type(
+                list[int]
```
guy1992l:
```suggestion
                typing.List[int]
```

mlrun/package/__init__.py
```diff
@@ -38,18 +40,13 @@
 
 
 def handler(
-    labels: Optional[dict[str, str]] = None,
```
guy1992l, line 41: add it back with deprecate

mlrun/config.py
```diff
@@ -856,13 +856,23 @@
         # Whether to enable packagers. True will wrap each run in the `mlrun.package.handler` decorator to log and parse
         # using packagers.
         "enabled": True,
+        # Whether to automatically unpack inputs with no type hints instead of leaving them as `mlrun.DataItem` objects.
```
guy1992l, line 859: add all the package configs to client spec

mlrun/package/packagers_manager.py
```diff
@@ -173,80 +173,115 @@ class YPackager(Packager):
...
-    ) -> Union[Artifact, dict, None, list[Union[Artifact, dict, None]]]:
+        self,
+        obj: Any,
+        log_hint: dict[str, str],
+        _unbundle_level: int | None = None,
+    ) -> Artifact | dict | None | list[Artifact | dict | None]:
         """
-        Pack an object using one of the manager's packagers. A `dict` ("**") or `list` ("*") unpacking syntax in the
-        log hint key packs the objects within them in separate packages.
+        Pack an object using one of the manager's packagers.
```
guy1992l, line 201: treat ** as * with warning

liranbg (approved):

PR #9199, opened 2026-01-13
[Serving] Resolve the final step within cyclic graph

davesh0812:
Had an option to include the last step within a cycle

https://iguazio.atlassian.net/browse/ML-11938


mlrun/serving/states.py
```diff
@@ -2890,8 +2889,9 @@ def get_first_function_step(step, current_function):
                 )
             default_final_step = self.final_step
 
-        elif len(self._start_steps) == 1:
+        elif len(self._start_steps) == 1 and not self.allow_cyclic:
             # find the final step in case if a simple sequence of steps
+            # default_final_step is used ony for feature sets therefore we won't be set it when cycles are allowed
```
gtopper:
```suggestion
            # default_final_step is used only for feature sets therefore it won't be set when cycles are allowed
```

tests/serving/test_async_flow.py
```diff
@@ -104,14 +104,15 @@ def test_async_basic():
...
-    logger.info(f"\nAsync Flow:\n{flow.to_yaml()}")
-    resp = server.test(body=[])
-
-    server.wait_for_completion()
+    try:
+        # plot the graph for test & debug
+        flow.plot(f"{results}/serving/async.png")
+        server = function.to_mock_server()
+        server.context.visits = {}
+        logger.info(f"\nAsync Flow:\n{flow.to_yaml()}")
```
gtopper: Try should start after server is created. Something like:
```suggestion
    # plot the graph for test & debug
    flow.plot(f"{results}/serving/async.png")
    logger.info(f"\nAsync Flow:\n{flow.to_yaml()}")
    server = function.to_mock_server()
    try:
        server.context.visits = {}
```

tests/serving/test_async_flow.py
```diff
@@ -1473,6 +1476,33 @@ def test_cyclic_to_first_step(method):
         server.wait_for_completion()
 
 
+@pytest.mark.parametrize("method", ["add_step", "to"])
```
gtopper, line 1480: Nice to link back to the issue:
```suggestion
# ML-11938
@pytest.mark.parametrize("method", ["add_step", "to"])
```

mlrun/serving/states.py
```diff
@@ -2890,8 +2889,9 @@ def get_first_function_step(step, current_function):
                 )
             default_final_step = self.final_step
 
-        elif len(self._start_steps) == 1:
+        elif len(self._start_steps) == 1 and not self.allow_cyclic:
             # find the final step in case if a simple sequence of steps
```
gtopper:
```suggestion
            # find the final step in case of a simple sequence of steps
```

gtopper (commented):
LGTM otherwise.

gtopper (approved):

PR #9205, opened 2026-01-14
[Authentication] Change user secret token from username to userid

liranbg:
### 📝 Description

Change user token secret naming convention to use `auth_info.user_id` instead of `auth_info.username`. This change ensures that user token secrets are stored using a stable identifier (`user_id` which is a UUID) rather than usernames which may contain characters invalid for Kubernetes resource names.

---

### 🛠️ Changes Made

- Changed `store_user_token_secret` to accept `auth_info` directly instead of `username`
- Added `auth_userid` label from `auth_info.user_id` for secret identification
- Updated `resolve_auth_token_secret_name` to use `user_id` parameter
- Added `InternalAnnotations` class in constants for annotation keys

---

### ✅ Checklist

---

### 🧪 Testing
- Unit tests + ig4 running KFP / Kubejob

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11976
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/server/py/services/api/crud/secrets.py
+++ b/server/py/services/api/crud/secrets.py
@@ -466,7 +466,7 @@ class Secrets(
             expiration = token_info["token_exp"]
 
             action = self.secrets_provider.store_user_token_secret(
-                username=auth_info.username,
+                auth_info=auth_info,
                 token_name=token_name,
                 token=token,
                 expiration=expiration,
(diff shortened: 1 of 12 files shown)


server/py/services/api/crud/secrets.py
```diff
@@ -497,27 +497,17 @@ def store_secret_tokens(
 
     def list_secret_tokens(
         self,
-        authenticated_username: str,
+        auth_info: mlrun.common.schemas.AuthInfo,
```
quaark, line 500: this method needs to support explicit user id as well for the admin

quaark (changes requested):
What I'm missing from the PR is the option to still supply a username (on list / revoke) and have mlrun use iguazio sdk to resolve the user id. If all that is waiting for ML-10775, then ignore this comment

server/py/services/api/crud/secrets.py
liranbg: This will be addressed on another PR which provides the ability for sys admin to list other users tokens

quaark (approved):

elbamit (approved):
Looking good. Minor question - don't you want to sanitize everything that is saved as a label? user_id and token_name?

PR #9211, opened 2026-01-16
[Projects] Ensure project spec owner

liranbg:
### 📝 Description

This PR ensures that the project owner is automatically set from the authenticated user's information when a project is created or stored without an explicit owner. Previously, projects could be created without an owner field being populated.

---

### 🛠️ Changes Made

- **Set project owner from auth info**: Added logic in `_enrich_project` to populate `project.spec.owner` from `auth_info.username` when the owner is not already set
- **Consolidated enrichment logic**: Merged `_enrich_project_create` into `_enrich_project` for a single enrichment path
- **Update local project object**: After `save_to_db`, the local project object is now updated with server-side changes (e.g., generated metadata)
- **Removed unused `session` parameter**: Cleaned up `_transform_project_record_to_schema` by removing the unused `session` argument
- **Removed legacy migration code**: Removed pre-0.6.0 project migration code that auto-fixed incomplete project records on read
- **Minor fixes**: Fixed Makefile `.PHONY` target typo, replaced deprecated `datetime.utcnow()` with `datetime.now(UTC)`

---

### ✅ Checklist

---

### 🧪 Testing

- Manually tested project creation via SDK to verify owner is populated

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11868
- Design docs links: 
- External links: 

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes

- The legacy migration code for pre-0.6.0 projects was removed as it's no longer needed (MLRun 0.6.0 was released years ago)
- Owner is only set if `project.spec.owner` is `None`, preserving explicitly provided owners

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -3756,7 +3756,8 @@ class MlrunProject(ModelObj):
         :store: if True, allow updating in case project already exists
         """
         self.export(filepath)
-        self.save_to_db(store)
+        project = self.save_to_db(store)
+        self.__dict__.update(project.__dict__)
         return self
 
     def save_to_db(self, store=True):
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -3373,7 +3373,16 @@ class MlrunProject(ModelObj):
                         function_definition, self, name
                     )
                 except FileNotFoundError as exc:
-                    message = f"File {exc.filename} not found while syncing project functions."
+                    filename = getattr(exc, "filename", None) or getattr(
+                        exc, "filename2", None
+                    )
+                    if not filename:
+                        filename = getattr(exc, "args", ["unknown"])[0]
+                        # "not found" is a common error message, remove it
+                        filename = filename.replace("not found", "").strip()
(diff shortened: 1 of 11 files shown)


mlrun/projects/project.py
```diff
@@ -3373,7 +3373,16 @@ def sync_functions(
                         function_definition, self, name
                     )
                 except FileNotFoundError as exc:
-                    message = f"File {exc.filename} not found while syncing project functions."
+                    filename = getattr(exc, "filename", None) or getattr(
+                        exc, "filename2", None
```
yaelgen, line 3377: what is this `filename2`?

yaelgen (approved):
added a question, also I would try to add a UT to make sure owner is set from auth_info only when not explicitly provided

mlrun/projects/project.py
liranbg: https://docs.python.org/3/library/exceptions.html#OSError.filename2

PR #9217, opened 2026-01-18
[Serving] Validate that `BaseStep`'s `after` parameter is a list

gtopper:
to prevent it being iterated over character by character, resulting in confusing errors to the user.

Added regression test.

[ML-11985](https://iguazio.atlassian.net/browse/ML-11985)

[ML-11985]: https://iguazio.atlassian.net/browse/ML-11985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/serving/states.py
+++ b/mlrun/serving/states.py
@@ -148,9 +148,7 @@ class BaseStep(ModelObj):
         self._parent = None
         self.comment = None
         self.context = None
-        if after and not isinstance(after, list):
-            raise mlrun.errors.MLRunInvalidArgumentError("after must be a list")
-        self.after = after or []
+        self.after = after
         self._next = None
         self.shape = shape
         self.on_error = None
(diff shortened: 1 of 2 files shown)


mlrun/serving/states.py
```diff
@@ -148,6 +148,8 @@ def __init__(
         self._parent = None
         self.comment = None
         self.context = None
+        if after and not isinstance(after, list):
+            raise mlrun.errors.MLRunInvalidArgumentError("after must be a list")
```
davesh0812: Why not to have the same behavior like we have in 'add_step' method
```suggestion
        if after:
            self.after = after if isinstance(after, list) else [after]
        else:
            self.after = []
```

davesh0812 (approved):

PR #9221, opened 2026-01-19
[Model Monitoring] Fix incorrect JSON fallback for v2+ writer (ML-11979)

alxtkr77:
## Summary
When `feature_analysis=True` is requested and parquet stats don't exist yet, the code incorrectly fell back to reading JSON stats files. However, JSON stats files are only created by v1 writer - v2+ writer stores stats exclusively in parquet format.

This caused unnecessary warnings in logs for v2+ writer deployments:
```
Stats file was not found at path v3io://.../stats/{endpoint_id}_current_stats.json
```

## Changes Made
- Added version check before JSON fallback in `get_model_endpoint()`
- JSON stats lookup now only runs when `writer_version == "v1"`
- For v2+ writer, empty stats are the expected state until drift analysis runs

## Testing
- Ran model monitoring endpoint unit tests (35 passed)
- Verified fix against vmdev76 with actual model endpoint
- Analyzed production logs confirming the issue pattern

## Reference
- Jira: ML-11979

## Breaking Changes
- No

### DIFF
--- a/mlrun/model_monitoring/db/_stats.py
+++ b/mlrun/model_monitoring/db/_stats.py
@@ -89,9 +89,10 @@ class ModelMonitoringStatsFile(abc.ABC):
             # Different errors are raised for S3 or local storage, see ML-8042
             FileNotFoundError,
         ) as err:
-            logger.warning(
-                "The Stats file was not found. It should have been created "
-                "as a part of the model endpoint's creation",
+            logger.debug(
+                "Stats file not found. This is expected for v2+ writer which stores "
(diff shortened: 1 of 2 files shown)


davesh0812 (commented):
This change may introduce a BC issue for endpoints created in version 1.10 that did not receive new invocations after upgrading to 1.11.0. In such cases, the endpoint may not return drift_measures and current_measures, even though they exist.

davesh0812 (approved):

PR #9223, opened 2026-01-19
[Runtime] Forbid Using Specific Service Accounts in MLRun Jobs

quaark:
### 📝 Description
This PR introduces enhancements to the service account validation mechanism in MLRun. It adds support for forbidden service accounts, improves the handling of service account validation, and updates the `AuthInfo` schema to include a `kind` attribute. These changes ensure stricter security and better configurability for service account usage.

---

### 🛠️ Changes Made
- Added `AuthInfoKind` to the `AuthInfo` schema to differentiate between user and service account types.
- Introduced forbidden service accounts in the configuration and validation logic.
- Updated `validate_service_account` methods to handle forbidden service accounts.
- Enhanced the `resolve_project_default_service_account` function to include forbidden service accounts.
- Modified the `process_function_service_account` function to support forbidden service accounts.
- Updated the Iguazio client to parse and handle the `resourceType` field in authentication responses.
- Added unit tests for the new service account validation logic and Iguazio client changes.

---

### ✅ Checklist
  - Cannot create system test for the feature for 2 reasons:
    - It is only relevant for IG4 and we are yet to support IG4 system tests
    - The system test mlrun client needs access to kubectl and/or run from a pod to get a SA token to test the feature

---

### 🧪 Testing
- Added unit tests for the `validate_service_account` method to verify behavior with allowed and forbidden service accounts.
- Added unit tests for the Iguazio client to ensure proper parsing of the `resourceType` field.
- Verified changes manually by running relevant scenarios in a development environment.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11619

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No


mlrun/common/schemas/auth.py
```diff
@@ -150,6 +156,9 @@ def get_member_ids(self) -> list[str]:
     def get_session(self) -> str:
         return self.data_session or self.session
 
+    def is_service_account(self) -> bool:
+        return self.kind == AuthInfoKind.service_account
```
theSaarco, line 160: Not strictly an issue, but it looks a bit funny for me to have an enum with 2 values, and then add a specific check for one of them...

mlrun/runtimes/base.py
```diff
@@ -254,7 +256,10 @@ def try_auto_mount_based_on_config(self):
         pass
 
     def validate_and_enrich_service_account(
-        self, allowed_service_account, default_service_account
+        self,
+        allowed_service_account,
```
theSaarco:
```suggestion
        allowed_service_accounts,
```

While you're at it 😄

server/py/framework/utils/clients/iguazio/v4.py
```diff
@@ -437,7 +440,7 @@ def _extract_error_message(self, response_body: dict) -> typing.Optional[str]:
     @staticmethod
     def _parse_auth_response_data(
         response_body: typing.Mapping[typing.Any, typing.Any],
-    ) -> tuple[str, str, list[str]]:
+    ) -> tuple[str, str, list[str], str]:
         """
         Validate and parse the authentication response body to extract the username, user ID, and group IDs.
```
theSaarco:
```suggestion
        Validate and parse the authentication response body to extract the username, user ID, group IDs and type of authentication (user or service account).
```

server/py/framework/api/utils.py
```diff
@@ -759,15 +761,20 @@ def process_function_service_account(function):
...
+    )
 
     function.validate_and_enrich_service_account(
-        allowed_service_accounts, default_service_account
+        allowed_service_accounts, forbidden_service_accounts, default_service_account
     )
 
 
-def resolve_project_default_service_account(project_name: str):
+def resolve_project_default_service_account(
```
theSaarco: The name of this function is becoming more and more inadequate - I think originally it returned just the default SA, but now it returns the default SA, allowed and forbidden SAs. So let's change the name of this function to something more appropriate - maybe `resolve_project_service_account_details`?

server/py/framework/api/utils.py
```diff
@@ -783,6 +790,38 @@ def resolve_project_default_service_account(project_name: str):
...
+        )
+
+    # If the auth info's service account is in the forbidden list, remove it from there to allow the current request to
+    # proceed.
+    if (
+        auth_info
+        and auth_info.is_service_account()
+        and auth_info.username in forbidden_service_accounts
+    ):
+        forbidden_service_accounts.remove(auth_info.username)
```
theSaarco, line 824: I am deliberating whether this check is really needed (and generally, whether a service with a given SA is allowed to run jobs with the same SA, even if the SA is considered forbidden for the project).
It's true that we want to prevent users from using specific SAs, and we don't necessarily want to block system services from doing it, but OTOH - these services will not really launch jobs, won't they? And even if they did, they will still be able to launch jobs with other SAs, simply not with their SAs - so I think this is redundant.

diff.log
```diff
@@ -0,0 +1,539 @@
+diff --git a/mlrun/common/schemas/__init__.py b/mlrun/common/schemas/__init__.py
```
theSaarco: ??? 
Remove.

server/py/framework/api/utils.py
theSaarco, line 824: And, if we decide that no need for this check, then the additional `auth_info` parameter in this function can be removed.

theSaarco (changes requested):
Some changes needed, and also there's a question we need to discuss and consider.

mlrun/common/schemas/auth.py
quaark: We usually stay away from booleans in cases like these. Just in case we have more than 2 cases in the future

server/py/framework/api/utils.py
quaark: Why though?
In the case where we have an external service that wants to run jobs as its own identity while we disallow any user to run jobs as that identity. That case is useful in my opinion.

diff.log
quaark: Oops

server/py/framework/api/utils.py
quaark: I think we still need it honestly

theSaarco (approved):
LGTM.

server/py/framework/api/utils.py
```diff
@@ -783,6 +790,39 @@ def resolve_project_default_service_account(project_name: str):
...
+        forbidden_service_accounts.extend(
+            [
+                service_account.strip()
+                for service_account in forbidden_service_accounts_secret.split(",")
+            ]
+        )
+
+    # If the auth info's service account is in the forbidden list, remove it from there to allow the current request to
+    # proceed.
+    # TODO: In the future we should avoid running jobs as a service account that is in the forbidden list altogether.
```
liranbg, line 818: this needs to be mapped to a ticket or it will be lost. what's the current flow we allow? can you be specific?

liranbg (approved):
overall, lgtm. minor question

PR #9227, opened 2026-01-20
[Auth] Token resolution logic for runtimes

elbamit:
### 📝 Description

This PR adds logic to the MLRun API to automatically determine which authentication token to use when none is explicitly provided by the user. The resolution is performed by constructing an `.igz.yml `file from the relevant Kubernetes secrets and using it to initialize an Iguazio SDK client. During client initialization, the existing token selection and validation logic is applied, allowing the resolved refresh token name to be retrieved.

When no token is specified, tokens are selected in the following order:
- The default token, if present
- The first valid (non-expired) token, chosen in alphabetical order

---

### 🛠️ Changes Made

- Added `resolve_token_from_igz_yml` method in Iguazio v4 client module that creates an Iguazio Client using an `.igz.yml `file, then retrieve the token.
- Added `get_user_secret_tokens_as_igz_yml_data` in K8s layer that fetchs user token(s) from k8s secrets in `igz.yml` format.
- Added `list_user_token_secret_values` in K8s layer that lists tokens for the user and return the token values.
- Added `resolve_auth_token_name` helper that fetches user tokens from k8s, construct the `igz.yml` then create the iguazio client to get the token.
- Refactor `resolve_auth_token_secret_name` (used by argo pods) and `enrich_and_validate_auth_token_name` (used by all other runtimes) to use the new `resolve_auth_token_name` helper method

---

### ✅ Checklist

---

### 🧪 Testing
  
  
- Unit tests for `resolve_token_from_igz_yml`
- Unit tests for `list_user_token_secret_values`
- Refactor tests for `enrich_and_validate_auth_token_name`
- Refactor tests for `resolve_auth_token_name`
- Unit tests for `get_user_secret_tokens_as_igz_yml_data`
---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11600
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


server/py/framework/utils/clients/iguazio/v4.py
```diff
@@ -153,6 +153,61 @@ def _revoke_offline_token():
...
+        Creates a temporary file with the provided YAML content and uses the
+        Iguazio SDK's token file resolution to find and validate the token.
+
+        :param igz_yml_content: YAML content with tokens in igz.yml format.
+        :param user_id: The user_id for error messages.
+        :param token_name: Specific token to validate (strict mode), or None (auto-discovery).
+        :return: The resolved token name.
+        :raises MLRunNotFoundError: If no valid token found or validation fails.
+        """
+        import tempfile
```
liranbg: import here? why?

liranbg (changes requested):
This needs to be redesigned wrt to why mlrun-api needs to make token resolution on behalf of the end-client. this could cause a huge performance hit on mlrun-api.

server/py/services/api/launcher.py
```diff
@@ -678,20 +679,23 @@ def _validate_retry(runtime_kind: str, retry: Optional["mlrun.model.Retry"]):
     def enrich_and_validate_auth_token_name(
         self, object: Union[mlrun.run.RunObject, mlrun.runtimes.RemoteRuntime]
     ):
+        if not (
+            mlrun.mlconf.is_iguazio_v4_mode()
+            and self._auth_info
+            and self._auth_info.user_id
```
liranbg: when's auth_info would be not set here?

server/py/services/api/utils/helpers.py
```diff
@@ -75,26 +77,54 @@ def resolve_client_default_kfp_image(
...
+    provided_token_name: typing.Optional[str],
+    user_id: str,
+) -> str:
+    """
+    Resolve the token name for a user using the iguazio SDK.
+
+    Resolution priority (handled by SDK):
+    1. User-provided token name (strict mode - must be valid)
+    2. "default" token
+    3. First valid token (alphabetically sorted)
```
liranbg: id rather not explicitly write it here as it is derived from another sdk. so the lgoic is actually subjected to current external sdk implementation

server/py/framework/utils/singletons/k8s.py
```diff
@@ -1440,6 +1440,55 @@ def get_user_token_secret_value(
...
+                f"No tokens found for user '{user_id}'"
+            )
+
+        secret_tokens = []
+        for token_info in all_token_infos:
+            try:
+                token_value = self.get_user_token_secret_value(
+                    user_id=user_id, token_name=token_info.name
+                )
+                secret_tokens.append({"name": token_info.name, "token": token_value})
```
liranbg: why do we need to get for each token?

liranbg (changes requested):

server/py/framework/utils/singletons/k8s.py
elbamit: You're right that's inefficient because the get operation also performs a list operation from k8s. I'll create a `list_user_token_secret_values` method that returns token values to minimize the number of k8s API calls.

server/py/services/api/launcher.py
elbamit: Well it's optional in the launcher's init. Not sure when this is actually happening.

liranbg: the server side launcher must have auth token - on CE deploymentsthe user_id might be empty but that's anyway wont be relevantw ehn you check if the deployment is igzv4

server/py/services/api/utils/helpers.py
```diff
@@ -75,26 +77,49 @@ def resolve_client_default_kfp_image(
     return image
 
 
+def resolve_auth_token_name(
+    provided_token_name: typing.Optional[str],
+    user_id: str,
+) -> str:
```
quaark, line 83: since it's optional:
```suggestion
def resolve_auth_token_name(
    user_id: str,
    provided_token_name: typing.Optional[str] = None,
) -> str:
```

quaark (changes requested):
From the code POV just a very minor comment, In general looks good.
But I have to say I'm concerned about this approach as it has a specific disadvantage. It means unless a specific token is selected, we always query N secrets per the amount of secrets the user had, while the previous approach stopped on the first successful secret token.
Large amounts of tokens in k8s may also create a massive file and leave the server vulnerable to memory attacks.
So I'm not sure we should ditch the logic completely in mlrun even though it may be code duplication, this could present an issue in high scale.

liranbg (approved):

PR #9239, opened 2026-01-22
[IG4] Fix enable_model_monitoring issues after removing V3IO support

danielperezz:
### 📝 Description

This PR fixes issues in enable_model_monitoring that were introduced after removing V3IO support, ensuring the flow works correctly in IG4 environments.

---

### 🛠️ Changes Made
- Don’t apply the access key in IG4 environments
- Return proper errors when attempting to use V3IO in IG4
- Mount the new environment variable to allow overriding the default store prefix from mlrun-api to the controller (used for controller-schedules files)

---

### ✅ Checklist

---

### 🧪 Testing
  
`enable_model_monitoring` is working as expected and all pods are starting successfully.
I wasn’t able to test end-to-end monitoring due to a failure in the model monitoring application base, caused by a blocking bug related to feature sets ([ML-11986](https://iguazio.atlassian.net/browse/ML-11986)￼). However, up to the application step, everything is running properly.
  

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11605?focusedCommentId=200787&sourceType=mention
- Complementary `mlefi` ticket: https://github.com/iguazio/mlefi/pull/323
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes

[ML-11986]: https://iguazio.atlassian.net/browse/ML-11986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/py/services/api/api/endpoints/nuclio.py
```diff
@@ -525,7 +525,7 @@ def _deploy_nuclio_runtime(
     )
 
     if monitoring_application or serving_to_monitor:
-        if not mlrun.mlconf.is_ce_mode():
+        if not mlrun.mlconf.is_ce_mode() and not mlrun.mlconf.is_iguazio_v4_mode():
```
Eyal-Danieli: I found several more places that we should update this condition in the same approach:
- [stream pod graph](https://github.com/mlrun/mlrun/blob/development/mlrun/model_monitoring/stream_processing.py#L65) 
- [TSDBfactory](https://github.com/mlrun/mlrun/blob/development/mlrun/model_monitoring/db/tsdb/__init__.py#L44)
- [genai_tutorial](https://raw.githubusercontent.com/mlrun/mlrun/refs/heads/development/docs/tutorials/genai-02-model-monitor-llm.ipynb#:~:text=%22if%20not%20mlrun.mlconf.is_ce_mode()%3A%5Cn%22%2C)

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -256,6 +257,16 @@ def deploy_model_monitoring_controller(
             fn = self._get_model_monitoring_controller_function(
                 image=controller_image, ignore_stream_already_exists_failure=overwrite
             )
+
+            # Propagate store-prefix override from the deploying pod into the controller if present.
+            monitoring_store_prefix_env = os.getenv(
```
Eyal-Danieli: The right approach is to use `mlrun.get_secret_or_env()` + I would store `MLRUN_MODEL_ENDPOINT_MONITORING__STORE_PREFIXES__DEFAULT` as a constant in `model_monitoring_constants.py` and use it as a constant

Eyal-Danieli (commented):
Great!
few comments

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -1684,9 +1696,9 @@ def _validate_and_get_tsdb_profile(
         if isinstance(
             tsdb_profile, mlrun.datastore.datastore_profile.DatastoreProfileV3io
         ):
-            if mlrun.mlconf.is_ce_mode():
+            if mlrun.mlconf.is_ce_mode() or mlrun.mlconf.is_iguazio_v4_mode():
                 raise mlrun.errors.MLRunInvalidMMStoreTypeError(
-                    "MLRun CE supports only TimescaleDB TSDB, received a V3IO profile for the TSDB"
+                    "V3IO TSDB profile is not supported, use TimescaleDB instead."
```
elbamit, line 1690: Is TimescaleDB supported for IG4?

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -1768,9 +1780,9 @@ def _validate_v3io_stream(
         self,
         v3io_profile: mlrun.datastore.datastore_profile.DatastoreProfileV3io,
     ) -> None:
-        if mlrun.mlconf.is_ce_mode():
+        if mlrun.mlconf.is_ce_mode() or mlrun.mlconf.is_iguazio_v4_mode():
             raise mlrun.errors.MLRunInvalidMMStoreTypeError(
-                "MLRun CE supports only Kafka streams, received a V3IO profile for the stream"
+                "V3IO stream profile is not supported, use Kafka streams instead."
```
elbamit, line 1774: Is Kafka supported for IG4?

docs/tutorials/src/model_monitoring_utils.py
```diff
@@ -22,7 +22,7 @@ def enable_model_monitoring(
         name=stream_profile_name, v3io_access_key=mlrun.mlconf.get_v3io_access_key()
     )
 
-    if mlrun.mlconf.is_ce_mode():
+    if mlrun.mlconf.is_ce_mode() or mlrun.mlconf.is_iguazio_v4_mode():
```
theSaarco: This change is done a lot of times in this PR, why not wrap this in a utility function in mlconf, something like `is_using_v3io()` or something similar? This will also help understand why we're checking this.

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -256,6 +256,17 @@ def deploy_model_monitoring_controller(
             fn = self._get_model_monitoring_controller_function(
                 image=controller_image, ignore_stream_already_exists_failure=overwrite
             )
+
+            # Propagate store-prefix override from the deploying pod into the controller if present.
+            monitoring_store_prefix_env = mlrun.get_secret_or_env(
+                key=mm_constants.DEFAULT_STORE_PREFIX_ENV, default=None
+            )
```
theSaarco: Not sure I understand the flow here. 
The env key `MLRUN_MODEL_ENDPOINT_MONITORING__STORE_PREFIXES__DEFAULT` will automatically translate on MLRun startup to an MLRun config value `model_endpoint_monitoring.store_prefixes.default`. What's more - project secrets are not mounted on MLRun API pod.
So, there's no reason to use `get_secret_or_env` here - instead you should get it from mlconf. 
If that wasn't the intention, then you shouldn't use this naming format. Env variables beginning with `MLRUN_` are always used to feed into the configuration.

theSaarco (changes requested):
Some small changes needed.

server/py/services/api/crud/model_monitoring/deployment.py
danielperezz: @theSaarco  When I didn’t override this value by propagating the env var to the controller, I observed that the controller still tried to use the default value from mlrun-config (in this case, with the V3IO prefix), which caused it to run into an error.

theSaarco: My concern wasn't passing the env var to the controller - you have to do that to override the MLRun config in the controller. My concern was using `get_secret_or_env` here.

Having said that, the correct way to do this would have been to include this config variable (`model_endpoint_monitoring.store_prefixes.default`) in the `client_spec` mechanism, and ensure that this value is passed to any pod being launched. 
And, as a matter of fact, I see that it is included in it:
```
model_endpoint_monitoring_store_prefixes=self._get_config_value_if_not_default(
    "model_endpoint_monitoring.store_prefixes"
),
```
Since you're modifying the default value by setting an env variable, this should have propagated to the controller code when it imported MLRun. So, I would say you need to check why this mechanism didn't work.

danielperezz: Apparently there was a bug with the model endpoint monitoring store_prefixes part in the connect method (in httpdb). After fixing it, the value is indeed propagated as needed via the `client_spec` mechanism, so this code is redundant. Thanks!

mlrun/db/httpdb.py
```diff
@@ -632,7 +632,7 @@ def connect(self, secrets=None) -> typing.Self:
             )
             for prefix in ["default", "user_space", "monitoring_application"]:
                 store_prefix_value = model_monitoring_store_prefixes.get(prefix)
-                if server_prefix_value is not None:
+                if store_prefix_value is not None:
```
theSaarco, line 635: 👑

theSaarco (approved):
LGTM

PR #9244, opened 2026-01-25
[Application] Support single-file source deployment with artifacts

Yacouby:
### 📝 Description
This PR implements client-side logic for Application Runtime to automatically upload local single-file sources as MLRun artifacts during deployment.

When a user defines an Application Runtime function with a local file path (e.g., `handler.py`), the deploy process now:

- Uploads the file to the artifact store ( stored under `_mlrun/sources` internal path to avoid conflicts/overrides )
- Registers it as an MLRun artifact
- Marks it with `mlrun/system-generated` label
- Stores the artifact URI in `fn.spec.build.source` for future use by the init container

---

### 🛠️ Changes Made

- Added `_upload_source_as_artifact` to the Application class.
- Added `function_name` and `system_generated` labels to MLRunInternalLabels.
- Updated `set_function` to support `.py` files for application runtime.
- Removed the test: `test_application_from_local_file_validation` because it tested the old behavior that rejected local .py files for application runtime which is now supported.

---

### ✅ Checklist

---

### 🧪 Testing
Added unit tests for artifact upload and set function with single file sources.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11982
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


mlrun/model.py
```diff
@@ -571,13 +571,15 @@ def source(self):
     def source(self, source):
         if source and not (
             source.startswith("git://")
+            # Allow store artifact URIs for single-file sources
+            or mlrun.datastore.is_store_uri(source)
```
yaelgen, line 575: Questions:
1. Will other runtimes also be effected by this?
2. If so, what happens if a user sets source="store://..." for a non-application runtime?

mlrun/projects/project.py
```diff
@@ -5934,7 +5934,17 @@ def _init_function_from_dict(
...
+        # For application runtime we set the source path directly, deploy will upload it as an artifact
+        if kind == mlrun.runtimes.RuntimeKinds.application:
+            func = new_function(
+                name,
+                image=image,
+                kind=kind,
+                handler=handler,
+                tag=tag,
+            )
+            func.spec.build.source = url
```
yaelgen, line 5946: What happens if the file path is invalid or doesn't exist at this point?
Should we validate the file exists here to fail fast, or defer to deploy()?

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -490,6 +493,8 @@ def deploy(
 
         :return: The default API gateway URL if created or True if the function is ready (deployed)
         """
+        # Upload local single-file source as artifact (if applicable)
+        self._upload_source_as_artifact()
```
yaelgen, line 497: What happens if the upload fails? Does deploy() continue or abort? 
Maybe worth adding a warning or log to the user in such cases

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -1050,3 +1055,63 @@ def _validate_set_probes_input(params: dict):
...
+            source=source,
+            artifact_key=artifact_key,
+            project=project_name,
+        )
+
+        # Upload the file as an artifact to an internal path with system-generated label
+        artifact = project.log_artifact(
+            item=artifact_key,
+            local_path=source,
+            artifact_path="_mlrun/sources",
```
yaelgen: should be a constant and also documented

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -1050,3 +1055,63 @@ def _validate_set_probes_input(params: dict):
...
+        self.spec.build.source = artifact.uri
+
+    @staticmethod
+    def _is_local_single_file(source: str) -> bool:
+        # Skip if the source is already a store URI
+        if mlrun.datastore.is_store_uri(source):
+            return False
+
+        # Skip if it's a remote URL (not a relative/local path)
+        if not is_relative_path(source) and not os.path.isabs(source):
```
yaelgen:
```suggestion
        if not (is_relative_path(source) or os.path.isabs(source)):
```
More readable imo

yaelgen (changes requested):
Looks good! added a few suggestions and questions

mlrun/runtimes/nuclio/application/application.py
TomerShor: Is this overriding the existing artifact path or appending to it?

mlrun/runtimes/nuclio/application/application.py
Yacouby: added try except with clear error message

mlrun/projects/project.py
Yacouby: `get_item_absolute_path` is called before we reach the logic here, which raises `FileNotFoundError` in case the file doesnt exisit

mlrun/model.py
Yacouby: This change only updates validation to allow store:// URIs. 
Other runtimes are not affected.. setting a store URI on a non-Application runtime will not trigger uploads or downloads and retains existing behavior.

mlrun/common/constants.py
```diff
@@ -27,6 +27,9 @@
 NUCLIO_LABEL_PREFIX = "nuclio.io/"
 RESERVED_TAG_NAME_LATEST = "latest"
 
+# Internal path for application runtime source artifacts (avoids user artifact conflicts)
+MLRUN_INTERNAL_ARTIFACT_PATH = "_mlrun/sources"
```
rokatyy: maybe `.mlrun/sources` to align with common convention of utilitary/hidden files?

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -1050,3 +1055,68 @@ def _validate_set_probes_input(params: dict):
...
+        logger.info(
+            "Uploading local source file as artifact",
+            source=source,
+            artifact_key=artifact_key,
+            project=project_name,
+        )
+
+        # Upload the file as an artifact to an internal path with system-generated label
+        try:
+            artifact = project.log_artifact(
```
rokatyy, line 1093: what happens if this artifact already exists? 

Also, this `mlrun.common.constants.MLRUN_INTERNAL_ARTIFACT_PATH` folder under each project specifically, right? What happens if we have the same application runtime name in 2 different projects, may it lead to a conflict of two the same artifact names or we are safe?

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -1050,3 +1055,68 @@ def _validate_set_probes_input(params: dict):
...
+        except Exception as exc:
+            raise mlrun.errors.MLRunRuntimeError(
+                f"Failed to upload source file '{source}' as artifact"
+            ) from exc
+
+        # Update the source to point to the artifact URI
+        self.spec.build.source = artifact.uri
+
+    @staticmethod
+    def _is_local_single_file(source: str) -> bool:
```
rokatyy: minor
```suggestion
    def _is_single_local_file(source: str) -> bool:
```
sounds better

rokatyy (commented):
Looks good!

Some comments and questions

mlrun/runtimes/nuclio/application/application.py
Yacouby: Nice catch ! updated artifact path to append to it

mlrun/common/constants.py
Yacouby: Using `_mlrun/sources` is intentional to indicate internal system-managed artifacts; making it a hidden path is not necessary since it’s already under an internal namespace and not meant to be accessed directly by users

mlrun/runtimes/nuclio/application/application.py
Yacouby: > what happens if this artifact already exists?

A: Each deploy creates a new artifact version with the same key

 
> Also, this `mlrun.common.constants.MLRUN_INTERNAL_ARTIFACT_PATH` folder under each project specifically, right? What happens if we have the same application runtime name in 2 different projects, may it lead to a conflict of two the same artifact names or we are safe?

A: The path is under each project's artifact path, so two functions with the same name in two different projects will have separate artifacts with no conflicts.

rokatyy: @Yacouby but artifact_key will be the same, won't it create a conflict?

mlrun/common/constants.py
rokatyy: > indicate internal system-managed artifacts

this is exactly the idea of `.mlrun` to show that this is mlrun internal dir (the same as .cursor or .github). If we don't have this `_<dir>` pattern anywhere, I think `.mlrun` is more conventional

mlrun/runtimes/nuclio/application/application.py
Yacouby: @rokatyy no, artifacts are project scoped so even if the key is the same, artifacts in different projects are completely separate.

mlrun/common/constants.py
Yacouby: Ok I take that 👍 , updated

mlrun/common/constants.py
```diff
@@ -27,6 +27,9 @@
 NUCLIO_LABEL_PREFIX = "nuclio.io/"
 RESERVED_TAG_NAME_LATEST = "latest"
 
+# Internal path for application runtime source artifacts (avoids user artifact conflicts)
+MLRUN_INTERNAL_ARTIFACT_PATH = "+/.mlrun/sources"
```
rokatyy, line 32: why `+` ?

Yacouby: to append to the existing artifact path when logging the artifact.. and not override it

yaelgen: @rokatyy for `extend_artifact_path` to use

rokatyy: @Yacouby didnt know that, add a comment for this please

yaelgen (approved):

rokatyy (approved):
LGTM

TomerShor (approved):

PR #9248, opened 2026-01-27
[Serving] Implement support for stream responses

gtopper:
Adds support for streaming responses in serving graphs, enabling real-time chunk-by-chunk HTTP responses (e.g., for LLM token streaming).

Key changes:
* New `set_streaming(enabled=True)` API on serving functions
* Async streaming handler that yields results as they're produced by graph steps
* Graph steps can now use generators to stream multiple chunks
* Updated nuclio handler for generator return type support

Using the new functionality introduced in storey 1.11.8 / mlrun/storey#605.

[ML-11876](https://iguazio.atlassian.net/browse/ML-11876)

Depends on https://github.com/nuclio/nuclio-jupyter/pull/197.

[ML-11876]: https://iguazio.atlassian.net/browse/ML-11876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/runtimes/nuclio/serving.py
```diff
@@ -889,6 +950,13 @@ def to_job(
                 f"Cannot convert function '{self.metadata.name}' to a job because it has child functions"
             )
 
+        if self.spec.streaming:
```
assaf758, line 953: Losing to_job for streaming (or having to continuously toggle set_streaming) is somewhat of a drawback - do we must go this route?
What if we ignore streaming in on_job deployment (flagging the model provider to collect all the response)

assaf758 (approved):
Looks good!
Please review with David or Roy as well.

mlrun/runtimes/nuclio/serving.py
gtopper: As discussed offline – I added [this optional story](https://iguazio.atlassian.net/browse/ML-12063) to the requirement.

TL;DR: this functionality is [excluded from the spec](https://iguazio.atlassian.net/wiki/spaces/ARC/pages/467763201/Streaming+support+in+serving+graphs#1.3.-Out-of-scope), but we may want to consider it going forward.

mlrun/serving/states.py
```diff
@@ -4023,7 +4023,7 @@ def _init_async_objects(context, steps, root):
                 and step.responder
             ):
                 # if responder step (return result), add Complete()
-                step.async_object.to(storey.Complete(full_event=True))
+                step.async_object.to(storey.Complete())
```
davesh0812: Are you sure this change don't any side effects?

davesh0812 (commented):
Looks good. As we discussed, let’s also add a system test for the error case.

gtopper:
> Looks good. As we discussed, let’s also add a system test for the error case.

I've added unit tests for this case, but the error case is currently blocked by [NUC-723](https://iguazio.atlassian.net/browse/NUC-723), which turns out to cover user-raised errors as well.

[NUC-723]: https://iguazio.atlassian.net/browse/NUC-723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

gtopper:
Created [ML-12066](https://iguazio.atlassian.net/browse/ML-12066) to track the error flow and Nuclio dependency.

[ML-12066]: https://iguazio.atlassian.net/browse/ML-12066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

davesh0812 (approved):

PR #9267, opened 2026-02-02
[Application] Add init container config for source loading

Yacouby:
### 📝 Description
This PR implements server-side orchestration for init container injection in `ApplicationRuntime` when source code needs to be loaded at pod startup.

The init container uses the mlrun CLI command `mlrun load-source` to fetch source code and extract it to the shared volume before the sidecar container starts.

---

### 🛠️ Changes Made
- Added `set_source_target` method to ApplicationRuntime for configuring custom target directory.
- Added init container configuration logic in server-side function compilation.

---

### ✅ Checklist

---

### 🧪 Testing
- Added unit tests for init container configuration logic.
- Added unit test for `set_source_target` method.
- Added system test to validate the end-to-end init container flow. 
The test deploys an application with single-file source, updates the source code, redeploys, and verifies the init container correctly loads the updated source into the sidecar.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11954 + https://iguazio.atlassian.net/browse/ML-11951
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [X] No

---


mlrun/runtimes/nuclio/application/application.py
```diff
@@ -887,8 +903,22 @@ def _build_application_image(
                 "Loading on build will be forced regardless of whether 'pull_at_runtime=True' was configured."
             )
 
+        # For store:// artifact sources, exclude from build context since the init container will load
+        # the source at runtime
+        source_for_init_container = None
+        if self.spec.build.source and mlrun.datastore.is_store_uri(
+            self.spec.build.source
+        ):
+            source_for_init_container = self.spec.build.source
```
rokatyy, line 915: add a comment explaining why this magic with cleaning up  `self.spec.build.source` is needed

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -297,6 +297,22 @@ def url(self, url):
...
+
+        :param target_dir: Absolute path inside the runtime container where the source code will be placed
+        """
+        if not target_dir:
+            raise mlrun.errors.MLRunInvalidArgumentError("target_dir is required")
+
+        if not target_dir.startswith("/"):
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"target_dir must be an absolute path, got: {target_dir}"
+            )
```
liranbg, line 312: why not ensure it starts with `/` and not fail on it?

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -735,3 +752,174 @@ def _add_secrets_config_to_function_spec(
...
+    )
+
+    return {
+        "name": mlrun.common.constants.SOURCE_LOADER_INIT_CONTAINER_NAME,
+        "image": init_container_image,
+        "command": ["mlrun", "load-source"],
+        "args": [source, "--project", project, "--target", target_dir],
+        "env": [
+            {"name": "MLRUN_PROJECT", "value": project},
+            {"name": "MLRUN_DBPATH", "value": mlrun.mlconf.httpdb.api_url},
```
rokatyy, line 888: don't we have constants for those?

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -887,8 +903,22 @@ def _build_application_image(
...
+        source_for_init_container = None
+        if self.spec.build.source and mlrun.datastore.is_store_uri(
+            self.spec.build.source
+        ):
+            source_for_init_container = self.spec.build.source
+            self.spec.build.source = None
+            logger.info(
+                "Source is a store:// artifact URI - excluding from build, "
+                "init container will load it at runtime",
+                source=source_for_init_container,
```
liranbg: it should be a debug log, if at all

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -899,6 +929,12 @@ def _build_application_image(
             with_mlrun=with_mlrun,
         )
 
+        # Restore source for init container configuration by the server
+        if source_for_init_container:
+            self.spec.build.source = source_for_init_container
```
liranbg: if build has failed, do you want to restore it? if so, use `try` / `finally`

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -735,3 +752,174 @@ def _add_secrets_config_to_function_spec(
             f"Unexpected function kind {function.kind}. Expected one of: "
             f"{mlrun.runtimes.RuntimeKinds.nuclio_runtimes()}"
         )
+
+
+def _init_container_needed(
```
rokatyy: name is too generic. Nuclio supports init containers as a generic feature. This is an init container for loading source, which should be reflected in a name as well

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -270,6 +270,23 @@ def _compile_function_config(
...
             )
 
+    # Configure init container for Application runtime when source needs runtime loading
+    if function.kind == mlrun.runtimes.RuntimeKinds.application:
+        if not sidecars:
+            logger.warning(
+                "No sidecar found for Application runtime, skipping init container configuration",
+                project=function.metadata.project,
+                function=function.metadata.name,
+            )
```
liranbg: what would be teh action item the user needs to handle in order to make this warning go away? or, in other words, why is that a warning at all? what's the impact?/why it is not an error then?

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -735,3 +752,174 @@ def _add_secrets_config_to_function_spec(
...
+        return True
+
+    is_git_source = source.startswith("git://")
+    is_archive_source = source.endswith(".tar.gz") or source.endswith(".zip")
+    pull_at_runtime = function.spec.build.load_source_on_run
+
+    return (is_git_source or is_archive_source) and pull_at_runtime
+
+
+def _configure_init_container(
```
rokatyy: same here, and could you please add more comments to this function?

server/py/services/api/crud/runtimes/nuclio/function.py
liranbg: I think the question should be "_should_fetch_source_code"

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -735,3 +752,174 @@ def _add_secrets_config_to_function_spec(
...
+        target_dir=target_dir,
+    )
+
+    logger.info(
+        "Configured init container for source loading",
+        project=function.metadata.project,
+        function=function.metadata.name,
+        source=source,
+        target_dir=target_dir,
+    )
```
liranbg: this is api log level. debug is enough

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -735,3 +752,174 @@ def _add_secrets_config_to_function_spec(
...
+        mlrun.mlconf.default_base_image,
+        client_version=client_version,
+        client_python_version=client_python_version,
+    )
+
+    return {
+        "name": mlrun.common.constants.SOURCE_LOADER_INIT_CONTAINER_NAME,
+        "image": init_container_image,
+        "command": ["mlrun", "load-source"],
+        "args": [source, "--project", project, "--target", target_dir],
```
liranbg, line 885: escape those argument to avoid shell injection

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -735,3 +752,174 @@ def _add_secrets_config_to_function_spec(
...
+    :param target_dir: Target directory for source code
+    """
+    # Add volume mount idempotently
+    sidecar_mounts = sidecar.setdefault("volumeMounts", [])
+    if not any(vm.get("name") == volume_name for vm in sidecar_mounts):
+        sidecar_mounts.append(volume_mount)
+
+    sidecar["workingDir"] = target_dir
+
+    # Add PYTHONPATH env var idempotently
```
liranbg: worth explaining why we need to set the pythonpath at all

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -735,3 +752,174 @@ def _add_secrets_config_to_function_spec(
...
+    function: mlrun.runtimes.nuclio.function.RemoteRuntime,
+    init_container: dict,
+):
+    """
+    Add init container to spec idempotently (replace if exists).
+
+    :param function: The function object to configure
+    :param init_container: Init container specification
+    """
+    init_container_name = init_container["name"]
```
rokatyy: check that name is not empty

liranbg (commented):
Neat work. some comments on logging / comments

rokatyy (commented):
Looks good!

Some questions and nitpickings below

mlrun/runtimes/nuclio/application/application.py
Yacouby: mm silently fixing user input could mask configuration mistakes. 
An explicit error ensures users are aware of the expected format and intentionally provide absolute paths..

server/py/services/api/crud/runtimes/nuclio/function.py
Yacouby: These are Kubernetes container arguments passed directly to the entrypoint, not through a shell. The values come from trusted sources (function spec, project name), so shell injection is not a concern here

server/py/services/api/crud/runtimes/nuclio/function.py
Yacouby: but it can never be empty because we define it in `_build_source_loader_init_container`

server/py/services/api/crud/runtimes/nuclio/function.py
Yacouby: no, across the code `MLRUN_DBPATH` is used

server/py/services/api/crud/runtimes/nuclio/function.py
rokatyy: @Yacouby it can be reused in the future by any other functions as it looks very generic

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -270,6 +270,21 @@ def _compile_function_config(
                 )
             )
 
+    # Configure init container for Application runtime when source needs runtime loading
+    if function.kind == mlrun.runtimes.RuntimeKinds.application:
+        if not sidecars:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"No sidecar found for Application runtime '{function.metadata.name}'. "
+            )
+        elif _should_fetch_source_code(function):
```
rokatyy: no need for elif since we anyway would have raised the exception in the condition above

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -735,3 +750,194 @@ def _add_secrets_config_to_function_spec(
...
+    sidecar_mounts = sidecar.setdefault("volumeMounts", [])
+    if not any(vm.get("name") == volume_name for vm in sidecar_mounts):
+        sidecar_mounts.append(volume_mount)
+
+    sidecar["workingDir"] = target_dir
+
+    # Set PYTHONPATH to target directory so the sidecar can import modules from the source
+    # code loaded by the init container
+    sidecar_env = sidecar.setdefault("env", [])
+    if not any(e.get("name") == "PYTHONPATH" for e in sidecar_env):
```
TomerShor: what if the `PYTHONPATH` env exists but it is different than `target_dir`?

Yacouby: handled it + unit test coverage

TomerShor (approved):
LGTM 👍

rokatyy (approved):

liranbg (approved):

PR #5309, opened 2024-03-20
[Model Monitoring] Remove monitoring batch job

jond01:
Resolves [ML-5752](https://iguazio.atlassian.net/browse/ML-5752) and [ML-5972](https://iguazio.atlassian.net/browse/ML-5972).
Partly address [ML-5819](https://iguazio.atlassian.net/browse/ML-5819) (`TrackingPolicy` removal).

TODO: 
- [x] Update system tests.
  - All the tests in `tests/system/model_monitoring/test_app.py` pass.
  - All the tests in `tests/system/model_monitoring/test_model_monitoring.py` pass, except for:
    - `TestModelMonitoringKafka` - pass except for the last two assertions.
    - `TestVotingModelMonitoring` and `TestModelMonitoringRegression` are currently skipped.
- [x] Add deprecations to `record_results`.

BC test intentionally fails due to the removal of the following API endpoint:
- POST: `/api/v1/projects/{project}/jobs/batch-monitoring`

Breaking changes (apart from the direct batch job removal):
* The stream pod will not be deployed via `serving_fn.set_tracking()`. Call `project.enable_model_monitoring()` explicitly.
* Remove `TrackingPolicy` class.
* Do not update the following TSDB table in batch inference flow: `users/pipelines/{project_name}/model-endpoints/events`. The relevant results are stored in the app-results TSDB.
https://github.com/mlrun/mlrun/blob/7737c58a5e179e047d4cfba4318616bcc9e40ed7/mlrun/model_monitoring/batch.py#L809-L832

[ML-5752]: https://iguazio.atlassian.net/browse/ML-5752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ML-5972]: https://iguazio.atlassian.net/browse/ML-5972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ML-5819]: https://iguazio.atlassian.net/browse/ML-5819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9272, opened 2026-02-03
[Model Monitoring] Add writer lag detection events and graph step

alxtkr77:
## Summary
- Add `MODEL_MONITORING_INFRA` entity kind and `MODEL_MONITORING_LAG_DETECTED` event kind to the alert schema
- Add `lag_detection` config defaults under `model_endpoint_monitoring` (threshold: 60min, cooldown: 30min, min: 5min)
- Add `WriterLagEventsGenerator` storey step that detects when writer events are stale (lag exceeds configurable threshold) and emits alert events with per-worker cooldown via `time.monotonic()`
- Override nuclio handler at module level to capture `worker_id` for entity identification (`{project}.writer.{worker_id}`)
- Wire `lag_events_generator` into the writer graph, converging with `alert_generator` on the shared `filter_none` → `alert_generator_api_call` path
- Route all event kinds (METRIC, RESULT, STATS) through the lag detection step
- Add 20 unit tests covering detection logic, cooldown behavior, routing, and graph topology

## Notes
- This PR covers ML-11674 (Phase 1 of ML-9954). The lag params are accepted by `WriterGraphFactory` but not yet wired through the SDK/API chain — that comes in ML-12079 (Phase 2).
- When `lag_threshold_seconds=0` (default), the step returns `None` for every event — effectively a no-op until configured.

## Test plan
- [x] `pytest tests/model_monitoring/test_lag_detection.py -v` — 20/20 passing
- [x] `pytest tests/model_monitoring/test_writer.py tests/model_monitoring/test_writer_graph.py -v` — no regressions
- [x] `ruff check` and `ruff format` clean on all changed files
- [x] Verify writer graph topology includes `lag_events_generator` step in V3IO and TimescaleDB variants
- [ ] System test: enable model monitoring with lag params, send stale event, verify alert fires (requires live deployment)

Reference: [ML-11674](https://iguazio.atlassian.net/browse/ML-11674)

[ML-11674]: https://iguazio.atlassian.net/browse/ML-11674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/model_monitoring/writer.py
+++ b/mlrun/model_monitoring/writer.py
@@ -46,25 +46,6 @@ from mlrun.model_monitoring.helpers import get_result_instance_fqn
 from mlrun.serving.utils import StepToDict
 from mlrun.utils import logger, now_date
 
-# Module-level worker ID captured by the nuclio handler override below.
-# When running inside a nuclio worker, ``handler()`` populates this before
-# each request so that ``WriterLagEventsGenerator`` can include the worker
-# identity in the lag event entity.
-NUCLIO_WORKER_ID: int | None = None
-
-
-def init_context(context):
-    from mlrun.runtimes import nuclio_init_hook
-
-    nuclio_init_hook(context, globals(), "serving_v2")
-
-
-def handler(context, event):
-    global NUCLIO_WORKER_ID
-    NUCLIO_WORKER_ID = getattr(context, "worker_id", None)
-    return context.mlrun_handler(context, event)
-
-
 _RawEvent = dict[str, Any]
 _AppResultEvent = NewType("_AppResultEvent", _RawEvent)
 
@@ -525,7 +506,7 @@ class WriterLagEventsGenerator(storey.MapClass):
         if lag_seconds < self.lag_threshold_seconds:
             return None
 
-        worker_id = NUCLIO_WORKER_ID if NUCLIO_WORKER_ID is not None else 0
+        worker_id = getattr(self.context, "worker_id", 0)
         now_mono = time.monotonic()
         last_emit = self._last_emit_ts.get(worker_id, -float("inf"))
         if (now_mono - last_emit) < self.lag_event_cooldown_seconds:
--- a/mlrun/model_monitoring/writer.py
+++ b/mlrun/model_monitoring/writer.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import json
+import time
 import typing
 from collections.abc import Callable
 from datetime import UTC, datetime
@@ -44,6 +45,25 @@ from mlrun.model_monitoring.db._stats import (
 from mlrun.model_monitoring.helpers import get_result_instance_fqn
 from mlrun.serving.utils import StepToDict
 from mlrun.utils import logger
+
+# Module-level worker ID captured by the nuclio handler override below.
+# When running inside a nuclio worker, ``handler()`` populates this before
+# each request so that ``WriterLagEventsGenerator`` can include the worker
+# identity in the lag event entity.
+NUCLIO_WORKER_ID: int | None = None
+
+
+def init_context(context):
+    from mlrun.runtimes import nuclio_init_hook
+
+    nuclio_init_hook(context, globals(), "serving_v2")
+
+
+def handler(context, event):
+    global NUCLIO_WORKER_ID
+    NUCLIO_WORKER_ID = getattr(context, "worker_id", None)
+    return context.mlrun_handler(context, event)
+
 
 _RawEvent = dict[str, Any]
 _AppResultEvent = NewType("_AppResultEvent", _RawEvent)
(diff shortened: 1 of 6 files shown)


mlrun/model_monitoring/writer.py
```diff
@@ -233,10 +253,83 @@ def do(self, event: _RawEvent) -> None:
...
+        end_infer_time = event.get(WriterEvent.END_INFER_TIME)
+        if end_infer_time is None:
+            return None
+
+        if isinstance(end_infer_time, str):
+            end_infer_time = datetime.fromisoformat(end_infer_time)
+
+        lag_seconds = (
+            datetime.now(tz=UTC) - end_infer_time.astimezone(UTC)
+        ).total_seconds()
```
danielperezz: Just a suggestion - use `mlrun.utils.now_date` instead

mlrun/model_monitoring/writer.py
```diff
@@ -233,10 +253,83 @@ def do(self, event: _RawEvent) -> None:
...
+        entity_id = f"{self.project}.writer.{worker_id}"
+
+        event_data = alert_objects.Event(
+            kind=alert_objects.EventKind.MODEL_MONITORING_LAG_DETECTED,
+            entity=alert_objects.EventEntities(
+                kind=alert_objects.EventEntityKind.MODEL_MONITORING_INFRA,
+                project=self.project,
+                ids=[entity_id],
+            ),
+            value_dict={
```
danielperezz: Consider adding:
`"app_name": event.get([WriterEvent.APPLICATION_NAME], "")`

mlrun/model_monitoring/writer.py
```diff
@@ -272,12 +367,29 @@ def apply_writer_graph(
...
+            lag_threshold_seconds=lag_threshold_seconds,
+            lag_event_cooldown_seconds=lag_event_cooldown_seconds,
+        )
         graph.add_step(
             "storey.Filter",
             name="filter_none",
             _fn="(event is not None)",
             after="alert_generator",
         )
+        graph["filter_none"].after_step("lag_events_generator")
```
danielperezz: Why not simply:
```
graph.add_step(
            "storey.Filter",
            name="filter_none",
            _fn="(event is not None)",
            after=["alert_generator", "writer_lag_events_generator"],
        )
```

mlrun/model_monitoring/writer.py
```diff
@@ -355,11 +467,11 @@ def select_outlets(self, event):
...
         logger.info("Selecting the outlet for the event", kind=kind)
         if kind == WriterEventKind.METRIC:
-            outlets = ["tsdb_metrics"]
+            outlets = ["tsdb_metrics", "lag_events_generator"]
         elif kind == WriterEventKind.RESULT:
-            outlets = ["tsdb_app_results", "alert_generator"]
+            outlets = ["tsdb_app_results", "alert_generator", "lag_events_generator"]
         elif kind == WriterEventKind.STATS:
-            outlets = ["stats_writer"]
+            outlets = ["stats_writer", "lag_events_generator"]
```
danielperezz: I don't think it's needed for STATS

mlrun/model_monitoring/writer.py
```diff
@@ -233,10 +253,83 @@ def do(self, event: _RawEvent) -> None:
         logger.info("Model monitoring writer finished handling event")
 
 
+class WriterLagEventsGenerator(storey.MapClass):
```
danielperezz: Consider moving this class to the end of this file to be consistent with the other steps that are implemented after WriterGraphFactory

mlrun/model_monitoring/writer.py
```diff
@@ -272,12 +367,29 @@ def apply_writer_graph(
             after="kind_choice_step",
             project=fn.metadata.project,
         )
+        lag_threshold_seconds = (
+            self.lag_threshold_minutes * 60 if self.lag_threshold_minutes else 0
```
danielperezz: I think it might be clearer if you set this `None` to declare it's disabled

mlrun/model_monitoring/writer.py
```diff
@@ -45,6 +46,25 @@
...
+NUCLIO_WORKER_ID: int | None = None
+
+
+def init_context(context):
+    from mlrun.runtimes import nuclio_init_hook
+
+    nuclio_init_hook(context, globals(), "serving_v2")
+
+
+def handler(context, event):
```
danielperezz: I  think you also need to pass explicit handler parameter when creating the writer function as we do in `_get_model_monitoring_controller_function` so the default handler won't be appended and override this one

danielperezz (changes requested):
Nice! had a few commnets.
Also, did you include Makefile changes on purpose?

alxtkr77:
> Nice! had a few commnets. Also, did you include Makefile changes on purpose?

Yes. The --config-prop incompatible.response.enum.increased:false was added intentionally. Without it, the BC test fails because we added new enum values (MODEL_MONITORING_INFRA, MODEL_MONITORING_LAG_DETECTED) that appear in response schemas. The openapi-diff tool treats response enum additions as breaking by default. This config tells it to allow that - which is correct since our StrEnums deserialize as plain strings on the client side.

mlrun/model_monitoring/writer.py
alxtkr77: Investigated the handler flow — the custom `handler()` and `init_context()` in writer.py work as-is without an explicit parameter. Nuclio discovers them by name from the source module.                                  
                                                                                                                                                                                                                            
  The flow is:                                                                                                                                                                                                              
  1. Nuclio calls `init_context()` → runs `nuclio_init_hook(context, globals(), "serving_v2")`                                                                                                                              
  2. This sets `context.mlrun_handler` to the serving graph handler                                                                                                                                                         
  3. Each request goes through our `handler()` which captures `worker_id` and delegates to `context.mlrun_handler`                                                                                                          
                                                                                                                                                                                                                            
  The writer uses `kind=serving`, and for serving runtimes the `handler` parameter in `code_to_function` is effectively ignored:                                                                                            
  - `default_handler` is always set to `""` for non-mlrun sub-kinds ([run.py:840](https://github.com/mlrun/mlrun/blob/development/mlrun/run.py#L840))                                                                       
  - `function_handler` only gets set if the handler contains `":"` ([run.py:841-842](https://github.com/mlrun/mlrun/blob/development/mlrun/run.py#L841))                                                                    
                                                                                                                                                                                                                            
  ---                                                                                                                                                                                                                       
                                                                                                                                                                                                                            
  Interestingly, looks like `_get_model_monitoring_controller_function`'s `handler="handler"` is redundant as well! Even for `kind=nuclio`, `nuclio.build_file()` already defaults to `"handler"`:                          
                                                                                                                                                                                                                            
  ```python                                                                                                                                                                                                                 
  # run.py:798                                                                                                                                                                                                              
  nuclio.build_file(filename, handler=handler or "handler", ...)                                                                                                                                                            
                                                                                                                                                                                                                            
  If handler is None, None or "handler" gives "handler" — same result. Then at https://github.com/mlrun/mlrun/blob/development/mlrun/run.py#L841, since "handler" doesn't contain ":", it falls through to get_in(spec,     
  "spec.handler") which returns "main:handler" from nuclio.build_file. The outcome is "main:handler" regardless.                                                                                                            
                                                                                                                                                                                                                            
  It was originally introduced in https://github.com/mlrun/mlrun/pull/3598 for the batch function (kind=job) and carried over through refactors (https://github.com/mlrun/mlrun/pull/5309) without anyone noticing it's a   
  no-op.                                                                                                                                                                                                                    

  I removed it from _get_model_monitoring_controller_function and added tests to verify the behavior:                                                                                                                       
                                                                                                                                                                                                                            
  class TestWriterHandlerParameter:                                                                                                                                                                                         
      def test_serving_function_handler_same_with_and_without_param(self):                                                                                                                                                  
          # kind=serving: handler="handler" is identical to omitting it                                                                                                                                                     
      def test_nuclio_function_handler_same_with_and_without_param(self):                                                                                                                                                   
          # kind=nuclio: same — handler="handler" is a no-op                                                                                                                                                                
      def test_colon_format_overrides_function_handler(self):                                                                                                                                                               
          # only "module:func" format actually overrides function_handler                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                              
  ```                                                                                                                                                                                                                       

  ```

danielperezz: @alxtkr77 
Let me correct myself, what's missing in the writer compared to the controller, is that the controller uses `kind=mlrun.run.RuntimeKinds.nuclio` in its `code_to_function` params (not `serving`)
Then `code_to_function` is calling:
```
name, spec, code = nuclio.build_file(
        filename,
        name=name,
        handler=handler or "handler",
        kind=sub_kind,
        ignored_tags=ignored_tags,
    )
```
For `serving` kind there is a `sub_kind`, while for `nuclio` kind, the `sub_kind` resolves to `None`.
Then in `build_file`:
```
 if kind:
      code = add_kind_footer(kind, config, code)
```
And `add_kind_footer` appends the `mlrun_footer` which contains the `init_context()` and the `handler()` to the end of the function's source file in case of serving function, which will override the handler you added.

So to summarize what your'e missing is:
- `kind=mlrun.run.RuntimeKinds.nuclio`
- Add `init_context()` to the end of the writer code file  (to replace the one appended by `mlrun_footer`).
- Consider moving the handler to end of the file too, for consistency.
- `handler="handler'` is probably also required

alxtkr77: @danielperezz Right, now I understand the problem you're raising.
But looking deeper, we don't need a custom handler() at all. The serving framework already propagates worker_id through the graph context: v2_serving_init → init_states → GraphContext(nuclio_context=context) → self.worker_id = nuclio_context.worker_id (server.py:1103). So every graph step already has self.context.worker_id available. 
So we can do ```worker_id = getattr(self.context, "worker_id", 0)``` instead of ```worker_id = NUCLIO_WORKER_ID if NUCLIO_WORKER_ID is not None else 0```

What do you think?

danielperezz: @alxtkr77 
Sounds good, I guess I missed it in my design.
Do you know if `self.context` is accessible from the steps of v2 of the graph? 
I guess at least some minimal changes are required to make it accessible.

alxtkr77: @danielperezz Yes, self.context is accessible from v2 graph steps. The serving framework passes it during init_object() (states.py:851-853) — if the step class accepts context in its __init__ (via explicit param or
  **kwargs), it gets the GraphContext.

storey.MapClass inherits from storey.Flow which has context=None in its signature, so self.context is available. GraphContext copies worker_id from the nuclio context at server.py:1103 . Added a unit test to cover it

danielperezz (approved):
LGTM

PR #9278, opened 2026-02-04
[Model Monitoring] Add support for streaming MRS

gtopper:
[ML-11879](https://iguazio.atlassian.net/browse/ML-11879)

Enables model monitoring for streaming `ModelRunnerStep`s (MRS). When a serving function has streaming enabled, the MM pipeline needs to aggregate streaming chunks into a single event before processing.

**Key changes:**
* Bumps storey to get the changes in https://github.com/mlrun/storey/pull/613
- **Collector insertion** (`server.py`): When `streaming=True`, dynamically inserts a `storey.Collector` step between each MRS and the MM pipeline to aggregate streaming chunks into a single event.
- **Chunk aggregation** (`system_steps.py`): Adds aggregation logic to `MonitoringPreProcessor` that detects collected streaming chunks and merges them — concatenating string outputs, summing numeric metrics, and taking first values for other fields.
- **ProcessEndpointEvent fixes** (`stream_processing.py`): Returns event with `body=None` instead of bare `None` on validation failure (fixes downstream `AttributeError`). Removes `microsec` `is_not_none` validation since streaming latency is calculated asynchronously by `Collector`.

## Tests
* Unit tests for chunk detection and all aggregation paths
* Graph structure test verifying `Collector` insertion when streaming is enabled
* System test (`test_monitoring_with_streaming_model_runner`) verifying end-to-end: streaming model → `Collector` → MM pipeline → model endpoint creation
* `test_app_flow` passed

[ML-11879]: https://iguazio.atlassian.net/browse/ML-11879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/serving/system_steps.py
```diff
@@ -58,6 +58,144 @@ def __init__(
...
+            return False
+
+        # Only detect dict lists with streaming indicators
+        # String lists are ambiguous (could be batch inputs)
+        first = body[0]
+        if not isinstance(first, dict):
+            return False
+
+        # Check for streaming chunk indicators in the first element
+        streaming_keys = {"output", "outputs", "metrics"}
```
royischoss: I know it a hard mistake to make but also user output can look like this. (I don't know about the list) , maybe adding it to the metadata a flag saying we are streaming result.

royischoss (commented):
Hey looks good, small comment on the way we check if an event is from streaming type.

tests/system/model_monitoring/test_model_monitoring.py
```diff
@@ -1161,6 +1161,86 @@ def test_monitoring_with_model_runner_batch_infer(self, with_training_set: bool)
...
+            name=endpoint_name,
+            project=self.project_name,
+            function_name=function_name,
+            function_tag="latest",
+            tsdb_metrics=True,
+        )
+        assert (
+            endpoint_with_metrics.status.last_request is not None
+        ), "Expected last_request to be set after streaming invoke"
+
```
tomerm-iguazio, line 1304: please add tsdb check + predications table (you can view `test_monitoring_with_model_runner_batch_infer` as an example)

mlrun/serving/system_steps.py
gtopper: Opened https://github.com/mlrun/storey/pull/614 to provide this functionality in the Collector.

tests/serving/test_tracking.py
tomerm-iguazio, line 1: It’s great to see several unit component tests. I would suggest adding an end-to-end test that includes single and multiple models and validates error scenarios using pytest parametrized tests.

tomerm-iguazio (changes requested):
Looks good, add 2 tests comments

mlrun/serving/system_steps.py
gtopper: Released in storey v1.11.15.

gtopper: Pushed change to use the new marker instead of this heuristic.

tests/serving/test_tracking.py
gtopper: For error scenarios, we have another story – [ML-12066](https://iguazio.atlassian.net/browse/ML-12066). For single + multiple models, I'll add a test, but keep in mind that this is only supported when the multiple models are not streaming models. I.e. what's supported is selecting (per event) one streaming models OR one or more non-streaming models.

gtopper: Also, I'm assuming by end-to-end, you mean a system test?

tests/system/model_monitoring/test_model_monitoring.py
gtopper: Done.

mlrun/serving/system_steps.py
```diff
@@ -406,14 +457,29 @@ def do(self, event):
...
+            event.body = self._aggregate_collected_chunks(
+                event.body if isinstance(event.body, list) else [event.body]
+            )
+
+        # When streaming chunks were collected from a multi-model MRS, only
+        # one model actually ran (streaming requires a single selected
+        # runnable). The body is the raw aggregated output, not keyed by
+        # model name, so narrow monitoring_data to the single selected model
+        # so the single-model path below is used.
+        if is_stream_collected and len(monitoring_data) > 1:
```
royischoss, line 477: in this context monitoring_data is a dict for a single model. so this check may pass but I don't think this is what you check here

royischoss (commented):
Hey small comment

royischoss (approved):

tests/serving/test_tracking.py
gtopper: Done.

tomerm-iguazio (approved):
Looks good:)

PR #9279, opened 2026-02-04
[Application] Support skipping image build for source changes

Yacouby:
### 📝 Description
This PR skips sidecar image rebuild on redeploy when only source code changes for Application Runtime.
When using single-file sources, the init container loads the source at runtime. This means code-only changes don't require rebuilding the sidecar image.

---

### 🛠️ Changes Made

- Add `requires_build` override that returns False for store:// URIs (init container handles source loading)
- Update `prepare_image_for_deploy` to only force build for non-store URIs (git/archives not yet supported by init container)
- Update `force_build` parameter docstring to clarify when it should be used

---

### ✅ Checklist

---

### 🧪 Testing

- Added parametrized unit test validating `requires_build`
- Updated system test to validate that sidecar image was not built on second deploy

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11953
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


mlrun/runtimes/nuclio/application/application.py
```diff
@@ -466,14 +466,39 @@ def pre_deploy_validation(self):
...
-        if self.spec.build.source and self.spec.build.load_source_on_run:
-            logger.warning(
-                "Application runtime requires loading the source into the application image. "
-                f"Even though {self.spec.build.load_source_on_run=}, loading on build will be forced."
-            )
-            self.spec.build.load_source_on_run = False
+        source = self.spec.build.source
+        if source and self.spec.build.load_source_on_run:
+            # store:// URIs are handled by init container at runtime, no need to force build
+            if not mlrun.datastore.is_store_uri(source):
```
TomerShor, line 472: you can create a private method `requires_load_on_build` that will later be extended

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -466,14 +466,39 @@ def pre_deploy_validation(self):
...
+        source = build.source
+
+        # store:// URIs are loaded by init container at runtime, not baked into image
+        if source and mlrun.datastore.is_store_uri(source):
+            source_requires_build = False
+        else:
+            # For other sources (git, archives), check load_source_on_run flag
+            source_requires_build = bool(source and not build.load_source_on_run)
+
+        return bool(build.commands or build.requirements or source_requires_build)
```
TomerShor, line 500: This should return `True` on first deploy, right?
How does it determine it on this case?

TomerShor (changes requested):
This only handles the build of the sidecar container by letting MLRun know it doesn't need to build it again. 
But the Nuclio container image (the reverse proxy) will still be rebuilt every time - this is another build we want to skip, and we need to configure it properly to tell Nuclio not to build the image.

mlrun/runtimes/nuclio/application/application.py
Yacouby: Not always - it depends on the function spec, not whether it's first or second deploy.
`requires_build` returns True if the spec has commands, requirements, or a source that isn't a store:// URI. 
For store:// source alone (no commands/requirements), it returns False even on first deploy.
It determines this purely from the spec content, not deployment history.

mlrun/runtimes/nuclio/application/application.py
Yacouby: This entire method (including the check) will be removed in the next PR when git/archive init container support is added because all source types will be handled by the init container at runtime - no workaround needed. So the method would only call super() with no additional logic

mlrun/runtimes/nuclio/application/application.py
TomerShor: I see, so in case we can commands or requirements we always build, even if the don't change between deployments?

Yacouby: This only checks if the spec needs building in general, but the actual build condition ( https://github.com/mlrun/mlrun/pull/9279/changes#diff-fd22288190456c7d8d3a35f0031b11a903575abcee9e74dceb0ed5a6ff8af639R546 ) also checks if an image already exists. So while this returns True when commands/requirements exist, the build only happens on first deploy or when force_build=True is passed.

TomerShor: Thanks for clarifying 🙏

TomerShor (approved):

PR #9294, opened 2026-02-09
[Package] Added `LogHint` model to pass in a function's `returns`

guy1992l:
### 📝 Description

Made a convenient class to use instead of `dict` for passsing complex returning log hints to MLRun functions. Added functionality and tests.

---

### 🛠️ Changes Made

- [x] Created `LogHint` to pass extra configurations like extra data, tag, labels, itemization and `kwargs` for the packager.
- [x] Deprecated `dict` as log hint (now warns user who uses it).
- [x] Added an option to pass `packing_kwargs` as part of a log hint string as well.
- [x] Changed doc strings of all `returns` argument accordingly. 
- [x] Made linking artifacts bidirectional (from packagers to context and from context to packagers). 
- [x] Introduced log hint and nested inputs into `mlrun_pipelines`.

---

### ✅ Checklist
- [X] If I introduced a deprecation:
  - [X] I followed the [Deprecation Guidelines](./DEPRECATION.md)
  - [X] I updated the relevant Jira ticket for documentation

---

### 🧪 Testing
  
  
Wrote unit tests.
---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11894?atlOrigin=eyJpIjoiNjBmODNkMTIxMDdhNDUzYTlmNDk3OWNlN2Q4YTI1NjIiLCJwIjoiaiJ9
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [X] No

---


mlrun/package/log_hint.py
```diff
@@ -0,0 +1,347 @@
...
+    """
+    The artifact key to log the object under.
+    """
+
+    tag: str = ""
+    """
+    The artifact tag to log the object under. Default is an empty string.
+    """
+
+    # TODO: Restore type to `bool | int` once migrated to Pydantic v2, which handles Union[bool, int]. Remove the
```
yaelgen, line 40: So itemized is typed as Any due to Pydantic v1 union limitations (bool | int).
Maybe worth clarifying this in the field docstring (not just the TODO), since it affects validation expectations for users.

mlrun/package/log_hint.py
```diff
@@ -0,0 +1,347 @@
...
+        # TODO: Remove in 1.13.0 - this method should only support parsing from the new LogHint format.
+        # Check for the old dict format and raise a deprecation warning:
+        if isinstance(obj, dict):
+            # Detect new LogHint-format dict (from .dict() serialization):
+            if set(obj.keys()).issubset(
+                set(cls.__fields__.keys())
+            ) and "*" not in obj.get("key", ""):
+                return super().parse_obj(obj=obj)
+            # Potentially old format, try to parse and raise a warning:
+            key = obj.pop("key")
```
yaelgen, line 127: I know the support will be deprecated in 1.13, but this mutates legacy dicts internally and will change the user dict. I see you are passing `.copy()` when you are passing the log_hint (log_hint.copy()) in some places, 
Maybe safer to copy inside parse_obj when obj is a dict, so callers don’t need to remember doing it?

mlrun/__init__.py
```diff
@@ -17,7 +17,6 @@
     "set_environment",
     "code_to_function",
     "import_function",
-    "handler",
```
yaelgen, line 20: `from mlrun import handler` used to work but now it won't, are we ok with that?

mlrun/package/log_hint.py
```diff
@@ -0,0 +1,347 @@
...
+    """
+    Metrics to log alongside the model artifact (only for model artifacts). To link to another package, write the key
+    and a '...' as the value. For more information, see the 'Linking artifacts' section at the ``Packager`` or
+    ``DefaultPackager`` documentation.
+    """
+
+    @validator("itemized")
+    def _validate_itemized(cls, v):  # noqa: N805
+        if not isinstance(v, bool | int):
+            raise ValueError(f"'itemized' must be bool or int, got {type(v).__name__}")
```
yaelgen, line 95: This raises `ValueError` for invalid itemized, while the rest of the package uses `MLRunInvalidArgumentError` for invalid arguments, is that intentional?

mlrun/package/__init__.py
```diff
@@ -29,20 +28,20 @@
...
     ArtifactType,
-    LogHintKey,
     StructFileSupportedFormat,
 )
 
 
 def handler(
     labels: dict[str, str] | None = None,  # TODO: Remove in MLRun 1.13.0
-    outputs: list[str | dict[str, str]] | None = None,
+    outputs: list[str | dict[str, str] | LogHint] | None = None,
```
yaelgen, line 44: So when users pass LogHint instances in the handler’s outputs (e.g. @handler(outputs=[LogHint(key="x"), "y"])), that list is passed as log_hints to ContextHandler.log_outputs, which calls LogHint.parse_obj(log_hint) for each item. So parse_obj can be called with an existing LogHint instance. If I understand correctly, right now that case falls through to super().parse_obj(obj=obj), which in Pydantic v1 expects a dict and is not meant for model instances. Maybe Add an early return for that case:
```
if isinstance(obj, LogHint):
    return obj
```
?

yaelgen (commented):
Looks good! a few questions and suggestions

mlrun/__init__.py
guy1992l: it will work, I added a deprication message with get attribute.

mlrun/package/log_hint.py
guy1992l: the dict is passed from outside the run and created again when loading from the yaml in local runtime. It won't affect user written values. The log hint copy is on purpose for the itemization.

mlrun/package/log_hint.py
guy1992l: See doc string:

```
Accepts the following types:
    * `bool`:
        - `True`: Recursively **unbundles** the object all the way down.
        - `False` (default): Logs the collection as a single, opaque artifact.
    * `int`: Specifies the maximum depth of **unbundling**. For example, `1` will itemize the top-level collection but
      log nested collections as single units.
```

mlrun/package/log_hint.py
guy1992l: I tried to mimic PyDantic value errors, it will come from there.

mlrun/package/__init__.py
guy1992l: nice catch. the validate model in v2 had it covered. will change.

mlrun/package/log_hint.py
yaelgen: I still think a defensive copy of the dict inside parse_obj for dict inputs is a safer and more consistent approach while we keep dict support, but if you’re comfortable with it I’m fine leaving it as is

mlrun/package/__init__.py
guy1992l: I already changed, but I remembered it used to work because I had a test and indeed see: https://github.com/mlrun/mlrun/pull/9294/changes#diff-567b6acccd9a70b8896e37bcf8bb94272f57b7f367a3ddb0796658c3fc6952f1R50

It actually make sense to call parse or validate on the object so in the 0.0000000001% chance someone will create a `LogHint` and edit it to break it and send it, it will be flagged. Maybe.

yaelgen (approved):

mlrun/package/log_hint.py
guy1992l: added copy :+1:

yaelgen (approved):

liranbg (approved):

PR #9285, opened 2026-02-05
[Model Monitoring] Wire lag detection params through SDK/API/CRUD

alxtkr77:
## Summary
Add `lag_threshold` and `lag_event_cooldown` parameters to the full `enable_model_monitoring()` call chain, allowing users to configure writer lag detection thresholds.

## Changes Made
- **SDK** (`project.py`): Add params with validation (`lag_threshold >= min_lag_threshold_minutes`)
- **DB Interface** (`base.py`, `httpdb.py`, `nopdb.py`, `sqldb.py`): Wire params through interface
- **API Endpoint** (`model_monitoring.py`): Add query parameters
- **CRUD** (`deployment.py`): Compute defaults from config and `base_period`, pass to `WriterGraphFactory`
- **Warning**: Emit warning when `base_period` is updated via `update_model_monitoring_controller()`
- **Tests**: Add 5 unit tests for validation and parameter forwarding

## Testing
- All 23 lag detection tests pass
- Validated SDK rejects `lag_threshold < 5` minutes
- Validated HTTPDB conditionally includes params (backward compatible)

## Reference
- Jira: ML-12079
- Parent feature: ML-9954

## Breaking Changes
None - new optional parameters with server-computed defaults

PR #9315, opened 2026-02-15
[Model Monitoring] Add SDK methods and wildcard cache for lag detection alerts

alxtkr77:
## Summary
- Add `set_model_monitoring_lag_alert` and `delete_model_monitoring_lag_alert` SDK methods to `MlrunProject`
- Add wildcard entity (`ids=["*"]`) support to the alerts service event cache, enabling a single alert to match lag events from any writer worker
- Clean up lag alerts automatically when model monitoring is disabled

### SDK (`mlrun/projects/project.py`)
- `set_model_monitoring_lag_alert()`: creates a single wildcard alert matching lag events from all writer workers
- `delete_model_monitoring_lag_alert()`: idempotent deletion using `contextlib.suppress`
- `disable_model_monitoring()`: cleans up lag alerts on disable with error handling

### Alerts service (`server/py/services/alerts/crud/events.py`)
- Two-tier event cache: `_cache` for exact entity matching, `_wildcard_cache` for `ids=["*"]` alerts
- Both exact and wildcard alerts fire independently when an event matches
- Type-safe `tuple[...]` annotations, `Optional[str]` narrowing for project parameter

### Shared constant (`mlrun/common/schemas/model_monitoring/constants.py`)
- `MonitoringAlertNames.WRITER_LAG_DETECTED = "writer-lag-detected"`

## Testing
- 8 new SDK unit tests (set/delete alert methods, notification wrapping, field validation)
- 11 new alerts cache unit tests (add/remove/delete operations, wildcard matching, mixed exact+wildcard)
- 1 new system test for end-to-end lag alert flow
- All 42 new tests passing locally

## Reference
- Jira: ML-11675
- Depends on: #9285 (lag detection param wiring)

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -11,6 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+import contextlib
 import datetime
 import getpass
 import glob
@@ -44,6 +46,7 @@ import mlrun.common.runtimes.constants
 import mlrun.common.schemas.alert
 import mlrun.common.schemas.artifact
 import mlrun.common.schemas.model_monitoring.constants as mm_constants
+import mlrun.common.schemas.notification
 import mlrun.common.secrets
 import mlrun.datastore.datastore_profile
 import mlrun.db
@@ -2738,8 +2741,15 @@ class MlrunProject(ModelObj):
             user_application_list=user_application_list,
         )
         if succeed and delete_resources:
-            if delete_resources:
-                logger.info("Model Monitoring disabled", project=self.name)
+            try:
+                self.delete_model_monitoring_lag_alert()
+            except Exception as exc:
+                logger.warning(
+                    "Failed to delete lag detection alerts during disable",
+                    project=self.name,
+                    error=mlrun.errors.err_to_str(exc),
+                )
+            logger.info("Model Monitoring disabled", project=self.name)
             if delete_user_applications:
                 logger.info(
                     "All the desired monitoring application were deleted",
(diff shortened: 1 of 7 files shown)


mlrun/projects/project.py
```diff
@@ -2757,6 +2766,75 @@ def disable_model_monitoring(
...
+        )
+        db = mlrun.db.get_run_db(secrets=self._secrets)
+        db.store_alert_config(alert_name, alert_data, project=self.name)
+
+    def delete_model_monitoring_lag_alert(self) -> None:
+        """Delete the writer lag detection alert for this project."""
+        db = mlrun.db.get_run_db(secrets=self._secrets)
+        with contextlib.suppress(mlrun.errors.MLRunNotFoundError):
+            db.delete_alert_config(
+                mm_constants.MonitoringAlertNames.WRITER_LAG_DETECTED, self.name
```
danielperezz: add `project_name=self.name` for readability

mlrun/projects/project.py
```diff
@@ -2757,6 +2766,75 @@ def disable_model_monitoring(
...
+            name=alert_name,
+            summary=(
+                "Model monitoring writer (worker {{entity}}) "
+                "in project {{project}} is experiencing lag."
+            ),
+            severity=alert_constants.AlertSeverity.MEDIUM,
+            entities=alert_constants.EventEntities(
+                kind=alert_constants.EventEntityKind.MODEL_MONITORING_INFRA,
+                project=self.name,
+                ids=["*"],
```
danielperezz, line 2813: This means we catch all `MODEL_MONITORING_LAG_DETECTED` in the same alert (not just the writer but also other possible future events)
@theSaarco What do you think? is that a problem?

danielperezz, line 2813: Also, consider adding a comment to clarify the meaning of `id=["*"]`

server/py/services/alerts/crud/events.py
```diff
@@ -27,36 +27,52 @@
...
 
-        if len(event_data.entity.ids) > 1:
-            return False
-
-        return bool(event_data.is_valid())
+        return False if len(event_data.entity.ids) > 1 else bool(event_data.is_valid())
 
     def add_event_configuration(self, project, event_kind, alert_id, entity_id):
-        self._cache.setdefault((project, event_kind, entity_id), set()).add(alert_id)
+        if entity_id == "*":
```
danielperezz, line 50: consider adding both here and in `alert.py` a comment/docstring explaining the behaviour of `"*"`

danielperezz (changes requested):
Looks great, a few small comments :)

mlrun/projects/project.py
alxtkr77: Good question. I looked into this and here's what I found:

Per the HLD (titled "Generate an Alert When Lags Are Detected in Writer/Applications"), the event kind `MODEL_MONITORING_LAG_DETECTED` was intentionally designed to be generic -- not writer-specific. The idea is that if applications also start lagging in the future, the same event kind would be reused.

So the wildcard `ids=["*"]` catching future non-writer lag events would actually be by design, not a bug. That said, here are the options I considered:

1. **Keep `*` as-is (recommended)** -- The HLD intent is that the lag alert covers all monitoring infra, not just the writer. If apps also lag, the user probably wants to know. I could rename the alert from `writer-lag-detected` to `monitoring-lag-detected` to better reflect this.

2. **Writer-specific event kind** (`MODEL_MONITORING_WRITER_LAG_DETECTED`) -- Scopes the alert to writers only, but contradicts the HLD which designed the event kind to be shared across writer and applications.

3. **Per-worker alerts with specific ids** (original HLD approach) -- Create one alert per worker with `ids=["{project}.writer.{worker_id}"]`. Naturally scopes to writers, but requires knowing and tracking the worker count on both server and client side.

4. **Prefix/regex matching on ids** -- Use `ids=["{project}.writer.*"]` instead of `ids=["*"]`. Scopes to writer entities without needing to know worker count, but requires adding glob/regex support to the alert matching logic (`_event_entity_matches`).

5. **Separate entity kind per component** -- e.g. `MODEL_MONITORING_WRITER` vs `MODEL_MONITORING_APPLICATION`. Then `*` with the writer entity kind only catches writer events. But introduces a new entity kind just for scoping.

I'd lean toward option 1 since it aligns with the HLD. What do you and @theSaarco think?

mlrun/projects/project.py
alxtkr77: Done, changed to keyword arg `project_name=self.name`.

mlrun/projects/project.py
alxtkr77: Added inline comment: `# Wildcard matches lag events from any writer worker entity`

server/py/services/alerts/crud/events.py
alxtkr77: Added docstring to `add_event_configuration()` explaining the wildcard vs exact-match behavior, and an inline comment in `alerts.py:_event_entity_matches()`.

mlrun/projects/project.py
danielperezz: After consulting with Saar, he thinks a prefix match would be the most suitable solution here, especially considering potential future use cases. How much additional complexity would this add?

assaf758: As I see it, when the future use-cases arrive, we can then update the logic here to be more specific.
Reminding that we are arriving at feature-complete, so I would like to wrap things up if possible.

mlrun/projects/project.py
```diff
@@ -2756,6 +2766,77 @@ def disable_model_monitoring(
...
+        alert_notifications = [
+            alert_constants.AlertNotification(notification=n) for n in notifications
+        ]
+
+        alert_name = mm_constants.MonitoringAlertNames.WRITER_LAG_DETECTED
+        alert_data = mlrun.alerts.alert.AlertConfig(
+            project=self.name,
+            name=alert_name,
+            summary=(
+                "Model monitoring writer (worker {{entity}}) "
```
theSaarco: This may be misleading if setting `count>1`, and the events arrive for different writer instances. This will show the ID of the arbitrary writer that passed the count threshold.

mlrun/projects/project.py
theSaarco: Just to note that while catching all invocations of this event type works per the HLD - the notification description doesn't fit (it mentions writer explicitly). So, if we keep this as-is for simplicity for now, let's make sure the texts fit it.

theSaarco (commented):
I'm ok with keeping it with global `*` for now, as long as some log/notification adjustments are made - see my comments.

mlrun/projects/project.py
alxtkr77: Good point. Removed `{{entity}}` from the summary entirely. Changed to: `"Model monitoring lag detected in project {{project}}."`

The specific worker/component details (worker_id, endpoint_name, etc.) are still available in the event's `value_dict` and visible in the alert history.

mlrun/projects/project.py
alxtkr77: Agreed. Updated the summary to `"Model monitoring lag detected in project {{project}}."` -- no longer mentions writer explicitly, so it fits if the event kind is reused by other components in the future.

alxtkr77: Also renamed the alert constant from `WRITER_LAG_DETECTED` / `"writer-lag-detected"` to `LAG_DETECTED` / `"monitoring-lag-detected"`, and updated all docstrings to remove writer-specific references.

danielperezz (approved):

PR #9321, opened 2026-02-16
[Package] Docs and Pack / Unpack Artifact Types Separation

guy1992l:
### 📝 Description

Added docs and tutorials for packagers and a separation for artifact types check.

---

### 🛠️ Changes Made

- [X] Some packagers might be needed only for packing (or unpacking, although it is more of a niche - explained in docs) so I added an option to separate the artifact types support. Examples for pack-only can be "result" (of course), "plot" for `matplotlib.Figure` to be saved as PNG, etc.
- [X] Added packagers docs, explaining the what, why and how.
- [x] Added end-to-end tutorial to use packagers.
- [X] Added explanation about how to write a custom packagers.
- [X] Added 3 tutorials for writing custom packagers around pack-only, round-trip (pack + unpack) and itemization.

---

### ✅ Checklist

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link:
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [X] No

---


docs/concepts/packagers/custom_packagers/index.md
```diff
@@ -0,0 +1,240 @@
...
+
+def pack_file(
+    self, obj: MyType, key: str, file_format: str = "json"
+) -> tuple[Artifact, dict]:
+    # Serialize to a temporary file
+    path = f"/tmp/{key}.{file_format}"
+    obj.save(path)
+
+    # Create the artifact
+    artifact = Artifact(key=key, src_path=path)
```
theSaarco: It might be useful to explain why you just create an artifact, rather than use the `log_XXX` methods - MLRun users are used to log artifacts, rather than creating them.

docs/concepts/packagers/custom_packagers/index.md
```diff
@@ -0,0 +1,240 @@
...
+```
+
+If your type has subclasses that should also be handled by this packager, set
+`PACK_SUBCLASSES = True`.
+
+### 3. Implement `pack_<type>()` methods
+
+Each packing method serializes the object and returns a tuple of `(Artifact, instructions_dict)`.
+The `instructions_dict` carries metadata needed to reconstruct the object when unpacking.
+For result types, return a plain `dict` with the key and value.
```
theSaarco: This line is a bit cryptic - it is the first time `result` types are mentioned. Also, the example doesn't cover this, so at least add an explanation what is meant or possible add an example for that.

docs/concepts/packagers/custom_packagers/index.md
```diff
@@ -0,0 +1,240 @@
...
+MLRun's {ref}`built-in packagers <packagers>` cover common Python types — scalars,
+collections, NumPy arrays, and Pandas DataFrames. But when your function produces or
+consumes a type that isn't built-in, the default behavior is to **pickle** the object
+with `cloudpickle` (or any configured pickling module of your choice). Pickle files are opaque, 
+not human-readable, and tied to specific Python and module versions — making them fragile and hard to inspect.
+
+A **custom packager** gives you full control over how your type is serialized and
+deserialized, producing readable, portable artifacts (PNG images, JSON configs, CSV
+tables) instead of pickle blobs.
+
```
theSaarco, line 13: Would be nice to have a short reminder here when `packing` is done (function output) and when `unpacking` is done (function input).

docs/concepts/packagers/custom_packagers/index.md
```diff
@@ -0,0 +1,240 @@
...
+    # Clean up the temp file after upload
+    self.add_future_clearing_path(path)
+
+    # Return artifact + instructions for unpacking
+    return artifact, {"file_format": file_format}
+```
+
+The method name determines the artifact type: `pack_file` handles `artifact_type="file"`,
+`pack_plot` handles `"plot"`, and so on.
```
theSaarco: Is there any validation done to ensure the result is of the correct artifact type? What if the user in the function created a different artifact type, will something fail?

docs/concepts/packagers/custom_packagers/index.md
```diff
@@ -0,0 +1,240 @@
...
+
+When running remotely, the worker needs access to your packager module. Set the project
+source with `pull_at_runtime=True` so the code is pulled before the function executes:
+
+```python
+project.set_source(source="./", pull_at_runtime=True)
+```
+
+Without this, the remote worker won't be able to import your packager class and the run
+will fail (if `is_mandatory=True`) or fall back to pickle (if `is_mandatory=False`).
```
theSaarco: That's not entirely accurate - you can still have project source and build image with this source as archive. You can also have your code available on shared storage and point the runtime's execution path to that path. The important thing is that the code is there on the runtime, you can achieve that in more than one way.

docs/concepts/packagers/index.md
```diff
@@ -0,0 +1,388 @@
...
+```{toctree}
+:maxdepth: 1
+
+packagers-tutorial
+custom_packagers/index
+```
+
+## What are packagers?
+
+Writing a function locally and running it remotely should feel identical. Locally, your
```
theSaarco:
```suggestion
Writing a function locally and running it remotely should feel identical. For example, running locally your
```

docs/concepts/packagers/index.md
```diff
@@ -0,0 +1,388 @@
...
+**Output flow:** your function `return` → Python object → packager `pack()` → `Artifact` / `Result` → artifact store
+
+## Why use packagers?
+
+Packagers offer several advantages over manual artifact handling and the legacy context-based API.
+
+### Better and faster learning curve
+
+With packagers you don't need to learn about `Artifact`s, `DataItem`s, or the MLRun context object. You write standard 
+Python with type hints and returning values — MLRun integrates into your existing code without changing it.
```
theSaarco:
```suggestion
Python with type hints and returning values — MLRun wraps your existing code without changing it.
```

docs/concepts/packagers/index.md
```diff
@@ -0,0 +1,388 @@
...
+MLRun automatically converts the `DataItem` to a DataFrame on input and logs the row count as a result and the
+cleaned DataFrame as a dataset artifact on output.
+
+### Uniformity of artifacts between users and projects
+
+ML engineers can define how artifacts are serialized once and have that convention enforced
+across every development notebook, CI pipeline, and production project in the organization.
+Because packagers standardize the serialization format, artifacts become truly portable — a
+DataFrame logged in one project can be consumed by a function in a completely different
+project without conversion steps or format mismatches.
```
theSaarco:
```suggestion
project without conversion steps or format mismatches (of course, given access is allowed across these projects).
```

docs/concepts/packagers/index.md
```diff
@@ -0,0 +1,388 @@
...
+
+Running at scale means dealing with multiple workers, jobs, nodes, resource allocation
+(GPUs, memory), and a distribution framework — concerns that normally require significant
+code changes. 
+
+Packagers are the key to seamless distribution: because they know exactly how
+to serialize and deserialize every argument, MLRun can automatically split inputs across
+workers on different jobs and nodes, assign the right resources to each, and gather the
+outputs back into a single result — all without touching your function code. A function
+with typed I/O can move from a single machine to a distributed cluster with zero changes.
```
theSaarco: Would remove the future-looking statements here - let's document what we have.

docs/concepts/packagers/index.md
```diff
@@ -0,0 +1,388 @@
...
+
+### Adaptive to user needs
+
+MLRun provides common built-in packagers with rich options and configurations. For
+example, you can control the output format with a single log hint:
+
+    returns=['data : dataset[format="parquet"]']
```
theSaarco: I would add an example of using the new `LogHint` as well as using the "legacy" format.

docs/concepts/packagers/index.md
```diff
@@ -0,0 +1,388 @@
...
+
+instead of writing manual pandas I/O and artifact construction.
+
+Beyond built-in packagers, MLRun supports **custom packagers** that you write and
+register in your project to handle domain-specific types. Custom packagers imported
+from the MLRun Hub are coming soon.
```
theSaarco: Again, would not document things that are not there yet.

docs/concepts/packagers/index.md
```diff
@@ -0,0 +1,388 @@
...
+| Field | Type | Description |
+|-------|------|-------------|
+| `key` | `str` | **Required.** The artifact key to log the object under. |
+| `artifact_type` | `str \| None` | The artifact type (e.g. `"dataset"`, `"model"`, `"result"`). If `None`, the packager's default is used. |
+| `tag` | `str` | Tag for the artifact. Default: `""`. |
+| `itemized` | `bool \| int` | Unbundling control. `False` (default): log as one artifact. `True`: fully unbundle. `int`: unbundle to that depth. |
+| `packing_kwargs` | `dict` | Extra keyword arguments passed to the packager's `pack_<type>()` method (e.g. `{"format": "parquet"}`). |
+| `labels` | `dict[str, str]` | Labels to add to the logged artifact. |
+| `extra_data` | `dict` | Extra data to attach to the artifact. Use `...` (Ellipsis) as a value to link to another package by key. |
+| `metrics` | `dict` | Metrics to log alongside a model artifact. Use `...` (Ellipsis) as a value to link to another package by key. |
```
theSaarco: It's hard to understand what the `...` description means. What do you mean by "value to link to another package"?

theSaarco (changes requested):
Very nice! Have some comments and suggestions (I didn't try to "compile" the code in the notebooks to see if it works, went mostly over the md files).

docs/api/mlrun.package/index.rst
```diff
@@ -3,36 +3,43 @@
...
    errors
 
-.. rubric:: Packagers
+.. rubric:: Built-in packager modules
 
-MLRun comes with the following list of modules, out of the box. All of the packagers listed here
-use the implementation of :ref:`DefaultPackager <mlrun.package.packagers.default\_packager.DefaultPackager>` and are
-available by default at the start of each run.
+MLRun includes the following built-in packager modules. All built-in packagers
+subclass :py:class:`~mlrun.package.packagers.default_packager.DefaultPackager` and
```
jillnogold, line 42: subclass :py:class:`~mlrun.package.packagers.default_packager.DefaultPackager`

docs/concepts/packagers/custom_packagers/index.md
```diff
@@ -0,0 +1,240 @@
...
+### `Packager` (full control)
+
+The base {py:class}`~mlrun.package.packager.Packager` class gives you complete control.
+You override `pack()` and `unpack()` directly and manage artifact-type routing, validation,
+and fallback behavior yourself. Use this only when `DefaultPackager`'s convention-based
+approach doesn't fit your needs.
+
+## The three patterns
+
+Custom packagers follow one of three patterns, depending on what your type needs:
```
jillnogold: Custom packagers follow one of four patterns, depending on what your type needs:

docs/concepts/packagers/custom_packagers/index.md
```diff
@@ -0,0 +1,240 @@
...
+from mlrun.package import ArtifactType
+from mlrun.package.packagers.default_packager import DefaultPackager
+
+
+class MyTypePackager(DefaultPackager): ...
+```
+
+### 2. Set class variables
+
+At minimum, set the type your packager handles and the default artifact type:
```
jillnogold: At the minimum, set the type your packager handles and the default artifact type:

docs/concepts/packagers/custom_packagers/index.md
```diff
@@ -0,0 +1,240 @@
...
+def unpack_file(self, data_item: mlrun.DataItem, file_format: str = "json") -> MyType:
+    # Download the artifact to a local path
+    local_path = data_item.local()
+
+    # Reconstruct the object
+    return MyType.load(local_path, format=file_format)
+```
+
+Each instruction parameter (e.g. `file_format`) must be **optional** (have a default
+value) so the method can also handle objects that were logged manually rather than
```
jillnogold: value) so that the method can also handle objects that were logged manually rather than

docs/concepts/packagers/custom_packagers/index.md
```diff
@@ -0,0 +1,240 @@
...
+
+### 5. Clean up temporary files
+
+If your `pack_*` or `unpack_*` methods write files to disk, call
+`self.add_future_clearing_path(path)` so MLRun deletes them after the artifact is
+uploaded. This prevents temporary files from accumulating on the worker.
+
+### 6. Set the priority (optional)
+
+The `PRIORITY` class variable (integer 1–10, lower = higher priority) controls which
```
jillnogold: The `PRIORITY` class variable (integer 1–10, 1 = highest priority) controls which

docs/concepts/packagers/custom_packagers/index.md
```diff
@@ -0,0 +1,240 @@
...
+type depends on runtime conditions rather than a fixed value.
+
+### `Packager` (full control)
+
+The base {py:class}`~mlrun.package.packager.Packager` class gives you complete control.
+You override `pack()` and `unpack()` directly and manage artifact-type routing, validation,
+and fallback behavior yourself. Use this only when `DefaultPackager`'s convention-based
+approach doesn't fit your needs.
+
+## The three patterns
```
jillnogold: ## The four patterns

docs/concepts/packagers/custom_packagers/index.md
guy1992l: the artifact type is not really an `Artifact` kind but more of how to create an Artifact in the flavor of the given artifact type. So the user can't really create something that will cause a failure. The artifact type is noted by the packager manager, the user needs to return an artifact and a set of instructions.

docs/concepts/packagers/index.md
guy1992l: I would rather leave the "legacy" dict format as it has no real "legacy" :)

docs/concepts/packagers/index.md
guy1992l: lets say I:
`return my_model, some_results, loss_plot, test_dataset`

I would want `some_results` to be in the `my_model`'s metrics, and all the rest in `my_model`'s extra data. So I can in the log hint specify all the keys I want with an `...` (python common char for fill me or will be filled) in the metrics and extra data dicts.

I'll try to explain it better in the docs.

docs/api/mlrun.package/index.rst
guy1992l: it is the same as my current change.

docs/concepts/packagers/custom_packagers/index.md
```diff
@@ -0,0 +1,276 @@
...
+class MyTypePackager(DefaultPackager): ...
+```
+
+### 2. Set class variables
+
+At the minimum, set the type your packager handles and the default artifact type:
+
+```python
+class MyTypePackager(DefaultPackager):
+    PACKABLE_OBJECT_TYPE = MyType
```
GiladShapira94: I would add an example for a type, allow users to copy and paste the example

docs/concepts/packagers/custom_packagers/index.md
GiladShapira94: Another question: if the user wants to log a Dataset, should they use the Dataset object, or is it always an artifact?

docs/concepts/packagers/custom_packagers/index.md
```diff
@@ -0,0 +1,276 @@
...
+  ```
+
+* **Build into the function image** — include the packager source in the function's
+  build so it is baked into the container image.
+
+* **Shared storage** — place the packager module on a shared volume and configure the
+  function's working directory to point there.
+
+If the packager module is missing at runtime, the run fails immediately when
+`is_mandatory=True`, or falls back to pickle when `is_mandatory=False`.
```
GiladShapira94: Maybe worth mentioning what is the default behavior?

docs/concepts/packagers/custom_packagers/index.md
```diff
@@ -0,0 +1,276 @@
...
+| `PACKABLE_OBJECT_TYPE` | `...` (any) | The Python type this packager handles. Used by `is_packable` and `is_unpackable` to match objects and type hints. |
+| `PACK_SUBCLASSES` | `False` | When `True`, this packager also handles subclasses of `PACKABLE_OBJECT_TYPE`. |
+| `DEFAULT_PACKING_ARTIFACT_TYPE` | `"object"` | The artifact type to use when the user doesn't specify one in the log hint. |
+| `DEFAULT_UNPACKING_ARTIFACT_TYPE` | `"object"` | The artifact type to use when unpacking a `DataItem` that wasn't originally packed by this packager (e.g. a manually logged artifact). |
+| `BUNDLE_FROM_LIST` | `False` | When `True`, the type can be initialized from a `list` to serve as a bundle container. |
+| `BUNDLE_FROM_DICT` | `False` | When `True`, the type can be initialized from a `dict` to serve as a bundle container. |
+
+`DefaultPackager` **auto-discovers** supported artifact types by scanning for methods
+independently: `pack_*` methods define packing artifact types and `unpack_*` methods
+define unpacking artifact types. If your class has `pack_file` but no `unpack_file`,
```
GiladShapira94: the `pack_file` name prefix is taken from the `DEFAULT_PACKING_ARTIFACT_TYPE`?

docs/concepts/packagers/custom_packagers/index.md
guy1992l: this is exactly that, just copy and change to your type. Plus, there are 3 packagers examples in the tutorials.

docs/concepts/packagers/custom_packagers/index.md
guy1992l: the other way around. You write your `pack_<type>` methods and then set the default one at the  `DEFAULT_PACKING_ARTIFACT_TYPE`.

docs/concepts/packagers/custom_packagers/index.md
guy1992l: there is no default value `def add_custom_packager(self, packager: str, is_mandatory: bool):`. There is no right or wrong here, whatever suites the situation according to the user, the user decides - always.

guy1992l: regarding this question (for some reason I can't answer you directly): 
```
Another question: if the user wants to log a Dataset, should they use the Dataset object, or is it always an artifact?
```
As long as the object is a subclass of `Artifact`, you can return it. `ModelArttifact`, `DatasetArtifact` (like for `numpy` and `pandas`), `LLMPromptArtifact`, etc.

GiladShapira94: Thanks,
1. But if the user does not define `is_mandatory` values, what will be the behavior? Or the user must set this flag.
2. Great, worth adding a note about that, that if the user wants to log a diffrent type of artifact, he should use those objects and refer to the most common `ModelArttifact`, `DatasetArtifact`, `LLMPromptArtifact`

docs/concepts/packagers/custom_packagers/index.md
GiladShapira94: thanks, just I think that worth to mention this dependancy

docs/concepts/packagers/custom_packagers/index.md
guy1992l: 1. no default value means the user MUST set the flag :)
2. will do.

docs/concepts/packagers/custom_packagers/index.md
theSaarco: I still don't see any explanation for that or an example.

docs/concepts/packagers/custom_packagers/index.md
theSaarco: This is still not explained.

docs/concepts/packagers/custom_packagers/index.md
theSaarco: You still say that you have to use `pull_at_runtime=True` - needs to change.

theSaarco (approved):
Approving, though there are some comments that were not addressed - if they were and I missed it, then please let me know.

docs/concepts/packagers/custom_packagers/index.md
guy1992l: written in the note at `custom-packagers-tutorials-overview.md`:

```
Inside a `pack_<artifact_type>` method you create and **return** an `Artifact` object — you do not
call `context.log_artifact()` or `context.log_dataset()`. The packager manager handles
the actual logging and uploading; the pack method's job is only to serialize the data and
describe the artifact.
```

docs/concepts/packagers/index.md
guy1992l: see Linknig artifacts. at `packagers-overview.md`

docs/concepts/packagers/custom_packagers/index.md
guy1992l: from `custotm-packagers-tutorial-overview.md`:

#### Result artifact type

A **special case** of a `pack_<artifact_type>()` method is the **result** artifact type — a scalar 
or simple value (`int`, `float`, `str`, `bool`) stored directly in run metadata (visible 
in `run.status.results` and in the MLRun UI without downloading anything). For result 
types, the pack method returns a plain `dict` with the key and value instead of an 
`(Artifact, instructions)` tuple:

```python
def pack_result(self, obj: MyType, key: str) -> dict:
    # Stored as run metadata, not as a file artifact
    return {key: obj.score}
```

`DefaultPackager` already provides a generic `pack_result` implementation, so you only
need to override it if you want custom extraction logic (e.g. pulling a specific field
from your type). The `"result"` type is always available for packing.

docs/concepts/packagers/custom_packagers/index.md
guy1992l: from `custom-packagers-tutroial-overview.md`:

### 8. Make the packager importable on the remote worker

When running remotely, the worker must be able to import your packager module. There
are several ways to achieve this:

* **Pull at runtime** (simplest) — set the project source with `pull_at_runtime=True`
  so the code is fetched before execution:

  ```python
  project.set_source(source="./", pull_at_runtime=True)
  ```

* **Build into the function image** — include the packager source in the function's
  build so it is baked into the container image

* **Shared storage** — place the packager module on a shared volume and configure the
  function's working directory to point there

If the packager module is missing at runtime, the run fails immediately when
`is_mandatory=True`, or falls back to pickle when `is_mandatory=False`.

liranbg (approved):

PR #9326, opened 2026-02-16
[Auth] Add bulk token deletion endpoint for user cleanup

yaelgen:
### 📝 Description

Add a new `DELETE /api/v1/user-secrets/tokens` endpoint for bulk-deleting all stored offline tokens (K8s secrets) 
for a given user. This is called by Orca during user deletion.
Since the user is already deactivated, token revocation is skipped and only the K8s secrets are removed.

---

### 🛠️ Changes Made
- Added `DeleteSecretTokensResponse` schema (deleted_count, failed_tokens)
- Added `DELETE /tokens` endpoint in `user_secrets.py` with admin authorization
- Added `delete_secret_tokens` async method in secrets CRUD - lists tokens,
  deletes each K8s secret in parallel via `asyncio.gather`, skips revocation
- Extracted `_delete_single_token` helper from `delete_secret_token`
- Added `delete_secret_tokens` to `HTTPRunDB` SDK client

---

### ✅ Checklist

---

### 🧪 Testing

- Unit tests for `delete_secret_tokens` CRUD: success, no tokens
- Tests verify that bulk delete skips revocation (asserts
  `revoke_offline_token` is not called)

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/IG4-1323
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


server/py/services/api/crud/secrets.py
```diff
@@ -574,36 +633,113 @@ def delete_secret_token(
...
+        :param username:
+            The username of the user whose tokens should be deleted.
+            For regular users, this must be their own username.
+            For system admins, this can be any user's username.
+        :param auth_info:
+            Authentication information of the requesting user.
+        :return: DeleteSecretTokensResponse with deleted_count and any failed_tokens.
+        """
+        # Resolve the target user_id from the username
+        target_user_id = self._resolve_target_user_id(auth_info, username)
```
liranbg: This function does not only "resolve" but rather "fetches" the user. so I suggest:
1. rename function to `_get_user_id(auth_info, username)`
2. run in threadpool since this perform IO blocking call under `async` function.

server/py/services/api/crud/secrets.py
```diff
@@ -574,36 +633,113 @@ def delete_secret_token(
...
+            "Deleting all secret tokens for user",
+            target_user_id=target_user_id,
+            target_username=username,
+            requesting_user=auth_info.username,
+        )
+
+        # List all tokens for the user
+        tokens: list[mlrun.common.schemas.SecretTokenInfo] = (
+            self.secrets_provider.list_user_token_secrets(user_id=target_user_id)
+        )
```
liranbg: another IO blocking call?

server/py/services/api/crud/secrets.py
```diff
@@ -574,36 +633,113 @@ def delete_secret_token(
...
+                    skip_revocation=True,
+                )
             )
-        except Exception as exc:
-            logger.error(
-                "Token revoked but failed to delete associated secret",
+            for token_info in tokens
+        ]
+
+        results = await asyncio.gather(*tasks, return_exceptions=True)
```
liranbg: for safeguards - I suggest using semaphore with batches of 10 calls to avoid cases where 1k of tokens creates 1k api calls.

liranbg (changes requested):
initial reivew

mlrun/db/httpdb.py
```diff
@@ -5485,6 +5469,51 @@ def delete_secret_token(
...
+            response = db.delete_secret_tokens()
+            print(f"Deleted {response.deleted_count} tokens")
+
+            # As a system admin, delete all tokens for a specific user
+            response = db.delete_secret_tokens(username="john_doe")
+        """
+        endpoint_path = "user-secrets/tokens"
+        params = None
+        if username is not None:
+            params = {"username": username}
```
rokatyy:
```suggestion
params = {"username": username} if username else None
```

server/py/services/api/crud/secrets.py
```diff
@@ -37,6 +40,9 @@
 import services.api
 import services.api.utils.events.events_factory as events_factory
 
+# Maximum number of concurrent K8s secret deletions during bulk token cleanup
+MAX_CONCURRENT_TOKEN_DELETIONS = 10
```
rokatyy: Do we want to add this configuration to mlrun config?

server/py/services/api/crud/secrets.py
```diff
@@ -574,36 +634,112 @@ def delete_secret_token(
...
+        async def _delete_with_semaphore(token_name: str):
+            async with semaphore:
+                await run_in_threadpool(
+                    self._delete_single_token,
+                    target_user_id=target_user_id,
+                    target_username=username,
+                    token_name=token_name,
+                    iguazio_client=iguazio_client,
+                    request_headers=auth_info.request_headers,
+                    skip_revocation=True,
```
liranbg, line 708: add a comment above to explain why do you choose to set it to True

server/py/services/api/crud/secrets.py
```diff
@@ -574,36 +634,112 @@ def delete_secret_token(
...
-            logger.error(
-                "Token revoked but failed to delete associated secret",
+        semaphore = asyncio.Semaphore(
+            mlrun.mlconf.secret_stores.kubernetes.concurrent_token_deletions
+        )
+
+        async def _delete_with_semaphore(token_name: str):
+            async with semaphore:
+                await run_in_threadpool(
+                    self._delete_single_token,
```
liranbg, line 701: if we only delete the k8s secrets here (given that skip_recovation=True) then why not using k8s  `delete_collection_namespaced_secret` method with label selector representing the user? that way, you will get a single api call to k8s and not need for that threadpool / semaphore all together

liranbg (changes requested):

server/py/services/api/crud/secrets.py
yaelgen: Good catch. It does require refactoring and retesting.
As agreed, Added a TODO and created a follow-up ticket: https://iguazio.atlassian.net/browse/IG4-1510

liranbg (approved):

rokatyy (approved):

PR #9332, opened 2026-02-17
[RemoteRuntime] Fix adding http trigger mode for sync

royischoss:
### 📝 Description

Fix https://iguazio.atlassian.net/browse/ML-12025 adding http trigger mode to trigger spec when using sync mode

---

### 🛠️ Changes Made

Added mode field for default http trigger of serving function + adding to `with_http` method the mode when sync mode is chosen, by default or explicit

---

### ✅ Checklist

---

### 🧪 Testing

added unit test to check the trigger spec, checked that serving behave as expected for both modes

---

### 🔗 References
- Ticket link: [Ticket](https://iguazio.atlassian.net/browse/ML-12025)
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/mlrun/runtimes/nuclio/function.py
+++ b/mlrun/runtimes/nuclio/function.py
@@ -561,9 +561,9 @@ class RemoteRuntime(KubeResource):
             )
         elif validate_nuclio_version_compatibility("1.15.3"):
             trigger._struct["mode"] = (
-                "async" if async_spec and async_spec.enabled else "sync"
+                "async" if getattr(async_spec, "enabled", False) else "sync"
             )
-            if async_spec.enabled:
+            if getattr(async_spec, "enabled", False):
                 trigger._struct["async"] = {
                     "maxConnectionsNumber": async_spec.max_connections,
                     "connectionAvailabilityTimeout": async_spec.connection_availability_timeout,
--- a/mlrun/runtimes/nuclio/function.py
+++ b/mlrun/runtimes/nuclio/function.py
@@ -513,7 +513,14 @@ class RemoteRuntime(KubeResource):
                 "Adding HTTP trigger despite the default HTTP trigger creation being disabled"
             )
 
-        if async_spec and async_spec.enabled:
+        nuclio_version_support_async = validate_nuclio_version_compatibility("1.15.3")
+        if async_spec is not None and not nuclio_version_support_async:
+            raise mlrun.errors.MLRunValueError(
+                "Async spec is only supported from Nuclio 1.15.3"
+            )
+
+        async_enabled = getattr(async_spec, "enabled", False)
+        if async_enabled:
             workers = 1 if workers is None else workers
         else:
             workers = 8 if workers is None else workers
@@ -555,15 +562,9 @@ class RemoteRuntime(KubeResource):
                 )
             trigger._struct["batch"] = batching_config
 
-        if async_spec and not validate_nuclio_version_compatibility("1.15.3"):
-            raise mlrun.errors.MLRunValueError(
-                "Async spec is only supported on Nuclio 1.15.3 and higher"
-            )
-        elif validate_nuclio_version_compatibility("1.15.3"):
-            trigger._struct["mode"] = (
-                "async" if getattr(async_spec, "enabled", False) else "sync"
-            )
-            if getattr(async_spec, "enabled", False):
+        if nuclio_version_support_async:
+            trigger._struct["mode"] = "async" if async_enabled else "sync"
+            if async_enabled:
                 trigger._struct["async"] = {
                     "maxConnectionsNumber": async_spec.max_connections,
                     "connectionAvailabilityTimeout": async_spec.connection_availability_timeout,
--- a/mlrun/runtimes/nuclio/function.py
+++ b/mlrun/runtimes/nuclio/function.py
@@ -503,8 +503,9 @@ class RemoteRuntime(KubeResource):
         :param batching_spec: BatchingSpec object that defines batching configuration.
             By default, batching is disabled.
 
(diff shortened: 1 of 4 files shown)


mlrun/runtimes/nuclio/function.py
```diff
@@ -555,13 +555,15 @@ def with_http(
...
-                trigger._struct["mode"] = "async"
+        if async_spec and not validate_nuclio_version_compatibility("1.15.3"):
+            raise mlrun.errors.MLRunValueError(
+                "Async spec is only supported on Nuclio 1.15.3 and higher"
+            )
+        elif validate_nuclio_version_compatibility("1.15.3"):
+            trigger._struct["mode"] = (
+                "async" if getattr(async_spec, "enabled", False) else "sync"
+            )
+            if getattr(async_spec, "enabled", False):
```
gtopper: I guess I'd extract `getattr(async_spec, "enabled", False)`.

mlrun/runtimes/nuclio/function.py
```diff
@@ -555,13 +555,15 @@ def with_http(
...
-                raise mlrun.errors.MLRunValueError(
-                    "Async spec is only supported on Nuclio 1.15.3 and higher"
-                )
-            if async_spec.enabled:
-                trigger._struct["mode"] = "async"
+        if async_spec and not validate_nuclio_version_compatibility("1.15.3"):
+            raise mlrun.errors.MLRunValueError(
+                "Async spec is only supported on Nuclio 1.15.3 and higher"
+            )
+        elif validate_nuclio_version_compatibility("1.15.3"):
```
gtopper: Should probably extract `validate_nuclio_version_compatibility("1.15.3")`.

mlrun/runtimes/nuclio/function.py
```diff
@@ -555,13 +555,15 @@ def with_http(
...
-        if async_spec:
-            if not validate_nuclio_version_compatibility("1.15.3"):
-                raise mlrun.errors.MLRunValueError(
-                    "Async spec is only supported on Nuclio 1.15.3 and higher"
-                )
-            if async_spec.enabled:
-                trigger._struct["mode"] = "async"
+        if async_spec and not validate_nuclio_version_compatibility("1.15.3"):
+            raise mlrun.errors.MLRunValueError(
+                "Async spec is only supported on Nuclio 1.15.3 and higher"
```
gtopper:
```suggestion
                "Async spec is only supported from Nuclio 1.15.3"
```

mlrun/runtimes/nuclio/function.py
```diff
@@ -555,13 +555,15 @@ def with_http(
                 )
             trigger._struct["batch"] = batching_config
 
-        if async_spec:
-            if not validate_nuclio_version_compatibility("1.15.3"):
```
gtopper, line 559: I think the original, nested structure was cleaner, actually.

server/py/services/api/tests/unit/api/test_nuclio.py
```diff
@@ -237,3 +237,61 @@ def test_mlrun_function_translation_to_nuclio(
...
+)
+def test_with_http_async_spec(
+    async_spec, expected_mode, expected_async_struct, expected_workers
+):
+    """Test with_http method with various async_spec configurations."""
+    func = mlrun.runtimes.nuclio.function.RemoteRuntime()
+
+    with patch(
+        "mlrun.runtimes.nuclio.function.validate_nuclio_version_compatibility",
+        return_value=True,
```
gtopper: Should probably test the negative case as well.

gtopper (commented):
Minor comments – LGTM otherwise!

mlrun/runtimes/nuclio/function.py
royischoss: The reason I changed it was because we want to affect mode field for this version even if the user didn't provided async_spec

royischoss: Changed the structure a bit checking it early

mlrun/runtimes/nuclio/function.py
```diff
@@ -513,7 +513,14 @@ def with_http(
                 "Adding HTTP trigger despite the default HTTP trigger creation being disabled"
             )
 
-        if async_spec and async_spec.enabled:
+        nuclio_version_support_async = validate_nuclio_version_compatibility("1.15.3")
+        if async_spec is not None and not nuclio_version_support_async:
```
gtopper, line 518: Minor issue here – in case of `AsyncSpec(enabled=False)`, we'll still raise the error, which shouldn't be the case. Might want to cover this in test as well.

server/py/services/api/crud/runtimes/nuclio/helpers.py
```diff
@@ -134,6 +135,9 @@ def enrich_function_with_ingress(config, mode, service_type):
             "workerAvailabilityTimeoutMilliseconds": 10000,  # 10 seconds
             "attributes": {},
         }
+        # Provide trigger mode for nuclio 1.15.3 and above.
+        if validate_nuclio_version_compatibility("1.15.3"):
+            http_trigger["mode"] = "sync"
```
gtopper, line 140: I just realized that the added test doesn't verify this fix. Consider adding another test, probably directly for `enrich_function_with_ingress()`, in `server/py/services/api/tests/unit/runtimes/test_nuclio.py`.

mlrun/runtimes/nuclio/function.py
gtopper: We discussed this and it turns out that this is on purpose, with the thinking that it's not correct to specify any AsyncSpec with an old version of Nuclio.

gtopper (approved):

PR #9263, opened 2026-02-01
[Serving] Add support for streaming models in MRS

gtopper:
[ML-11878](https://iguazio.atlassian.net/browse/ML-11878)

Enables Model subclasses to return generators from `predict()` / `predict_async()` and have them correctly propagate as stream responses through `ModelRunnerStep`.

* Override `is_streaming()` method in `Model` to detect generator-based predict methods.
* Bumps storey for newly-added `storey.ParallelExecutionRunnable.is_streaming()`.
* Add unit tests for streaming MRS with naive and process-based execution mechanisms.
* Extend system test to cover both ordinary (`Map`-based) streaming step, and `ModelRunnerStep` with a streaming `Model`.

[ML-11878]: https://iguazio.atlassian.net/browse/ML-11878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9337, opened 2026-02-18
[Tests] Drop timing assertion in new system test

gtopper:
As it cannot be relied on in CI.

Following test failure in https://github.com/mlrun/mlrun/actions/runs/22123290962/job/63948573423.

This test was added in #9263.


royischoss (approved):
LGTM

PR #9335, opened 2026-02-17
[Model Providers] Add streaming support

gtopper:
[ML-11936](https://iguazio.atlassian.net/browse/ML-11936)

Integrate the model providers API with the streaming API, and implement streaming in both OpenAI and HuggingFace providers.

### Changes

- **`ModelProvider` base class**: Added `supports_streaming` flag and `invoke_stream()` / `async_invoke_stream()` methods (default `NotImplementedError`).
- **`OpenAIProvider`**: Implemented sync and async streaming via `client.chat.completions.create(stream=True)`. Extracted `_prepare_stream_kwargs` and `_extract_stream_token` helpers to share logic between sync/async paths.
- **`HuggingFaceProvider`**: Implemented sync streaming using `TextIteratorStreamer` with a background generation thread. No `async_invoke_stream` — storey's `ParallelExecutionRunnable` handles async execution of the sync generator.
- **`LLModel`**: Overrode `is_streaming()` to check both `server.streaming` and `provider.supports_streaming`. `predict()` / `predict_async()` return the provider's stream generator directly in streaming mode. Added init-time validation that rejects unsupported provider + streaming combinations.

### Tests

- **Unit tests**: Provider streaming methods (both providers), `LLModel` streaming integration (`tests/serving/test_streaming.py`).
- **Integration tests**: End-to-end provider streaming against real APIs (`tests/integration/model_providers/`).
- **System tests**: Deployed Nuclio streaming functions with chunked HTTP response validation (`tests/system/datastore/model_providers/`).

[ML-11936]: https://iguazio.atlassian.net/browse/ML-11936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9343, opened 2026-02-19
[Model Monitoring] Fix tracking for streaming `LLModel` responses

gtopper:
## Summary

Follow up on #9335. Adds a streaming tracking system test as per [ML-12182](https://iguazio.atlassian.net/browse/ML-12182), and fixes two related issues in the monitoring pipeline for streaming `LLModel` responses.

### Fix 1: Skip `result_path` for stream-collected scalars

When streaming is enabled for `LLModel` through `ModelRunnerStep`, the `Collector` gathers text tokens and the `MonitoringPreProcessor` aggregates them into a plain string. However, `reconstruct_request_resp_fields` then attempted to extract outputs via `result_path` from that string, raising `MLRunInvalidArgumentError` because it expected a dict. As a result, monitoring events were silently dropped and no predictions were recorded in TSDB or parquet.

The fix ignores `result_path` for stream-collected events when the aggregated body is a scalar (not a dict or list), since the body is raw output — not wrapped in a result-path structure.

### Fix 2: Remove broken `None` padding in `MapFeatureNames`

`MapFeatureNames._map_dictionary_values` padded missing prediction values with `None` when the schema (from the DB) had more fields than the actual data. For streaming models, the schema declares `["answer", "usage"]` but streaming only produces `["answer"]`, so `usage` was padded to `None`. This caused a silent failure in `storey.TSDBTarget`: it inferred `usage` as a column from the dict keys but skipped `None` during data extraction, resulting in a column count mismatch (`ValueError: 10 columns passed, passed data had 9 columns`). As a result, `endpoint_features` records were never written to the EVENTS TSDB table, and the error was swallowed.

The fix removes the `None` padding. `zip` naturally stops at the shorter list, so missing fields are simply not included rather than being set to `None` and causing downstream failures.

## Changes

* **mlrun/serving/system_steps.py**: Skip `result_path` extraction in `reconstruct_request_resp_fields` when the event is stream-collected and the body is not a dict/list.
* **mlrun/model_monitoring/stream_processing.py**: Remove `None` padding in `MapFeatureNames._map_dictionary_values` — let `zip` stop naturally when prediction values are shorter than the schema.
* **mlrun/datastore/\_\_init\_\_.py**: Truncate verbose `_DummyStream` log output to avoid flooding test logs.
* **tests/serving/test_tracking.py**: Add unit tests for stream-collected events with `result_path` set and multi-field output schemas.
* **tests/system/datastore/model_providers/test_remote_model_tracking.py**: Add `test_llmodel_streaming_tracking` system test — deploys a streaming function with MockModelProvider + model monitoring, verifies chunked HTTP streaming response, and validates that monitoring data (model endpoint, TSDB predictions, TSDB endpoint_features, parquet) is recorded correctly.

[ML-12182]: https://iguazio.atlassian.net/browse/ML-12182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/system/datastore/model_providers/test_remote_model_tracking.py
```diff
@@ -409,3 +411,89 @@ def send_event(event, delay):
         assert len(error_df) == 1
         error_dict = error_df.head(1).to_dict(orient="records")[0]
         assert error_dict["error_count"] == 1
+
+    @pytest.mark.parametrize(
+        "execution_mechanism",
+        ["process_pool", "dedicated_process", "naive", "asyncio", "thread_pool"],
+    )
+    def test_llmodel_streaming_tracking(self, execution_mechanism):
```
tomerm-iguazio, line 419: can you add problematic invocation that will raise an error?
you can see what I did by myself with `error_input`.

- Add this check also to unit tests, with checking mock stream
- Here in the system test, check the error table as we do in other tests.

tomerm-iguazio (changes requested):
Nice, I have 1 comment

tests/system/datastore/model_providers/test_remote_model_tracking.py
gtopper: Opened https://github.com/mlrun/storey/pull/616 to fix an issue exposed by the new tests.

gtopper: Updated PR with new tests skipped and new test extension commented out until storey PR is merged and released.

tomerm-iguazio (approved):
Approved:)

PR #9346, opened 2026-02-19
[Application] Improve single-file source artifact user experience

Yacouby:
### 📝 Description
This PR improves the single-file source deployment UX for Application Runtime:
- Preserve original filename: When the init container downloads a source artifact, it now uses `artifact.spec.src_path` to restore the original filename instead of the artifact-key-based name (`my-app-source.py`). 

- Auto-restore local path between deploys: `spec.build.source` is temporarily swapped to the `store:// `URI during deploy and restored to the original local path afterward. Subsequent deploys automatically re-upload the file without the user having to manually reset `spec.build.source`.

---

### 🛠️ Changes Made
- `mlrun/utils/clones.py` — Use a`rtifact.spec.src_path` for the download filename 
- `mlrun/runtimes/nuclio/application/application.py` — Store artifact URI in `_artifact_source_uri` instead of overwriting `spec.build.source`; swap/restore in deploy() so the server gets the URI while the user-facing source stays as the local path

---

### ✅ Checklist

---

### 🧪 Testing
Updated unit and system tests accrodingly

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-12184
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


tests/system/runtimes/test_application.py
```diff
@@ -449,7 +446,7 @@ def _create_simple_flask_application(self, name="simple-flask-app", source=None)
         function.spec.args = [
             "-m",
             "flask",
-            f"--app={name}-source",
+            "--app=simple_flask_app",
```
liranbg, line 449: why not taking from name? or the `self._simple_flask_app`?

liranbg (commented):
lgtm, minor question

tests/system/runtimes/test_application.py
Yacouby: - `name` is the function name with - not the Flask module name.
- `self._simple_flask_app` includes the `.py` extension and Flask does not expect it

can change it to remove the suffix but hardcoded is simpler in this case..

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -527,8 +527,13 @@ def deploy(
...
         :return: The default API gateway URL if created or True if the function is ready (deployed)
         """
-        # Upload local single-file source as artifact (if applicable)
+        # Upload local source as artifact. The server needs the store:// URI to configure the init container, but we
+        # restore the local path afterward, so subsequent deploys re-upload the file.
+        original_local_source = self.spec.build.source
         self._upload_source_as_artifact()
+        artifact_uri = getattr(self, "_artifact_source_uri", None)
+        if artifact_uri:
+            self.spec.build.source = artifact_uri
```
TomerShor: shouldn't this whole section be inside `_upload_source_as_artifact`? 
because if `source` is empty it isn't needed

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -564,8 +569,11 @@ def deploy(
             "Successfully deployed function.",
         )
 
+        # Restore the original local source path so subsequent deploys re-upload automatically
+        if artifact_uri and original_local_source:
+            self.spec.build.source = original_local_source
```
TomerShor: what if there is an exception raised in one of the functions above (e.g. `super().deploy`)? we would not reset this value

TomerShor (changes requested):

TomerShor (approved):
✅

PR #9358, opened 2026-02-23
[Serving] Add wildcard star path matching to API handler

jond01:
## Summary

Implements wildcard `*` path matching for the serving API handler - [ML-11658](https://iguazio.atlassian.net/browse/ML-11658).

### SDK (`mlrun/runtimes/nuclio/serving.py`)
- New `APIHandlerConfig._validate_path()` validates `*` rules at `add_endpoint_handler` time — fails early on the client side
- Rejects `*` that is not the last character (e.g. `/api/*/users`)
- Rejects paths with more than one `*`
- Raises `MLRunValueError`

### API Handler (`mlrun/serving/api_handler.py`)
- Three-phase matching with strict precedence: **exact → template → star**; insertion order respected within each group
- Star patterns compiled to a plain prefix string (`/api/*` → prefix `/api/`)
- `/api` does **not** match `/api/*`; `/apiv2/x` does **not** match `/api/*`
- All wildcard validation errors use `MLRunValueError`

### Tests (`tests/serving/test_api_handler.py`)
- 127 unit tests, all passing
- `TestStarPatternMatching`: 14 tests — basic match, validation, precedence, insertion order, method isolation, prefix boundary, allow/forbid
- `TestAPIHandlerConfig`: 3 new SDK-side validation tests
- Template insertion-order precedence test

### System test (`tests/system/runtimes/test_serving.py`)
- New `test_api_handler_wildcard_path`: serving function with `include_url_info=True` and `POST /api/wildcard/*`, asserts `mlrun_request_path`
- New asset: `tests/system/runtimes/assets/wildcard_path_handler.py`

## Testing
- Unit: `pytest tests/serving/test_api_handler.py` — 127 passed
- System - passed.

[ML-11658]: https://iguazio.atlassian.net/browse/ML-11658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/runtimes/nuclio/serving.py
+++ b/mlrun/runtimes/nuclio/serving.py
@@ -138,14 +138,16 @@ class APIHandlerConfig(mlrun.model.ModelObj):
         :param path: Normalized path (with leading ``/``) to validate.
         :raises mlrun.errors.MLRunValueError: If the path contains an invalid ``*`` pattern.
         """
-        if "*" not in path:
+        star_count = path.count("*")
+        if star_count == 0:
             return
-        if not path.endswith("*"):
+        # We know there is a wildcard, validate its position and count
+        if path[-1] != "*":
             raise mlrun.errors.MLRunValueError(
                 f"Invalid endpoint path '{path}': "
                 f"wildcard '*' must be at the end of the path"
             )
-        if path.count("*") > 1:
+        if star_count > 1:
             raise mlrun.errors.MLRunValueError(
                 f"Invalid endpoint path '{path}': "
                 f"wildcard '*' must appear only once at the end of the path"
--- a/mlrun/runtimes/nuclio/serving.py
+++ b/mlrun/runtimes/nuclio/serving.py
@@ -59,17 +59,19 @@ serving_subkind = "serving_v2"
 class APIHandlerConfig(mlrun.model.ModelObj):
     """Configuration for API handler in serving graph"""
 
-    _dict_fields = ["enabled", "endpoints", "body_map"]
+    _dict_fields = ["enabled", "endpoints", "body_map", "include_url_info"]
 
     def __init__(
         self,
         enabled: bool = True,
         endpoints: dict[str, dict] | None = None,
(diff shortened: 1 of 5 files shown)


mlrun/runtimes/nuclio/serving.py
```diff
@@ -124,6 +126,31 @@ def _normalize_path(path: str) -> str:
...
+
+        * ``*`` may only appear once.
+        * ``*`` must be the last character in the path.
+
+        :param path: Normalized path (with leading ``/``) to validate.
+        :raises mlrun.errors.MLRunValueError: If the path contains an invalid ``*`` pattern.
+        """
+        if "*" not in path:
+            return
+        if not path.endswith("*"):
```
assaf758: since you already know there is a "x" in path, an easy optimization would be to just to check if path[-1] is "*".

mlrun/runtimes/nuclio/serving.py
```diff
@@ -124,6 +126,31 @@ def _normalize_path(path: str) -> str:
...
+            raise mlrun.errors.MLRunValueError(
+                f"Invalid endpoint path '{path}': "
+                f"wildcard '*' must be at the end of the path"
+            )
+        if path.count("*") > 1:
```
assaf758: Checking for no "*" is roughly the same as counting their number  (both O(N)) - so again an easy optimization would be to start with having the count, and then abort the function if it's 0 or > 1.

assaf758 (commented):
review still in progress

mlrun/runtimes/nuclio/serving.py
jond01: You're right :) I thought about it, fixed now

mlrun/runtimes/nuclio/serving.py
jond01: Changed together with the other comment

assaf758 (approved):

PR #9378, opened 2026-02-26
[Model Monitoring] Use per-project Kafka consumer group

alxtkr77:
## Summary

- Set the Kafka consumer group to the topic name (already unique per project+function) instead of the shared profile group
- Prevents cross-project rebalance storms that cause data loss when multiple monitoring projects share the same consumer group
- Migrate committed offsets from the old shared group to the new per-topic group on upgrade

From [Kafka Consumer Group Rebalance (Lydtech)](https://www.lydtechconsulting.com/blog/kafka-consumer-group-rebalance---part-1-of-2):
> "If a service has multiple consumers that listen to mutually exclusive topics but that share the same group.id then any rebalance triggered by any one consumer would still affect the other consumers in the group."

## Changes Made

- **`server/py/services/api/crud/model_monitoring/deployment.py`**:
  - `group=kafka_profile.group` → `group=topic` for per-project+function isolation
  - Add `_migrate_consumer_group_offsets()` — on upgrade, copies committed offsets from the old `"serving"` group to the new per-topic group
- **`mlrun/datastore/datastore_profile.py`**: Change `DatastoreProfileKafkaStream.group` default from `"serving"` to `None`
- **`mlrun/projects/project.py`**: Warn in `set_model_monitoring_credentials()` when Kafka profile has explicit `group` (ignored for model monitoring)
- **`server/py/services/api/tests/unit/crud/model_monitoring/test_deployment.py`**: Tests for consumer group assignment and offset migration

## Backward Compatibility

On upgrade, `_migrate_consumer_group_offsets()` copies committed offsets from the old `"serving"` group to the new per-topic group so consumers resume where they left off.

## Reference

- Jira: ML-11979
- [Kafka Consumer Group Rebalance - Lydtech](https://www.lydtechconsulting.com/blog/kafka-consumer-group-rebalance---part-1-of-2)

### DIFF
--- a/server/py/services/api/crud/model_monitoring/deployment.py
+++ b/server/py/services/api/crud/model_monitoring/deployment.py
@@ -441,12 +441,21 @@ class MonitoringDeployment:
         topic = mlrun.common.model_monitoring.helpers.get_kafka_topic(
             project=self.project, function_name=function_name
         )
+        if kafka_profile.group is not None:
+            logger.warning(
+                "Kafka profile 'group' is ignored for model monitoring;"
+                " using topic name as consumer group to prevent"
+                " cross-project rebalance storms",
+                project=self.project,
+                configured_group=kafka_profile.group,
+                effective_group=topic,
+            )
         profile_attributes = kafka_profile.attributes()
         stream_source = mlrun.datastore.sources.KafkaSource(
             brokers=kafka_profile.brokers,
             topics=[topic],
             # Use topic as consumer group to isolate per project+function,
-            # preventing cross-project rebalance storms (ML-11979).
+            # preventing cross-project rebalance storms.
             group=topic,
             initial_offset=kafka_profile.initial_offset,
             partitions=kafka_profile.partitions,
@@ -474,7 +483,7 @@ class MonitoringDeployment:
                     # from the old shared consumer group to per-topic groups.
                     self._migrate_consumer_group_offsets(
                         kafka_profile=kafka_profile,
-                        old_group=kafka_profile.group,
+                        old_group="serving",
                         new_group=topic,
                         topic=topic,
                     )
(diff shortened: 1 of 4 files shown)


server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -445,7 +445,9 @@ def _apply_and_create_kafka_source(
         stream_source = mlrun.datastore.sources.KafkaSource(
             brokers=kafka_profile.brokers,
             topics=[topic],
-            group=kafka_profile.group,
+            # Use topic as consumer group to isolate per project+function,
+            # preventing cross-project rebalance storms (ML-11979).
+            group=topic,
```
assaf758, line 450: Question is - do we want to just ignore the user's `DatastoreProfileKafkaStream.group`?
Alternatively we can:
- Remove the default "serving" for `group`
- Use topic if group was not specified
- Mention that the group should be uniq across projects (force it?)

alxtkr77: Changed the default `group` from `"serving"` to `None`. When the user explicitly sets a group, we log a warning that it's ignored for model monitoring (consumer group is always the topic name for isolation). The offset migration now hardcodes `"serving"` as the old group since that's the historical value in Kafka.

server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -441,11 +441,22 @@ def _apply_and_create_kafka_source(
         topic = mlrun.common.model_monitoring.helpers.get_kafka_topic(
             project=self.project, function_name=function_name
         )
+        if kafka_profile.group is not None:
+            logger.warning(
```
assaf758: I don't think the user will see it (unless he checks in the mlrun-api logs, which is less likely).
How about checking in set_model_monitoring_credentials? The group field is not private..

alxtkr77: Moved the warning to `set_model_monitoring_credentials()` in `project.py` — resolves the stream profile client-side and warns if `group` is set.

assaf758 (approved):

PR #9383, opened 2026-02-26
[Serving] Fix QueueStep initialization 

davesh0812:
##### Overview

Refactors step initialization to support multi-function deployments. Introduces a centralized `_should_init` flag that controls whether steps initialize their class objects, ensuring efficient resource usage.

##### Problem

- All steps initialize class objects regardless of their assigned function
- No unified initialization control mechanism across step types
- `function` property only in `TaskStep`, not available to all step types
- Duplicate `_is_local_function()` logic across classes

###### 1. Centralized Control via `_should_init` Flag

Added to `BaseStep`:
```python
self._should_init = self._is_local_function(context)
```

###### 2. Unified Pattern

All steps follow this pattern:
```python
def init_object(self, context, namespace, mode="sync", reset=False, engine="async", **extra_kwargs):
    super().init_object(context, namespace, mode, reset, engine, **extra_kwargs)
    if not self._should_init:
        return  # skip initialization for non-local functions
    # ... step-specific initialization ...
```

**Updated Classes:** `TaskStep`, `RouterStep`, `ModelRunnerStep`, `QueueStep`, `FlowStep`, `RootFlowStep`, `HubTaskStep`

###### 3. Key Changes

- Moved `function` property to `BaseStep` (all steps support multi-function assignment)
- Moved `_is_local_function()` to `BaseStep`
- Extracted `set_class_object_or_handler()` in `TaskStep` and `HubTaskStep`
- Refactored `ModelRunnerStep.init_object()` to `_init_class_object_and_handler()`
- Extracted `_init_shared_resources()` in `RootFlowStep`
- Added `engine` parameter to all `init_object()` methods

###### 4. Special Note: RootFlowStep

**`RootFlowStep` is always local** - it always initializes because it manages the entire flow and shared resources. The `_should_init` check is not skipped for RootFlowStep; it always proceeds with initialization.

##### Key Points

- ✅ `self.context` is **always** set (even for non-local functions)
- ✅ Heavy initialization only happens when `_should_init=True`
- ✅ Backward compatible - no breaking changes
- ✅ Single-function deployments work unchanged

---

### ✅ Checklist

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-12209
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [ ] No

---

### DIFF
--- a/mlrun/runtimes/nuclio/serving.py
+++ b/mlrun/runtimes/nuclio/serving.py
@@ -805,7 +805,7 @@ class ServingRuntime(nuclio_function.RemoteRuntime):
                     class_args,
                     name=key,
                     handler=handler,
-                    function=child_function,
+                    function=child_function or graph.function,
                     model_endpoint_creation_strategy=creation_strategy,
                     endpoint_type=schemas.EndpointType.LEAF_EP,
                 )
--- a/mlrun/serving/states.py
+++ b/mlrun/serving/states.py
@@ -3689,6 +3689,8 @@ class RootFlowStep(FlowStep):
             if isinstance(step, mlrun.serving.MonitoredStep)
         }
 
+    def _is_local_function(self, context, current_function=None) -> bool:
+        return True
 
 class HubTaskStep(TaskStep):
     """hub task execution step, runs a class or handler from a hub"""
--- a/mlrun/serving/states.py
+++ b/mlrun/serving/states.py
@@ -3692,6 +3692,7 @@ class RootFlowStep(FlowStep):
     def _is_local_function(self, context, current_function=None) -> bool:
         return True
 
+
 class HubTaskStep(TaskStep):
     """hub task execution step, runs a class or handler from a hub"""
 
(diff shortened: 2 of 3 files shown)


mlrun/serving/states.py
```diff
@@ -2649,9 +2670,20 @@ def __init__(
...
+        mode: str = "sync",
+        reset: bool = False,
+        engine: str = "async",
+        **extra_kwargs,
+    ):
+        super().init_object(context, namespace, mode, reset, engine, **extra_kwargs)
+        if not self._should_init:
+            # skip further initialization for non-local functions
+            return
+        if engine == "sync" and self.path:
```
royischoss, line 2686: I am not sure if I know the use of mode. but why do we have two fields? to know where storey is flow?

mlrun/runtimes/nuclio/serving.py
```diff
@@ -805,7 +805,7 @@ def add_model(
                     class_args,
                     name=key,
                     handler=handler,
-                    function=child_function,
+                    function=child_function or graph.function,
```
royischoss, line 808: what changed here? the way we behave to None?

royischoss (commented):
LGTM are we covered for those changes in tests? even unit tests

PR #9262, opened 2026-02-01
[Serving] Add a basic API handler option to the graph

jond01:
### 📝 Description

This PR adds an API handler feature to MLRun serving graphs, enabling fine-grained control over which HTTP endpoints are accessible.

### Key Changes

**Core Implementation:**
- Added `_APIHandlerStep` class for routing and validating serving requests based on configured endpoints
- Added `APIHandlerConfig` class to manage endpoint configurations (path, HTTP method, action)
- Introduced `APIHandlerAction` enum with `ALLOW` and `FORBID` actions
- Implemented `_add_api_handler_step_to_graph()` to automatically inject the API handler as the first step in serving graphs

**API Enhancements:**
- Added `set_api_handler_config()` method to `ServingRuntime` for configuring endpoint access control
- Helper functions `_combine_serving_endpoint_key()` and `_split_serving_endpoint_key()` for endpoint key management

**Testing:**
- Comprehensive unit tests covering `APIHandlerConfig`, `_APIHandlerStep`, endpoint key helpers, and graph integration
- System tests for end-to-end validation
- Mock server integration tests for various HTTP methods and paths

### Usage Example

```python
from mlrun.runtimes.nuclio.serving import APIHandlerConfig
from mlrun.common.schemas.serving import APIHandlerAction
from http import HTTPMethod

config = APIHandlerConfig()
config.add_endpoint_handler("/predict", HTTPMethod.POST, APIHandlerAction.ALLOW)
config.add_endpoint_handler("/admin", HTTPMethod.POST, APIHandlerAction.FORBID)

serving_fn.set_api_handler_config(config)
```

---

### ✅ Checklist
- [x] I updated the documentation (if applicable)
- [x] I have tested the changes in this PR
- [x] I confirmed whether my changes are covered by system tests
  - [x] If yes, I ran all relevant system tests and ensured they passed before submitting this PR

---

### 🔗 References

- Ticket link: [ML-11670](https://iguazio.atlassian.net/browse/ML-11670)
- Design docs links: linked in the feature.

---

### 🚨 Breaking Changes?

- [x] No

[ML-11670]: https://iguazio.atlassian.net/browse/ML-11670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9392, opened 2026-03-02
[Serving] Remove `context.current_event` that prevented last stream offset from being committed

gtopper:
Fixes: [ML-11979](https://iguazio.atlassian.net/browse/ML-11979), [ML-12131](https://iguazio.atlassian.net/browse/ML-12131), possibly [ML-11919](https://iguazio.atlassian.net/browse/ML-11919).

The API handler PR (#9262) added `context.current_event = event` in `v2_serving_handler` to pass HTTP metadata to `_APIHandlerStep`. This was redundant since the step is registered with `full_event=True`, which already provides the full Nuclio event to `do()`. The stale reference on the long-lived `Context` object prevented storey's weakref-based offset commit mechanism from collecting the last event, causing committed=0 in stream stats.

This went undetected because storey had an overcommitting bug (introduced in mlrun/storey#604, storey v1.11.7) that committed offsets on every event regardless of whether the event had been fully processed. That bug masked the reference leak since offsets were committed eagerly before the weakref mechanism was even consulted. When storey fixed the overcommitting bug (mlrun/storey#618, storey v1.11.18), the weakref-based commit path became the sole mechanism again, and the pinned last event was exposed as uncommittable.

Remove the unused `context.current_event` assignment and the dead fallback code in `_APIHandlerStep`. Fix tests to pass `MockEvent` objects matching the real `full_event=True` behavior.

Fixes `test_app_flow` after releasing the `storey==1.11.16` pin.

Smoked passed – https://github.com/mlrun/mlrun/actions/runs/22617698080.

[ML-11979]: https://iguazio.atlassian.net/browse/ML-11979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ML-12131]: https://iguazio.atlassian.net/browse/ML-12131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ML-11919]: https://iguazio.atlassian.net/browse/ML-11919?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alxtkr77:
### Architectural concern: the weakref-based commit mechanism is fragile

The PR fixes *this* instance, but the underlying design means any future code — including user-written custom steps — can reintroduce the same class of bug with zero warnings.

Consider a perfectly reasonable user-written step:

```python
class DiffDetector(MapClass):
    """Compute difference from previous event — a natural streaming pattern."""

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self._last_event = None

    def do(self, event):
        if self._last_event:
            event["diff"] = event["value"] - self._last_event["value"]
        self._last_event = event  # holds strong ref → offset never committed
        return event
```

This is idiomatic Python. Nothing looks wrong. But `self._last_event` pins the previous event, `weakref.ref(event)()` never returns `None`, `_EventOffset.is_ready_to_commit()` stays `False`, and since storey commits sequentially, **all subsequent offsets are blocked too**. The result: silently growing Kafka consumer lag and data loss on rebalance — with no error, no warning, nothing in any log.

The core issue is in storey's `_EventOffset` ([sources.py:270-276](https://github.com/mlrun/storey/blob/development/storey/sources.py#L270-L276)):

```python
class _EventOffset:
    def __init__(self, event):
        self.event_weakref = weakref.ref(event)  # correctness depends on GC

    def is_ready_to_commit(self):
        return self.event_weakref() is None  # None = no strong refs = "done"
```

This makes Kafka offset correctness dependent on Python's garbage collector and the absence of stray references across the entire codebase and all user code. That invariant is unenforceable — you can't lint for it, you can't test for it, and you can't document it effectively enough to prevent it.

**Suggestion:** Open a storey issue to track moving toward explicit completion signaling (e.g., structural tracking based on graph topology, or explicit ack at terminal steps with a timeout safety net). The weakref approach trades offset correctness for memory safety — but there are designs that provide both. This doesn't need to block the current PR, but it should be tracked.

mlrun/serving/api_handler.py
```diff
@@ -305,17 +305,7 @@ def _extract_query_params(
...
-                nuclio_fields = event.fields
-            elif (
-                self.context
-                and hasattr(self.context, "current_event")
-                and hasattr(self.context.current_event, "fields")
-                and self.context.current_event.fields
-            ):
-                nuclio_fields = self.context.current_event.fields
-
+            nuclio_fields = getattr(event, "fields", None)
```
alxtkr77, line 308: ### Suggestion: Replace MagicMock with `nuclio_sdk.Event` in tests

Several tests (mostly in `test_body_map.py`) still pass `MagicMock()` as the event to `step.do()`. This causes tests to silently exercise wrong code paths.

`MagicMock` is a test double that auto-creates any attribute on access — accessing `mock.anything` returns another `MagicMock` instead of raising `AttributeError`. This is useful when you need a stand-in for a complex dependency, but harmful when the real object is simple and the code under test branches on attribute presence or truthiness.

In these tests, `MagicMock` is used like this:

```python
event = MagicMock()
event.method = HTTPMethod.POST
event.path = "/predict"
event.body = {"payload": {"data": [1, 2]}}
result = step.do(event)
```

The test only sets `method`, `path`, and `body` — but `MagicMock` silently provides every other attribute too. When the production code does:

```python
nuclio_fields = getattr(event, "fields", None)
if nuclio_fields:
    for key, values in nuclio_fields.items():
        ...
```

`getattr(event, "fields", None)` does NOT return `None`. Since `event` is a `MagicMock`, accessing `.fields` returns an auto-created `MagicMock` object, which is truthy. So the code enters the `if nuclio_fields:` branch and calls `.items()` on a `MagicMock`. The test passes, but it's exercising the Nuclio query-params parsing path instead of the no-query-params path that a real event without query parameters would take.
Same goes for MockEvent. It's still worse than nuclio_sdk.Event because it's missing attributes the real event has. For example `MockEvent(body={"data": "test"}, method="GET", path="/api/resource/42") ` has no `fields` attribute at all

`nuclio_sdk.Event` is a plain data class with zero setup cost — no I/O, no external dependencies — and it's exactly what Nuclio passes to the handler in production with `full_event=True`. Its `fields` attribute defaults to `{}` (falsy), so the code correctly skips the branch:

```python
event = nuclio_sdk.Event(body={"payload": {"data": [1, 2]}}, method="POST", path="/predict")
result = step.do(event)
```

alxtkr77 (changes requested):

mlrun/serving/api_handler.py
gtopper: We want to remove the dependency on nuclio_sdk (once [NUC-497](https://iguazio.atlassian.net/browse/NUC-497) is done), so this isn't relevant.

Furthermore, this PR only removes uses of `MagicMock`, it doesn't add any.

gtopper:
> ### Architectural concern: the weakref-based commit mechanism is fragile
> ...
> This doesn't need to block the current PR, but it should be tracked.

* I don't think a bug fix is the place for an architectural discussion.
* The current architecture assigns the responsibility to the step to keep a reference to the event until it's safe to commit, and not beyond that. Other architectures have their own tradeoffs. Explicit signaling would fail if the event is filtered out, and would require all terminal steps (which can be any step, not just targets) to implement the signaling.
* Feel free to open a ticket for it if you think this is actually something we should reconsider.

mlrun/serving/api_handler.py
alxtkr77: I am saying that abusing mocking is not a very good practice. If you can use a real object - use it. Also - if you are going to get rid of the nuclio_sdk something else will be introduced instead. And you want the tests to fail. Otherwise how you will know that they are still valid? 
As for your remark for 'only removing MagicMock ' - be good :) Fix other's problems when you see it

gtopper: * We're removing the mlrun _dependency_ on `nuclio_sdk`, not the library itself.
* Events are actually duck typed by design. The event does not need to be a `nuclio_sdk.Event`. It can also, for example, be a `storey.Event`.
* Are you asking that I change the rest of the instances where `MagicMock` is used and replace them with `MockEvent`? I'd suggest we can do that in an additional PR if you like.

alxtkr77:
> > ### Architectural concern: the weakref-based commit mechanism is fragile
> > ...
> > This doesn't need to block the current PR, but it should be tracked.
> 
> * I don't think a bug fix is the place for an architectural discussion.
> * The current architecture assigns the responsibility to the step to keep a reference to the event until it's safe to commit, and not beyond that. Other architectures have their own tradeoffs. Explicit signaling would fail if the event is filtered out, and would require all terminal steps (which can be any step, not just targets) to implement the signaling.
> * Feel free to open a ticket for it if you think this is actually something we should reconsider.

alxtkr77:
Opened https://iguazio.atlassian.net/browse/ML-12224

assaf758:
Great catch Gal!
Few preliminary comments:
1. Please use the .github/pull_request_template.md
2. Please run Smoke test workflow
3. I believe there is system-tests for api-handler, Jonathan can recommend the system-tests to run?

alxtkr77 (approved):

gtopper:
@assaf758 Smoke passed – https://github.com/mlrun/mlrun/actions/runs/22617698080.

gtopper:
@assaf758 all tests in `TestServingAPIHandler` passed except for `test_api_handler_forbidden`, which is a preexisting issue (broken in dev) that I fix in #9397.

tests/serving/test_api_handler.py
```diff
@@ -1675,186 +1675,109 @@ def test_run_path_template_method_not_allowed(self) -> None:
...
         step = _APIHandlerStep(config=config)
-        step.context = context
+        event = MockEvent(body={"data": "test"}, path="/test")
 
         with pytest.raises(
             mlrun.errors.MLRunBadRequestError, match="HTTP method not found"
         ):
-            step.do({"data": "test"})
-
-    def test_run_no_path_in_context(self) -> None:
```
jond01, line 1786: Please convert the test instead of removing it

jond01 (commented):
LGTM, except for the comment

tests/serving/test_api_handler.py
gtopper: Thanks, done.

assaf758 (approved):

PR #9393, opened 2026-03-02
[KFP] Enhance pipeline run deletion with retry logic for transient KFP errors

liranbg:
### 📝 Description

Fixes pipeline run deletion failures caused by transient Kubeflow Pipelines (KFP) errors (e.g. "invalid connection" from KFP's internal DB pool) during project deletion. Adds retry logic when listing pipelines and corrects a bug where the experiment-deletion loop waited on the wrong futures.

---

### 🛠️ Changes Made

- **Retry logic for pipeline listing**  
  In `delete_pipelines_runs`, the call to `list_pipelines` is now wrapped with `mlrun.utils.helpers.retry_until_successful` (2s backoff, 60s timeout, verbose logging) so transient KFP connection/API errors are retried instead of failing the entire project deletion. `MLRunInvalidArgumentError` is treated as fatal and not retried.

- **Bugfix: experiment deletion loop**  
  The loop that processes experiment-deletion futures was incorrectly using `delete_run_futures` in `concurrent.futures.as_completed(...)`. It now correctly uses `delete_experiment_futures`, so we wait on and handle the experiment-deletion futures rather than the pipeline-run-deletion futures.

---

### ✅ Checklist

---

### 🧪 Testing

- UT

---

### 🔗 References

- Ticket link: https://iguazio.atlassian.net/browse/ML-12062
- Design docs links: —
- External links: —

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes

- Retry parameters (backoff=2s, timeout=60s) are chosen to tolerate short-lived KFP connection issues without blocking project deletion for too long.
- `fatal_exceptions=(mlrun.errors.MLRunInvalidArgumentError,)` ensures invalid-argument errors are not retried.

### DIFF
--- a/server/py/services/api/crud/pipelines.py
+++ b/server/py/services/api/crud/pipelines.py
@@ -127,7 +127,15 @@ class Pipelines(
     def delete_pipelines_runs(
         self, db_session: sqlalchemy.orm.Session, project_name: str
     ):
-        _, _, project_pipeline_runs = self.list_pipelines(
+        # Retry listing pipelines to handle transient KFP connection errors
+        # (e.g. "invalid connection" from KFP's internal DB pool)
+        _, _, project_pipeline_runs = mlrun.utils.helpers.retry_until_successful(
+            backoff=2,
+            timeout=60,
+            logger=mlrun.utils.logger,
+            verbose=True,
+            _function=self.list_pipelines,
+            fatal_exceptions=(mlrun.errors.MLRunInvalidArgumentError,),
             db_session=db_session,
             project=project_name,
             format_=mlrun.common.formatters.PipelineFormat.metadata_only,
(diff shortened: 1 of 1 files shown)


elbamit:
Isn't timeout=60s for listing pipelines will fail when deleting a project that has many pipelines?

quaark (approved):

quaark:
> Isn't timeout=60s for listing pipelines will fail when deleting a project that has many pipelines?

@elbamit the retry here is to handle connection issues which terminate the request immediately, and the timeout only decides whether to retry once more or fail if the request failed. It won't stop a successful request that takes longer than 60s.
If we thought there's a case where the request takes a long time and finally fails, then yes we might want a longer timeout here, though this isn't that case.

PR #9396, opened 2026-03-03
[Auth] Skip invalid tokens during sync

elbamit:
### 📝 Description

When igz.yml contains multiple tokens and one is invalid (e.g. malformed JWT), the entire import mlrun flow would fail because it tries to sync and store all the tokens, but `extract_and_validate_tokens_info` raises an error on the invalid token.

This PR adds a parameter that allows skipping the error raising and just logging a warning on the invalid token.

---

### 🛠️ Changes Made

- Add a `skip_invalid` parameter to `extract_and_validate_tokens_info` that logs a warning and continues past invalid tokens.

---

### ✅ Checklist

---

### 🧪 Testing
  
  
- UT + manual test

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11974
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [ ] No

---


rokatyy (commented):
LGTM, just one question - would not we want to fail of all the tokens haven't passed a validation?

mlrun/auth/utils.py
```diff
@@ -288,44 +293,55 @@ def extract_and_validate_tokens_info(
...
-                    "Stored tokens can only belong to the authenticated user."
-                )
-
-            # Store token info
-            token_values[secret_token.name] = {
-                "token_exp": decoded_token.get(Claims.EXPIRATION),
-                "token": secret_token.token,
-            }
+                raise
         else:
```
Yacouby, line 346: this else branch outside the try/except, so it still raises even when skip_invalid=True. ss that intentional? If so, might be worth a comment explaining why name errors are treated differently

mlrun/auth/utils.py
```diff
@@ -263,6 +263,7 @@ def load_and_prepare_secret_tokens(
         ],
         authenticated_id=auth_user_id,
         filter_by_authenticated_id=True,
+        skip_invalid=True,
```
Yacouby: skip_invalid=True is hardcoded — so even when raise_on_error=True, token validation errors are silently skipped. Should skip_invalid be exposed as a parameter here to keep the behavior consistent?

Yacouby (commented):
minor questions

elbamit:
> LGTM, just one question - would not we want to fail of all the tokens haven't passed a validation?

@rokatyy In httpdb `connect`, 2 operations happen sequentially:
- self._init_token_provider()
- mlrun.secrets.sync_secret_tokens()
https://github.com/mlrun/mlrun/blob/fb4df91992f55ae1846ee0292d39b8afa3d47826/mlrun/db/httpdb.py#L693-L697

If all tokens are malformed, then the _init_token_provider() fails before even reaching the syncing part, where I made those changes.

There is a rare race condition where _init_token_provider() passes, but before sync_secret_tokens() runs, the file changes (or the token that succeeded in _init_token_provider() is now expired). I'll add a failure for such case just to be on the safe side.

mlrun/auth/utils.py
elbamit: Not intentional. Fixing it

Yacouby (approved):

mlrun/auth/utils.py
elbamit: skip_invalid is actually kind of the opposite of raise_on_error. I'll tie them together to maintain the same behavior

mlrun/auth/utils.py
```diff
@@ -288,48 +294,67 @@ def extract_and_validate_tokens_info(
...
+                message = (
+                    f"Duplicate token name '{secret_token.name}' found in request payload, "
+                    "only first occurrence is synced to the backend."
+                )
+            else:
+                message = "Token with invalid name found in request payload"
+            if skip_invalid:
+                mlrun.utils.logger.warning(message)
+                continue
+            raise mlrun.errors.MLRunInvalidArgumentError(message)
```
rokatyy: use `raise_or_log_error` from utils

rokatyy (approved):

liranbg (approved):

PR #9400, opened 2026-03-03
[Application] Add workdir support for pull-at-runtime

Yacouby:
### 📝 Description
Add workdir support for the pull-at-runtime flow in ApplicationRuntime. 
Previously, `spec.workdir` was only applied during the build-time flow. 
In the pull-at-runtime flow, the sidecar's workingDir always defaulted to target_dir, ignoring workdir entirely. 

---

### 🛠️ Changes Made
- Apply `spec.workdir` to the sidecar's `workingDir` and `PYTHONPATH`. Supports both relative (joined with target_dir) and absolute paths.
- Restore `build.source` after `from_image` when `load_source_on_run` is enabled, so the server can still configure the init container for runtime source loading (since `from_image` clears `build.source` to prevent Nuclio from rebuilding).
- Correct artifact versioning description in the app runtime doc NB and document explicit `store://` URI usage.

---

### ✅ Checklist

---

### 🧪 Testing
- Add parametrized unit test for workdir resolution in init container configuration. 
- Add workdir validation to git source system test. 
- Use `build_pod` instead of `application_image` to verify rebuilds in the system tests since application_image  stays the same across rebuilds while `build_pod` changes to a new Kaniko pod name each time a build is triggered.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-12213
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


docs/runtimes/application.ipynb
```diff
@@ -1,843 +1,844 @@
...
+      "mimetype": "text/x-python",
+      "name": "python",
+      "nbconvert_exporter": "python",
+      "pygments_lexer": "ipython3",
+      "version": "3.9.13"
+    }
+  },
+  "nbformat": 4,
+  "nbformat_minor": 4
+}
```
rokatyy: endline

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -1021,6 +1021,11 @@ def _configure_application_sidecar(self):
         ):
             self.from_image(reverse_proxy_image)
 
+            # from_image() clears build.source to prevent Nuclio from rebuilding from it, but the server needs
+            # build.source to configure the init container for runtime source loading
+            if self.spec.build.load_source_on_run and self.status.application_source:
+                self.spec.build.source = self.status.application_source
+
```
rokatyy: won't it break other flows?

docs/runtimes/application.ipynb
```diff
@@ -1,843 +1,844 @@
...
+      "codemirror_mode": {
+        "name": "ipython",
+        "version": 3
+      },
+      "file_extension": ".py",
+      "mimetype": "text/x-python",
+      "name": "python",
+      "nbconvert_exporter": "python",
+      "pygments_lexer": "ipython3",
+      "version": "3.9.13"
```
rokatyy, line 839: unrelated, but why is it py 3.9?

rokatyy (commented):
Very nice findings, the only concern I have is about BC wrt `source` reassignment

mlrun/runtimes/nuclio/application/application.py
Yacouby: No, this is guarded by `self.spec.build.load_source_on_run`, which is only True in the pull-at-runtime flow. In all other flows (build-time, from-image without pull-at-runtime), `load_source_on_run` is False/None, so this block is skipped entirely.

docs/runtimes/application.ipynb
Yacouby: mm not sure, this is pre-existing notebook metadata and not part of my changes. ( so when i go to the NB cell I do not see this version at all )

mlrun/runtimes/nuclio/application/application.py
TomerShor: @Yacouby so if `load_source_on_run==True` we won't skip the image building (since `build.source` won't be empty)?
If so that's not ideal, since we still want to skip the build

Yacouby: nuclio image:
my restore puts `build.source` back, but `functionSourceCode` stays empty.
Nuclio needs both to trigger a rebuild , with no `functionSourceCode`, there's nothing to compile, so it skips regardless of `build.source`

and regarding the application image:
the restore only happens after the sidecar build step, and it's only used by the server to configure the init container. so we are safe also here

server/py/services/api/tests/unit/runtimes/test_application.py
```diff
@@ -268,6 +268,56 @@ def test_configure_init_container_idempotent(self, db: Session, client: TestClie
             f"{mlrun.common.constants.DEFAULT_SOURCE_CODE_TARGET_DIR}:/user/custom/path"
         )
 
+    @pytest.mark.parametrize(
+        "workdir,expected_workdir,expected_pythonpath",
```
liranbg: so if `expected_workdir` always is `expected_pythonpath` you may just remove it and assert on test that output
but if you see a test case where they are different, add it here.

liranbg (commented):
lgtm. minor test comment

mlrun/runtimes/nuclio/application/application.py
Yacouby: image

rokatyy: after looking at the code again, maybe we can just remove the clean up of the source at all? And then we can remove the duplicate here as well https://github.com/mlrun/mlrun/blob/560e0a8e889d451efb60f00609026be6ed2e12ec/mlrun/runtimes/nuclio/application/application.py#L576-L577

Yacouby: removing it may be risky as it would break existing flow, it should be empty when the function reaches the server, otherwise `compile_nuclio_archive_config` sets Nuclio's codeEntryType (archive/git) which could trigger a rebuild even though we want to skip it.

Instead, I moved the logic to the server side and now fall back to `status.application_source` when `build.source` is empty. This way `build.source` stays cleared for Nuclio, but the init container still gets the correct source URI.

rokatyy (approved):
Nice!

TomerShor (approved):

PR #9404, opened 2026-03-04
[Nuclio] Fix JSONDecodeError when invoking function with HEAD method

jond01:
## Summary

- **Bug ([ML-12228](https://iguazio.atlassian.net/browse/ML-12228)):** `RemoteRuntime.invoke()` raised `JSONDecodeError` for any `method="HEAD"` call against a deployed Nuclio function
- **Root cause:** HTTP HEAD responses carry no body (RFC 9110 §9.3.2); Nuclio correctly strips it but keeps `Content-Type: application/json`. The previous code unconditionally called `json.loads(resp.content)`, crashing on the empty `b""`
- **Fix:** Guard `json.loads` with an emptiness check — `if data and resp.headers["content-type"] == "application/json"`

## Changes

- `mlrun/runtimes/nuclio/function.py` — one-line fix in `RemoteRuntime.invoke()`
- `tests/serving/test_remote.py` — three unit tests (mocked HTTP session, no deployment needed)
- `tests/system/runtimes/test_nuclio.py` — system regression test: deploys a plain `kind="nuclio"` function, asserts POST returns parsed JSON and HEAD returns `b""` without raising
- `tests/system/runtimes/assets/echo_handler.py` — new asset: plain Nuclio handler that echoes the request body as JSON (used by the system test)

## Test plan

- [x] Run unit tests: `pytest tests/serving/test_remote.py::test_invoke_head_returns_empty_bytes tests/serving/test_remote.py::test_invoke_post_returns_parsed_json tests/serving/test_remote.py::test_invoke_non_json_content_type_returns_bytes`
- [x] Run system test against a live cluster: `pytest tests/system/runtimes/test_nuclio.py::TestNuclioRuntime::test_invoke_head_method_does_not_raise`

[ML-12228]: https://iguazio.atlassian.net/browse/ML-12228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/serving/test_remote.py
+++ b/tests/serving/test_remote.py
@@ -466,15 +466,52 @@ def test_remote_function_step(rundb_mock, httpserver, engine):
 
 
 # ---------------------------------------------------------------------------
-# ML-12228 – RemoteRuntime.invoke() with HEAD method
+# RemoteRuntime.invoke() with HEAD method
 # ---------------------------------------------------------------------------
 
 
-def _make_nuclio_fn(address: str) -> mlrun.runtimes.RemoteRuntime:
-    fn = cast(
-        mlrun.runtimes.RemoteRuntime, mlrun.new_function("test-ml12228", kind="nuclio")
+def test_invoke_head_returns_empty_bytes() -> None:
+    """HEAD responses have no body; invoke() must return b"" without raising."""
+    fn = _make_nuclio_fn()
+    fn._http_session = MagicMock()
+    fn._http_session.request.return_value = _mock_response(b"", "application/json")
+
+    result = fn.invoke("/", method="HEAD")
+    assert result == b""
+
+
+def test_invoke_post_returns_parsed_json() -> None:
+    """Non-empty JSON body is parsed correctly."""
+    fn = _make_nuclio_fn()
+    fn._http_session = MagicMock()
+    fn._http_session.request.return_value = _mock_response(
+        b'{"echo": "hello"}', "application/json"
     )
-    fn.status.address = address
+
+    result = fn.invoke("/", body={"data": "hello"})
+    assert result == {"echo": "hello"}
+
+
+def test_invoke_non_json_content_type_returns_bytes() -> None:
+    """Responses with non-JSON content-type are returned as raw bytes."""
+    fn = _make_nuclio_fn()
+    fn._http_session = MagicMock()
+    fn._http_session.request.return_value = _mock_response(b"plain text", "text/plain")
+
+    result = fn.invoke("/", method="GET")
+    assert result == b"plain text"
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _make_nuclio_fn() -> mlrun.runtimes.RemoteRuntime:
+    fn = cast(
+        mlrun.runtimes.RemoteRuntime, mlrun.new_function("test-nuclio", kind="nuclio")
+    )
+    fn.status.address = "http://localhost:8080"
     return fn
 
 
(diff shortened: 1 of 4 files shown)


tests/system/runtimes/test_nuclio.py
```diff
@@ -881,6 +881,43 @@ def test_async_http_mode_serving_graph(self, with_code):
...
+        self._logger.debug("Deploying function")
+        function.deploy()
+
+        # POST: body present, JSON parsed normally.
+        self._logger.debug("Invoking POST – should return parsed JSON")
+        result = function.invoke("/", body={"data": "ml12228"})
+        assert result == {"echo": {"data": "ml12228"}}
+
+        # HEAD: body stripped by HTTP; invoke() must return b"" without raising.
+        self._logger.debug("Invoking HEAD – must not raise (ML-12228 fix)")
```
TomerShor: All this `ML-12228` in the comments and logs is redundant. We get it - there was a bug.
```suggestion
        self._logger.debug("Invoking HEAD – must not raise")
```

tests/serving/test_remote.py
```diff
@@ -460,3 +463,62 @@ def test_remote_function_step(rundb_mock, httpserver, engine):
...
+# ---------------------------------------------------------------------------
+# ML-12228 – RemoteRuntime.invoke() with HEAD method
+# ---------------------------------------------------------------------------
+
+
+def _make_nuclio_fn(address: str) -> mlrun.runtimes.RemoteRuntime:
+    fn = cast(
+        mlrun.runtimes.RemoteRuntime, mlrun.new_function("test-ml12228", kind="nuclio")
+    )
+    fn.status.address = address
```
TomerShor: Why is the address needed if you anyway mock everything?

tests/serving/test_remote.py
```diff
@@ -460,3 +463,62 @@ def test_remote_function_step(rundb_mock, httpserver, engine):
...
+    return fn
+
+
+def _mock_response(
+    content: bytes,
+    content_type: str,
+    status: HTTPStatus = HTTPStatus.OK,
+) -> requests.Response:
```
TomerShor: Move private functions below public tests

TomerShor (changes requested):

tests/system/runtimes/test_nuclio.py
jond01: Done — removed all ML-12228 references from logs, test data, and the function name.

tests/serving/test_remote.py
jond01: The address is still needed — even though the HTTP session is mocked, `_resolve_invocation_url` runs first to build the URL. Without an address (and no `internal_invocation_urls` / `external_invocation_urls`), it calls `_get_state()` to fetch state from the API server, which would fail in a unit test. Kept the address hardcoded inside the helper instead of passing it as a parameter.

tests/serving/test_remote.py
jond01: Done — moved both helpers below the public tests.

tests/serving/test_remote.py
```diff
@@ -460,3 +463,67 @@ def test_remote_function_step(rundb_mock, httpserver, engine):
...
+
+    result = fn.invoke("/", method="HEAD")
+    assert result == b""
+
+
+def test_invoke_post_returns_parsed_json() -> None:
+    """Non-empty JSON body is parsed correctly."""
+    fn = _make_nuclio_fn()
+    fn._http_session = MagicMock()
+    fn._http_session.request.return_value = _mock_response(
```
TomerShor: This is repeated in all new tests - You can refactor this setup using a pytest fixture, which provides reusable test resources.
e.g.:
```python
@pytest.fixture
def mocked_nuclio_fn():
    fn = _make_nuclio_fn()
    fn._http_session = MagicMock()
    return fn

def test_invoke_post_returns_parsed_json(mocked_nuclio_fn):
    ...
```

TomerShor (commented):
Last comment

tests/serving/test_remote.py
jond01: Done — introduced a `nuclio_fn` fixture for the repeated setup. `mock_response` is kept as a plain module-level function since it takes parameters (a factory fixture would require a nested callable).

TomerShor (approved):
LGTM.
I suggest adding some rules to your Claude, to align with the repo's best practices and style (like the previous comments, favoring explicit variables (`func` or `function` over `fn`), and the rest of our conventions

assaf758 (approved):

PR #9410, opened 2026-03-05
[Serving] Block custom handler usage in jobs from serving functions

royischoss:
### 📝 Description

Prevent specifying a custom handler when executing jobs that originate from serving functions while using `project.run_function`.

---

### 🛠️ Changes Made

 Blocking user handler for running a serving origin job.
---

### ✅ Checklist
- [x] Please run [Smoke-tests](https://github.com/mlrun/mlrun/actions/workflows/system-tests-opensource.yml) workflow, providing it the PR number as input (upon success, the workflow run will add label "Smoke tests: Pass" to the RP)

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-12215
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [ ] No

---

### DIFF
--- a/mlrun/projects/operations.py
+++ b/mlrun/projects/operations.py
@@ -187,6 +187,13 @@ def run_function(
                             If not provided, the default backoff delay is 30 seconds.
     :return: MLRun RunObject or PipelineNodeWrapper
     """
+    if (
+        isinstance(function, mlrun.runtimes.KubejobRuntime) and function.serving_spec
+    ) and handler is not None:
+        raise mlrun.errors.MLRunInvalidArgumentError(
+            "handler cannot be specified when running a KubeJobRuntime with a serving spec"
+        )
+
     if artifact_path:
         warnings.warn(
             "'artifact_path' parameter is deprecated in 1.10.0 and will be removed in 1.12.0, "
--- a/mlrun/projects/operations.py
+++ b/mlrun/projects/operations.py
@@ -188,8 +188,10 @@ def run_function(
     :return: MLRun RunObject or PipelineNodeWrapper
     """
     if (
-        isinstance(function, mlrun.runtimes.KubejobRuntime) and function.serving_spec
-    ) and handler is not None:
+        isinstance(function, mlrun.runtimes.KubejobRuntime)
+        and function.serving_spec
+        and handler is not None
+    ):
         raise mlrun.errors.MLRunInvalidArgumentError(
             "handler cannot be specified when running a KubeJobRuntime with a serving spec"
         )
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -4187,8 +4187,8 @@ class MlrunProject(ModelObj):
             )
 
         :param function:        name of the function (in the project) or function object
-        :param handler:         name of the function handler, for serving function/job handler should not be provided as
-                                the default handler is defined in the serving spec.
+        :param handler:         name of the function handler. should not be set for serving graph deployed as job,
+                                as in this case the handler is predefined
         :param name:            execution name
         :param params:          input parameters (dict)
         :param hyperparams:     hyper parameters
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -4187,7 +4187,8 @@ class MlrunProject(ModelObj):
             )
 
         :param function:        name of the function (in the project) or function object
-        :param handler:         name of the function handler
+        :param handler:         name of the function handler, for serving functions handler should not be provided as
+                                the default handler is defined in the serving spec.
         :param name:            execution name
         :param params:          input parameters (dict)
         :param hyperparams:     hyper parameters
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -4251,6 +4251,14 @@ class MlrunProject(ModelObj):
             )
         output_path = output_path or artifact_path
 
+        if (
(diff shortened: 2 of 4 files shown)


mlrun/serving/states.py
```diff
@@ -1900,9 +1897,11 @@ def select_outlets(self, event) -> Collection[str] | None:
         #  fall to default behavior of routing to all valid outlets
         all_outlets = list(self._name_to_outlet.keys())
         if is_batched:
-            all_outlets.remove(f"{self.name}_error_raise")
+            if (error_raiser := f"{self.name}_error_raise") in all_outlets:
+                all_outlets.remove(error_raiser)
         else:
-            all_outlets.remove(f"{self.name}_unpacker")
+            if (unpacker := f"{self.name}_unpacker") in all_outlets:
+                all_outlets.remove(unpacker)
```
yaelgen: Just to make sure, If `_error_raise` or `_unpacker` is missing from outlets, does silently skipping the removal produce the correct routing behavior?

mlrun/projects/project.py
```diff
@@ -4251,6 +4252,14 @@ def run_function(
...
         output_path = output_path or artifact_path
 
+        if (
+            isinstance(function, mlrun.runtimes.KubejobRuntime)
+            and function.serving_spec
+        ) and handler is not None:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                "handler cannot be specified when running a KubeJobRuntime with a serving spec"
+            )
+
```
yaelgen: This check only guards `project.run_function()`, I'm not sure if this is something that's being done, but users can also reach the same code path via `mlrun.run_function()` (operations.py) or fn.run() directly - both bypass this validation. 
Consider moving the check into `operations.run_function()` so it covers at least both entry points.

yaelgen (changes requested):
Nice! added a question and a suggestion

mlrun/serving/states.py
royischoss: This can silently produce wrong results, but this code runs in storey and most times the error might not be raised, or raised without the real context or without the real code line, so we rather see it fails in mlrun.

royischoss: removing it

yaelgen (approved):
🚀

mlrun/projects/operations.py
```diff
@@ -187,6 +187,13 @@ def my_pipe(url=""):
                             If not provided, the default backoff delay is 30 seconds.
     :return: MLRun RunObject or PipelineNodeWrapper
     """
+    if (
```
assaf758, line 190: The parentheses are redundant

mlrun/projects/project.py
```diff
@@ -4187,7 +4187,8 @@ def run_function(
             )
 
         :param function:        name of the function (in the project) or function object
-        :param handler:         name of the function handler
+        :param handler:         name of the function handler, for serving function/job handler should not be provided as
+                                the default handler is defined in the serving spec.
```
assaf758:
```suggestion
        :param handler:         name of the function handler. should not be set for serving graph deployed as job, as in this case the handler is predefined 
```

tests/runtimes/test_to_job.py
```diff
@@ -272,3 +272,52 @@ def test_to_job_backward_compatibility():
...
+    "use_operations_run_function",
+    [False, True],
+    ids=["project.run_function", "operations.run_function"],
+)
+def test_run_function_with_job_from_serving_fails_with_handler(
+    use_operations_run_function,
+):
+    """Test that run_function raises error when handler is specified for a job from serving function.
+
+    When a serving function is converted to a job via to_job(), the job retains a serving_spec.
```
assaf758:
```suggestion
    When a serving function is deployed to a job via to_job(), the job containes a serving_spec.
```

tests/runtimes/test_to_job.py
```diff
@@ -272,3 +272,52 @@ def test_to_job_backward_compatibility():
...
+
+    # Create a serving function
+    serving_fn = mlrun.new_function(name="test-serving", kind="serving")
+    serving_fn.spec.image = "mlrun/mlrun"  # Set an image to avoid build requirements
+
+    # Set up a simple graph
+    graph = serving_fn.set_topology("flow", engine="async")
+    graph.to(name="step1", handler="handler")
+
+    # Convert to job - this creates a KubejobRuntime with serving_spec
```
assaf758: Convert => Deploy

tests/runtimes/test_to_job.py
```diff
@@ -272,3 +272,52 @@ def test_to_job_backward_compatibility():
...
+
+    When a serving function is converted to a job via to_job(), the job retains a serving_spec.
+    Running such a job with a custom handler should fail because the serving spec already
+    defines the default handler (execute_graph).
+
+    This test validates both:
+    - project.run_function (use_operations_run_function=False)
+    - operations.run_function directly (use_operations_run_function=True)
+    """
+    from mlrun.projects.operations import run_function as operations_run_function
```
assaf758: more a matter of style - but consider moving to just above operations_run_function, such that it's executed just on the "operation" parametrization.

assaf758 (commented):
Looks good.
minor comments inline.

assaf758 (approved):

PR #9418, opened 2026-03-08
[Project] Add mgmt permission check for project owner changes

Yacouby:
### 📝 Description
This PR Fixes a permission gap where patching a project's owner in MLRun only checked `/resources/projects/{project} `, while Orca expects `/mgmt/projects/{project}/owner`. 
This caused System Admin users who have full `/mgmt/**` permissions but not `/resources/projects/{project}` to fail when changing project owners on IG4.

Additionally closes a secondary gap: users with regular writer privileges on a project could change the owner without the required management-level permission.
The patch-project endpoint now performs fine-grained permission checks:
If the patch modifies `spec.owner`, check `/mgmt/projects/{project}/owner` (update)
If the patch modifies any other fields, check` /resources/projects/{project} `(update)
If both are modified, both checks must pass

---

### 🛠️ Changes Made
- Added `project_owner` to `AuthorizationResourceTypes` enum
- Extracted `_verify_patch_project_permissions` to replace the single permission check in `patch_project`. Scoped the fine-grained check to IG4 only, preserving original IG3 behavior.

---

### ✅ Checklist

---

### 🧪 Testing
Added unit tests verifying that the correct permission check (mgmt-level or regular) is invoked based on the patch contents, and that access-denied errors propagate correctly.

- IG3 BC Validation (tested on vmdev26):
Verified that the patch-project permission behavior is unchanged after deploying the code changes. 

- IG4 Validation (tested on vmdev223ig4):
Tested with a project owned by normal-user and admin (with System Admin + Developer policies).

Before the fix, admin patching the project owner returned:
`MLRunAccessDeniedError: Failed patching project bc-test-ig4 details: MLRunAccessDeniedError('Not allowed to update resource /resources/projects/bc-test-ig4')`

After the fix, admin can successfully patch the owner. 

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-12255
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


server/py/services/api/api/endpoints/projects.py
```diff
@@ -560,6 +556,60 @@ async def load_project(
...
+            project_name,
+            "",
+            mlrun.common.schemas.AuthorizationAction.update,
+            auth_info,
+            resource_namespace=mlrun.common.schemas.AuthorizationResourceNamespace.mgmt,
+        )
+
+    # Run the regular resource-level check when non-owner fields are touched, or as a fallback when the patch doesn't
+    # modify the owner either (e.g. empty body), so that no request bypasses permission verification entirely.
+    if modifies_other or not modifies_owner:
```
TomerShor, line 594: What's use case of having `modifies_other=False and modifies_owner=False`? do we allow patching without body?

server/py/services/api/tests/unit/api/test_projects.py
```diff
@@ -234,6 +234,81 @@ async def test_project_permissions_update_when_exists(
...
 
 
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "patch_body, expect_mgmt_check, expect_regular_check",
+    [
+        ({"spec": {"owner": "new_owner"}}, True, False),
+        ({"spec": {"description": "updated"}}, False, True),
+        ({"spec": {"owner": "new_owner", "description": "updated"}}, True, True),
+        ({}, False, True),
```
TomerShor, line 245: add a case for a non-empty patch_body with something other than spec

TomerShor (commented):
Nicely done! minor comments

server/py/services/api/api/endpoints/projects.py
Yacouby: the endpoint accepts `project: dict` with no upstream validation, so an empty {} body is technically valid and will reach the permission check. This fallback ensures that even an empty patch still goes through the regular project-update permission check, so no request bypasses authorization entirely.

TomerShor (approved):

PR #9425, opened 2026-03-09
[Serving] Fix _MappedBody unpacking bypassed in async engine

jond01:
## Summary

Fixes [ML-12217](https://iguazio.atlassian.net/browse/ML-12217): `AttributeError: '_RequestContext' object has no attribute 'split'` when using `body_map` with a generator step and `engine="async"`.

- `_init_async_objects()` wraps user step handlers in `storey.Map`, which calls `handler(event.body)` as a positional arg, bypassing `TaskStep.run()` and its `_MappedBody → **kwargs` unpacking
- `_APIHandlerStep.do()` sets `event.body` to `_RequestContext` (a `_MappedBody`/`dict` subclass) when `body_map` is configured — this arrived in the handler as a dict instead of being spread as `**kwargs`, causing `AttributeError` on the first string operation
- The downstream `async_generator is not JSON serializable` error in Nuclio was a symptom: the generator body never executed because the handler received a dict instead of a string

**Fix:** extend `_MappedBodyAwareHandler` with an `async_call` method for the async engine path:
- `TaskStep.run()` (sync engine) uses `__call__` as before — picklable, works with storey's `max_processes` multiprocessing
- `_init_async_objects()` passes the bound `async_call` method to `storey.Map` — storey's `asyncio.iscoroutinefunction` check naturally returns `True` for a bound `async def` method, with no private-API sentinels needed

## Test plan

- [x] `TestBodyMapMappedBodyUnpacking::test_sync_engine_unpacks_body_map` — verifies `engine="sync"` path
- [x] `TestBodyMapMappedBodyUnpacking::test_async_engine_unpacks_body_map` — verifies `engine="async"` path (the regression)
- [x] `test_configure_model_runner_step_max_threads_processes[max_processes]` — verifies picklability is preserved
- [x] System test `test_api_handler_with_body_mapping` now parametrized over both engines

[ML-12237]: https://iguazio.atlassian.net/browse/ML-12237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ML-12217]: https://iguazio.atlassian.net/browse/ML-12217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/serving/states.py
+++ b/mlrun/serving/states.py
@@ -937,12 +937,7 @@ class TaskStep(BaseStep):
                 )
 
             body = _extract_input_data(self.input_path, event.body)
-            if isinstance(body, _MappedBody):
-                # body_map-transformed bodies are unpacked as **kwargs
-                # so handler signatures like def fun(book: str) work
-                result = self._handler(**body, **kwargs)
-            else:
-                result = self._handler(body, *args, **kwargs)
+            result = _MappedBodyAwareHandler(self._handler)(body, *args, **kwargs)
             event.body = _update_result_body(self.result_path, event.body, result)
         except Exception as exc:
             if self._on_error_handler:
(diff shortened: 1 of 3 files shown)


mlrun/serving/states.py
```diff
@@ -4137,8 +4167,14 @@ def _init_async_objects(context, steps, root):
                         f"Step '{step.name}' does not have a handler that can be called"
                     )
                 # if regular class, wrap with storey Map
+                wrapped = _MappedBodyAwareHandler(step._handler)
+                if inspect.iscoroutinefunction(step._handler):
+                    # Pass the bound async method so asyncio.iscoroutinefunction detects it correctly
+                    handler = wrapped.async_call
```
royischoss, line 4173: We need to make sure that `test_async_http_mode_serving_graph` passes because this touches the async behavior.

mlrun/serving/states.py
```diff
@@ -4137,8 +4167,14 @@ def _init_async_objects(context, steps, root):
                         f"Step '{step.name}' does not have a handler that can be called"
                     )
                 # if regular class, wrap with storey Map
+                wrapped = _MappedBodyAwareHandler(step._handler)
+                if inspect.iscoroutinefunction(step._handler):
```
royischoss, line 4171: we might need to make this check a little less strict with `inspect.iscoroutine(obj)`

royischoss (commented):
Hey LGTM two small comments

mlrun/serving/states.py
jond01: From what I see [`inspect.iscoroutinefunction`](https://docs.python.org/3/library/inspect.html#inspect.iscoroutinefunction) is the more general check, is catches also coroutines wrapped in `functools.partial`

mlrun/serving/states.py
jond01: Passed for me ✅

assaf758 (approved):

mlrun/serving/states.py
royischoss: import inspect

async def foo():
    return 1

coro = foo()

print(inspect.iscoroutinefunction(foo))   # True
print(inspect.iscoroutinefunction(coro))  # False

PR #9432, opened 2026-03-10
[API] Fix OPA manifest distribution race condition

liranbg:
### 📝 Description

When MLRun runs as project-leader on multi-pod deployments, operations immediately after `create_project()` (e.g. `set_secrets`, `create_function`) can fail with 403 because OPA manifest propagation to other pods' sidecars is eventually consistent.

```
Flow:
1. Pod A handles create_project → tells Orca to create OPA policies → waits up to 10s for its own sidecar
2. SDK immediately calls set_secrets → load-balanced to Pod B → Pod B's OPA sidecar doesn't have the manifest yet → 403
```

This PR fixes the race by populating the existing OPA owner cache (`_allowed_project_owners_cache`) from the shared DB (source of truth) whenever `ensure_project` runs and the requesting user is the project owner. Pods that haven't received the OPA manifest yet can then allow the owner via cache (30s TTL) instead of returning 403.

Additionally, introduces `ProjectFormatCustomSelection` — a generic server-side mechanism for querying only specific project DB columns (bypassing full ORM load and pickle deserialization), enabling lightweight DB access for `ensure_project` and future flows.

---

### 🛠️ Changes Made

- **New `ProjectFormatCustomSelection` pipeline** (`server/py/framework/utils/project_formats.py`): Generic server-side format type that allows selecting specific project DB columns (`name`, `owner`, `created`), with a `build()` method that maps raw column values to a minimal `Project` schema. Extensible for future custom-field flows.
- **DB layer** (`server/py/framework/db/sqldb/db.py`): `list_projects` now handles `ProjectFormatCustomSelection` by querying only the requested columns directly (no ORM model load, no pickle deserialization).
- **`ensure_project`** (`server/py/framework/utils/projects/member.py`): Changed from `name_only` to `ProjectFormatCustomSelection(["name", "owner"])` format, and added OPA owner cache population when the requesting user matches the project owner.
- **`project_secrets` endpoints** (`project_secrets.py`): Replaced `get_project` calls with `ensure_project` to both verify project existence and populate the OPA owner cache.

---

### ✅ Checklist

---

### 🧪 Testing

Unit + manual patching and test

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11996
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes

- The fix mitigates the race for the **project owner only**. Non-owners may still see 403 on a pod that hasn't received the OPA manifest yet; that behavior is unchanged and acceptable for this ticket.
- `ProjectFormatCustomSelection` is designed to be generic and extensible — future flows that need lightweight project field access can reuse it by adding columns to `ProjectFormatCustom` and mapping them in `build()`.
- The nuclio and iguazio v4 client `list_projects` type hints still use `ProjectFormat` (not `ProjectFormatType`); they are never called with custom selection so this is cosmetic and can be updated separately.


server/py/framework/utils/projects/member.py
```diff
@@ -45,9 +46,11 @@ def ensure_project(
         auth_info: mlrun.common.schemas.AuthInfo = mlrun.common.schemas.AuthInfo(),
     ):
         try:
+            # Using minimal format to access spec.owner for OPA cache population
+            # while avoiding the overhead of fetching large fields (functions, workflows, artifacts)
             project = self.get_project(
                 db_session,
-                format_=mlrun.common.formatters.ProjectFormat.name_only,
+                format_=mlrun.common.formatters.ProjectFormat.minimal,
```
theSaarco: This does have some minor performance implications, since the `minimal` actually gets the full record from the DB, it just drops most of the fields before returning. 
Did you explore the logic before it got removed - how was this done using the `name_only` approach? Was there no validation that the current user is really the owner registered in the DB?

server/py/services/api/api/endpoints/project_secrets.py
```diff
@@ -38,14 +38,14 @@ async def store_project_secrets(
...
 ):
-    # Doing a specific check for project existence, because we want to return 404 in the case of a project not
-    # existing, rather than returning a permission error, as it misleads the user. We don't even care for return
-    # value.
+    # Using ensure_project instead of get_project to both verify project existence (returning 404 rather than
+    # a misleading permission error) and populate the OPA owner cache, mitigating the OPA manifest propagation
+    # race condition on multi-pod deployments.
     await run_in_threadpool(
-        framework.utils.singletons.project_member.get_project_member().get_project,
+        framework.utils.singletons.project_member.get_project_member().ensure_project,
```
theSaarco, line 45: Interesting - was this an oversight where all other endpoints moved to use `ensure` and only the project-secrets endpoint didn't? I see the original code was added in 2021 (by me 😄) and wasn't changed since.

theSaarco (approved):
Looks good - some clarification questions inside on the implementation.

server/py/framework/utils/projects/member.py
liranbg: the formatting introduced after it was removed.

server/py/services/api/api/endpoints/project_secrets.py
liranbg: I guess it was forgotten :) there are still more endpoints missing (see pr description under "Additional Notes")

TomerShor (approved):

server/py/services/api/api/endpoints/project_secrets.py
liranbg: solved by adding custom-wise columns to pull. see below commits

theSaarco (approved):
LGTM.

liranbg:
docs check failed - but passed manually on rtd @ https://app.readthedocs.org/projects/mlrun/builds/31839729/

PR #9454, opened 2026-03-16
[v3io] Fix v3io api usage as part of the url usage

tomerm-iguazio:
### 📝 Description
This PR improves how MLRun handles V3IO datastore configuration, so it can work correctly whether the V3IO endpoint is provided through the URL, environment variables/secrets, or a datastore profile. It also expands the system test coverage to validate these different configuration paths and ensure behavior stays consistent across scenarios.

---

### 🛠️ Changes Made
This PR extends V3IO datastore support to accept the V3IO API endpoint from an explicit datastore profile field and from the `V3IO_API` secret/env var, in addition to the existing config fallback.

It updates V3IO URL parsing/normalization (including adding https:// when needed) and adds a V3IO-specific as_df override to avoid double-appending the endpoint when reading dataframes.
System tests were expanded to cover both “endpoint embedded in the v3io://... URL” and “endpoint provided via profile/config” scenarios.

---

### ✅ Checklist

---

### 🧪 Testing
V3io system tests + paramatized feature store system test.

---

### 🔗 References
- Ticket link: [ML-12096](https://iguazio.atlassian.net/browse/ML-12096)
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [ ] No

---

### 🔍️ Additional Notes

[ML-12096]: https://iguazio.atlassian.net/browse/ML-12096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/datastore/v3io.py
```diff
@@ -49,6 +53,8 @@ def __init__(self, parent, schema, name, endpoint="", secrets: dict | None = Non
         elif self.endpoint.startswith("http://"):
             self.endpoint = self.endpoint[len("http://") :]
             self.secure = False
+        else:
```
alxtkr77, line 72: **Bug (Critical):** When the endpoint comes from `parse_url` (embedded in URL), the scheme is already stripped — e.g. `remote-host:8081` has no `http://` prefix. So this `else` branch fires unconditionally for URL-embedded endpoints and sets `secure=True`.

But `v3io://` means HTTP (vs `v3ios://` for HTTPS). Line 41 (`self.secure = self.kind == "v3ios"`) already correctly set `False`, and this `else` clobbers it.

Result: `v3io://remote-host:8081/path` → `https://remote-host:8081` instead of the expected `http://remote-host:8081`.

**Fix**: Remove the `else` branch entirely. The initial assignment on line 41 already handles this correctly.

mlrun/datastore/v3io.py
```diff
@@ -236,3 +242,33 @@ def rm(self, path, recursive=False, maxdepth=None):
...
+        subpath,
+        columns=None,
+        df_module=None,
+        format="",
+        start_time=None,
+        end_time=None,
+        time_column=None,
+        additional_filters=None,
+        **kwargs,
+    ):
```
alxtkr77, line 274: The `as_df` override exists because removing the `parse_url` normalization means v3io-fs now gets the endpoint in both `storage_options["v3io_api"]` and the URL's netloc — causing a double-append. I understand the intent.

However, `.replace(self.endpoint, "", 1)` does a bare substring search which is fragile if the hostname string appears elsewhere in the path. A safer structural approach:

```python
def as_df(self, url, subpath, **kwargs):
    if url:
        parsed = urlparse(url)
        if parsed.netloc:
            url = f"{parsed.scheme}://{parsed.path}"
    return super().as_df(url=url, subpath=subpath, **kwargs)
```

This removes the netloc structurally via `urlparse` rather than doing a substring search.

tests/datastore/test_datastores.py
```diff
@@ -306,14 +306,15 @@ def test_verify_data_stores_are_not_cached_in_api_when_not_needed():
     assert obj3._store._secrets == {}
 
 
-def test_verify_data_stores_are_cached_when_not_api():
+@pytest.mark.parametrize("system_endpoint", ["some-system", None])
+def test_verify_data_stores_are_cached_when_not_api(system_endpoint):
     user1_secrets = {"V3IO_ACCESS_KEY": "user1-access-key"}
-    user1_objpath = "v3io://some-system/some-dir/user1"
+    user1_objpath = f"v3io://{system_endpoint}/some-dir/user1"
```
alxtkr77, line 314: When `system_endpoint=None`, this f-string produces `"v3io://None/some-dir/user1"` — a URL with the literal hostname `"None"`, not a URL with no host.

The no-endpoint case should produce `"v3io:///some-dir/user1"` (triple slash). Needs conditional logic:

```python
user1_objpath = f"v3io://{system_endpoint}/some-dir/user1" if system_endpoint else "v3io:///some-dir/user1"
```

(Same for user2 and user3.)

alxtkr77:
## Review Summary

The core fix (removing the `v3io://` → `v3io:///` normalization in `parse_url`) is the right approach — it was the root cause preventing endpoint extraction from v3io URLs in the datastore layer. The `DatastoreProfileV3io.v3io_api` addition and `get_storage_options()` fix are also good.

### Remaining gaps in storey `V3ioDriver` instantiation

The PR fixes the `V3ioStore` datastore layer but the same problem persists in the storey `V3ioDriver` creation paths:

| Code path | Gap |
|---|---|
| `ResourceCache.get_table()` (store_resources.py:82) | No `access_key` passed to `V3ioDriver` — always falls back to `V3IO_ACCESS_KEY` env var |
| `StreamStoreyTarget` (storeytargets.py:125) | No `webapi` passed to `V3ioDriver` — always falls back to `V3IO_API` env var |

These paths are used by feature store online serving and stream targets, so a user embedding the endpoint in a `v3io://` URL or using a datastore profile will still hit the env var fallback there.

mlrun/datastore/utils.py
```diff
@@ -315,12 +315,9 @@ def valid_entries_only(self, input_dict: dict) -> dict:
 
 
 def parse_url(url):
```
alxtkr77, line 317: **Backward compatibility concern:** Removing this normalization changes the semantics of `v3io://something/path`.

Before: `v3io://mycontainer/data` → normalized to `v3io:///mycontainer/data` → `hostname=None`, `path=/mycontainer/data` → endpoint falls back to `mlconf.v3io_api`, path is `/mycontainer/data`.

After: `v3io://mycontainer/data` → `hostname=mycontainer`, `path=/data` → endpoint is `mycontainer`, path is `/data`.

Users who wrote `v3io://container/path` (double slash) expecting the default endpoint will now silently get `container` interpreted as a hostname. The correct form for no-endpoint is `v3io:///container/path` (triple slash), but existing code/configs may rely on the old double-slash behavior.

This is a **silent breakage** — no error raised, just wrong endpoint resolution. Consider adding a deprecation warning when `parse_url` detects a v3io URL with a netloc that looks like a container name (no dots/colons), or at minimum document this as a breaking change.

PR #9485, opened 2026-03-22
[Mounts] Support S3 and secret_env auto mount types in client-side auto_mount()

alxtkr77:
### 📝 Description
Extend `auto_mount()` to handle `s3` and `secret_env` auto mount types, delegating to existing `mount_s3()` and `set_env_vars_from_secret()` functions. Previously only `pvc` was supported as a config-driven auto mount type; S3 and secret-env users had to manually apply mount modifiers.

---

### 🛠️ Changes Made
- Added `s3` and `secret_env` branches in `auto_mount()` (`mlrun/runtimes/mounts.py`)
- Added 3 unit tests covering both new paths and the error case (`tests/runtimes/test_mounts.py`)

---

### ✅ Checklist

---

### 🔗 References
- Ticket link: [IG4-1658](https://iguazio.atlassian.net/browse/IG4-1658)

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes

[IG4-1658]: https://iguazio.atlassian.net/browse/IG4-1658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/runtimes/mounts.py
+++ b/mlrun/runtimes/mounts.py
@@ -371,6 +371,8 @@ def auto_mount(
     - k8s PVC volume when both pvc_name and volume_mount_path are set
     - k8s PVC volume when env var is set: MLRUN_PVC_MOUNT=<pvc-name>:<mount-path>
     - k8s PVC volume if it's configured as the auto mount type
+    - S3 credentials when configured as the auto mount type
+    - Secret-based env vars when configured as the auto mount type
     - iguazio v3io volume when V3IO_ACCESS_KEY and V3IO_USERNAME env vars are set
 
     """
@@ -388,6 +390,10 @@ def auto_mount(
     # parameters may still be declared - use them in that case.
     if config.storage.auto_mount_type == "pvc":
         return mount_pvc(**config.get_storage_auto_mount_params())
+    if config.storage.auto_mount_type == "s3":
+        return mount_s3(**config.get_storage_auto_mount_params())
+    if config.storage.auto_mount_type == "secret_env":
+        return set_env_vars_from_secret(**config.get_storage_auto_mount_params())
     if "V3IO_ACCESS_KEY" in os.environ:
         return mount_v3io(name=volume_name or "v3io")
 
--- a/mlrun/runtimes/mounts.py
+++ b/mlrun/runtimes/mounts.py
@@ -388,12 +388,13 @@ def auto_mount(
         )
     # In the case of CE when working remotely, no env variables will be defined but auto-mount
     # parameters may still be declared - use them in that case.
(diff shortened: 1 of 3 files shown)


rokatyy (commented):
@alxtkr77 could you please explain why this is needed and in which flows, I'm not sure I follow the idea of `auto_mount` itself as all the runtimes seems to be using https://github.com/mlrun/mlrun/blob/560e0a8e889d451efb60f00609026be6ed2e12ec/mlrun/runtimes/pod.py#L1215 function to mount on the function.

alxtkr77:
> @alxtkr77 could you please explain why this is needed and in which flows, I'm not sure I follow the idea of `auto_mount` itself as all the runtimes seems to be using
> 
> https://github.com/mlrun/mlrun/blob/560e0a8e889d451efb60f00609026be6ed2e12ec/mlrun/runtimes/pod.py#L1215
> 
> function to mount on the function.

try_auto_mount_based_on_config() is called automatically by the client launcher (enrich_runtime()) at .run() time. It already handles s3 and secret_env correctly via AutoMountType.get_modifier() — so any function that goes through the standard launcher path doesn't need this fix.           
auto_mount() is the explicit public API that users call as function.apply(auto_mount()). When a user does this:                                                                                                                                                                                    
1. auto_mount() runs and applies the mount (or raises ValueError without this fix)
2. apply() sets spec.disable_auto_mount = True, so try_auto_mount_based_on_config() will skip at launch time (it sees disable_auto_mount=True and returns early)

So the two paths are mutually exclusive: explicit auto_mount() wins over the automatic path. The bug is that on IG4/CE-with-S3, if a user or notebook calls function.apply(auto_mount()) — which is common,
it's in the docs and examples — it immediately raises ValueError("Failed to automount") before the function is even submitted.

mlrun/runtimes/mounts.py
```diff
@@ -388,6 +390,10 @@ def auto_mount(
     # parameters may still be declared - use them in that case.
     if config.storage.auto_mount_type == "pvc":
         return mount_pvc(**config.get_storage_auto_mount_params())
+    if config.storage.auto_mount_type == "s3":
+        return mount_s3(**config.get_storage_auto_mount_params())
```
rokatyy: I think we should do smth like:

```suggestion
        auto_mount_type = AutoMountType(type_)
        return auto_mount_type.get_modifier()(**config.get_storage_auto_mount_params())
```
instead of hardcoding each type with corresponding function again here

rokatyy (approved):

PR #9489, opened 2026-03-24
[Runtimes] Fix envvar order

liranbg:
### 📝 Description

Fix a bug where user-set environment variables (e.g., `AWS_ACCESS_KEY_ID`) are silently overridden by global/project secret injection during pod creation. When a user explicitly sets an env var via the batch run wizard or `function.set_env()`, the global secret (or project secrets with matching keys) replaces the user's value with a `secretKeyRef`, so the custom value never reaches the pod.

---

### 🛠️ Changes Made

- Added `_has_user_set_plain_env()` static method to `BaseRuntimeHandler` that distinguishes user-set plain-value env vars (`.value`) from secret-injected vars (`.value_from`)
- Added guard in the global secret injection loop (`add_k8s_secrets_to_spec`) to skip keys the user already set as plain env vars
- Added the same guard in the project secret injection loop
- Fixed unused loop variable: `for key, value` → `for key` in global secrets iteration
- Fixed return type annotations for `_list_pods_paginated` and `_list_crd_objects_paginated` (they yield, not return lists)
- Added `import typing` to `base.py` and `k8s.py` for generator type hints
- Added 4 regression tests in `test_base.py` covering all override scenarios

---

### ✅ Checklist

---

### 🧪 Testing

- UT + manual

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-12330
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


server/py/services/api/tests/unit/runtimes/test_base.py
```diff
@@ -0,0 +1,274 @@
+# Copyright 2023 Iguazio
```
Yacouby:
```suggestion
# Copyright 2026 Iguazio
```

server/py/services/api/runtime_handlers/base.py
```diff
@@ -426,13 +427,17 @@ def add_k8s_secrets_to_spec(
                     global_secret_name
                 )
             )
-            for key, value in global_secrets.items():
+            for key in global_secrets.keys():
```
Yacouby: Nit
```suggestion
            for key in global_secrets:
```

Yacouby (changes requested):
The try/finally config cleanup in the tests can be replaced with monkeypatch which handles the restore automatically and more standard pytest way to temporarily override config values.

Yacouby (approved):

PR #9493, opened 2026-03-24
[Docs] Update examples in serving graph getting started

jillnogold:


docs/serving/getting-started.md
```diff
@@ -69,130 +69,245 @@ For an example of writing the minimal serving functions, see [Minimal sklearn se
 See the full [V2 Model Server (SKLearn) example](https://github.com/mlrun/functions/blob/master/functions/src/v2_model_server/v2_model_server.ipynb) that 
 tests one or more classifier models against a held-out dataset.
 
-## Example of advanced data processing and serving ensemble
+## Streaming serving function
```
gilad-shaham, line 128: We should add why streaming is helpful, this is relevant when the responses are large, and when there response can be sent incrementally to the user, for example, an LLM output displayed to the user as the tokens as generated.

docs/serving/getting-started.md
```diff
@@ -69,130 +69,245 @@ For an example of writing the minimal serving functions, see [Minimal sklearn se
...
-to wait for the async event loop to complete.
-  
-```python
-server = function.to_mock_server()
-resp = server.test("/v2/models/m2/infer", body={"inputs": data})
-server.wait_for_completion()
-``` 
 
-And deploy the graph as a real-time Nuclio serverless function with one command:
+## ModelRunnerStep with proxy models for a shared model
```
gilad-shaham: I think we should show MRS early on in this page, before streaming. Maybe even before model router as in the future we'll remove the router option, so the most basic example is MRS.

gilad-shaham (changes requested):

docs/serving/getting-started.md
```diff
@@ -69,130 +69,245 @@ For an example of writing the minimal serving functions, see [Minimal sklearn se
...
+{Admonition}Important
+`function.invoke()` does not support streaming responses — it buffers the
+entire response and tries to parse it as a single JSON object. Use `requests` with
+`stream=True` instead.
+``` 
+### Define the streaming step
 
-Create a function of type serving from code and set the graph topology to `async flow`.
+A streaming step is any step whose `do()` method is a generator (sync or async).
+Each yielded value becomes a separate HTTP chunk in the response.
```
gtopper, line 144: This is actually not guaranteed in Nuclio. The number of HTTP response chunks could be anywhere from 1 to the number of chunks that arrive at the responder step (which doesn't necessarily match what is emitted by some other step in the graph – e.g. some chunks might be filtered out before they reach the responder).

docs/serving/getting-started.md
```diff
@@ -69,130 +69,245 @@ For an example of writing the minimal serving functions, see [Minimal sklearn se
...
 
-Use the graph `error_handler` if you want an error from the graph or a step to be fed into a specific state (catcher). See the full description in {ref}`pipelines-error-handling`.
+Use `requests` with `stream=True` and iterate over chunks as they arrive.
 
-Specify which step is the responder (returns the HTTP response) using the `step.respond()` method. 
-If the responder is not specified, the graph is non-blocking.
+```
+{Admonition}Note
+HTTP chunk boundaries are not guaranteed to align 1:1 with yielded values.
+The network stack or proxies may coalesce multiple chunks into a single read.
```
gtopper, line 191: Right, this is what I was referring to in my earlier comment.

docs/serving/getting-started.md
```diff
@@ -69,130 +69,245 @@ For an example of writing the minimal serving functions, see [Minimal sklearn se
...
 ```python
-# define a new real-time serving function (from code) with an async graph
-project = mlrun.get_or_create_project("myproj")
+graph.to("generator").to("evaluator").to("choice").cycle_to(["generator"]).to(
+    "post_process"
+).respond()
+```
+As an alternative to the choice step, you can implement {py:class}`~mlrun.serving.states.ModelRunnerSelector.select_outlets` in the evaluator step. See the usage in [Prevent infinite loops](#prevent-infinite-loops).
+
+Iteration tracking is automatic, you do not need to add counters manually in the step code. The default number of iterations is 10_000.
```
MichalEshchar: 10,000 - but I'm not sure it is the correct number

docs/serving/getting-started.md
```diff
@@ -69,130 +69,245 @@ For an example of writing the minimal serving functions, see [Minimal sklearn se
...
+import storey
+
+
+class MyChoiceStep(storey.Choice):
+    def select_outlets(self, event):
+        if event["should_continue"]:
+            return ["continue"]
+        return ["stop"]
+```
+The list returned by `select_outlets` must include only valid step names that follow the current step in the graph flow. If the current step is the responder, use Complete as the outlet name to exit the graph and return the response.
```
MichalEshchar, line 317: It would be great to add here a chart of a cyclic graph

docs/serving/getting-started.md
```diff
@@ -69,130 +69,245 @@ For an example of writing the minimal serving functions, see [Minimal sklearn se
...
 
-## Example of serving function using Kafka queue and serving child function
+## Serving function using Kafka queue and serving child function
```
MichalEshchar, line 319: need some more explanation regarding the motivation to use queue

docs/serving/getting-started.md
jillnogold: moved up

docs/serving/getting-started.md
jillnogold: added

docs/serving/getting-started.md
jillnogold: deleted

docs/serving/getting-started.md
jillnogold: added

docs/serving/getting-started.md
jillnogold: added

docs/serving/getting-started.md
jillnogold: fixed (100)

MichalEshchar (approved):

gilad-shaham (approved):

PR #9390, opened 2026-03-02
[Docs] Add graph API handler

jillnogold:
### 📝 Description
[ML-9565](https://iguazio.atlassian.net/browse/ML-9565)
[ML-12054](https://iguazio.atlassian.net/browse/ML-12054)

[ML-9565]: https://iguazio.atlassian.net/browse/ML-9565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ML-12054]: https://iguazio.atlassian.net/browse/ML-12054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9496, opened 2026-03-24
[Docs] Add graph API handler documentation [1.11.x]

jond01:
## Summary
[ML-12054](https://iguazio.atlassian.net/browse/ML-12054) | [ML-9565](https://iguazio.atlassian.net/browse/ML-9565)

- Adds `docs/serving/api-handler.md` documenting the graph API handler feature
- Adds the new page to the serving-graph TOC

The doc covers: configuration (`APIHandlerConfig`, `add_endpoint_handler`, `add_body_mapping`), path matching priority (exact → template → wildcard), parameter extraction as kwargs, `include_url_info`, and a complete runnable example.

All code snippets were validated against the real implementation using a mock server.

Notable corrections vs the draft PR (#9390):
- Removed unimplemented `mlrun_url_path_segments` / `mlrun_url_query_params` fields; documents actual `mlrun_request_path` field
- Removed out-of-scope admin API use-cases (model monitoring, add/remove models)
- `body_map` is global and applies to all ALLOW endpoints — documented with explicit note
- No distinction between class-based and function handler signatures (both use kwargs)

[ML-12054]: https://iguazio.atlassian.net/browse/ML-12054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ML-9565]: https://iguazio.atlassian.net/browse/ML-9565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/serving/api-handler.md
+++ b/docs/serving/api-handler.md
@@ -120,8 +120,13 @@ config.add_body_mapping("titles", "$['store']['book'][*]['title']")
 Rules:
 - A single match → the value is returned as-is.
 - Multiple matches → a list is returned.
-- No match → the request fails with a 422 (Unprocessable Entity) error.
-- `body_map` applies to **all** ALLOW endpoints in the config. Every allowed request must supply a JSON dict body that satisfies all configured mappings.
+- No match → the request fails with HTTP 422 (Unprocessable Entity).
+- Non-JSON or non-dict body → the request fails with HTTP 422 (Unprocessable Entity).
+- `body_map` applies to **all** ALLOW endpoints in the config. Because the mapping is global, a `body_map` that matches one endpoint's body format may cause 422 errors for other endpoints whose requests do not share that format.
+
+```{note}
+The strict 422 behavior for missing body mappings and non-JSON bodies is a known limitation in the current release and will be relaxed in a future release to allow more flexible body handling.
+```
 
 To remove a mapping: `config.remove_body_mapping("model_name")`
 
--- a/docs/serving/api-handler.md
+++ b/docs/serving/api-handler.md
@@ -163,9 +163,9 @@ def step_handler(body, completion_id, limit, mlrun_request_path, **kwargs):
     # mlrun_request_path="/v1/chat/completions/abc123"  — from include_url_info
     ...
 
+
 class MyStep:
-    def do(self, body, completion_id, limit, mlrun_request_path, **kwargs):
-        ...
+    def do(self, body, completion_id, limit, mlrun_request_path, **kwargs): ...
 ```
 
 ## Complete example
--- a/docs/serving/api-handler.md
+++ b/docs/serving/api-handler.md
@@ -61,10 +61,10 @@ The configuration is serialized into `serving_fn.spec.api_handler_config` and is
 
 ```python
 config.add_endpoint_handler(
-    path,                                   # URL path, e.g. "/v1/chat" or "/v1/chat/*"
-    http_method=HTTPMethod.POST,            # HTTPMethod enum or string ("GET", "POST", ...)
-    action=APIHandlerAction.ALLOW,          # ALLOW or FORBID
-    description=None,                       # Optional human-readable description
+    path,  # URL path, e.g. "/v1/chat" or "/v1/chat/*"
+    http_method=HTTPMethod.POST,  # HTTPMethod enum or string ("GET", "POST", ...)
+    action=APIHandlerAction.ALLOW,  # ALLOW or FORBID
+    description=None,  # Optional human-readable description
 )
 ```
 
@@ -142,8 +142,8 @@ A `GET /v1/chat/completions/abc123?limit=10` request passes the following keywor
 
 ```python
 {
-    "completion_id": "abc123",                            # from path template
-    "limit": "10",                                        # from query string
+    "completion_id": "abc123",  # from path template
+    "limit": "10",  # from query string
     "mlrun_request_path": "/v1/chat/completions/abc123",  # from include_url_info
 }
 ```
@@ -153,14 +153,12 @@ A `GET /v1/chat/completions/abc123?limit=10` request passes the following keywor
 Extracted parameters are passed as keyword arguments to the handler function or `do()` method:
 
 ```python
-def predict(model_name, inputs, **kwargs):
-    ...
+def predict(model_name, inputs, **kwargs): ...
 ```
 
 ```python
 class InferenceStep:
-    def do(self, model_name, inputs, **kwargs):
-        ...
+    def do(self, model_name, inputs, **kwargs): ...
 ```
 
 ## Complete example
@@ -182,11 +180,13 @@ serving_fn = project.set_function(
     image="mlrun/mlrun",
 )
 
+
 # --- Define the graph step that processes chat completions ---
 # model_name and messages are passed as keyword arguments by the API handler
 def chat_handler(model_name, messages, mlrun_request_path, **kwargs):
     reply = f"Received {len(messages)} message(s) for model '{model_name}'"
     return {"reply": reply, "path": mlrun_request_path}
+
 
 graph = serving_fn.set_topology("flow", engine="sync")
 graph.to(name="chat", handler=chat_handler).respond()
--- a/docs/serving/api-handler.md
+++ b/docs/serving/api-handler.md
@@ -1,0 +1,238 @@
+(api-handler)=
+# API handler
+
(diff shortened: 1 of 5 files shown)


docs/serving/api-handler.md
```diff
@@ -0,0 +1,238 @@
...
+
+| Configured path           | Matches                             | Does not match   |
+|---------------------------|-------------------------------------|------------------|
+| `/v1/models`              | `/v1/models`                        | `/v1/models/gpt` |
+| `/v1/models/{model_name}` | `/v1/models/gpt`, `/v1/models/bert` | `/v1/models`     |
+| `/v1/*`                   | `/v1/chat`, `/v1/models/gpt`        | `/v1`            |
+
+## Extracting request parameters
+
+When the handler allows a request, it may extract parameters from three sources:
```
jillnogold: When the handler allows a request, it can extract parameters from three sources:

docs/serving/api-handler.md
```diff
@@ -0,0 +1,238 @@
...
+
+An API handler is a graph step that is automatically prepended to a serving graph when configured. It validates incoming HTTP requests against a set of user-defined endpoints, extracts parameters from path templates, query strings, and the request body, and passes them to the next step in the graph.
+
+Use an API handler to:
+- Implement industry-defined REST API schemas on your serving graph (for example, the OpenAI chat-completion interface for LLMs).
+- Gate access to specific paths using ALLOW and FORBID rules.
+
+API handlers are active only for HTTP-triggered invocations. When an event arrives through a non-HTTP trigger such as a stream, the API handler is bypassed (the path is always `/` in that case).
+
+**Supported runtimes**: Nuclio functions with an HTTP trigger, and the mock server (local testing).
```
theSaarco:
```suggestion
**Supported runtimes**: Serving functions with an HTTP trigger, and the mock server (local testing).
```

docs/serving/api-handler.md
```diff
@@ -0,0 +1,238 @@
...
+
+```python
+config.remove_endpoint_handler("/v1/models", HTTPMethod.GET)
+```
+
+## Path matching rules
+
+Endpoints are matched in the following priority order:
+
+1. **Exact paths** — no wildcards or template parameters, e.g. `/v1/models`. O(1) dict lookup.
```
theSaarco:
```suggestion
1. **Exact paths** — no wildcards or template parameters, e.g. `/v1/models`. 
```

Since we don't mention complexity for other types of paths, let's not put it here as well.

docs/serving/api-handler.md
```diff
@@ -0,0 +1,238 @@
...
+
+| Source                   | How to configure                 | Available as               |
+|--------------------------|----------------------------------|----------------------------|
+| Path template parameters | `{param}` in the path pattern    | keyword argument           |
+| Query string parameters  | automatic                        | keyword argument           |
+| Request body fields      | `body_map` (JSONPath, see below) | keyword argument           |
+
+The extracted parameters are passed to the next step as keyword arguments. If the same parameter name appears in more than one source, the request fails with a 400 error.
+
+**Priority when no conflicts exist**: path > query > body_map (highest to lowest).
```
theSaarco: What does that mean? If there are no conflicts, then priority between what?

docs/serving/api-handler.md
```diff
@@ -0,0 +1,238 @@
...
+
+If no parameters are extracted and `include_url_info` is not enabled, the original event body is passed through unchanged.
```
theSaarco: This sentence is a bit confusing.
What exactly does `include_url_info` control? The above 3 types of parameters are passed anyway, right?
If `include_url_info` is not configured and there are no param extracted, the event body will not be passed? It should be passed anyway if the URL is valid, right?

docs/serving/api-handler.md
```diff
@@ -0,0 +1,238 @@
...
+
+### Body mapping (`body_map`)
+
+`body_map` maps parameter names to [JSONPath](https://goessner.net/articles/JsonPath/) expressions. It is **global** — the same mapping applies to every configured endpoint. The request body must be a JSON dict when `body_map` is configured.
```
theSaarco: I think it's better to reference the official RFC here - https://datatracker.ietf.org/doc/html/rfc9535
Users can look up specific explanations by other contributors themselves.

docs/serving/api-handler.md
```diff
@@ -0,0 +1,238 @@
...
+Extracted parameters are passed as keyword arguments to the handler function or `do()` method:
+
+```python
+def predict(model_name, inputs, **kwargs): ...
+```
+
+```python
+class InferenceStep:
+    def do(self, model_name, inputs, **kwargs): ...
+```
```
theSaarco, line 169: Perhaps use the example above (with `completion_id` etc.) and add a comment how this will be passed to the functions.

docs/serving/api-handler.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Complete example
```
theSaarco, line 171: Nice!

theSaarco (changes requested):
Looks good! Have some suggestions for improvement.

docs/serving/api-handler.md
jond01: You are right, in the case of conflict between the same parameter name in the path (a "path param") and a parameter in the body (through) the `body_map`, an error is raised. That is true for other conflicts as well.

Removed this sentence.

docs/serving/api-handler.md
jond01: Rephrased this paragraph.
`include_url_info` - by default it's `False`, but if it's true, the normalized path is passed to the graph handler as `mlrun_request_path` parameter.
The parameters are passed, the question is just how - in a dictionary, or as plain parameters.

docs/serving/api-handler.md
```diff
@@ -0,0 +1,240 @@
...
+config.add_body_mapping("inputs", "$.data.inputs")
+
+# Multiple matches (e.g. all book titles) return a list
+config.add_body_mapping("titles", "$['store']['book'][*]['title']")
+```
+
+Rules:
+- A single match → the value is returned as-is.
+- Multiple matches → a list is returned.
+- No match → the request fails with a 422 (Unprocessable Entity) error.
```
theSaarco: Is this the behavior? Because this is not right. It should not fail the processing. See slack discussion for reasoning.

docs/serving/api-handler.md
```diff
@@ -0,0 +1,240 @@
...
+- `body_map` applies to **all** ALLOW endpoints in the config. Every allowed request must supply a JSON dict body that satisfies all configured mappings.
```
theSaarco: This is not true, and why the above comment is written. We cannot force the incoming requests to all follow the same structure. We must be more flexible in that sense, and not fail if missing body mappings exist.
If we had a mapping per incoming URL, then it would make sense, but given that the mappings are global, this behavior is not reasonable.

theSaarco (changes requested):
Seems that we have an issue with the feature spec, reflected in the docs. We need to see what we do with this behavior.

mlrun/runtimes/nuclio/serving.py
```diff
@@ -191,9 +191,9 @@ def add_endpoint_handler(
         action: schemas.serving.APIHandlerAction = schemas.serving.APIHandlerAction.ALLOW,
         description: str | None = None,
     ) -> None:
-        """Add an endpoint handler configuration.
+        r"""Add an endpoint handler configuration.
```
jillnogold: """Add an endpoint handler configuration.

jond01: Remove 👍

docs/serving/api-handler.md
jond01: This change in behavior is tracked in [ML-12376](https://iguazio.atlassian.net/browse/ML-12376), and is implemented in #9528. The target version for it is 1.11.1.

The docs in this version include the following note:

> The strict 422 behavior for missing body mappings and non-JSON bodies is a known limitation in the current release and will be relaxed in a future release to allow more flexible body handling.

docs/serving/api-handler.md
jond01: Please see the comment below

jillnogold (approved):

theSaarco (approved):
LGTM

PR #9497, opened 2026-03-24
[Docs] Improve artifact doc

jillnogold:
[ML-9732](https://iguazio.atlassian.net/browse/ML-9732)

- The order of the documentation does not make sense since artifacts are used in the data item, which appears before the artifacts section. (fixed in previous PR)
- The data stores define how to work with different storage providers, and I don’t think it should be the first section. (fixed in previous PR)
- No example for how to log something as an artifact (“dataset_artifact = project.log_dataset(key="test-michal",df=example_df,format ="csv")
- View artifacts – only explain how to do so using the UI but no code examples.

[ML-9732]: https://iguazio.atlassian.net/browse/ML-9732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/store/artifacts.md
MichalEshchar: Do we still need to mention those versions?

MichalEshchar (approved):

assaf758 (approved):

PR #9501, opened 2026-03-25
[Alerts] Add cooldown period support for auto-reset alerts

Yacouby:
### 📝 Description
This PR adds optional `cooldown_period` on alerts (when reset_policy=auto) and persists `cooldown_end_time` on `alert_states`, so firing an alert can keep it active for a configured duration before auto-reset (instead of resetting immediately). 

---

### 🛠️ Changes Made
Schema
  - Added `cooldown_period` field to `AlertConfig` & `AlertTemplate`

  SDK
  - Added `cooldown_period` to `dict_fields`, `_init_`, and docstring

  Config
  - Added `cooldown_reset_interval`: 15 seconds under alerts section

  DB 
  - Added `cooldown_end_time` column to `AlertState`
  - Added composite index `(active, cooldown_end_time)` on `AlertState`
  - Updated `store_alert_state()` signature with `cooldown_end_time` and `clear_cooldown`

  Alerts CRUD
  - `validate_alert_cooldown_period()` — new method validating cooldown period format, reset policy constraint, and minimum value
  - `handle_notification()` — computes `cooldown_end_time`, passes it to `store_alert_state`, skips immediate auto-reset when cooldown is set
  - `reset_alert() `— added `obj={}` to clear stale JSON, passes `clear_cooldown=True`
  - `should_reset_alert()`— added `cooldown_period` to functional parameters

**Bug Fix (pre-existing):**
 Fixed a bug where `reset_alert()` never cleared the full_object JSON column in `alert_states` on reset. 
This caused stale `event_timestamps` and `last_activation_id` to persist in
  the DB after a manual reset. 
On service restart, `populate_caches()` would restore these stale timestamps into memory, potentially causing the alert to fire prematurely on the next event.
Fixed by passing `obj={}` to `store_alert_state()` in `reset_alert()`.

---

### ✅ Checklist

---

### 🧪 Testing
Unit tests:
- Added `test_alerts_cooldown.py`: validates `cooldown_period` (vs reset_policy and min duration), reset-on-config-change, cooldown vs immediate-auto notification behavior, and `reset_alert` kwargs
- Refactored `reset_alert_caches` into `server/py/services/alerts/tests/unit/crud/conftest.py` and wired `test_alerts.py` to use it so alert LRU caches are cleared safely between tests.
- Extended `generate_alert_data` in `utils.py` with `cooldown_period` for test data.
- DB unit tests.

Patch code to vmdev26:
Performed mlrun migrations successfully

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-12297
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


server/py/services/alerts/crud/alerts.py
```diff
@@ -296,11 +299,27 @@ def _handle_notification(
...
             state["active"] = True
             state_obj["last_activation_id"] = activation_id
 
+        cooldown_end_time = None
+        if alert.cooldown_period:
+            cooldown_td = framework.utils.helpers.string_to_timedelta(
+                alert.cooldown_period, raise_on_error=False
+            )
+            if cooldown_td is not None:
+                cooldown_end_time = datetime.datetime.now(datetime.UTC) + cooldown_td
```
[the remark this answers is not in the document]
Yacouby: periodic job will be added in a following PR 👍

mlrun/alerts/alert.py
```diff
@@ -27,6 +27,7 @@ class AlertConfig(ModelObj):
         "summary",
         "severity",
         "reset_policy",
+        "cooldown_period",
```
rokatyy, line 30: Now that I branched out the code I see that there is already another cooldown_period in Notification object, that means completely different thing:

https://github.com/mlrun/mlrun/blob/16a39a25fcd7e19e269d620d52b6508c0f10950f/mlrun/common/schemas/alert.py#L130-L138

I think that's smth worth raising and discussing as it might be confusing for the end user. cc: @theSaarco

server/py/framework/db/sqldb/db.py
```diff
@@ -6994,6 +6996,10 @@ def store_alert_state(
         state.active = active
         if obj is not None:
             state.full_object = obj
+        if cooldown_end_time is not None:
+            state.cooldown_end_time = cooldown_end_time
+        if clear_cooldown:
+            state.cooldown_end_time = None
```
rokatyy, line 7003: are they mutually exclusive?

server/py/framework/db/sqldb/models.py
```diff
@@ -836,6 +843,11 @@ class AlertState(Base, framework.db.sqldb.base.BaseModel):
             default=None,
         )
         active = Column(BOOLEAN, default=False)
+        cooldown_end_time = Column(
+            framework.db.sqldb.sql_types.DateTime(timezone=True),
```
rokatyy: why timezone=True? we don't do this in other places

server/py/services/api/migrations/versions/3f2c1331622a_add_cooldown_end_time_to_alert_states.py
```diff
@@ -0,0 +1,57 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "alert_states",
+        sa.Column(
+            "cooldown_end_time",
+            mysql.DATETIME(timezone=True, fsp=3),
```
rokatyy: why not fsp=6?

rokatyy (commented):
Looks good!

Some question and suggestions below. Also, could you please add comments near new conditions checks with whys?

server/py/framework/db/sqldb/db.py
Yacouby: Callers should pass either a new `cooldown_end_time` or `clear_cooldown=True`, not both. 
If both were passed, the implementation clears last (`clear_cooldown` wins). 
I’ll add a one-line comment above these branches to document that contract.

server/py/framework/db/sqldb/models.py
Yacouby: we do have it, e.g in:
https://github.com/mlrun/mlrun/blob/3810f84654f99713c5ae71b6070792486378782f/server/py/framework/db/sqldb/models.py#L945
+
https://github.com/mlrun/mlrun/blob/3810f84654f99713c5ae71b6070792486378782f/server/py/framework/db/sqldb/models.py#L961

server/py/services/api/migrations/versions/3f2c1331622a_add_cooldown_end_time_to_alert_states.py
Yacouby: its just to be aligned with `activation_time` & `reset_time`

rokatyy: @Yacouby if this is the only reason, then I think we should use fsp as with activation_time, there is a strong reason why we are doing this (mainly old indexing problem and table might be huge). With states table, we don't have these problems, so we can use 6 fsp straight away

server/py/framework/db/sqldb/models.py
Yacouby: Removed `timezone=True` from the column definition - MicroSecondDateTime already applies it

mlrun/alerts/alert.py
Yacouby: I asked it in the ticket and he responded to keep it as is ( first answer here : https://iguazio.atlassian.net/browse/ML-11994?focusedCommentId=209471 )

rokatyy (approved):

server/py/framework/db/sqldb/db.py
```diff
@@ -6994,6 +6996,11 @@ def store_alert_state(
         state.active = active
         if obj is not None:
             state.full_object = obj
+        # These two are mutually exclusive: pass cooldown_end_time to set it, or clear_cooldown=True to null it.
+        if cooldown_end_time is not None:
+            state.cooldown_end_time = cooldown_end_time
+        if clear_cooldown:
+            state.cooldown_end_time = None
```
TomerShor, line 7003: Won't it be simpler if we only have the one `cooldown_end_time` argument and allow setting it to 0?
Then:
- if `0` - set to None
- elif not `None` - set the time

TomerShor (commented):
Overall LGTM - 1 question

server/py/framework/db/sqldb/db.py
Yacouby: I'd prefer to keep the two separate parameters, since using 0 mixes types (datetime and int) and hides the intent at the call site, and a single `cooldown_end_time` would still be awkward for “don’t change the column” vs “set to NULL” on partial updates. `clear_cooldown=True` keeps that explicit.

TomerShor (approved):

PR #9505, opened 2026-03-26
[Model Monitoring] Reduce drift_table_plot artifact size via configurable include_plotlyjs

jond01:
## Summary

Relates to [ML-12267](https://iguazio.atlassian.net/browse/ML-12267) — `drift_table_plot` artifact creation takes much longer than before on IG4.

The original issue was not reproduced, but based on the symptom (slow artifact creation/upload), the suspected cause is artifact size: by default, `plotly.Figure.to_html()` embeds the full Plotly.js bundle (~3.4 MB) inline in every HTML artifact.

This PR is an effort to improve the situation under that assumption:

- Add `include_plotlyjs: bool | str = True` parameter to `PlotlyArtifact.__init__`, passed directly to [`plotly.io.to_html(include_plotlyjs=...)`](https://plotly.com/python-api-reference/generated/plotly.io.to_html.html). This makes the behavior configurable for all Plotly artifacts.
- Add `include_plotlyjs: bool | str = "cdn"` parameter to `FeaturesDriftTablePlot.produce`, defaulting to `"cdn"` for the drift table plot specifically. With `"cdn"`, the exported HTML references Plotly.js from the official CDN instead of embedding it, reducing the artifact from ~3.4 MB to ~8 KB.

## Test plan

- [x] Existing unit tests pass
- [x] Verify `drift_table_plot` artifact size is reduced when `include_plotlyjs="cdn"`
- [x] Verify artifact still renders correctly in MLRun UI (requires network access to CDN)

[ML-12267]: https://iguazio.atlassian.net/browse/ML-12267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/artifacts/plots.py
+++ b/mlrun/artifacts/plots.py
@@ -79,13 +79,23 @@ class PlotlyArtifact(Artifact):
         figure: typing.Optional["Figure"] = None,
         key: str | None = None,
         target_path: str | None = None,
+        include_plotlyjs: bool | str = True,
     ) -> None:
         """
         Initialize a Plotly artifact with the given figure.
 
-        :param figure:      Plotly figure ('plotly.graph_objs.Figure' object) to save as an artifact.
-        :param key:         Key for the artifact to be stored in the database.
-        :param target_path: Path to save the artifact.
+        :param figure:           Plotly figure ('plotly.graph_objs.Figure' object) to save as an artifact.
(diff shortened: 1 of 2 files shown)


jond01:
Closing this as the original bug could not be reproduced.

PR #9515, opened 2026-03-29
[SDK] Fix HTTPDB http client socket leak

liranbg:
### 📝 Description

`HTTPRunDB.api_call()` creates a new `HTTPSessionWithRetry` on every POST/PUT request, replacing `self.session` without closing the previous one. Each orphaned session retains its urllib3 `PoolManager` (configured with `pool_maxsize=64`), keeping TCP sockets alive until GC. In long-running processes this causes `CLOSE_WAIT` socket accumulation and eventually `OSError: [Errno 24] Too many open files`.

---

### 🛠️ Changes Made

- **`mlrun/db/httpdb.py`**: Changed `api_call()` to create the session once and reuse it. For POST/PUT, the retry policy is updated in-place via `session.update_retry_methods()` instead of replacing the entire session.
- **`mlrun/utils/http.py`**: Added `update_retry_methods()` to `HTTPSessionWithRetry` that updates both the internal `_retry_methods` frozenset and the mounted adapter's `Retry` object without recreating the session or adapter.
- **`tests/rundb/test_httpdb_sessions.py`**: New test file with 8 tests covering session reuse across all HTTP methods (parametrized), mixed-method reuse, leak prevention, and `update_retry_methods()` correctness (frozenset update, adapter update, no-adapter guard, integration with `api_call`).

---

### ✅ Checklist

---

### 🧪 Testing

- unit tests + manual testings

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-12369
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes

The root cause was that retry policies are baked into the `urllib3.util.retry.Retry` object at `HTTPAdapter` construction time. The old code created a fresh session per POST/PUT to get the right retry config for each path. The fix updates the retry config in-place on the existing session, which also improves performance by preserving the connection pool across requests.

### DIFF
--- a/mlrun/utils/http.py
+++ b/mlrun/utils/http.py
@@ -232,7 +232,14 @@ class HTTPSessionWithRetry(requests.Session):
         :param retry_on_post: Whether POST requests should be retried.
         :param retry_on_put:  Whether PUT requests should be retried.
         """
-        self._retry_methods = self._resolve_retry_methods(retry_on_post, retry_on_put)
+        new_methods = self._resolve_retry_methods(retry_on_post, retry_on_put)
+
+        # Skip Retry object re-allocation when the allowed methods haven't changed
+        # consecutive calls with the same policy (common case) don't need a new Retry object
+        if new_methods == self._retry_methods:
+            return
+
+        self._retry_methods = new_methods
         if hasattr(self, "_http_adapter"):
             self._http_adapter.max_retries = urllib3.util.retry.Retry(
                 total=self.max_retries,
--- a/mlrun/utils/http.py
+++ b/mlrun/utils/http.py
@@ -222,6 +222,26 @@ class HTTPSessionWithRetry(requests.Session):
             return False
         return True
 
+    def update_retry_methods(self, retry_on_post: bool, retry_on_put: bool) -> None:
+        """Update the retry method set on the session and its mounted adapter.
+
+        This allows reusing a single session across requests with different
+        retry policies (e.g., POST paths that are retriable vs. non-retriable),
+        avoiding the overhead of creating a new session per request.
+
+        :param retry_on_post: Whether POST requests should be retried.
+        :param retry_on_put:  Whether PUT requests should be retried.
+        """
+        self._retry_methods = self._resolve_retry_methods(retry_on_post, retry_on_put)
+        if hasattr(self, "_http_adapter"):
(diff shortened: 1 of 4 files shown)


tests/rundb/test_httpdb_sessions.py
```diff
@@ -0,0 +1,160 @@
+# Copyright 2023 Iguazio
```
Yacouby:
```suggestion
# Copyright 2026 Iguazio
```

mlrun/utils/http.py
```diff
@@ -222,6 +222,26 @@ def err_chain(err):
...
+        """Update the retry method set on the session and its mounted adapter.
+
+        This allows reusing a single session across requests with different
+        retry policies (e.g., POST paths that are retriable vs. non-retriable),
+        avoiding the overhead of creating a new session per request.
+
+        :param retry_on_post: Whether POST requests should be retried.
+        :param retry_on_put:  Whether PUT requests should be retried.
+        """
+        self._retry_methods = self._resolve_retry_methods(retry_on_post, retry_on_put)
```
Yacouby: worth adding an early return if the resolved methods are the same as the current ones, to avoid allocating a new Retry object unnecessarily.

Yacouby (commented):
minor suggestion

tests/rundb/test_httpdb_sessions.py
```diff
@@ -0,0 +1,160 @@
...
+        for _ in range(num_calls):
+            db.api_call("POST", "some/path")
+
+        assert len(sessions_created) == 1, (
+            f"Expected 1 session created for {num_calls} POST calls, "
+            f"got {len(sessions_created)}"
+        )
+
+
+def test_update_retry_methods_updates_frozenset():
```
elbamit: What about `POST=False PUT=False` and `POST=True PUT=True`?

elbamit (approved):

tests/rundb/test_httpdb_sessions.py
liranbg: ill parameterize it

mlrun/utils/http.py
liranbg: small rare optimization but will add

Yacouby (approved):

PR #9516, opened 2026-03-29
[Auth] Tolerate missing auth tokens for scheduled runs

elbamit:
### 📝 Description

When a user's auth token secret is deleted, scheduled runs fail inside `enrich_and_validate_auth_token_name` before `store_run()` is ever called. This means no run record is created in the DB and the failure is completely invisible to the user.
This PR makes scheduled runs tolerate a missing token so the run record is still persisted and the failure is visible. Direct (non-scheduled) runs preserve the existing fail-fast behaviour.

---

### 🛠️ Changes Made

- Wrap `resolve_auth_token_name` in a try/except which does not raise for scheduled runs (detected via `schedule_name` label).

---

### ✅ Checklist

---

### 🧪 Testing
  
  

Unit tests:
- Scheduled run tolerates missing token
- Direct run still raises on missing token (both error types)
- Warning log includes `user_id` and `schedule_name` context
- Scheduled run with valid token enriches normally

Patch IG4 system and saw runs scheduled runs getting created and failing at runtime

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-12339
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


server/py/services/api/launcher.py
```diff
@@ -717,10 +717,28 @@ def enrich_and_validate_auth_token_name(
...
+            if not is_scheduled:
+                raise
+            mlrun.utils.logger.warning(
+                "Auth token not found for scheduled run; proceeding without token mount",
+                user_id=user_id,
+                schedule_name=(object.metadata.labels or {}).get(
+                    mlrun.common.schemas.constants.LabelNames.schedule_name
+                ),
+            )
+            return
```
liranbg: add a comment explaining why you allow such case

server/py/services/api/launcher.py
```diff
@@ -717,10 +717,28 @@ def enrich_and_validate_auth_token_name(
...
+        except (mlrun.errors.MLRunBadRequestError, mlrun.errors.MLRunNotFoundError):
+            if not is_scheduled:
+                raise
+            mlrun.utils.logger.warning(
+                "Auth token not found for scheduled run; proceeding without token mount",
+                user_id=user_id,
+                schedule_name=(object.metadata.labels or {}).get(
+                    mlrun.common.schemas.constants.LabelNames.schedule_name
+                ),
+            )
```
Yacouby: This long thing is called twice, store it in a variable and then use it

server/py/services/api/tests/unit/test_launcher.py
```diff
@@ -596,6 +596,149 @@ def test_enrich_and_validate_auth_token_name_iguazio_v4_token_not_found(
...
+        metadata=mlrun.model.RunMetadata(
+            labels={schedule_label: "my-schedule"},
+        ),
+        spec=mlrun.model.RunSpec(auth=None),
+    )
+
+    launcher.enrich_and_validate_auth_token_name(run)
+
+    # token_name must NOT be set — method returned before calling set_auth_token_name
+    assert run.spec.auth is None or "token_name" not in (run.spec.auth or {})
```
Yacouby: This is hard to read. cleaner:
```suggestion
    assert not run.spec.auth or run.spec.auth.get("token_name") is None
```

server/py/services/api/launcher.py
```diff
@@ -717,10 +717,28 @@ def enrich_and_validate_auth_token_name(
...
+        is_scheduled = bool(
+            (object.metadata.labels or {}).get(
+                mlrun.common.schemas.constants.LabelNames.schedule_name
+            )
         )
+        try:
+            token_name = services.api.utils.helpers.resolve_auth_token_name(
+                user_id=user_id, provided_token_name=provided_token_name
+            )
+        except (mlrun.errors.MLRunBadRequestError, mlrun.errors.MLRunNotFoundError):
```
Yacouby: Is it certain that only `MLRunBadRequestError`/`MLRunNotFoundError` are raised, what about `MLRunUnauthorizedError` for example ?

server/py/services/api/tests/unit/test_launcher.py
```diff
@@ -596,6 +596,149 @@ def test_enrich_and_validate_auth_token_name_iguazio_v4_token_not_found(
...
+
+@pytest.mark.parametrize(
+    "exc",
+    [
+        mlrun.errors.MLRunBadRequestError("token gone"),
+        mlrun.errors.MLRunNotFoundError("token revoked"),
+    ],
+    ids=["bad_request", "not_found"],
+)
+def test_enrich_and_validate_auth_token_name_scheduled_run_logs_warning(
```
Yacouby: This test overlaps with the first one.
The first test already implicitly verifies the warning path (no exception) and this one re-patches the same path and also verifies no exception is raised, just to check log kwargs. 
Merging them both into one, since they test the same code path.

server/py/services/api/launcher.py
Yacouby: This early return skips `set_auth_user_id` and the `_auth_info.user_id` backfill.
The comments below say those are needed for run retries and for downstream code (e.g. `_mount_secret_token_to_runtime`). Even when the token is revoked, we still have a `user_id` ..should we still persist/propagate it here and only skip setting `token_name`?

Yacouby (changes requested):
Suggestions + some questions

Yacouby (approved):

PR #9517, opened 2026-03-29
[Serving] Fix body_map to be usable across endpoints with different body formats

jond01:
## Summary

Fixes ML-12376: `body_map` was raising 422 errors for endpoints that don't match the configured JSONPath expressions, making it impossible to share a serving graph across endpoints with different body formats.

### Changes

**Bug fixes:**
- `body_map` no longer raises 422 when the request body is not a dict — non-dict bodies silently skip body_map extraction, so endpoints with different payload formats (e.g. one JSON, one plain text) can coexist in the same graph
- `body_map` no longer raises 422 when a JSONPath expression has no match — unmatched paths are silently skipped instead of erroring

**Refactor (internal):**
- Removed `MLRunUnprocessableEntityError` from `mlrun/errors.py` (was only used for the two cases above)
- Unified `_MappedBody`, the local `_RequestContext` in `api_handler.py`, and `_RequestContextWithBody` in `utils.py` into a single `_RequestContext` class in `utils.py`
  - The unified class carries the original event body as `original_body` and merges path/query/body_map params with conflict detection
  - `url_params` (reserved `mlrun_` prefix) are merged last without conflict checking
- `_MappedBodyAwareHandler` now consistently calls handlers as `fn(original_body, **params)` when the body is a `_RequestContext`, giving handlers clean access to both the raw body and all extracted parameters

## Test plan

- [x] `pytest tests/serving/test_body_map.py tests/serving/test_api_handler.py` — all 154 tests pass
- [x] Verify non-dict body no longer raises 422 (`test_body_map_non_dict_body_skipped`)
- [x] Verify missing JSONPath match no longer raises 422 (`test_body_map_missing_params_skipped`)
- [x] Verify dict body with no matching fields still passes original body through (`test_run_body_map_dict_body_no_fields_match`)

### DIFF
--- a/mlrun/serving/api_handler.py
+++ b/mlrun/serving/api_handler.py
@@ -32,66 +32,7 @@ import mlrun.serving.states
 import mlrun.serving.utils as serving_utils
 import mlrun.utils
 from mlrun.common.schemas.serving import _APIEndpointKeys
-from mlrun.serving.utils import _MappedBody
-
-
-class _RequestContext(_MappedBody):
-    """Unified container for request parameters from multiple sources.
-
-    Consolidates parameters from body_map (JSONPath extraction), path templates,
-    and query strings into a single dict-like object that unpacks as kwargs.
-
-    Priority order: path > query > body_map (path parameters have highest priority)
-    Conflicts are not allowed - request will fail if same parameter appears in multiple sources.
-
-    Additionally, system-injected ``url_params`` (e.g. ``mlrun_request_path``) are merged
-    in after the conflict check without raising conflicts - they use the reserved ``mlrun_``
-    prefix to avoid collisions with user-defined parameters.
-    """
-
-    def __init__(
-        self,
-        path_params: dict[str, str] | None = None,
-        query_params: dict[str, str | list[str]] | None = None,
-        body_params: dict[str, Any] | None = None,
-        url_params: dict[str, Any] | None = None,
-    ):
-        merged = {}
-        sources = [
-            ("body_map", body_params or {}),
-            ("query", query_params or {}),
-            ("path", path_params or {}),
-        ]
-
-        # Track conflicts for better error messages
-        param_sources: dict[str, list[str]] = {}
-
-        for source_name, params in sources:
-            for key, value in params.items():
-                if key in merged:
-                    param_sources.setdefault(key, []).append(source_name)
-                else:
-                    param_sources[key] = [source_name]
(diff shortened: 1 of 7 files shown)


mlrun/serving/api_handler.py
```diff
@@ -455,18 +391,19 @@ def do(
                         query_params=query_params,
                         url_params=url_params,
                     )
-                    request_context = _RequestContext(
+                    original_body = event.body if hasattr(event, "body") else event
```
assaf758: if there is no body, why not set original_body to None?

mlrun/serving/api_handler.py
```diff
@@ -420,33 +357,32 @@ def do(
...
-                        )
-                    except mlrun.errors.MLRunUnprocessableEntityError:
-                        raise
-                    except Exception as exc:
-                        raise mlrun.errors.MLRunBadRequestError(
-                            f"Failed to process body_map transformation: {exc}"
-                        ) from exc
+                    if isinstance(body, dict):
+                        try:
+                            body_params = dict(self._apply_parsed_body_map(body))
```
assaf758: why do you need to cast it to dict, this is what _apply_parsed_body_map returns

mlrun/serving/api_handler.py
```diff
@@ -455,18 +391,19 @@ def do(
...
+                        original_body=original_body,
                         body_params=body_params,
                         path_params=path_params,
                         query_params=query_params,
                         url_params=url_params,
                     )
 
-                    # Replace event body with unified context
                     if hasattr(event, "body"):
-                        event.body = request_context
```
assaf758, line 467: why do we need to check for body existence and branch accordingly setting the request_context in body or the event.
why not settle on always setting event.body to the request_context

assaf758 (commented):
looks good,
some comments inline

mlrun/serving/api_handler.py
jond01: Right

mlrun/serving/api_handler.py
jond01: Fixed

mlrun/serving/api_handler.py
jond01: Right, I simplified it. The condition was actually a dead code.
Edit: the code was updated, I don't know why it doesn't show it on the comment.

assaf758 (approved):

PR #9525, opened 2026-03-31
[Mounts] Fix auto_mount() ignoring server-configured S3 mount type when MLRUN_PVC_MOUNT is set

alxtkr77:
## Summary

- `auto_mount()` now checks the server-configured `auto_mount_type` before the client-side `MLRUN_PVC_MOUNT` env var
- When the server explicitly sets `auto_mount_type=s3` (or `secret_env`, `pvc`, etc.), that takes precedence over `MLRUN_PVC_MOUNT` on the client

## Root cause

On IG4 with external Jupyter, `MLRUN_PVC_MOUNT=mlrun-jupyter:/home/jovyan/` is set on the Jupyter node. `auto_mount()` checked this env var **before** the server-configured `auto_mount_type=s3`, so it always returned `mount_pvc` instead of `mount_s3`. Spark pods received PVC mounts instead of S3 credentials, causing `NoAuthWithAWSException`.

`try_auto_mount_based_on_config()` (used by the launcher) was not affected — it goes directly to the server config. The bug was only in the `auto_mount()` standalone function used in notebooks.

## Changes

- `mlrun/runtimes/mounts.py`: Moved the explicit `auto_mount_type` config check before the `MLRUN_PVC_MOUNT` env var check in `auto_mount()`
- `tests/runtimes/test_mounts.py`: Added test verifying S3 mount takes precedence over PVC env var

## Test plan

- [x] `pytest tests/runtimes/test_mounts.py` — all pass
- [x] `pytest tests/runtimes/test_base.py::TestAutoMount::test_auto_mount_function_with_pvc_config` — pass
- [ ] Manual test on vmdev212ig4 external Jupyter with `MLRUN_PVC_MOUNT` set

Reference: ML-12370

### DIFF
--- a/mlrun/runtimes/mounts.py
+++ b/mlrun/runtimes/mounts.py
@@ -382,19 +382,25 @@ def auto_mount(
             volume_mount_path=volume_mount_path,
             volume_name=volume_name or "shared-persistency",
         )
+    # When auto_mount_type is explicitly configured (not the default "auto"),
+    # honour it regardless of env variables like MLRUN_PVC_MOUNT.  This ensures
+    # that an admin-configured S3 or secret_env mount type takes precedence over
+    # a PVC env var that may be set on the client machine (e.g., external Jupyter).
+    # Lazy import to avoid circular dependency (pod.py imports mounts.py at module level).
+    from mlrun.runtimes.pod import AutoMountType, _filter_modifier_params
+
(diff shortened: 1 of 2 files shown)


liranbg (changes requested):
Fix PR description

assaf758 (approved):

PR #9063, opened 2025-12-16
[Model Monitoring] Optimize stream pod avoiding duplication in tsdb

royischoss:
### 📝 Description

Optimize model monitoring stream pod, avoiding working on duplicate events in TSDB steps

---

### 🛠️ Changes Made

This change includes:
1. Splitting the graph earlier between parquet to tsdb branches, avoiding flatten for events that target to the TSDB
2. Small change in the handling of NOP event
3. Mapping the first event of batch to the TSDB while this event hold all necessary data
4. v3io - include events table, added a MapFeatureNamesTSDB for some fields need to be covered

TDEngine and TimeScaleDB: 
image

V3io:
image

---

### ✅ Checklist
- [ ] I updated the documentation (if applicable)
- [x] I have tested the changes in this PR
- [x] I confirmed whether my changes are covered by system tests
  - [x] If yes, I ran all relevant system tests and ensured they passed before submitting this PR
  - [ ] I updated existing system tests and/or added new ones if needed to cover my changes
- [ ] If I introduced a deprecation:
  - [ ] I followed the [Deprecation Guidelines](./DEPRECATION.md)
  - [ ] I updated the relevant Jira ticket for documentation

---

### 🧪 Testing

Manual for v3io and TimeScaleDB, `test_app_flow` with ModelRunnerStep using both v3io and TimeScaleDB TSDB

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-11639 , https://iguazio.atlassian.net/browse/ML-9667
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

PR #9526, opened 2026-03-31
[Model Monitoring] Fix histogram data drift crash on empty sample data

gtopper:
### 📝 Description

A stream processing topology change in v1.11.0 (#9063) introduced a timing gap where TSDB data becomes available before Parquet data flushes. This exposed three latent bugs in the histogram data drift application (present since v1.7.0) that caused `KeyError` and `NoneType` crashes when the application was triggered before sample data was available. [ML-12380](https://iguazio.atlassian.net/browse/ML-12380)

---

### 🛠️ Changes Made

- Add `sample_df.empty` guard in `do_tracking()` to skip processing when no sample data is available for the given interval (mirrors the existing guard in the local `evaluate()` path)
- Iterate the intersection of `feature_stats` and `sample_df_stats` columns in `_compute_metrics_per_feature()` instead of assuming all reference features exist in the sample
- Guard `_get_shared_features_sample_stats()` against missing keys in `sample_df_stats`
- Return `event` from `_ApplicationErrorHandler.do()` to prevent `None` from propagating downstream in the serving graph

---

### ✅ Checklist

---

### 🧪 Testing

- 4 new unit tests added covering all fixed scenarios:
  - `TestEmptySampleDf::test_do_tracking_with_empty_sample_df` — verifies graceful handling of empty sample data
  - `TestMetricsPerFeature::test_compute_metrics_mismatched_features` — verifies intersection-based iteration when feature sets differ
  - `TestGetSharedFeaturesSampleStats::test_missing_keys` — verifies missing keys don't cause `KeyError`
  - `TestApplicationErrorHandler::test_do_returns_event` — verifies error handler returns the event
- All 18 tests in `test_histogram_data_drift.py` pass (14 existing + 4 new)
- Verified in k8s logs (project `jvlhtrcbaa`) that the production error matches the reproduced `KeyError: 'sepal_length_cm'`

Smoke passed – https://github.com/mlrun/mlrun/actions/runs/23795615062.

---

### 🔗 References
- Ticket link: [ML-12380](https://iguazio.atlassian.net/browse/ML-12380)
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes

**Root cause**: The regression trigger is commit #9063 which moved `FilterNOP` earlier in the stream processing pipeline, causing the TSDB branch to diverge before `flatten_events`/`MapFeatureNames`. This creates a window where TSDB has data but `ParquetTarget` hasn't flushed yet. The controller queries TSDB, finds data, triggers the app — but the app reads an empty `sample_df` from Parquet.

This PR does not revert the topology change. Instead, it adds defensive guards so the application gracefully skips processing when sample data is unavailable. The only consequence is that the first monitoring cycle after deployment may silently skip drift computation for some endpoints; subsequent cycles work normally once Parquet catches up.

The topology issue is non-deterministic (depends on flush timing), which explains why QA tests could pass on one RC and fail on another.

[ML-12380]: https://iguazio.atlassian.net/browse/ML-12380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ML-12380]: https://iguazio.atlassian.net/browse/ML-12380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/model_monitoring/applications/histogram_data_drift.py
+++ b/mlrun/model_monitoring/applications/histogram_data_drift.py
@@ -165,6 +165,11 @@ class HistogramDataDriftApplication(ModelMonitoringApplicationBase):
             monitoring_context.sample_df_stats
         )
         common_features = set(feature_stats.columns) & set(sample_df_stats.columns)
+        if common_features != set(feature_stats.columns):
+            monitoring_context.logger.warning(
+                "Some reference features are missing from the sample data",
+                missing_features=set(feature_stats.columns) - common_features,
+            )
         for feature_name in common_features:
             sample_hist = np.asarray(sample_df_stats[feature_name])
             reference_hist = np.asarray(feature_stats[feature_name])
--- a/mlrun/model_monitoring/applications/histogram_data_drift.py
+++ b/mlrun/model_monitoring/applications/histogram_data_drift.py
@@ -259,7 +259,7 @@ class HistogramDataDriftApplication(ModelMonitoringApplicationBase):
         return stats
 
     @staticmethod
-    def _get_common_features_sample_stats(
+    def _get_shared_features_sample_stats(
         monitoring_context: mm_context.MonitoringApplicationContext,
     ) -> mlrun.common.model_monitoring.helpers.FeatureStats:
         """
@@ -338,7 +338,7 @@ class HistogramDataDriftApplication(ModelMonitoringApplicationBase):
 
         if self._produce_plotly_artifact:
             self._log_plotly_table_artifact(
-                sample_set_statistics=self._get_common_features_sample_stats(
+                sample_set_statistics=self._get_shared_features_sample_stats(
                     monitoring_context
                 ),
                 inputs_statistics=monitoring_context.feature_stats,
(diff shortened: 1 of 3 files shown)


mlrun/model_monitoring/applications/histogram_data_drift.py
```diff
@@ -164,7 +164,8 @@ def _compute_metrics_per_feature(
         sample_df_stats = monitoring_context.dict_to_histogram(
             monitoring_context.sample_df_stats
         )
-        for feature_name in feature_stats:
+        common_features = set(feature_stats.columns) & set(sample_df_stats.columns)
```
royischoss, line 167: we might silence here a needed failure when there is a mismatch with different column name

royischoss (commented):
hey looks good small comment

mlrun/model_monitoring/applications/histogram_data_drift.py
gtopper: Added a warning.

royischoss (approved):
LGTM

PR #9557, opened 2026-04-11
[Bug] fix(project): re-apply parameters after project setup to fix git source ignoring params

### LINKED ISSUE #9275: [Bug]: project parameters ignored when using source as github repo.
### MLRun Version checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of MLRun CE.

### Reproducible Example

```python
    "project_name",
    parameters={
        "source": "git://github.com/my_repo#main",
        "default_image": "my_image",
    },
)


# and in project setup

def setup(project: mlrun.projects.MlrunProject) -> mlrun.projects.MlrunProject:
    source = project.get_param("source")
    secrets_file = project.get_param("secrets_file")
    default_image = project.get_param("default_image")
```

### Issue Description

The parameters passed in get_or_create_project are ignored by project setup. project.get_param() does not get them. 

this issue doesn't happen using source from s3:// only source from repo. 

### Expected Behavior

expected behavior is that source from repo would be the same from a parameter perspective as a source from s3. 

### Installation OS

Mac

### Installation Method

Kubernetes

### Python Version

'3.11.13'

### MLRun Version

'1.9.2'

### Additional Information

_No response_

hirenkumar-n-dholariya:
## Bug

Parameters passed to get_or_create_project() are ignored when loading from a git:// source. project.get_param() returns None
instead of the provided values.

## Root Cause

In load_project(), parameters are applied to project.spec.params BEFORE project.setup() is called. For git:// sources, setup()
re-initializes the project from the cloned repo, overwriting project.spec.params with values from the repo's project.yaml,
discarding the user-provided parameters.

This does not affect s3:// sources because setup() does not reinitialize the project spec in that case.

## Fix

Re-apply user-provided parameters after project.setup() completes:

  if parameters:
      for key, val in parameters.items():
          project.spec.params[key] = val
      if to_save:
          project.save()

This ensures get_param() returns correct values for all source types.

## Reproducible Example

      "project_name",
      parameters={
          "source": "git://github.com/my_repo#main",
          "default_image": "my_image",
      },
  )

  # Before fix: returns None
  # After fix: returns "my_image"
  project.get_param("default_image")

## Impact

- Fixes parameter passing for git:// sources
- No behavior change for s3:// or other sources
- Parameters applied twice is safe - second apply is identical values

## References

- Fixes #9275


yaelgen (changes requested):
Thanks for the contribution! 

I traced through the `load_project` flow for `git:// ` sources and want to make sure we're addressing the right root cause.
In the current code, user-provided parameters are applied to `project.spec.params` (line 431-434) before setup() runs (line 455). 
The setup() function receives the project object with those params already set - it doesn't re-read `project.yaml` or reset `project.spec.params` on its own.

Based on this, I wasn't able to identify where the params get overwritten at the framework level. Maybe some interaction with `get_or_create_project` when the project already exists in the DB from a previous run.
Would you be able to share a minimal reproducible example (the project_setup.py, project.yaml, and the calling code)? That would help us pinpoint whether the fix should be here or somewhere else in the flow.

Also, one concern with the fix: with `save=True`, the project is already saved twice (line 452 and inside `_run_project_setup` at line 653). Adding a third save could overwrite intentional transformations that setup() made to parameters, like in case a setup script reads a raw param and replaces it with an enriched value, re-applying the originals would undo that.

hirenkumar-n-dholariya:
> Thanks for the contribution!
> 
> I traced through the `load_project` flow for `git:// ` sources and want to make sure we're addressing the right root cause. In the current code, user-provided parameters are applied to `project.spec.params` (line 431-434) before setup() runs (line 455). The setup() function receives the project object with those params already set - it doesn't re-read `project.yaml` or reset `project.spec.params` on its own.
> 
> Based on this, I wasn't able to identify where the params get overwritten at the framework level. Maybe some interaction with `get_or_create_project` when the project already exists in the DB from a previous run. Would you be able to share a minimal reproducible example (the project_setup.py, project.yaml, and the calling code)? That would help us pinpoint whether the fix should be here or somewhere else in the flow.
> 
> Also, one concern with the fix: with `save=True`, the project is already saved twice (line 452 and inside `_run_project_setup` at line 653). Adding a third save could overwrite intentional transformations that setup() made to parameters, like in case a setup script reads a raw param and replaces it with an enriched value, re-applying the originals would undo that.

Thanks for the detailed trace-through, really helpful context!

**On the root cause**: 
You are right that for a fresh load_project call, project.spec.params is set before setup() runs and setup() doesn't reset it. The issue we observed specifically happens when get_or_create_project finds the project already exists in the DB from a previous run - in that path, the project is first hydrated from the DB (which already has its own params snapshot), and the user-provided params passed into the call get silently dropped or overwritten by the DB version before setup() ever sees them. That's the race condition we're trying to close.

# First run - creates project in DB with threshold=0.8
    "myproj", context="./",
    url="git://github.com/myorg/myproj.git",
    parameters={"threshold": 0.8}
)
print(project.params)  # correctly shows 0.8

# Second run - project now exists in DB
    "myproj", context="./",
    url="git://github.com/myorg/myproj.git",
    parameters={"threshold": 0.9}  # intended override
)
print(project.params)  # BUG: shows 0.8 (from DB), not 0.9

**On the double-save concern**: 
Valid point! To avoid undoing intentional setup() transformations, we could instead only apply caller-provided params if setup() hasn't already modified them. I would be happy to revise; could you advise on the intended contract for parameters in this flow?

yaelgen:
> > Thanks for the contribution!
> > I traced through the `load_project` flow for `git:// ` sources and want to make sure we're addressing the right root cause. In the current code, user-provided parameters are applied to `project.spec.params` (line 431-434) before setup() runs (line 455). The setup() function receives the project object with those params already set - it doesn't re-read `project.yaml` or reset `project.spec.params` on its own.
> > Based on this, I wasn't able to identify where the params get overwritten at the framework level. Maybe some interaction with `get_or_create_project` when the project already exists in the DB from a previous run. Would you be able to share a minimal reproducible example (the project_setup.py, project.yaml, and the calling code)? That would help us pinpoint whether the fix should be here or somewhere else in the flow.
> > Also, one concern with the fix: with `save=True`, the project is already saved twice (line 452 and inside `_run_project_setup` at line 653). Adding a third save could overwrite intentional transformations that setup() made to parameters, like in case a setup script reads a raw param and replaces it with an enriched value, re-applying the originals would undo that.
> 
> Thanks for the detailed trace-through, really helpful context!
> 
> **On the root cause**: You are right that for a fresh load_project call, project.spec.params is set before setup() runs and setup() doesn't reset it. The issue we observed specifically happens when get_or_create_project finds the project already exists in the DB from a previous run - in that path, the project is first hydrated from the DB (which already has its own params snapshot), and the user-provided params passed into the call get silently dropped or overwritten by the DB version before setup() ever sees them. That's the race condition we're trying to close.
> 
> # First run - creates project in DB with threshold=0.8
> project = mlrun.get_or_create_project( "myproj", context="./", url="git://github.com/myorg/myproj.git", parameters={"threshold": 0.8} ) print(project.params) # correctly shows 0.8
> 
> # Second run - project now exists in DB
> project = mlrun.get_or_create_project( "myproj", context="./", url="git://github.com/myorg/myproj.git", parameters={"threshold": 0.9} # intended override ) print(project.params) # BUG: shows 0.8 (from DB), not 0.9
> 
> **On the double-save concern**: Valid point! To avoid undoing intentional setup() transformations, we could instead only apply caller-provided params if setup() hasn't already modified them. I would be happy to revise; could you advise on the intended contract for parameters in this flow?


Thanks for the clarification and the detailed scenario!

I traced through the "second run" case against the code and wasn't able to reproduce the described behavior. In `get_or_create_project`, the first attempt calls `load_project(url="myproj")` (the name, not the git URL). Inside load_project, since `"://"` not in "myproj", it takes the DB path (line 409-415) and loads the project with the old params. But then user-provided parameters are applied on top (line 431-434), before setup() runs:

```
if parameters:
    for key, val in parameters.items():
        project.spec.params[key] = val
```
So the in-memory project returned should have threshold=0.9. The git URL isn't used at all in this path -`get_or_create_project` passes name (not url) as the URL argument in its first `load_project` call (line 536-537).

I also tested this locally - created a project with `parameters={"threshold": 0.8}`, then ran `get_or_create_project` again with `parameters={"threshold": 0.9}.` The project already existed in the DB with 0.8, but `print(project.params)` correctly returned 0.9. The DB still holds 0.8 since get_or_create_project uses save=False on the DB load path, but the returned object has the correct override.

One thing worth noting: since `save=False` is used (line 545), the updated params are only in memory - the DB retains the previous value. If downstream code (e.g., a remote workflow runner) re-reads from DB, it would get the stale value. But that's a different issue..
There are scenarios where params could get overwritten - but they'd come from the project's own files rather than the framework. For example, if project_setup.py explicitly reassigns params (e.g., project.spec.params = {...} or project.spec.params["threshold"] = some_default), or if project.yaml defines a params section that gets loaded before the user-provided values are applied. In those cases it would look like the framework is dropping params, but it's actually the project's setup logic overriding them.

Would you be able to share a minimal reproducible example that demonstrates the issue (the project_setup.py, project.yaml, and calling code)? That would help us pinpoint whether the overwrite is coming from the framework or from the project's own configuration.

hirenkumar-n-dholariya:
> > > Thanks for the contribution!
> > > I traced through the `load_project` flow for `git:// ` sources and want to make sure we're addressing the right root cause. In the current code, user-provided parameters are applied to `project.spec.params` (line 431-434) before setup() runs (line 455). The setup() function receives the project object with those params already set - it doesn't re-read `project.yaml` or reset `project.spec.params` on its own.
> > > Based on this, I wasn't able to identify where the params get overwritten at the framework level. Maybe some interaction with `get_or_create_project` when the project already exists in the DB from a previous run. Would you be able to share a minimal reproducible example (the project_setup.py, project.yaml, and the calling code)? That would help us pinpoint whether the fix should be here or somewhere else in the flow.
> > > Also, one concern with the fix: with `save=True`, the project is already saved twice (line 452 and inside `_run_project_setup` at line 653). Adding a third save could overwrite intentional transformations that setup() made to parameters, like in case a setup script reads a raw param and replaces it with an enriched value, re-applying the originals would undo that.
> > 
> > 
> > Thanks for the detailed trace-through, really helpful context!
> > **On the root cause**: You are right that for a fresh load_project call, project.spec.params is set before setup() runs and setup() doesn't reset it. The issue we observed specifically happens when get_or_create_project finds the project already exists in the DB from a previous run - in that path, the project is first hydrated from the DB (which already has its own params snapshot), and the user-provided params passed into the call get silently dropped or overwritten by the DB version before setup() ever sees them. That's the race condition we're trying to close.
> > # First run - creates project in DB with threshold=0.8
> > project = mlrun.get_or_create_project( "myproj", context="./", url="git://github.com/myorg/myproj.git", parameters={"threshold": 0.8} ) print(project.params) # correctly shows 0.8
> > # Second run - project now exists in DB
> > project = mlrun.get_or_create_project( "myproj", context="./", url="git://github.com/myorg/myproj.git", parameters={"threshold": 0.9} # intended override ) print(project.params) # BUG: shows 0.8 (from DB), not 0.9
> > **On the double-save concern**: Valid point! To avoid undoing intentional setup() transformations, we could instead only apply caller-provided params if setup() hasn't already modified them. I would be happy to revise; could you advise on the intended contract for parameters in this flow?
> 
> Thanks for the clarification and the detailed scenario!
> 
> I traced through the "second run" case against the code and wasn't able to reproduce the described behavior. In `get_or_create_project`, the first attempt calls `load_project(url="myproj")` (the name, not the git URL). Inside load_project, since `"://"` not in "myproj", it takes the DB path (line 409-415) and loads the project with the old params. But then user-provided parameters are applied on top (line 431-434), before setup() runs:
> 
> ```
> if parameters:
>     for key, val in parameters.items():
>         project.spec.params[key] = val
> ```
> 
> So the in-memory project returned should have threshold=0.9. The git URL isn't used at all in this path -`get_or_create_project` passes name (not url) as the URL argument in its first `load_project` call (line 536-537).
> 
> I also tested this locally - created a project with `parameters={"threshold": 0.8}`, then ran `get_or_create_project` again with `parameters={"threshold": 0.9}.` The project already existed in the DB with 0.8, but `print(project.params)` correctly returned 0.9. The DB still holds 0.8 since get_or_create_project uses save=False on the DB load path, but the returned object has the correct override.
> 
> One thing worth noting: since `save=False` is used (line 545), the updated params are only in memory - the DB retains the previous value. If downstream code (e.g., a remote workflow runner) re-reads from DB, it would get the stale value. But that's a different issue.. There are scenarios where params could get overwritten - but they'd come from the project's own files rather than the framework. For example, if project_setup.py explicitly reassigns params (e.g., project.spec.params = {...} or project.spec.params["threshold"] = some_default), or if project.yaml defines a params section that gets loaded before the user-provided values are applied. In those cases it would look like the framework is dropping params, but it's actually the project's setup logic overriding them.
> 
> Would you be able to share a minimal reproducible example that demonstrates the issue (the project_setup.py, project.yaml, and calling code)? That would help us pinpoint whether the overwrite is coming from the framework or from the project's own configuration.

Thank you for testing this locally and for the detailed trace-through!

You are right - looking at the code again with your explanation, lines 431-434 do correctly apply the caller-provided parameters before setup() runs, and get_or_create_project passes the name (not the git URL) to load_project on the DB path, so the git URL is not involved in that second-run scenario.

I believe the original issue in #9275 was likely caused by the project's own project_setup.py overriding params (e.g. project.spec.params["key"] = default), rather than a framework-level bug.

Given this, I will close this PR as the fix is not needed at the framework level. The correct guidance for users facing this issue would be to check their project_setup.py for any unconditional param assignments, and instead use:

    if "key" not in project.spec.params:
        project.spec.params["key"] = default_value

This preserves caller-provided values while still setting defaults in setup().

Thanks again for the thorough review - I learned a lot about the load_projectflow from this discussion.

hirenkumar-n-dholariya:
Hi @yaelgen 
Hope you are doing well. Could you please take a look at my above comments and give your suggestions/approval to merge the PR.

https://github.com/mlrun/mlrun/pull/9557#issuecomment-4255725728

PR #9558, opened 2026-04-12
[Pipelines] Fix step-to-run correlation for long pipeline names

liranbg:
### 📝 Description

Fix step-to-run correlation failure when pipeline + step names produce pod hostnames exceeding the 63-character kubelet truncation limit.

When KFP pipeline names are long, the kubelet truncates the pod hostname to 63 characters. The `runner_pod` label was set using `socket.gethostname()`, which returns this truncated value, causing the API server to fail matching pipeline steps to their MLRun runs (missing `run_uid` in the pipeline graph).

The fix injects `MLRUN_POD_NAME` via the Kubernetes downward API (`metadata.name`) into pipeline step pods, and uses it (with `socket.gethostname()` as fallback) when setting the `runner_pod` label.

---

### 🛠️ Changes Made

- **`mlrun/__main__.py`**: Use `MLRUN_POD_NAME` env var (falling back to `socket.gethostname()`) for the `runner_pod` label
- **`pipeline-adapters/mlrun-pipelines-kfp-v1-8/src/mlrun_pipelines/ops.py`**: Inject `MLRUN_POD_NAME` env var from Kubernetes downward API (`metadata.name`) into pipeline step containers
- **`pipeline-adapters/mlrun-pipelines-kfp-v2/src/mlrun_pipelines/ops.py`**: Same injection for KFP v2 adapter using `kfp_k8s.use_field_path_as_env`
- **`pipeline-adapters/*/pyproject.toml`**: Version bump to 0.7.0 for all three pipeline adapter packages
- **`tests/system/runtimes/test_kfp.py`**: Add system test `test_kfp_long_pipeline_name_steps_have_run_uid` that runs a pipeline with long names and asserts both steps have `run_uid`
- **`tests/system/base.py`**: Skip `igz_mgmt.Client` initialization when `V3IO_ACCESS_KEY` is not configured (supports igz4 systems)

---

### ✅ Checklist

---

### 🧪 Testing
  
  

- Ran `test_kfp_long_pipeline_name_steps_have_run_uid` system test against vmdev103ig4 lab — pipeline completed successfully with both steps having `run_uid` populated
- The test uses a pipeline name of 39 chars + step names of 35/30 chars, ensuring the combined pod hostname exceeds 63 chars

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-12372
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes

- The `MLRUN_POD_NAME` env var is only injected into KFP pipeline step pods (not regular MLRun jobs), and the fallback to `socket.gethostname()` ensures backward compatibility for environments where the env var is not set.
- The `tests/system/base.py` change is a minor fix to support running system tests against igz4 systems where `V3IO_ACCESS_KEY` is not needed.

### DIFF
--- a/mlrun/__main__.py
+++ b/mlrun/__main__.py
@@ -294,8 +294,11 @@ def run(
 
     if workflow:
         runobj.metadata.labels[mlrun_constants.MLRunInternalLabels.workflow] = workflow
+        # Use the full pod name (via MLRUN_POD_NAME downward API env var) for
+        # unambiguous step-to-run correlation. socket.gethostname() is truncated
+        # to 63 chars by the kubelet, which can cause lookup mismatches.
         runobj.metadata.labels[mlrun_constants.MLRunInternalLabels.runner_pod] = (
-            socket.gethostname()
+            environ.get("MLRUN_POD_NAME", socket.gethostname())
         )
 
     if db:
--- a/pipeline-adapters/mlrun-pipelines-kfp-common/pyproject.toml
+++ b/pipeline-adapters/mlrun-pipelines-kfp-common/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "mlrun-pipelines-kfp-common"
-version = "0.6.4"
+version = "0.7.0"
 description = "MLRun Pipelines adapter package for providing KFP common functionality"
 readme = "README.md"
 requires-python = ">=3.11, <3.12"
--- a/pipeline-adapters/mlrun-pipelines-kfp-v1-8/pyproject.toml
+++ b/pipeline-adapters/mlrun-pipelines-kfp-v1-8/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "mlrun-pipelines-kfp-v1-8"
-version = "0.6.4"
+version = "0.7.0"
 description = "MLRun Pipelines adapter package for providing KFP 1.8 compatibility"
 readme = "README.md"
 requires-python = ">=3.11, <3.12"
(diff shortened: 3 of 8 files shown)


elbamit (approved):
LGTM. This won't fix already existing pipelines, only new ones. But I think that's reasonable.

mlrun/__main__.py
```diff
@@ -294,8 +294,11 @@ def run(
 
     if workflow:
         runobj.metadata.labels[mlrun_constants.MLRunInternalLabels.workflow] = workflow
+        # Use the full pod name (via MLRUN_POD_NAME downward API env var) for
+        # unambiguous step-to-run correlation. socket.gethostname() is truncated
+        # to 63 chars by the kubelet, which can cause lookup mismatches.
         runobj.metadata.labels[mlrun_constants.MLRunInternalLabels.runner_pod] = (
-            socket.gethostname()
+            environ.get("MLRUN_POD_NAME", socket.gethostname())
```
yaelgen, line 301: Nice, what about old runs already stored in the DB with a truncated hostname? After the update, the read side will search by the full pod name, but the DB still has the truncated version. Maybe we can add a fallback that tries a prefix match on the read side too so if the lookup by full name returns no results, retry with the first 63 characters.

yaelgen: I see Amit's comment now, and actually a read-side fallback using truncation is possible but risky, since if the pipeline name is long enough to consume most of the 63 chars, multiple steps could produce the same truncated hostname, causing the wrong run to be matched to a step. 
So it's reasonable to scope this to new pipelines only. Worth noting in the PR description.

Also, model monitoring also sets runner_pod via socket.gethostname() in context.py and base.py. It's a different use case (not kfp related), so not sure if we care really, but if we want to be consistent and avoid a similar issue surfacing there in the future, we could apply the same environ.get("MLRUN_POD_NAME", socket.gethostname()) pattern.

liranbg: Indeed there's no neat way to track old pipelines - but for new one, the fix does its magic.

PR #9568, opened 2026-04-13
[Artifacts] Add CodeArtifact type for function/workflow code

alxtkr77:
## 📝 Description

Introduce `CodeArtifact` (kind=`code`) for registering function/workflow source
code as MLRun artifacts. Lays the foundation for ML-11980 / ML-11981 (loading
functions/workflows from artifacts via `store://`).

The new artifact type carries metadata the consumer needs to resolve and run
the code: `language`, `code_type` (`function`|`workflow`), and `requirements`
(a list of pip-style dependency strings). Application runtime is migrated to
use `log_code_file()` so code it auto-uploads is now tagged with the proper
artifact kind.

## 🛠️ Changes Made

- New `mlrun/artifacts/code.py`: `CodeArtifact`, `CodeArtifactSpec`, `CodeArtifactCodeType` enum.
- Register `"code": CodeArtifact` in `mlrun/artifacts/manager.py`.
- Export `CodeArtifact`, `CodeArtifactSpec`, `CodeArtifactCodeType` from `mlrun/artifacts/__init__.py`.
- Add `code` to `ArtifactCategories` enum and update `from_kind()`, `to_kinds_filter()`, `all()`.
- New `log_code_file()` SDK method on `MlrunProject` and `MLClientCtx`.
- Application runtime (`_upload_source_as_artifact`) now calls `log_code_file(language="python", code_type="function")` instead of the generic `log_artifact`.

## ✅ Checklist

- [x] `make fmt` clean
- [x] `make lint` clean
- [x] Unit tests pass (151 tests across `tests/artifacts/test_artifacts.py` + `tests/runtimes/test_application.py`)
- [x] No secrets in diff
- [ ] Documentation — tracked separately in **ML-12364** (CodeArtifact documentation and tutorial)

## 🧪 Testing

Unit tests added in `tests/artifacts/test_artifacts.py`:
- CodeArtifact kind, spec fields (`language`, `code_type`, `requirements`)
- `code_type` default (`function`), enum validation
- `log_code_file()` on both `MlrunProject` and `MLClientCtx`
- Serialization round-trip
- `ArtifactCategories.code` filter behavior

Application-runtime test updated to expect `kind='code'` on the auto-uploaded artifact.

## 🔗 References

- Jira: [ML-12337](https://iguazio.atlassian.net/browse/ML-12337) — New type of artifact: `CodeArtifact`
- Jira: [ML-12358](https://iguazio.atlassian.net/browse/ML-12358) — Implement CodeArtifact type, `log_code_file()` SDK, and application-runtime integration (this PR)
- Blocks: ML-11980 (`store://` function loading) and ML-11981 (`store://` workflow loading)
- HLD: https://iguazio.atlassian.net/wiki/spaces/Dataflows/pages/853508103

## 🚨 Breaking Changes

No. New artifact kind is additive; application-runtime switch from
`log_artifact` to `log_code_file` produces an artifact with `kind='code'`
instead of the generic default, but existing artifacts are untouched (no
migration performed, per ML-12337 design).

[ML-12337]: https://iguazio.atlassian.net/browse/ML-12337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/artifacts/code.py
+++ b/mlrun/artifacts/code.py
@@ -16,6 +16,22 @@ import mlrun.common.types
 
 from .base import Artifact, ArtifactSpec
 
+_PYTHON_SUFFIXES = (".py", ".ipynb")
+
+
+def _derive_language_from_path(path: str | None) -> str | None:
+    """Derive a language value from a file path's suffix.
+
+    Returns ``"python"`` for ``.py``/``.ipynb``, ``""`` for archives and unknown
+    suffixes (caller knows there's a path but can't infer the language), and
+    ``None`` when no path is available (nothing to infer from).
+    """
+    if not path:
+        return None
+    if path.lower().endswith(_PYTHON_SUFFIXES):
+        return "python"
+    return ""
+
 
 class CodeArtifactCodeType(mlrun.common.types.StrEnum):
     function = "function"
@@ -87,7 +103,13 @@ class CodeArtifact(Artifact):
         :param format:       Optional file format
         :param target_path:  Absolute target path
         :param src_path:     Path to the local code file or archive
-        :param language:     Programming language and version (e.g. "python:3.9")
+        :param language:     Programming language (e.g. ``"python"``).
+                             Free-text advisory metadata — no validation or
+                             enforcement is applied, and the value is not consulted
+                             at resolution or execution time.
+                             When ``None``, derived from the ``target_path`` (or
+                             ``src_path``) suffix: ``.py``/``.ipynb`` → ``"python"``,
+                             archives/unknown → ``""``, no path → stays ``None``.
         :param code_type:    Type of code: "function" or "workflow" (default: "function")
         :param requirements: List of dependency strings (e.g. ["pandas>=2.0", "numpy"])
         """
@@ -99,6 +121,8 @@ class CodeArtifact(Artifact):
             src_path=src_path,
             **kwargs,
         )
+        if language is None:
+            language = _derive_language_from_path(target_path or src_path)
         self.spec.language = language
         self.spec.code_type = CodeArtifactCodeType(
             code_type or CodeArtifactCodeType.function
--- a/mlrun/artifacts/code.py
+++ b/mlrun/artifacts/code.py
@@ -16,22 +16,6 @@ import mlrun.common.types
 
 from .base import Artifact, ArtifactSpec
 
-_PYTHON_SUFFIXES = (".py", ".ipynb")
-
-
-def _derive_language_from_path(path: str | None) -> str | None:
-    """Derive a language value from a file path's suffix.
-
-    Returns ``"python"`` for ``.py``/``.ipynb``, ``""`` for archives and unknown
-    suffixes (caller knows there's a path but can't infer the language), and
-    ``None`` when no path is available (nothing to infer from).
-    """
-    if not path:
-        return None
-    if path.lower().endswith(_PYTHON_SUFFIXES):
-        return "python"
-    return ""
-
 
 class CodeArtifactCodeType(mlrun.common.types.StrEnum):
     function = "function"
@@ -146,3 +130,20 @@ class CodeArtifact(Artifact):
     @spec.setter
     def spec(self, spec):
         self._spec = self._verify_dict(spec, "spec", CodeArtifactSpec)
+
+
+_PYTHON_SUFFIXES = (".py", ".ipynb")
+
+
+def _derive_language_from_path(path: str | None) -> str | None:
+    """Derive a language value from a file path's suffix.
+
+    Returns ``"python"`` for ``.py``/``.ipynb``, ``""`` for archives and unknown
+    suffixes (caller knows there's a path but can't infer the language), and
+    ``None`` when no path is available (nothing to infer from).
+    """
+    if not path:
+        return None
+    if path.lower().endswith(_PYTHON_SUFFIXES):
+        return "python"
+    return ""
--- a/mlrun/artifacts/code.py
+++ b/mlrun/artifacts/code.py
@@ -1,0 +1,114 @@
+# Copyright 2026 Iguazio
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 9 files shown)


tests/artifacts/test_artifacts.py
```diff
@@ -769,3 +769,155 @@ def test_artifact_producer_parse_uri(uri, expected_parsed_result):
...
+            artifact.spec.code_type
+            == mlrun.artifacts.code.CodeArtifactCodeType.function
+        )
+        assert artifact.spec.language is None
+        assert artifact.spec.requirements is None
+
+    def test_create_with_language_and_code_type(self):
+        artifact = mlrun.artifacts.CodeArtifact(
+            key="my-code",
+            language="python:3.9",
```
liranbg, line 788: python:3.9 is deprecated. 
I know this is just a free text, but worth putting something that is forever green? `python`. also, why specifying the version makes sense?

liranbg (commented):
minor

tests/artifacts/test_artifacts.py
alxtkr77: To have version tag is part of the [ML-12337] requirements. @theSaarco , any comments on that?

mlrun/artifacts/code.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def __init__(
+        self,
+        key=None,
+        body=None,
+        format=None,
+        target_path=None,
+        src_path=None,
+        language=None,
+        code_type=None,
```
theSaarco:
```suggestion
        code_type: CodeArtifactCodeType | None =None,
```

Since you've defined this class, why not use it?
Looking at the code, it looks like you always assume it's a string, but why not accept the enum as well?
In general, I'm missing type hints here (and in the spec as well).

tests/artifacts/test_artifacts.py
theSaarco: Since the artifact points at a fixed piece of code, it makes sense to note what version of the language it supports. For example, your artifact's code may run on py3.11 but not support 3.9.
This is completely the user's choice - this is not used to deploy python or anything, it is just a descriptive field. So, it can be `python` and it can be `python:3.9` (and it can be `biton` for all we care). Since this is a test we can go both ways.
I wouldn't use 3.9 anyway since it's already deprecated.

theSaarco (commented):
Minor comment (lack of type hints).

mlrun/artifacts/code.py
```diff
@@ -0,0 +1,148 @@
...
+
+    Returns ``"python"`` for ``.py``/``.ipynb``, ``""`` for archives and unknown
+    suffixes (caller knows there's a path but can't infer the language), and
+    ``None`` when no path is available (nothing to infer from).
+    """
+    if not path:
+        return None
+    if path.lower().endswith(_PYTHON_SUFFIXES):
+        return "python"
+    return ""
```
liranbg: private to last

mlrun/runtimes/nuclio/application/application.py
```diff
@@ -1165,11 +1165,13 @@ def _upload_source_as_artifact(self) -> tuple[str | None, str | None]:
...
 
-        # Upload the file as an artifact to an internal path with system-generated label
+        # Upload the file as a code artifact to an internal path with system-generated label
         try:
-            artifact = project.log_artifact(
-                item=artifact_key,
+            artifact = project.log_code_file(
+                key=artifact_key,
                 local_path=source,
+                language="python",
```
liranbg: @Yacouby - is this flow only bound to python only? can't we use it to deploy other programming language as well? e.g deploy shell scripts?

tests/artifacts/test_artifacts.py
```diff
@@ -769,3 +769,189 @@ def test_artifact_producer_parse_uri(uri, expected_parsed_result):
         deepdiff.DeepDiff(parsed_result, expected_parsed_result, ignore_order=True)
         == {}
     )
+
+
+class TestCodeArtifact:
+    def test_create_with_defaults(self):
+        artifact = mlrun.artifacts.CodeArtifact(key="my-code")
```
liranbg: avoid using classes for pytest.

mlrun/runtimes/nuclio/application/application.py
alxtkr77: removing - no need to specify language here, you are right

liranbg (approved):

mlrun/projects/project.py
```diff
@@ -1782,6 +1783,66 @@ def log_dataset(
...
         return item
 
+    def log_code_file(
+        self,
+        key,
+        local_path=None,
+        body=None,
+        tag="",
+        artifact_path=None,
+        upload=None,
```
Yacouby: Shoudln't this default be True and not None ?

mlrun/execution.py
```diff
@@ -818,6 +819,70 @@ def log_dataset(
...
+        self,
+        key,
+        local_path=None,
+        body=None,
+        tag="",
+        artifact_path=None,
+        upload=True,
+        labels=None,
+        target_path="",
+        db_key=None,
```
Yacouby, line 833: This param is missing from `MlrunProject.log_code_file`, is this intentional ?

mlrun/projects/project.py
```diff
@@ -1782,6 +1783,66 @@ def log_dataset(
...
+        labels=None,
+        target_path="",
+        language=None,
+        code_type: str | mlrun.artifacts.code.CodeArtifactCodeType | None = None,
```
Yacouby: better to import it and use it directly

mlrun/execution.py
```diff
@@ -818,6 +819,70 @@ def log_dataset(
...
+        language=None,
+        code_type: str | mlrun.artifacts.code.CodeArtifactCodeType | None = None,
```
Yacouby: same as https://github.com/mlrun/mlrun/pull/9568/changes#r3131132487

mlrun/runtimes/nuclio/application/application.py
Yacouby: @liranbg It accepts any file extension

Yacouby (changes requested):
Suggestions

Yacouby (approved):

PR #9572, opened 2026-04-14
[Alerts] Add cooldown period support for auto-reset alerts

Yacouby:
### 📝 Description
This PR adds background periodic task that resets alerts whose cooldown has ended, so they return to an inactive state and can fire again.

Previously, auto-reset could run as soon as a notification was sent; with `cooldown_period`, the alert stays active until cooldown elapses. 
Manual reset still clears cooldown immediately when appropriate.

---

### 🛠️ Changes Made

- DB: `list_alerts_pending_cooldown_reset` on the DB interface + SQL implementation to list active alerts due for cooldown reset.
- Alerts CRUD: `reset_cooled_down_alerts` to process pending rows and reset via existing reset logic.
- Alerts service: Start `run_function_periodically` on startup (with `cooldown_reset_interval`) calling `_reset_cooled_down_alerts`; and tolerate failures without crashing the loop.
-  Cooldown "0" fix: cooldown_period="0"/"0s" is now treated as no cooldown (immediate reset), consistent with the feature spec. Validation no longer rejects it.

---

### ✅ Checklist

---

### 🧪 Testing

- Unit tests: Added for CRUD and DB.
- System tests: `test_alert_cooldown_auto_reset` (active during cooldown, second event during cooldown, then inactive after cooldown + periodic interval margin), 
- `test_alert_cooldown_manual_reset` (manual reset clears cooldown and allows re-trigger).
- Manual: Deploy server changes to vmdev, trigger an alert with reset_policy=auto and cooldown_period, confirm pod logs show periodic task startup and state transitions match API.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-12299
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


server/py/services/alerts/crud/alerts.py
```diff
@@ -664,10 +676,16 @@ def reset_alert(
                 f"Alert {name} for project {project} does not exist"
             )
 
-        # MANUAL and cooldown alerts track the last activation so it can be marked as resolved on reset
+        # MANUAL and real-cooldown (> 0) alerts track the last activation so it can be marked as resolved on reset
```
TomerShor, line 679: what is `real-cooldown`?

server/py/services/alerts/crud/alerts.py
```diff
@@ -686,6 +704,34 @@ def reset_alert(
         self._get_alert_state_cached().cache_remove(session, alert.id)
         self._clear_alert_states(alert.id)
 
+    def reset_cooled_down_alerts(self, session: sqlalchemy.orm.Session) -> None:
+        """Reset alerts whose cooldown period has elapsed."""
+        alerts = (
```
TomerShor: for explicitness sake
```suggestion
        alerts_to_reset = (
```

TomerShor (commented):
Looking great!
really minor comments

server/py/services/alerts/crud/alerts.py
Yacouby: cooldown period which is > 0... the comment clarifies that only alerts with an actual positive cooldown duration (not zero) need to track the last activation for resolving on reset

TomerShor (approved):

PR #9574, opened 2026-04-15
[Tests] Fix flaky TestMonitoringAppFlow::test_app_flow

assaf758:
### 📝 Description
`test_app_flow[True-True]` intermittently fails with `AssertionError: last_request is None`
because monitoring infra functions (stream/controller/writer) were still being deployed in the background
while inference was already running, causing events to be silently dropped.

---

### 🛠️ Changes Made
- Add `wait_for_deployment=True` to `enable_model_monitoring` in `_submit_controller_and_deploy_writer`
  so MM infra is fully up before inference starts.
- Move `_submit_controller_and_deploy_writer` into the `ThreadPoolExecutor` alongside
  `_set_and_deploy_monitoring_apps` and `_deploy_model_serving`, so all three deploy
  concurrently instead of sequentially, keeping setup time low.

---

### ✅ Checklist
- [x] Please run [Smoke-tests](https://github.com/mlrun/mlrun/actions/workflows/system-tests-opensource.yml) workflow, providing it the PR number as input (upon success, the workflow run will add label "Smoke tests: Pass" to the RP)

---

### 🧪 Testing
Test-only change. The fix addresses a race condition observed in CI run
[24426761963](https://github.com/mlrun/mlrun/actions/runs/24426761963).

---

### 🔗 References
- Ticket link:
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?
- [x] No

---

### 🔍️ Additional Notes
Most other tests in the same file already pass `wait_for_deployment=True` — this aligns
`_submit_controller_and_deploy_writer` with that pattern.

### DIFF
--- a/tests/system/model_monitoring/test_app.py
+++ b/tests/system/model_monitoring/test_app.py
@@ -856,16 +856,18 @@ class TestMonitoringAppFlow(TestMLRunSystemModelMonitoring, _V3IORecordsChecker)
         self._log_model(with_training_set)
 
         with concurrent.futures.ThreadPoolExecutor() as executor:
-            executor.submit(
+            infra_future = executor.submit(
                 self._submit_controller_and_deploy_writer,
                 _DefaultDataDriftAppData in self.apps_data,
             )
-            executor.submit(self._set_and_deploy_monitoring_apps)
-            future = executor.submit(
+            monitoring_apps_future = executor.submit(self._set_and_deploy_monitoring_apps)
+            serving_future = executor.submit(
                 self._deploy_model_serving, with_training_set, with_model_runner
             )
 
-        serving_fn = future.result()
+        infra_future.result()
+        monitoring_apps_future.result()
+        serving_fn = serving_future.result()
         self._add_error_alert()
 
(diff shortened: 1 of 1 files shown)


tests/system/model_monitoring/test_app.py
jond01, line 869: Since we now wait for the infra deployment - we can be more informative with deployment errors and not continue the test when it fails:
```suggestion
with concurrent.futures.ThreadPoolExecutor() as executor:
    infra_future = executor.submit(
        self._submit_controller_and_deploy_writer,
        _DefaultDataDriftAppData in self.apps_data,
    )
    executor.submit(self._set_and_deploy_monitoring_apps)
    future = executor.submit(
        self._deploy_model_serving, with_training_set, with_model_runner
    )

infra_future.result()
```

jond01 (approved):
Great idea!
I think it was originally the order due to constraints about infra first, before any MM serving function could be deployed. The test makes more sense now.

I left a small suggestion.

jond01 (approved):

PR #9577, opened 2026-04-15
[Docs] Sort the vale ignore list

jond01:
Running `make lint-docs` now checks the sort of the items in the file `.github/styles/MLRun/ignore.txt`.

Add a new target `make sort-vale-ignore` to ease the sort.


rokatyy (approved):
Nice!

PR #9592, opened 2026-04-20
[Runtimes] Add async handler support for job functions

jond01:
## Summary

Implements [ML-12442](https://iguazio.atlassian.net/browse/ML-12442): Support jobs async handler.

MLRun job handlers can now be defined as `async def` functions. The runtime automatically detects when a handler returns a coroutine and drives it to completion before committing the run result. No user-facing API changes are required.

```python
async def fetch_data(context: mlrun.MLClientCtx) -> int:
    context.logger.info("Async handler started")
    await asyncio.sleep(1)  # simulate async I/O
    result_value = 42
    context.log_result("async_result", result_value)
    return result_value

# Run exactly as a sync handler — no changes needed
run = fn.run()
print(run.output("async_result"))  # 42
```

## Changes

**Core implementation (`mlrun/runtimes/local.py`)**
- `_run_async_handler(coro)` — drives a coroutine to completion. Handles two execution contexts:
  - No running loop (K8s pod, plain Python): delegates to `asyncio.run()`.
  - Running loop (Jupyter / Tornado): submits `asyncio.run()` to a `ThreadPoolExecutor(max_workers=1)` in a new thread, giving identical blocking semantics without nesting loops (same pattern as Django's `asgiref.sync.async_to_sync`).
- `_normalize_handler_output(fn)` — decorator applied once to the handler before the packagers branch in `exec_from_params`: awaits coroutines via `_run_async_handler`, raises `MLRunRuntimeError` for sync/async generators (not supported as handler return types), passes everything else through unchanged.
  - Applied as a decorator so both the packagers-enabled and plain execution paths share the same async/generator handling without duplication.
  - Uses `inspect.iscoroutine()` (not `asyncio.iscoroutine()`) to avoid a Python 3.11 quirk where `asyncio.iscoroutine()` returns `True` for legacy generator objects.

**Documentation (`docs/concepts/submitting-tasks-jobs-to-functions.md`)**
- Added *Async handlers* section with usage example and a note that sync/async generators are not supported as handler return types.
- Added *Handlers inside a class* section documenting the `ClassName::method_name` handler syntax, including a subsection on passing constructor arguments via the `_init_args` key in `params`.

**Unit tests (`tests/runtimes/`)**
- `assets/async_handlers.py` — handler fixtures: `async_handler`, `async_handler_with_error`, `sync_handler`, `sync_generator_handler`, `async_generator_handler`, `AsyncHandlerClass` (instance/classmethod/staticmethod), `InitArgsHandlerClass`, `SyncInitArgsHandlerClass`.
- `test_async_handler.py` — 25 tests covering:
  - Async handler completes and logs output (both execution paths: `packagers_enabled=False/True`).
  - Sync handler regression (both paths).
  - Exception propagation from async handler.
  - Sync and async generator returns raise `MLRunRuntimeError` matching `"generator"`.
  - Full-stack Jupyter path: `_launch` called from within a running event loop exercises the `ThreadPoolExecutor` branch.
  - Unit tests for `_run_async_handler` directly (value return, exception propagation, running-loop path).
  - Async instance method, classmethod, and staticmethod handlers inside a class.
  - Constructor argument injection via `_init_args` (async and sync).

**System tests (`tests/system/runtimes/`)**
- `assets/async_handler.py` — `fetch_data` async handler with type hints.
- `test_kubejob.py` — `test_async_handler_completes` parametrized over `local=[True, False]`, covering both local and remote K8s execution.

## Test plan

- [x] Unit tests: `pytest tests/runtimes/test_async_handler.py -v` — 25 passed
- [x] System tests (requires live MLRun CE env): `pytest tests/system/runtimes/test_kubejob.py::TestKubejobRuntime::test_async_handler_completes -v`
- [x] Confirm sync handler regression tests still pass (covered by `test_sync_handler_unaffected`)
- [x] Confirm generator handlers raise `MLRunRuntimeError` with a clear message

[ML-12442]: https://iguazio.atlassian.net/browse/ML-12442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/runtimes/local.py
```diff
@@ -42,6 +42,7 @@
 from ..model import RunObject
 from ..utils import get_handler_extended, get_in, logger, set_paths
 from ..utils.clones import extract_source
+from ..utils.helpers import _resolve_handler_output
```
liranbg, line 45: If that's a public function - change its signature. Avoid importing internal / private functions from other modules

mlrun/utils/helpers.py
```diff
@@ -1938,6 +1941,63 @@ async def run_in_threadpool(func, *args, **kwargs):
     return await asyncio.to_thread(func, *args, **kwargs)
 
 
+def _run_async_handler(coro: Coroutine[Any, Any, _T]) -> _T:
```
liranbg, line 1944: Private functions are expected to be below public functions

tests/runtimes/assets/async_handlers.py
```diff
@@ -0,0 +1,48 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+
+
+async def async_handler(context):
+    """Async handler: simulates I/O, logs a result, and returns a value."""
+    await asyncio.sleep(0)  # yield to event loop to prove cooperative scheduling works
```
liranbg: I think the await on asyncio is pretty self explanatory no need to add the comment

If you do decide do it, then since comment is long, I suggest moving it atop the function call and not inline

tests/runtimes/test_async_handler.py
```diff
@@ -0,0 +1,150 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Unit tests for async job handler support (ML-11839).
+
+Tests cover:
```
liranbg: Why do we need that here? Looks like ai slop. Please unslop and also, keep in mind these things gets stale pretty quick

tests/runtimes/test_async_handler.py
```diff
@@ -0,0 +1,150 @@
...
+from mlrun.utils.helpers import _run_async_handler
+
+assets_path = pathlib.Path(__file__).parent / "assets"
+_HANDLER_FILE = str(assets_path / "async_handlers.py")
+
+_parametrize_packagers = pytest.mark.parametrize("packagers_enabled", [False, True])
+
+
+@pytest.fixture(autouse=True)
+def _restore_packagers_enabled():
```
liranbg, line 31: Why private ?

liranbg (changes requested):

mlrun/runtimes/local.py
jond01: It is private in the sense of external callers, but not inside mlrun itself. I think it's legitimate, and it's not intended to be public.

tests/runtimes/test_async_handler.py
jond01: In this case, the fixture is not meant to be called by a specific test, but is applied automatically in this test module.

tests/runtimes/test_async_handler.py
jond01: Not a slop, I reviewed it, but yes, it can be removed.

tests/runtimes/assets/async_handlers.py
jond01: Removed

mlrun/utils/helpers.py
jond01: It is not followed in this module, and this function is close in its nature to `run_in_threadpool` which is above it.
What do you think about opening a new module under `mlrun/utils/`?

mlrun/runtimes/local.py
liranbg: Id move the `_resolve_handler_output` to runtimes/ and change its signature to `resolve_handler_output` no reason for it to be "utils.helpers" - utils-helpers is not "anyone's helper" but rather "utilities helpers" - could be and aim to be wrapping utilities functions. for runtime-specific functions - use its module.

and, there's no reason to hold private functions and import them from other modules.

mlrun/utils/helpers.py
liranbg: > It is not followed in this module

that's not a good reason to add more noise, right?

> opening a new module

sure. if you plan on moving more funcitons from helpers.py to this module - do it on different pr to avoid mixing refactor+feature on same one

guy1992l (changes requested):
Async handler helper function seems to me like it should have been a decorator. So decorate it once, then if packagers are enabled, you can decorate it again with the packagers decorator. You don't need to decorate with async handler in two places, if packagers are enabled and if they are not...

Because now it won't be needed in packagers, you should move it from utils to simply be declared in local.py.

mlrun/runtimes/local.py
jond01: I moved the helper functions to the `local.py` file, and now they are used only inside the same module.

mlrun/utils/helpers.py
jond01: https://github.com/mlrun/mlrun/pull/9592#discussion_r3131558454

jond01:
> Async handler helper function seems to me like it should have been a decorator. So decorate it once, then if packagers are enabled, you can decorate it again with the packagers decorator. You don't need to decorate with async handler in two places, if packagers are enabled and if they are not...
> 
> Because now it won't be needed in packagers, you should move it from utils to simply be declared in local.py.

Thanks for the suggestion. It is now implemented with a decorator on the original handler, rather than only the branches of `packagers_enabled=True/False`.

guy1992l (approved):
Nice, clean and simple

mlrun/runtimes/local.py
```diff
@@ -476,6 +481,68 @@ def flush(self):
         self.terminal.flush()
 
 
+_T = typing.TypeVar("_T")
+
+
+def _run_async_handler(coro: Coroutine[typing.Any, typing.Any, _T]) -> _T:
```
liranbg: move private functions *below* public functions. convention.

tests/runtimes/test_async_handler.py
```diff
@@ -0,0 +1,167 @@
...
+    original = mlrun.mlconf.packagers.enabled
+    yield
+    mlrun.mlconf.packagers.enabled = original
+
+
+def _launch(
```
liranbg: move below public

tests/runtimes/test_async_handler.py
```diff
@@ -0,0 +1,167 @@
...
+
+    async def call_from_running_loop() -> mlrun.run.RunObject:
+        return _launch("async_handler", packagers_enabled=packagers_enabled)
+
+    result = asyncio.run(call_from_running_loop())
+    assert result.status.state == "completed"
+    assert result.status.results.get("async_result") == 42
+
+
+def test__run_async_handler_returns_value() -> None:
```
liranbg: why the `__`?

tests/system/runtimes/test_kubejob.py
```diff
@@ -878,6 +879,31 @@ async def test_job_from_serving_runtime_from_inside_running_asyncio_loop(self):
         finally:
             v3io_client.close()
 
+    def _make_async_handler_function(
```
liranbg: move below public

liranbg (changes requested):

mlrun/runtimes/local.py
jond01: Ok, moved

tests/runtimes/test_async_handler.py
jond01: It's a testing module, I don't think this convention is relevant here, but ok

liranbg (approved):

docs/concepts/submitting-tasks-jobs-to-functions.md
```diff
@@ -9,53 +9,162 @@ Learn how to submit a job using `run_function`, and use the `RunObject` to track
...
+    handler="PlainClass::run",
+)
+run = fn.run()
+print(run.output("result"))  # 42
+```
+
+#### Passing arguments to the constructor
+
+To initialize the class with custom arguments, define them in `__init__` and
+pass them at run time via the special `"_init_args"` key inside `params`.
```
jillnogold: pass them at runtime via the special `"_init_args"` key inside `params`.

jond01: Changed it 👍

docs/concepts/submitting-tasks-jobs-to-functions.md
```diff
@@ -9,53 +9,162 @@ Learn how to submit a job using `run_function`, and use the `RunObject` to track
...
+    def __init__(self, context: mlrun.MLClientCtx, threshold: float = 0.5):
+        context.logger.debug("Initializing the class")
+        self.threshold = threshold
+
+    def run(self, context: mlrun.MLClientCtx) -> None:
+        value = 0.9  # extract the value from the context
+        context.log_result("above_threshold", value > self.threshold)
+```
+
+Pass constructor arguments at run time:
```
jillnogold: Pass constructor arguments at runtime:

jillnogold (approved):

docs/concepts/submitting-tasks-jobs-to-functions.md
```diff
@@ -9,53 +9,162 @@ Learn how to submit a job using `run_function`, and use the `RunObject` to track
...
+It is supported also in Jupyter notebooks.
 ```
 
-MLRun also supports iterative jobs that can run and track multiple child jobs (for hyperparameter tasks, AutoML, etc.). 
+MLRun also supports iterative jobs that can run and track multiple child jobs (for hyperparameter tasks, AutoML, etc.).
 See {ref}`hyper-params` for details and examples.
 
+### Handlers inside a class
+
+You can place set function handlers to methods inside a class and reference them with the
```
assaf758:
```suggestion
You can set function handlers to methods inside a class and reference them with the
```

assaf758 (commented):
Great job!
typo inline

docs/concepts/submitting-tasks-jobs-to-functions.md
jond01: Fixed

assaf758 (approved):

PR #9578, opened 2026-04-15
[Invoke] Use https for external urls unless schema is set explicitly

rokatyy:
### 📝 Description
Use https for external urls unless schema is set explicitly

---

### 🛠️ Changes Made

---

### ✅ Checklist
- [ ] I updated the documentation (if applicable)
- [x] I have tested the changes in this PR
- [x] I confirmed whether my changes are covered by system tests
  - [ ] If yes, I ran all relevant system tests and ensured they passed before submitting this PR
  - [ ] I updated existing system tests and/or added new ones if needed to cover my changes
- [ ] If I introduced a deprecation:
  - [ ] I followed the [Deprecation Guidelines](./DEPRECATION.md)
  - [ ] I updated the relevant Jira ticket for documentation
- [ ] Please run [Smoke-tests](https://github.com/mlrun/mlrun/actions/workflows/system-tests-opensource.yml) workflow, providing it the PR number as input (upon success, the workflow run will add label "Smoke tests: Pass" to the RP)

---

### 🧪 Testing
unit, run application runtime system tests

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-12409
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

PR #9593, opened 2026-04-21
[Tests] Fix system tests broken by https external URL change

Yacouby:
### 📝 Description
This PR: https://github.com/mlrun/mlrun/pull/9578 changed `_resolve_invocation_url` to use `https://` instead of `http://` for external invocation URLs. 
This broke several system tests that either asserted the full URL (including scheme) or used raw `requests.post()` calls without `verify=False`.

---

### 🛠️ Changes Made

- URL assertions made scheme-agnostic: Changed assertions to compare only hostnames using `urlparse().hostname`, so they don't break when the URL scheme changes.

- Added `verify=False` to raw `requests.post()` calls: Tests that use `requests.post()` directly now pass `verify=False` to support self-signed certificate clusters. 

---

### ✅ Checklist

---

### 🧪 Testing
Failing system tests pass successfully on my vmdev.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/ML-12455
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---


elbamit (commented):
Using `urlparse().hostname` you only check the hostname, not the entire endpoint. This means that differences in the endpoint itself will not be detected.
example
```
# will both pass the assertion
https://api.example.com/v1/run
https://api.example.com/v2/run
```
Isn't that breaking the test?

Yacouby:
> Using `urlparse().hostname` you only check the hostname, not the entire endpoint. This means that differences in the endpoint itself will not be detected. example
> 
> ```
> # will both pass the assertion
> https://api.example.com/v1/run
> https://api.example.com/v2/run
> ```
> 
> Isn't that breaking the test?

@elbamit 
No - the test's purpose is to verify that the function was deployed successfully and is reachable, which is validated by the `function.invoke()` call that follows. The URL comparison was just a sanity check that `deploy()` and` get_url()` point to the same host, not a test of the URL format itself.

liranbg (approved):

TomerShor (approved):

PR #9596, opened 2026-04-21
[Model Monitoring] Use per-function Kafka consumer group

alxtkr77:
## Summary

Re-addresses ML-11979 — cross-function rebalance on the shared `"serving"` Kafka consumer group causes MM data loss under scale (missed events → missing drift status / last prediction, observed in REG-1918). The original fix ([#9378](https://github.com/mlrun/mlrun/pull/9378)) was reverted by [#9504](https://github.com/mlrun/mlrun/pull/9504) after two regressions were reported in 1.11.0-rc43 — ML-12356 and ML-12357.

This PR lands the same per-function group idea with narrower scope and hardened migration.

## Changes Made

- Derive MM consumer group as `f"{profile.group or 'serving'}_{topic}"` inside `_apply_and_create_kafka_source` only — a custom `DatastoreProfileKafkaStream.group` is honoured as a namespace prefix instead of being discarded.
- Leave `DatastoreProfileKafkaStream.group` default untouched (`"serving"`) — avoids the ML-12356 blast radius on non-MM Kafka users.
- Preserve pre-PR re-raise behaviour when `ignore_stream_already_exists_failure=False` (controller with `overwrite=False`) — avoids ML-12357.
- Add `_migrate_kafka_consumer_group_offsets` to copy committed offsets from the legacy shared group to the new per-function group on upgrade:
  - Source group is `kafka_profile.group` (defaulting to `"serving"`), not hardcoded.
  - Self-detecting idempotency via `list_consumer_group_offsets(new_group)` — no operator flag.
  - Raises on Kafka failure (no silent fallback to `initial_offset=earliest` and topic replay).
  - Marked `# TODO: Remove in 1.14.0` per DEPRECATION.md (two-minor-version window).
- Update `_enrich_kafka_topic_stats` to create one `KafkaConsumer` per topic with the matching per-function `group_id`, closed in `finally` (also fixes an existing leak).
- Introduce module-level `_LEGACY_KAFKA_GROUP = "serving"` and `_kafka_base_group()` helper — single source of truth for the legacy default.

## References

- Jira: [ML-11979](https://iguazio.atlassian.net/browse/ML-11979)
- Original (reverted) PR: [#9378](https://github.com/mlrun/mlrun/pull/9378)
- Revert PR: [#9504](https://github.com/mlrun/mlrun/pull/9504)
- Regressions filed against #9378 that drove the revert: [ML-12356](https://iguazio.atlassian.net/browse/ML-12356), [ML-12357](https://iguazio.atlassian.net/browse/ML-12357)

## Breaking Changes

**No.** `DatastoreProfileKafkaStream` behavior is unchanged for non-MM callers (feature store Kafka ingestion, user serving graphs). The per-function group rename is confined to model-monitoring code paths; the one-time offset migration preserves consumption position for upgraders.

[ML-11979]: https://iguazio.atlassian.net/browse/ML-11979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/py/services/api/crud/model_monitoring/deployment.py
```diff
@@ -78,6 +78,33 @@
 )
 BASE_PERIOD_LOOKUP_TABLE = {1: 1, 2: 2, 20: 3, 60: 5, 120: 10, float("inf"): 20}
 
+# Fallback for ``DatastoreProfileKafkaStream.group`` when the profile has
+# no group set (``None``). Matches the pydantic default on that field so
+# an explicit-None profile behaves the same as the default one, and so
+# the legacy migration source group (pre-ML-11979) is well-defined.
```
gtopper: I guess there's no need to mention ML-11979 11 times :)

alxtkr77: Fair :) Dropped all 10 mentions in a162c7997 — per [AGENTS.md](https://github.com/mlrun/mlrun/blob/development/AGENTS.md#code-style--conventions) the ticket belongs in the PR description, not scattered through the code.

assaf758 (approved):
Nice!

PR #9605, opened 2026-04-25
[Workflow] Fix GitHub App token expiry during cross-repo UI build

jmcdonagh88:
## Summary

This PR migrates GitHub Actions workflows from Personal Access Tokens (PATs) to GitHub App tokens and adds a shared token refresh mechanism for long-running jobs.

### Token expiry fix for cross-repo UI build (`private-release.yaml`)
- Replaced `convictional/trigger-workflow-and-wait` with an inline bash script that generates JWTs, creates installation tokens, and automatically refreshes them every 55 minutes during the polling loop
- Prevents `401 Bad credentials` failures when the UI build is queued longer than 1 hour

### Shared token refresh script (`.github/scripts/github-app-token.sh`)
- New shared bash script centralising GitHub App JWT generation, installation token creation, and a background refresh daemon
- The daemon periodically refreshes the token and updates `env.yml` on disk so long-running system tests always have a valid token

### PAT to GitHub App migration
- **`upgrade-lock.yaml`**: Replaced `MLRUN_PR_PAT` with GitHub App token for `peter-evans/create-pull-request`
- **`system-tests-enterprise.yml`**: Replaced `SYSTEM_TEST_GITHUB_ACCESS_TOKEN` PAT with GitHub App token scoped to `mlrun,private-system-tests`; added background token refresh daemon
- **`system-tests-opensource.yml`**: Same migration as enterprise; added background token refresh daemon

### Docker cleanup improvements (`build-internal.yaml`)
- Cleanup step now runs on every build (`if: always()`), not just private releases
- Prunes all unused images (`docker image prune -af`) and build cache older than 30 days
- Uses `|| true` to prevent cleanup failures from failing the job

### Stale venv fix (`release-promotion.yaml`)
- Added `--clear` flag to `python3.11 -m venv` to prevent stale venv directories from previous runs causing `ensurepip` failures due to dangling symlinks

## Test plan
- [x] Token generation and refresh validated via dry-run workflow
- [x] Full private release workflow completed successfully on this branch (all image builds + UI build + PyPI publish)
- [x] Release promotion workflow validated (PyPI venv fix confirmed working)
- [x] Verify `upgrade-lock.yaml` GitHub App has `Pull requests: Write` permission (currently failing with `Resource not accessible by integration`)
- [ ] System test workflows to be validated on next scheduled/manual run


.github/workflows/private-release.yaml
```diff
@@ -130,23 +130,99 @@ jobs:
...
+        env:
+          GH_APP_ID: ${{ secrets.GH_APP_ID }}
+          GH_APP_PRIVATE_KEY: ${{ secrets.GH_APP_PRIVATE_KEY }}
+          OWNER: ${{ github.repository_owner }}
+          UI_REF: ${{ needs.prepare-inputs.outputs.ui_ref }}
+          VERSION: ${{ needs.prepare-inputs.outputs.version }}
+          POLL_INTERVAL: 60
+          TIMEOUT_MINUTES: 120
+          TOKEN_LIFETIME: 3300
+        run: |
```
liranbg: why not using `convictional/trigger-workflow-and-wait@f69fa9eedd3c62a599220f4d5745230e237904be`?

jmcdonagh88: @liranbg ,

The convictional/trigger-workflow-and-wait action receives the GitHub App token once at step start and has no mechanism to refresh it. GitHub App installation tokens expire after 1 hour. When the UI build gets queued behind other jobs on the mlrun-release runners (we've seen 50+ minute queue times), the token expires mid-poll, and the action fails with 401 Bad credentials.

The inline script replaces it by refreshing the token every 55 minutes during the polling loop, generating a fresh JWT and requesting a new installation token. This isn't possible with convictional/trigger-workflow-and-wait since it doesn't expose a hook for token refresh, as far as I am aware

liranbg (approved):

PR #9607, opened 2026-04-26
[API] Fix DB backup auth on password-protected MySQL

liranbg:
### 📝 Description

The DB backup step that runs before migrations was authenticating only with `-h/-P/-u` from the parsed DSN — the password parsed out of the DSN was dropped on the floor. On deployments where the MySQL DSN carries a password (e.g. `mysql+pymysql://user:secret@host:3306/db`), `mysqldump` (and the restore-time `mysql` client) failed to authenticate, breaking the pre-migration backup.

The password now flows through to the subprocess via the `MYSQL_PWD` environment variable, so it never appears in the shell command string that gets debug-logged, and never shows up in `ps`/`/proc/<pid>/cmdline`.

---

### 🛠️ Changes Made
- `server/py/services/api/utils/db/backup.py`:
  - Added `DBBackupUtil._mysql_password_env(dsn_data)` — returns `{"MYSQL_PWD": password}` when the DSN has a password, else `None`.
  - `_backup_database_mysql` and `_load_database_backup_mysql` pass that env to `_run_shell_command`.
  - `_run_shell_command` now accepts an optional `env` dict, merges it into the inherited `os.environ`, and forwards it to `subprocess.Popen`. The env contents are intentionally not logged.

---

### ✅ Checklist

---

### 🧪 Testing
- Unit tests in `server/py/services/api/tests/unit/utils/db/test_db_backup.py`:
  - Updated `test_backup_and_load_mysql` to assert `_run_shell_command` is invoked with `env=None` when the DSN has no password (no behavior change for that path).
  - Added `test_backup_and_load_mysql_with_password` covering a DSN of the form `mysql+pymysql://root:s3cret@mlrun-db:3306/mlrun`. Asserts `MYSQL_PWD` is passed via `env` for both backup and restore commands, and that the password never appears in the command string itself.
- All 6 tests in the file pass; `ruff check` and `ruff format --check` pass on the touched files.

---

### 🔗 References
- Ticket link:
- Design docs links:
- External links:
  - [MySQL `MYSQL_PWD` environment variable](https://dev.mysql.com/doc/refman/8.0/en/environment-variables.html)

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes
- `MYSQL_PWD` is preferred over `-p<password>` here because `_run_shell_command` debug-logs the full command string; embedding the password would have leaked it into logs (and into `ps` output). The env-var approach keeps the password out of both.
- `MySQL`'s docs note `MYSQL_PWD` is visible via `/proc/<pid>/environ` to other users on the same host. If that becomes a concern in shared-host deployments, a follow-up can switch to a `--defaults-extra-file` tempfile (mode `0600`).
- No change to the SQLite path or to behavior when the DSN has no password.

### DIFF
--- a/server/py/services/api/utils/db/backup.py
+++ b/server/py/services/api/utils/db/backup.py
@@ -12,11 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import contextlib
 import datetime
 import os
 import pathlib
 import shutil
 import subprocess
+import tempfile
 
 import mlrun
 import mlrun.common.db.dialects
@@ -108,14 +110,15 @@ class DBBackupUtil:
 
         mlrun.utils.logger.debug("Backing up mysql DB data", backup_path=backup_path)
         dsn_data = DBUtil.get_parsed_dsn().as_dict()
-        self._run_shell_command(
-            "mysqldump --single-transaction --routines --triggers "
-            f"--max_allowed_packet={mlrun.mlconf.httpdb.db.backup.max_allowed_packet} "
-            f"-h {dsn_data['host']} "
-            f"-P {dsn_data['port']} "
-            f"-u {dsn_data['username']} "
-            f"{dsn_data['database']} > {backup_path}"
-        )
+        with self._mysql_defaults_file(dsn_data) as defaults_arg:
+            self._run_shell_command(
+                f"mysqldump {defaults_arg}--single-transaction --routines --triggers "
+                f"--max_allowed_packet={mlrun.mlconf.httpdb.db.backup.max_allowed_packet} "
+                f"-h {dsn_data['host']} "
+                f"-P {dsn_data['port']} "
+                f"-u {dsn_data['username']} "
+                f"{dsn_data['database']} > {backup_path}"
+            )
 
     def _load_database_backup_mysql(self, backup_file_name: str) -> None:
         """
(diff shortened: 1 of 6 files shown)


server/py/services/api/utils/db/backup.py
```diff
@@ -200,17 +212,22 @@ def _get_sqlite_db_file_path() -> str:
...
-    def _run_shell_command(command: str) -> int:
+    def _run_shell_command(command: str, env: dict[str, str] | None = None) -> int:
         mlrun.utils.logger.debug(
             "Running shell command",
             command=command,
         )
+        # Merge `env` into the inherited environment so the subprocess keeps
+        # PATH etc. The values in `env` are intentionally not logged — callers
+        # use it to pass credentials (e.g. MYSQL_PWD).
+        subprocess_env = {**os.environ, **env} if env else None
```
Yacouby:
```suggestion
        subprocess_env = {**os.environ, **env} if env is not None else None
```

server/py/services/api/utils/db/backup.py
```diff
@@ -135,9 +136,20 @@ def _load_database_backup_mysql(self, backup_file_name: str) -> None:
...
 
+    @staticmethod
+    def _mysql_password_env(dsn_data: dict) -> dict[str, str] | None:
+        # Pass the password via MYSQL_PWD so it's not embedded in the shell
+        # command (which is logged) or visible in `ps`. The mysql/mysqldump
+        # clients read this env var when no -p flag is given.
+        password = dsn_data.get("password")
+        if not password:
+            return None
+        return {"MYSQL_PWD": password}
```
Yacouby: `MYSQL_PWD` is deprecated as of MySQl 8.4 : 
https://dev.mysql.com/doc/refman/8.4/en/environment-variables.html#:~:text=MYSQL_PWD%20is%20deprecated%20as%20of%20MySQL%208.4%3B%20expect%20it%20to%20be%20removed%20in%20a%20future%20version%20of%20MySQL.

liranbg: Fixed https://github.com/mlrun/mlrun/pull/9607/commits/aa93f4c4f4dcdd3bef506e1f2786fb9a238ba380

server/py/services/api/utils/db/backup.py
```diff
@@ -130,13 +133,45 @@ def _load_database_backup_mysql(self, backup_file_name: str) -> None:
...
+        # or in `/proc/<pid>/environ`. Unlike MYSQL_PWD, this is not deprecated
+        # in MySQL 8.4+. The file is unlinked on context exit.
+        password = dsn_data.get("password")
+        if not password:
+            yield ""
+            return
+
+        # Escape backslash and double-quote so values with these characters
+        # round-trip through the option file format.
+        escaped = password.replace("\\", "\\\\").replace('"', '\\"')
```
Yacouby, line 160: How about adding a test case with a password with backslash to verify the option file content is escaped correctly?

Yacouby (commented):
one minor suggestion

Yacouby (approved):

PR #9608, opened 2026-04-26
[Automation] Add kubectl mode and DB migration trigger to patch_remote

liranbg:
### 📝 Description

Adds two opt-in capabilities to `automation/patch_igz/patch_remote.py` to make patching faster and friendlier when working from a workstation that already has cluster access via `kubectl`, without changing the existing SSH-based flow.

The existing `ssh` mode remains the default and is unchanged.

Usage example

```sh
./automation/patch_igz/patch_remote.py \
    -c automation/patch_igz/patch_env.yml \
    --kubeconfig ~/.kube/mydev.yaml
```

or, with migration

```sh
./automation/patch_igz/patch_remote.py \
    -c automation/patch_igz/patch_env.yml \
    --kubeconfig ~/.kube/mydev.yaml \
    --migrate
```

---

### 🛠️ Changes Made

- **`MODE=kubectl` / `--mode kubectl`**: run `kubectl` locally against the target cluster instead of SSH-ing into a data node. `KUBECONFIG` (config field) or `--kubeconfig` points at the right cluster.
  - `SSH_USER` / `SSH_PASSWORD` are no longer required in this mode.
  - `DATA_NODES` is optional unless pushing to per-node Iguazio datanode registries (multi-node image replication).
- **Mode inference**: when `MODE` is unset, kubectl mode is inferred if `KUBECONFIG` is provided (config field or `--kubeconfig` flag); otherwise the legacy `ssh` default is used. Explicit `MODE` always wins, so kubectl mode against the kubectl default (`KUBECONFIG` env / `~/.kube/config`) still works via `MODE: kubectl` with no `KUBECONFIG` field.
- **`--migrate` / `-mig`**: after the api rollout, trigger DB migrations on `mlrun-api-chief` via the official `/operations/migrations` endpoint.
  - Authenticates with the pod's k8s service-account token + `x-igz-authenticator-kind: sa` header (Iguazio v4 verifies it as an SA session).
  - Disables the chief deployment's DB backup mode first (idempotent) to work around the api container's missing `mysqldump` credentials when a migration is actually pending.
  - Refuses to run together with `--skip-api`.
- **Config/CLI/docs updates**: `patch_env_template.yml` now documents `MODE` and `KUBECONFIG` and clarifies which fields are required per mode; `README.md` describes the two execution modes and inference rules.

---

### ✅ Checklist

---

### 🧪 Testing

- Manually patched a remote Iguazio lab system

---

### 🔗 References
- Ticket link:
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

Existing SSH flow is unchanged. New behavior is opt-in via `MODE` / `KUBECONFIG` / `--migrate`.

---

### 🔍️ Additional Notes

- The migration trigger intentionally bypasses `mlrun.get_run_db()`: its `connect()` syncs client-spec which enables `auth_with_oauth_token`, and `IGTokenProvider.__init__` eagerly fetches a token that isn't available inside the api pod. The snippet constructs `HTTPRunDB` directly and injects the SA bearer onto the session so both the trigger `POST` and the polling `GET`s carry it.

### DIFF
--- a/automation/patch_igz/patch_remote.py
+++ b/automation/patch_igz/patch_remote.py
@@ -38,7 +38,11 @@ coloredlogs.install(level=log_level, logger=logger, fmt=fmt)
 
 
 class Constants:
-    mandatory_fields = {"DATA_NODES", "SSH_USER", "SSH_PASSWORD", "DOCKER_REGISTRY"}
+    mode_ssh = "ssh"
+    mode_kubectl = "kubectl"
+    valid_modes = {mode_ssh, mode_kubectl}
+    common_mandatory_fields = {"DOCKER_REGISTRY"}
+    ssh_mandatory_fields = {"DATA_NODES", "SSH_USER", "SSH_PASSWORD"}
     api_container = "mlrun-api"
     log_collector_container = "mlrun-log-collector"
     api = "api"
@@ -69,6 +73,9 @@ class MLRunPatcher:
         no_build: bool,
         no_push: bool,
         namespace: str,
+        mode: str = "",
+        kubeconfig: str = "",
+        migrate: bool = False,
     ):
         self._config = yaml.safe_load(conf_file)
         patch_yaml_data = yaml.safe_load(patch_file)
@@ -76,6 +83,11 @@ class MLRunPatcher:
         self._reset_db = reset_db
         self._image_tag = image_tag
         self._patch_log_collector_image = bool(patch_log_collector_image)
+        self._mode = self._resolve_mode(mode)
+        kubeconfig_value = kubeconfig or self._config.get("KUBECONFIG") or ""
+        self._kubeconfig = (
+            os.path.expanduser(kubeconfig_value) if kubeconfig_value else ""
+        )
         self._validate_config()
         self._patch_mlrun_image = patch_mlrun_image
         self._skip_patch_api = skip_patch_api
@@ -84,10 +96,16 @@ class MLRunPatcher:
         self._no_push = no_push
         self._namespace = self._resolve_namespace(namespace)
         self._docker_client = docker.from_env()
+        self._ssh_client: paramiko.SSHClient | None = None
+        self._migrate = migrate
         if self._skip_patch_api and self._patch_alerts:
             raise ValueError("Cannot skip api and patch alerts at the same time")
+        if self._migrate and self._skip_patch_api:
+            raise ValueError(
+                "--migrate requires the api to be patched (drop --skip-api)"
+            )
 
-        cluster_data_nodes = self._config["DATA_NODES"]
+        cluster_data_nodes = self._config.get("DATA_NODES") or []
         if not isinstance(cluster_data_nodes, list):
             cluster_data_nodes = [cluster_data_nodes]
         self._cluster_data_nodes = cluster_data_nodes
@@ -122,8 +140,7 @@ class MLRunPatcher:
             )
             self._push_docker_images(built_images)
 
-        # Connect to the first node and start deployment patching process
-        node = self._cluster_data_nodes[0]
+        node = self._cluster_data_nodes[0] if self._cluster_data_nodes else None
         self._connect_to_node(node)
 
         if self._patch_log_collector_image:
@@ -146,6 +163,9 @@ class MLRunPatcher:
                     self._rollout_deployment()
 
                 self._wait_deployment_ready()
+
+                if self._migrate:
+                    self._run_db_migrations()
 
             finally:
                 # Check status of pods after deployment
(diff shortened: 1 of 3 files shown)


automation/patch_igz/patch_remote.py
```diff
@@ -449,6 +491,109 @@ def _wait_for_pods_readiness(self):
...
+        # eagerly fetches a token we don't have. Construct HTTPRunDB directly
+        # and inject the pod's k8s service-account token on the session so
+        # every api_call (POST + polling GETs) carries the SA bearer.
+        snippet = "\n".join(
+            [
+                "import sys, mlrun.db.httpdb as httpdb",
+                "db = httpdb.HTTPRunDB('http://localhost:8080')",
+                "db.token_provider = None",
+                "db.session = db._init_session(False, False)",
+                "with open('/var/run/secrets/kubernetes.io/serviceaccount/token') as f:",
```
Yacouby, line 567: what if the file doesn't exist ? (e.g., SA token mounting disabled on the pod)
can use try/except here

automation/patch_igz/patch_remote.py
```diff
@@ -696,6 +855,25 @@ def _resolve_docker_registry(self):
...
+    help="Force execution mode. Defaults to ssh, or kubectl when --kubeconfig/KUBECONFIG is set.",
+)
+@click.option(
+    "--kubeconfig",
+    default="",
+    type=str,
+    help="Path to a kubeconfig file. Setting this implies kubectl mode unless --mode is given.",
+)
+@click.option(
+    "-mig",
```
Yacouby, line 878: isnt there some convention for this to be a single-char ? if so we can drop it and keep only --migrate
since it is also conufsed with the above -m option

automation/patch_igz/patch_remote.py
```diff
@@ -574,6 +719,9 @@ def _exec_local(
         return output
 
     def _exec_remote(self, cmd: list[str], live=False) -> str:
+        if self._mode == Constants.mode_kubectl:
+            return self._exec_kubectl_local(cmd, live=live)
```
Yacouby, line 728: When this command fails in non-live mode, the failure output is swallowed ?

automation/patch_igz/patch_remote.py
```diff
@@ -246,7 +266,10 @@ def _push(image: str):
                 fut.result()
 
     def _validate_config(self):
-        missing_fields = Constants.mandatory_fields - set(self._config.keys())
+        mandatory_fields = set(Constants.common_mandatory_fields)
+        if self._mode == Constants.mode_ssh:
+            mandatory_fields |= Constants.ssh_mandatory_fields
+        missing_fields = mandatory_fields - set(self._config.keys())
```
Yacouby: validation should also assert the value is non-empty.  If someone sets DATA_NODES: null (or leaves it empty) in their YAML the validation passes but then `self._cluster_data_nodes[0]` silently falls back to `None`, and `_connect_to_node(None)` crashes with a confusing error

Yacouby (changes requested):
minor

automation/patch_igz/patch_remote.py
liranbg: there's no such deployment for igz where SA is disabled

automation/patch_igz/patch_remote.py
liranbg: not really

automation/patch_igz/patch_remote.py
liranbg: fixed

automation/patch_igz/patch_remote.py
liranbg: fixed

Yacouby (approved):

automation/patch_igz/patch_remote.py
```diff
@@ -122,8 +140,7 @@ def patch(self):
             )
             self._push_docker_images(built_images)
 
-        # Connect to the first node and start deployment patching process
-        node = self._cluster_data_nodes[0]
+        node = self._cluster_data_nodes[0] if self._cluster_data_nodes else None
```
elbamit, line 143: If someone uses ssh mode and passes `DATA_NODES=[""]`, we'll get stuck in `_connect_to_node(node=None)` in ssh mode.

elbamit (approved):
Minor comment, looks good

automation/patch_igz/patch_remote.py
liranbg: existing bug, right? well, I hope that wont happen for the user :)

PR #9609, opened 2026-04-27
[Code Artifacts] Run jobs from store:// code artifacts

alxtkr77:
## 📝 Description

This is the first PR in a multi-PR split delivering the CodeArtifact runtime feature (parent: [ML-11980](https://iguazio.atlassian.net/browse/ML-11980)). It introduces support for running **job-kind** functions sourced from a `store://` code artifact URI, and adds the server-side helper that merges artifact requirements into the function build spec at deploy/run time.

Subsequent PRs will add Nuclio dual-mode deploy, project zip-export of store:// code, and end-to-end system tests.

## 🛠️ Changes Made

**SDK side (mlrun/):**
- `set_function(func="store://...")` — store:// URIs are accepted as function sources and stored as-is in `spec.build.source`, resolved at run/deploy time.
- `extract_source()` — delegates to `load_source_code()` for store:// URIs (consistent with git/archive sources).
- `_load_store_artifact()` — now returns the **target directory** (not the file path) so callers like `_pre_run` can use it as the workdir + add it to `sys.path`.
- `LocalRuntime` passes `project=` to `extract_source()` so the local runner can resolve store:// artifacts.
- Refactor: `is_store_uri` moved to `mlrun.utils.helpers` (next to `DB_SCHEMA`); `mlrun.datastore` keeps re-export for backward compatibility.

**Server side (server/py/):**
- New helper `services.api.utils.functions.enrich_function_from_code_artifact(function, project)` — resolves a `store://` URI and merges the artifact's `spec.requirements` into `function.spec.build.requirements` using `mlrun.utils.merge_requirements()` (user requirements win).
- `ServerSideLauncher.enrich_runtime()` calls the helper before `requires_build()`, so job/KFP/spark/MPI runtimes get artifact requirements merged into their build spec, triggering an image build when needed.
- `KubeRuntimeHandler._get_cmd_args()` defaults `load_source_on_run=True` for store:// jobs **only when unset** (`is None`); explicit user values (`True`/`False`) are honored. Logs a `logger.debug` for visibility.

## ✅ Checklist

- [x] `make fmt` clean
- [x] `make lint` clean (on changed files)
- [x] Unit tests added/updated and passing (39 cases)
- [x] No FastAPI endpoint changes (no new auth/authz needed)
- [x] No DB migrations
- [x] No changes to `version.json`, `locked-requirements.txt`, or `.github/workflows/`

## 🧪 Testing

Ran 39 unit-test cases across the touched surface — all pass:

```
server/py/services/api/tests/unit/utils/test_functions.py        (5 tests for enrich_function_from_code_artifact)
server/py/services/api/tests/unit/runtime_handlers/test_kubejob.py (2 tests for store:// load_source_on_run defaulting + explicit-False preservation)
tests/utils/test_clones.py                                        (3 tests for extract_source store:// delegation)
tests/projects/test_project.py                                    (4 tests for set_function store:// across job/nuclio/serving/with_repo)
```

System tests for the end-to-end flow will land in a later PR in this series.

## 🔗 References

- Jira (this PR): [ML-12479](https://iguazio.atlassian.net/browse/ML-12479)
- Parent feature: [ML-11980](https://iguazio.atlassian.net/browse/ML-11980)
- Related: ML-12337 (CodeArtifact type, merged in #9568); follow-ups ML-12480 (Nuclio dual-mode), ML-12481 (zip export), ML-12482 (system tests)

## 🚨 Breaking Changes

**No** breaking changes for users.

Internal note for reviewers: `mlrun/utils/clones._load_store_artifact()` (private helper) now returns a directory path instead of a file path. Verified the only caller (`load_source_code`) passes the result through unchanged, and the docstring was updated.

[ML-11980]: https://iguazio.atlassian.net/browse/ML-11980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ML-12479]: https://iguazio.atlassian.net/browse/ML-12479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/py/services/api/tests/unit/utils/test_functions.py
```diff
@@ -0,0 +1,94 @@
...
+
+    with patch(
+        "mlrun.datastore.get_store_resource",
+        side_effect=Exception("artifact not found"),
+    ):
+        try:
+            enrich_function_from_code_artifact(func, "proj")
+            assert False, "Should have raised"
+        except mlrun.errors.MLRunInvalidArgumentError as exc:
+            assert "Cannot resolve code artifact" in str(exc)
```
Yacouby: replace this code with `pytest.raises`. would be cleaner

server/py/services/api/runtime_handlers/kubejob.py
```diff
@@ -122,6 +123,20 @@ def _get_cmd_args(
...
+        if (
+            runtime.spec.build.source
+            and mlrun.utils.is_store_uri(runtime.spec.build.source)
+            and runtime.spec.build.load_source_on_run is None
+        ):
+            logger.debug(
+                "Defaulting load_source_on_run=True for store:// job source",
+                source=runtime.spec.build.source,
+            )
+            runtime.spec.build.load_source_on_run = True
```
Yacouby: This method typically shouldn't mutate the object it reads. If the method gets called more than once. Shouldn't this defaulting happen earlier in the run flow (e.g. in `enrich_runtime`) where mutations are expected?

server/py/services/api/utils/functions.py
```diff
@@ -133,3 +135,35 @@ def build_function(
...
+    """Resolve store:// code artifact and merge its requirements into function build spec.
+
+    :param function: The function object to enrich
+    :param project:  Project name for artifact resolution
+    """
+    # Falls back to status.application_source for Nuclio (Application runtime)
+    # redeploys where spec.build.source is cleared mid-deploy. No-op for jobs.
+    source = function.spec.build.source or getattr(
+        function.status, "application_source", None
+    )
```
Yacouby, line 163: The comment says this fallback is for Nuclio redeploys, but `enrich_function_from_code_artifact` is only called from `ServerSideLauncher.enrich_runtime`, which doesn't handle Nuclio deploys

server/py/services/api/utils/functions.py
```diff
@@ -133,3 +135,35 @@ def build_function(
...
+    if not source or not mlrun.utils.is_store_uri(source):
+        return
+
+    try:
+        artifact = mlrun.datastore.get_store_resource(source, project=project)
```
Yacouby, line 168: `get_store_resource` can return any artifact kind. 
If artifact.kind != "code", then `getattr(artifact.spec, "requirements", None)` silently returns None — no requirements merged and no feedback to the user. Should there be a kind check here, or at least a warning when the source resolves to a non-code artifact?

mlrun/utils/clones.py
```diff
@@ -191,9 +191,16 @@ def clone_git(url: str, context: str, secrets=None, clone: bool = True):
...
+def extract_source(source: str, workdir=None, secrets=None, clone=True, project=None):
     if not source:
         return
+    if is_store_uri(source):
+        target_dir = workdir or os.path.realpath("./code")
+        return load_source_code(
+            source_uri=source,
+            target_dir=target_dir,
+            project=project,
+        )
```
Yacouby, line 207: The clone parameter is silently ignored for `store://` URIs, worth a short comment clarifying this is intentional

Yacouby, line 207: Another comment here:
The secrets parameter from `extract_source` is not forwarded. `_load_store_artifact` may need credentials if the artifact is on a secured store and secrets are provided via the secrets parameter so in this case they'd be silently dropped.

Yacouby (changes requested):

server/py/services/api/utils/functions.py
alxtkr77: Intentional — this is scaffolding for the follow-up Nuclio dual-mode PR (ML-12480), which will add the actual call site from a Nuclio deploy path

Yacouby (approved):

PR #9618, opened 2026-04-27
[API] Publish DB migration lifecycle events to Iguazio v4

liranbg:
### 📝 Description

Add an Iguazio v4 events client that publishes
`Platform.MLRun.DB.Migration{Required,Started,Completed,Failed}` via the
orca SDK's `publish_event` endpoint

Bonus fix: the DB backup step is now wrapped inside the same try/except as schema and data migrations, so a backup failure correctly transitions the API to `migrations_failed` and emits the `Failed` event (previously the exception bubbled out without state transition).

---

### 🛠️ Changes Made
- Added `services/api/utils/events/iguazio_v4.py` — events client built on the iguazio (orca) Python SDK with auth via the in-cluster service-account token.
- Extended `BaseEventClient` with a default-no-op `generate_db_migration_event(action, error, duration_seconds, scope, versions)`.
- Updated `EventsFactory` to select the v4 client when `is_iguazio_v4_mode()`, with the existing v3 path preserved for non-v4 environments.
- Wired `_migrate_existing_data` in `initial_data.py` to publish Required/Started/Completed/Failed at the right transitions, capturing duration via `time.monotonic()` and version context via `_resolve_needed_operations`.
- Exposed `latest_revision` and `get_current_revision()` on `AlembicUtil` so the publish helper can include alembic source/target revisions in event details.
- Added `MigrationEventActions` enum (`required`/`started`/`completed`/`failed`) and `EventClientKinds.iguazio_v4 = "iguazio-v4"` to `mlrun.common.schemas.events`.

---

### ✅ Checklist

---

### 🧪 Testing

Unit tests
Lab verification
- Real migration cycle (introduced a temporary alembic head) emitted `Required → Started → Completed` with correct schema revisions, scope, and duration in iguazio's events store.
- Verified events are persisted via `iguazio.Client.list_event_activations` from the SDK.

---

### 🔗 References
- Ticket link: https://iguazio.atlassian.net/browse/IG4-2137
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/server/py/services/api/initial_data.py
+++ b/server/py/services/api/initial_data.py
@@ -152,19 +152,21 @@ def _migrate_existing_data(
     db_util = DBUtil()
     db_util.wait_for_db_liveness()
     db_util.set_configurations()
-    migration_scope: list[str] = []
-    migration_versions: dict = {}
+    migration_operations: dict[str, tuple] = {}
     if engine.name != mlrun.common.db.dialects.Dialects.SQLITE:
         (
             is_migration_needed,
             is_backup_needed,
-            migration_scope,
-            migration_versions,
+            migration_operations,
         ) = _resolve_needed_operations(alembic_util)
     else:
         # ON SQLite, we don't have schema migrations, so we don't need to check for them
         is_migration_needed = False
         is_backup_needed = False
+
+    migration_scope, migration_versions = _operations_to_scope_and_versions(
+        migration_operations
+    )
 
     if not perform_migrations_if_needed and is_migration_needed:
         state = mlrun.common.schemas.APIStates.waiting_for_migrations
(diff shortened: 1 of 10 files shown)


server/py/services/api/utils/events/base.py
```diff
@@ -54,3 +54,26 @@ def generate_project_secret_event(
...
         pass
+
+    def generate_db_migration_event(
+        self,
+        action: mlrun.common.schemas.MigrationEventActions,
+        error: BaseException | str | None = None,
+        duration_seconds: float | None = None,
+        scope: list[str] | None = None,
+        versions: dict | None = None,
+    ):
```
daanvdk-mck: Base method has no return type annotation while implementation does, should be added.

server/py/services/api/initial_data.py
```diff
@@ -164,33 +170,64 @@ def _migrate_existing_data(
...
-    if is_backup_needed:
-        mlrun.utils.logger.info("DB Backup is needed, backing up...")
-        db_backup = services.api.utils.db.backup.DBBackupUtil()
-        db_backup.backup_database()
-
     mlrun.utils.logger.info("Creating initial data")
     mlrun.mlconf.httpdb.state = mlrun.common.schemas.APIStates.migrations_in_progress
+    migration_start_monotonic: float | None = None
+    if is_migration_needed:
+        migration_start_monotonic = time.monotonic()
```
daanvdk-mck: Does it make sense that now the publish time is included in this? I think it would read cleaner if this was set in the below is_migration_needed block (no need to reason about when it is set and if it thus usable if its in the same block opposed to two blocks that have the same condition)

elbamit (approved):
LGTM

server/py/services/api/initial_data.py
```diff
@@ -226,7 +263,7 @@ def update_default_configuration_data():
 
 def _resolve_needed_operations(
     alembic_util: services.api.utils.db.alembic.AlembicUtil,
-) -> tuple[bool, bool]:
+) -> tuple[bool, bool, list[str], dict]:
```
daanvdk-mck: I am not too sure about the decoupling of the scopes and their related versions in this return type, I feel like this goes against the 'make invalid states unrepresentable' principle. Technically this function can now return a schema scope but have no schema revisions set and the same for data. Maybe instead we could combine the two in one dict: `{"schema": (current_revision, expected_revision), "data": (current_version, expected_version)}` so that if one key is present the revisions have to be present as well. This is a small easily inspectable internal helper though so not too important.

liranbg: Makes completely sense.

daanvdk-mck (approved):

PR #9631, opened 2026-04-29
[Workflow] Update K3S registries configuration to support Artifactory authentication

tal-haim:
📝 Description
This PR updates the System Tests Open Source workflow to route ghcr.io / gcr.io image pulls through Artifactory instead of direct GHCR auth.
It is intended to avoid long silent image-pull stalls during CE install (notably kfp-visualization-server) and to avoid exposing a hardcoded Artifactory URL in public workflow code.

🛠️ Changes Made
- Added Artifactory-based mirror configuration for: ghcr.io, gcr.io
- Switched to secret-based registry host: ARTIFACTORY_DOCKER_REGISTRY_URL
- Added fail-fast validation if ARTIFACTORY_DOCKER_REGISTRY_URL is missing.

### DIFF
--- a/.github/workflows/system-tests-opensource.yml
+++ b/.github/workflows/system-tests-opensource.yml
@@ -360,17 +360,22 @@ jobs:
       env:
         DOCKER_USERNAME: ${{ secrets.DOCKER_HUB_DOCKER_REGISTRY_USERNAME }}
         DOCKER_PASSWORD: ${{ secrets.DOCKER_HUB_DOCKER_REGISTRY_PASSWORD }}
-        GHCR_USERNAME: ${{ github.actor }}
-        GHCR_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
+        ARTIFACTORY_REGISTRY_URL: ${{ secrets.ARTIFACTORY_DOCKER_REGISTRY_URL }}
+        ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_DOCKER_REGISTRY_USERNAME }}
+        ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_DOCKER_REGISTRY_PASSWORD }}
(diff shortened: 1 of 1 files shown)


.github/workflows/system-tests-opensource.yml
```diff
@@ -360,17 +360,22 @@ jobs:
       env:
         DOCKER_USERNAME: ${{ secrets.DOCKER_HUB_DOCKER_REGISTRY_USERNAME }}
         DOCKER_PASSWORD: ${{ secrets.DOCKER_HUB_DOCKER_REGISTRY_PASSWORD }}
-        GHCR_USERNAME: ${{ github.actor }}
-        GHCR_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
+        ARTIFACTORY_REGISTRY_URL: ${{ secrets.ARTIFACTORY_DOCKER_REGISTRY_URL }}
```
assaf758, line 363: Do we have ARTIFACTORY_DOCKER_REGISTRY_URL secret configured in mlrun org? I don't see it in mlrun repo..
Where does it point to in artifactory?

assaf758 (approved):
Looks good!
Some clarifying questions inline.
Also, for first-time image use (where the artifactory cache is cold) I suspect we would be in the same problem and the smoke-test will fail on timeout? Or do you think rate-limiting will not kick in after this change?

.github/workflows/system-tests-opensource.yml
assaf758: Merged it (as anyway smoke is broken) and ran [smoke](https://github.com/mlrun/mlrun/actions/runs/25118322043/job/73613064072) - seems ARTIFACTORY_DOCKER_REGISTRY_URL is indeed missing

tal-haim: @assaf758 added

PR #9635, opened 2026-04-30
[RemoteModels] Add HF_ENDPOINT and HF_MAX_WORKERS to HuggingFace provider 

tomerm-iguazio:
### 📝 Description
Added `HF_ENDPOINT` and `HF_MAX_WORKERS` to the HuggingFace provider, as secret and profile fields. These environment variables are required to handle network restrictions when downloading models. 

In addition, In the HuggingFace profile, false values like device=0 and trust_remote_code=False were previously dropped silently but will now be correctly forwarded. 

---

### 🛠️ Changes Made

---

### ✅ Checklist

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link: [ML-12484](https://iguazio.atlassian.net/browse/ML-12484)
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [ ] No

---

### 🔍️ Additional Notes

[ML-12484]: https://iguazio.atlassian.net/browse/ML-12484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mlrun/datastore/datastore_profile.py
```diff
@@ -619,9 +619,10 @@ def url(self, subpath):
 
 class HuggingFaceProfile(DatastoreProfile):
     type: str = pydantic.v1.Field("huggingface")
-    _private_attributes = ("token", "model_kwargs")
+    _private_attributes = ("token", "endpoint", "model_kwargs")
```
davesh0812: I think that only the token should be private, wdyt?

davesh0812 (commented):
should we add also auth option to the hf endpoint?

mlrun/datastore/datastore_profile.py
tomerm-iguazio: removed, tnx!

mlrun/datastore/datastore_profile.py
```diff
@@ -622,21 +622,25 @@ class HuggingFaceProfile(DatastoreProfile):
...
             "HF_TOKEN": self.token,
+            "HF_ENDPOINT": self.endpoint,
             "HF_DEVICE": self.device,
             "HF_DEVICE_MAP": self.device_map,
             "HF_TRUST_REMOTE_CODE": self.trust_remote_code,
+            "HF_MAX_WORKERS": self.max_workers,
             "HF_MODEL_KWARGS": self.model_kwargs,
         }
-        return {k: v for k, v in keys.items() if v}
+        return {k: v for k, v in keys.items() if v is not None}
```
davesh0812, line 643: Worth calling out this filter change in the PR description — switching from if v to if v is not None affects all existing fields, not just the new ones. Values like device=0 and trust_remote_code=False were previously dropped as falsy and will now be forwarded. I think the new behavior is correct (device=0 was effectively broken before), but it's a behavior change beyond what the title suggests and should be documented.

davesh0812 (approved):

PR #9642, opened 2026-04-30
[Serving] Add APIHandler support for multiple models mapping

tomerm-iguazio:
### 📝 Description
This PR replaces the global body_map mechanism with a per-endpoint BodyMappings object attached directly to each EndpointConfig. All matching endpoints (star, template, exact) contribute their body maps to a merged result, with more specific endpoints winning on conflict - giving a natural inheritance model where a star defines base parameters and specific endpoints extend or override them. Access control (ALLOW/FORBID) is still determined solely by the most specific match.

This is a breaking change — the old global body_map, add_body_mapping, and remove_body_mapping APIs on APIHandlerConfig are removed, and callers must migrate to BodyMappings + input_body_mappings on add_endpoint_handler.

---

### 🛠️ Changes Made

- `BodyMappings` — new class holding per-endpoint input (and future output) parameter mappings. Supports add_mapping (with overwrite+warn on duplicate source or destination), remove_mapping, and mandatory flag enforcement at request time. JSONPath expressions are compiled once at step init and validated eagerly — invalid expressions raise MLRunValueError immediately.

- `EndpointConfig` — extended with input_body_mappings: BodyMappings | None and output_body_mappings: BodyMappings | None parameters on add_endpoint_handler.

- `_APIHandlerStep._merge_body_maps` — new method that merges body maps from all matched endpoints (lowest to highest priority). Handles two conflict types: same destination (higher priority source wins), same source with different destination (stale destination is removed to prevent the value being passed to two params).

- `_compile_patterns` — now raises MLRunValueError for overlapping template endpoints (e.g. /a/{key} and /a/{user_id} on same method).

Breaking change:

- `APIHandlerConfig.body_map, add_body_mapping()` and `remove_body_mapping()` are removed. Callers must migrate to attaching a BodyMappings instance via input_body_mappings on add_endpoint_handler. The old global body map was applied to all endpoints uniformly — the new model requires explicit attachment per endpoint.

---

### ✅ Checklist

---

### 🧪 Testing
  
  

---

### 🔗 References
- Ticket link: [ML-12459](https://iguazio.atlassian.net/browse/ML-12459)
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [x] Yes (explain below)
- [ ] No

As described above.

---

### 🔍️ Additional Notes
This PR will not handle conflicts between query, path and body parameters in the new format (body mapping per endpoint).
This issue will be addressed in a separate ticket: [ML-12515](https://ecliptos.atlassian.net/browse/ML-12515)

[ML-12459]: https://iguazio.atlassian.net/browse/ML-12459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/serving/api-handler.md
```diff
@@ -102,37 +103,91 @@ When the handler allows a request, it can extract parameters from three sources:
...
+)
 ```
 
+`add_mapping` parameters:
+
+| Parameter          | Description                                                                 |
+|--------------------|-----------------------------------------------------------------------------|
+| `destination_path` | Name of the keyword argument passed to the next step.                       |
+| `source_json_path` | JSONPath expression evaluated against the request body dict.                |
+| `mandatory`        | If `True` (default `False`), a missing field fails the request with HTTP 400. |
```
jond01:
```suggestion
| `mandatory`        | If `True` (default `False`), a missing field fails the request with the error code HTTP 400 - bad request. |
```

docs/serving/api-handler.md
```diff
@@ -102,37 +103,91 @@ When the handler allows a request, it can extract parameters from three sources:
...
-- `body_map` applies to **all** ALLOW endpoints in the config. Because the mapping is global, a `body_map` that matches one endpoint's body format may cause 422 errors for other endpoints whose requests do not share that format.
+- No match on a **mandatory** field → the request fails with HTTP 400 (Bad Request).
+- No match on an optional field → the parameter is silently omitted.
+- Non-dict body → body mappings are silently skipped.
+- Calling `add_mapping` with a duplicate `destination_path` or `source_json_path` overwrites the existing entry and logs a warning.
 
-```{note}
-The strict 422 behavior for missing body mappings and non-JSON bodies is a known limitation in the current release and will be relaxed in a future release to allow more flexible body handling.
-```
+To remove a mapping by destination: `bm.remove_mapping("model_name")`
```
jond01: `"model_name"` - mention whether it is the source or destination.

mlrun/serving/api_handler.py
```diff
@@ -420,65 +458,80 @@ def do(
...
             raise
 
-    def _match_endpoint(
-        self, method: HTTPMethod, path: str
-    ) -> tuple[str | None, dict[str, str]]:
-        """Find matching endpoint key for the given method and path.
+    def _raise_not_found_endpoint(self, method: str, normalized_path: str) -> None:
+        # Check if path exists with any method (for 405 vs 404 distinction)
+        path_exists = any(
+            e.path == normalized_path for e in self.config.endpoints.values()
```
royischoss: The comparison here can miss cases of /users/{id} == /users/123 → always False, can cause 404 instead of 405

mlrun/runtimes/nuclio/serving.py
```diff
@@ -56,64 +53,207 @@
...
-            ) from e
+        for endpoint_key, ep in endpoints.items():
+            if isinstance(ep, EndpointConfig):
+                self._endpoints[endpoint_key] = ep
+            else:
+                # dict → deserialization path
+                # Manually deserialize nested BodyMappings since ModelObj does not do it automatically
+                body_mappings_dict = ep.get("input_body_mappings")
+                input_body_mappings = (
+                    BodyMappings.from_dict(body_mappings_dict)
```
royischoss, line 328: please add a comment for old dictionary format

mlrun/serving/api_handler.py
```diff
@@ -64,34 +74,35 @@ def __init__(
...
-        self._parsed_body_map = {}
-        if self.config.body_map:
-            for param_name, jsonpath_expr in self.config.body_map.items():
-                try:
-                    self._parsed_body_map[param_name] = jsonpath_parse(jsonpath_expr)
-                except (JsonPathLexerError, JsonPathParserError) as exc:
-                    raise mlrun.errors.MLRunInvalidArgumentError(
-                        f"Invalid JSON path expression for parameter '{param_name}': "
-                        f"'{jsonpath_expr}'. Error: {exc}"
-                    ) from exc
```
jond01, line 77: This is a regression - you do not check for errors in the JSON path. Please recover this functionality

mlrun/runtimes/nuclio/serving.py
```diff
@@ -232,56 +381,16 @@ def remove_endpoint_handler(
...
-            )  # Multiple matches return list
-        """
-        # Validate JSONPath expression by parsing it
-        try:
-            jsonpath_parse(json_path)
-        except (JsonPathLexerError, JsonPathParserError) as exc:
-            raise mlrun.errors.MLRunValueError(
-                f"Invalid JSON path expression for parameter '{parameter_name}': "
-                f"'{json_path}'. Error: {exc}"
-            ) from exc
```
jond01, line 266: Same here

tests/serving/test_api_handler.py
```diff
@@ -672,16 +820,22 @@ def test_multiple_star_patterns_different_methods(self) -> None:
...
+        assert len(matches) == 1
+        m = matches[0]
+        ep, params = m.endpoint, m.path_params
+        assert ep.get_endpoint_key() == "POST:/write/*"
+        assert params == {}
         # GET /write should NOT match (wrong method)
-        match, params = step._match_endpoint(HTTPMethod.GET, "/write/something")
-        assert match is None
+        matches = step._collect_endpoint_matches(HTTPMethod.GET, "/write/something")
+        assert matches == []
```
royischoss, line 836: _merge_body_maps has no direct unit test — the existing test_precedence_* tests verify which endpoints are collected, but don't exercise the merge logic itself (same destination conflict, same source / different destination
  conflict). A request that hits both a star and a specific endpoint with overlapping mappings could silently pass the wrong value through.                                                                                             
                                                                                                                                                                                                                                      
  Suggested test to add:                                                                                                                                                                                                                
                                                                                                                                                                                                                                      
  def test_merge_body_maps_star_overridden_by_specific(self) -> None:
      star_bm = BodyMappings()                                                                                                                                                                                                          
      star_bm.add_mapping("$.model", destination_path="model")
      star_bm.add_mapping("$.stream", destination_path="stream")                                                                                                                                                                        
                                                                                                                                                                                                                                        
      specific_bm = BodyMappings()                                                                                                                                                                                                      
      specific_bm.add_mapping("$.model", destination_path="model_name")  # same source, new dest                                                                                                                                        
                                                                                                                                                                                                                                        
      config = APIHandlerConfig()
      config.add_endpoint_handler("/v1/*", HTTPMethod.POST, APIHandlerAction.ALLOW, input_body_mappings=star_bm)                                                                                                                        
      config.add_endpoint_handler("/v1/predict", HTTPMethod.POST, APIHandlerAction.ALLOW, input_body_mappings=specific_bm)                                                                                                              
   
      step = _APIHandlerStep(config=config)                                                                                                                                                                                             
      matches = step._collect_endpoint_matches(HTTPMethod.POST, "/v1/predict")                                                                                                                                                        
      merged = step._merge_body_maps(matches)                                                                                                                                                                                           
                                                                                                                                                                                                                                      
      assert "model_name" in merged       # specific wins                                                                                                                                                                               
      assert "model" not in merged        # stale dest removed                                                                                                                                                                        
      assert "stream" in merged           # inherited from star

royischoss (commented):
Hey looks good small comments

jond01 (commented):
Partial review

mlrun/runtimes/nuclio/serving.py
```diff
@@ -56,64 +53,207 @@
...
+                )
+
+        # Check for duplicate source_json_path or destination_path — overwrite and warn if found.
+        # destination_path=None means "validate existence only"; don't treat two None-dest
+        # mappings as duplicates of each other — they may guard different source fields.
+        for i, existing in enumerate(self.mappings):
+            dest_conflict = (
+                destination_path is not None
+                and existing.get("destination_path") == destination_path
+            )
```
jond01, line 176: This scales as N^2.

My AI:

Building up N mappings via successive add_mapping calls is O(N²): each call scans the full list of existing mappings.

  In practice N is tiny (a handful of mappings per endpoint), so this is unlikely to matter at runtime. But the data structure choice is the root cause —
  self.mappings is a list[dict], so every duplicate check is a linear scan.

  If you want O(N) total, the storage should be two dicts (one keyed by source_json_path, one keyed by destination_path) with the list reconstructed only for
  serialization, or a single dict keyed by destination_path with a reverse-index for source. That also cleans up _merge_body_maps, remove_mapping, and the
  to_dict/from_dict paths.

tests/serving/test_api_handler.py
tomerm-iguazio: `test_same_source_different_dest_specific_wins` in test_body_map.py:987 already covers this case - stale destination removal and specific-wins-over-star merge logic are both exercised there.

mlrun/runtimes/nuclio/serving.py
```diff
@@ -56,64 +53,207 @@
...
+        )
+
+    def remove_mapping(self, destination_path: str) -> None:
+        """Remove the mapping with the given destination_path. No-op if not found.
+
+        :param destination_path: The destination key to remove.
+        """
+        self.mappings = [
+            m for m in self.mappings if m.get("destination_path") != destination_path
+        ]
```
jond01: Why not index by `destination_path`? It would make this and the previous comment much cleaner

mlrun/serving/api_handler.py
tomerm-iguazio: JSONPath validation is handled in _compile_patterns at `_APIHandlerStep.__init__ `(line 202-208) — invalid expressions raise MLRunValueError with the endpoint key and expression.
Adding it to add_mapping too would parse the same expression twice for no real benefit.

jond01: But there is a benefit - immediate response to the user when the JSON Path is invalid.
Initializing the graph's step happens only at deployment or mock testing.

royischoss (approved):
LGTM

mlrun/runtimes/nuclio/serving.py
```diff
@@ -56,64 +55,256 @@
...
-        body_map: dict[str, str] | None = None,
+        endpoints: dict[str, dict | EndpointConfig] | None = None,
         include_url_info: bool = False,
     ):
         self.enabled = enabled
-        self._endpoints = endpoints or {}
-        self._body_map = body_map or {}
+        self._endpoints: dict[str, EndpointConfig] = {}
+        if endpoints:
+            self.endpoints = endpoints
```
jond01: Why not keeping
```suggestion
        self._endpoints: dict[str, EndpointConfig] = endpoints or {}
```
?

mlrun/runtimes/nuclio/serving.py
```diff
@@ -177,47 +368,56 @@ def _validate_http_method(http_method: HTTPMethod | str) -> HTTPMethod:
...
         :param http_method: HTTP method for the endpoint (``HTTPMethod`` enum or string like ``"GET"``, ``"POST"``)
-        :param action: Action to take for this endpoint (:py:class:`~mlrun.common.schemas.serving.APIHandlerAction`)
+        :param action: Action to take for this endpoint (:py:class:`~mlrun.common.schemas.APIHandlerAction`)
         :param description: Optional description of the endpoint
+        :param input_body_mappings: Optional input :class:`BodyMappings` for this endpoint (REST → graph).
+            If ``None``, the request body is passed through as-is.
         :raises mlrun.errors.MLRunValueError: If the path contains an invalid wildcard ``*`` pattern
         """
         http_method = self._validate_http_method(http_method)
         path = self._normalize_path(path)
```
jond01: Move this normalization to `EndpointConfig` class

mlrun/runtimes/nuclio/serving.py
```diff
@@ -56,64 +55,256 @@
...
+
+    def __init__(
+        self,
+        path: str = "",
+        http_method: HTTPMethod | str = HTTPMethod.POST,
+        action: schemas.APIHandlerAction = schemas.APIHandlerAction.ALLOW,
+        description: str | None = None,
+        input_body_mappings: BodyMappings | None = None,
+    ) -> None:
+        self.path = path
```
jond01: You have to normalize the path here. Otherwise, the two paths `/api/v1/models` and `api/v1/models` will appear in the `APIHandlerConfig` as two different endpoints, but in a Nuclio runtime they are the same.

mlrun/runtimes/nuclio/serving.py
```diff
@@ -56,64 +55,256 @@
...
+        self.http_method = (
+            HTTPMethod(http_method) if isinstance(http_method, str) else http_method
+        )
```
jond01: Please use `_validate_http_method`, which handles this more accurately.

mlrun/serving/api_handler.py
```diff
@@ -157,48 +190,58 @@ def _compile_patterns(
...
-                        f"parameter(s) in pattern '{compiled_pattern.pattern}'. "
-                        f"Rename the body_map parameter(s) or the path template "
-                        f"placeholder(s) to avoid ambiguity."
+            # Compile input_body_mappings for this endpoint (any pattern type)
+            if ep.input_body_mappings:
+                compiled_map: dict[str, tuple[Any, bool]] = {}
+                for mapping in ep.input_body_mappings.mappings:
+                    try:
+                        compiled_expr = jsonpath_ng.parse(mapping["source_json_path"])
+                    except Exception as e:
```
jond01: You should check specific errors, not every execption.

mlrun/serving/api_handler.py
```diff
@@ -64,34 +74,35 @@ def __init__(
             self.config = mlrun.runtimes.nuclio.serving.APIHandlerConfig()
         self.context = context
 
-        # Parse JSONPath expressions during initialization for performance and early error detection
-        self._parsed_body_map = {}
-        if self.config.body_map:
-            for param_name, jsonpath_expr in self.config.body_map.items():
-                try:
-                    self._parsed_body_map[param_name] = jsonpath_parse(jsonpath_expr)
-                except (JsonPathLexerError, JsonPathParserError) as exc:
```
jond01, line 73: You should catch only these errors

tests/system/runtimes/test_serving.py
```diff
@@ -136,42 +161,35 @@ def test_api_handler_forbidden(self) -> None:
 
     @pytest.mark.parametrize("engine", ["sync", "async"])
     def test_api_handler_with_body_mapping(self, engine: str) -> None:
-        """Test API handler with body_map JSONPath extraction."""
+        """Test API handler with per-endpoint input_body_mappings JSONPath extraction."""
```
jond01, line 164: Did it pass?

tests/serving/test_api_handler.py
```diff
@@ -1560,44 +1719,12 @@ class TestEndpointKeyHelpers:
...
         """Test combining method and path into endpoint key"""
-        key = _combine_serving_endpoint_key(HTTPMethod.GET, "/api/test")
+        key = combine_serving_endpoint_key(HTTPMethod.GET, "/api/test")
         assert key == "GET:/api/test"
 
-        key = _combine_serving_endpoint_key(HTTPMethod.POST, "/predict")
+        key = combine_serving_endpoint_key(HTTPMethod.POST, "/predict")
         assert key == "POST:/predict"
 
-    def test_split_serving_endpoint_key(self) -> None:
```
jond01, line 1569: Why did you remove these tests?

tests/serving/test_api_handler.py
jond01, line 1: Where do you have a test where the destination of the JSON path is taken from the most specific endpoint and there are overlaps?

jond01 (changes requested):

jond01:
> * `_compile_patterns` — now raises MLRunValueError for overlapping template endpoints (e.g. /a/{key} and /a/{user_id} on same method), consistent with Flask/FastAPI behavior.

The claim isn't accurate — FastAPI silently accepts overlapping templates and routes to the first registered match. It doesn't raise at startup or at request
  time.


Minimal repro:

```py
from fastapi import FastAPI
from fastapi.testclient import TestClient

app = FastAPI()

@app.get("/a/{key}")
def by_key(key: str):
    return {"matched": "by_key", "key": key}

@app.get("/a/{user_id}")
def by_user(user_id: str):
    return {"matched": "by_user", "user_id": user_id}

client = TestClient(app)
print(client.get("/a/foo").json())
# {'matched': 'by_key', 'key': 'foo'}   ← first-registered wins, no error
```

mlrun/runtimes/nuclio/serving.py
tomerm-iguazio: I don't want to lost the hint annotation, so I update it with your code + hint annotation

tests/serving/test_api_handler.py
tomerm-iguazio: `split_serving_endpoint_key` was removed, and the test for it was deleted along with it.

tests/serving/test_api_handler.py
tomerm-iguazio: yep, `test_same_source_different_dest_specific_wins`

mlrun/serving/utils.py
```diff
@@ -182,12 +183,67 @@ def to_dict(
...
 
 
-def _combine_serving_endpoint_key(method: HTTPMethod, path: str) -> str:
+def combine_serving_endpoint_key(method: HTTPMethod, path: str) -> str:
     """Combine method and path to create a unique endpoint key"""
     return f"{method.value}:{path}"
 
 
-def _split_serving_endpoint_key(endpoint_key: str) -> tuple[HTTPMethod, str]:
+def split_serving_endpoint_key(endpoint_key: str) -> tuple[HTTPMethod, str]:
```
jond01: ☝️

tests/serving/test_api_handler.py
jond01: Well, it was not removed - see https://github.com/mlrun/mlrun/pull/9642#discussion_r3232750901.

Edit: indeed, I see that you don't use it anymore - so please remove it if not needed :)

jond01 (approved):
Approved up to this comment: https://github.com/mlrun/mlrun/pull/9642#discussion_r3232751805

Good work!!

mlrun/serving/utils.py
tomerm-iguazio: tnx!

davesh0812 (approved):

PR #9643, opened 2026-04-30
[Model Monitoring/Test] Fix test_app_flow smoke test on open-source CE

assaf758:
### 📝 Description

`test_app_flow[True-True]` (the system smoke test for model monitoring) was failing for several independent reasons. This PR fixes them so the smoke gate passes against open-source CE on k3s. Each commit is a self-contained fix:

1. **Parallel app deploys** — `_set_and_deploy_monitoring_apps` had a `max_workers=1 if (os.cpu_count() or 1) <= 8 else None` guard that serialised monitoring-app deploys on small CPUs (the 8-CPU CI runner included). With ~13 min per kaniko build, `evidently-app-test` only began building at T+16 min and the test hit its 30-min `pytest-timeout` 2 min before it finished. The guard was originally there to avoid concurrent GitHub clones; that concern is moot now that the wheel is served from a local HTTP server.

2. **MM functions honour `function_defaults.image_by_kind.nuclio`** — `MonitoringDeployment.deploy_*` (controller / writer / stream / histogram-data-drift) hardcoded `image="mlrun/mlrun"` as the default, bypassing the configured nuclio default image. This meant an mlrun-ce deployment that pre-staged its base image into a local registry and configured `MLRUN_FUNCTION_DEFAULTS__IMAGE_BY_KIND__NUCLIO` accordingly still saw MM infra functions cold-pull from `ghcr.io`. See https://github.com/mlrun/mlrun/pull/9644

3. **Plumb `None` default through the SDK→API chain** — the previous fix (commit 2) added a `None`-resolution at the API CRUD layer, but every layer above it (`Project.enable_model_monitoring`, `RunDBInterface`, `HTTPRunDB`, `NopDB`, FastAPI endpoint) had its own hardcoded `"mlrun/mlrun"` default that overrode it before the call ever reached the resolver. Switch all four call sites to `None` and have `HTTPRunDB` skip the `image` query param when the caller didn't specify one, so the server-side resolution actually fires.

4. **Test-side workaround for ML-12522** — `_resolve_invocation_url` was changed in #9578 to default scheme-less external URLs to `https://`, which broke NodePort/plain-HTTP setups (smoke runs on k3s via NodePort) with `SSL: WRONG_VERSION_NUMBER`. The proper fix lives in [ML-12522](https://ecliptos.atlassian.net/browse/ML-12522) — route API-gateway invocations through `APIGateway.invoke_url` so `_resolve_invocation_url` can default back to `http://`. Until that's resolved, `TestMonitoringAppFlow._deploy_model_serving` post-mutates `serving_fn.status.external_invocation_urls` to prefix `http://`, but only for IPv4 `host:port` form (NodePort) — iguazio's TLS-terminated ingress URLs (hostname/path form) keep the `https://` default.

---

### 🛠️ Changes Made

- `tests/system/model_monitoring/test_app.py`
  - `TestMonitoringAppFlow._set_and_deploy_monitoring_apps`: drop the CPU-count gate on `ThreadPoolExecutor`.
  - `TestMonitoringAppFlow._deploy_model_serving`: post-deploy, force `http://` on NodePort-form `external_invocation_urls` (IPv4 + port + no path). New helper `_workaround_ml_12522_force_http_for_nodeport`.
- `server/py/services/api/crud/model_monitoring/deployment.py`: change `image: str = "mlrun/mlrun"` defaults to `image: str | None = None` and resolve `None` to `mlrun.mlconf.function_defaults.image_by_kind.nuclio` in the four `deploy_*` methods.
- `server/py/services/api/api/endpoints/model_monitoring.py`: same `None` default for the `enable_model_monitoring` FastAPI endpoint.
- `mlrun/projects/project.py`, `mlrun/db/httpdb.py`, `mlrun/db/base.py`, `mlrun/db/nopdb.py`: same `None` default through the SDK→API chain. `HTTPRunDB.enable_model_monitoring` omits the `image` query param when `None`.

---

### ✅ Checklist
- [x] Please run [Smoke-tests](https://github.com/mlrun/mlrun/actions/workflows/smoke-tests.yml) workflow, providing it the PR number as input (upon success, the workflow run will add label "Smoke tests: Pass" to the RP)

---

### 🧪 Testing

`test_app_flow[True-True]` smoke test passes end-to-end on open-source CE / k3s after these changes.

Reference smoke run on `feature/smoke` (with the equivalent commit set): https://github.com/mlrun/mlrun/actions/runs/25248470887 — `6 passed, 1 skipped, 736 deselected, 150 warnings in 762.09s (0:12:42)`.

---

### 🔗 References
- Ticket link: [ML-12522](https://ecliptos.atlassian.net/browse/ML-12522) — proper fix; test-side workaround in commit 4 will be removed once ML-12522 lands.
- Related: [#9578](https://github.com/mlrun/mlrun/pull/9578) — introduced the `https://` default that necessitated commit 4's workaround.

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

`enable_model_monitoring`'s default `image` value changes from the literal `"mlrun/mlrun"` to `None` (resolved server-side to `mlrun.mlconf.function_defaults.image_by_kind.nuclio`, whose default value is also `"mlrun/mlrun"`). Behavior is unchanged for callers passing an explicit `image` and for environments using the default `image_by_kind.nuclio` config.

---

### 🔍️ Additional Notes

Once ML-12522 lands and `_resolve_invocation_url` defaults back to `http://`, commit 4 (`deab343c3`) can be dropped — the test workaround becomes a harmless no-op (the `https://` default it bypasses no longer fires).

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -2654,7 +2654,7 @@ class MlrunProject(ModelObj):
     def enable_model_monitoring(
         self,
         base_period: int = 10,
-        image: str = "mlrun/mlrun",
+        image: str | None = None,
         *,
         deploy_histogram_data_drift_app: bool = True,
         wait_for_deployment: bool = False,
@@ -2675,7 +2675,8 @@ class MlrunProject(ModelObj):
                                                   (which is also the minimum value for production environments).
         :param image:                             The image of the model monitoring controller, writer, monitoring
                                                   stream & histogram data drift functions, which are real time nuclio
-                                                  functions. By default, the image is mlrun/mlrun.
+                                                  functions. Defaults to
+                                                  ``mlrun.mlconf.function_defaults.image_by_kind.nuclio``.
         :param deploy_histogram_data_drift_app:   If true, deploy the default histogram-based data drift application:
             :py:class:`~mlrun.model_monitoring.applications.histogram_data_drift.HistogramDataDriftApplication`.
             If false, and you want to deploy the histogram data drift application
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -2742,7 +2742,7 @@ class MlrunProject(ModelObj):
     def update_model_monitoring_controller(
         self,
         base_period: int = 10,
-        image: str = "mlrun/mlrun",
+        image: str | None = None,
         *,
         wait_for_deployment: bool = False,
     ) -> None:
@@ -2753,7 +2753,8 @@ class MlrunProject(ModelObj):
                                     is triggered. By default, the base period is 10 minutes.
         :param image:               The image of the model monitoring controller, writer & monitoring
                                     stream functions, which are real time nuclio functions.
-                                    By default, the image is mlrun/mlrun.
+                                    Defaults to
+                                    ``mlrun.mlconf.function_defaults.image_by_kind.nuclio``.
         :param wait_for_deployment: If true, return only after the deployment is done on the backend.
                                     Otherwise, deploy the controller on the background.
         """
--- a/tests/system/model_monitoring/test_app.py
+++ b/tests/system/model_monitoring/test_app.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import concurrent.futures
+import ipaddress
 import json
 import pickle
 import tempfile
@@ -566,7 +567,7 @@ class TestMonitoringAppFlow(TestMLRunSystemModelMonitoring, _V3IORecordsChecker)
     def _workaround_ml_12522_force_http_for_nodeport(
         fn: mlrun.runtimes.nuclio.serving.ServingRuntime,
     ) -> None:
-        """Workaround for https://ecliptos.atlassian.net/browse/ML-12522.
+        """Workaround for ML-12522.
 
         Nuclio reports ``status.external_invocation_urls`` as scheme-less.
         Since mlrun #9578, ``_resolve_invocation_url`` prepends ``https://``
@@ -580,8 +581,6 @@ class TestMonitoringAppFlow(TestMLRunSystemModelMonitoring, _V3IORecordsChecker)
         invocations through ``APIGateway.invoke_url`` and
         ``_resolve_invocation_url`` defaults back to ``http://``.
         """
-        import ipaddress
-
         for i, url in enumerate(fn.status.external_invocation_urls or []):
             if "://" in url or "/" in url:
                 continue
--- a/tests/system/model_monitoring/test_app.py
+++ b/tests/system/model_monitoring/test_app.py
@@ -559,7 +559,38 @@ class TestMonitoringAppFlow(TestMLRunSystemModelMonitoring, _V3IORecordsChecker)
             serving_fn.spec.image = serving_fn.spec.build.image = cls.image
 
         serving_fn.deploy()
+        cls._workaround_ml_12522_force_http_for_nodeport(serving_fn)
         return serving_fn
+
+    @staticmethod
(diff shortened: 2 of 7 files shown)


tests/system/model_monitoring/test_app.py
```diff
@@ -562,8 +559,39 @@ def _deploy_model_serving(
...
 
         serving_fn.deploy()
+        cls._workaround_ml_12522_force_http_for_nodeport(serving_fn)
         return serving_fn
 
+    @staticmethod
+    def _workaround_ml_12522_force_http_for_nodeport(
+        fn: mlrun.runtimes.nuclio.serving.ServingRuntime,
+    ) -> None:
+        """Workaround for https://ecliptos.atlassian.net/browse/ML-12522.
```
gtopper: Probably better leave the URL out because
1. this is open source code
2. in case the domain changes again in the future
3. users with access know where to find it
```suggestion
        """Workaround for ML-12522.
```

gtopper: Also opened #9645 to address inconsistency in existing code.

mlrun/projects/project.py
jond01, line 1: What about update_model_monitoring_controller? It also has

```py
        image: str = "mlrun/mlrun",
```

Also in the httpdb method and others, including the API endpoint

tests/system/model_monitoring/test_app.py
```diff
@@ -562,8 +559,39 @@ def _deploy_model_serving(
...
+        setups (open-source CE / k3s) with ``SSL: WRONG_VERSION_NUMBER``.
+        Iguazio setups expose functions via TLS-terminated ingress
+        (hostname/path form) and must keep the ``https://`` default.
+
+        Only rewrite the NodePort form (IPv4 ``host:port`` with no path)
+        to ``http://`` here. Remove once ML-12522 routes API-gateway
+        invocations through ``APIGateway.invoke_url`` and
+        ``_resolve_invocation_url`` defaults back to ``http://``.
+        """
+        import ipaddress
```
jond01: Why importing here and not top-level?

jond01 (approved):
Nice, I had two minor comments

jond01:
> 1. that concern is moot now that the wheel is served from a local HTTP server.

I don't understand that point

tests/system/model_monitoring/test_app.py
```diff
@@ -562,8 +559,39 @@ def _deploy_model_serving(
...
+
+        for i, url in enumerate(fn.status.external_invocation_urls or []):
+            if "://" in url or "/" in url:
+                continue
+            host = url.split(":", 1)[0]
+            try:
+                ipaddress.ip_address(host)
+            except ValueError:
```
jond01, line 590: In what situations we expect this?

assaf758: The ValueError can happen when nuclio external-url is scheme-less, path-less hostname (for example "my-service:8080").
In practice this doesn't happen — ig3/ig4 ingress always has a path and CE NodePort always uses an IP.
So it's probably just a defensive guard, but anyway the code will hopefully be removed shortly when ml-12522 is resolved.

assaf758:
> > 1. that concern is moot now that the wheel is served from a local HTTP server.
> 
> I don't understand that point

See the "brother" PR that I pointed to - https://github.com/mlrun/mlrun/pull/9644
We were having github rate-limiting us due test_app[True,True] requesting 8 concurrent kaniko builds, each doing `python pip install github..`. So as part of the test preparation, we are building an mlrun pip wheel and serving it to kaniko from a local http server.

PR #9647, opened 2026-05-04
[Secrets] Secret auto mount with cleartext

theSaarco:
### 📝 Description
`set_env_vars_from_secret` (auto-mount type `secret_env`) can mount Kubernetes Secret keys as pod environment variables, but it has no way to also inject plain `key=value` environment variables in the same call. This forces users to put non-sensitive values (e.g. a storage account name) into a secret.
Thid PR adds `cleartext_env: str | dict[str, str] | None = None` to the function signature. Parse the string form (semicolon-separated `key:value` pairs, splitting on the *first* colon per token) into a dict at function-entry time, outside the closure.
Inside the closure, after the existing secret-mounting logic, iterate the dict and call `runtime.set_env(k, v)` for each pair.

---

#### Files Modified

| File | Change |
|------|--------|
| `mlrun/runtimes/mounts.py` | Added `cleartext_env` param, inline string-parsing block, closure loop; updated docstring with two new `Usage` examples and `:param cleartext_env:` |
| `tests/common_fixtures.py` | Extended `assert_env_from_secret` with optional `cleartext_env: dict[str, str] | None` param; asserts plain `{"name": k, "value": v}` entries |
| `tests/runtimes/test_base.py` | Added `@pytest.mark.parametrize("with_cleartext", [True, False])` — 4-combo matrix; plain-string and base64-JSON paths both assert via updated helper |
| `tests/runtimes/test_mounts.py` | Parameterized `test_auto_mount_secret_env` over `with_keys x with_cleartext` (4 combos); added `test_set_env_vars_from_secret_cleartext_env_string_validation` (4 cases) |

#### Key Implementation Details

- **Fail-fast**: `MLRunInvalidArgumentError` is raised at factory-call time (before the closure is
  returned) when a string token is missing its `:` separator. Error is caught early, not inside
  the closure.
- **Colon-in-value**: In string form, `token.find(":")` finds the *first* colon, so values like
  `http://example.com` work correctly in dict form; in string form only the leftmost `:` is the
  delimiter (colons in values not supported for string form — documented in docstring).
- **Defensive copy**: `dict(cleartext_env)` is used when input is already a dict, so mutation of
  the caller's dict after the call has no effect on the closure.
- **Backward compatibility**: `_filter_modifier_params` (pod.py) only passes keys present in the
  params dict that match the function signature. Old configs without `cleartext_env` never forward
  the key, so the default `None` is used transparently.

#### Edge Cases Handled

1. `cleartext_env=None` (default) — `cleartext_env_dict = {}` — no extra env vars (backward compat).
2. `cleartext_env={}` (empty dict) — `dict({}) = {}` — no extra env vars.
3. Whitespace around keys/values in string form — `.strip()` applied to tokens and split halves.
4. Trailing `;` in string form — empty tokens skipped via `if not token: continue`.
5. `keys` absent from auto-mount params dict — `_filter_modifier_params` drops it, `keys_list=[]`,
   `envFrom` path taken alongside cleartext vars (AC6/AC8).

---

### ✅ Checklist

---

### 🧪 Testing
Unit tests:
**Test command**: `python -m pytest tests/runtimes/test_mounts.py tests/runtimes/test_base.py -x -v`
**Result**: 53 passed, 25 warnings

**New tests added**: 12
- 8 parametrized `test_auto_mount_secret_env` combos (`with_keys x with_cleartext`, across both
  test files and both params formats).
- 4 string-validation cases in `test_set_env_vars_from_secret_cleartext_env_string_validation`.

---

### 🔗 References
- Ticket link: ML-12495
- Design docs links:
- External links:

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [X] No

---


liranbg (approved):
neat!

rokatyy (approved):
🚀

PR #9649, opened 2026-05-04
[Nuclio] Extend init container to vanilla Nuclio/Serving for store:// sources

alxtkr77:
### 📝 Description

Vanilla Nuclio and Serving functions cannot be deployed from a `store://`
CodeArtifact source today — Nuclio's native builder errors out with
"Couldn't resolve code entry type from source" because store:// is not a
type Nuclio knows. Until now the source-loader init-container path that
solves this was gated to Application runtime only.

This PR removes that gate for store:// sources on vanilla Nuclio and
Serving runtimes: the existing `mlrun load-source` init container is
re-used to fetch the artifact at deploy time, and a small generated
loader stub re-exports the user's real handler at runtime.

---

### 🛠️ Changes Made

- `services/api/crud/runtimes/nuclio/function.py::_compile_function_config`:
  for non-Application kinds with a store:// source, stash the URI to
  `status.application_source`, clear `spec.build.source` so the Nuclio
  builder doesn't see it, generate a dedicated loader stub as
  `functionSourceCode`, override `spec.handler` to point at the loader,
  and configure the source-loader init container.
- `_configure_source_loader_init_container`: `sidecar` parameter is now
  optional. When `None` (vanilla Nuclio/Serving), patches the main
  function container via the new `_patch_main_function_container` helper.
- `_patch_main_function_container`: injects PYTHONPATH into
  `function.spec.config["spec.env"]` (volumes are wired by the parent).
- `_ensure_source_loader_init_container`: dedupes init containers in BOTH
  `function.spec.config` and `function.spec.base_spec` — without the
  `base_spec` dedupe, redeploys hit "Duplicate value" because
  `extend_config` later merges `base_spec` with `config` and produces a
  duplicate.
- `NuclioStatus` gains `application_source` (URI stash for redeploy) and
  `original_handler` (handler stash for redeploy).
- `mlrun.common.constants.STORE_URI_HANDLER_LOADER_MODULE`: name of the
  generated loader module. A dedicated module name avoids the
  `/opt/nuclio` `sys.path[0]` shadow that would otherwise hide the
  user's real handler module loaded by the init container.

Git/archive sources with `pull_at_runtime` stay on the native Nuclio
path (`codeEntryType="git"/"archive"`) — only `store://` is redirected,
since Nuclio handles git/archive natively but not store://.

---

### ✅ Checklist

- [x] I updated the documentation (if applicable) — N/A, internal compile path

---

### 🧪 Testing

**Unit tests** (`server/py/services/api/tests/unit/crud/runtimes/nuclio/test_helpers.py`):

- `test_should_fetch_source_code_for_nuclio_with_store_uri`
- `test_compile_nuclio_function_with_store_source_has_init_container[nuclio|serving]`
- `test_nuclio_store_source_preserved_on_redeploy`
- `test_compile_nuclio_function_with_store_source_redeploy_is_idempotent` —
  pins the "Duplicate value" regression: init-container, MLRUN_REAL_HANDLER
  env, functionSourceCode, original_handler all stay single/stable on rebuild.
- `test_compile_nuclio_function_with_store_source_requires_handler` —
  fail-fast at compile time if no handler is provided.
- `test_compile_nuclio_function_with_custom_source_code_target_dir` —
  override flows through to the loader stub embedded path.
- `test_compile_nuclio_function_with_git_source_skips_loader_branch` —
  negative test: git sources don't accidentally pick up the loader.
- `test_patch_main_function_container_pythonpath_merge_logic` (4
  parametrized cases): empty / prepend / already-present / multi-entry.

**System tests** (`tests/system/runtimes/test_nuclio_store_uri.py` — new):

- `test_deploy_and_artifact_update`: log_code_file v1 → deploy → invoke
  (asserts init container + status.application_source preserved + v1
  response). Re-log v2 under same key+tag → redeploy → invoke (asserts
  v2 response). Proves the init container fetches the *current* artifact
  on each deploy, not a cached copy.
- `test_cross_runtime_same_artifact`: same store:// CodeArtifact consumed
  by both job and nuclio runtimes (different handler files per runtime).

Both system tests run end-to-end on vmdev76 with `MLRUN_TEST_IMAGE` set
to a runnable mlrun image.

---

### 🔗 References

Jira (this PR): ML-12480
Parent feature: ML-11980

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### DIFF
--- a/mlrun/common/constants.py
+++ b/mlrun/common/constants.py
@@ -23,6 +23,19 @@ MLRUN_SERVING_SPEC_PATH = (
 DEFAULT_SOURCE_CODE_TARGET_DIR = "/home/mlrun_code"
 SOURCE_LOADER_INIT_CONTAINER_NAME = "mlrun-source-loader"
 SOURCE_CODE_VOLUME_NAME = "mlrun-source-code"
+# Module name for a generated stub that re-exports the user's real handler
+# under a fixed name. The dedicated namespaced name avoids the sys.path[0]
+# (=/opt/nuclio) shadow that would otherwise hide the real handler module
+# loaded into the source-loader init container's output directory.
+STORE_URI_HANDLER_LOADER_MODULE = "_mlrun_store_uri_loader"
+# BUMP THIS WHENEVER THE LOADER STUB BODY CHANGES.
+# The version marker is baked as the first line of the loader stub
+# (`# stub_version=<N>`); on (re)deploy, an existing stub whose marker
+# doesn't match the current value is re-baked, so a stub-body fix in
+# mlrun reaches existing functions on their next deploy. If you forget
+# to bump this when changing the stub, deployed functions stay on the
+# old bytes forever.
+STORE_URI_LOADER_STUB_VERSION = "1"
 MLRUN_FUNCTIONS_ANNOTATION = "mlrun/mlrun-functions"
 MYSQL_MEDIUMBLOB_SIZE_BYTES = 16 * 1024 * 1024
 MLRUN_LABEL_PREFIX = "mlrun/"
--- a/mlrun/common/constants.py
+++ b/mlrun/common/constants.py
@@ -23,18 +23,9 @@ MLRUN_SERVING_SPEC_PATH = (
 DEFAULT_SOURCE_CODE_TARGET_DIR = "/home/mlrun_code"
 SOURCE_LOADER_INIT_CONTAINER_NAME = "mlrun-source-loader"
 SOURCE_CODE_VOLUME_NAME = "mlrun-source-code"
-# Module name for a generated stub that re-exports the user's real handler
-# under a fixed name. The dedicated namespaced name avoids the sys.path[0]
-# (=/opt/nuclio) shadow that would otherwise hide the real handler module
-# loaded into the source-loader init container's output directory.
 STORE_URI_HANDLER_LOADER_MODULE = "_mlrun_store_uri_loader"
-# BUMP THIS WHENEVER THE LOADER STUB BODY CHANGES.
-# The version marker is baked as the first line of the loader stub
-# (`# stub_version=<N>`); on (re)deploy, an existing stub whose marker
-# doesn't match the current value is re-baked, so a stub-body fix in
-# mlrun reaches existing functions on their next deploy. If you forget
-# to bump this when changing the stub, deployed functions stay on the
-# old bytes forever.
+# Bump when the loader stub body changes — existing functions re-bake on
+# next deploy via the `# stub_version=<N>` marker.
 STORE_URI_LOADER_STUB_VERSION = "1"
 MLRUN_FUNCTIONS_ANNOTATION = "mlrun/mlrun-functions"
 MYSQL_MEDIUMBLOB_SIZE_BYTES = 16 * 1024 * 1024
--- a/server/py/services/api/crud/runtimes/nuclio/function.py
+++ b/server/py/services/api/crud/runtimes/nuclio/function.py
@@ -277,53 +277,13 @@ def _compile_function_config(
             "Ensure the application image is set via 'spec.image' or 'with_sidecar()'."
         )
 
-    # Configure init container when source needs runtime loading.
-    if _should_fetch_source_code(function):
-        if function.kind == mlrun.runtimes.RuntimeKinds.application:
-            # Gate the project-secret envFrom mount on source type, not runtime
-            # kind: store:// sources may resolve through DataStore profiles
-            # whose private members live in mlrun-project-secrets-<project>;
-            # git/archive sources don't need that resolution.
-            application_source = function.spec.build.source or getattr(
-                function.status, "application_source", ""
-            )
-            _configure_source_loader_init_container(
-                function,
-                # Application runtime has exactly one sidecar (the user's application container)
-                sidecar=sidecars[0],
-                client_version=client_version,
-                client_python_version=client_python_version,
-                add_project_secrets=mlrun.datastore.is_store_uri(application_source),
-            )
-        else:
-            # Vanilla Nuclio/Serving. Nuclio's native builder resolves git://
-            # and archive sources directly (codeEntryType="git"/"archive");
-            # only store:// is unsupported and gets routed through the init
-            # container path. Anything else with load_source_on_run=True is
-            # an unsupported combination and raises.
-            source = function.spec.build.source or getattr(
-                function.status, "application_source", ""
-            )
-            if mlrun.datastore.is_store_uri(source):
-                _install_store_uri_loader(
-                    function,
-                    project=project,
-                    client_version=client_version,
-                    client_python_version=client_python_version,
-                )
-            else:
-                # _should_fetch_source_code returned True, source is not
-                # store:// — must be git/archive + load_source_on_run on a
-                # non-Application kind. The init-container redirect was only
-                # designed for store://; other source types belong on
-                # Application kind (where the sidecar handles them).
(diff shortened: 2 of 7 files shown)


tests/system/runtimes/test_nuclio_store_uri.py
```diff
@@ -0,0 +1,260 @@
...
+            handler=self._function_handler,
+            image=self._function_image,
+        )
+        assert nuclio_fn.spec.build.source == nuclio_uri
+
+        job_fn = self.project.set_function(
+            func=job_uri,
+            name="cross-runtime-job",
+            kind="job",
+            handler="handler.my_func",
```
Yacouby, line 247: is this the acceptable form ? isnt it with the colon ?

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -798,30 +843,314 @@ def _should_fetch_source_code(
...
+            "if _CODE_DIR not in sys.path:\n"
+            "    sys.path.insert(0, _CODE_DIR)\n"
+            "\n"
+            '_real = os.environ.get("MLRUN_REAL_HANDLER", "")\n'
+            "if not _real:\n"
+            "    raise RuntimeError(\n"
+            '        "MLRUN_REAL_HANDLER env var not set; cannot resolve "\n'
+            '        "user handler for store:// Nuclio function."\n'
+            "    )\n"
+            '_module_name, _, _func_name = _real.rpartition(":")\n'
```
Yacouby: isn't it better to use partition as done in https://github.com/mlrun/mlrun/blob/a108aff2137090d309182f7477aaf5f159abace7/mlrun/utils/helpers.py#L1506

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -798,30 +843,314 @@ def _should_fetch_source_code(
     return (is_git_source or is_archive_source) and pull_at_runtime
 
 
+def _install_store_uri_loader(
```
Yacouby, line 856: very long function and hard to read, splitting it would make the flow easier to follow

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -798,30 +843,314 @@ def _should_fetch_source_code(
...
+    loader_module = mlrun.common.constants.STORE_URI_HANDLER_LOADER_MODULE
+    loader_handler = f"{loader_module}:handler"
+
+    if is_serving_topology:
+        # If the function was previously deployed with an explicit handler
+        # (loader stub baked, MLRUN_REAL_HANDLER env injected) and the user
+        # is now switching to topology, clear the loader-shaped state so
+        # the spec doesn't carry dead bytes. The serving wrapper handler
+        # is wired downstream by `_set_source_code_and_handler`.
+        if function.spec.function_handler == loader_handler:
```
Yacouby: this is a dead code, since in the above `is_serving_topology`, you required `function_handler` to be falsy

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -963,31 +1381,50 @@ def _patch_sidecar_for_source(
...
+    # base_spec["spec"]["env"] into spec.config["spec.env"] on deploy, and
+    # without removing the stale PYTHONPATH from base_spec a redeploy
+    # accumulates a duplicate entry alongside the freshly-set one.
+    base_spec = function.spec.base_spec or {}
+    base_env = mlrun.utils.get_in(base_spec, "spec.env", []) or []
+    if base_env:
+        mlrun.utils.update_in(
+            base_spec,
+            "spec.env",
+            [e for e in base_env if e.get("name") != "PYTHONPATH"],
```
Yacouby, line 1379: This removes every `PYTHONPATH` entry from `base_spec`, not just the mlrun-managed one. 
so if a user set `PYTHONPATH` only in `base_spec` (e.g. via `function.spec.base_spec` directly), it would be silently dropped on every redeploy.

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -798,30 +843,314 @@ def _should_fetch_source_code(
...
+        # If the previously-baked stub is loader-shaped, drop it so Nuclio
+        # rebuilds with the wrapper wiring.
+        if function.spec.build.functionSourceCode:
+            try:
+                decoded = base64.b64decode(
+                    function.spec.build.functionSourceCode
+                ).decode("utf-8")
+            except Exception:
+                decoded = ""
+            if "MLRUN_REAL_HANDLER" in decoded:
```
Yacouby: why not using also here the `stub_version_marker` ?

Yacouby (changes requested):
Looks good overall, some comments inline.
Also, please go over the whole changes and trim the multi-line comments as most of them explain what the code does rather than why, which is already clear from the code itself. keep only the ones that capture a non-obvious constraint or a specific design decision a future reader wouldn't derive on their own.

tests/system/runtimes/test_nuclio_store_uri.py
alxtkr77: This is kind="job", not nuclio. mlrun jobs use dot form (module.function);

server/py/services/api/crud/runtimes/nuclio/function.py
alxtkr77: Refactored

server/py/services/api/crud/runtimes/nuclio/function.py
alxtkr77: Fixed

mlrun/common/constants.py
```diff
@@ -23,6 +23,19 @@
...
+# loaded into the source-loader init container's output directory.
+STORE_URI_HANDLER_LOADER_MODULE = "_mlrun_store_uri_loader"
+# BUMP THIS WHENEVER THE LOADER STUB BODY CHANGES.
+# The version marker is baked as the first line of the loader stub
+# (`# stub_version=<N>`); on (re)deploy, an existing stub whose marker
+# doesn't match the current value is re-baked, so a stub-body fix in
+# mlrun reaches existing functions on their next deploy. If you forget
+# to bump this when changing the stub, deployed functions stay on the
+# old bytes forever.
+STORE_URI_LOADER_STUB_VERSION = "1"
```
Yacouby: shorten the comments here

Yacouby: if they are needed at all

server/py/services/api/crud/runtimes/nuclio/function.py
```diff
@@ -802,30 +796,323 @@ def _should_fetch_source_code(
...
+            client_version=client_version,
+            client_python_version=client_python_version,
+        )
+        return
+
+    raise mlrun.errors.MLRunInvalidArgumentError(
+        f"Function '{function.metadata.name}' (kind={function.kind!r}) has "
+        f"source={source!r} with load_source_on_run=True, which is supported "
+        "on Application kind only. Either set kind='application' or use a "
+        "store:// CodeArtifact source."
```
Yacouby: the log here should be fixed now, as it says "Application kind only"

Yacouby (changes requested):
minor

mlrun/common/constants.py
alxtkr77: ok

server/py/services/api/crud/runtimes/nuclio/function.py
alxtkr77: ok

Yacouby (approved):

PR #9651, opened 2026-05-05
[Runtimes] Make job + module:func handlers work for runtime-loaded source

alxtkr77:
## 📝 Description

PR #9609 (ML-11980 / pr-b) shipped `store://` CodeArtifact support for jobs but the canonical `module:function` handler form was broken end-to-end whenever the source had to be loaded at runtime (`store://`, `git://`, archive). Three separate mlrun code paths each accepted `.` as a module-prefix separator but missed `:`. Cluster validation on vmdev76 surfaced a fourth site in `mlrun/__main__.py` that the launcher-side fix doesn't cover.

This PR also fills the system-test coverage gap pr-b shipped without — `kind=job` + `store://` and `kind=application` + `store://` end-to-end flows now have dedicated tests.

## 🛠️ Changes Made

**Code fixes (4 line-level + 1 small refactor)**

- `mlrun/launcher/local.py` — kind selector in `ClientLocalLauncher._create_local_function_for_execution` accepts `:` (covers `function.run(local=True)` path)
- `mlrun/__main__.py` — parallel kind selector in the `mlrun run` CLI accepts `:` (covers the K8s job pod path: `mlrun run --from-env --handler mod:func --source store://...`)
- `mlrun/runtimes/local.py` — `LocalRuntime._pre_run` points `spec.command` at the extracted `<module>.py` and strips the module prefix from `runobj.spec.handler` when source is loaded at runtime
- `mlrun/launcher/base.py` — auto-name separator list adds `:` so `run.metadata.name` stays DNS-1123 valid for `module:func` handlers
- `mlrun/utils/helpers.py` + `mlrun/run.py` — extracted `split_handler_module_and_function` helper, reused from `code_to_function` and the new `_pre_run` logic

**Tests**

- 3 new parametrized unit tests in `tests/utils/test_helpers.py` and `tests/launcher/test_local.py` (helper round-trip, kind-selector branch, run-name stripping)
- 2 new system tests: `tests/system/runtimes/test_jobs_store_uri.py` and `tests/system/runtimes/test_applications_store_uri.py` — both passed on vmdev76
- System tests default image to `mlrun/mlrun` (matching `test_archives.py` / `test_kubejob.py`); `MLRUN_TEST_IMAGE` env var overrides

## ✅ Checklist

- [x] Format/lint clean (`ruff format`, `ruff check`)
- [x] Unit tests pass locally (557 in launcher / runtimes / utils sweep)
- [x] System tests pass on vmdev76 (job + application)
- [x] No public API docs needed updating (handler form already documented; helper is internal)
- [x] No secrets in diff
- [x] Independent of pr-c (ML-12480 Nuclio init container) — neither test uses pr-c-only symbols

## 🧪 Testing

**Unit (local)**
```
pytest tests/utils/test_helpers.py tests/launcher/ \
       tests/runtimes/test_local.py tests/runtimes/test_function.py
# 557 passed
```

**System (vmdev76)**
```
pytest tests/system/runtimes/test_jobs_store_uri.py \
       tests/system/runtimes/test_applications_store_uri.py
# 2 passed in 4:17
```

## 🔗 References

- Jira: [ML-11980](https://iguazio.atlassian.net/browse/ML-11980) (parent feature: load functions from artifact)
- Jira: [ML-12479](https://iguazio.atlassian.net/browse/ML-12479) (pr-b deficiency)
- Builds on: #9609 (pr-b — already merged)

## 🚨 Breaking Changes

**No.** All changes are gated behind conditions that don't trigger for existing flows:
- New colon-handler paths only activate when handler contains `:`
- `_pre_run` command-pointing only activates when `spec.command` is empty AND module name is set AND the file exists in `target_dir`
- Refactored `split(":")` in `code_to_function` produces identical results for all bare and single-colon inputs (multi-colon `"a:b:c"` now yields `"b:c"` instead of `"b"` — pinned by the new helper test; this form is exotic and undocumented)

[ML-11980]: https://iguazio.atlassian.net/browse/ML-11980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mlrun/runtimes/local.py
+++ b/mlrun/runtimes/local.py
@@ -285,6 +285,20 @@ class LocalRuntime(BaseRuntime, ParallelRunner):
                 if os.path.isfile(candidate):
                     self.spec.command = candidate
                     runobj.spec.handler = func_name
+                else:
+                    # Surface the missing-file case with a useful breadcrumb
+                    # — without this log, a typo in the handler module name
+                    # falls through to a cryptic ImportError downstream when
+                    # _get_handler tries to load an empty spec.command.
+                    logger.warning(
+                        "module:func handler refers to a module that wasn't "
+                        "found in the extracted source directory; the run "
+                        "will likely fail to load the handler",
+                        handler=runobj.spec.handler or self.spec.default_handler,
+                        module_name=module_name,
+                        candidate=candidate,
+                        target_dir=target_dir,
+                    )
 
         if execution._current_workdir:
             execution._old_workdir = os.getcwd()
--- a/mlrun/runtimes/local.py
+++ b/mlrun/runtimes/local.py
@@ -39,6 +39,7 @@ from nuclio import Event
 
 import mlrun
 import mlrun.common.constants as mlrun_constants
+import mlrun.utils.helpers
 from mlrun.lists import RunList
 from mlrun.package import handler as mlrun_handler_decorator
 
@@ -265,6 +266,25 @@ class LocalRuntime(BaseRuntime, ParallelRunner):
                 execution._current_workdir = os.path.join(target_dir, workdir)
             else:
                 execution._current_workdir = workdir or target_dir
+
+            # When the source was extracted (e.g. from a store:// CodeArtifact)
+            # rather than baked into the function image, spec.command is empty -
+            # _get_handler / load_module would then have no file to import, and
+            # a `module:func` handler can't be resolved. Mirror the convention
+            # used by mlrun.code_to_function: spec.command holds the file,
+            # spec.handler holds just the function name; load_module loads the
+            # file, _search_in_namespaces finds the function in the loaded
+            # module.
+            module_name, func_name = (
+                mlrun.utils.helpers.split_handler_module_and_function(
+                    runobj.spec.handler or self.spec.default_handler or ""
+                )
+            )
+            if not self.spec.command and module_name:
+                candidate = os.path.join(target_dir, f"{module_name}.py")
+                if os.path.isfile(candidate):
+                    self.spec.command = candidate
+                    runobj.spec.handler = func_name
 
         if execution._current_workdir:
             execution._old_workdir = os.getcwd()
--- a/mlrun/runtimes/local.py
+++ b/mlrun/runtimes/local.py
@@ -267,14 +267,9 @@ class LocalRuntime(BaseRuntime, ParallelRunner):
             else:
                 execution._current_workdir = workdir or target_dir
 
-            # When the source was extracted (e.g. from a store:// CodeArtifact)
-            # rather than baked into the function image, spec.command is empty -
-            # _get_handler / load_module would then have no file to import, and
-            # a `module:func` handler can't be resolved. Mirror the convention
-            # used by mlrun.code_to_function: spec.command holds the file,
-            # spec.handler holds just the function name; load_module loads the
-            # file, _search_in_namespaces finds the function in the loaded
-            # module.
+            # Source extracted at runtime (store:// CodeArtifact, git, archive)
+            # leaves spec.command empty, so a "module:func" handler can't resolve.
+            # Mirror code_to_function: command -> file, handler -> bare function.
             module_name, func_name = (
                 mlrun.utils.helpers.split_handler_module_and_function(
                     runobj.spec.handler or self.spec.default_handler or ""
@@ -286,10 +281,8 @@ class LocalRuntime(BaseRuntime, ParallelRunner):
                     self.spec.command = candidate
                     runobj.spec.handler = func_name
                 else:
-                    # Surface the missing-file case with a useful breadcrumb
-                    # — without this log, a typo in the handler module name
-                    # falls through to a cryptic ImportError downstream when
-                    # _get_handler tries to load an empty spec.command.
+                    # Warn so a handler-module typo doesn't end in a cryptic
+                    # ImportError downstream from an empty spec.command.
                     logger.warning(
                         "module:func handler refers to a module that wasn't "
                         "found in the extracted source directory; the run "
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -1485,6 +1485,28 @@ def get_class(class_name, namespace=None):
     return class_object
 
 
+def split_handler_module_and_function(handler: str) -> tuple[str, str]:
+    """Split mlrun's ``"module:function"`` handler form into ``(module, function)``.
+
+    mlrun's canonical handler format on the wire is ``"<module>:<function>"`` —
+    e.g. ``"trainer:train_model"``. Internally on the runtime spec, however,
+    ``spec.handler`` typically holds JUST the function name (the module is
+    implicit in ``spec.command`` or in the loaded source file). This helper
+    bridges the two: callers split once and decide what to do with each part.
+
+    Bare function names (no colon) are returned with an empty module:
+    ``"my_func"`` -> ``("", "my_func")``.
+
+    :param handler: Handler string in either ``"module:function"`` or bare
+                    ``"function"`` form. Falsy values return ``("", "")``.
+    :returns: ``(module_name, function_name)``.
+    """
+    if not handler or ":" not in handler:
+        return "", handler or ""
+    module, _, function = handler.partition(":")
+    return module, function
+
+
 def get_function(function, namespaces, reload_modules: bool = False):
     """Return function callable object from function name string
 
--- a/mlrun/utils/helpers.py
+++ b/mlrun/utils/helpers.py
@@ -1485,7 +1485,7 @@ def get_class(class_name, namespace=None):
     return class_object
 
 
-def split_handler_module_and_function(handler: str) -> tuple[str, str]:
+def split_handler_module_and_function(handler: str | None) -> tuple[str, str]:
     """Split mlrun's ``"module:function"`` handler form into ``(module, function)``.
 
     mlrun's canonical handler format on the wire is ``"<module>:<function>"`` —
--- a/tests/launcher/test_local.py
+++ b/tests/launcher/test_local.py
@@ -107,8 +107,10 @@ def test_create_local_function_for_execution_with_enrichment():
 @pytest.mark.parametrize(
     "module_qualified_handler",
     [
-        "module.submodule.fn",  # dotted form (already works)
-        "handler:my_func",  # canonical mlrun module:func form (the new case)
+        # dotted form — already works
+        "module.submodule.fn",
+        # canonical mlrun module:func form — the new case
+        "handler:my_func",
     ],
 )
 def test_create_local_function_for_execution_picks_local_for_module_qualified_handler(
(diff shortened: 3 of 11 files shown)


tests/launcher/test_local.py
```diff
@@ -323,3 +360,34 @@ def test_validate_run_retries_invalid_runtime():
...
+)
+def test_enrich_run_name_strips_handler_module_prefix(handler_value, expected_suffix):
+    """The auto-generated run.metadata.name must strip the module prefix from
+    every handler form mlrun accepts. Critically, the result must be a valid
+    DNS-1123 name - no `:` allowed - or the run will be rejected by K8s.
+    Order matters: the `::` separator (class-method) must split BEFORE `:`
+    so MyClass::run_method becomes 'run_method', not 'method'."""
+    launcher = mlrun.launcher.local.ClientLocalLauncher(local=False)
+    runtime = mlrun.code_to_function(
+        name="my-fn", kind="job", filename=str(func_path), handler=handler
```
Yacouby: shouldnt be `handler=handler_value` ?
it is now using other handler thats not what it claims to test

mlrun/runtimes/local.py
```diff
@@ -266,6 +267,25 @@ def _pre_run(self, runobj: RunObject, execution: MLClientCtx):
...
+            module_name, func_name = (
+                mlrun.utils.helpers.split_handler_module_and_function(
+                    runobj.spec.handler or self.spec.default_handler or ""
+                )
+            )
+            if not self.spec.command and module_name:
+                candidate = os.path.join(target_dir, f"{module_name}.py")
+                if os.path.isfile(candidate):
+                    self.spec.command = candidate
+                    runobj.spec.handler = func_name
```
Yacouby, line 282: If `module:func` is given but `module.py` doesn't exist in `target_dir`, this silently skips, and execution fails later with a  import/load error. Consider adding a `logger.warning` so the user gets a useful message pointing at the missing file.

mlrun/runtimes/local.py
```diff
@@ -266,6 +267,25 @@ def _pre_run(self, runobj: RunObject, execution: MLClientCtx):
...
+            # module.
+            module_name, func_name = (
+                mlrun.utils.helpers.split_handler_module_and_function(
+                    runobj.spec.handler or self.spec.default_handler or ""
+                )
+            )
+            if not self.spec.command and module_name:
+                candidate = os.path.join(target_dir, f"{module_name}.py")
+                if os.path.isfile(candidate):
+                    self.spec.command = candidate
```
Yacouby, line 281: This block is the heart of the fix but has no unit test coverage.

tests/system/runtimes/test_jobs_store_uri.py
```diff
@@ -0,0 +1,84 @@
...
+
+    project_name = "job-store-uri-system-test"
+
+    def custom_setup(self):
+        super().custom_setup()
+        self._handler_filename = "handler.py"
+        self._function_image = os.environ.get("MLRUN_TEST_IMAGE", "mlrun/mlrun")
+
+    def _log_code_artifact(self, key: str) -> str:
+        """Log the assets/handler.py file as a CodeArtifact and return its
```
Yacouby, line 40: just making sure, this file already exists in the assets ?

tests/launcher/test_local.py
```diff
@@ -104,6 +104,43 @@ def test_create_local_function_for_execution_with_enrichment():
     assert runtime.spec.allow_empty_resources
 
 
+@pytest.mark.parametrize(
+    "module_qualified_handler",
+    [
+        "module.submodule.fn",  # dotted form (already works)
+        "handler:my_func",  # canonical mlrun module:func form (the new case)
```
Yacouby: move comments to be above the code

tests/utils/test_helpers.py
```diff
@@ -2346,3 +2346,21 @@ def test_remove_image_protocol_prefix(image, expected):
...
+@pytest.mark.parametrize(
+    "handler,expected",
+    [
+        ("trainer:train_model", ("trainer", "train_model")),
+        ("my_func", ("", "my_func")),
+        ("", ("", "")),
+        (None, ("", "")),
+        # leading/trailing whitespace is preserved by partition (caller's
+        # responsibility) — this is just documenting current behavior:
+        ("a:b:c", ("a", "b:c")),  # only the FIRST colon splits
```
Yacouby: move comment to above the code line

tests/system/runtimes/test_applications_store_uri.py
```diff
@@ -0,0 +1,112 @@
...
+        super().custom_setup()
+        self._app_filename = "simple_flask_app.py"
+        self._function_image = os.environ.get("MLRUN_TEST_IMAGE", "mlrun/mlrun")
+
+    def _log_code_artifact(self, key: str) -> str:
+        """Log the assets/simple_flask_app.py file as a CodeArtifact and
+        return its store:// URI."""
+        local_path = os.path.join(self.assets_path, self._app_filename)
+        self.project.log_code_file(key=key, local_path=local_path)
+        return f"store://artifacts/{self.project_name}/{key}"
```
Yacouby: `log_code_file` returns the artifact object. Consider using `artifact.uri` instead of constructing the URI manually — this ensures the test uses the exact URI the system stored rather than relying on an implicit format that could change or include a tag suffix.

tests/system/runtimes/test_jobs_store_uri.py
```diff
@@ -0,0 +1,84 @@
...
+        store:// URI."""
+        local_path = os.path.join(self.assets_path, self._handler_filename)
+        self.project.log_code_file(key=key, local_path=local_path)
+        return f"store://artifacts/{self.project_name}/{key}"
```
Yacouby: Same comment as in https://github.com/mlrun/mlrun/pull/9651/changes#r3190512743

Yacouby (changes requested):
Some comments inline

tests/launcher/test_local.py
alxtkr77: Removed `handler=handler` from `code_to_function`; the runtime now has no default handler so `handler_value` is the only handler input flowing through `_enrich_run`.

mlrun/runtimes/local.py
alxtkr77: Added `logger.warning` in the `else` branch with handler / module_name / candidate / target_dir, so a missing module surfaces a useful breadcrumb instead of falling through to a cryptic `ModuleNotFoundError` downstream. Kept it as a warning rather than a raise to preserve namespace-resolution fall-through (e.g. dotted modules like `pkg.mod:func` where the file is at `target_dir/pkg/mod.py` and `_search_in_namespaces` resolves it).

mlrun/runtimes/local.py
alxtkr77: Added two tests in `tests/runtimes/test_local.py`: `test_pre_run_points_command_at_extracted_module` (happy path — pins `spec.command` lands at the extracted file and the handler prefix is stripped) and `test_pre_run_warns_when_module_not_in_extracted_dir` (pins the warning message + that `spec.command` stays unset on miss). Both monkeypatch `extract_source` to control the target dir.

tests/system/runtimes/test_jobs_store_uri.py
alxtkr77: Yes — `tests/system/runtimes/assets/handler.py` is the existing handler used by sibling system tests; `my_func` produces deterministic outputs we assert on (`accuracy = p1 * 2`). Expanded the `_log_code_artifact` docstring to make the asset reuse explicit.

tests/launcher/test_local.py
alxtkr77: Done — parametrize entry comments moved above the values.

tests/utils/test_helpers.py
alxtkr77: Done — moved above. Also rewrote the comment text since the original mentioned whitespace preservation but the parametrize doesn't include any whitespace cases; the actual point is `partition()` semantics vs `split()`.

tests/system/runtimes/test_applications_store_uri.py
alxtkr77: Done — `_log_code_artifact` now captures the returned `CodeArtifact` and returns `artifact.uri`. Docstring updated.

tests/system/runtimes/test_jobs_store_uri.py
alxtkr77: Same fix applied here — `_log_code_artifact` returns `artifact.uri`.

mlrun/utils/helpers.py
```diff
@@ -1485,6 +1485,28 @@ def get_class(class_name, namespace=None):
     return class_object
 
 
+def split_handler_module_and_function(handler: str) -> tuple[str, str]:
```
Yacouby: should be: `str | None`  ?

mlrun/runtimes/local.py
```diff
@@ -266,6 +267,39 @@ def _pre_run(self, runobj: RunObject, execution: MLClientCtx):
...
                 execution._current_workdir = workdir or target_dir
 
+            # When the source was extracted (e.g. from a store:// CodeArtifact)
+            # rather than baked into the function image, spec.command is empty -
+            # _get_handler / load_module would then have no file to import, and
+            # a `module:func` handler can't be resolved. Mirror the convention
+            # used by mlrun.code_to_function: spec.command holds the file,
+            # spec.handler holds just the function name; load_module loads the
+            # file, _search_in_namespaces finds the function in the loaded
+            # module.
```
Yacouby: long comment, can be shorter for a faster read

tests/runtimes/test_local.py
```diff
@@ -36,3 +36,78 @@ def test_run_exec_verbose_stderr(return_code):
...
+    """`LocalRuntime._pre_run` points spec.command at the extracted module
+    file and strips the module prefix from the run handler when the source
+    is loaded at runtime (e.g. store:// CodeArtifact, git, archive).
+
+    Without this, _get_handler would call load_module(file_name="",
+    handler="module:func", ...), produce no module, and fail downstream
+    with ModuleNotFoundError.
+    """
+    import mlrun
+    import mlrun.runtimes.local
```
Yacouby: These should be at the top of the file, not repeated inside each test.

tests/runtimes/test_local.py
```diff
@@ -36,3 +36,78 @@ def test_run_exec_verbose_stderr(return_code):
...
+def test_pre_run_warns_when_module_not_in_extracted_dir(tmp_path, monkeypatch, caplog):
+    """When `module:func` is given but `<module>.py` doesn't exist in the
+    extracted source dir, _pre_run logs a warning instead of silently
+    falling through. Pinning this prevents the regression where a handler
+    typo produces a cryptic ImportError far from the entry point.
+    """
+    import logging
+
+    import mlrun
+    import mlrun.runtimes.local
```
Yacouby: These should be at the top of the file, not repeated inside each test.

tests/runtimes/test_local.py
```diff
@@ -36,3 +36,78 @@ def test_run_exec_verbose_stderr(return_code):
...
+        lambda *args, **kwargs: str(target_dir),
+    )
+
+    runtime = mlrun.runtimes.local.LocalRuntime()
+    runtime.spec.build.source = "store://artifacts/proj/missing_code"
+    run = mlrun.run.RunObject()
+    run.spec.handler = "missing_module:my_func"
+    execution = mlrun.run.MLClientCtx.from_dict(run.to_dict(), autocommit=False)
+
+    with caplog.at_level(logging.WARNING, logger="mlrun"):
```
Yacouby: MLRun uses structlog, not stdlib logging, so caplog may not capture the warning if structlog isn't propagating to stdlib logging in the test environment. Consider monkeypatching logger.warning directly and asserting it was called instead.

Yacouby (commented):
Looks good, minor suggestions

Yacouby (approved):

PR #9655, opened 2026-05-05
[Projects] Support store:// CodeArtifact URIs in set_workflow

alxtkr77:
### 📝 Description

Phase 2 of ML-11980/11981 (after the Phase 1 function-side `store://` PRs):
enable `project.set_workflow(workflow_path="store://artifacts/<proj>/<key>")`
end-to-end. The runner pod (or any caller of `WorkflowSpec.get_source_file`)
resolves the URI via the existing `mlrun.utils.clones.load_source_code`
helper and validates `artifact.spec.code_type == "workflow"`.

---

### 🛠️ Changes Made

**Client side (`mlrun/projects/project.py`):**

- `_validate_file_path`: relax the suffix-required rejection for
  `store://` URIs only. Other remote URLs (`s3://`, `gs://`, etc.) still
  require a file suffix.
- `set_workflow`: when `workflow_path` is `store://` and the artifact is
  reachable client-side, validate `code_type == "workflow"`; raise
  `MLRunInvalidArgumentError` on mismatch. If unreachable (offline,
  missing creds), defer to the runner pod's authoritative check.

**Runner pod side (`mlrun/projects/pipelines.py`):**

- `WorkflowSpec.get_source_file`: gains optional `project_name` and
  `secrets` parameters + a `store://` branch that downloads via
  `load_source_code` into `<context>/.mlrun/artifacts/` and re-validates
  `code_type`.
- `_PipelineRunner._get_handler` and `_KFPRunner.save`: thread
  `project.metadata.name` and project secrets through to
  `get_source_file` so `engine='kfp'` compile-side resolution works for
  credential-protected stores.
- `_RemoteRunner.run`: extract the existing project-local-path
  relativization into a static helper `relativize_workflow_path` and
  skip it for any `://` URL so the runner pod sees URLs verbatim.

**Helper (`mlrun/utils/clones.py`):**

- `load_source_code` returns `(target_dir, file_path | None)`. `file_path`
  is set for `store://` (single-file CodeArtifacts) and `None` for
  git/archive sources where there's no canonical entry file.
  `extract_source` and the `mlrun load-source` CLI take `target_dir`;
  `WorkflowSpec.get_source_file` takes `file_path`.

---

### ✅ Checklist

- [x] I updated the documentation (if applicable) — N/A, internal SDK code path
---

### 🧪 Testing

**System test** (`tests/system/projects/test_workflow_store_uri.py` —
new file):

- `TestWorkflowStoreUri::test_run_workflow_from_store_artifact`:
  registers a workflow CodeArtifact, `set_workflow` with the `store://`
  URI, runs via `engine="kfp"`, asserts state == `Succeeded`. Verified
  end-to-end on vmdev76 in ~70 s.

---

### 🔗 References

- Ticket link: ML-11981

---

### 🚨 Breaking Changes?

- [ ] Yes (explain below)
- [x] No

---

### 🔍️ Additional Notes

The workflow runner pod is a regular KubeJob whose main container
auto-receives all project secrets as env vars (via
`BaseRuntimeHandler._add_k8s_secrets_to_spec`). So
`load_source_code(secrets=None)` works for any project-secret-protected
backend (S3, GCS, Azure) without any explicit secret threading — unlike
the Phase 1 Application init container path, which required an explicit
`envFrom` mount because init containers don't inherit main-container env.
Compile-side callers (`engine='kfp'`) still thread `project._secrets`
through, so users running pipelines on their local machine with
credential-protected stores work without setting env vars manually.


mlrun/projects/pipelines.py
```diff
@@ -114,16 +145,54 @@ def get_source_file(self, context=""):
...
+                    f"{mlrun.artifacts.CodeArtifactCodeType.workflow.value!r}."
+                )
+            target_dir = os.path.join(context or ".", ".mlrun", "artifacts")
+            os.makedirs(target_dir, exist_ok=True)
+            _, file_path = mlrun.utils.clones.load_source_code(
+                source_uri=self.path,
+                target_dir=target_dir,
+                project=project_name,
+                secrets=secrets,
+            )
```
liranbg: to reduce the complexity of this function, I think it worth movig it to an internal function and call it. this will also help unit-testing it.

mlrun/projects/pipelines.py
```diff
@@ -845,6 +922,33 @@ class _RemoteRunner(_PipelineRunner):
 
     engine = "remote"
 
+    @staticmethod
+    def relativize_workflow_path(workflow_path: str, code_path: str) -> str:
+        """Rewrite a project-local workflow path to ``./relative`` form.
```
liranbg: what does it rewrite? this is a bit misleading. it just resolve a workflow relative path. I think also method name can be changed

mlrun/projects/project.py
```diff
@@ -1422,6 +1422,74 @@ def set_workflow(
...
+                code_type = workflow_artifact.spec.code_type
+                if (
+                    code_type is not None
+                    and code_type != mlrun.artifacts.CodeArtifactCodeType.workflow
+                ):
+                    raise mlrun.errors.MLRunInvalidArgumentError(
+                        f"Workflow path {workflow_path!r} resolves to a code "
+                        f"artifact with code_type={code_type!r}; expected "
+                        f"{mlrun.artifacts.CodeArtifactCodeType.workflow.value!r}."
+                    )
```
liranbg: also here, can be moved to a function handling the work. keeping current function more coherent

mlrun/projects/project.py
```diff
@@ -1422,6 +1422,74 @@ def set_workflow(
...
+                mlrun.errors.MLRunNotFoundError,
+                mlrun.errors.MLRunBadRequestError,
+                mlrun.errors.MLRunUnauthorizedError,
+                mlrun.errors.MLRunAccessDeniedError,
+                mlrun.errors.MLRunPreconditionFailedError,
+            ):
+                # The DB returned a definitive "no, that's wrong" answer
+                # (typo'd key, malformed URI, perms). Propagate so the user
+                # sees it now, not after a 5-10 minute runner-pod round trip.
+                raise
```
liranbg: error prone to any missing 4xx issues - not sure why you decided to go with "cherry-picking" exceptions and not pick on only those you can safely rely that is "transient"

mlrun/projects/project.py
```diff
@@ -1422,6 +1422,74 @@ def set_workflow(
...
+                    exc=mlrun.errors.err_to_str(exc),
+                )
+                workflow_artifact = None
+            if workflow_artifact is not None:
+                if workflow_artifact.kind != "code":
+                    raise mlrun.errors.MLRunInvalidArgumentError(
+                        f"Workflow path {workflow_path!r} resolves to a "
+                        f"{workflow_artifact.kind!r} artifact; expected a "
+                        "code artifact (kind='code')."
+                    )
```
liranbg: can workflow kind be a constant?

liranbg (changes requested):

mlrun/projects/pipelines.py
```diff
@@ -845,6 +986,38 @@ class _RemoteRunner(_PipelineRunner):
 
     engine = "remote"
 
+    @staticmethod
+    def make_workflow_path_relative(workflow_path: str, code_path: str) -> str:
```
liranbg:
```suggestion
    def resolve_relative_workflow_path(workflow_path: str, code_path: str) -> str:
```

I think is a better name + stands better with verb-convention of "get/resolve"

tests/system/projects/test_workflow_store_uri.py
```diff
@@ -0,0 +1,147 @@
...
+        """Log the workflow source as a CodeArtifact and return its store:// URI."""
+        workflow_src = self._write_pipeline_to_tempfile()
+        artifact = self.project.log_code_file(
+            key=artifact_key,
+            local_path=workflow_src,
+            code_type="workflow",
+        )
+        return artifact.uri
+
+    def test_run_workflow_from_store_artifact_kfp_engine(self):
```
liranbg: look like both function eventually do the same, so could be a good option for parameterized

in addition, why not putting it under exstising system/runtimes/test_kfp ?

mlrun/projects/pipelines.py
```diff
@@ -151,6 +181,98 @@ def clear_tmp(self):
...
+    :raises MLRunNotFoundError: if ``artifact`` is ``None``.
+    :raises MLRunInvalidArgumentError: if kind or code_type mismatch.
+    """
+    if artifact is None:
+        raise mlrun.errors.MLRunNotFoundError(
+            f"Workflow path {workflow_path!r} did not resolve to an artifact "
+            "(the store returned no resource; the URI may be malformed or "
+            "point at a missing link target)."
+        )
+    if artifact.kind != mlrun.artifacts.CodeArtifact.kind:
```
theSaarco, line 212: You should first validate that `artifact` itself is of type `Artifact`. The `get_store_resource` function may return artifacts, but also can return `FeatureSet` or `FeatureVector`. The check you're doing here will probably work in that case, since these objects also have a `kind` that is different, but logically you're assuming that you already have an artifact which is not guaranteed.

mlrun/projects/pipelines.py
```diff
@@ -151,6 +181,98 @@ def clear_tmp(self):
...
+            f"Workflow path {workflow_path!r} resolves to a code artifact "
+            f"with code_type={code_type.value!r}; expected "
+            f"{mlrun.artifacts.CodeArtifactCodeType.workflow.value!r}."
+        )
+
+
+def _try_validate_remote_workflow_artifact(
+    workflow_path: str, project_name: str
+) -> None:
+    """Client-side pre-check for a ``store://`` workflow path.
```
theSaarco: Not sure I understand the use-case where this is needed. The `get_store_resource` method doesn't download the actual object, it just gets the artifact metadata from the DB. This is expected to work properly in any client-side setup with connectivity to MLRun backend.

mlrun/projects/project.py
```diff
@@ -1429,6 +1446,21 @@ def set_workflow(
...
+        if embed and workflow_path and "://" in workflow_path:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"embed=True is not supported for remote workflow_path "
+                f"({workflow_path!r}); use embed=False (the default) so the "
+                "workflow source is fetched at execution time."
+            )
+
+        if workflow_path and mlrun.datastore.is_store_uri(workflow_path):
+            # Surface kind/code_type misuses now; the runner pod re-validates.
+            mlrun.projects.pipelines._try_validate_remote_workflow_artifact(
```
theSaarco: See my comment in the implementation - why should this be tolerant to not being able to download the artifact?

theSaarco (changes requested):
Looks good - some minor fixes needed.

mlrun/projects/pipelines.py
alxtkr77: ok

tests/system/projects/test_workflow_store_uri.py
alxtkr77: Parametrized  - the two methods collapsed into `test_run_workflow_from_store_artifact` with `engine` as the parameter.
On relocation: the 7 tests in `system/runtimes/test_kfp.py` all test KFP runtime quirks (mount, retry, termination, step-name truncation, ...). Our test is about `set_workflow` + workflow CodeArtifact URI handling - it happens to use `engine='kfp'` as the simplest driver, but the thing under test is project-level.

mlrun/projects/pipelines.py
alxtkr77: ok

mlrun/projects/pipelines.py
alxtkr77: I overcomplicated things, right

mlrun/projects/project.py
alxtkr77: same here

mlrun/projects/pipelines.py
```diff
@@ -151,6 +181,89 @@ def clear_tmp(self):
             os.remove(self._tmp_path)
 
 
+# Project-context subdir where store:// workflow downloads land.
+_WORKFLOW_ARTIFACTS_SUBDIR = os.path.join(".mlrun", "artifacts")
```
liranbg: why is this a constant

mlrun/projects/pipelines.py
```diff
@@ -151,6 +181,89 @@ def clear_tmp(self):
...
+
+
+def _validate_workflow_code_artifact(artifact, workflow_path: str) -> None:
```
liranbg, line 186: why these functions are not part of the class (methods)?

alxtkr77: Just followed module level pattern - many helper functions are defined as stand alone functions in this file

mlrun/projects/pipelines.py
alxtkr77: It represents the directory we use to download the code to. Important enough to be a constant. Consolidated it with project '.mlrun/code' directory and made it a global constant. It actually represents the same thing: the folder we have the code downloaded

mlrun/projects/project.py
```diff
@@ -1429,6 +1447,21 @@ def set_workflow(
...
+        if embed and workflow_path and "://" in workflow_path:
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"embed=True is not supported for remote workflow_path "
+                f"({workflow_path!r}); use embed=False (the default) so the "
+                "workflow source is fetched at execution time."
+            )
+
+        if workflow_path and mlrun.datastore.is_store_uri(workflow_path):
+            # Surface kind/code_type misuses now; the runner pod re-validates.
+            mlrun.projects.pipelines._validate_remote_workflow_artifact(
```
theSaarco: Since you're calling this method externally to the module, don't name it as private.

mlrun/projects/pipelines.py
```diff
@@ -151,6 +183,85 @@ def clear_tmp(self):
...
+        and code_type != mlrun.artifacts.CodeArtifactCodeType.workflow
+    ):
+        raise mlrun.errors.MLRunInvalidArgumentError(
+            f"Workflow path {workflow_path!r} resolves to a code artifact "
+            f"with code_type={code_type.value!r}; expected "
+            f"{mlrun.artifacts.CodeArtifactCodeType.workflow.value!r}."
+        )
+
+
+def _validate_remote_workflow_artifact(workflow_path: str, project_name: str) -> None:
```
theSaarco:
```suggestion
def validate_remote_workflow_artifact(workflow_path: str, project_name: str) -> None:
```

Being called from another module, no reason to name it as "private".

mlrun/projects/pipelines.py
```diff
@@ -151,6 +183,85 @@ def clear_tmp(self):
...
+    """Resolve, validate, and download a ``store://`` workflow artifact.
+
+    Runner-pod authoritative path; all errors propagate.
+
+    :returns: Local filesystem path to the downloaded workflow file.
+    """
+    artifact = mlrun.datastore.get_store_resource(
+        workflow_path, project=project_name, secrets=secrets, data_store_secrets=secrets
+    )
+    _validate_workflow_code_artifact(artifact, workflow_path)
```
theSaarco: Simply call `validate_remote_workflow_artifact` instead (probably want to add `secrets` and `data_store_secrets` to its parameters as well).

theSaarco (changes requested):
Some cleaning-up suggestions.

mlrun/projects/pipelines.py
alxtkr77: ok

mlrun/projects/pipelines.py
alxtkr77: ok

mlrun/projects/project.py
alxtkr77: ok

theSaarco (approved):
LGTM

tests/system/projects/test_workflow_store_uri.py
```diff
@@ -0,0 +1,162 @@
...
+    def _setup_store_workflow(self, artifact_key: str) -> str:
+        """Log the workflow source as a CodeArtifact and return its store:// URI."""
+        workflow_src = self._write_pipeline_to_tempfile()
+        artifact = self.project.log_code_file(
+            key=artifact_key,
+            local_path=workflow_src,
+            code_type="workflow",
+        )
+        return artifact.uri
+
```
liranbg: move below public functions

tests/system/projects/test_workflow_store_uri.py
```diff
@@ -0,0 +1,162 @@
...
+_PIPELINE_BODY = (
+    "from kfp import dsl\n"
+    "import mlrun\n"
+    "\n"
+    "funcs = {}\n"
+    "\n"
+    '@dsl.pipeline(name="store-uri pipeline")\n'
+    "def kfpipeline():\n"
+    '    funcs["describe"].as_step(name="describe-step")\n'
+)
```
liranbg: constant can be removed, used once, can be inline to the function uses it

tests/system/projects/test_workflow_store_uri.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+@tests.system.base.TestMLRunSystem.skip_test_if_env_not_configured
+class TestWorkflowStoreUri(tests.system.base.TestMLRunSystem):
```
liranbg: it worth moving the test to existing (TestKFP).

liranbg (changes requested):

tests/system/projects/test_workflow_store_uri.py
liranbg: Our convention is a single test file per runtime. adjust the class name and what is needed for adhere the convention

alxtkr77: ok

tests/system/projects/test_workflow_store_uri.py
alxtkr77: ok

tests/system/projects/test_workflow_store_uri.py
alxtkr77: ok

liranbg (approved):

PR #9656, opened 2026-05-05
[Projects] Download store:// function and workflow code during zip export

alxtkr77:
### 📝 Description                                                                                                                                                                                         
                                                                                                                                                                                                             
  When `MlrunProject.export()` produces a zip archive, functions and workflows whose source is a `store://` CodeArtifact URI currently end up referenced verbatim in the embedded `project.yaml`. The zip is
  therefore not self-contained — it cannot be unpacked on a different cluster (or after the source DB is gone), because the `store://` URI is only meaningful relative to the cluster it was created against.
                                                                                                                                                                                                             
  This PR makes zip-export self-contained: when `archive_code=True`, each `store://` function/workflow source is resolved to its target, downloaded into a temporary staging directory, and the zip is built 
  with the bytes included plus `project.yaml` rewritten to reference local paths inside the zip (`.mlrun/code/<basename>_<sha256(target_path)[:16]><ext>`). YAML-only export keeps `store://` URIs verbatim —
   same-cluster round-trip already resolves them via pr-a/pr-b's `_init_function_from_dict` `store://` branch.                                                                                               
                                                                                                                                                                                                             
  ### 🛠️  Changes Made                    
                                                                                                                                                                                                             
  `mlrun/projects/project.py`:                                                                         
  - New module-level helper `_download_store_artifact_for_export(staging_dir, store_uri, project_name)` resolves the URI via `mlrun.datastore.get_store_resource`, downloads the bytes into the caller's
  staging directory, and returns `(real_path, zip_arcname)`. Returns `None` on recoverable failures (artifact missing, network/IO); raises `MLRunInvalidArgumentError` for non-`CodeArtifact` resolutions
  (e.g., user passed a model or feature-set URI as a function source).
  - New module-level helper `_stage_store_downloads_for_export(spec, staging_dir, project_name)` walks `_function_definitions` and `_workflows`, calls the download helper for each `store://` URI it finds,
  and returns deepcopied dicts with paths rewritten plus `(staged_downloads, failures)` bookkeeping. Pure — no side effects on `self.spec`.                                                                  
  - New module-level helper `_build_export_zip(filepath, project_dir, files_filter, staged_downloads)` owns the `tempfile.NamedTemporaryFile` + `zipfile.ZipFile` + remote-upload step.
  - `MlrunProject.export()` itself shrinks to ~70 LOC: a `contextlib.ExitStack` manages the staging tmpdir, the deepcopy/swap pattern keeps the live spec invariant across the YAML write, and any download  
  failures abort the export with `MLRunRuntimeError` (fail-fast — a partial zip silently shipping `store://` URIs would defeat the purpose of a self-contained archive).                                     
                                          
  `tests/projects/test_project.py`:                                                                                                                                                                          
  - 5 mock-free integration tests using the existing `rundb_mock` fixture, all `tmp_path`-isolated:                                                                                                          
    - `test_export_to_zip_downloads_store_function_code`                                                                                                                                                     
    - `test_export_to_yaml_keeps_store_uri`                                                                                                                                                                  
    - `test_export_to_zip_raises_on_unresolvable_store_uri`                                            
    - `test_export_to_zip_raises_on_download_failure`                                                                                                                                                        
    - `test_export_to_zip_downloads_store_workflow_code`                                                                                                                                                     
                                                                                                                                                                                                             
  ### ✅ Checklist                                                                                                                                                                                           
                                                                                                       
  - [x] I updated the documentation (if applicable) — N/A, internal SDK path                                                                                                                                 
    - [x] If yes, I ran all relevant system tests and ensured they passed before submitting this PR — N/A; covered by mock-free integration tests against real `mlrun.datastore` resolution                  
    - [x] I updated existing system tests and/or added new ones if needed to cover my changes — see above
                                                                                                                                                                                                             
  ### 🧪 Testing                                                                                       
                                                                                                                                                                                                             
  - 5 mock-free integration tests exercise the real `mlrun.datastore` `store:// → file://` resolution path via the in-memory `rundb_mock` fixture (no patching of the helper itself).                        
  - Why no system tests: the export-side logic is pure SDK; the load-side is unchanged from pr-a/b/c (`_init_function_from_dict`'s `store://` branch). A real-cluster round-trip would mostly verify         
  v3io/cluster infrastructure that this PR does not touch.                                                                                                                                                   
                                                                                                                                                                                                             
  ### 🔗 References                      
                                                                                                                                                                                                             
  - Ticket link: ML-11980                                                                              
                                                                                                                                                                                                             
  ### 🚨 Breaking Changes?                                                                             
                                          
  - [ ] Yes                                   
  - [x] No

### DIFF
--- a/mlrun/projects/project.py
+++ b/mlrun/projects/project.py
@@ -3980,57 +3980,63 @@ class MlrunProject(ModelObj):
         project_dir = pathlib.Path(project_file_path).parent
         project_dir.mkdir(parents=True, exist_ok=True)
 
-        original_function_definitions = self.spec._function_definitions
-        original_workflows = self.spec._workflows
-        working_function_definitions = original_function_definitions
-        working_workflows = original_workflows
-        staged_downloads: list[tuple[pathlib.Path, str]] = []
+        # On-disk project.yaml always carries the ORIGINAL store:// URIs,
+        # which resolve at load time. Pre-zip-export PRs already wrote the
+        # yaml here; rewriting it would leave the user's context dir holding
+        # paths that only resolve inside the zip, breaking
+        # load_project(<context>) afterwards.
+        with open(project_file_path, "w") as fp:
+            fp.write(self.to_yaml())
 
-        with contextlib.ExitStack() as stack:
-            if archive_code:
-                staging_dir = stack.enter_context(
-                    tempfile.TemporaryDirectory(prefix="mlrun-export-")
-                )
-                (
-                    working_function_definitions,
-                    working_workflows,
-                    staged_downloads,
-                    download_failures,
-                ) = _stage_store_downloads_for_export(
-                    spec=self.spec,
-                    staging_dir=staging_dir,
-                    project_name=self.metadata.name,
-                )
-                if download_failures:
-                    # Fail-fast: a partially-rewritten YAML would silently
-                    # ship store:// URIs that won't resolve on the destination
-                    # cluster, defeating the point of a self-contained zip.
-                    # Per-URI warnings were already logged by the helper.
-                    raise mlrun.errors.MLRunRuntimeError(
-                        f"Failed to download {len(download_failures)} "
-                        f"store:// reference(s) during zip export: "
-                        f"{download_failures}. See per-URI warning logs for "
-                        f"the specific error of each."
-                    )
+        if not archive_code:
+            return
 
-            # Install rewritten copies for to_yaml(), restore originals so
-            # callers see no mutation. No-op when archive_code=False.
+        with tempfile.TemporaryDirectory(prefix="mlrun-export-") as staging_dir:
+            (
+                working_function_definitions,
+                working_workflows,
+                staged_downloads,
+                download_failures,
+            ) = _stage_store_downloads_for_export(
+                spec=self.spec,
+                staging_dir=staging_dir,
+                project_name=self.metadata.name,
+            )
+            if download_failures:
+                # Fail-fast: a partially-rewritten YAML would silently ship
+                # store:// URIs that won't resolve on the destination
+                # cluster, defeating the point of a self-contained zip.
+                # Per-URI warnings were already logged by the helper.
+                raise mlrun.errors.MLRunRuntimeError(
+                    f"Failed to download {len(download_failures)} "
+                    f"store:// reference(s) during zip export: "
+                    f"{download_failures}. See per-URI warning logs for "
+                    f"the specific error of each."
+                )
+
+            # Capture rewritten YAML as a string (NOT to disk under the
+            # user's context). Writing it to <context>/project.yaml would
+            # leave behind a YAML whose .mlrun/code/... paths resolve only
+            # inside the zip, breaking subsequent load_project(<context>).
+            # The rewritten YAML is added to the zip directly by
+            # _build_export_zip via writestr("project.yaml", ...).
+            original_function_definitions = self.spec._function_definitions
+            original_workflows = self.spec._workflows
             try:
                 self.spec._function_definitions = working_function_definitions
                 self.spec._workflows = working_workflows
-                with open(project_file_path, "w") as fp:
-                    fp.write(self.to_yaml())
+                rewritten_yaml = self.to_yaml()
             finally:
                 self.spec._function_definitions = original_function_definitions
                 self.spec._workflows = original_workflows
 
-            if archive_code:
-                _build_export_zip(
-                    filepath=filepath,
-                    project_dir=project_dir,
-                    files_filter=include_files or "**",
-                    staged_downloads=staged_downloads,
-                )
+            _build_export_zip(
+                filepath=filepath,
+                project_dir=project_dir,
+                files_filter=include_files or "**",
+                staged_downloads=staged_downloads,
+                rewritten_yaml=rewritten_yaml,
+            )
 
     def set_model_monitoring_credentials(
         self,
@@ -6133,31 +6139,16 @@ def _download_store_artifact_for_export(
               inside ``staging_dir``; ``zip_arcname`` is the POSIX-style
               relative path that the caller embeds in the exported
               ``project.yaml`` AND uses as the arcname when adding the file
-              to the zip. Returns ``None`` on recoverable failure (artifact
-              missing, network/IO error).
-    :raises:  MLRunInvalidArgumentError if the resolved store:// URI does not
-              point to a code-bearing artifact (e.g., user passed a feature-
-              set URI as a function source).
+              to the zip. Returns ``None`` on any recoverable failure:
+              artifact missing, wrong artifact kind (not a CodeArtifact),
+              empty download body, or network/IO error. The caller logs the
+              specific error per-URI and aggregates all failures into a
+              single ``MLRunRuntimeError`` at end of the walk.
     """
     # Do not log target_path: for some datastores it may be a presigned URL
     # carrying credentials. store_uri is the public handle and is safe.
     try:
-        try:
-            artifact = mlrun.datastore.get_store_resource(
-                store_uri, project=project_name
-            )
-        except AttributeError as exc:
-            # Upstream get_store_resource calls .get(...) on the read_artifact
-            # result before its own None-check (store_resources.py:191), so a
-            # missing artifact surfaces as `'NoneType' object has no attribute
-            # 'get'`. Translate that one specific case to MLRunNotFoundError;
-            # any other AttributeError is a real bug or unrelated upstream
-            # failure — propagate so it's visible.
-            if "NoneType" not in str(exc):
-                raise
-            raise mlrun.errors.MLRunNotFoundError(
-                f"store:// artifact not found: {store_uri}"
-            ) from exc
+        artifact = mlrun.datastore.get_store_resource(store_uri, project=project_name)
         if artifact is None:
             raise mlrun.errors.MLRunNotFoundError(
                 f"store:// artifact not found: {store_uri}"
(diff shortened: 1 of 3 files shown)


mlrun/projects/project.py
```diff
@@ -6082,6 +6117,192 @@ def _set_as_current_active_project(project: MlrunProject):
...
+            # result before its own None-check (store_resources.py:191), so a
+            # missing artifact surfaces as `'NoneType' object has no attribute
+            # 'get'`. Translate that one specific case to MLRunNotFoundError;
+            # any other AttributeError is a real bug or unrelated upstream
+            # failure — propagate so it's visible.
+            if "NoneType" not in str(exc):
+                raise
+            raise mlrun.errors.MLRunNotFoundError(
+                f"store:// artifact not found: {store_uri}"
+            ) from exc
```
yaelgen: `store_resources.get_store_resource` calls `resource.get(...)` at line 191 before its own `None` check, which is what produces `NoneType` object has no attribute `get`. 
Coupling export correctness to that exact CPython error string is fragile: a different Python build, a wrapped exception, or a mock returning a non-None falsy value all break the translation silently and we end up raising instead of warn-and-skip. Can we fix it at the source instead by adding:
```
if not resource: 
    return None (or raise MLRunNotFoundError) 
```
in `store_resources.py` and drop this whole except AttributeError block?

mlrun/projects/project.py
```diff
@@ -6082,6 +6117,192 @@ def _set_as_current_active_project(project: MlrunProject):
...
+            # ModelArtifact, DatasetArtifact, FeatureSet, FeatureVector — all
+            # have store:// URIs but none should be used as a function or
+            # workflow source. Surface a typed error rather than silently
+            # download bytes that won't execute as Python.
+            raise mlrun.errors.MLRunInvalidArgumentError(
+                f"store:// URI {store_uri!r} resolved to {type(artifact).__name__}, "
+                f"expected CodeArtifact (function/workflow source must be a "
+                f"code artifact)"
+            )
+        target_path = artifact.get_target_path()
```
yaelgen, line 6171: `artifact.get_target_path()` can return `None`. `os.path.basename(None)` raises TypeError, which isn't in the recoverable-exception handler, it leaks instead of warn-and-skip. 
You can add:

```
target_path = artifact.get_target_path()
if not target_path:
    raise mlrun.errors.MLRunNotFoundError(
        f"store:// artifact has no target path: {store_uri}"
    )
```

mlrun/projects/project.py
```diff
@@ -6082,6 +6117,192 @@ def _set_as_current_active_project(project: MlrunProject):
...
+        filename = os.path.basename(target_path)
+        # Hash on target_path (not store_uri) so different URIs that resolve
+        # to the same backing file produce the same local filename — and so
+        # different files with the same basename get distinct names.
+        # 16 hex = 64 bits, birthday-safe up to ~4 billion artifacts.
+        path_hash = hashlib.sha256(target_path.encode()).hexdigest()[:16]
+        basename, ext = os.path.splitext(filename)
+        unique_filename = f"{basename}_{path_hash}{ext}"
+        real_path = pathlib.Path(staging_dir) / unique_filename
+        mlrun.get_dataitem(target_path).download(str(real_path))
```
yaelgen, line 6192: No check after download() that the file is non-empty. A backend returning HTTP 200 with an empty body ships a broken zip silently.

mlrun/projects/project.py
```diff
@@ -6082,6 +6117,192 @@ def _set_as_current_active_project(project: MlrunProject):
...
+            "Failed to download code artifact for export",
+            store_uri=store_uri,
+            project=project_name,
+            error=mlrun.errors.err_to_str(exc),
+        )
+        return None
+
+
+def _stage_store_downloads_for_export(
+    spec,
```
yaelgen:
```suggestion
    spec: ProjectSpec,
```

mlrun/projects/project.py
```diff
@@ -6082,6 +6117,192 @@ def _set_as_current_active_project(project: MlrunProject):
...
+    staging_dir: str,
+    project_name: str,
+) -> tuple[dict, dict, list[tuple[pathlib.Path, str]], list[str]]:
+    """Walk function and workflow definitions, download store:// sources into
+    ``staging_dir``, and return rewritten copies of the spec dicts plus the
+    bookkeeping the caller needs to finish the export.
+
+    Source-bearing fields handled (Pre-submit audit #1 — polymorphic-branch
```
yaelgen: `Pre-submit audit #1 ...` ?

mlrun/projects/project.py
```diff
@@ -3978,24 +3979,58 @@ def export(self, filepath=None, include_files: str | None = None):
...
+                        f"the specific error of each."
+                    )
+
+            # Install rewritten copies for to_yaml(), restore originals so
+            # callers see no mutation. No-op when archive_code=False.
+            try:
+                self.spec._function_definitions = working_function_definitions
+                self.spec._workflows = working_workflows
+                with open(project_file_path, "w") as fp:
+                    fp.write(self.to_yaml())
```
yaelgen: `project_file_path` is `<context>/project.yaml` when `archive_code=True`. This write puts the rewritten .mlrun/code/... paths on disk, the finally restores in-memory state but not the file. 
After export("proj.zip") the context dir's `project.yaml` has paths that only resolve inside the zip - `load_project(<context>)` will fail. Pre-PR the on-disk YAML kept `store://` URIs that resolved at load time, so this is a regression.

you can write the rewritten yaml to a temp file and add it to the zip as `arcname="project.yaml"`, skipping the `project_file_path` write entirely on the zip branch.

mlrun/projects/project.py
```diff
@@ -6082,6 +6117,192 @@ def _set_as_current_active_project(project: MlrunProject):
...
+        path_hash = hashlib.sha256(target_path.encode()).hexdigest()[:16]
+        basename, ext = os.path.splitext(filename)
+        unique_filename = f"{basename}_{path_hash}{ext}"
+        real_path = pathlib.Path(staging_dir) / unique_filename
+        mlrun.get_dataitem(target_path).download(str(real_path))
+        # POSIX-style separators in the YAML so the zip unpacks correctly
+        # on any OS the recipient uses.
+        arcname = f".mlrun/code/{unique_filename}"
+        return real_path, arcname
+    except mlrun.errors.MLRunInvalidArgumentError:
```
yaelgen: `MLRunInvalidArgumentError` (wrong artifact kind) propagates immediately and aborts the loop mid-iteration, while `MLRunNotFoundError`/`IO` errors are aggregated into failures after the full loop. If entry 4 of 10 has a ModelArtifact URI, entries 1-3 were downloaded for nothing, 5-10 are never validated, and the user gets a different exception type than they'd expect.

Either funnel MLRunInvalidArgumentError into failures too, or document in the docstring why type mismatches short-circuit while IO errors aggregate.

mlrun/projects/project.py
```diff
@@ -6082,6 +6117,192 @@ def _set_as_current_active_project(project: MlrunProject):
...
+              staged_downloads, failures)`` where the first two are deepcopies
+              of ``spec._function_definitions`` and ``spec._workflows`` with
+              ``store://`` URIs rewritten to local arcnames; ``staged_downloads``
+              is a list of ``(real_path_in_staging, zip_arcname)`` pairs to
+              write into the zip; ``failures`` lists names whose download
+              returned ``None`` (best-effort skip).
+    """
+    working_function_definitions = deepcopy(spec._function_definitions)
+    working_workflows = deepcopy(spec._workflows)
+    staged_downloads: list[tuple[pathlib.Path, str]] = []
```
yaelgen: `staged_downloads` as a list silently accumulates duplicates. `_set_function` stores both "my-func" and "my-func:v1" in `_function_definitions` for tagged functions - the walker hits both entries, downloads the same `store://` URI twice, and calls `zipf.write(same_arcname)` twice.
why not using `dict[arcname, real_path]` so the zip-write is idempotent?

yaelgen (changes requested):

mlrun/projects/project.py
```diff
@@ -6248,6 +6123,217 @@ def _set_as_current_active_project(project: MlrunProject):
...
+            "Failed to download code artifact for export",
+            store_uri=store_uri,
+            project=project_name,
+            error=mlrun.errors.err_to_str(exc),
+        )
+        return None
+
+
+def _stage_store_downloads_for_export(
+    spec: "ProjectSpec",
```
yaelgen: I think it can be unquoted since the class is defined earlier in the same module

mlrun/projects/project.py
```diff
@@ -6248,6 +6123,217 @@ def _set_as_current_active_project(project: MlrunProject):
...
+    staged_downloads: dict[str, pathlib.Path] = {}
+    failures: list[str] = []
+
+    for name, func_def in working_function_definitions.items():
+        store_uri = None
+        if isinstance(func_def, dict):
+            source = func_def.get("url", "")
+            if source and mlrun.datastore.is_store_uri(source):
+                store_uri = source
+        elif hasattr(func_def, "spec") and hasattr(func_def.spec, "build"):
```
yaelgen, line 6258: The BaseRuntime branch and its matching rewrite (func_def.spec.build.source = arcname) aren't exercised by any of the 5 new tests - diff-coverage flagged these lines for exactly that reason. All 5 tests call set_function(func=store_uri, ...), which stores a dict in _function_definitions, the runtime-object path is never hit.

That matters because a typo in the rewrite, a getattr quirk, or a deepcopy issue on a runtime instance (which has significantly more internal state than a dict) would silently ship a `store://` URI in the exported zip - the exact regression this PR is fixing, just on the parallel code path.

Suggested test:
```
rt = mlrun.code_to_function(name="rt-fn", kind="job", filename=<local_py>)
rt.spec.build.source = code_artifact.uri  # store:// URI
project.spec.functions = [rt]
project.export(zip_path)
# assert the in-zip project.yaml has .mlrun/code/... under spec.build.source
```

yaelgen (changes requested):
minor

mlrun/projects/project.py
alxtkr77: ok

mlrun/projects/project.py
alxtkr77: Test added. Test failed on first run — ProjectSpec.functions was silently rewriting the runtime's spec.build.source to "./" because  startswith("") matches everything when project source is unset. Dormant since 2020. Fixed in the same commit.

yaelgen (approved):
LGTM :)

